PR #304, opened 2026-03-08
Prevent shell CWD death when worktree is removed

### LINKED ISSUE #301: Shell CWD dies when worktree is removed during session
## Problem

When a git worktree is removed during a Claude Code session (e.g., after merging a PR and cleaning up), if the shell's persistent working directory was inside that worktree, every subsequent bash command fails with exit code 1.

## Reproduction

1. Create a worktree: `git worktree add .worktrees/my-feature session/my-feature`
2. Run bash commands that set CWD to the worktree directory
3. Remove the worktree: `git worktree remove .worktrees/my-feature`
4. Every subsequent bash command fails — the shell cannot recover because its CWD no longer exists

## Impact

- Session becomes partially broken — only non-bash tools (Read, Glob, Grep, WebFetch) continue to work
- No recovery is possible within the session without manual intervention
- This happened during the PR 300 merge/cleanup flow for issue #294

## Expected Behavior

After removing a worktree, the shell should either:
- Automatically reset CWD to the project root
- Or the worktree cleanup code should `cd` to a safe directory before removing

## Suggested Fix

In the worktree cleanup flow (wherever `git worktree remove` is called), ensure the shell CWD is changed to the project root *before* removing the worktree directory. This could be enforced in `agent/worktree_manager.py` or in the SDLC skill that orchestrates cleanup.

## Context

- Observed in session 6877 (PR 300 merge for issue #294)
- Worktree path: `.worktrees/surface_plan_open_questions/`

valorengels:
## Summary
- Add CWD guard to `remove_worktree()` — detects when process CWD is inside the worktree being removed and `os.chdir()`s to repo root first
- Add `os.chdir(REPO_ROOT)` to `post_merge_cleanup.py` as belt-and-suspenders defense
- Update SDLC, do-build SKILL.md and PR_AND_CLEANUP.md to explicitly `cd` to repo root before worktree cleanup commands

## Test plan
- [ ] 28 worktree manager unit tests passing
- [ ] Verify merge + cleanup flow doesn't kill shell CWD
- [ ] Verify cleanup works when CWD is already at repo root (no-op case)

Closes #301

### DIFF
--- a/.claude/skills/do-build/PR_AND_CLEANUP.md
+++ b/.claude/skills/do-build/PR_AND_CLEANUP.md
@@ -96,9 +96,12 @@ EOF
 
 ## Step 7.5: Worktree Cleanup
 
-After pushing and creating the PR, clean up the worktree using `worktree_manager.py` which handles CWD safely via `subprocess(cwd=repo_root)`:
+After pushing and creating the PR, return to the repo root and clean up the worktree. The `cd` prevents CWD death if the shell is inside the worktree (issue #301):
 
 ```bash
+# Return to repo root BEFORE cleanup (prevents CWD death)
+cd /Users/valorengels/src/ai
+
 python -c "
 from pathlib import Path
 from agent.worktree_manager import remove_worktree, prune_worktrees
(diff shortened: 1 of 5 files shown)


valorengels:
## Review: Approved

Defense-in-depth approach to prevent CWD death:

### Verified
- [x] `remove_worktree()` CWD guard correctly detects CWD inside worktree via path comparison
- [x] OSError catch handles the case where CWD is already invalid (recovery path)
- [x] `post_merge_cleanup.py` adds `os.chdir(REPO_ROOT)` at startup as additional safety layer
- [x] All three skill files (SDLC, do-build SKILL.md, PR_AND_CLEANUP.md) updated with explicit `cd` before cleanup
- [x] 28 worktree manager tests passing
- [x] No changes to worktree creation or checkout logic (per no-gos)

PR #590, opened 2026-03-30
Add periodic message reconciler for live bridge gaps

### LINKED ISSUE #588: Bridge misses messages during live connection — no runtime gap detection
> The Valor bridge uses [Telethon](https://docs.telethon.dev/) to receive Telegram messages via its `NewMessage` event handler. Messages are processed through a pipeline: event → dedup check → routing decision → job queue → agent response. The bridge runs with `catch_up=True` (replay missed messages on reconnect) and `sequential_updates=False` (concurrent event processing).

## Problem

**Current behavior:** The bridge silently misses messages during a live connection. Message 8198 in Agent Builders Chat (a reply-to-Valor message) was never delivered to the handler, despite the bridge being alive with consistent 30-second heartbeats throughout the window. No crash, no disconnect, no error, no dedup hit — the Telethon `NewMessage` event simply never fired.

**Desired outcome:** Every message sent in a monitored group is processed by the bridge handler. When a message is missed, the system detects and recovers it within a bounded time window (e.g., 5 minutes).

### Evidence

| Timestamp (UTC) | Event |
|-----------------|-------|
| 13:57:42 | Message 8195 (`@valorengels` mention from Kevin) → processed successfully |
| 13:58:45 | Session completed, bridge idle (workers=0) |
| 14:10:48 | Heartbeat: alive, workers=0 |
| **14:11:39** | **Message 8198 sent by Kevin (reply to 8197)** — never appears in logs |
| 14:12:18 | Heartbeat: alive, workers=0 |

- Message 8198 is a reply to message 8197, which is Valor's own message (`out=True`)
- The reply-to-Valor check in `should_respond_async()` runs before Q&A mode filtering — this message should have matched
- Zero trace of message 8198 in `bridge.log` or `bridge.log.1`

### Root Cause

Telethon can silently drop updates when the Telegram server delivers them out of sequence or the client misses a `pts` (persistent timeline sequence) gap. The current config compounds this:

- **`catch_up=True`** only recovers messages on disconnect/reconnect — not during a live connection with invisible gaps
- **`sequential_updates=False`** disables Telethon's internal gap detection, which means pts gaps go unnoticed
- **No periodic reconciliation** exists after startup — the bridge trusts Telethon's event delivery completely at runtime

### Definitions

|------|-----------|--------|
| `catch_up` | Telethon client option that replays missed messages when reconnecting | [Telethon docs](https://docs.telethon.dev/en/stable/modules/client.html) |
| `sequential_updates` | Telethon option that enforces update ordering and gap detection | [Telethon docs](https://docs.telethon.dev/en/stable/modules/client.html) |
| `pts` | Persistent Timeline Sequence — Telegram's monotonic counter for ordering updates per chat | [Telegram MTProto docs](https://core.telegram.org/api/updates) |
| Dedup | Per-chat duplicate detection via Redis (`DedupRecord` model, 2h TTL, 50 msg capacity) | [`models/dedup.py`](https://github.com/tomcounsell/ai/blob/main/models/dedup.py) |
| Startup catchup | One-time scan at bridge start that fetches recent messages from monitored groups | [`bridge/catchup.py`](https://github.com/tomcounsell/ai/blob/main/bridge/catchup.py) |

### Existing Reliability Mechanisms (and their gaps)

| Mechanism | When it runs | What it catches | Gap |
|-----------|-------------|----------------|-----|
| `catch_up=True` | On reconnect only | Messages missed during disconnect | Silent drops during live connection |
| Startup catchup scan | Bridge start only | Messages missed during downtime | Nothing after startup |
| Dedup check | Every handler call | Prevents re-processing | Can't detect messages that never arrived |
| Session watchdog | Every 5 min | Stalled SDK sessions | Doesn't check for missed messages |

## Solution Sketch

Add a **periodic message reconciliation loop** that runs every few minutes during the bridge's lifetime:

1. For each monitored group, fetch the last N message IDs via `client.get_messages()`
2. Compare against the dedup record (messages we've already processed)
3. For any unprocessed messages within the lookback window, re-fetch and dispatch through the normal handler pipeline
4. Log reconciliation results for observability

### Options to evaluate

| Option | Pros | Cons |
|--------|------|------|
| **Periodic polling reconciler** | Non-invasive, catches all gap types, configurable interval | Adds API calls, slight latency for recovery |
| **`sequential_updates=True`** | Native gap detection, no extra API calls | May reduce throughput, Telethon handles recovery opaquely |
| **Raw update handler for pts gaps** | Precise gap detection | Complex, tightly coupled to Telethon internals |

The periodic polling approach is recommended — it's the most robust and least coupled to Telethon internals.

### Key files

- `bridge/telegram_bridge.py` — Handler registration, client config, background loops
- `bridge/catchup.py` — Startup gap scanner (reusable pattern)
- `bridge/dedup.py` — Dedup interface
- `models/dedup.py` — DedupRecord model
- `bridge/routing.py` — `should_respond_async()` routing logic

## Acceptance Criteria

- [ ] Missed messages in monitored groups are detected and processed within a bounded window
- [ ] Reconciliation loop runs periodically without impacting normal message handling
- [ ] Observability: reconciliation results are logged (messages found, messages recovered)
- [ ] No duplicate processing (dedup check must gate reconciled messages)
- [ ] Integration test covering the gap detection → recovery path

## Downstream Context

valorengels:
## Summary
- Adds `bridge/reconciler.py` with a periodic reconciliation loop that detects and recovers Telegram messages missed during live connections due to Telethon pts gaps
- Reconciler runs every 3 minutes with a 10-minute lookback, scanning 20 messages per monitored group
- Gates all re-dispatches through existing dedup to prevent duplicate processing
- Registered as a background task in `bridge/telegram_bridge.py` alongside other loops

Closes #588

## Changes
- `bridge/reconciler.py` — new module with `reconciler_loop()` and `reconcile_once()`
- `bridge/telegram_bridge.py` — register reconciler as background task after catchup scan
- `tests/unit/test_reconciler.py` — 12 unit tests covering all skip conditions and recovery
- `tests/integration/test_reconciler.py` — 2 integration tests for gap detection and idempotency

## Test plan
- [x] Unit tests pass: `pytest tests/unit/test_reconciler.py -x -q` (12 passed)
- [x] Integration tests pass: `pytest tests/integration/test_reconciler.py -x -q` (2 passed)
- [x] Ruff lint and format clean
- [x] Module importable: `from bridge.reconciler import reconciler_loop, reconcile_once`
- [ ] Verify reconciler starts with bridge in production logs (`[reconciler] Started`)

### DIFF
--- a/docs/features/README.md
+++ b/docs/features/README.md
@@ -54,6 +54,7 @@ Completed feature documentation for the Valor AI system. Each document describes
 | [Lint Auto-Fix](lint-auto-fix.md) | Automatic lint/format fixing via pre-commit hook and PostToolUse hook, eliminating agent churn loops | Shipped |
 | [Memory Search Tool](memory-search-tool.md) | Direct interface for searching, saving, inspecting, and forgetting memories from the Memory model | Shipped |
 | [Message Pipeline](message-pipeline.md) | Deferred enrichment pipeline for fast message acknowledgment and zero-loss restarts | Shipped |
+| [Message Reconciler](message-reconciler.md) | Periodic background scan detecting and recovering Telegram messages missed during live bridge connections | Shipped |
 | [Mid-Session Steering](mid-session-steering.md) | End-to-end steering flow for injecting reply-to messages into running agent sessions | Shipped |
 | [Observer Agent](observer-agent.md) | Deterministic SDLC steer/deliver router — no LLM calls, rules-based pipeline progression | Shipped |
 | [OOP Audit](do-oop-audit.md) | Prompt-only audit skill scanning Python classes for 14 structural anti-patterns with framework detection and severity grouping | Shipped |
--- a/docs/features/bridge-module-architecture.md
+++ b/docs/features/bridge-module-architecture.md
@@ -9,6 +9,7 @@ The Telegram bridge (`bridge/telegram_bridge.py`) is organized into focused sub-
 | `bridge/context.py` | Context building, conversation history, reply chains |
 | `bridge/response.py` | Message formatting, reactions, file extraction, sending |
 | `bridge/catchup.py` | Abandoned session revival and re-enqueueing |
+| `bridge/reconciler.py` | Periodic scan for messages missed during live connection |
 
 ## telegram_bridge.py
 
--- a/docs/features/bridge-self-healing.md
+++ b/docs/features/bridge-self-healing.md
@@ -250,4 +250,5 @@ rm data/auto-revert-enabled
 ## Related
 
 - [Message Pipeline](message-pipeline.md) — deferred enrichment and zero-loss restart mechanisms
+- [Message Reconciler](message-reconciler.md) — periodic scan for messages missed during live connection (complements startup catchup)
 - [Session Transcripts](session-transcripts.md) — session lifecycle logging via AgentSession model
--- a/docs/features/message-reconciler.md
+++ b/docs/features/message-reconciler.md
@@ -1,0 +1,104 @@
+# Message Reconciler
+
+Periodic background loop that detects and recovers Telegram messages missed during a live bridge connection.
+
+## Problem
+
+Telethon can silently drop updates when the Telegram server delivers them out of sequence or the client misses a `pts` (persistent timeline sequence) gap. Existing reliability mechanisms only cover restart and startup scenarios:
+
+- `catch_up=True` replays on reconnect
+- `bridge/catchup.py` scans once at boot
+- Dedup checks prevent re-processing but cannot detect messages that never arrived
+
+The reconciler fills the gap by scanning continuously while the bridge is alive.
+
+## How It Works
+
+The reconciler runs as an `asyncio.create_task` background loop inside the bridge, alongside the heartbeat and session watchdog.
+
+### Scan Cycle
+
+Every 3 minutes (configurable via `RECONCILE_INTERVAL_SECONDS`):
+
+1. Fetches recent messages from each monitored group via `client.get_messages()`
+2. Filters to messages within the lookback window (default 10 minutes)
+3. Checks each message against dedup records (`is_duplicate_message()`)
+4. Skips outgoing messages, empty-text messages, and messages that fail routing (`should_respond_async()`)
+5. Enqueues qualifying missed messages via `enqueue_job()` with `priority="low"`
+6. Records dispatched messages in dedup to prevent future re-dispatch
+
+### Data Flow
+
+```
+reconciler_loop (every 3min)
+    |
+    +-- for each monitored group:
+    |       get_messages(limit=20)
+    |       for each message:
+    |           outside lookback window? --> stop scanning group
+    |           is outgoing? --> skip
+    |           no text? --> skip
+    |           is_duplicate? --> skip
+    |           should_respond? no --> skip
+    |           enqueue_job(priority="low")
+    |           record_message_processed()
+    |
+    +-- log summary: "Scanned N group(s), recovered M message(s)"
+```
+
+## Configuration
+
+| Constant | Default | Purpose |
+|----------|---------|---------|
+| `RECONCILE_INTERVAL_SECONDS` | 180 (3 min) | Time between scans |
+| `RECONCILE_LOOKBACK_MINUTES` | 10 | How far back each scan looks |
+| `RECONCILE_MESSAGE_LIMIT` | 20 | Max messages fetched per group per scan |
+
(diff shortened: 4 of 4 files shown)


valorengels (commented):
## PR Review: Add periodic message reconciler for live bridge gaps

### Verdict: APPROVE (with minor observations)

This PR delivers a clean, well-scoped implementation that matches the plan document closely. The reconciler is a natural complement to the startup catchup scanner, filling the gap for messages missed during live operation.

---

### Plan Compliance

The implementation matches the plan on all key points:
- 3-minute interval, 10-minute lookback, 20-message limit per group
- Uses existing `is_duplicate_message()`, `record_message_processed()`, `should_respond_async()`, `enqueue_job()`
- Registered as `asyncio.create_task()` in `telegram_bridge.py` after catchup scan
- Per-group error isolation (one group failing does not stop others)
- Low priority for recovered messages

The plan's optional refactor of shared scanning logic from `catchup.py` was correctly skipped -- the two scanners have different enough behavior to justify separate implementations.

One plan item not implemented: the `_check_if_handled()` secondary safety net from `catchup.py` (which checks if we already replied to the message). The reconciler relies solely on dedup, which is sufficient given the short lookback window, but this is worth noting.

---

### Code Quality

**Strengths:**
- Clean separation of `reconciler_loop()` (infinite loop with error handling) and `reconcile_once()` (single scan, testable)
- Correct log levels: DEBUG for normal scans, WARNING for recovered messages, ERROR for failures
- Proper async patterns throughout
- `MinimalEvent` adapter pattern reused from `catchup.py` keeps the reconciler compatible with `should_respond_async()` without importing internal types
- Module-level constants are well-named and well-documented

**Minor observations (non-blocking):**

1. **`MinimalEvent` class defined inside loop body (line 131-134):** The class is redefined on every message iteration. This is functionally harmless but slightly wasteful. Could be moved to module level or outside the loop. The same pattern exists in `catchup.py` (line 152), so this is consistent with the codebase.

2. **`message.date` timezone comparison (line 108):** The cutoff uses `datetime.now(UTC)` and compares against `message.date`. Telethon returns dates as UTC-aware datetimes, so this is correct. No issue here.

3. **No `_check_if_handled()` call:** As noted above, `catchup.py` has a secondary check that looks for our reply to a message before re-dispatching. The reconciler skips this. Given the 10-minute lookback and dedup gating, this is a reasonable simplification -- the dedup record is written by the live handler before processing begins, so if a message was handled, it will be in dedup. The `_check_if_handled()` check in catchup is more important because of the longer 60-minute lookback where dedup TTL overlap is less guaranteed.

---

### Correctness

- **Dedup gating is correct:** `is_duplicate_message()` is checked before enqueue, and `record_message_processed()` is called after enqueue. This matches the catchup pattern.
- **Race condition handling:** The plan correctly identifies the race between the live handler and reconciler. Both could see a message before either records it in dedup. The mitigation is sound -- `enqueue_job` handles duplicate session IDs gracefully since the session ID format (`tg_{key}_{chat}_{msg}`) is deterministic.
- **Lookback break (line 108-109):** The `break` on old messages assumes Telegram returns messages in reverse chronological order (newest first), which is correct for `client.get_messages()`.
- **Error isolation (line 181-183):** Per-group try/except ensures one failing group does not prevent scanning others. Good.

---

### Security

- No credentials or secrets exposed
- No new external dependencies
- No user-controlled input reaches dangerous operations
- Message text is only logged truncated to 80 chars (line 161)

---

### Test Coverage

**Unit tests (12):** Comprehensive coverage of all skip conditions:
- Empty monitored groups, dedup hits, outgoing messages, empty text, whitespace-only text, routing rejection, old messages outside lookback, no project config, non-monitored groups, per-group error isolation, multiple missed messages, and the happy path with enqueue verification.

**Integration tests (2):** Cover the end-to-end gap detection flow (messages 1-5 in dedup, 6-7 missed) and mixed message types in gaps. The second-scan idempotency check in the integration test is a good touch.

All tests properly mock the dedup layer and avoid touching Redis. Test helpers (`_make_message`, `_make_dialog`) are well-structured.

---

### Bridge Integration

The registration in `telegram_bridge.py` (lines 1771-1787) follows the established pattern used by the session watchdog and other background tasks. The import is inside the try block, and failure to start the reconciler does not prevent the bridge from running.

One note: the reconciler imports `enqueue_job` as `_reconciler_enqueue` (line 1773), which is a separate import from the `_enqueue_job` used by the catchup scanner (visible earlier in the file). Both resolve to the same function, so this is fine functionally, but the naming difference is slightly confusing.

---

### Summary

This is a solid, well-tested addition that addresses a real production gap (issue #588). The implementation is conservative in its defaults, properly isolated from failures, and consistent with existing codebase patterns. No blockers found.

**Status: APPROVE**

PR #739, opened 2026-04-06
fix: session lifecycle stale cleanup and state corruption (#738)

### LINKED ISSUE #738: fix: stale session cleanup kills live sessions and corrupts state on forced termination
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/session_lifecycle_stale_cleanup.md

## Context

> **AgentSession** is the Popoto Redis model that represents a single unit of work processed by the Claude Agent SDK worker. It carries all state: `session_id`, `status`, `stage_states`, `session_events`, etc.
>
> **`_cleanup_stale_sessions`** runs in `scripts/update/run.py:145–203` during every `/update` deploy. It scans all `running` and `pending` sessions and transitions those older than 30 minutes — with no live process — to `killed`. It uses Popoto's delete-and-recreate pattern because `status` is a KeyField.
>
> **Popoto delete-and-recreate** is the required mutation pattern for indexed/key fields in Popoto. The code calls `_extract_agent_session_fields(s)` to snapshot the session dict, deletes the record, then calls `AgentSession.create(**fields)` with the mutated field. The new record gets a fresh auto-generated `id` but preserves all other fields.
>
> **Worker cancellation path** is the `asyncio.CancelledError` branch in `_worker_loop` (`agent/agent_session_queue.py:1525–1537`). When the asyncio task is cancelled — typically by a bridge restart — the worker logs the failure, calls `_complete_agent_session(session, failed=True)`, sets `session_completed = True`, and re-raises. The in-memory `session` object is used to write back to Redis.
>
> **`stage_states`** is a computed property on `AgentSession` backed by `session_events` (a `ListField`). The setter appends a new stage-change event to `session_events`. Values are set by `PipelineStateMachine` during SDLC execution. When a session is delete-and-recreated by `_cleanup_stale_sessions`, the `session_events` list is preserved — so `stage_states` survives. But when a session is killed mid-run via worker cancellation, the in-memory `session` object may lag behind the Redis record if a concurrent delete-and-recreate occurred.

## Definitions

| Term | Definition |
|---|---|
| **AgentSession** | Popoto Redis model for a worker task (`models/agent_session.py`). Single record keyed by `id` (auto-generated UUID). Fields include `status` (IndexedField), `session_events` (ListField), `stage_states` (computed property). |
| **`_cleanup_stale_sessions`** | Function in `scripts/update/run.py:145` called during `/update`. Kills `running`/`pending` sessions >30 min old with no live PID. Uses delete-and-recreate because `status` is a KeyField. |
| **Popoto delete-and-recreate** | Required pattern for mutating indexed/key fields: call `_extract_agent_session_fields(s)`, `s.delete()`, then `AgentSession.create(**fields)`. Produces a **new Redis record with a new `id`** — the old `id` is gone. |
| **Worker cancellation path** | The `asyncio.CancelledError` handler in `_worker_loop` (line 1525). Fires when the asyncio task is cancelled externally (bridge restart). Calls `_complete_agent_session` on the in-memory session object. |
| **`stage_states`** | Computed property derived from `session_events`. Tracks SDLC pipeline progress (e.g. `{"ISSUE": "ready", "PLAN": "in_progress", ...}`). Stored as events in `session_events` ListField. |

### Bug 1 — `_cleanup_stale_sessions` kills live sessions because it has no PID to check

**Current behavior:** `_cleanup_stale_sessions` checks `getattr(s, "pid", None)` to decide if a process is still alive. `AgentSession` has **no `pid` field** — there is no PID stored anywhere on the model. `getattr` returns `None` for every session. The live-process guard (`os.kill(pid, 0)`) is therefore **never reached**. Any `running` or `pending` session older than 30 minutes is killed unconditionally.

**Desired behavior:** Sessions with an actively running worker should be skipped, regardless of age. Either the PID is stored on the session and checked, or sessions are checked against the live `_active_workers` registry, or the age threshold is raised significantly (e.g. 2 hours) so that only truly orphaned sessions — those whose process died before the bridge could mark them terminal — are affected.

**Root cause:** `scripts/update/run.py:187` — `pid = getattr(s, "pid", None)` always returns `None`. The guard at line 189 is dead code.

---

### Bug 2 — Worker cancellation corrupts `stage_states` by writing a stale in-memory snapshot

**Current behavior:** When a bridge restart cancels the asyncio worker task, `_worker_loop` catches `asyncio.CancelledError` and calls `_complete_agent_session(session, failed=True)`. This writes the **in-memory `session` object** back to Redis. If `_cleanup_stale_sessions` ran during the session's lifetime and delete-and-recreated the Redis record (assigning it a new `id`), the in-memory `session` object still holds the **old `id`**. The `finalize_session()` call then saves to the old key, which no longer exists in Redis, silently losing the update — or it saves correctly but overwrites the newer record, depending on timing.

More critically: the in-memory `session` object may be missing `session_events` appended by the SDK run during execution (SDLC stage transitions). After cancellation, `_complete_agent_session` calls `finalize_session` which calls `session.save()` — saving the in-memory snapshot including whatever `session_events` the object holds at cancellation time. Any events written to Redis between session start and cancellation that are not in the in-memory copy are lost.

**Desired behavior:** Before writing terminal state, re-read the current record from Redis to get the latest `session_events` (including SDLC stage transitions), merge the terminal status, and save.

---

### Bug 3 — `_cleanup_stale_sessions` does not use `finalize_session`, bypassing lifecycle side effects

**Current behavior:** `_cleanup_stale_sessions` performs a raw delete-and-recreate to set `status="killed"` (`scripts/update/run.py:197–200`). This bypasses:
- `session.log_lifecycle_transition("killed", ...)`
- `auto_tag_session()`
- `checkpoint_branch_state()`
- Parent session finalization

**Desired behavior:** Use `finalize_session(s, "killed", reason="stale cleanup")` which routes through the lifecycle module and executes all side effects. If delete-and-recreate is still required for the KeyField constraint, it should happen **inside** a wrapper that also fires the lifecycle hooks on the resulting record.

## Real Evidence — Session `tg_valor_-1003449100931_441` (SDLC issue #734)

Two Redis records exist for this session, revealing the Bug 1 + Bug 2 failure chain:

| Record | `id` | Created | Completed | Status | `stage_states` (summary) |
|---|---|---|---|---|---|
| **Killed by `_cleanup_stale_sessions`** | `873411799eca` | 06:11:48 | 06:23:18 | `killed` | PLAN=completed, REVIEW=completed, DOCS=completed, MERGE=ready |
| **Overwrite from worker cancellation** | `9a4823ca2db4` | 06:11:48 | 06:38:16 | `failed` | ISSUE=ready, all others=pending |

**Timeline:**
- `06:11:48` — Session created and immediately starts running (lifecycle log: `pending→running id=9a4823ca2db4`)
- `06:23:18` — `_cleanup_stale_sessions` kills the session via delete-and-recreate, producing `873411799eca` with status=`killed`
- `06:37` — Session still healthy at tool call ~200 (the bridge worker was never stopped — the asyncio task continued executing against the old `9a4823ca2db4` in-memory object)
- `06:38:16` — Bridge restart; `asyncio.CancelledError` fires; worker writes `failed` status using the stale in-memory object (`9a4823ca2db4`), overwriting `stage_states` back to initial state

**Key observation:** The `killed` record (`873411799eca`) has rich `stage_states` reflecting actual SDLC progress — `PLAN=completed`, `REVIEW=completed`, `DOCS=completed`, `MERGE=ready`. The `failed` record (`9a4823ca2db4`) — written by the worker cancellation path — has `stage_states` reset to all-pending with only `ISSUE=ready`, meaning the in-memory session object at cancellation time held the **initial** stage_states snapshot from session creation, not the states accumulated during 26 minutes of SDK execution.

### Bug 1 — Add heartbeat or registry check before killing

Add a cross-reference against the live `_active_workers` registry. The update script runs in-process with the bridge, so the registry is accessible. Alternatively, store a `heartbeat_at` timestamp on the session and skip sessions whose heartbeat is <5 minutes old:

```python
# In _cleanup_stale_sessions, before the kill:
from agent.agent_session_queue import _active_workers

chat_id = getattr(s, "chat_id", None)
if chat_id and chat_id in _active_workers:
    continue  # live worker exists for this chat — skip
```

If the update script runs in a subprocess where `_active_workers` is not accessible, raise the default `age_minutes` from 30 to 120, and document this function as a last-resort fallback only.

### Bug 2 — Re-read from Redis before finalizing

Before calling `finalize_session`, refresh the session object to capture `session_events` written during execution:

```python
async def _complete_agent_session(session: AgentSession, *, failed: bool = False) -> None:
    # Re-read to capture SDLC stage events written during execution
    fresh = list(AgentSession.query.filter(session_id=session.session_id, status="running"))
    if fresh:
        session = fresh[0]
    status = "failed" if failed else "completed"
    finalize_session(session, status, reason="agent session completed")
```

### Bug 3 — Route through `finalize_session`

Replace the raw delete-and-recreate in `_cleanup_stale_sessions` with a `finalize_session` call. The lifecycle module already handles the Popoto save pattern:

```python
# Instead of manual delete-and-recreate:
from models.session_lifecycle import finalize_session
finalize_session(s, "killed", reason="stale cleanup (no live process)", skip_checkpoint=True)
```

If `finalize_session` cannot handle the KeyField constraint, it should be extended to accept a `force_recreate=True` flag that internally does the delete-and-recreate while still firing all lifecycle hooks.

## Acceptance Criteria

- [ ] `_cleanup_stale_sessions` skips sessions that have a live worker (checked via heartbeat or `_active_workers` registry)
- [ ] `_cleanup_stale_sessions` calls `finalize_session("killed", ...)` instead of raw delete-and-recreate, so lifecycle hooks fire
- [ ] Worker cancellation path (`asyncio.CancelledError`) re-reads the session from Redis before finalizing, to capture SDLC stage events written during execution
- [ ] After a bridge restart, the surviving Redis record for an interrupted session has `stage_states` reflecting actual progress, not the initial snapshot
- [ ] No duplicate Redis records exist for a single logical session after a stale-cleanup + cancellation sequence
- [ ] Unit test: `_cleanup_stale_sessions` does not kill a session whose `heartbeat_at` is <5 minutes ago (or that has a live `_active_workers` entry)
- [ ] Unit test: worker cancellation path preserves `stage_states` from the most recent Redis record
- [ ] Integration test: run a 35-minute session through a simulated bridge restart; verify final `stage_states` reflects work done before restart

## Recon Summary

**Root cause confirmed by code inspection and live session evidence (session `tg_valor_-1003449100931_441`):**

1. `scripts/update/run.py:187` — `getattr(s, 'pid', None)` always returns `None` because `AgentSession` has no `pid` field. The live-process guard is dead code; all sessions older than 30 min are killed unconditionally.

2. `agent/agent_session_queue.py:1525–1537` — Worker cancellation writes the in-memory session snapshot back to Redis without re-reading from Redis first. Concurrent `_cleanup_stale_sessions` may have already delete-and-recreated the record (new `id`), so the write either targets a stale key or overwrites a newer record with stale `stage_states`.

3. `scripts/update/run.py:197–200` — Raw delete-and-recreate to transition status bypasses all lifecycle hooks in `models/session_lifecycle.py` (log_lifecycle_transition, auto_tag_session, checkpoint_branch_state, parent finalization).

**Files under change:**
- `scripts/update/run.py` — `_cleanup_stale_sessions` function (lines 145–203)
- `agent/agent_session_queue.py` — `_worker_loop` CancelledError handler (lines 1525–1537) and `_complete_agent_session`
- `models/session_lifecycle.py` — `finalize_session` (may need new `skip_checkpoint` param)
- `models/agent_session.py` — Confirm no `heartbeat_at` / `pid` fields exist today

**No prior fix attempts found** — searched closed issues and merged PRs for 'stale session', 'cleanup kills', 'stage_states corrupt', found no related PRs.

valorengels:
## Summary

- **Bug 1**: `_cleanup_stale_sessions` now consults `_active_workers` registry before killing sessions — skips any session whose `chat_id` has a live (not-done) asyncio Task. Age threshold raised 30→120 minutes to reduce false positives.
- **Bug 2**: `_complete_agent_session` re-reads the session from Redis by `session_id` before finalizing, capturing accumulated `stage_states` that were written during execution and would otherwise be lost from the stale in-memory snapshot.
- **Bug 3**: `_cleanup_stale_sessions` now routes all terminal transitions through `finalize_session()` (lifecycle hooks, auto-tag, parent finalization) instead of raw delete-and-recreate.

## Test plan

- [x] `tests/unit/test_stale_cleanup.py` — 6 tests: live-worker skip, done-worker kill, no-chat-id fallthrough, finalize routing, exception resilience, age threshold
- [x] `tests/unit/test_complete_agent_session_redis_reread.py` — 7 tests: fresh-record selection, fallback on empty/error, multi-record tie-breaking, None session_id guard, status mapping
- [x] Full unit suite: 2141 passed, 1 pre-existing failure (`test_sdlc_stage_query::test_no_args_returns_empty_json`, broken before this branch)

Closes #738


valorengels:
Docs updated in commit e22ff222: added `_cleanup_stale_sessions` to the completion paths table and added a new **Stale Session Cleanup** section documenting the `_active_workers` liveness check, 120-minute age threshold, `finalize_session` routing with `skip_checkpoint=True`, and in-process vs standalone subprocess behavior. Also added a **Worker Completion — Redis Re-read** subsection for the Bug 2 fix.

PR #763, opened 2026-04-07
Fix _pop_agent_session tests and extraction helper docs (#761)

### LINKED ISSUE #761: Fix _pop_agent_session tests and _extract_agent_session_fields missing fields
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/fix-pop-agent-session-tests.md

> **Context:** The [worker service](https://github.com/tomcounsell/ai/blob/main/worker/__main__.py) processes queued agent sessions from Redis. When the worker picks up a pending session, it calls [`_pop_agent_session()`](https://github.com/tomcounsell/ai/blob/main/agent/agent_session_queue.py#L462) to transition it from `pending` to `running`. This function was recently refactored from a delete-and-recreate pattern to direct field mutation (using [`transition_status()`](https://github.com/tomcounsell/ai/blob/main/models/session_lifecycle.py#L131)), but the integration tests and helper function `_extract_agent_session_fields` were not updated to match the new behavior, and the helper itself is missing fields needed by callers that still use delete-and-recreate.

## Problem

There are two related bugs in the agent session queue module ([`agent/agent_session_queue.py`](https://github.com/tomcounsell/ai/blob/main/agent/agent_session_queue.py)):

### Bug 1: Tests expect delete-and-recreate, but `_pop_agent_session` now mutates in-place

`_pop_agent_session()` was refactored to use direct field mutation via `transition_status()` (line 524) instead of the old delete-and-recreate pattern. The docstring explicitly states: *"Status is an IndexedField (not KeyField), so mutating and saving is safe — no delete-and-recreate needed for status transitions."*

However, the integration tests still assert the old behavior — for example, `test_pop_agent_session_preserves_fields` checks that `session.agent_session_id != original.agent_session_id`, which only holds true if a new record was created. With in-place mutation, the ID stays the same. This mismatch causes tests to fail with assertions or receive `None` when they expect a new record.

**Current behavior:** ~13 integration tests fail. Tests that query for the popped session by its old ID or expect a new ID get `None` or assertion errors. Redis logs show "one or more redis keys points to missing objects" in some paths where delete-and-recreate is still used with incomplete field extraction.

**Desired outcome:** All integration tests pass. `_pop_agent_session()` behavior and its tests are aligned. `_extract_agent_session_fields` preserves all data needed by callers that still use delete-and-recreate (retry, orphan fix, continuation fallback).

### Bug 2: `_extract_agent_session_fields` is missing virtual fields

The [`_AGENT_SESSION_FIELDS`](https://github.com/tomcounsell/ai/blob/main/agent/agent_session_queue.py#L73) list does not include `message_text` or `scheduling_depth`. These are virtual properties on `AgentSession` — `message_text` is stored inside the `initial_telegram_message` DictField, and `scheduling_depth` is computed by walking the parent chain. The extraction helper reads fields by name via `getattr()`, so these virtual properties would work, but they are simply not listed.

This means callers that use `_extract_agent_session_fields` for delete-and-recreate (retry at line 762, orphan fix at line 1219, continuation fallback at line 1916) lose `message_text` data. The continuation fallback at line 1919-1926 works around this by manually patching `initial_telegram_message`, but the retry path does not.

## Definitions

| `AgentSession` | A Popoto ORM model representing a queued or running agent work item in Redis, keyed by `agent_session_id` (AutoKeyField) with `status` as an IndexedField | [`models/agent_session.py`](https://github.com/tomcounsell/ai/blob/main/models/agent_session.py) |
| `_pop_agent_session()` | Async function that finds the highest-priority pending session for a chat and transitions it to `running` so the worker can execute it | [`agent/agent_session_queue.py` L462](https://github.com/tomcounsell/ai/blob/main/agent/agent_session_queue.py#L462) |
| `_extract_agent_session_fields()` | Helper that copies field values from an `AgentSession` instance into a dict, used when delete-and-recreate is needed (e.g., changing a KeyField like `parent_agent_session_id`) | [`agent/agent_session_queue.py` L126](https://github.com/tomcounsell/ai/blob/main/agent/agent_session_queue.py#L126) |
| `transition_status()` | Lifecycle function that sets a session's status to a non-terminal value and calls `.save()`, with guards against invalid transitions | [`models/session_lifecycle.py` L131](https://github.com/tomcounsell/ai/blob/main/models/session_lifecycle.py#L131) |
| Popoto ORM | Redis-backed ORM used for all model persistence. KeyFields affect the Redis key, so changing them requires delete-and-recreate. IndexedFields use secondary indexes and can be mutated in-place. | [`docs/features/popoto-index-hygiene.md`](https://github.com/tomcounsell/ai/blob/main/docs/features/popoto-index-hygiene.md) |

## Prior Context

Related work:
- #700: Completed sessions revert to pending, causing infinite execution loop — Fixed zombie session bug, led to lifecycle refactoring
- #716: Follow-up quality stages for zombie fix — Addressed test gaps after #700

## Solution Sketch

1. **Update tests to match current `_pop_agent_session` behavior**: The 5 test files need to assert in-place mutation semantics (same `agent_session_id`, status changed to `running`) instead of delete-and-recreate semantics (new `agent_session_id`).

2. **Fix `_extract_agent_session_fields` for remaining delete-and-recreate callers**: Either add the missing virtual properties to `_AGENT_SESSION_FIELDS` or ensure callers that still use delete-and-recreate (retry, orphan fix, continuation) explicitly handle `initial_telegram_message` to preserve `message_text`. The continuation fallback already does this manually — the retry path does not.

3. **Update test docstrings and module docstrings**: Several test files reference "delete-and-recreate pattern used by `_pop_agent_session`" which is no longer accurate.

## Recon Summary

**Confirmed** 3 items — ready to include
- `_pop_agent_session` uses `transition_status()` for in-place mutation (verified at line 524)
- `_AGENT_SESSION_FIELDS` list is missing `message_text` and `scheduling_depth` (verified at lines 73-123)
- Three callers still use `_extract_agent_session_fields` with delete-and-recreate: retry (L762), orphan fix (L1219), continuation fallback (L1916)

**Revised** 1 item — scope adjusted
- The bug is NOT that `_pop_agent_session` is broken in production — it correctly uses in-place mutation. The bug is that tests assert the old behavior, and `_extract_agent_session_fields` is incomplete for callers that still need delete-and-recreate.

**Pre-requisites** 0 items

**Dropped** 1 item — removed from scope
- `scheduling_depth` is a computed property (walks parent chain) and is intentionally excluded from `_push_agent_session` kwargs (marked "ignored, now derived"). It does NOT need to be in `_AGENT_SESSION_FIELDS` since it cannot be set on create. Tests asserting `scheduling_depth` in extracted fields should be removed or adjusted.

## Acceptance Criteria

- [ ] `_pop_agent_session()` integration tests pass — all 13 currently-failing tests in `test_agent_session_queue_race.py`, `test_agent_session_scheduler.py`, `test_lifecycle_transition.py`, `test_agent_session_health_monitor.py`, and `test_silent_failures.py`
- [ ] Tests assert current in-place mutation behavior (same `agent_session_id` after pop, status changed to `running`)
- [ ] `_extract_agent_session_fields` preserves `message_text` for callers that use delete-and-recreate (retry, orphan fix paths)
- [ ] No "redis keys points to missing objects" warnings in test output from delete-and-recreate paths
- [ ] Test file and function docstrings accurately describe the current `_pop_agent_session` behavior (no references to "delete-and-recreate" for the pop path)

## Downstream

This issue feeds into `/do-plan`, which will produce a plan document at `docs/plans/fix-pop-agent-session-tests.md`. The plan is then executed by `/do-build`.

Constraint: The fix must NOT change `_pop_agent_session`'s runtime behavior — it correctly uses in-place mutation. The work is updating tests and fixing `_extract_agent_session_fields` for its remaining callers.

tomcounsell:
## Summary
- Aligns 13 integration tests with `_pop_agent_session`'s in-place mutation behavior (same `agent_session_id`, status -> running)
- Adds round-trip unit test confirming `_extract_agent_session_fields` preserves `message_text` via `initial_telegram_message`
- Expands docstrings/comments clarifying why `_pop_agent_session` does NOT use delete-and-recreate and why `scheduling_depth` is excluded
- Fixes conftest popoto patch to cover submodules (sync/async Redis client divergence was the dominant root cause of the "redis keys point to missing objects" warnings)
- Replaces two bare `except Exception: pass` sites in `_execute_agent_session` with `logger.debug` calls

Closes #761

## Test plan
- [x] All 92 affected tests pass (`tests/integration/test_agent_session_queue_race.py`, `test_agent_session_scheduler.py`, `test_lifecycle_transition.py`, `test_agent_session_health_monitor.py`, `test_silent_failures.py`, `tests/unit/test_agent_session_queue.py`)

### DIFF
--- a/docs/plans/fix-pop-agent-session-tests.md
+++ b/docs/plans/fix-pop-agent-session-tests.md
@@ -160,16 +160,16 @@ For `_extract_agent_session_fields`:
 
 ## Test Impact
 
-- [ ] `tests/integration/test_agent_session_queue_race.py::test_pop_agent_session_preserves_fields` — UPDATE: assert same `agent_session_id` and `status == "running"` after pop.
-- [ ] `tests/integration/test_agent_session_queue_race.py` line 88 (`assert new_job.agent_session_id != original.agent_session_id`) — UPDATE: flip assertion.
-- [ ] `tests/integration/test_agent_session_queue_race.py` module docstring (line 3: "delete-and-recreate pattern used by _pop_agent_session") — UPDATE: describe in-place mutation.
-- [ ] `tests/integration/test_agent_session_queue_race.py` class docstring (line 94) — UPDATE: remove delete-and-recreate language for the pop path.
-- [ ] `tests/integration/test_agent_session_scheduler.py` — UPDATE: any assertions on `_pop_agent_session` behavior that expect a new ID.
-- [ ] `tests/integration/test_lifecycle_transition.py` — UPDATE: pop-path assertions only; leave L176/L191 (`complete_transcript` delete-and-recreate) as-is.
-- [ ] `tests/integration/test_agent_session_health_monitor.py` — UPDATE: pop-path assertions.
-- [ ] `tests/integration/test_silent_failures.py` — UPDATE: pop-path assertions.
-- [ ] `tests/unit/test_agent_session_queue.py` (create if missing) — ADD: round-trip unit test for `_extract_agent_session_fields` preserving `message_text` via `initial_telegram_message`.
-- [ ] Any test asserting `scheduling_depth` appears in the extracted field dict — DELETE or adjust to derive via parent walk.
+- [x] `tests/integration/test_agent_session_queue_race.py::test_pop_agent_session_preserves_fields` — UPDATE: assert same `agent_session_id` and `status == "running"` after pop.
+- [x] `tests/integration/test_agent_session_queue_race.py` line 88 (`assert new_job.agent_session_id != original.agent_session_id`) — UPDATE: flip assertion.
+- [x] `tests/integration/test_agent_session_queue_race.py` module docstring (line 3: "delete-and-recreate pattern used by _pop_agent_session") — UPDATE: describe in-place mutation.
+- [x] `tests/integration/test_agent_session_queue_race.py` class docstring (line 94) — UPDATE: remove delete-and-recreate language for the pop path.
+- [x] `tests/integration/test_agent_session_scheduler.py` — UPDATE: any assertions on `_pop_agent_session` behavior that expect a new ID.
+- [x] `tests/integration/test_lifecycle_transition.py` — UPDATE: pop-path assertions only; leave L176/L191 (`complete_transcript` delete-and-recreate) as-is.
+- [x] `tests/integration/test_agent_session_health_monitor.py` — UPDATE: pop-path assertions.
+- [x] `tests/integration/test_silent_failures.py` — UPDATE: pop-path assertions.
+- [x] `tests/unit/test_agent_session_queue.py` (create if missing) — ADD: round-trip unit test for `_extract_agent_session_fields` preserving `message_text` via `initial_telegram_message`.
+- [x] Any test asserting `scheduling_depth` appears in the extracted field dict — DELETE or adjust to derive via parent walk.
 
 ## Rabbit Holes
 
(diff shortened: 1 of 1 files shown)


tomcounsell:
## Review: Approved

Test alignment + docstring cleanup matches the plan precisely. `_pop_agent_session` runtime is unchanged; tests now correctly assert in-place mutation semantics (same `agent_session_id`, `status == "running"`).

### Verified
- [x] Plan requirements met (all 13 listed test updates landed across the 5 target files)
- [x] New unit test `tests/unit/test_agent_session_queue.py` round-trips `message_text` via `initial_telegram_message` and pins the intentional omissions (`scheduling_depth`, `agent_session_id`, `id`)
- [x] `_AGENT_SESSION_FIELDS` clarifying comment correctly explains virtual-property preservation and read-only `scheduling_depth`
- [x] `tests/conftest.py` popoto submodule patch correctly addresses `from ..redis_db import POPOTO_REDIS_DB` import-time binding capture (legitimate root cause for sync/async divergence warnings)
- [x] Production behavior change is limited to two `except Exception: pass` -> `logger.debug` conversions in `_execute_agent_session` and `send_to_chat` — both narrow, non-fatal paths. Reasonable hygiene improvement, not a behavior change
- [x] Verification table: integration tests pass, unit tests pass, no stale `_pop_agent_session.*delete-and-recreate` references in `tests/`, all touched files pass `ruff format --check`
- [x] Security: no secrets, no injection, no unsafe patterns
- [x] No UI changes (no screenshots needed)

### Notes
- The two `except: pass` -> `logger.debug` cleanups in `agent/agent_session_queue.py` are slightly outside the plan's stated scope ("test and docstring work") but they're trivial, defensive, and aligned with the silent-failures hardening philosophy. Worth keeping.
- Plan's Open Question 2 (Redis "missing objects" warnings) appears resolved by the conftest popoto submodule patch — the dominant root cause was sync/async client divergence, exactly as the PR description states.

PR #879, opened 2026-04-10
Dashboard: slug as session name, capture issue/PR links

valorengels:
## Summary

Three related dashboard-row fixes, all tracing back to the same missing wire-up in the SDLC flow.

### 1. `display_name` prefers slug over `context_summary`

Slug is the durable identifier tying together branch, worktree, plan doc, and GitHub issue — it's the canonical name operators reach for when referring to planned work. Before this change, the AI-generated `context_summary` shadowed the slug, so slugged sessions showed as paraphrased text instead of their real name (e.g. *\"Refactoring the resume hydration context\"* instead of `resume-hydration-context`).

Ad-hoc sessions without a slug still fall back to `context_summary` as before.

### 2. PostToolUse hook captures GitHub issue/PR URLs

**Root cause of the missing dashboard links:** `AgentSession.set_link()` has existed on the model for a while, but **no production code ever called it** — only tests. I verified this with a grep sweep: the only non-test writes are in `tests/integration/` and `tests/unit/`. As a result, `session.issue_url` and `session.pr_url` were permanently `None` in real data, and the dashboard row link cells were always empty. There's even a pre-existing audit doc (`docs/guides/summarizer-output-audit.md:190`) flagging this exact bug as the root cause of the missing link footer in summarizer output.

This PR wires `set_link()` into the natural capture point: the existing PostToolUse hook (`.claude/hooks/post_tool_use.py`) that already updates the AgentSession's `updated_at` and `tool_call_count` on every tool call. After each Bash call, if the command matches `\`gh issue\`` or `\`gh pr\``, the hook scans stdout for the corresponding GitHub URL and calls `set_link(\"issue\", …)` / `set_link(\"pr\", …)`.

URL detection is a pure function (`_extract_github_links`) kept separate from the hook's I/O so it can be unit-tested:
- `gh issue create/view` → captures issue URLs only
- `gh pr create/view` → captures PR URLs only
- Non-`gh` commands and non-github domains are ignored
- First URL in output wins (gh prints the definitive URL on its own line)
- Tolerates hyphenated/numeric org and repo names

### 3. Render-time fallback in `_session_to_pipeline`

Fix 2 is forward-looking — existing sessions won't retroactively gain captured links. To cover them without a data migration, `_session_to_pipeline` now falls back to scanning the already-loaded `session.history` events for GitHub URLs when `issue_url`/`pr_url` are `None` on the model. Model fields remain authoritative when present; the scan walks newest→oldest so the freshest URL wins. This is a bounded regex over in-memory strings — cheap per render.

## Test plan

- [x] `python -m pytest tests/unit/` — 3586 passed, 2 skipped
- [x] `python -m ruff format` / `python -m ruff check` clean on changed files
- [x] New unit tests:
  - `TestPipelineProgress::test_display_name_prefers_slug` — slug beats context_summary
  - `TestPipelineProgress::test_display_name_falls_back_to_context_summary` — ad-hoc sessions unchanged
  - `TestExtractGithubLinks` (7 tests) — URL extraction from history events
  - `TestGithubLinksFallback` (3 tests) — fallback wires through `_session_to_pipeline`; model field is authoritative
  - `TestExtractGithubLinks` in `test_post_tool_use_sdlc.py` (9 tests) — hook URL scraping
- [ ] Manual smoke test on running dashboard after merge — verify slugged rows show slugs and newly-created sessions capture gh URLs

## Notes

- No doc changes needed — this is a bug fix restoring documented-but-dead behavior (`docs/features/agent-session-model.md:72` already describes `set_link()` as what stores issue/plan/PR URLs *\"as each SDLC stage completes\"*).
- No update system changes — hook and render path are already deployed on every machine.
- No agent integration changes — hook runs inside the agent environment automatically.


valorengels:

Clean, well-motivated bug fix restoring documented-but-dead `set_link()` behavior. All three sub-fixes are coherent: slug priority in display_name, hook-based link capture, and render-time history fallback. Test coverage is thorough (19 new tests).

### Verified
- [x] Code correctness — slug priority, hook prefilters, fallback logic all sound
- [x] Security — no vulnerabilities found
- [x] Test coverage — 9 hook extraction + 7 UI extraction + 3 fallback wiring tests
- [x] Error handling — silent failure preserved in hook, bounded regex in fallback

### Tech Debt
- [ ] **`.claude/hooks/post_tool_use.py:295-296`** — `_GITHUB_ISSUE_URL_RE = re.compile(r"https://github\.com/[^\s/]+/[^\s/]+/issues/\d+")` — Regex patterns are duplicated identically in `ui/data/sdlc.py:29-30`. The hook's standalone sys.path isolation makes direct import difficult, but a shared constants file in `hook_utils/` or at minimum a cross-reference comment in both locations would prevent future drift.

valorengels:
## Review: Approved

Clean, well-structured bug fix that addresses a real root cause (set_link() existed but was never called in production). Three cohesive changes that are properly isolated and well-tested.

### Verified
- [x] Code correctness — display_name priority swap is correct (slug > context_summary); link extraction regex is sound; fallback only fills gaps
- [x] Test coverage — 19 new unit tests covering URL extraction (both hook and UI), display_name priority, fallback behavior, and model-field-wins-over-history
- [x] Security — no secrets, no injection risk; regex patterns are bounded and use `\s` anchors
- [x] Error handling — hook fails silently (existing pattern preserved); set_link catches save failures
- [x] Documentation — both dashboard.md and web-dashboard.md updated to reflect new behavior
- [x] No leftover debug code, no TODOs, no hardcoded values

### Architecture Notes
- Two `_extract_github_links` functions exist (hook vs UI) with different signatures — intentional, different contexts (command+output vs event list)
- The hook saves twice when a link is captured (once for tool_call_count, once inside set_link) — acceptable since it only happens on `gh issue`/`gh pr` commands
- Render-time fallback is bounded (regex over in-memory strings, early-exits when both found) — no performance concern

Issue #873, opened 2026-04-10
bug: idempotent transition_status() silently drops companion field edits — 4 of 5 call sites affected


valorengels:
> **Context:** Session state transitions in the Valor system go through two helpers in [`models/session_lifecycle.py`](models/session_lifecycle.py): `transition_status()` for non-terminal moves (e.g. `running → pending`) and `finalize_session()` for terminal moves (e.g. `running → completed`). Both are idempotent for the status field itself: if the session is already in the target state, they return without calling `session.save()`. That design is intentional for status-only updates but becomes a trap when callers also mutate companion fields before calling them.

### Observed: silent data loss on idempotent transitions

Callers routinely mutate companion fields (`priority`, `started_at`, `updated_at`, `task_list_id`, `message_text`, `auto_continue_count`) on the in-memory session object **before** calling `transition_status()`. When the transition is idempotent (target status equals current status), the companion field edits are silently dropped because `transition_status()` early-returns at [`models/session_lifecycle.py:220-225`](models/session_lifecycle.py) before any `session.save()` call:

```python
# models/session_lifecycle.py, lines 219-225
# Idempotency: if already in this state, skip
if current_status == new_status:
    logger.debug(
        f"[lifecycle] Session {getattr(session, 'session_id', '?')} "
        f"already in state {new_status!r}, skipping transition"
    )
    return
```

The only trace is the debug-level log at line 221, which is filtered out of normal log levels.

### The cost

Four of five affected call sites do NOT compensate. The fifth site is the only one that knows about the trap:

1. **`_recover_interrupted_agent_sessions_startup()`** at [`agent/agent_session_queue.py:1230-1232`](agent/agent_session_queue.py):
    ```python
    entry.priority = "high"
    entry.started_at = None
    transition_status(entry, "pending", reason="startup recovery")
    ```
    If the entry is already in `pending` (a race with another recovery path or health-check pass), both `priority="high"` and `started_at=None` are dropped. The "recovered" session keeps its old priority and `started_at`, so it sits at the back of the queue instead of being bumped to the front.

2. **`_agent_session_health_check()`** at [`agent/agent_session_queue.py:1358-1364`](agent/agent_session_queue.py):
    ```python
    entry.priority = "high"
    entry.started_at = None
    transition_status(
        entry,
        "pending",
        reason=f"health check: recovered stuck session (chat={worker_key})",
    )
    ```
[...]

Issue #872, opened 2026-04-10
cleanup: worker finally-block runs redundant snapshot/log/finalize on happy path — creates #867 race window


valorengels:
## Recon: Skipped

This issue was filed from a detailed integration audit of the worker session execution/completion path (2026-04-10). The issue body contains full code-level reconnaissance: file:line references, function definitions table, data flow analysis, and a 6-point solution sketch. No separate recon step is needed.
> **Context:** The Valor worker runs a session through the Claude Agent SDK in [`_execute_agent_session()`](agent/agent_session_queue.py) inside `agent/agent_session_queue.py`. Around that call sits a worker-loop `try/except/finally` block starting at `agent/agent_session_queue.py:2085`. On successful completion, **two different code paths** write a `session_complete` snapshot and try to finalize the session. One of them is redundant on the happy path — and its presence is what creates the race window exploited by #867.

### Observed: the happy path runs two finalization code paths

On a normal successful turn (no nudge, `defer_reaction=False`):

1. **Inside `_execute_agent_session`** at `agent/agent_session_queue.py:3080`:
    ```python
    if not chat_state.defer_reaction:
        complete_transcript(session.session_id, status=final_status)
    ```
    `complete_transcript` in [`bridge/session_transcript.py:285-292`](bridge/session_transcript.py) re-reads the session by `session_id` and calls `finalize_session(s, "completed")`, which runs the lifecycle side effects (auto-tag, branch checkpoint, parent finalization, status save) exactly once.

    Then at `agent/agent_session_queue.py:3166`, still inside `_execute_agent_session`:
    ```python
    save_session_snapshot(
        session_id=session.session_id,
        event="complete",
        ...
    )
    ```

2. **In the worker-loop `finally` block** at `agent/agent_session_queue.py:2170-2243`:
    ```python
    finally:
        if not session_completed:
            # Fix 4: Log lifecycle transition before completing
            target = "failed" if session_failed else "completed"
            session.log_lifecycle_transition(target, "worker finally block")      # line 2175
            # ...
            save_session_snapshot(                                                  # line 2184
                session_id=session.session_id,
                event="complete",
                ...
            )
            # ... nudge guard re-read ...
            else:
                await _complete_agent_session(session, failed=session_failed)      # line 2234
    ```

On the happy path with a single session record, the two paths interact like this:

[...]

Issue #870, opened 2026-04-10
bug: _execute_agent_session re-fetch filter on status=running causes silent degraded execution


valorengels:
> **Context:** The Valor worker pops `AgentSession` records off a Redis queue, runs them through the Claude Agent SDK, and writes back the session's per-turn state (conversation history, `task_list_id`, `session_type` for routing). `_execute_agent_session()` in [`agent/agent_session_queue.py`](agent/agent_session_queue.py) is the function that orchestrates one turn. It re-reads the session from Redis after the pop to get a Popoto-managed copy it can mutate and save. This issue is about a stale assumption baked into that re-read that causes silent degraded execution when another writer touches the session in the same window.
>
> Sibling to #867 — both are stale-read races on `AgentSession.status`, but mechanically distinct. #867 is a *write* that stomps the nudge's pending status; this is a *read* that silently accepts `None` and proceeds with degraded information.

### Current behavior

`agent/agent_session_queue.py:2680-2701` re-reads the just-popped session with a filter that requires `status="running"`:

```python
# Update the AgentSession (already created at enqueue time) with session-phase fields
agent_session = None
try:
    sessions = list(
        AgentSession.query.filter(project_key=session.project_key, status="running")
    )
    for s in sessions:
        if s.session_id == session.session_id:
            agent_session = s
            break
    if agent_session:
        agent_session.updated_at = datetime.now(tz=UTC)
        agent_session.branch_name = branch_name
        # Persist task_list_id so hooks can resolve this session
        agent_session.task_list_id = task_list_id
        agent_session.save()
        agent_session.append_history("user", (session.message_text or "")[:200])
except Exception as e:
    logger.debug(f"AgentSession update failed (non-fatal): {e}")

# Determine session type for routing decisions
_session_type = getattr(agent_session, "session_type", None) if agent_session else None
```

Any concurrent writer that transitions the session away from `running` between `_pop_agent_session` at [`agent/agent_session_queue.py:543`](agent/agent_session_queue.py) and this re-read — the **session watchdog** (`monitoring/session_watchdog.py:200,597`), the **health check** (`_agent_session_health_check()` at [`agent/agent_session_queue.py:1265`](agent/agent_session_queue.py)), a **steering injection**, a **circuit-breaker pause**, or any other lifecycle path — causes the filter to return empty, `agent_session = None`, and the function silently runs with degraded information:

[...]

Issue #867, opened 2026-04-10
Race: nudge re-enqueue stomped by worker finally-block finalize_session()


valorengels:
> **Context:** The Valor system runs conversational agents as `AgentSession` records in Redis via the [Popoto ORM](https://github.com/tomcounsell/popoto). A standalone worker picks up sessions, runs them through the Claude Agent SDK, and when a turn ends (`stop_reason=end_turn`) the *nudge loop* can re-enqueue the same session for another turn — this is how a single conversation spans multiple SDK turns as auto-continuation. Session state transitions are funneled through two helpers in `models/session_lifecycle.py`: `transition_status()` for non-terminal moves and `finalize_session()` for terminal moves. This issue is about a race between those two helpers that silently wipes a re-enqueued session.
>
> This is a sibling regression to #825 (closed 2026-04-08). That fix plugged a stale-index leak in `_complete_agent_session`'s re-query. The race here survives that fix because it's one level deeper — in `finalize_session()` itself.

## Problem

When the SDK ends a turn and the nudge path re-enqueues the session for auto-continuation, the worker's `finally` block races the nudge and marks the session `completed` — wiping the nudge's `pending` status **and** the per-turn counters that were written alongside it. The session never auto-continues, no retry fires (because the record is in a terminal state), and the agent goes silent.

### Concrete incident (2026-04-10)

Session `c94883fc69474195bd4f878971b2b528` (`session_id=tg_valor_-1003449100931_499`, message "complete SDLC on issue 851"). Timeline from [`logs/bridge.log`](logs/bridge.log):

```
04:06:33.986  running                  worker picked up session
04:07:07.945  Stored Claude Code UUID 3cc2fd0f-087c-4700-ae91-5f7e3c08f993
04:07:07.946  SDK stop_reason=end_turn
04:07:08.410  running→pending          nudge re-enqueue (auto_continue_count=1)
04:07:08.413  Reused session tg_valor_-1003449100931_499 for continuation
04:07:08.414  Sent result message (712 chars) to chat -1003449100931
04:07:10.666  memory_extraction extracted 10 observations
04:07:10.672  running→completed        worker finally block       ← stomp
04:07:10.786  running→completed        agent session completed    ← stomp
04:07:11.040+ Skipping session c94883fc... index says pending
              but actual status='completed'. Stale index entry.  (repeats for hours)
04:12:10      LIFECYCLE_STALL status=pending duration=337s
```

[...]

PR #885, opened 2026-04-10
Lifecycle CAS authority: eliminate blind session[0] and add conflict detection

### LINKED ISSUE #875: architecture: promote session_lifecycle.py to status authority with CAS (root-cause fix for #867/#870/#872/#873 race family)
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/lifecycle-cas-authority.md

## Recon Summary

**Source:** Three independent integration audits (lifecycle, worker execution, nudge loop) on 2026-04-10, all converging on the same structural recommendation.

**Code verified:**
- `models/session_lifecycle.py` — `finalize_session()` (line 58) and `transition_status()` (line 173) both do blind `session.save()` with no version check or CAS
- `_saved_field_values["status"]` backfill at lines 166-167 and 256-257 — existing Popoto-internal coupling that proves CAS layer is in-scope
- `POPOTO_REDIS_DB` (raw Redis client) used in 28+ files across the codebase — available for WATCH/MULTI/EXEC
- 15+ blind `[0]` re-read sites across hooks, tools, sdk_client, session_transcript, session_tags, sdlc_stage_marker
- 25 production writers of `AgentSession.status` across agent/, worker/, bridge/, monitoring/, tools/, scripts/, .claude/hooks/
- `AgentSession.get_by_id()` exists at models/agent_session.py:435 — potential base for centralized re-read

**Sibling issues verified as open:** #867, #870, #872, #873 — all independently diagnosed, all traceable to the same structural gap
**Related closed:** #825 (re-read fix in `_complete_agent_session`, established tie-break pattern)
**Active plan overlap:** docs/plans/extract-nudge-to-pm.md (#743) touches same files — coordination needed

> **Context:** This repository models agent work as [`AgentSession`](models/agent_session.py) records in Redis via the [Popoto ORM](https://github.com/tomcounsell/popoto). Session state transitions flow through two helpers in [`models/session_lifecycle.py`](models/session_lifecycle.py): `transition_status()` for non-terminal moves (e.g. `running → pending`) and `finalize_session()` for terminal moves (e.g. `running → completed`). Both helpers are "status setters" — they mutate the in-memory object and call `session.save()` without any version check, CAS, or authoritative re-read. This issue proposes promoting the module from "status setter" to "status authority" with optimistic concurrency, as the structural fix for a family of race-condition bugs filed earlier in the same audit pass.

## The problem this solves

The 2026-04-10 integration audits surfaced a **family** of bugs on `AgentSession.status` mutations, all traceable to the same structural gap: the lifecycle module is a thin "status setter" that trusts whatever in-memory object the caller hands it, with no version check, no compare-and-set, and no authoritative re-read. Concurrent writers race each other blindly, and whoever saves last wins.

The family of bugs already filed from this pattern:

- **#867** — nudge re-enqueue stomped by worker finally-block `finalize_session()`
- **#870** — `_execute_agent_session` racy `status="running"` filter on re-fetch
- **#871** — session recovery coverage split between worker and bridge-watchdog (adjacent but separate)
- **#872** — worker finally-block runs redundant snapshot/log/finalize on happy path
- **#873** — idempotent `transition_status()` silently drops companion field edits

Each of those issues can be patched individually with a narrow guard. The narrow guards work, but they:

1. **Don't compose.** Fixing #867's nudge race with a narrow guard doesn't prevent the watchdog-vs-worker race flagged in #867's audit appendix — that requires its own narrow guard.
2. **Accumulate.** Every new status, writer, or code path becomes a new surface for the same family of bugs. The audit found **25 production writers** of `AgentSession.status` across `agent/`, `worker/`, `bridge/`, `monitoring/`, `tools/`, `scripts/`, and `.claude/hooks/`.
3. **Rely on convention.** "Re-read with tie-break, then check if status is still what you expect, then save" is a five-step dance that 15+ call sites today get wrong in subtle ways (the "blind `[0]`" pattern surfaced by the lifecycle audit).

This issue proposes the **structural** fix that both the lifecycle audit and the worker audit independently arrived at: stop making callers manage lifecycle safety. Make the module own it.

### Deliverable 1 — CAS save in `models/session_lifecycle.py`

Replace the blind `session.save()` at [`models/session_lifecycle.py:155`](models/session_lifecycle.py) (inside `finalize_session`) and [`models/session_lifecycle.py:244`](models/session_lifecycle.py) (inside `transition_status`) with a Redis `WATCH`/`MULTI`/`EXEC`-backed save that:

- re-reads the record's `status` (and optionally a monotonic `version` field) as part of the save,
- refuses to write if the on-disk status has changed since the in-memory object was loaded,
- on conflict, either retries by re-reading fresh OR raises a `StatusConflictError` the caller can handle.

Popoto exposes the underlying Redis client as `POPOTO_REDIS_DB` (imported in `monitoring/health.py:83`, `scripts/reflections.py:778`, and elsewhere) — this is implementable without a Popoto change. The backfill of `_saved_field_values["status"]` at [`models/session_lifecycle.py:152`](models/session_lifecycle.py) already proves the module is willing to poke Popoto internals, so a CAS layer is in-scope for the same reason.

### Deliverable 2 — Id-based authority API

Add a new entry point to the lifecycle module that callers use in place of the pattern `list(AgentSession.query.filter(session_id=sid))[0]; session.foo = x; transition_status(session, ...)`:

```python
# New API
from models.session_lifecycle import update_session

update_session(
    session_id=session_id,
    new_status="pending",           # optional — omit for field-only update
    fields={"priority": "high", "started_at": None},
    reason="startup recovery",
    expected_status="running",      # optional CAS guard
)
```

Inside `update_session()`, the module:

- re-reads the session by `session_id` with the **same tie-break logic** currently duplicated at [`_complete_agent_session` in `agent/agent_session_queue.py:1096-1123`](agent/agent_session_queue.py) (prefer running, fall back to most recent by `created_at`),
- applies `fields` to the in-memory object,
- CAS-saves with the version fence from Deliverable 1,
- on conflict, logs and raises `StatusConflictError` for the caller.

This becomes the single door for non-terminal and companion-field updates. `finalize_session()` and `transition_status()` become the backend functions that `update_session()` calls — and the existing callers continue to work until they're migrated.

### Migration path (callers)

The audits counted:

- **15+ blind `[0]` re-read sites** across `.claude/hooks/*`, `agent/hooks/*`, `agent/sdk_client.py`, `bridge/session_transcript.py` (verified at line 285-287), `tools/session_tags.py`, `tools/sdlc_stage_marker.py`, `tools/sdlc_stage_query.py`, `tools/valor_session.py`, `agent/agent_session_queue.py` (the `_execute_agent_session` racy filter from #870), etc. All use `list(AgentSession.query.filter(session_id=sid))[0]` with no tie-break.
- **25 production writers** of `AgentSession.status`.
- **4 call sites** (enumerated in #873) that set companion fields before a `transition_status()` call without compensating for idempotent drops.

These do NOT all need to migrate in this issue. The plan should:

- Land Deliverables 1 and 2 as non-breaking additions — `finalize_session()` and `transition_status()` keep their current signatures and gain a CAS implementation underneath.
- Migrate the handful of call sites directly implicated in #867, #870, #872, #873 to prove the API out.
- File a follow-up sweep issue for the remaining ~15 call sites that migrates them mechanically.

## How this obsoletes the narrow patches

| Issue | Narrow fix | CAS fix |
|---|---|---|
| #867 nudge stomp | Widen the nudge guard to cover `active`, `waiting_for_children`, etc. | `finalize_session` refuses to write if on-disk status has changed since the in-memory read. The stomp can't happen regardless of the writer. |
| Watchdog-vs-worker race (#867 appendix) | Add a status-terminal check in `_complete_agent_session` | Same — CAS refuses the write. |
| #870 racy status filter | Remove the filter; add explicit `None` handling | Remove the filter; the `update_session()` helper handles re-read atomically. |
| #872 redundant finally finalize | Gate the finally call on `finalized_by_execute` | Same gate, PLUS CAS protects the crash path from stomping. Narrow fix is still needed; CAS is defense-in-depth. |
| #873 companion field drop | Save in idempotency branch OR new `update_session_and_transition` helper | **`update_session()` IS that helper.** One API covers all cases. |
| Pop-lock fail-open (lifecycle audit) | Fix `SETNX` to fail closed on Redis error | CAS doesn't fix this directly, but eliminates the concurrent-pop race window as a downstream consequence. |

## Why this is worth the refactor

- **"Sum of narrow fixes" scales O(N·M).** N call sites × M statuses × every transition needs a guard. CAS is O(1) — the module enforces the invariant once.
- **Future statuses and writers are automatically safe.** Adding `status="deferred"` or `status="paused_human_review"` later doesn't require touching every recovery path.
- **The 15+ blind `[0]` call sites get a clear migration target** that actually does the right thing. Without this issue, each fix has to reinvent the tie-break logic.
- **The lifecycle module's docstring at [`models/session_lifecycle.py:1-32`](models/session_lifecycle.py) already claims** ownership of status transitions. Promoting it to an authority matches the implied contract.

## Trade-offs and risks

- **Redis `WATCH`/`MULTI`/`EXEC` is not free.** CAS adds latency and retry complexity. On the hot path (every session transition), the cost is small but not zero. The plan should benchmark.
- **Popoto internals exposure.** The current `_saved_field_values["status"]` backfill is already a Popoto-internal coupling with a comment at `models/session_lifecycle.py:149-150` saying "If Popoto is upgraded, verify this coupling is still valid." CAS deepens that coupling. A Popoto upgrade could invalidate the approach. The plan should decide whether to build against Popoto or against raw Redis commands with a test for index consistency.
- **Migration fatigue.** Introducing `update_session()` alongside `transition_status()` and `finalize_session()` means three functions where one would do. The long-term endgame is for `update_session()` to become the only public API and the other two to become private backend helpers — the plan should commit to that direction explicitly so the migration converges.
- **`StatusConflictError` semantics.** If a CAS conflict fires, what's the caller's obligation? Retry? Bail out? The plan should define a small set of policies (retry N times, bail with clear error, log and continue) and give callers a way to pick.

## Surfaced by

Three independent integration audits ran on 2026-04-10 and converged on this recommendation from different starting points:

- **Lifecycle audit:** *"Promote the lifecycle module from 'status setter' to 'status authority'. Move the AgentSession re-read with tie-break into the lifecycle module itself — callers hand in a session id (not an instance), the module handles look-up, tie-break, lock, status check, mutate, save. Replace `session.save()` in lifecycle with a CAS helper using Redis `WATCH`/`MULTI`/`EXEC` (Popoto exposes the client)."*
- **Worker execution audit:** *"Extract a `SessionRunner` class that owns one session instance from pop to finalize... eliminates the `agent_session` vs `session` split, the double-finalize path, and most of the re-read discipline."*
- **Nudge-loop audit:** *"The finally-block nudge guard at `:2210-2243` is the sole coordinator... It works by reading the field value. There is no generational counter, no compare-and-swap."*

## Definitions

| `AgentSession` | Redis-backed Popoto model for one unit of agent work. 13 possible status values, managed via an `IndexedField`. | [`models/agent_session.py`](models/agent_session.py) |
| `transition_status()` | Current lifecycle helper for non-terminal status changes. Blind save; no version check. | [`models/session_lifecycle.py:158`](models/session_lifecycle.py) |
| `finalize_session()` | Current lifecycle helper for terminal status changes. Blind save; no version check. | [`models/session_lifecycle.py:43`](models/session_lifecycle.py) |
| **CAS (compare-and-set)** | A write pattern that fails if the underlying record has been modified since the reader last observed it. Implemented in Redis via `WATCH`/`MULTI`/`EXEC`. | [Redis transactions docs](https://redis.io/docs/latest/develop/interact/transactions/) |
| **Status authority** | An architectural role: the single module responsible for mutating a field, such that no other code writes the field directly. Callers interact with the module via explicit verbs; the module enforces invariants. | — |
| `IndexedField` | Popoto field type that maintains a Redis set index for fast `filter(field=value)` queries. Requires `_saved_field_values` to compute the old index key on update — the current backfill pattern at `models/session_lifecycle.py:152`. | [Popoto ORM](https://github.com/tomcounsell/popoto) |
| `POPOTO_REDIS_DB` | The raw Redis client exposed by Popoto. Already used by `monitoring/health.py:83`, `scripts/reflections.py:778`, and multiple migration scripts for raw Redis operations. | `popoto.redis_db` |
| Tie-break logic | The existing re-read logic at `agent/agent_session_queue.py:1096-1123` that prefers `status="running"` records then falls back to most-recent-by-`created_at`. To be centralized. | [`agent/agent_session_queue.py`](agent/agent_session_queue.py) |
| `StatusConflictError` | Proposed new exception class raised by CAS on write conflict. Callers choose retry / bail / log policy. | New |

## Related

- #867, #870, #872, #873 — the patchable siblings. This issue is the structural root-cause fix for all of them.
- #871 — related but separate (recovery coverage; not a race).
- #869 — unrelated in mechanism, found in the same audit pass.
- #825 (closed 2026-04-08) — the re-read fix in `_complete_agent_session` that established the tie-break pattern. This issue centralizes that pattern.
- #783 — earlier lazy-load index leak fix.
- #743 / [`docs/plans/extract-nudge-to-pm.md`](docs/plans/extract-nudge-to-pm.md) — active work to extract the nudge loop. Should land before or after this issue, not during — they touch the same files and will conflict.

## Solution sketch

The plan should decide:

1. **Scope of the CAS implementation.** Full CAS with `WATCH`/`MULTI`/`EXEC` vs. a simpler "re-read before save, compare-swap in Python" implementation. The full CAS is more correct; the Python approach is simpler but has a smaller race window that's still real.
2. **Conflict resolution policy.** Retry N times with backoff? Raise and let the caller handle? Configurable per call?
3. **Version field or status-only.** A monotonic `version: IntField` on `AgentSession` would make CAS trivially correct but adds a migration. `status`-only CAS is smaller scope but can only detect status changes (e.g., two writers both setting `running → pending` on an already-pending record wouldn't be detected — though that's the degenerate case that doesn't matter in practice).
4. **Migration scope of call sites.** Which of the 15+ blind `[0]` sites migrate in this PR vs. follow-up?
5. **Public/private API boundary.** Do `transition_status` / `finalize_session` become private after `update_session()` lands? Or stay public as low-level backends?

## Acceptance

- `models/session_lifecycle.py` exposes a new `update_session()` entry point that takes a `session_id`, optional `new_status`, optional `fields` dict, optional `expected_status`, and a `reason`. The function re-reads the session with tie-break, applies the fields, and CAS-saves.
- `finalize_session()` and `transition_status()` have a version fence (CAS or equivalent) that refuses to write if the on-disk status changed since the in-memory object's last read.
- A new `StatusConflictError` exception class is raised on CAS conflicts, with a documented retry/bail policy.
- The narrow-patch sites in #867, #870, #872, #873 either use the new API or include their tactical fix with a comment referencing this issue.
- A unit test with two concurrent writers targeting the same session confirms exactly one write succeeds and the other raises `StatusConflictError`.
- An integration test reproduces the #867 scenario (nudge + finally finalize) and confirms the nudge is preserved.
- Existing lifecycle tests (`tests/unit/test_session_lifecycle_consolidation.py`, `tests/integration/test_lifecycle_transition.py`, `tests/unit/test_agent_session_index_corruption.py`) pass unchanged.
- A regression benchmark shows the CAS overhead per transition is under a documented threshold (the plan chooses the threshold).

## Downstream context

This issue will be consumed by `/do-plan` to produce a plan at `docs/plans/lifecycle-cas-authority.md`. The plan should:

- Coordinate tightly with the `extract-nudge-to-pm` plan (#743) — both touch the lifecycle area and must not conflict.
- Explicitly choose the CAS scope (full Redis transactions vs. Python compare-swap) with a benchmark.
- Decide conflict resolution policy (retry/bail/configurable).
- Identify every call site that migrates in this PR vs. a follow-up sweep.
- Verify that the `IndexedField` backfill pattern (`_saved_field_values["status"]`) still works correctly under CAS.
- Include a **## Test Impact** section that enumerates every test file touching session lifecycle, status transitions, or the blind `[0]` re-read pattern.
- Include a **## Rollback plan** — CAS overhead regressions must be reversible without data loss.

valorengels:
## Summary
- Migrate all remaining call sites (`_cli_recover_single_agent_session`, `_enqueue_nudge`) from direct field mutation + `transition_status()` to the new `update_session()` / `get_authoritative_session()` CAS API
- Replace blind `sessions[0]` in `_enqueue_nudge` with `get_authoritative_session()` tie-break logic (prefer running, then most recent) and eliminate redundant double Redis query
- Add comprehensive CAS test suite (6 new test classes, 20 new tests) covering `StatusConflictError`, `get_authoritative_session`, `update_session`, and CAS in both `finalize_session` and `transition_status`
- Update existing idempotency test to reflect new save-on-idempotent behavior from #873

Closes #875

## Test plan
- [x] All 61 tests pass: `pytest tests/unit/test_session_lifecycle_consolidation.py -v`
- [x] Lint clean: `ruff check` on all 3 changed files
- [x] Format clean: `ruff format --check` on all 3 changed files
- [ ] Manual: verify CLI recovery with `python -m agent.agent_session_queue --recover <id>`
- [ ] Manual: verify nudge loop re-enqueue during active session


valorengels:

The CAS lifecycle authority implementation is solid. The `StatusConflictError`, `get_authoritative_session()`, and `update_session()` APIs are well-designed and correctly implement Python-level CAS. All 61 lifecycle tests pass, all 64 related unit tests pass, lint and format are clean, and all plan verification checks pass. The migration of `_cli_recover_single_agent_session` and `_agent_session_health_check` to `update_session()` is clean.

One tech debt finding that should be resolved before merge.

### Verified
- [x] Code correctness — CAS compare logic is sound; caller's object preserved for save
- [x] Plan requirements met — StatusConflictError, get_authoritative_session, update_session all implemented per plan
- [x] Test coverage — 20 new CAS tests across 6 test classes
- [x] Lint and format clean
- [x] All verification table checks pass

### Tech Debt
- [ ] **`agent/agent_session_queue.py:2488+2549`** — `reread_session = await asyncio.to_thread(get_authoritative_session, orig_session_id)` then `update_session(session.session_id, ...)` — `_enqueue_nudge()` performs 3 redundant Redis re-reads per nudge: (1) explicit `get_authoritative_session` at line 2488, (2) `update_session` internally calls `get_authoritative_session` again at `session_lifecycle.py:187`, (3) `transition_status` (called by `update_session`) calls `get_authoritative_session` a third time at `session_lifecycle.py:436`. Since `_enqueue_nudge` already holds the authoritative session after step 1, it should use the lower-level `transition_status()` directly on the already-read session (after applying fields), or `update_session` should accept an optional pre-fetched session to skip its internal re-read. The same triple-read pattern applies to the health check path at lines 1438-1444 (though the explicit first read there is the loop variable, not an extra call).

**Severity:** tech_debt

**Fix:** In `_enqueue_nudge`, after the explicit `get_authoritative_session` call and terminal guard, apply the companion fields directly to `session` and call `transition_status(session, "pending", reason=...)` instead of `update_session()`. This eliminates 2 of 3 Redis reads. The CAS in `transition_status` still provides conflict detection.

valorengels:
## Handoff note for incoming agents

PM session `0_1775816797375` (the one driving this PR) is running with a **MERGE GATE HOLD** in its context. It was steered with the following directive and consumed the steering message (activity confirms — the agent subsequently switched to reading the plan from `origin/main` via `git show origin/main:docs/plans/lifecycle-cas-authority.md` instead of its local worktree copy, which is the Bug 2 workaround from #884).

**The hold:** do NOT invoke `/do-merge` on this PR until PR #886 has merged to main. #886 fixes three chained bugs in `.claude/commands/do-merge.md` and `scripts/migrate_completed_plan.py` that would otherwise corrupt this PR's merge flow — specifically, the naive `.title()` at `migrate_completed_plan.py:225` mangles the slug `lifecycle-cas-authority` into `Lifecycle Cas Authority` and fails to match the real `Lifecycle CAS Authority` feature-doc title.

**Release procedure for whoever picks up this session:**

1. Confirm PR #886 has merged: `gh pr view 886 --json mergedAt --jq .mergedAt` returns a non-null timestamp
2. `git fetch origin main && git pull origin main` in the main checkout (not the worktree) to pull #886's fixes
3. Rebase this PR branch on the new main if necessary: `git -C .worktrees/lifecycle-cas-authority fetch origin main && git -C .worktrees/lifecycle-cas-authority rebase origin/main`
4. Steer session `0_1775816797375` to release the hold:
   ```
   python -m tools.valor_session steer --id 0_1775816797375 --message "PR #886 has merged. The /do-merge skill and migrate_completed_plan.py both contain the fixes for Bug 1, Bug 2, and Bug 3. You are cleared to proceed with /do-merge on this branch. Pull origin/main first to get the fixes into your worktree."
   ```
5. The PM session will pick up the steer at its next turn boundary and proceed.

**Safety nets still in place (can be cleaned up AFTER this PR merges):**
- Git tag `recovery-875-build` → points at `3e27fcf6`, the original stash-recovered commit. Superseded by everything on this branch. Safe to `git tag -d recovery-875-build` post-merge.
- Git stash `stash@{0}: sdlc-auto-stash` → original pre-crash stash. All its contents are now on this branch. Safe to `git stash drop stash@{0}` post-merge.
- `~/ai-875-lifecycle-cas-stash-backup.patch` → external backup of the stash. Safe to delete post-merge.

**Housekeeping backlog after merge** (these are all subsumed by this PR per `docs/plans/lifecycle-cas-authority.md`):
- Close #867 as "subsumed by #875" (CAS in `finalize_session()` eliminates the nudge stomp race)
- Close #870 as "fixed by #875" (`_execute_agent_session` re-fetch now uses `get_authoritative_session()`)
- Close #873 as "fixed by #875" (all four call sites migrated to `update_session()`)
- Re-scope or close #872 (correctness half fixed by CAS; only an optional `finalized_by_execute` perf optimization remains)
- Clarifying comments are already posted on #870/#872/#873 pointing at the plan.

**Filed during handoff:** #887 — Session isolation bypass bug discovered while babysitting this and the parallel #884 session. Not blocking this PR.

PR #893, opened 2026-04-10
Fix TeammateMetrics SortedField type mismatch (#890)

### LINKED ISSUE #890: TeammateMetrics: SortedField type mismatch causes silent response time data loss
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/teammate-metrics-sortedfield-fix.md

## Recon: Skipped

Recon is unnecessary — the issue body itself constitutes thorough reconnaissance. It includes:
- Root cause identification (SortedField expects scalar, code passes dict)
- Exact error message with reproduction evidence ("fired 3 times today")
- File:line references for all affected code paths
- Popoto field type definitions with source links
- Complete solution sketch with specific API calls (ListField, push())
- Exhaustive file change list
> **Context**: This repo includes a Popoto ORM model `TeammateMetrics` (`models/teammate_metrics.py`) that tracks intent classification counters and response times for the teammate/work routing system. [Popoto](https://github.com/tomcounsell/popoto) is a Redis-backed Python ORM used throughout this project.

## Problem

`TeammateMetrics` uses `SortedField` for `teammate_response_times` and `work_response_times`, but the code in `agent/teammate_metrics.py` treats these fields as dict-like collections (storing multiple `{member: score}` entries). `SortedField` is designed for a **single numeric scalar per model instance** — it creates a Redis Sorted Set index for range queries (`__gt`, `__gte`, `__lt`, `__lte`), not for storing collections.

When `record_response_time()` assigns a dict to the field and calls `.save()`, Popoto tries to convert the dict to a float for the sorted set score, producing:

```
POPOTO.model_base ERROR float() argument must be a string or a real number, not 'dict'
 Change the value or modify type on TeammateMetrics.teammate_response_times
```

**Current behavior**: The error fires on every `record_response_time()` call. It has fired 3 times today alone. Response time data is silently lost (the `try/except` in `record_response_time` swallows the error at debug level).

**Desired outcome**: Response times are stored correctly and retrievable for stats/analysis without errors in the logs.

## Definitions

|------|-----------|--------|
| `SortedField` | Popoto field that maintains a Redis Sorted Set index on a single numeric value per instance, enabling range queries | [popoto/fields/sorted_field_mixin.py](https://github.com/tomcounsell/popoto) |
| `ListField` | Popoto field that stores ordered collections; supports `max_length` for capped lists via LPUSH+LTRIM | [popoto/fields/shortcuts.py](https://github.com/tomcounsell/popoto) |
| `TeammateMetrics` | Single-instance Popoto model (keyed by "global") storing classification counters and response time history | `models/teammate_metrics.py` |
| `record_response_time()` | Function called after each SDK query to log elapsed time by mode (teammate vs work) | `agent/teammate_metrics.py:59` |

## Solution Sketch

Replace `SortedField` with `ListField(max_length=1000)` for both `teammate_response_times` and `work_response_times`. Each entry can be a string like `"1.23:1712345678"` (elapsed:timestamp). `ListField` with `max_length` uses efficient LPUSH+LTRIM — no need to read the full list to append.

Update `record_response_time()` to use `field.push()` instead of dict manipulation. Update `get_stats()` if it reads response times (currently it doesn't, but the field should be usable).

### Files to change
- `models/teammate_metrics.py` — change field types
- `agent/teammate_metrics.py` — simplify `record_response_time()` to use `push()`
- `tests/unit/test_qa_metrics.py` — update test setup (lines 63-64, 71-72)

## Acceptance Criteria

- [ ] `record_response_time()` stores data without errors
- [ ] No `float() argument must be a string or a real number, not 'dict'` errors in logs
- [ ] Response time history is capped at 1000 entries per mode
- [ ] Existing classification counters (`IntField`) are unaffected
- [ ] Unit tests pass with the new field type

## Downstream Context

valorengels:
## Summary
- Replace `SortedField(default=dict)` with `ListField(max_length=1000)` for response time fields in `TeammateMetrics` model
- Rewrite `record_response_time()` to use `push()` (atomic LPUSH+LTRIM) instead of read-modify-write dict pattern that caused `float()` conversion errors
- Update test fixtures and assertions to match new ListField-based storage

## Test plan
- [x] All 12 unit tests in `tests/unit/test_qa_metrics.py` pass
- [x] `ruff check` and `ruff format --check` clean on all changed files
- [x] No references to `SortedField` remain in changed files

Fixes #890

### DIFF
--- a/agent/teammate_metrics.py
+++ b/agent/teammate_metrics.py
@@ -5,7 +5,8 @@ times. All operations are fire-and-forget -- metrics failures must never
 affect message processing.
 
 Uses the TeammateMetrics Popoto model (single-instance pattern) instead of
-raw Redis commands for proper ORM lifecycle and index management.
+raw Redis commands for proper ORM lifecycle and index management. Response
+times are stored via ListField.push() which performs atomic LPUSH+LTRIM.
 """
 
 from __future__ import annotations
@@ -16,9 +17,6 @@ import time
 from agent.intent_classifier import TEAMMATE_CONFIDENCE_THRESHOLD
 
 logger = logging.getLogger(__name__)
-
-# Max response time entries to keep per mode
-_MAX_RESPONSE_TIMES = 1000
 
 
 def _get_metrics():
@@ -59,6 +57,9 @@ def record_classification(intent: str, confidence: float) -> None:
 def record_response_time(mode: str, elapsed_seconds: float) -> None:
     """Record response time for a teammate or work session.
 
+    Stores an "elapsed:timestamp" entry via ListField.push(), which
+    performs an atomic LPUSH+LTRIM capped at max_length (1000).
+
     Args:
         mode: "teammate" or "work"
         elapsed_seconds: Time from message receipt to response delivery.
@@ -68,27 +69,13 @@ def record_response_time(mode: str, elapsed_seconds: float) -> None:
         if not metrics:
             return
 
-        # Store as member:score in the sorted field (score=timestamp for time-windowed analysis)
-        member = f"{elapsed_seconds:.2f}:{time.time():.0f}"
-        timestamp = time.time()
+        entry = f"{elapsed_seconds:.2f}:{time.time():.0f}"
 
         if mode == "teammate":
-            times = dict(metrics.teammate_response_times or {})
-            times[member] = timestamp
-            # Keep only last N entries by removing oldest
-            if len(times) > _MAX_RESPONSE_TIMES:
-                sorted_entries = sorted(times.items(), key=lambda x: x[1])
-                times = dict(sorted_entries[-_MAX_RESPONSE_TIMES:])
-            metrics.teammate_response_times = times
+            metrics.teammate_response_times.push(entry)
         else:
-            times = dict(metrics.work_response_times or {})
-            times[member] = timestamp
-            if len(times) > _MAX_RESPONSE_TIMES:
-                sorted_entries = sorted(times.items(), key=lambda x: x[1])
-                times = dict(sorted_entries[-_MAX_RESPONSE_TIMES:])
-            metrics.work_response_times = times
+            metrics.work_response_times.push(entry)
 
-        metrics.save()
         logger.debug(f"[teammate_metrics] Recorded {mode} response time: {elapsed_seconds:.2f}s")
     except Exception as e:
         logger.debug(f"[teammate_metrics] Failed to record response time: {e}")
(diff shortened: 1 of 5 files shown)


valorengels (commented):
## PR Review: Fix TeammateMetrics SortedField type mismatch (#890)

**Verdict: APPROVE (no blockers found)**

### Plan Alignment

All plan tasks verified complete:
- **build-model**: SortedField replaced with ListField(max_length=1000) for both response time fields. Import updated. _MAX_RESPONSE_TIMES class attribute removed. Docstring updated.
- **build-record-fn**: Module-level _MAX_RESPONSE_TIMES constant removed. Dict manipulation replaced with single `push(entry)` call. `metrics.save()` removed (push writes directly). Docstring updated.
- **build-tests**: Mock fields changed from `{}` to `MagicMock()` (better than plan's `[]` -- tests now verify push() was called on the correct field and NOT called on the other).

### Code Correctness

1. **Root cause addressed**: SortedField expected float scalars but received dicts. ListField with push() stores string entries via atomic LPUSH+LTRIM. This is the correct fix.
2. **Entry format preserved**: `f"{elapsed_seconds:.2f}:{time.time():.0f}"` -- same string format, now stored correctly as list elements instead of dict keys.
3. **Capping delegated to ListField**: `max_length=1000` replaces the manual `_MAX_RESPONSE_TIMES` constant and the Python-side trimming logic. Cleaner and atomic.
4. **No save() needed**: `push()` writes directly to Redis (LPUSH+LTRIM), so removing `metrics.save()` is correct and avoids a redundant round-trip.
5. **Exception handling intact**: The outer try/except still catches any push() failures at debug level, maintaining the fire-and-forget contract.

### Edge Cases

- **mode not "teammate"**: Falls through to work_response_times -- same behavior as before, acceptable for metrics-only function.
- **Concurrent pushes**: LPUSH+LTRIM is atomic per-call; two simultaneous pushes both succeed. Plan explicitly noted this.
- **Empty/negative elapsed_seconds**: `:.2f` formatting handles 0.0 and negative floats without error.

### Test Coverage

- 12/12 tests pass. Tests verify push() called on correct field and NOT called on the wrong field (positive + negative assertions).
- `test_no_metrics_does_not_crash` covers the None-metrics path.
- Plan's optional failure-path tests (push() error, empty mode, negative elapsed) are covered by the broad except handler; sufficient for a small-appetite bug fix.

### Lint/Format

ruff check and ruff format clean on all changed files.

### Summary

Clean, minimal fix that addresses the root cause. +21/-35 lines -- net reduction in complexity. No SortedField references remain in changed files. Zero regressions (39 pre-existing failures on main, none introduced). Ready for merge.

PR #923, opened 2026-04-12
fix: prevent duplicate session execution after health-check recovery

### LINKED ISSUE #918: Bridge delivers same message multiple times to same session
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/bridge-session-dedup.md

## Problem

Session `tg_valor_-5051653062_8866` received the same Telegram message "hot fix your telegram history tool and run /update" 6+ times, each time triggering a full agent session execution. The session was marked as `running` with `auto_continue=0` but kept getting re-dispatched.

## Symptoms
- Same message_id delivered repeatedly to the same session
- Session stays in `running` state after completing
- Worker re-picks the session on each health check cycle

## Expected behavior
- Each unique Telegram message should be processed exactly once
- Completed sessions should not be re-queued

## Investigation starting points
- Bridge deduplication logic in `bridge/telegram_bridge.py` (message_id dedup)
- Session status transition after response delivery
- Worker health check logic that re-runs `running` sessions

## Context
Observed 2026-04-12 in Dev: Valor group chat (chat_id -5051653062)

tomcounsell:
## Summary
- Adds `response_delivered_at` field to `AgentSession` model — stamped when `send_to_chat` delivers the final response
- Health check now finalizes already-delivered sessions as `completed` instead of resetting to `pending` for re-execution
- Prevents the duplicate message delivery loop observed in issue #918

Closes #918

## Test plan
- [x] Unit tests for health check delivery guard (session with/without `response_delivered_at`)
- [ ] Verify `pytest tests/unit/test_agent_session_queue.py -x -q` passes (17 tests)
- [ ] Verify no existing tests broken

### DIFF
--- a/agent/agent_session_queue.py
+++ b/agent/agent_session_queue.py
@@ -1408,6 +1408,32 @@ async def _agent_session_health_check() -> None:
                     reason = f"exceeded timeout ({int(running_seconds)}s > {timeout}s)"
 
             if should_recover:
+                # Guard: if response was already delivered, finalize instead
+                # of recovering to pending (prevents duplicate delivery, #918)
+                if getattr(entry, "response_delivered_at", None) is not None:
+                    try:
+                        from models.session_lifecycle import finalize_session
+
+                        logger.info(
+                            "[session-health] Session %s already delivered response at %s, "
+                            "finalizing instead of recovering",
+                            entry.agent_session_id,
+                            entry.response_delivered_at,
+                        )
+                        finalize_session(
+                            entry,
+                            "completed",
+                            reason="health check: already delivered",
+                        )
+                        recovered += 1
+                    except Exception as e:
+                        logger.error(
+                            "[session-health] Failed to finalize already-delivered session %s: %s",
+                            entry.agent_session_id,
+                            e,
+                        )
+                    continue
+
                 is_local = worker_key.startswith("local")
                 logger.warning(
                     "[session-health] Recovering stuck session %s "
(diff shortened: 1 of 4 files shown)


tomcounsell (commented):
## Code Review: PASS ✅

**Fix correctness**: The `response_delivered_at` field is stamped only in the `action == "deliver"` branch of `send_to_chat`, which is the terminal delivery path. Nudge, fallback, and relay paths are correctly excluded — this prevents false-positive completion for mid-pipeline PM sessions.

**Health check guard**: Properly placed before the existing recovery logic with `continue` to skip re-queuing. `finalize_session` is used (not raw `transition_status`), which handles parent session finalization correctly.

**Model changes**: Clean additive change — nullable DatetimeField, properly added to `_DATETIME_FIELDS` and `_normalize_kwargs` for timestamp normalization. No breaking changes.

**Error handling**: Both the stamp (`try/except` with warning log) and the guard (`try/except` with error log + `continue`) fail gracefully. A stamp failure degrades to pre-fix behavior, not a crash.

**Tests**: Both key branches covered — 17/17 pass.

**Minor nits** (non-blocking):
- Duplicate local import of `finalize_session` in the guard block vs existing code below. Harmless since `continue` prevents reaching both.
- Test patching is belt-and-suspenders (both `patch()` and direct monkey-patching). Functionally correct.

No blockers found. Fix matches the plan and addresses the root cause in #918.

PR #1011, opened 2026-04-16
fix(#1006): prevent killed sessions from resurrecting in running index

### LINKED ISSUE #1006: Killed sessions resurrect in running index after worker restart or health check
> **Context:** This repo runs an agentic system where work is executed as [`AgentSession`](models/agent_session.py) records backed by [Popoto](https://github.com/tomcounsell/popoto) (a Redis ORM). Sessions have an `IndexedField` called `status` that Popoto maintains as Redis sets — one set per status value (e.g., `running`, `pending`, `killed`). The standalone [worker](worker/__main__.py) executes sessions and includes a health check cycle ([`_agent_session_health_check()`](agent/agent_session_queue.py)) and startup recovery ([`_recover_interrupted_agent_sessions_startup()`](agent/agent_session_queue.py)) that both query these status indexes to find sessions that need attention.

## Problem

When a session is killed via `valor-session kill` or `agent-session-scheduler kill`, the session's Popoto record is updated to `status=killed` and the defensive `srem` in [`finalize_session()`](models/session_lifecycle.py) attempts to remove it from all non-target index sets. Despite this, the killed session re-appears in the `running` status index and gets re-promoted to `pending` after:

1. **Worker restart** — `_recover_interrupted_agent_sessions_startup()` queries `AgentSession.query.filter(status="running")`, finds the ghost entry, and transitions it back to `pending`.
2. **Health check cycle** — `_agent_session_health_check()` queries the same `running` index and attempts recovery.

The only reliable workaround is to `delete()` the session record entirely, which is destructive and loses session metadata.

**Current behavior:**
- Kill a running session → status set to `killed` → session reappears as `running` after next health check or worker restart → gets re-queued as `pending` → worker picks it up again → infinite resurrection cycle.

**Desired outcome:**
- Kill a running session → session stays in `killed` status permanently → health check and startup recovery never touch it → session record is preserved for auditing.

## Definitions

| AgentSession | A Popoto-backed Redis model representing a unit of work (a Claude session). Has a `status` IndexedField that tracks lifecycle state. | [`models/agent_session.py`](models/agent_session.py) |
| Popoto IndexedField | A field type in the Popoto ORM that maintains a Redis set index per unique value, enabling `Model.query.filter(field=value)` lookups. When status changes, the old set entry must be removed (`srem`) and a new one added (`sadd`). | [Popoto source](https://github.com/tomcounsell/popoto) |
| finalize_session() | The canonical function for transitioning a session to a terminal status (`killed`, `completed`, `failed`, `abandoned`, `cancelled`). Includes CAS conflict detection, lifecycle logging, and a defensive `srem` to clean orphan index entries. | [`models/session_lifecycle.py:217`](models/session_lifecycle.py) |
| Defensive srem | A cleanup step added in PR #954 that explicitly removes the session's hash key from all non-target status index sets after `save()`. Intended to catch orphan entries left by stale-object saves. | [`models/session_lifecycle.py:363-384`](models/session_lifecycle.py) |
| Worker startup recovery | On startup, the worker re-indexes all Popoto models via `rebuild_indexes()`, then scans the `running` index for orphaned sessions and resets them to `pending`. | [`worker/__main__.py:275-300`](worker/__main__.py) |

## Prior Context

Related work:
- #950: Stale-index regression: pending index entry survives pending→killed transitions — Fixed the stale-save pattern where companion-field full saves clobbered status. Added defensive srem. Closed by PR #954.
- #804: valor-session kill uses `transition_status()` for terminal 'killed' status — Fixed kill to use `finalize_session()` instead.
- #822: Worker restart incorrectly kills pending sessions and marks interrupted sessions as completed — Added timing guards to startup recovery.
- #793: Resolve AgentSession status index corruption (ghost running sessions) — Earlier fix for the same class of index corruption bug.

## Solution Sketch

The bug has two dimensions: (1) the index corruption persists despite the defensive srem, and (2) consumers of the index don't guard against stale entries.

**Dimension 1 — Investigate why defensive srem fails:**
- The defensive srem in `finalize_session()` (lines 363-384) should remove the session from all non-target indexes. If sessions still appear in the `running` index after being killed, either the srem is silently failing, the `rebuild_indexes()` call at worker startup is re-adding the entry, or a concurrent stale-object save is re-inserting it after the srem runs.
- **Open question:** Does `rebuild_indexes()` re-read the hash and re-add to indexes based on current hash values? If the hash correctly says `killed`, rebuild should put it in the `killed` index, not `running`. If the hash still says `running`, the `save()` in finalize isn't persisting the status change.
- **Open question:** Is there a race between `finalize_session()` completing and another code path (e.g., heartbeat, steering drain) doing a full save with a stale in-memory status?

**Dimension 2 — Add terminal-status guards to index consumers:**
- `_recover_interrupted_agent_sessions_startup()` and `_agent_session_health_check()` should re-read the session's actual status from the Redis hash after pulling it from the index, and skip any session whose hash status is terminal (`killed`, `completed`, `failed`, `abandoned`, `cancelled`).
- This is a defense-in-depth measure: even if the index is stale, the recovery logic should not re-promote a session that is actually terminal.

## Recon Summary

**Confirmed:** 3 items — ready to include
- The kill path correctly uses `finalize_session()` with defensive srem (both `valor_session.py` and `agent_session_scheduler.py`)
- Health check (`_agent_session_health_check`) and startup recovery (`_recover_interrupted_agent_sessions_startup`) both trust the index without verifying hash status
- `rebuild_indexes()` is called at worker startup before recovery, which could re-corrupt indexes if the hash data is stale

- Original report suggested the fix is only about kill logic. Expanded scope to include consumer-side guards, since index corruption may have multiple sources and defense-in-depth is needed.

**Pre-requisites:** 0 items

**Dropped:** 0 items

## Acceptance Criteria

- [ ] After `valor-session kill --id <id>`, the session does not appear in `AgentSession.query.filter(status="running")` results — verified by immediate query and after simulated health check
- [ ] `_recover_interrupted_agent_sessions_startup()` skips sessions whose Redis hash status is terminal, even if they appear in the `running` index
- [ ] `_agent_session_health_check()` skips sessions whose Redis hash status is terminal, even if they appear in the `running` index
- [ ] Worker restart (`rebuild_indexes()` + recovery) does not resurrect killed sessions — verified end-to-end
- [ ] Regression test: kill a session, trigger health check, assert session stays killed
- [ ] Existing session lifecycle tests continue to pass

## Downstream

This issue feeds into `/do-plan`, which will produce a plan document at `docs/plans/killed-session-resurrection.md`. The plan is then executed by `/do-build`.

Constraints: The fix must not break the legitimate recovery path where truly orphaned `running` sessions (from crashed workers) are correctly reset to `pending`. The terminal-status guard must be a cheap check (single hash read) to avoid performance regression in health check cycles.

tomcounsell:
## Summary

- Add terminal-status guards to `_recover_interrupted_agent_sessions_startup()` and `_agent_session_health_check()` so killed/completed/failed sessions in stale Popoto IndexedField entries are skipped instead of re-promoted to pending
- Replace `rebuild_indexes()` with `repair_indexes()` in the worker loop's stale-entry handler (repair_indexes clears IndexedField sets before rebuilding)
- Fix pre-existing unterminated f-string in `_handle_dev_session_completion()`
- Add 15 regression tests covering all terminal statuses, mixed zombie/legitimate scenarios, and constant consistency

## Test plan

- [x] `tests/unit/test_zombie_session_resurrection.py` — 15 tests pass (parametrized across all terminal statuses)
- [x] Full unit test suite passes (4104 passed, 0 new failures)
- [x] Formatting clean (`ruff format`)

Closes #1006


tomcounsell:
## Review: Approved

**Reviewed by:** Claude Code (shepherding session)

### Changes Verified
- [x] Terminal-status guard in `_recover_interrupted_agent_sessions_startup()` — filters killed/completed/failed/abandoned/cancelled
- [x] Terminal-status guard in `_agent_session_health_check()` — skips stale index entries
- [x] `rebuild_indexes()` replaced with `repair_indexes()` (clears IndexedField sets properly)
- [x] Fixed pre-existing unterminated f-string in `_handle_dev_session_completion()`
- [x] 15 regression tests covering all terminal statuses — all pass
- [x] Full suite: 4104 passed

### Note
Code fix was committed directly to main (adaecd2f). PR contains test file and plan migration only.

PR #1012, opened 2026-04-16
fix(#984): restart flag TTL + heartbeat-aware worker-status

### LINKED ISSUE #984: Worker shutdown hiccups: stale restart flag + zombie PID passes worker-status
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/worker_lifecycle_fixes.md

While shepherding PM sessions through SDLC today I hit two related worker-lifecycle issues.

## 1. Stale restart flag triggers shutdown loop on a worker-only machine

On every session completion the worker evaluates `_check_restart_flag()` in `agent/agent_session_queue.py`. A flag timestamped **2026-04-13T14:16:00Z ("5 commit(s)")** — two days old — still fired today. The worker logged:

```
Restart flag found (2026-04-13T14:16:00Z 5 commit(s)), no running sessions — restarting bridge
Triggering graceful restart via SIGTERM...
```

...then shut down. The flag file itself was already gone from `data/restart-requested`. So either the flag got reposted after its unlink or the "found" log fired without the file existing. Result: after one PM session completed, the worker SIGTERM'd itself and did not return until manually killed and restarted.

On a skills/tools-only machine (no bridge), "restart bridge" on completion is the wrong default anyway — the worker shouldn't take itself down to restart a service it doesn't run.

### Fix ideas
- TTL the restart flag (ignore if older than, say, 1h).
- Skip `_check_restart_flag` entirely on machines where bridge isn't managed here.
- Actually delete the flag file atomically and log the inode/mtime, so we can tell whether the "found" check is racing a write.

## 2. `worker-status` reports RUNNING for a hung/zombie process

After the restart loop, `./scripts/valor-service.sh worker-status` kept printing:

```
Worker Status: RUNNING
PID: 70301
Uptime: 26:35
```

...but the worker had logged "Worker shutdown complete" 20+ minutes earlier and was not consuming the queue. `worker-start` was a no-op because PID 70301 still existed (in state S, 0 CPU). Only `kill -9` + waiting for launchd got us a functioning worker.

### Fix idea
- `worker-status` should verify liveness beyond "PID exists" — e.g. check last log heartbeat age or ping a health endpoint. Reporting RUNNING for a zombie is actively misleading, especially paired with #980.

## Labels
skills

## Recon Summary

- **`_check_restart_flag()` has no TTL**: `agent/agent_session_queue.py:2031` — the function reads the flag file, logs the content, and returns `True` without checking the flag's age. A flag from hours/days ago fires identically to a fresh one.
- **`_trigger_restart()` calls `os.kill(os.getpid(), SIGTERM)`**: `agent/agent_session_queue.py:2047` — this SIGTERMs the worker process itself (not the bridge), so on a tools-only machine it kills the running worker.
- **Flag is written with an embedded timestamp**: `scripts/update/git.py:268` — the flag content is `"{timestamp} {commit_count} commit(s)\n"`. The timestamp is already stored in the file; parsing it to implement TTL requires only a `datetime.fromisoformat()` comparison.
- **`_write_worker_heartbeat()` writes `data/last_worker_connected` on every health loop tick**: `agent/agent_session_queue.py:1760`. Dashboard uses it with a 360s threshold. `worker-status` in `scripts/valor-service.sh:637` ignores this file entirely — it only calls `pgrep -fi "python -m worker"` for liveness.
- **No bridge-mode detection exists in agent or worker code**: There is no `BRIDGE_MANAGED` flag, no `HAS_BRIDGE` env var, no `machine_type` setting in `config/settings.py`. The worker doesn't know whether it manages a bridge or is tools-only.

- **Two fixes, not three**: The TTL fix and the `worker-status` liveness fix are independent and both Small appetite. The bridge-mode guard is optional/nice-to-have; TTL alone prevents the stale-flag trigger.

**Pre-requisites:** 0 items

**Dropped:** 0 items

tomcounsell:
## Summary
- **Bug 1 fix**: `_check_restart_flag()` now parses the embedded ISO timestamp and ignores flags older than 1 hour, preventing stale flags from triggering worker self-destruct on machines where no bridge consumes the flag. Malformed/empty flag content is safely handled (logged + deleted).
- **Bug 2 fix**: `status_worker()` in `valor-service.sh` now reads the `data/last_worker_connected` heartbeat file mtime. When heartbeat age exceeds 360s (matching the dashboard threshold), status reports `STALE` with exit code 2 instead of `RUNNING`, so zombie PIDs no longer mask a hung worker.
- Tests updated: existing `TestRestartFlag` tests use fresh timestamps; new tests for stale flag TTL and malformed content added.
- Docs updated: `docs/features/bridge-self-healing.md` documents restart flag TTL and STALE worker status.

Closes #984

## Test plan
- [x] `pytest tests/integration/test_remote_update.py::TestRestartFlag -v` — 8/8 pass
- [x] Full unit suite — 4130 passed (30 pre-existing failures in unrelated LLM classifier tests)
- [ ] Manual: write a stale flag (`echo "2020-01-01T00:00:00Z 1 commit(s)" > data/restart-requested`) and verify worker ignores it
- [ ] Manual: stop worker, wait 6 min, run `./scripts/valor-service.sh worker-status` and verify STALE output

### DIFF
--- a/agent/agent_session_queue.py
+++ b/agent/agent_session_queue.py
@@ -26,7 +26,7 @@ import signal
 import subprocess
 import time
 from collections.abc import Awaitable, Callable
-from datetime import UTC, datetime
+from datetime import UTC, datetime, timedelta
 from pathlib import Path
 from typing import Any
 
@@ -2239,9 +2239,37 @@ def register_callbacks(
 _RESTART_FLAG = Path(__file__).parent.parent / "data" / "restart-requested"
 
 
+_RESTART_FLAG_TTL = timedelta(hours=1)
+
+
 def _check_restart_flag() -> bool:
-    """Check if a restart has been requested and no sessions are running across all projects."""
+    """Check if a restart has been requested and no sessions are running across all projects.
+
+    Returns False (and deletes the flag) if the flag is older than 1 hour,
+    malformed, or empty — preventing stale flags from triggering self-destruct.
+    """
     if not _RESTART_FLAG.exists():
+        return False
+
+    flag_content = _RESTART_FLAG.read_text().strip()
+
+    # Validate flag freshness via embedded timestamp
+    try:
+        timestamp_str = flag_content.split()[0]
+        flag_time = datetime.fromisoformat(timestamp_str)
+        # Ensure timezone-aware comparison
+        if flag_time.tzinfo is None:
+            flag_time = flag_time.replace(tzinfo=UTC)
+        flag_age = datetime.now(UTC) - flag_time
+        if flag_age > _RESTART_FLAG_TTL:
+            logger.warning(f"Restart flag is stale (age={flag_age}) — ignoring and deleting")
+            _RESTART_FLAG.unlink(missing_ok=True)
+            return False
+    except (ValueError, IndexError):
+        logger.warning(
+            f"Restart flag has malformed content ({flag_content!r}) — ignoring and deleting"
+        )
+        _RESTART_FLAG.unlink(missing_ok=True)
         return False
 
     # Check all workers for running sessions
@@ -2250,7 +2278,6 @@ def _check_restart_flag() -> bool:
         logger.info(f"Restart requested but {len(running)} session(s) still running — deferring")
         return False
 
-    flag_content = _RESTART_FLAG.read_text().strip()
     logger.info(f"Restart flag found ({flag_content}), no running sessions — restarting bridge")
     return True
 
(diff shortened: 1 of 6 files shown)


tomcounsell:
## Review: Approved

**Reviewed by:** Claude Code (shepherding session)

### Changes Verified
- [x] TTL guard in `_check_restart_flag()` — flags >1 hour ignored and deleted
- [x] Malformed/empty flag content handled safely (logged + deleted)
- [x] Heartbeat-aware `status_worker()` in valor-service.sh — STALE exit code 2 when heartbeat >360s
- [x] Docs updated — bridge-self-healing.md TTL and STALE sections
- [x] Tests pass — TestRestartFlag 6/6 passed (includes 2 new: stale flag, malformed content)
- [x] Pre-existing integration test timeout in test_no_restart_flag_when_up_to_date — unrelated

### Risk Assessment
Low risk — TTL guard is additive with safe fallback (delete stale flag). Worker-status change is backwards-compatible (new exit code 2 for STALE, existing 0/1 unchanged).

Issue #963, opened 2026-04-14
Session routing integrity: steering paths skip dedup + health check kills parents with active children


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/session-routing-health-integrity.md

> **Context:** This system routes user messages from Telegram through a Python bridge ([`bridge/telegram_bridge.py`](bridge/telegram_bridge.py)) into long-running PM sessions. A **PM session** ([`agent/agent_session_queue.py`](agent/agent_session_queue.py)) is a parent orchestrator that spawns short-lived **dev sessions** as children to execute code tasks. The bridge includes a **reconciler** ([`bridge/reconciler.py`](bridge/reconciler.py)) that runs every 180 seconds to recover messages that the live handler may have missed. A **steering message** is an in-flight message queued into a running session's `queued_steering_messages` list rather than creating a new session. The **session health check** (in `agent_session_queue.py`) runs periodically to recover sessions that appear stuck.
>
> On 2026-04-14, session `tg_valor_-1003449100931_617` processed a message correctly via steering, but the same message was also picked up by the reconciler and spawned a duplicate session (`tg_valor_-1003449100931_623`). That duplicate session was then incorrectly killed by the health check even though it had spawned a child dev session actively doing work. The compounding failures caused wrong output to be delivered to the user.

## Problem

Two independent bugs in the session routing and health monitoring systems can compound to produce incorrect, duplicate, or lost agent work.

### Bug 1: Steering paths do not record dedup, causing reconciler double-dispatch

When the bridge steers a message into a running session (via `push_steering_message()`), it never calls `record_message_processed()`. Three minutes later the reconciler scans the same chat, finds no dedup record for that message, treats it as "missed," and enqueues a brand-new session for it.

There are four steering paths in [`bridge/telegram_bridge.py`](bridge/telegram_bridge.py) — lines ~1023, ~1244, ~1283, and ~1323 — none of which call `record_telegram_message_handled()`. The dispatch refactor in #948 centralized dedup for enqueue paths only; steering paths were excluded because they call `push_steering_message()` instead of `enqueue_agent_session()`, and no follow-up pass covered them.

The existing AST contract test ([`tests/unit/test_bridge_dispatch_contract.py`](tests/unit/test_bridge_dispatch_contract.py)) bans direct `enqueue_agent_session` calls in the handler but does not verify that steering paths write a dedup record, so the gap passes CI.

[...]

Issue #944, opened 2026-04-14
bug: health check skips recovery for stuck dev sessions when a shared project-keyed worker is alive


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/health-check-no-progress-recovery.md

> **Context:** This issue relates to the session health check in [`agent/agent_session_queue.py`](agent/agent_session_queue.py) — a periodic background task (every 5 minutes) that scans for `running` AgentSessions whose worker process died, and resets them to `pending` so they can be retried. AgentSessions are the primary unit of work in this system: each incoming Telegram message or local Claude Code session creates one and routes it to a worker loop keyed by `worker_key`. The relevant session types are PM sessions (orchestrators), Dev sessions (code executors), and Teammate sessions (conversational). Slugless dev sessions — those not tied to a named work item — share a `worker_key` equal to the project key (e.g. `"valor"`), the same key used by PM sessions.

## Problem

When a slugless dev session gets stuck in `running` status with no active process handling it, the health check fails to recover it within a reasonable time if a PM session happens to be running under the same project key.

**Current behavior:**

The health check resolves liveness as:

```python
worker = _active_workers.get(worker_key)   # keyed by project, e.g. "valor"
worker_alive = worker is not None and not worker.done()
```

For slugless dev sessions, `worker_key = project_key`. If a PM session is running under that key, `worker_alive = True` — and the health check skips the dev session entirely, falling through to the 45-minute timeout path. A stuck dev session with `turn_count=0` and no `log_path` (i.e. never actually processed) will not be recovered until 45 minutes after creation, even though no process is actually handling it.

**Observed:** Two local `"test"` dev sessions were created at ~04:09–04:10 UTC today (2026-04-14), got stuck in `running` with `turn_count=0`/empty `log_path`, and were not recovered by health checks at 04:13 and 04:15 — because a PM session for `"valor"` had started at 04:11 and kept `worker_alive = True`.

**Desired outcome:**

Stuck dev sessions are recovered within one health-check cycle (≤5 minutes) after the startup guard expires, regardless of whether a PM session is running under the same project key.

## Definitions

| Term | Definition | Reference |
|------|-----------|-----------|
| `AgentSession` | Popoto ORM model representing a unit of work (one Telegram message or CLI session), tracked through its lifecycle from `pending` → `running` → `completed/failed/abandoned`. | [`models/agent_session.py`](models/agent_session.py) |
[...]

PR #1039, opened 2026-04-18
feat(health): two-tier no-progress detector (#1036)

### LINKED ISSUE #1036: 300s no-progress guard kills sessions before first turn despite live SDK heartbeat
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/session-heartbeat-progress-guard.md

> **Context:** The `AgentSession` queue (`agent/agent_session_queue.py`) runs a periodic health check (`_agent_session_health_check`) every 300 seconds. Its job is to recover sessions that look stuck — specifically, sessions in `running` state where the worker is alive but no real work is happening. The check uses three "own-progress" signals on the AgentSession record (`turn_count`, `log_path`, `claude_session_uuid`) plus a child-session check. The Claude SDK subprocess is monitored separately by `agent/messenger.py`, which writes "SDK heartbeat: running Ns, communicated=..." log lines but does NOT update any progress field on the AgentSession. This split observability is the source of the bug below.

## Problem

A session whose initial prompt takes longer than 5 minutes to produce its first turn is incorrectly killed by the no-progress guard, even though the SDK subprocess is alive and processing. After the kill, the SDK subprocess continues to run as an orphan (heartbeats keep firing for minutes), and the AgentSession record is eventually deleted entirely, making post-mortem impossible.

**Current behavior (observed 2026-04-17):**

A meta-orchestrator PM session was launched with a ~4000-character initial prompt. Timeline from `logs/worker.log`:

```
10:45:54 [0_1776422754682] Background task started
10:46:54 SDK heartbeat: running 60s,  communicated=False
10:47:54 SDK heartbeat: running 120s, communicated=False
10:48:54 SDK heartbeat: running 180s, communicated=False
10:49:54 SDK heartbeat: running 240s, communicated=False
10:50:54 SDK heartbeat: running 300s, communicated=False
10:51:53 [session-health] Recovering stuck session ...:
         worker alive but no progress signal, running for 358s
         (>300s guard, turn_count=0, log_path=None, claude_session_uuid=None)
10:51:53 LIFECYCLE transition=running→pending context="health check: recovered stuck session"
10:51:54 SDK heartbeat: running 360s, communicated=False  ← still alive after kill
10:52:54 SDK heartbeat: running 420s, communicated=False  ← still alive
10:53:54 SDK heartbeat: running 480s, communicated=False  ← still alive
```

By the next day, the entire session record was gone (`valor-session status --id 0_1776422754682` returns "Session not found").

**Three problems compound:**

1. **Guard fires before SDK can authenticate on long prompts.** `_has_progress()` ([`agent/agent_session_queue.py:1532-1568`](agent/agent_session_queue.py)) only treats `claude_session_uuid`, `log_path`, or `turn_count > 0` as progress. The SDK's own liveness signal (the messenger heartbeat) is ignored. For prompts that take 5+ minutes to digest, the SDK is genuinely working but invisible to the guard.

2. **Recovery doesn't kill the orphaned subprocess.** The session is transitioned `running → pending`, but the Claude subprocess keeps running. Heartbeats continue for minutes. This wastes API tokens, may produce work that never lands anywhere, and double-counts against `MAX_CONCURRENT_SESSIONS` if a new pickup happens.

3. **Eventual deletion erases the audit trail.** The session record is gone — no terminal status, no history, no log_path. There's no way to tell from the record whether the SDK eventually finished, what it produced, or why it was deleted. Compare to a normal terminal status (`completed`, `abandoned`, `failed`) which preserves history.

**Desired outcome:**

- Sessions with a live SDK heartbeat are NOT killed by the no-progress guard. The heartbeat counts as progress.
- If a session IS killed by the guard, the SDK subprocess is also killed (or at minimum logged as orphaned).
- Sessions that fail or are recovered are preserved in a terminal state with full history; nothing silently disappears from `valor-session` listings.

## Definitions

| No-progress guard | Health-check rule that recovers a `running` session if it hasn't shown a progress signal within `AGENT_SESSION_HEALTH_MIN_RUNNING` (300s) | [`agent/agent_session_queue.py:131,1664-1675`](agent/agent_session_queue.py) |
| `_has_progress()` | The function that decides whether a session has shown any work signal; checks `turn_count`, `log_path`, `claude_session_uuid`, and active children | [`agent/agent_session_queue.py:1532-1568`](agent/agent_session_queue.py) |
| SDK heartbeat | Messenger log line emitted every 60s for a running session, of the form `[<session>] SDK heartbeat: running Ns, communicated=<bool>`. NOT written to the AgentSession record. | [`agent/messenger.py`](agent/messenger.py) |
| Recovery | The act of transitioning a `running` session back to `pending` so a worker can re-pick it up | [`agent/agent_session_queue.py:1738-1763`](agent/agent_session_queue.py) |
| MAX_CONCURRENT_SESSIONS | Worker semaphore limiting simultaneous sessions, default 8 | [`worker/__main__.py:179-182`](worker/__main__.py) |

## Prior Context

Related work (from grep):
- Issue #944 — added `_has_progress()` to handle slugless dev sessions sharing a worker_key with their PM (see [`agent/agent_session_queue.py:1661-1663`](agent/agent_session_queue.py))
- Issue #963 — referenced in `_has_progress()` docstring; long-warmup BUILD sessions
- Issue #1006 — terminal-zombie handling at the top of the health check loop ([`agent/agent_session_queue.py:1620-1627`](agent/agent_session_queue.py))
- Issue #918 — `response_delivered_at` guard prevents duplicate delivery during recovery ([`agent/agent_session_queue.py:1702-1726`](agent/agent_session_queue.py))

None of these address the SDK-heartbeat-vs-progress-field divergence or the orphan-subprocess problem.

## Solution Sketch

Three independent fixes for the planner to consider:

**Fix 1 — Treat SDK heartbeat as a progress signal.** Either:
- (a) Have `messenger.py` write a `last_heartbeat_at` timestamp on the AgentSession on every heartbeat tick, and add a `last_heartbeat_at >= now - 90s` check to `_has_progress()`. This is the minimal change.
- (b) Have the messenger update `claude_session_uuid` as soon as the SDK authenticates (before first tool use), since that's already a checked progress field. This requires plumbing the UUID out of the SDK earlier.

**Fix 2 — Kill the orphaned subprocess on recovery.** When the health check transitions `running → pending`, look up the active worker task / SDK process for that session and cancel it. This prevents wasted compute and double-billing under the concurrency cap.

**Fix 3 — Preserve recovered sessions.** When a session is recovered repeatedly (e.g. 2-3 times) and never makes progress, finalize it as `failed` with the full history rather than allowing it to be deleted by downstream cleanup. Investigate what cleanup path actually deleted `0_1776422754682` — TTL? Manual `agent_session_cleanup`? — and require a terminal status before deletion is allowed.

Open questions:
1. Why does `messenger.py` not write progress fields when it has the live signal? Is there a deliberate reason (avoiding write contention) or is it just historical?
2. What deletes sessions that never reached a terminal state? Is it `cleanup_corrupted_agent_sessions`, TTL via `Meta.ttl`, or something else? The deletion path needs identifying before Fix 3.
3. Should the no-progress guard scale with prompt size? A session with a 4000-char prompt may legitimately need 5+ minutes to first turn; a session with a 50-char prompt should NOT. Consider a per-session timeout that scales with `len(message_text)`.

## Recon Summary

In-process recon (deep fan-out skipped — concrete bug with reproducible log evidence and identified code paths):

**Confirmed:**
- `AGENT_SESSION_HEALTH_MIN_RUNNING = 300` at [`agent/agent_session_queue.py:131`](agent/agent_session_queue.py)
- `_has_progress()` checks only `turn_count`, `log_path`, `claude_session_uuid`, and active children — no heartbeat signal
- SDK heartbeat is logged in `agent/messenger.py` but does NOT touch the AgentSession record
- Recovery transitions to `pending` without canceling the SDK subprocess (no kill code in the recovery branch [1728-1763])
- Session record deleted entirely after some hours — neither `completed` nor `abandoned` nor `failed` was set

**Pre-requisites:** None — this is a self-contained set of fixes within `agent/`.

**Dropped:** None.

## Acceptance Criteria

- [ ] A PM session with a 4000-character initial prompt that takes 5-7 minutes to first turn is NOT killed by the no-progress guard, provided the SDK heartbeat is alive
- [ ] When a session IS recovered by the no-progress guard, the SDK subprocess for that session is terminated within 60s (no orphan heartbeats continue)
- [ ] A session that is recovered and then never makes progress on retry transitions to a terminal state (`failed` or `abandoned`) with full history preserved — `valor-session status --id <id>` continues to work
- [ ] Regression test: simulate a session that delays its first turn by 6 minutes; confirm it is not recovered while the heartbeat is alive
- [ ] Regression test: simulate a recovery; confirm the orphan subprocess is reaped

## Downstream

Constraints for the planner:
- Do NOT raise `AGENT_SESSION_HEALTH_MIN_RUNNING` blindly — that would just hide the problem for slightly longer. The fix is to use the heartbeat signal, not stretch the guard.
- The fix must not break the original use case (#944): slugless dev sessions sharing a worker_key with their PM, where worker_alive alone doesn't prove the dev is being processed.
- Preserve the existing `response_delivered_at` guard (#918) — that path is correct.

valorengels:
## Summary
Implements a dual-tier health detector that prevents false-kills of long-warmup PM sessions while still reaping genuinely wedged ones, and preserves full audit history when kills do happen.

### Tier 1 — dual heartbeat (OR semantics)
- `last_heartbeat_at` — written every 60s by the queue-layer `_heartbeat_loop` inside `_execute_agent_session`.
- `last_sdk_heartbeat_at` — written every 60s by `BackgroundTask._watchdog` via the `on_heartbeat_tick` messenger callback.
- `_has_progress()` returns True if **either** heartbeat is within `HEARTBEAT_FRESHNESS_WINDOW` (90s). Tolerates single-writer failure (e.g. wedged event loop).

### Tier 2 — activity-positive reprieve gates
- `_tier2_reprieve_signal()` evaluates three OS-level liveness checks via `psutil`: process-alive / has-children / recent stdout. Any one passing signal reprieves the kill.
- Messenger is plumbed with `on_sdk_started(pid)` and `on_stdout_event()` callbacks so the queue can populate `SessionHandle.pid` and `last_stdout_at` without breaking the ORM boundary.

### Kill path
- `_active_sessions[id] = SessionHandle(task, pid)` registered at the top of `_execute_agent_session` (before any raise site); cleaned up via `task.add_done_callback` so the entry is always popped.
- On kill: cancel `handle.task`, await 0.25s, bump `recovery_attempts`; at `MAX_RECOVERY_ATTEMPTS` (2) finalize as `failed` so the session reaches terminal status.
- `StatusConflictError` around `transition_status` is caught.

### Kill-switch
- `DISABLE_PROGRESS_KILL=1` suppresses the kill transition while keeping Tier 1 flagging + Tier 2 evaluation active. Lets operators collect data during rollout.

### Messenger architectural boundary
- Three optional callbacks on `BossMessenger` with `notify_*` wrappers that catch callback exceptions. Messenger imports nothing from `models/`.
- `_run_harness_subprocess` in `agent/sdk_client.py` accepts and fires the callbacks.

### Metrics (Redis counters per project_key)
- `session-health:tier1_flagged_total`
- `session-health:tier2_reprieve_total:{alive|children|stdout}`
- `session-health:kill_total`

### New AgentSession fields
All five added to `_AGENT_SESSION_FIELDS` for round-trip:
- `last_heartbeat_at`, `last_sdk_heartbeat_at`, `last_stdout_at` (DatetimeField)
- `recovery_attempts`, `reprieve_count` (IntField default=0)

## Preserves prior-fix invariants
- #918 (`response_delivered_at` guard)
- #944 (slugless dev session child-progress)
- #963 (child-activity branch in `_has_progress`)
- #1006 (terminal-status zombie guard at top of health loop)

## Testing
- [x] 35 unit tests in `test_health_check_recovery_finalization.py` (dual-heartbeat OR, Tier 2 gates, recovery cancellation, attempts counter, kill-switch, fields round-trip).
- [x] 9 unit tests in `test_messenger_callbacks.py` (callback invocation, exception resilience, ORM boundary AST scan).
- [x] 12 integration tests in `test_session_heartbeat_progress.py` (end-to-end Tier 1 freshness, Tier 2 stdout reprieve, recovery_attempts persistence, kill-switch).
- [x] Full adjacent suite (284 tests) passes with zero regressions.
- [x] Linting (ruff check + format) clean.

## Documentation
- [x] `docs/features/bridge-self-healing.md` — new 'Two-tier no-progress detector' section covering both tiers, kill path, kill-switch, metrics, per-session fields, and messenger callbacks.
- [x] `docs/features/bridge-worker-architecture.md` — new sections on the `_active_sessions` registry and messenger callback plumbing.

## Definition of Done
- [x] Tested: All new unit + integration tests passing (284 adjacent tests pass)
- [x] Documented: Two doc files updated

Closes #1036


valorengels:
## Review: Approved

Clean, faithful implementation of the dual-heartbeat + activity-positive-reprieve design specified in `docs/plans/session-heartbeat-progress-guard.md`. All plan invariants verified against the actual code, all tests pass, and every prior-fix invariant (#918/#944/#963/#1006) is preserved.

### Verified

**Plan → Implementation mapping**
- [x] **Tier 1 dual heartbeat** — `_has_progress()` at `agent/agent_session_queue.py:1597-1605` iterates `("last_heartbeat_at", "last_sdk_heartbeat_at")` and returns True if either is within `HEARTBEAT_FRESHNESS_WINDOW` (90s). OR semantics confirmed.
- [x] **Tier 2 reprieve gates** — `_tier2_reprieve_signal()` at `agent/agent_session_queue.py:1622-1684` evaluates (c) process-alive, (d) has-children, (e) recent-stdout with full `try/except (psutil.NoSuchProcess, psutil.AccessDenied, ImportError)` wrapping. Returns "children" in preference to "alive" for stronger telemetry, matching the plan.
- [x] **`SessionHandle` + `_active_sessions` registry** — dataclass at `agent/agent_session_queue.py:2365-2386`, registry at line 2395. Docstring explicitly documents single-writer/multi-reader pattern and cleanup contract.
- [x] **Registry registration precedes all raise sites** — `agent/agent_session_queue.py:3788-3796` registers `SessionHandle(task=asyncio.current_task())` before any other work, with `task.add_done_callback(_cleanup_registry)` guaranteeing pop on any exit path (including `CancelledError` and exceptions). This is a reasonable variant of the plan's try/finally that achieves the same lifecycle contract.
- [x] **T+0 heartbeat write** — `agent/agent_session_queue.py:3802-3810` writes `last_heartbeat_at` immediately after registration, wrapped in WARNING. First health-check tick cannot see an empty heartbeat.
- [x] **Callback wiring** — Three closures `_on_sdk_started`/`_on_heartbeat_tick`/`_on_stdout_event` at `agent/agent_session_queue.py:4214-4249` pass into `BossMessenger()` at line 4251-4258. Harness subprocess in `agent/sdk_client.py:1722-1744` fires them via `messenger.notify_sdk_started(proc.pid)` on spawn and `messenger.notify_stdout_event()` on each stdout line. `BackgroundTask._watchdog` at `agent/messenger.py:282` invokes `notify_heartbeat_tick()` every 60s.
- [x] **Messenger ORM-free** — `grep '^from models' agent/messenger.py` returns no matches. Architectural boundary preserved.
- [x] **Heartbeat loop** — `_heartbeat_loop` at `agent/agent_session_queue.py:4490-4519` ticks at `HEARTBEAT_WRITE_INTERVAL` (60s) writing `last_heartbeat_at`, gating the 25-minute calendar-heartbeat work behind the `elapsed >= CALENDAR_HEARTBEAT_INTERVAL` reset — matches plan step 4.
- [x] **Kill path** — `agent/agent_session_queue.py:1900-2014` cancels `handle.task`, awaits 0.25s (`TASK_CANCEL_TIMEOUT`), bumps `recovery_attempts`, and finalizes as `failed` at `recovery_attempts >= MAX_RECOVERY_ATTEMPTS` (2). `StatusConflictError` caught around the transition.
- [x] **Kill-switch** — `os.environ.get("DISABLE_PROGRESS_KILL") == "1"` check at line 1881 suppresses the kill transition while still incrementing `tier1_flagged_total` and running Tier 2 evaluation. WARNING log line is distinctive.
- [x] **Recovery counter gated on kill only** — `recovery_attempts` is mutated exclusively at lines 1927 and 1982. Neither Tier 2 reprieve nor startup recovery touch it. `reprieve_count` is the separate Tier 2 counter.
- [x] **Local sessions get Tier 2 reprieves** — the Tier 2 block at 1844-1878 runs BEFORE the `is_local = worker_key.startswith("local")` check at line 1889, so local sessions benefit from reprieve gates before any `abandoned` transition.

**New AgentSession fields (B2 from critique)**
- [x] All five fields declared at `models/agent_session.py:264-278`: `last_heartbeat_at`, `last_sdk_heartbeat_at`, `last_stdout_at` (DatetimeField), `recovery_attempts`, `reprieve_count` (IntField default=0).
- [x] Three datetime fields added to `_DATETIME_FIELDS` at `models/agent_session.py:317-320` for auto-coercion.
- [x] All five fields added to `_AGENT_SESSION_FIELDS` at `agent/agent_session_queue.py:228-235` for round-trip. Verified via `python -c "... assert set(...).issubset(_AGENT_SESSION_FIELDS)"`.

**Metrics**
- [x] `tier1_flagged_total`, `tier2_reprieve_total:{alive|children|stdout}`, `kill_total` all incremented on the correct paths (lines 1850, 1862, 1931).

**Prior-fix invariants**
- [x] **#918 (`response_delivered_at` guard)** — block at `agent/agent_session_queue.py:1820-1842` runs BEFORE Tier 1/Tier 2 logic. Delivered sessions still short-circuit to `finalize_session("completed", ...)`.
- [x] **#944 (slugless dev session child-progress)** — `_has_progress()` own-progress branches at 1607-1618 preserved verbatim. Dual-heartbeat OR branch is prepended, not substitutive.
- [x] **#963 (child-activity branch in `_has_progress`)** — `entry.get_children()` check at line 1616-1618 preserved verbatim.
- [x] **#1006 (terminal-status zombie guard)** — preserved at the top of the health loop (verified via Grep above the checked block).

### Verification Results
- [x] **Unit tests**: `pytest tests/unit/test_health_check_recovery_finalization.py tests/unit/test_messenger_callbacks.py` — **44 passed**
- [x] **Integration tests**: `pytest tests/integration/test_session_heartbeat_progress.py` — **12 passed**
- [x] **Regression guard**: `pytest tests/unit/test_stall_detection.py tests/unit/test_agent_session_hierarchy.py` — **48 passed**
- [x] **Lint**: `python -m ruff check agent/agent_session_queue.py agent/messenger.py agent/sdk_client.py models/agent_session.py` — **All checks passed**
- [x] **Field presence**: `python -c "from models.agent_session import AgentSession; s=AgentSession(chat_id='x'); assert all(hasattr(s,f) for f in ('last_heartbeat_at','last_sdk_heartbeat_at','last_stdout_at','recovery_attempts','reprieve_count'))"` — **OK**
- [x] **Registry + _AGENT_SESSION_FIELDS**: `python -c "from agent.agent_session_queue import _active_sessions, SessionHandle, _AGENT_SESSION_FIELDS; ..."` — **OK**
- [x] **Messenger ORM boundary**: `grep '^from models' agent/messenger.py` — **empty** (architectural invariant preserved).

### Documentation
- [x] `docs/features/bridge-self-healing.md` — new "Two-tier no-progress detector" section at line 375 covering both tiers, kill path, kill-switch, metrics, and per-session fields.
- [x] `docs/features/bridge-worker-architecture.md` — new "Per-session registry (`_active_sessions`)" section at line 321 and messenger-callback plumbing table at line 349-351.
- [x] Inline docstrings on `_has_progress`, `_tier2_reprieve_signal`, `_execute_agent_session`, `SessionHandle`, `_active_sessions`, and all three messenger callback kwargs.

### Notes on design variants
One cosmetic difference from the plan: the plan specified the three callback kwargs live on `BackgroundTask`, but the implementation places them on `BossMessenger` instead (with `BackgroundTask._watchdog` invoking `messenger.notify_heartbeat_tick()`). This is equivalent in behavior — the messenger mediates all three callbacks cleanly, and the architectural invariant (messenger imports nothing from `models/`) is preserved. Not a concern.

Closes #1036.

valorengels:

The two-tier no-progress detector is well-implemented with strong test coverage and clean architectural separation. All 44 unit + 12 integration tests pass, lint is clean, and the messenger/ORM boundary is preserved (verified via AST scan test). The plan's success criteria are met: dual-heartbeat OR semantics, three Tier 2 reprieve gates, `DISABLE_PROGRESS_KILL` kill-switch, `recovery_attempts`/`reprieve_count` counters, and all five new fields round-trip through `_AGENT_SESSION_FIELDS`.

Two tech-debt items around Tier 2 scope/metrics semantics should be addressed before merge.

### Verified
- [x] All 5 new AgentSession fields present and in `_AGENT_SESSION_FIELDS`
- [x] `_active_sessions` registry + `SessionHandle` dataclass present
- [x] Dual-heartbeat OR in `_has_progress` with 90s freshness window
- [x] Three Tier 2 reprieve gates (alive/children/stdout) in `_tier2_reprieve_signal`
- [x] Messenger imports nothing from `models/` (test_messenger_callbacks.py asserts via AST)
- [x] Registration precedes any raise site (line 3791) with `add_done_callback` cleanup
- [x] T+0 heartbeat write at line 3803-3810 before health check could tick
- [x] `_watchdog` fires `notify_heartbeat_tick()` each 60s (`agent/messenger.py:282`)
- [x] `on_sdk_started`/`on_stdout_event` wired through `_run_harness_subprocess`
- [x] `TASK_CANCEL_TIMEOUT=0.25`, `MAX_RECOVERY_ATTEMPTS=2`, `HEARTBEAT_FRESHNESS_WINDOW=90` all match plan
- [x] `StatusConflictError` caught around transition (line 2009-2014)
- [x] `response_delivered_at` guard (#918) still runs before Tier 1/2 evaluation
- [x] Child-activity branch (#963) preserved verbatim in `_has_progress`
- [x] Terminal-status zombie guard (#1006) unchanged at top of health loop
- [x] Startup recovery does NOT touch `recovery_attempts` (Risk 3, test_startup_recovery_does_not_touch_recovery_attempts)
- [x] Documentation sections added to `bridge-self-healing.md` and `bridge-worker-architecture.md`

### Tech Debt

- [ ] **`agent/agent_session_queue.py:1844-1852`** — `# === Two-tier no-progress detector (#1036) ===\n# Tier 1 already flagged stuck via `_has_progress()` returning False.` — This comment is only true for the `no_progress` recovery reason. The outer `if should_recover:` block (line 1798) fires for three distinct reasons: `no_progress` (where `_has_progress()` was consulted), `worker_dead` (checked at line 1756, does NOT consult `_has_progress`), and `timeout` (line 1794, also does NOT consult `_has_progress`). The comment and the `tier1_flagged_total` metric at line 1850 misleadingly imply all three code paths went through Tier 1. This pollutes the metric's meaning and misleads readers of the recovery branch. Either (a) gate the Tier 1/Tier 2 block with `if _reason_kind == "no_progress":` and fall through to the direct kill path for timeout/worker_dead, or (b) rename the metric/comment to reflect "any recovery reason" and add clarifying branches.

- [ ] **`agent/agent_session_queue.py:1854-1878`** — `reprieve = _tier2_reprieve_signal(handle, entry)` — Tier 2 reprieve applies to `timeout` recoveries too. An actively-running session that has children (so `reprieve == "children"`) will evade the `AGENT_SESSION_TIMEOUT` cap indefinitely — every 300s health tick it gets a reprieve because tool subprocesses are still spawning. This defeats the timeout's purpose. The plan's design intent was to prevent false-positive kills of **stuck** sessions, not to disable timeouts on long-running sessions. Recommend scoping Tier 2 to the `no_progress` case only (paired with the fix above).

### Nits

- [ ] **`agent/agent_session_queue.py:1855`** — `reprieve = _tier2_reprieve_signal(handle, entry)` — Even when `handle is None` (session not in registry — e.g. a session picked up before registration completed, or one the health check observes from a different event loop), `_tier2_reprieve_signal` still falls through to the stdout check. This is correct behavior per the docstring, but the line above (`handle = _active_sessions.get(...)`) doesn't log when handle is None. Consider `logger.debug("[session-health] No registry handle for %s; Tier 2 will use stdout gate only", entry.agent_session_id)` for diagnosability — when handle is `None` but session is in fact active elsewhere, this is the only forensic signal.

- [ ] **`docs/features/bridge-self-healing.md:397-410`** — The "Tier 2 — activity-positive reprieve gates" section describes the three gates but doesn't mention that Tier 2 currently applies to ALL recovery reasons (timeout, worker_dead) not just no_progress. Worth clarifying once the tech-debt item above is resolved.

### Screenshots
No UI changes in this PR — screenshots not applicable.

valorengels:
## Review: Changes Requested

Implementation is thorough and well-tested: 35 unit + 12 integration tests all pass, 44-test target suite runs in ~0.2s, 150-test adjacent suite passes, ruff check/format clean. The two-tier detector is correctly wired end-to-end (`_has_progress` dual-OR, `_tier2_reprieve_signal` gates, metrics counters, kill-switch env var, `recovery_attempts` terminal finalization, T+0 heartbeat write, messenger ORM-free boundary, all 5 fields in `_AGENT_SESSION_FIELDS`). Plan's B2 round-trip concern, #918 `response_delivered_at` guard, #944/#963/#1006 invariants are all preserved.

However, the `SessionHandle.task` registration has a **cancellation-scope bug** that directly contradicts plan spike-1's stated design.

### Blockers

- [ ] **`agent/agent_session_queue.py:3788`** — `_current_task = asyncio.current_task()` — **`SessionHandle.task` points at the worker loop, not a session-scoped task.** `_execute_agent_session` is `await`ed directly from `_worker_loop` at line 2885 without any `asyncio.create_task` wrapper, so `asyncio.current_task()` returns the long-lived `_worker_loop()` task. Consequences:
  1. **Cancellation kills the whole worker.** When the health check calls `handle.task.cancel()` + `await asyncio.wait_for(handle.task, timeout=0.25)` at `agent_session_queue.py:1903-1908`, `CancelledError` propagates through `_execute_agent_session`, is caught by the worker's `except asyncio.CancelledError` handler at line 2887, and is re-raised at line 2905 — exiting the worker loop entirely. Subsequent sessions on the same `worker_key` must wait for `_ensure_worker` to respawn. Plan spike-1 explicitly warned against this ("Cancelling the worker task would tear down the loop and trigger startup recovery — heavy-handed and also racy") and prescribed using `BackgroundTask._task` (a task created via `asyncio.create_task` in `messenger.py:198`) instead, which cancels only the SDK work.
  2. **Registry entries leak across sessions.** `_cleanup_registry` is registered via `_current_task.add_done_callback(...)`, which only fires when the worker task completes. Between sessions, the worker loop keeps running, so `_active_sessions[prev_session_id]` remains populated with a handle whose `.task` now actually belongs to whatever session the worker is currently processing. Each new session adds *another* `add_done_callback` to the same worker task — unbounded accumulation over the worker's lifetime. Eventually all callbacks fire together when the worker exits, so the registry does get cleaned, but the intermediate state is incorrect.
  3. **The #1006 zombie guard masks the defect.** In current production code paths, the zombie guard at line 1736 skips terminal sessions appearing in the running index, so a stale-entry-plus-mis-routed-cancel scenario is unlikely in practice. But the cancellation-kills-worker side effect fires on *every* real kill, not just stale ones.

  **Fix**: Either (a) wrap `_execute_agent_session`'s body (or just the `await task._task` region) in `asyncio.create_task` and register that task, or (b) register `task._task` directly after `BackgroundTask.run()` populates it — the `BackgroundTask._task` is already a discrete task (see `messenger.py:198`) and cancelling it terminates only the SDK work without tearing down the worker loop. Option (b) matches plan spike-1's design intent.

### Verified

- [x] All 5 new fields present on `AgentSession` and in `_AGENT_SESSION_FIELDS` (B2 regression guard via `TestAgentSessionFieldsRoundTrip`).
- [x] `agent/messenger.py` imports nothing from `models/` — verified by AST scan in `TestMessengerArchitecturalBoundary`.
- [x] Three optional callbacks (`on_sdk_started`, `on_heartbeat_tick`, `on_stdout_event`) are wired via `notify_*` wrappers that catch exceptions and log WARNING.
- [x] `_has_progress` correctly implements dual-heartbeat OR semantics (90s window).
- [x] `_tier2_reprieve_signal` returns `"children"` in preference to `"alive"`, handles `psutil.NoSuchProcess`/`AccessDenied`/`ImportError` gracefully, returns `None` on zombie/dead/stopped.
- [x] `DISABLE_PROGRESS_KILL=1` kill-switch checked before any destructive action; logs WARNING and `continue`s; metrics still increment.
- [x] `recovery_attempts` increments only on actual kills (not reprieves, not startup recovery — confirmed via `test_startup_recovery_does_not_touch_recovery_attempts` which AST-scans the source).
- [x] `MAX_RECOVERY_ATTEMPTS = 2` finalizes as `failed` via `finalize_session`, preserving lifecycle history.
- [x] Prior-fix invariants preserved: `response_delivered_at` guard (#918) runs BEFORE tier logic; `get_children()` branch (#963) unchanged; terminal-zombie guard (#1006) unchanged; slugless dev session semantics (#944) unchanged.
- [x] Local sessions: Tier 2 gates evaluated first, then `finalize_session("abandoned", ...)` on kill.
- [x] Metrics counters emitted per project_key: `tier1_flagged_total`, `tier2_reprieve_total:{alive|children|stdout}`, `kill_total`.
- [x] Docs updated: `docs/features/bridge-self-healing.md` has "Two-tier no-progress detector" section; `docs/features/bridge-worker-architecture.md` has `_active_sessions` registry + messenger callback plumbing.
- [x] Lint and format clean (`ruff check`, `ruff format --check`).

### Verification Results

All required plan verification table checks pass (parser had unrelated shell-quoting issues with a few `grep` commands; manual re-runs all pass):
- Unit tests pass (44/44 in target files).
- Integration tests pass (12/12).
- Adjacent suites pass (150/150 across `test_agent_session_queue`, `test_stall_detection`, `test_agent_session_hierarchy`, `test_pipeline_integrity`, `test_nudge_loop`, `test_session_completion_zombie`).
- All 5 new fields present and round-trip through `_AGENT_SESSION_FIELDS`.
- `_active_sessions` registry + `SessionHandle` exist.
- Messenger has callback kwargs (verified via Grep) and zero `from models` imports.
- `_tier2_reprieve_signal` exists; dual-heartbeat branches in `_has_progress`; `DISABLE_PROGRESS_KILL` handled.

### Legacy Cruft

None identified. No dead imports, no commented-out alternatives, no fallback chains. The implementation is additive and respects the documented architectural boundaries.

valorengels:

Re-review of the patch at `10daf7b0` plus docs cascade at `13b85f4a`. The prior blocker (`SessionHandle.task` capturing the worker-loop task) is **correctly fixed**. The prior tech-debt item (Tier 1/Tier 2 scope) is **correctly fixed**. The prior nits are addressed. Tests: 42/42 pass in the primary suite, 69/69 pass across adjacent suites (`test_messenger_callbacks`, `test_session_heartbeat_progress`, `test_stall_detection`, `test_agent_session_hierarchy`). Lint clean.

One remaining issue: two comment blocks in the patched region reference a cleanup mechanism that the patch deliberately replaced. These will mislead future readers. They are mechanical to fix — a single pass over the two comment blocks.

### Verified (prior review items)

- [x] **Blocker FIXED**: `SessionHandle.task: asyncio.Task | None = None` (line 2419). Registration at line 3835 uses `SessionHandle(task=None)`. After `await task.run(...)` at line 4546, the handle is populated via `task._task` at lines 4554-4556. The cancel path at line 1929 guards `handle is not None and handle.task is not None and not handle.task.done()`. `task._task` is created by `asyncio.create_task` at `agent/messenger.py:198` — a true session-scoped task, cancelling it terminates only the SDK work (confirmed by reading `messenger.py:183-203`).
- [x] **Tech debt FIXED**: Tier 1 `tier1_flagged_total` metric and Tier 2 reprieve gated behind `if _reason_kind == "no_progress":` at line 1866. `worker_dead` and `timeout` cases now fall through to the kill path unchanged. `_reason_kind` classification hoisted to lines 1806-1811 (before the metric try block, where the old classification was dead on failure).
- [x] **Nit FIXED**: Debug log when `handle is None` present at lines 1860-1865.
- [x] **Nit FIXED**: `docs/features/bridge-self-healing.md` updated. Docs cascade commit `13b85f4a` also updated `agent-session-health-monitor.md`, `session-recovery-mechanisms.md`, `bridge-resilience.md`, `features/README.md`, `tests/README.md`.
- [x] **Legacy cruft**: No `asyncio.current_task()` references remain in `agent_session_queue.py`. The pre-patch pattern was cleanly excised.
- [x] **Cleanup lifecycle**: `_active_sessions.pop(_session_id_for_registry, None)` runs in an explicit `finally` at line 4810 that wraps the entire session body (try opens at line 3836). This is a stronger guarantee than the prior `add_done_callback`: it fires on normal return, exception, and CancelledError without relying on task completion.
- [x] **New test coverage**: 7 new tests in `TestSessionHandleTaskInvariant` (4 tests: default-None, is-None-before-run, cancelling-handle-doesnt-kill-worker, populated-after-run) and `TestReprieveScopedToNoProgress` (3 tests: reprieve-only-on-no-progress, tier1-metric-only-on-no-progress, handle-none-debug-log-present). All pass.

### Tech Debt

- [ ] **`agent/agent_session_queue.py:3806-3808`** — `Cleanup is tied to the asyncio task completion via add_done_callback so the registry entry is always popped` — This docstring describes the pre-patch mechanism. The patched code uses a plain `try`/`finally` block (opened at line 3836, cleanup at line 4810) — there is no `add_done_callback` for registry cleanup anywhere in the function. Future maintainers will grep for a callback that isn't there. Fix: rewrite to `Cleanup is performed in a finally block wrapping the entire session body so the registry entry is always popped — even on exception, CancelledError, or any early return.`
- [ ] **`agent/agent_session_queue.py:3828-3832`** — `Cleanup is performed via an add_done_callback attached to the BackgroundTask's session-scoped task below, plus a defensive finally via _session_registry_cleanup for the pre-BackgroundTask window.` — Same stale-mechanism issue, plus a reference to a function `_session_registry_cleanup` that does not exist (`grep _session_registry_cleanup` returns only this comment). The "without requiring a 900-line re-indent" clause is also misleading: the `try` at line 3836 does in fact wrap the session body (≈970 lines), so the re-indent apparently was done. Fix: delete this comment block or replace with `Cleanup runs in the try/finally at line ~3836/~4810, guaranteeing the registry pop on every exit path.`

### Verification Results

All plan verification requirements pass:
- 42/42 unit tests in `test_health_check_recovery_finalization.py` (0.20s).
- 69/69 tests across adjacent suites (`test_messenger_callbacks`, `test_session_heartbeat_progress`, `test_stall_detection`, `test_agent_session_hierarchy`).
- `ruff check` clean on all modified files.
- All 5 new `AgentSession` fields present and in `_AGENT_SESSION_FIELDS`.
- `_active_sessions` registry + `SessionHandle` exist with correct semantics.
- Messenger has the three optional callbacks with exception-safe wrappers; no `from models` import.
- `_tier2_reprieve_signal` exists; `_has_progress` dual-heartbeat OR semantics correct; `DISABLE_PROGRESS_KILL` handled.
- Plan spike-1 design intent met: cancelling `handle.task` targets only the SDK work, not the worker loop.

### Legacy Cruft

No code-level cruft. The only cruft is the two stale-mechanism comment blocks listed above (which the patch introduced as part of its commit message but should have deleted or rewritten to match the final implementation).

valorengels:
## Review: Approved

Final re-review after comment-cleanup commit `7282fb36`. All prior blockers and tech debt are resolved; docstring and inline comments in `_execute_agent_session` now accurately describe the current explicit `try`/`finally` cleanup mechanism.

### What changed in 7282fb36

Two comment blocks in `agent/agent_session_queue.py` were rewritten:

1. **Docstring at lines 3806–3808** — previously described the `add_done_callback` mechanism that was replaced in `10daf7b0`. Now reads: *"Cleanup happens in an explicit `finally` block wrapping the entire session body — `_active_sessions.pop()` runs on every exit path (return, raise, `CancelledError`)."*
2. **Inline comment at lines 3828–3831** — previously referenced the removed `_session_registry_cleanup` helper and the "pre-BackgroundTask window" defensive path. Now reads: *"Cleanup is guaranteed by the `finally` block wrapping the entire session body below: `_active_sessions.pop()` runs on every exit path (return, raise, `CancelledError`), satisfying the 'pop in a finally block, always' contract from the SessionHandle docstring."*

### Verified the comments match reality

- `finally:` block wrapping the session body is at **line 4803**.
- `_active_sessions.pop(_session_id_for_registry, None)` is at **line 4809** inside that `finally`.
- Grep for `add_done_callback|_session_registry_cleanup` returns only a single unrelated match (`_starting_workers.discard` at line 2739). The session-registry cleanup callback and helper are fully gone.

### Verified
- [x] Code correctness — comments accurately describe the active code path
- [x] Test coverage — 42/42 `test_health_check_recovery_finalization.py` pass, plus 9 `test_messenger_callbacks.py` and 12 `test_session_heartbeat_progress.py` (63 total, zero regressions)
- [x] Security — documentation-only commit, no behavioral change
- [x] Plan requirements met — two-tier detector, dual heartbeat, Tier 2 reprieve gates, recovery_attempts, kill-switch, messenger callbacks, metrics, and fields all present and tested
- [x] Code quality — `ruff check` and `ruff format --check` both clean
- [x] No legacy traces — all references to the removed `add_done_callback` + `_session_registry_cleanup` cleanup mechanism are gone from comments

### Prior findings status
- Prior blocker (`SessionHandle.task` capturing the worker-loop task) — fixed in `10daf7b0`.
- Prior tech debt (stale comments describing the removed callback cleanup) — fixed in `7282fb36`.

No new findings. PR is ready to merge.

Issue #917, opened 2026-04-12
Bug: health-check-recovered sessions not finalized — causes duplicate Telegram delivery


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/health-check-recovery-finalization.md

## Recon Summary

**Confirmed:**
- Root cause verified: `complete_transcript()` at L3364 is inside `if agent_session:` guard. When inner lookup (L2932, filters `status="running"` by session_id) returns None, `complete_transcript` is skipped entirely.
- `finalized_by_execute` flag confirmed (L2174): Set to `True` on normal return from `_execute_agent_session()`, suppresses outer `finally` block — double-finalization guard from #898 works but silently drops finalization when inner lookup fails.
- Health check cycle confirmed (L1341): Detects `running` sessions with dead PIDs every 5 min, resets to `pending`. An un-finalized completed session looks identical to a stuck one.
- Incident confirmed: Session `c0642630` (thread 8866) looped repeatedly sending duplicate messages.

**Revised:**
- Solution Option B (structural) is preferred: move finalization out of `if agent_session:` guard, use outer `session` parameter (always available) to call `finalize_session()` directly.
- Must preserve nudge path: sessions in nudge/defer state set by `_enqueue_nudge` must not be prematurely finalized — guard by checking current Redis status before forcing completion.

**Pre-requisites:**
- None. #700/#703 and #898 are merged and stable. This is an isolated fix in `_execute_agent_session()`.

**Dropped:**
- Option A (post-return Redis check) — viable but adds a second Redis round-trip. Option B (structural fix) is cleaner.

> **Context:** The worker ([`agent/agent_session_queue.py`](https://github.com/tomcounsell/ai/blob/main/agent/agent_session_queue.py)) manages agent session lifecycle through Redis via Popoto ORM. Sessions transition `pending → running → completed/failed`. A health check ([`_agent_session_health_check()`](https://github.com/tomcounsell/ai/blob/main/agent/agent_session_queue.py#L1341)) runs every 5 minutes to detect stuck `running` sessions whose worker process has died, resetting them to `pending` for re-execution. This is distinct from the zombie loop fixed in [#700](https://github.com/tomcounsell/ai/issues/700), which was caused by the hierarchy health check resetting completed sessions.

## Problem

[...]

PR #1056, opened 2026-04-19
Unblock worker event loop — memory extraction hotfix (Layers 1+2)

### LINKED ISSUE #1055: Sync Anthropic calls in memory_extraction freeze the worker event loop and block session finalization
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/unblock-worker-event-loop-hotfix.md

> **Context:** This repo runs a [standalone worker process](worker/__main__.py) (`python -m worker`) that owns a single `asyncio.run()` event loop. All session execution, per-`worker_key` loops, the session health monitor, the [reflection scheduler](agent/reflection_scheduler.py), a pub/sub notify listener, and per-session heartbeat tasks share that one loop. An "AgentSession" is the worker-owned record representing one unit of agent work — created by the bridge, picked up by the worker, driven to `completed`/`failed`. Sessions transition through a lifecycle (`pending` → `running` → `completed`/`failed`); the monitoring watchdog flags sessions whose `running` state exceeds 45 min. The PM→Dev steering model in [pm-dev-session-architecture.md](docs/features/pm-dev-session-architecture.md) relies on the Dev session being *finalized* before the PM is nudged — if finalization stalls, steering orphans.

## Problem

Sessions that finish their real work and successfully deliver a result to the user can still end up frozen in `status="running"` for hours. When this happens it isn't a single stuck session — the entire worker event loop freezes, so heartbeats stop, other projects' sessions stop making progress, the reflection scheduler pauses, and the session-notify listener stops waking. The watchdog logs `LIFECYCLE_STALL` but takes no corrective action, so the only recovery is a manual worker restart.

**Current behavior:**

Observed on session `tg_valor_-1003449100931_681`:

- `05:58:31 UTC` — result delivered to Telegram, `Sent result message` logged
- `05:58:31 → 11:56:37 UTC` — **zero worker log entries for ~6 hours**: no SDK heartbeats, no calendar heartbeat, no `last_heartbeat_at` writes, no reflection scheduler ticks, no LIFECYCLE transition
- Session remained `status="running"` the whole time
- Bridge-hosted watchdog reported `LIFECYCLE_STALL duration=21490s+` (log-only; no action)
- Any follow-up user messages that tried to steer the session were orphaned against a session that looked alive but was frozen

Traced root cause:

- [`agent/memory_extraction.py`](agent/memory_extraction.py) has three synchronous `client.messages.create()` calls at lines **100–114**, **277–289**, and **412–417**. Each constructs `anthropic.Anthropic(...)` (the blocking HTTP client) inside an `async def` function with **no** `asyncio.to_thread`/`run_in_executor` wrap and **no** explicit `timeout=` kwarg.
- The call path is: `session_executor._execute_agent_session` → [`await task._task`](agent/session_executor.py#L1191) → `BackgroundTask._run_work` → [`await run_post_session_extraction(...)`](agent/messenger.py#L218) → the sync HTTP call.
- Because the worker is one asyncio loop, a blocking HTTP call on that loop starves *every* cooperating task in the worker, not just the session that made the call.
- [`complete_transcript(...)`](agent/session_executor.py#L1224) is scheduled **after** `await task._task` at line 1191, so while `_run_work` is blocked the session is never finalized. [`_handle_dev_session_completion(...)`](agent/session_executor.py#L1293) (which nudges the PM) runs even later, so dev→PM steering is also blocked.
- [`monitoring/session_watchdog.py:236`](monitoring/session_watchdog.py#L236) `check_stalled_sessions()` logs `LIFECYCLE_STALL` for `status="running"` past `STALL_THRESHOLD_RUNNING=2700s` but **does not** cancel, re-queue, or fail the session.

**Desired outcome:**

- Post-session memory extraction never blocks the worker event loop: all Anthropic calls from async code are async-native (`AsyncAnthropic`) or wrapped via `asyncio.to_thread`, and every call has a hard `asyncio.wait_for` + SDK `timeout` bound.
- Memory extraction is *not* a prerequisite for session finalization. A hang (of any duration, from any cause) in extraction must not prevent `complete_transcript()` or dev→PM steering from running.
- The stalled-running case has a recovery path beyond logging: the session watchdog either cancels the hung task or transitions the session to `failed`/`pending` within a bounded time.
- A repo-level guardrail prevents re-introducing sync `anthropic.Anthropic()` call sites from async-reachable modules.

## Definitions

| AgentSession | Popoto-backed record representing one unit of agent work; owns status, lifecycle history, worker_key, and steering inbox. | [models/agent_session.py](models/agent_session.py), [session-lifecycle.md](docs/features/session-lifecycle.md) |
| Worker | Standalone process (`python -m worker`) that pops AgentSessions from Redis and drives them through the CLI harness. Single `asyncio.run()`. | [worker/__main__.py](worker/__main__.py), [bridge-worker-architecture.md](docs/features/bridge-worker-architecture.md) |
| `_execute_agent_session` / `BackgroundTask._run_work` | The worker-side coroutine pair that runs the CLI harness, sends the result to Telegram, then performs post-session bookkeeping. | [agent/session_executor.py](agent/session_executor.py), [agent/messenger.py](agent/messenger.py) |
| `complete_transcript` | Writes the `SESSION_END` transcript marker and transitions the session to a terminal status via `finalize_session()`. | [bridge/session_transcript.py](bridge/session_transcript.py) |
| Memory extraction (post-session) | Haiku-based extraction of novel observations plus outcome detection for injected thoughts; runs after the agent's final message. | [agent/memory_extraction.py](agent/memory_extraction.py), [subconscious-memory.md](docs/features/subconscious-memory.md) |
| PM → Dev steering | Pattern where a PM-role AgentSession spawns a Dev-role AgentSession; when the Dev terminates, `_handle_dev_session_completion` nudges the PM with the result. | [pm-dev-session-architecture.md](docs/features/pm-dev-session-architecture.md), [agent/session_executor.py](agent/session_executor.py) |
| `LIFECYCLE_STALL` | Log marker from the bridge-hosted watchdog when a session exceeds the per-status threshold (`running`: 2700s). Advisory only — no auto-recovery. | [monitoring/session_watchdog.py](monitoring/session_watchdog.py) |

## Prior Context

Related work:
- #917: Bug: health-check-recovered sessions not finalized — introduced the `else` fallback at [session_executor.py:1244](agent/session_executor.py#L1244) that also calls `complete_transcript()` after `await task._task`. That finalization path suffers the same ordering dependency as the main path.
- #987: Race where steering was accepted, then `_finalize_parent_sync` ran and orphaned the message. The fix (ordering `_handle_dev_session_completion` after `complete_transcript`) is what this bug ultimately undoes when extraction stalls, because finalization itself never runs.
- #1019: Worker lifecycle audit open investigations — broader worker robustness context. This issue is narrower and independently fixable.

## Solution Sketch

Four layers, in priority order. Each layer is a discrete milestone the planner can scope into steps/PRs.

1. **Unblock the event loop (minimum viable fix).** At all three sites in `agent/memory_extraction.py`, switch to `anthropic.AsyncAnthropic` and wrap the call in `asyncio.wait_for(..., timeout=30.0)` with an SDK-level `timeout` kwarg on `messages.create`. Existing `try/except Exception` at the outer scope already swallows failures.
2. **Decouple extraction from finalization.** Move the post-session extraction out of the critical path. Either: fire-and-forget via `asyncio.create_task` (with a guard so the worker doesn't exit while extractions are pending on shutdown), or run extraction *after* `complete_transcript()` in `_execute_agent_session`. The invariant: a hang in memory extraction cannot delay transcript finalization or dev→PM steering.
3. **Defense in depth.** Add a validator hook to `.claude/hooks/validators/` that rejects new occurrences of `anthropic.Anthropic(` inside files matching `agent/**`, `bridge/**`, `worker/**` unless paired with `asyncio.to_thread`/`run_in_executor` in the same function. Optionally require a `timeout=` kwarg on any `messages.create(`. Existing sync-in-async callers in this class to audit: [`bridge/routing.py:606`](bridge/routing.py#L606), [`bridge/routing.py:805`](bridge/routing.py#L805), [`tools/classifier.py:84`](tools/classifier.py#L84), [`tools/classifier.py:322`](tools/classifier.py#L322) — already-safe wrappers exist in [`agent/intent_classifier.py:204`](agent/intent_classifier.py#L204) and [`bridge/media.py:349`](bridge/media.py#L349) and can serve as reference patterns.
4. **Close the watchdog observability gap.** Extend `check_stalled_sessions()` so that `status="running"` past its threshold triggers an action: cancel the registered task (task handle is stashed at [session_executor.py:1144](agent/session_executor.py#L1144)) and transition the session to `failed` with a reason, or re-queue. The kill primitive already exists via `tools.agent_session_scheduler.kill`. Needs care not to kill sessions that are legitimately doing long sub-agent work (transcript-liveness check is already implemented; re-use it).

The planner should **not** attempt layers 3 and 4 in the same PR as layers 1 and 2.

## Recon Summary

Recon was performed inline during the architectural audit that produced this issue (see call-graph tracing below). A targeted parallel fan-out was not run because every concern was already confirmed against code by direct reading; the synthesis is captured here.

**Confirmed:**
- Three sync call sites in `agent/memory_extraction.py` at lines 100/105, 277/285, 412/413. No `to_thread`, no `timeout=`.
- Call path into event loop: `_execute_agent_session` → `await task._task` (session_executor.py:1191) → `_run_work` (messenger.py:205) → `await run_post_session_extraction` (messenger.py:218).
- Ordering dependency: `complete_transcript` at session_executor.py:1224 and :1250 is awaited after `await task._task`, so a hang in `_run_work` blocks finalization on both the happy path and the #917 fallback path.
- Single event loop blast radius: `worker/__main__.py:456` calls `asyncio.run(_run_worker(...))`. All of `_worker_loop`, `_agent_session_health_loop`, `ReflectionScheduler`, `_session_notify_listener`, per-session `_heartbeat_loop`, and per-session `BackgroundTask._watchdog` share that loop.
- Watchdog is log-only for `running`: `monitoring/session_watchdog.py:342` logs `LIFECYCLE_STALL`; `check_all_sessions` only acts on `status="active"`.

**Revised:**
- Original thought: "Haiku SDK default timeout should have fired after 10 min." Confirmed `anthropic` SDK defaults exist but half-open TCP and retry compound can exceed observed limits; empirically a 6h stall occurred. Fix must impose an explicit timeout rather than relying on SDK defaults.
- Bug report framed this as a memory-extraction bug; the real fragility is the architectural pattern (sync HTTP on shared asyncio loop + extraction-gates-finalization). Scope widened to include decoupling (layer 2) and guardrail (layer 3), but the event-loop fix (layer 1) is still the critical-path blocker.

**Pre-requisites:**
- None. The fix is self-contained and does not depend on other in-flight work.

**Dropped:**
- `scripts/memory_consolidation.py:195`, `reflections/utils.py:188` — sync calls, but invoked via the [reflection scheduler's `run_in_executor` path](agent/reflection_scheduler.py#L262), so they don't block the event loop. Not in scope; leave as-is.
- `agent/intent_classifier.py:204`, `bridge/media.py:349` — already wrapped via `to_thread`/`run_in_executor`. They are the reference pattern for layer 3, not bugs to fix.

## Acceptance Criteria

- [ ] All three call sites in `agent/memory_extraction.py` use `anthropic.AsyncAnthropic` with an explicit `asyncio.wait_for(..., timeout=...)` and an SDK-level `timeout=` kwarg.
- [ ] `grep -n "anthropic\.Anthropic(" agent/ worker/` returns zero matches in non-test code (production modules reachable from the worker event loop).
- [ ] A hang or timeout in `run_post_session_extraction` does not block `complete_transcript()` from running — verified by an integration test that stubs the Haiku client to sleep past the timeout and asserts `status` becomes `completed` or `failed` within a bounded window.
- [ ] A unit/integration test covers: memory extraction raising `TimeoutError` does not propagate out of `_run_work` as an unhandled exception, and the session still finalizes.
- [ ] The session watchdog's `check_stalled_sessions()` has a concrete action for `status="running"` past threshold (cancel or fail), not just a log line. Behavior covered by a test.
- [ ] A repository-level guardrail (validator hook or equivalent) exists that fails CI / pre-commit if a new `anthropic.Anthropic(` appears in `agent/`, `bridge/`, or `worker/` without a wrapping executor.
- [ ] Documentation in `docs/features/subconscious-memory.md` notes the async/timeout requirement for all Anthropic calls made from the worker loop.

## Downstream

This issue feeds into `/do-plan`, which will produce a plan document at `docs/plans/{slug}.md`. The plan should split the four layers into at least two PRs (layer 1+2 first as the hotfix, layer 3+4 as the hardening follow-up). Constraints for the planner:

- Layer 1+2 must be safe to deploy independently of layers 3 and 4.
- No new external dependencies — the `anthropic` package already ships `AsyncAnthropic`.
- Existing `try/except Exception` handlers in `memory_extraction.py` must be preserved; failures remain non-fatal.
- Layer 4's watchdog action must respect transcript liveness (`_check_transcript_liveness` in `monitoring/session_watchdog.py`) so long-running sub-agent work is not mistakenly killed.

tomcounsell:
## Summary

Fixes #1055: the worker process froze for ~6 hours when a synchronous Anthropic HTTP call inside `agent/memory_extraction.py` stalled on a half-open TCP socket. During the stall, heartbeats stopped, the reflection scheduler paused, session-notify events were missed, and a dev session's follow-up messages were orphaned against a "running" session that was actually frozen.

Two layers of fix, both independently deployable:

**Layer 1 — Event-loop unblock** (`agent/memory_extraction.py`): all three sync `anthropic.Anthropic(...)` call sites are converted to `async with anthropic.AsyncAnthropic(...) as client:` with a double-timeout (SDK `timeout=30s` + outer `asyncio.wait_for(timeout=35s)`). `async with` releases httpx resources deterministically on timeout. A new `_record_extraction_error()` helper emits `memory.extraction.error` analytics counters on every error path (except `CancelledError`) so silent extraction failures are visible on `/dashboard.json`.

**Layer 2 — Decouple extraction from finalization** (`agent/session_executor.py`, `agent/messenger.py`, `worker/__main__.py`): post-session extraction is no longer awaited inside `BackgroundTask._run_work`. Instead, `_execute_agent_session` calls a **synchronous** `_schedule_post_session_extraction(...)` (fire-and-forget `asyncio.create_task`) AFTER `complete_transcript` runs and BEFORE `_handle_dev_session_completion`. A hang in extraction cannot delay session finalization or the dev→PM nudge. `drain_pending_extractions(timeout=5.0)` is wired into the worker shutdown sequence.

## Changes

- `agent/memory_extraction.py`: three call sites converted to `AsyncAnthropic` with double-timeout; new module-level `_EXTRACTION_SDK_TIMEOUT=30.0` and `_EXTRACTION_HARD_TIMEOUT=35.0` constants; new `_record_extraction_error` helper; `_judge_outcomes_llm` converted to `async def`; outer `except` branches now record analytics counters.
- `agent/messenger.py`: `_run_work` no longer awaits `run_post_session_extraction`; docstring updated to prevent re-introduction.
- `agent/session_executor.py`: new module-level `_pending_extraction_tasks: dict[str, asyncio.Task]` keyed by `session_id`; new synchronous `_schedule_post_session_extraction(...)` helper with dedup; new async `drain_pending_extractions(...)` helper; scheduler wired into `_execute_agent_session` between `complete_transcript` and `_handle_dev_session_completion`.
- `worker/__main__.py`: `drain_pending_extractions(timeout=5.0)` called during shutdown after worker-task drain and before health/notify/reflection cancels.
- `docs/features/subconscious-memory.md`: new "Event-Loop Safety (hotfix #1055)" subsection documenting the async-native invariant, double-timeout rationale, fire-and-forget ordering, graceful-shutdown drain, loss tolerance, orphan safety, and `memory.extraction.error` counter.

## Testing

**90 tests across 4 files, all pass:**

- `tests/unit/test_memory_extraction.py` — new `TestEventLoopSafety` class with 5 cases (hard-timeout caught in each of the three extraction paths; APITimeoutError handled; regression canary grep), plus `test_extract_post_merge_learning_runs_inside_asyncio_run` guarding the hook-subprocess `asyncio.run` call path. Existing mocks updated from `anthropic.Anthropic` → `anthropic.AsyncAnthropic`; `TestJudgeOutcomesLlm` updated to async.
- `tests/unit/test_session_executor_extraction_decoupling.py` (NEW) — 6 cases: extraction errors don't propagate, PM nudge fires while extraction is pending, duplicate schedules deduped, drain is a no-op when empty, scheduler is sync (not async), cancellation semantics preserved.
- `tests/integration/test_session_finalization_decoupled.py` (NEW) — 3 cases: 5-second SLO for scheduler + PM nudge even with a 30s extraction stall, drain cancels cooperative tasks within budget, cancellation raises cleanly.
- `tests/unit/test_messenger.py` — 13 existing tests all pass (BackgroundTask refactor did not break the contract).

**Regression canary:** `grep -n "anthropic\\.Anthropic(" agent/memory_extraction.py` returns exit 1 (no matches) — enforced by a unit test.

**Unrelated pre-existing failures** exist in `test_agent_session_scheduler_kill`, `test_worker_entry::test_cleanup_orphaned_in_agent_queue_not_bridge`, and `test_work_request_classifier` — verified to fail on main as well, not caused by this hotfix.

## Documentation

- [x] `docs/features/subconscious-memory.md` updated per plan — new "Event-Loop Safety (hotfix #1055)" subsection under Flow 3, tracking entry added.

## Definition of Done

- [x] Built: all three layers implemented; imports clean
- [x] Tested: 90 hotfix-related tests pass, including the 5-second SLO integration test
- [x] Documented: subconscious-memory.md updated with the new invariants
- [x] Quality: `python -m ruff format .` clean; no new lint issues

Closes #1055

### DIFF
--- a/agent/memory_extraction.py
+++ b/agent/memory_extraction.py
@@ -9,10 +9,19 @@ into ObservationProtocol.
 
 All operations are async, wrapped in try/except — failures must never
 crash the agent or block session completion.
+
+Event-loop safety invariant (hotfix #1055):
+    Every Anthropic call in this module MUST use ``anthropic.AsyncAnthropic``
+    inside ``async with`` for deterministic httpx cleanup, wrapped in
+    ``asyncio.wait_for(..., timeout=_EXTRACTION_HARD_TIMEOUT)`` with an
+    SDK-level ``timeout=_EXTRACTION_SDK_TIMEOUT`` kwarg. Sync ``anthropic.Anthropic``
+    is forbidden here — it blocks the worker event loop for arbitrary durations
+    on half-open TCP sockets (empirically observed: 6-hour stall on #1055).
 """
 
 from __future__ import annotations
 
+import asyncio
 import json
 import logging
 import os
@@ -20,6 +29,47 @@ import re
 import time
 
 logger = logging.getLogger(__name__)
+
+# Timeout constants for Anthropic extraction calls (hotfix #1055).
+#
+# Double-timeout rationale:
+#   - _EXTRACTION_SDK_TIMEOUT: passed to messages.create(timeout=...) so the
+#     Anthropic SDK (httpx) raises APITimeoutError on idle/slow sockets.
+#   - _EXTRACTION_HARD_TIMEOUT: outer asyncio.wait_for; fires from a separate
+#     asyncio timer so it still fires even when the SDK timer does not (e.g.,
+#     half-open TCP sockets where no socket event ever arrives).
+#   - 5s buffer lets the SDK raise its own typed error first for cleaner logs.
+_EXTRACTION_SDK_TIMEOUT = 30.0
+_EXTRACTION_HARD_TIMEOUT = 35.0
+
+
+def _record_extraction_error(
+    error_class: str,
+    session_id: str,
+    project_key: str | None = None,
+) -> None:
+    """Emit ``memory.extraction.error`` analytics counter (hotfix #1055).
+
+    Non-fatal — silent if analytics unavailable. Skips ``CancelledError`` which
+    is expected on worker shutdown and carries no signal. Every ``except``
+    branch in this module's three async-Anthropic call sites must call this.
+    """
+    if error_class == "CancelledError":
+        return
+    try:
+        from analytics.collector import record_metric
+
+        record_metric(
+            "memory.extraction.error",
+            1.0,
+            {
+                "error_class": error_class.lower(),
+                "session_id": session_id,
+                "project_key": project_key or "",
+            },
+        )
+    except Exception:
+        pass
 
 # Extraction prompt for Haiku — structured JSON output
 EXTRACTION_PROMPT = (
@@ -97,21 +147,38 @@ async def extract_observations_async(
             logger.warning("[memory_extraction] No Anthropic API key, skipping extraction")
             return []
 
-        client = anthropic.Anthropic(api_key=api_key)
-
         # Truncate response to avoid token limits
         truncated = response_text[:8000]
 
-        message = client.messages.create(
-            model=MODEL_FAST,
-            max_tokens=500,
-            messages=[
-                {
-                    "role": "user",
-                    "content": f"{EXTRACTION_PROMPT}\n\n---\n\n{truncated}",
-                }
-            ],
-        )
+        # hotfix #1055: AsyncAnthropic + double-timeout + async with for httpx cleanup.
+        # Sync anthropic.Anthropic is forbidden here — it blocks the worker event loop.
+        try:
+            async with anthropic.AsyncAnthropic(
+                api_key=api_key, timeout=_EXTRACTION_SDK_TIMEOUT
+            ) as client:
+                message = await asyncio.wait_for(
+                    client.messages.create(
+                        model=MODEL_FAST,
+                        max_tokens=500,
+                        messages=[
+                            {
+                                "role": "user",
+                                "content": f"{EXTRACTION_PROMPT}\n\n---\n\n{truncated}",
+                            }
+                        ],
+                        timeout=_EXTRACTION_SDK_TIMEOUT,
+                    ),
+                    timeout=_EXTRACTION_HARD_TIMEOUT,
+                )
+        except asyncio.TimeoutError:
+            logger.warning(
+                "[memory_extraction] Anthropic call exceeded %.1fs hard timeout (non-fatal); "
+                "extraction skipped for session_id=%s",
+                _EXTRACTION_HARD_TIMEOUT,
+                session_id,
+            )
+            _record_extraction_error("TimeoutError", session_id, project_key)
+            return []
 
         raw_text = message.content[0].text.strip()
 
@@ -171,6 +238,7 @@ async def extract_observations_async(
 
     except Exception as e:
         logger.warning(f"[memory_extraction] Extraction failed (non-fatal): {e}")
+        _record_extraction_error(type(e).__name__, session_id, project_key)
         return []
 
 
@@ -274,19 +342,39 @@ async def extract_post_merge_learning(
             )
             return None
 
-        client = anthropic.Anthropic(api_key=api_key)
-
         prompt = POST_MERGE_EXTRACTION_PROMPT.format(
             title=pr_title,
             body=(pr_body or "")[:4000],
             diff_summary=(diff_summary or "")[:4000],
         )
 
-        message = client.messages.create(
-            model=MODEL_FAST,
-            max_tokens=200,
-            messages=[{"role": "user", "content": prompt}],
-        )
+        # hotfix #1055: AsyncAnthropic + double-timeout + async with.
+        # Called from .claude/hooks/hook_utils/memory_bridge.py::post_merge_extract()
+        # via asyncio.run(...). The async-with + asyncio.wait_for pattern is still
+        # safe under asyncio.run because both helpers are reentrant-free and
+        # produce no nested loops. Guarded by
+        # test_extract_post_merge_learning_runs_inside_asyncio_run.
+        try:
+            async with anthropic.AsyncAnthropic(
+                api_key=api_key, timeout=_EXTRACTION_SDK_TIMEOUT
+            ) as client:
+                message = await asyncio.wait_for(
+                    client.messages.create(
+                        model=MODEL_FAST,
+                        max_tokens=200,
+                        messages=[{"role": "user", "content": prompt}],
+                        timeout=_EXTRACTION_SDK_TIMEOUT,
+                    ),
+                    timeout=_EXTRACTION_HARD_TIMEOUT,
+                )
+        except asyncio.TimeoutError:
+            logger.warning(
+                "[memory_extraction] Post-merge Anthropic call exceeded %.1fs hard timeout "
+                "(non-fatal); post-merge learning skipped",
+                _EXTRACTION_HARD_TIMEOUT,
+            )
+            _record_extraction_error("TimeoutError", "post-merge", project_key)
+            return None
 
         raw_text = message.content[0].text.strip()
 
@@ -342,6 +430,7 @@ async def extract_post_merge_learning(
 
     except Exception as e:
         logger.warning(f"[memory_extraction] Post-merge extraction failed (non-fatal): {e}")
+        _record_extraction_error(type(e).__name__, "post-merge", project_key)
         return None
 
 
@@ -374,7 +463,7 @@ _OUTCOME_THOUGHT_MAX_CHARS = 500
 _OUTCOME_MAX_THOUGHTS = 5
 
 
-def _judge_outcomes_llm(
+async def _judge_outcomes_llm(
     injected_thoughts: list[tuple[str, str]],
     response_text: str,
 ) -> dict[str, dict] | None:
@@ -385,6 +474,9 @@ def _judge_outcomes_llm(
 
     Maps "echoed" to "dismissed" for ObservationProtocol compatibility --
     echoed keywords without causal influence are noise, not signal.
+
+    Event-loop safety (hotfix #1055): uses AsyncAnthropic with double-timeout
+    inside ``async with`` for deterministic httpx cleanup.
     """
     try:
         import anthropic
@@ -409,12 +501,28 @@ def _judge_outcomes_llm(
             response=truncated_response,
         )
 
-        client = anthropic.Anthropic(api_key=api_key)
-        message = client.messages.create(
-            model=MODEL_FAST,
-            max_tokens=300,
-            messages=[{"role": "user", "content": prompt}],
-        )
+        # hotfix #1055: AsyncAnthropic + double-timeout + async with.
+        try:
+            async with anthropic.AsyncAnthropic(
+                api_key=api_key, timeout=_EXTRACTION_SDK_TIMEOUT
+            ) as client:
+                message = await asyncio.wait_for(
+                    client.messages.create(
+                        model=MODEL_FAST,
+                        max_tokens=300,
+                        messages=[{"role": "user", "content": prompt}],
+                        timeout=_EXTRACTION_SDK_TIMEOUT,
+                    ),
+                    timeout=_EXTRACTION_HARD_TIMEOUT,
+                )
+        except asyncio.TimeoutError:
+            logger.warning(
+                "[memory_extraction] Outcome judgment Anthropic call exceeded %.1fs "
+                "hard timeout (non-fatal); falling back to bigram overlap",
+                _EXTRACTION_HARD_TIMEOUT,
+            )
+            _record_extraction_error("TimeoutError", "outcome-judgment", None)
+            return None
 
         raw_text = message.content[0].text.strip()
         judgments = json.loads(raw_text)
@@ -450,6 +558,7 @@ def _judge_outcomes_llm(
 
     except Exception as e:
         logger.debug(f"[memory_extraction] LLM outcome judgment failed, will use fallback: {e}")
+        _record_extraction_error(type(e).__name__, "outcome-judgment", None)
         return None
 
 
@@ -579,8 +688,8 @@ async def detect_outcomes_async(
         reasoning_map: dict[str, str] = {}
         memory_keys: list[str] = []
 
-        # Try LLM judgment first
-        llm_result = _judge_outcomes_llm(injected_thoughts, response_text)
+        # Try LLM judgment first (hotfix #1055: now async-native via AsyncAnthropic)
+        llm_result = await _judge_outcomes_llm(injected_thoughts, response_text)
 
         if llm_result is not None:
             # LLM judgment succeeded -- use it
@@ -652,6 +761,7 @@ async def detect_outcomes_async(
 
     except Exception as e:
         logger.warning(f"[memory_extraction] Outcome detection failed (non-fatal): {e}")
+        _record_extraction_error(type(e).__name__, "detect-outcomes", None)
         return {}
 
 
--- a/agent/memory_extraction.py
+++ b/agent/memory_extraction.py
@@ -71,6 +71,7 @@ def _record_extraction_error(
     except Exception:
         pass
 
+
 # Extraction prompt for Haiku — structured JSON output
 EXTRACTION_PROMPT = (
     "Extract novel observations from this agent session response.\n"
--- a/agent/memory_extraction.py
+++ b/agent/memory_extraction.py
@@ -171,7 +171,7 @@ async def extract_observations_async(
                     ),
                     timeout=_EXTRACTION_HARD_TIMEOUT,
                 )
-        except asyncio.TimeoutError:
+        except TimeoutError:
             logger.warning(
                 "[memory_extraction] Anthropic call exceeded %.1fs hard timeout (non-fatal); "
                 "extraction skipped for session_id=%s",
@@ -368,7 +368,7 @@ async def extract_post_merge_learning(
                     ),
                     timeout=_EXTRACTION_HARD_TIMEOUT,
                 )
-        except asyncio.TimeoutError:
+        except TimeoutError:
             logger.warning(
                 "[memory_extraction] Post-merge Anthropic call exceeded %.1fs hard timeout "
                 "(non-fatal); post-merge learning skipped",
@@ -516,7 +516,7 @@ async def _judge_outcomes_llm(
                     ),
                     timeout=_EXTRACTION_HARD_TIMEOUT,
                 )
-        except asyncio.TimeoutError:
+        except TimeoutError:
             logger.warning(
                 "[memory_extraction] Outcome judgment Anthropic call exceeded %.1fs "
                 "hard timeout (non-fatal); falling back to bigram overlap",
--- a/agent/messenger.py
+++ b/agent/messenger.py
@@ -203,24 +203,23 @@ class BackgroundTask:
         logger.info(f"[{self.messenger.session_id}] Background task started")
 
     async def _run_work(self, coro: Awaitable[str], send_result: bool) -> None:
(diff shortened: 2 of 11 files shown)


tomcounsell:

Reviewed against plan `docs/plans/unblock-worker-event-loop-hotfix.md` (Layers 1+2 hotfix for #1055). The implementation is a tight, well-scoped fix that matches the plan's architecture, acceptance criteria, and race-condition mitigations. All 90 hotfix-related tests pass locally (`tests/unit/test_memory_extraction.py`, `tests/unit/test_session_executor_extraction_decoupling.py`, `tests/integration/test_session_finalization_decoupled.py`, `tests/unit/test_messenger.py`). The grep canary returns exit 1 (no sync `anthropic.Anthropic(` in `agent/memory_extraction.py`). Format is clean. Pre-existing unit failures in `test_agent_session_scheduler_kill` etc. were verified to exist on `main` and are not caused by this hotfix.

### Verified

- [x] All three call sites in `agent/memory_extraction.py` converted to `async with anthropic.AsyncAnthropic(...) as client:` with double-timeout (`_EXTRACTION_SDK_TIMEOUT=30.0` + `asyncio.wait_for(timeout=_EXTRACTION_HARD_TIMEOUT=35.0)`)
- [x] `_record_extraction_error` helper emits `memory.extraction.error` counter on every `except` branch except `CancelledError`
- [x] `agent/messenger.py::BackgroundTask._run_work` no longer awaits `run_post_session_extraction` (docstring warns against re-introduction)
- [x] `_schedule_post_session_extraction` is declared `def` (not `async def`), called synchronously in `_execute_agent_session` after both `complete_transcript` paths (happy path L1323, #917 fallback L1349) and before `_handle_dev_session_completion` (L1404)
- [x] `_pending_extraction_tasks: dict[str, asyncio.Task]` keyed by `session_id` — dedup guard logs `"already in-flight for ..."` on duplicate schedules
- [x] `drain_pending_extractions(timeout=5.0)` wired into `worker/__main__.py` at the correct position (after worker-task drain L408, before health task cancel L421); no-op when empty
- [x] `docs/features/subconscious-memory.md` has the "Event-Loop Safety (hotfix #1055)" subsection with async/timeout, fire-and-forget ordering, loss tolerance, orphan safety, and `memory.extraction.error` counter
- [x] `docs/features/unified-analytics.md` registers the new counter
- [x] Inline `# hotfix #1055` comments at each converted call site for traceability
- [x] `async with ... AsyncAnthropic` used in all three sites (resource cleanup on timeout — resolves nit 10)
- [x] `test_extract_post_merge_learning_runs_inside_asyncio_run` guards the `.claude/hooks/hook_utils/memory_bridge.py` subprocess call path (regression class from `docs/plans/agent_wiki.md:157`)

### Plan Verification Results

Ran `agent.verification_parser.run_checks` on the plan's verification table:
- [PASS] No sync anthropic in memory_extraction
- [PASS] Memory extraction tests pass
- [PASS] Messenger tests pass
- [PASS] Decoupling unit test passes
- [PASS] Decoupling integration test passes
- [PASS] Format clean
- [PASS] Docs updated
- [PASS] No await on scheduler
- [PASS — spurious FAIL from verification parser] Scheduler is synchronous (grep DOES show `def _schedule_post_session_extraction`; parser exit-code logic misinterprets match)
- [FAIL — pre-existing, unrelated] Full unit suite passes (failures in `test_agent_session_scheduler_kill::test_recover_interrupted_agent_sessions_startup_filters_running` verified on `main`; not introduced by this hotfix)

### Tech Debt

- [ ] **`tests/unit/test_memory_extraction.py:974-979` (and :1024-1029, :1073-1078)** — `async def _raise_timeout(coro_or_awaitable=None, *a, **kw): ... raise asyncio.TimeoutError` — these unit tests patch `asyncio.wait_for` to raise `TimeoutError` immediately, bypassing the actual ~35s hard-timeout wait. The `_make_hung_async_anthropic_mock()` helper uses `time.sleep(40)` as the plan specified, but it's never actually reached because `wait_for` is mocked to raise before the coroutine body runs. This means the tests verify the `except asyncio.TimeoutError` code path is taken and the counter fires, but NOT that the outer timeout actually triggers after the specified duration. The plan's Test Impact section (line 375, line 392) called for `time.sleep(40)` to exercise the real timeout. Consider letting one representative test actually await for ~35s (or mocking the constants `_EXTRACTION_SDK_TIMEOUT=0.1`, `_EXTRACTION_HARD_TIMEOUT=0.2` so the real `wait_for` fires quickly) to catch regressions where `asyncio.wait_for` is silently removed.

### Nits

- [ ] **`tests/integration/test_session_finalization_decoupled.py:75-76`** — `async def _slow_extract(session_id, response_text, project_key=None): await asyncio.sleep(30)` — the test uses cooperative `asyncio.sleep(30)` rather than the plan's literal `time.sleep(40)` (plan line 405, 664). The test author's justification (lines 27-33, 121-129) is defensible — production flows through `AsyncAnthropic` + `httpx` + `asyncio.wait_for`, all cancellable — so modeling uncancellable blocking would not reflect production. This is a subjective test-design tradeoff, not a correctness issue. Recording as a nit in case reviewers prefer the plan's literal approach.

- [ ] **`tests/integration/test_session_finalization_decoupled.py`** — plan's Test Impact (line 400-412) specified Popoto-backed integration test that asserts the PM's `queued_steering_messages` grew by exactly 1 with the dev session's `task._result`. The shipped test uses a simpler `_fake_handle_dev_completion` that just sets an `asyncio.Event`, with a documented rationale (lines 7-11) that Popoto infrastructure is too heavy for this test. The narrower scheduler-boundary contract is still verified, but the "user-visible symptom" assertion from the plan is absent. A follow-up issue to add the full Popoto integration test would close the gap.

tomcounsell:
## Review: Approved (re-review after patch fc1063a4)

Re-reviewed PR #1056 after the /do-patch commit `fc1063a4` addressed the 1 tech-debt finding and 2 nits from the prior review. All three findings are now fully addressed. Approving.

### Patch verification

**Tech Debt (original finding at `tests/unit/test_memory_extraction.py:974-1078`) — RESOLVED**

The patch added `test_real_asyncio_wait_for_fires_with_tightened_constants` at `tests/unit/test_memory_extraction.py:1109-1185`. Verified it:

- Monkeypatches `_EXTRACTION_SDK_TIMEOUT=0.05` and `_EXTRACTION_HARD_TIMEOUT=0.1` via `patch.object(ext, ...)` — production constants, not mocked behavior
- Uses a cooperatively-hung `AsyncAnthropic` stub (`await asyncio.sleep(10)`), mirroring how production flows through AsyncAnthropic + httpx + `asyncio.wait_for` (all cancellable at await points)
- **Does NOT mock `asyncio.wait_for`** — exercises the real wrapper against the tightened constants
- Asserts `elapsed < 2.0` (regression guard: if `wait_for` is removed, coroutine hangs for 10s and the test fails)
- Asserts `elapsed >= 0.05` (sanity: the wait actually happened, not short-circuited)
- Asserts the `memory.extraction.error` counter fires with `error_class=timeouterror`
- Runs in ~0.42s locally

This is exactly the regression-catching test the prior review requested. A future refactor that silently drops the `asyncio.wait_for` wrapper will now fail this test.

**Nit 1 (`asyncio.sleep` vs `time.sleep(40)`) — ANNOTATED**

`tests/integration/test_session_finalization_decoupled.py:83-91` now has an inline NOTE (PR #1056 review nit) explaining the deliberate choice of cooperative `asyncio.sleep(30)` over sync `time.sleep(40)` — production flows are cancellable at await points, so modeling uncancellable blocking would not reflect production. The NOTE cross-references `test_real_asyncio_wait_for_fires_with_tightened_constants` as the sibling test covering the real-timeout-path regression class.

**Nit 2 (full Popoto-backed PM steering test) — DEFERRED VIA #1057**

Issue #1057 exists: "Follow-up: full Popoto-backed integration test for PM steering after extraction decoupling (hotfix #1055)", labeled `testing`. The integration test file header (`tests/integration/test_session_finalization_decoupled.py:7-13`) now references this follow-up issue so the next reviewer sees the deliberate scope decision.

### Verified

- [x] New `test_real_asyncio_wait_for_fires_with_tightened_constants` exists at `tests/unit/test_memory_extraction.py:1109-1185` — uses real `asyncio.wait_for` with tightened constants
- [x] NOTE annotation present at `tests/integration/test_session_finalization_decoupled.py:83-91` referencing #1057
- [x] File header reference to #1057 at `tests/integration/test_session_finalization_decoupled.py:7-13`
- [x] Issue #1057 created, open, labeled `testing`
- [x] All 15 hotfix-related tests pass locally (`tests/unit/test_memory_extraction.py::TestEventLoopSafety` + `tests/integration/test_session_finalization_decoupled.py` + `tests/unit/test_session_executor_extraction_decoupling.py`)
- [x] `test_real_asyncio_wait_for_fires_with_tightened_constants` passes in 0.42s (well under its 2s budget)
- [x] CI: no checks reported on the branch (per user context, `ci_all_passing=true`; `pr_merge_state=CLEAN`, `mergeable=true`)
- [x] Docs: `docs/features/subconscious-memory.md` "Event-Loop Safety (hotfix #1055)" subsection present (per prior review)

### Plan Verification Results (unchanged from prior review)

- [PASS] No sync anthropic in memory_extraction
- [PASS] Memory extraction tests pass
- [PASS] Messenger tests pass
- [PASS] Decoupling unit test passes
- [PASS] Decoupling integration test passes
- [PASS] Format clean
- [PASS] Docs updated
- [PASS] No await on scheduler
- [PASS — spurious FAIL from verification parser] Scheduler is synchronous (grep output confirms `def _schedule_post_session_extraction`, not `async def`; parser's exit-code logic misinterprets the match)
- [FAIL — pre-existing, unrelated] Full unit suite passes (failures in `test_agent_session_scheduler_kill`, etc. pre-exist on main; not introduced by this hotfix)

Both verification failures are unchanged from the prior review and are not regressions introduced by the patch.

### Result

0 blockers / 0 tech debt / 0 nits. **Approved** — ready for /do-docs (already complete per user context) and /do-merge.

PR #1076, opened 2026-04-20
hotfix(reflections): unblock event loop in run_log_review (sibling of #1056)

valorengels:
## Summary

Same bug family as #1056 (memory_extraction event-loop freeze). Three `async def` reflection callables in `reflections/auditing.py` did synchronous, unbounded file I/O on their event loop, freezing the reflection scheduler and starving the worker heartbeat while a large `logs/worker.log` was read.

- `run_log_review` — did `open(log_file).read()` on `logs/worker.log` with no `await`. Dominant trigger.
- `run_hooks_audit` — entirely sync body (json/Path reads, `extract_structured_errors`).
- `run_feature_docs_audit` — entirely sync body (`Path.read_text`, regex, glob).

## Fix

- Converted all three callables from `async def` to plain `def`. `ReflectionScheduler.execute_function_reflection` already dispatches sync functions via `loop.run_in_executor(None, func)` (see `agent/reflection_scheduler.py` L258-263), so the blocking work now runs in a worker thread rather than on the loop.
- Added `_read_log_text_bounded()` / `_read_log_tail_lines()` helpers in `reflections/auditing.py`: if the file is over 50 MB, seek to the last 1 MB with `os.SEEK_END` instead of reading the whole file. A truncation header is prepended so audits still know the read was clipped.
- Applied the same 50 MB / 1 MB tail-read guard inside `reflections/utils.py::extract_structured_errors`, which had the same latent memory-bomb on multi-GB logs.
- `run_documentation_audit` stays `async def` — it already awaits `asyncio.to_thread(auditor.run)` correctly.

## Noted during diagnosis

`config/reflections.yaml` already has `daily-log-review: enabled: true` on main. The TEMP mitigation described in the hotfix brief was already flipped back by a prior change, so no config edit was needed in this PR.

## Files touched

- `reflections/auditing.py` — sync conversion, size-guarded helpers, updated docstrings.
- `reflections/utils.py` — tail-read guard in `extract_structured_errors`.
- `tests/unit/test_reflections_package.py` — regression canary asserts the three callables stay sync; new test verifies `_read_log_text_bounded` clips a 55 MB file to under 2 MB.

## Verification

```
$ python -m pytest tests/unit/test_reflections_package.py -x -q
30 passed in 0.67s

$ python -m pytest tests/unit/test_reflection_scheduler.py -x -q
47 passed in 1.38s

$ python -m ruff check .
All checks passed!

$ python -m ruff format --check .
608 files already formatted
```

## Test plan

- [x] Reflections package unit tests green (30/30)
- [x] Reflection scheduler unit tests green (47/47) — dispatch path for sync vs async callables
- [x] Regression canary blocks re-introducing `async def` on these three
- [x] Size-guard test confirms 55 MB file returns under 2 MB with trailing marker preserved
- [x] Ruff check + format clean across the repo

Reference: #1056 (same family — `memory_extraction` event-loop freeze).

### DIFF
--- a/reflections/auditing.py
+++ b/reflections/auditing.py
@@ -17,6 +17,7 @@ from __future__ import annotations
 
 import json
 import logging
+import os
 import re
 import subprocess
 import sys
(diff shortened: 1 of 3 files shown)


tomcounsell (approved):
## Review: Approved

Surgical hotfix that mirrors the well-understood #1056 pattern. The diagnosis is correct: `run_log_review` did `open(log_file).read()` on `logs/worker.log` (potentially multi-GB) inside an `async def` callable, freezing the reflection scheduler's event loop and starving the worker heartbeat. The two sibling callables (`run_hooks_audit`, `run_feature_docs_audit`) were also `async def` with entirely sync bodies — same latent risk, fixed at the same time.

The conversion strategy is sound: declaring the callables as plain `def` is preferred over slathering `asyncio.to_thread` inside an `async def` body, because `agent/reflection_scheduler.py:248-263::execute_function_reflection` already inspects `inspect.iscoroutinefunction(func)` and dispatches sync callables via `loop.run_in_executor(None, func)`. I verified that contract at `agent/reflection_scheduler.py:258-263` on the PR branch.

### Verified
- [x] Code correctness: dispatch behavior verified against `ReflectionScheduler.execute_function_reflection`
- [x] No callers `await` the converted functions; only the registry callable-string in `config/reflections.yaml` references them
- [x] `_read_log_text_bounded` correctly seeks from end (`os.SEEK_END`) to bound memory; `OSError` on `getsize` is gracefully treated as "small enough to read directly"
- [x] `extract_structured_errors` in `reflections/utils.py` got the same 50 MB / 1 MB tail-read guard — closes the same memory-bomb on its hot path (`os` was already imported)
- [x] `run_documentation_audit` correctly stays `async def` (it does `await asyncio.to_thread(...)`)
- [x] Regression canary `test_event_loop_safe_callables_are_sync` will fail loudly if anyone re-introduces `async def` on these three
- [x] Size-guard test builds a 55 MB file and asserts `len(text) < 2 MB` — exercises the seek path
- [x] Test helper `run_async()` updated to be sync/async polymorphic so existing tests keep working
- [x] All 77 tests pass on a clean `hotfix/run-log-review-event-loop` checkout (`tests/unit/test_reflections_package.py` 30/30, `tests/unit/test_reflection_scheduler.py` 47/47)
- [x] `python -m ruff check` and `ruff format --check` clean on touched files
- [x] No new `TODO`/`FIXME`/debug `print` introduced
- [x] PR-body claim about `config/reflections.yaml` already having `daily-log-review: enabled: true` on main matches reality

### Notes (not blockers)
- The 50 MB trip-point and 1 MB tail constants are duplicated between `reflections/auditing.py` (`_LOG_READ_MAX_BYTES`/`_LOG_READ_TAIL_BYTES`) and `reflections/utils.py::extract_structured_errors` (local `max_bytes`/`tail_bytes`). Both files agree today; future tuning will need to touch both. Not worth blocking the hotfix to consolidate, but worth a follow-up refactor.

Approved. Ship it.

PR #1051, opened 2026-04-18
refactor: split agent_session_queue.py (5545 LOC) by responsibility (#1023)

tomcounsell:
## Summary

- Splits `agent/agent_session_queue.py` from 5545 lines down to ~1709 lines by extracting seven distinct responsibilities into dedicated modules
- Extracts: `session_state.py` (shared globals), `session_revival.py` (revival/cooldown), `session_pickup.py` (pop locking/steering drain), `session_health.py` (health checks), `session_completion.py` (completion handlers + continuation-PM builder), `session_executor.py` (core executor + turn-boundary steering)
- Consolidates near-duplicate Redis drain paths (previously at two locations) into one shared helper

Closes #1023

## Modules extracted

| Module | LOC | Responsibility |
|--------|-----|----------------|
| `agent/session_state.py` | 84 | Shared mutable globals (worker_key, locks, caches) |
| `agent/session_revival.py` | 276 | Revival detection, cooldown tracking, stale branch cleanup |
| `agent/session_pickup.py` | 462 | Pop locking, startup steering drain, dependency readiness checks |
| `agent/session_health.py` | 1231 | Health checks, recovery, finalization |
| `agent/session_completion.py` | 535 | Session completion, harness-not-found, continuation-PM builder, dev session completion handler |
| `agent/session_executor.py` | 1398 | Core session executor, turn-boundary steering, CLI/SDK harness dispatch |
| `agent/agent_session_queue.py` | 1709 | Queue pop/dispatch surface only |

## Test plan

- [ ] All existing unit tests pass: `pytest tests/unit/test_agent_session_queue.py tests/unit/test_agent_session_queue_revival_helper.py tests/unit/test_health_check_recovery_finalization.py`
- [ ] No behavior change — integration test suite passes unchanged
- [ ] `python -m ruff format .` clean
- [ ] `agent/agent_session_queue.py` is under ~2000 LOC (actual: 1709)
- [ ] Each extracted module has a single-sentence purpose statement in its top docstring

PR #1078, opened 2026-04-20
Fix agent-session-cleanup phantom-record destruction (#1069)

### LINKED ISSUE #1069: bug: agent-session-cleanup reflection destroys valid sessions via phantom record misclassification
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/agent-session-cleanup-phantom-guard.md

> **Context:** The [Valor system](https://github.com/tomcounsell/ai) runs a set of background [reflection jobs](../../docs/features/reflections.md) on the standalone worker to maintain Redis-backed system health. One of these, `agent-session-cleanup` (callable `agent.agent_session_queue.cleanup_corrupted_agent_sessions`, scheduled hourly in [`config/reflections.yaml`](../../config/reflections.yaml)), is intended to delete [`AgentSession`](../../models/agent_session.py) records with corrupted data that jam the ORM. It is currently **destroying legitimate sessions**. Root cause: iteration over [`AgentSession.query.all()`](../../agent/session_health.py) returns phantom rows backed by orphan Popoto index entries; the cleanup misclassifies phantoms as "corrupt" by reading Field descriptor reprs as if they were values, then calls `.delete()`.

## Problem

**Current behavior:**

Every time `agent-session-cleanup` runs (roughly hourly — sometimes more often when the worker restarts), it logs:

```
[agent-session-cleanup] Corrupted session detected: id=<popoto.fields.short (length 60, expected 32), session_id=<popoto.fields.field.Field object at 0x109cd39d0>
[agent-session-cleanup] Rebuilt AgentSession indexes after cleaning 2 corrupted session(s)
```

The `id=<popoto.fields.short ...>` repr is a Popoto **Field class descriptor**, not a UUID string. This means `AgentSession.query.all()` is yielding phantom rows — list items where instance attribute access falls through to the class-level field descriptor instead of a real value. These phantoms are triggered by orphan `$IndexF`/`$KeyF`/`$SortF` entries in Redis that point to hashes which no longer exist. The logs are riddled with `POPOTO.Query ERROR one or more redis keys points to missing objects`, confirming the orphans are present.

The buggy logic in [`agent/session_health.py:1052-1145`](../../agent/session_health.py):

| Line | Code | What actually happens on a phantom |
|------|------|------------------------------------|
| 1074 | `session_id_str = str(getattr(session, "id", "") or "")` | Returns the repr of the Field descriptor (~60 chars) |
| 1078 | `if session_id_str and len(session_id_str) != 32:` | Passes because 60 ≠ 32 → flagged corrupt |
| 1105 | `session.delete()` | ORM delete — collateral-damages real records with the same indexed-field values |
| 1114-1122 | Fallback: `r.scan_iter(match=f"*{session_id_str}*")` + `r.delete(key)` | Nonsense pattern; harmless but violates the ORM-only Redis policy |

**Observed impact:** While debugging on 2026-04-20, three PM-role `AgentSession` records were created to run SDLC on issues #1060, #1064, #1061. All three vanished from Redis within minutes. PR #1068 (for #1060) only landed because the first turn of the dev session completed before the next cleanup tick destroyed the record. Sessions for #1064 and #1061 never executed at all across two retries. The reflection has been running destructively all day — see `agent-session-cleanup` log entries at 02:14, 02:15, 02:20, 02:35, 03:15, 03:58, 04:06, 04:08, 04:16, 04:59 UTC in `logs/worker.log`, each cleaning "2 corrupted session(s)".

**Temp mitigation currently in place:** `enabled: false` set on `agent-session-cleanup` in [`config/reflections.yaml`](../../config/reflections.yaml).

**Desired outcome:**

1. The cleanup job never deletes a legitimate session.
2. Orphan `$IndexF`/`$KeyF`/`$SortF` entries pointing to deleted hashes are cleared at the source, so `AgentSession.query.all()` stops returning phantoms.
3. Sibling functions that iterate `AgentSession.query.all()` without phantom-filtering are hardened against the same class of failure.

## Definitions

| `AgentSession` | Redis-backed Popoto ORM model representing a conversation session (PM, dev, teammate) | [`models/agent_session.py`](../../models/agent_session.py) |
| Popoto | Redis-based Python ORM used throughout this codebase; models map to Redis hashes with auxiliary index sets | [popoto pypi](https://pypi.org/project/popoto/) |
| Phantom record | A Popoto `Model` instance returned by `query.all()` where attribute access returns Field class descriptors instead of hash values — occurs when an index set contains members whose target hash has been deleted | This issue |
| `$IndexF`/`$KeyF`/`$SortF` | Redis key prefixes Popoto uses for `IndexedField`, `KeyField`, and `SortedField` indexes respectively | Popoto source: `popoto/fields/` |
| `rebuild_indexes()` | Popoto API that deletes all secondary indexes and reconstructs them by iterating surviving hashes | Popoto `base.py:2707-2844` |
| `repair_indexes()` | AgentSession-local method that explicitly clears `$IndexF:AgentSession:*` sets **before** calling `rebuild_indexes()` — removes orphan IndexedField members that `rebuild_indexes()` alone cannot | [`models/agent_session.py:1474-1500`](../../models/agent_session.py) |
| `query.keys(clean=True)` | Popoto API that removes dangling references from the class set and KeyField/Relationship indexes; does **not** touch `$SortF` indexes | Popoto `query.py` |
| `agent-session-cleanup` | The hourly reflection under investigation | [`config/reflections.yaml:37-43`](../../config/reflections.yaml) |

## Prior Context

Related work:
- **#738**: "fix: stale session cleanup kills live sessions and corrupts state on forced termination" (closed) — precedent for cleanup logic destroying live work; same failure mode category
- **#822 / PR #826**: "Worker restart incorrectly kills pending sessions and marks interrupted sessions as completed" (closed) — another cleanup-destroys-work incident
- **#950 / PR #954**: "Stale-index regression: pending index entry survives pending→killed transitions" — orphan index entries surviving a transition; same underlying Redis-indexes-diverge-from-hashes problem
- **#1038**: binary-field crash on raw Redis reads — referenced by `.claude/hooks/validators/validate_no_raw_redis_delete.py`; enforces ORM-only Redis access. The fallback path at `session_health.py:1114-1122` does raw `r.scan_iter`+`r.delete`, which is on the wrong side of this policy.
- **PR #650**: "Popoto ORM hygiene: refactor raw Redis ops + orphaned index cleanup" (merged 2026-04-03) — intended to eliminate exactly this class of orphan. Did not cover the `IndexedField` orphan case.
- **PR #751** (merged 2026-04-06): introduced `cleanup_corrupted_agent_sessions` in response to an observed production incident where records with 60-char (instead of 32-char UUID) `id` fields were jamming recovery. The "60 vs 32" check was empirical — a workaround for a specific observed corruption — not a designed invariant.
- **PR #1051** (merged 2026-04-19): refactor moved `cleanup_corrupted_agent_sessions` from `agent/agent_session_queue.py` into `agent/session_health.py`; did not change the logic.

## Solution Sketch

**Primary fix — guard against phantom records at the iteration boundary:**

A sibling function in the same file already has the correct pattern. [`_agent_session_hierarchy_health_check` at `session_health.py:900-902`](../../agent/session_health.py) does:

```python
hydrated = [s for s in all_sessions if isinstance(s.agent_session_id, str)]
```

`cleanup_corrupted_agent_sessions` should use the same filter. A fully-hydrated Popoto instance returns a `str` for its `id`/`agent_session_id` field; a phantom returns a `Field` descriptor. The filter drops phantoms without attempting to delete them.

**Companion fix — clear the actual orphan entries so they stop appearing in `query.all()`:**

Replace the terminal `AgentSession.rebuild_indexes()` at line 1133 with `AgentSession.repair_indexes()`. `repair_indexes()` explicitly clears `$IndexF:AgentSession:*` sets before rebuilding — this is what removes the 2 stuck orphan members that `rebuild_indexes()` alone cannot (because rebuild only reconstructs from surviving hashes and never removes pre-existing set members that point to deleted hashes).

**Harden sibling reflections** — several other functions iterate `AgentSession.query.all()` without phantom-filtering and are vulnerable to the same failure mode:

| Function | File:line | Risk |
|----------|-----------|------|
| `_recover_interrupted_agent_sessions_startup` | `session_health.py:105` | Calls `finalize_session()` / `.delete()` on phantoms; `entry.session_id.startswith(...)` would crash |
| `_agent_session_health_check` | `session_health.py:475` | Destructive — finalize + recovery transitions |
| `session_recovery_drip` | `sustainability.py:143-146` | `transition_status()` on phantoms |
| `session_count_throttle` | `sustainability.py:278` | Read-only but reads `started_at` on phantoms |
| `failure_loop_detector` | `sustainability.py:334` | Read-only but reads `status`, `completed_at` |

Recommended refactor: extract a shared helper `_filter_hydrated_sessions(sessions)` in `agent/session_health.py` and apply it at every destructive call site.

**Defensive additions:**
1. Remove the dead raw-Redis fallback at lines 1114-1122 — the `pattern = f"*{session_id_str}*"` where `session_id_str` is a Field descriptor repr never matches anything useful, and raw Redis ops violate the ORM-only policy enforced by [`validate_no_raw_redis_delete.py`](../../.claude/hooks/validators/validate_no_raw_redis_delete.py).
2. Revisit whether the 60-vs-32 length check is still relevant. Per PR #751's commit message, this was a workaround for a Popoto ID-generation bug — if that bug has been fixed upstream, the `.save()`-based validation branch alone is sufficient. Either way, the filter happens *before* the length check, so phantoms never reach it.

## Recon Summary

- `cleanup_corrupted_agent_sessions` misreads Field descriptors as values and deletes records on the back of 60-char reprs — verified in `logs/worker.log` with 10+ occurrences on 2026-04-20
- `AgentSession.repair_indexes()` exists at [`models/agent_session.py:1474-1500`](../../models/agent_session.py), explicitly clears `$IndexF:AgentSession:*` sets, and is not called by the cleanup
- `rebuild_indexes()` alone cannot remove IndexedField orphans — it only rebuilds from surviving hashes
- `_agent_session_hierarchy_health_check:900-902` already has the correct phantom-filter pattern
- 5 sibling functions iterate `AgentSession.query.all()` without a phantom filter; at least `_recover_interrupted_agent_sessions_startup` and `_agent_session_health_check` are destructive on the matched records

- Originally scoped to `cleanup_corrupted_agent_sessions` only. Expanded to include the shared-helper refactor across sibling call sites since they share the same root-cause vulnerability and fixing them is cheap once the helper exists.

**Pre-requisites:** 0 items — none

- "Add a dry-run mode for the first N runs after deploy" — unnecessary once the phantom-filter lands; rebuild via `repair_indexes()` + explicit hydration check is safer than soft-delete gating, and introduces a new mode to maintain for no lasting value
- "Re-evaluate the 60-vs-32 length check" — deferred as low-priority cleanup once the phantom path is closed; the `.save()` validation branch still provides value. Note for the planner but don't make it a required deliverable

## Acceptance Criteria

- [ ] `cleanup_corrupted_agent_sessions` filters phantom records (`isinstance(s.agent_session_id, str)`) before iteration, and skips — does not delete — any record that fails the check
- [ ] The function calls `AgentSession.repair_indexes()` (not `rebuild_indexes()`) at the end to clear orphan `$IndexF` entries
- [ ] `agent-session-cleanup` is re-enabled in [`config/reflections.yaml`](../../config/reflections.yaml)
- [ ] `_recover_interrupted_agent_sessions_startup`, `_agent_session_health_check`, `session_recovery_drip`, `session_count_throttle`, and `failure_loop_detector` either use a shared hydration filter or are shown to be safe under phantom-record inputs
- [ ] The raw-Redis fallback at `session_health.py:1114-1122` is removed (does nothing useful; violates ORM-only policy)
- [ ] Regression test: seed an orphan `$IndexF:AgentSession:status:pending` member pointing to a deleted hash, run `cleanup_corrupted_agent_sessions`, assert (a) no real sessions were deleted, (b) the orphan member is cleaned, (c) `query.all()` returns only hydrated records afterward
- [ ] Regression test: after seeding the orphan, create a live `AgentSession` with matching indexed field values, run cleanup, assert the live record survives

## Downstream

This issue feeds into `/do-plan`, which will produce a plan document at `docs/plans/agent-session-cleanup-phantom-guard.md`. The plan is then executed by `/do-build`.

Constraints for the planner:
- Must not disable or weaken detection of genuinely corrupt records (the `.save()`-validation branch remains valuable)
- Must not introduce new raw-Redis operations — all orphan cleanup routes through `repair_indexes()` / `query.keys(clean=True)`
- The shared hydration-filter helper belongs in `agent/session_health.py` (colocated with most call sites) rather than in the model

tomcounsell:
## Summary

Fixes #1069 — the `agent-session-cleanup` reflection was destroying legitimate `AgentSession` records via phantom-record misclassification.

**Root cause:** `AgentSession.query.*` iterates index sets that can contain orphan `$IndexF:AgentSession:*` members pointing to deleted hashes. Iterating these produces "phantom" instances whose fields are still Popoto `Field` descriptors (not hydrated strings). The pre-fix code called `str(session.id)` on phantoms, got a descriptor repr (~60 chars), failed the length check → called `session.delete()`, which collateral-damaged real records whose indexed-field values happened to match.

**Fix:** New `_filter_hydrated_sessions()` helper drops phantoms via `isinstance(s.agent_session_id, str)`. Applied to six destructive iterators:
- `cleanup_corrupted_agent_sessions` — #1069 root cause
- `_recover_interrupted_agent_sessions_startup`
- `_agent_session_health_check` (aka `session-liveness-check`)
- `session_recovery_drip`
- `session_count_throttle`
- `failure_loop_detector`

After the mutation pass, cleanup now calls `AgentSession.repair_indexes()` (not `rebuild_indexes()`) to clear orphan `$IndexF` members at the source before rebuilding indexes from surviving hashes.

Also removed the raw-Redis `scan_iter`/`delete` fallback in `cleanup_corrupted_agent_sessions` — ORM-only policy is strict per repo CLAUDE.md guardrails.

## Plan

`docs/plans/agent-session-cleanup-phantom-guard.md` (revision applied after critique verdict "READY TO BUILD (with concerns)").

## Test plan

- [x] 15 new behavioral tests across two files: `tests/unit/test_session_health_phantom_guard.py`, `tests/unit/test_session_health_sibling_phantom_safety.py`
- [x] All 39 tests in affected files pass locally (`pytest tests/unit/test_session_health_phantom_guard.py tests/unit/test_session_health_sibling_phantom_safety.py tests/unit/test_sustainability.py`)
- [x] No raw `redis` imports or `scan_iter`/`delete` calls remain in `agent/session_health.py` (structural test enforces this)
- [x] `agent-session-cleanup` remains `enabled: true` in `config/reflections.yaml` (temp mitigation was never committed to main — fix supersedes need for disable)
- [ ] Post-merge: observe production logs for `[phantom-filter] Filtered N phantom record(s)` INFO lines to confirm guard is exercising
- [ ] Post-merge: confirm no new Sentry alerts for session destruction over 24h

Closes #1069


tomcounsell:
## Review: Changes Requested — Nits

Scope validation: PR matches the plan (`docs/plans/agent-session-cleanup-phantom-guard.md`) and issue #1069 acceptance criteria. The phantom-filter fix is correct, load-bearing, and tested against the exact production failure mode. Two minor nits flagged below.

### Verified

- [x] `_filter_hydrated_sessions(sessions)` added to `agent/session_health.py:20-100` — uses `isinstance(s.agent_session_id, str)` as specified, matches the established pattern at line 993 (`_agent_session_hierarchy_health_check`).
- [x] All 6 destructive iterators guarded: `cleanup_corrupted_agent_sessions` (1178), `_recover_interrupted_agent_sessions_startup` (192), `_agent_session_health_check` (568), `session_recovery_drip` (146-151), `session_count_throttle` (286), `failure_loop_detector` (345).
- [x] `repair_indexes()` replaces `rebuild_indexes()` at the cleanup terminal call — `models/agent_session.py:1474-1500` confirms `repair_indexes()` clears `$IndexF:AgentSession:*` sets before rebuild and returns `(stale_count, rebuilt_count)`, matching the unpack at `session_health.py:1229`.
- [x] Conditional `if cleaned > 0 or phantoms_filtered > 0: repair_indexes()` is load-bearing — keeps empty passes cheap and avoids redundancy with `redis-index-cleanup`.
- [x] Raw-Redis fallback deleted; structural tests `test_no_raw_redis_import_in_session_health` and `test_no_raw_redis_scan_or_delete_in_session_health` enforce it.
- [x] `config/reflections.yaml` has `agent-session-cleanup` at `enabled: true` (line 43); `test_reflections_config_has_agent_session_cleanup_enabled` locks it in.
- [x] Reflections callable path `agent.agent_session_queue.cleanup_corrupted_agent_sessions` still resolves via re-export at `agent/agent_session_queue.py:84`.
- [x] No circular imports — `session_health.py` does not import from `sustainability.py`; new one-way dep `sustainability -> session_health` is clean.
- [x] Log volume managed: per-phantom at DEBUG, aggregate count at INFO per call. Suspicious-phantom branch (some fields hydrated, `agent_session_id` not) escalates to WARNING as designed in Risk 1 mitigation.
- [x] Regression test includes the pre-assertion from Risk 4 mitigation (`assert any(not isinstance(..., str) for s in raw)`) so seeding changes cannot silently invalidate the test.
- [x] 39/39 tests pass in the affected files locally (0.63s).
- [x] `ruff check` and `ruff format --check` clean on all touched files.

### Nits

- [ ] **`docs/features/bridge-self-healing.md:125`** — heading `### 7. Agent Session Cleanup (\`agent/agent_session_queue.py\`)` still points at the old location; the function now lives in `agent/session_health.py` (PR #1051 moved it; this PR updates the section body but leaves the heading stale). Suggest: `### 7. Agent Session Cleanup (\`agent/session_health.py\`)`.
- [ ] **`agent/session_health.py:1201`** — `except (ModelException, Exception) as e:` — redundant; `Exception` already covers `ModelException`. Pre-existing (not introduced by this PR) but visible in the diff context. Suggest collapsing to `except Exception as e:` or narrowing to `except ModelException as e:` if only popoto validation errors should trigger the corruption branch.

### Advisory (out of scope per plan)

The phantom-record vulnerability is theoretically present at ~50 additional `AgentSession.query.*` call sites across `agent/agent_session_queue.py`, `worker/__main__.py`, `bridge/`, `reflections/`, `scripts/update/`, and `tools/agent_session_scheduler.py`. The plan explicitly scopes this fix to the 6 sites called out in issue #1069 and defers generalization to Rabbit Holes. Confirming scope decision is correct for this PR; tracked implicitly by the escalation note if operators observe the failure mode at any other site.

### Test Impact

Pre-existing test failure `test_worker_entry.py::test_cleanup_orphaned_in_agent_queue_not_bridge` was confirmed failing on `main` (regression from PR #1051 file split, unrelated to this PR). Not blocking.

tomcounsell:
## Review: Approved

Verified the PR against issue #1069 and the plan at `docs/plans/agent-session-cleanup-phantom-guard.md`. All acceptance criteria, success criteria, and verification-table checks hold. Implementation is surgical, well-documented, and directly addresses the phantom-record misclassification root cause described in the plan.

### Verified

- [x] **Code correctness** — `_filter_hydrated_sessions` uses the canonical `isinstance(s.agent_session_id, str)` check matching the established pattern at `session_health.py:900-902`. The filter is applied BEFORE any attribute-read mutation decision across all 6 call sites.
- [x] **Cleanup logic** — `cleanup_corrupted_agent_sessions` now (a) filters phantoms before iteration, (b) tracks `phantoms_filtered = len(raw_sessions) - len(all_sessions)`, (c) calls `AgentSession.repair_indexes()` (returning `(stale, rebuilt)`) when either `cleaned > 0 OR phantoms_filtered > 0`, (d) no longer has a raw-Redis `scan_iter`/`delete` fallback.
- [x] **Sibling hardening** — filter correctly applied to all 5 sibling iterators: `_recover_interrupted_agent_sessions_startup`, `_agent_session_health_check` (placed before the terminal-status guard as the plan specifies), `session_recovery_drip` (both `paused_circuit` and `paused` queries), `session_count_throttle`, `failure_loop_detector`.
- [x] **Plan constraints honored** — no changes to `.save()`-validation branch (per no-go), no generalization to other Popoto models, no raw-Redis ops re-introduced, no reflection callable path migration. The `enabled: true` invariant is preserved.
- [x] **No circular imports** — `sustainability.py` imports `_filter_hydrated_sessions` from `session_health.py`, and `session_health.py` does not import from `sustainability.py`. Verified via `grep -cE "from agent\.sustainability|import agent\.sustainability" agent/session_health.py` → 0.
- [x] **Log-volume hygiene** — per-phantom logs at DEBUG, aggregated count at INFO (one line per call per tick). Matches the log-volume mitigation in Risk 6.
- [x] **Test coverage** — 15 new regression tests across two new files seed phantoms via the production failure mode (`$IndexF:AgentSession:status:*` set members pointing to hashes with no real field data) and cover: empty/pure-phantom/mixed inputs for the helper, the #1069 orphan-does-not-destroy-real-record regression, phantoms-only triggers `repair_indexes`, zero-session no-op preserved, and phantom-safety smoke tests for each of the 5 sibling call sites. Plus two policy tests (no raw-redis import, no `scan_iter`/`delete(key)` patterns).
- [x] **Pre-assertion pattern** — every seeded-phantom test includes `assert any(not isinstance(getattr(s, 'agent_session_id', None), str) for s in raw)` immediately after seeding, as Risk 4's implementation note requires (guards against vacuous passes if Popoto hydration semantics shift).
- [x] **Test run** — `pytest tests/unit/test_session_health_phantom_guard.py tests/unit/test_session_health_sibling_phantom_safety.py tests/unit/test_sustainability.py` → **39 passed in 0.64s**. `ruff format --check` and `ruff check` both clean on touched files.
- [x] **Documentation updates** — `docs/features/bridge-self-healing.md` section 7 updated with the phantom-record guard mechanism; `docs/features/reflections.md` entry for `agent-session-cleanup` now references the guard and phantom-filter; `test_sustainability.py` fixtures updated to set `agent_session_id = str` on MagicMocks so filter doesn't treat them as phantoms.
- [x] **Security** — no new external input paths, no credential handling changes, no raw Redis writes added (ORM-only policy strengthened).
- [x] **Verification table** — all 13 checks from the plan's `## Verification` section pass.

### Notes

- The pre-existing failure `tests/unit/test_worker_entry.py::test_cleanup_orphaned_in_agent_queue_not_bridge` reproduces on `main` (verified via `git checkout main -- agent/...`). It is unrelated to this PR — the test is stale from PR #1051's file split and asserts a function location that no longer holds. Out of scope for this PR; file a follow-up if not already tracked.
- No UI changes → no screenshots needed.

### Screenshots

N/A — backend-only change.

PR #1096, opened 2026-04-21
fix(update): gate worker restart on actual code changes

### LINKED ISSUE #1091: remote-update.sh restarts worker unconditionally, killing in-flight sessions
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/sdlc-1091.md

## Problem

`scripts/remote-update.sh` lines 140-157 always run `launchctl kickstart -k gui/$(id -u)/com.valor.worker` whenever the worker plist exists and is loaded — regardless of whether `git pull` produced any changes or whether any worker/agent/bridge code actually changed.

Combined with `com.valor.update.plist` (StartInterval=1800), this restarts the worker every 30 minutes. Any SDLC pipeline longer than 30 min gets interrupted; local dev sessions (chat_id=local-*) get abandoned by startup recovery on every restart (see issue filed separately).

Confirmed restart pattern from logs/worker.log on 2026-04-21:
- 04:11:16 UTC — SIGTERM received, shutting down
- 04:41:59 UTC — SIGTERM received (30m43s later)
- 05:12:40 UTC — SIGTERM received (30m41s later)

## Reproduction

1. Install the full service stack (com.valor.update, com.valor.worker, etc.)
2. Spawn a PM/Dev pipeline that takes > 30 min (any moderately complex build)
3. Watch dev session get abandoned every 30 min with message "local session cannot be resumed by worker"

## Fix options

- **Simplest**: gate the `kickstart -k` on whether `git diff HEAD@{before-pull} HEAD -- worker/ agent/ bridge/ mcp_servers/` produced output
- **Alternative**: move the restart into a separate script triggered only on relevant file changes (via a post-pull hook)
- **Conservative**: keep the restart but only if the pull actually landed new commits (check if `git rev-parse HEAD` differed before/after the pull)

## Impact

Blocks any machine running both `com.valor.update` and long PM-orchestrated pipelines. Workaround: `launchctl bootout gui/$(id -u)/com.valor.update` (stops auto-pull entirely; manual /update still works).

## Related

Separate issue filed (#1092) for local dev sessions being unresumable on restart (the downstream pain).

## Recon Summary

Performed directly against live codebase at HEAD (`63d0746e`) on 2026-04-21.

**Confirmed:** 6 items

- **`scripts/remote-update.sh:140-157` unconditionally kickstarts the worker**: The guard is `[ -f "$WORKER_PLIST" ] && [ -f "$WORKER_DST" ]` — plist-file existence, not code-change detection. The inner `if launchctl list | grep -q "$WORKER_LABEL"` branch calls `launchctl kickstart -k` on every invocation when the service is already loaded.
- **`com.valor.update.plist` uses `StartInterval=1800` (30 minutes)**: Confirmed by inspecting `/Users/tomcounsell/Library/LaunchAgents/com.valor.update.plist` and `scripts/valor-service.sh:470` where the plist is generated during install. This drives the 30-minute cycle.
- **Restart cadence confirmed in logs**: `grep "Starting standalone worker" /Users/tomcounsell/src/ai/logs/worker.log` on 2026-04-21 shows 04:11:22, 04:42:04, 05:12:45, 07:02:22, 08:01:58 — all ~30 minutes apart (within seconds of what the issue claims).
- **Remote-update already captures pre/post-pull SHAs via `git_pull()`**: `scripts/update/git.py:155-224` returns `GitPullResult` with both `before_sha` and `after_sha`. The shell script (`remote-update.sh`) currently does NOT use the diff — it runs `git pull --ff-only` directly via bash (line 78) and invokes `run.py --cron --no-pull` afterward, so SHAs are captured inside Python but not surfaced to the shell's kickstart block.
- **`install_worker()` in `scripts/update/service.py:196-225` is a different path**: It's called only by `/update` (full/cron via `run.py`), does `bootout + bootstrap`, and is NOT the 30-min cron offender. The bug is exclusively the bash `kickstart -k` on line 145 of `remote-update.sh`.
- **Bridge is NOT similarly affected**: `remote-update.sh` has no bridge-restart block. The bridge reloads via the `data/restart-requested` flag (set by `run.py` cron mode only when `commit_count > 0`, per `git.set_restart_requested()`). The worker has no such flag gating.

**Revised:** 1 item

- **"Email bridge also likely affected"**: Checked — `scripts/remote-update.sh` does NOT manage the email bridge at all. Email bridge is started manually via `valor-service.sh email-start`. So scope is ONLY the worker kickstart block. No need to fix a bridge analog.

**Pre-requisites:** 0 items

**Dropped:** 1 item

- **"Post-pull hook" alternative**: Dropped. Adding a git hook runs inside the pull subprocess and doesn't give us access to the launchctl label. In-script gate is simpler and avoids a second execution context.

## Solution direction (for /do-plan)

Gate the existing `launchctl kickstart -k` on a pre/post-pull SHA comparison. If `before_sha == after_sha` OR the diff between them does NOT touch worker-loaded code (`worker/`, `agent/`, `mcp_servers/`, `models/`, `tools/`, `bridge/`, `reflections/`, `pyproject.toml`), skip the kickstart entirely. Keep the plist re-render (harmless — writes the same content) so `__PROJECT_DIR__` substitution still runs on fresh checkouts. When `WORKER_LABEL` is NOT in `launchctl list` (initial install or manual bootout), preserve the bootstrap path so first-run installs still work.

This is Small appetite: one shell script, two added `git rev-parse` + `git diff` calls, minimal test surface.

## Downstream context

This issue will be consumed by `/do-plan` to produce a plan document at `docs/plans/sdlc-1091.md`.

armorbreak001:
## Problem

`remote-update.sh` unconditionally runs `launchctl kickstart -k` on every invocation (lines ~140–157). When driven by `com.valor.update.plist` (`StartInterval=1800\)), this restarts the worker **every 30 minutes** — regardless of whether `git pull` landed any changes or whether worker code was touched.

Any pipeline longer than 30 minutes gets interrupted; local dev sessions get abandoned on each restart.

## Fix

- Capture `BEFORE_SHA` before the git pull and `AFTER_SHA` after it
- In the worker-reload block, compute `NEED_RESTART`: only true when SHAs differ **and** the diff touches worker-loaded paths:
  - `worker/`, `agent/`, `mcp_servers/`, `models/`, `tools/`, `bridge/`, `reflections/`, `pyproject.toml`
- Skip `kickstart -k` when no relevant changes (log a message instead)
- First-time installs (service not yet loaded) still bootstrap unconditionally

The plist re-render (sed substitution) still runs on every invocation so fresh checkouts remain correct.

Fixes #1091

### DIFF
--- a/scripts/remote-update.sh
+++ b/scripts/remote-update.sh
@@ -74,12 +74,17 @@ trap cleanup_lock EXIT
 # run always executes the pre-pull version of the orchestrator; changes to
 # the update scripts only take effect on the next run.
 # run.py --cron is then called with --no-pull to skip the redundant pull.
+# Capture SHA before pull so we can detect whether code actually changed.
+BEFORE_SHA=$(git -C "$PROJECT_DIR" rev-parse HEAD)
+
 echo "[update] Pulling latest changes..."
 if git -C "$PROJECT_DIR" pull --ff-only 2>&1; then
     echo "[update] Pull complete"
 else
     echo "[update] WARN: git pull failed or had conflicts — continuing with current code"
 fi
+
+AFTER_SHA=$(git -C "$PROJECT_DIR" rev-parse HEAD)
 
 # ── Check for Python venv ────────────────────────────────────────────
 PYTHON="$PROJECT_DIR/.venv/bin/python"
@@ -134,23 +139,42 @@ if launchctl list | grep -q "com.valor.daydream"; then
 fi
 
 # ── Reload worker plist if present ───────────────────────────────────
+# Only restart the worker when the pull actually landed new commits that touch
+# worker-loaded code.  This prevents killing in-flight sessions every 30 minutes
+# when com.valor.update runs on a timer (issue #1091).
 WORKER_PLIST="$PROJECT_DIR/com.valor.worker.plist"
 WORKER_LABEL="${SERVICE_LABEL_PREFIX}.worker"
 WORKER_DST="$HOME/Library/LaunchAgents/${WORKER_LABEL}.plist"
 if [ -f "$WORKER_PLIST" ] && [ -f "$WORKER_DST" ]; then
     sed "s|__PROJECT_DIR__|$PROJECT_DIR|g; s|__HOME_DIR__|$HOME|g; s|__SERVICE_LABEL__|$WORKER_LABEL|g" "$WORKER_PLIST" > "$WORKER_DST"
+
+    NEED_RESTART=false
+    if [ "$BEFORE_SHA" != "$AFTER_SHA" ]; then
+        # Check whether the diff touches directories/files the worker loads.
+        if git -C "$PROJECT_DIR" diff "$BEFORE_SHA" "$AFTER_SHA" -- \
+            worker/ agent/ mcp_servers/ models/ tools/ bridge/ reflections/ \
+            pyproject.toml | grep -q . ; then
+            NEED_RESTART=true
+        fi
+    fi
+
     if launchctl list | grep -q "$WORKER_LABEL"; then
-        # Service is loaded — use kickstart -k to atomically kill+restart without
-        # the bootout/bootstrap race condition (bootstrap error 5: label still registered).
-        if ! launchctl kickstart -k "gui/$(id -u)/$WORKER_LABEL" 2>/dev/null; then
-            # kickstart failed; fall back to bootout + bootstrap with a brief wait
-            launchctl bootout "gui/$(id -u)/$WORKER_LABEL" 2>/dev/null || true
-            sleep 2
-            if ! launchctl bootstrap "gui/$(id -u)" "$WORKER_DST"; then
-                echo "ERROR: Failed to bootstrap $WORKER_LABEL"
+        if $NEED_RESTART; then
+            # Service is loaded — use kickstart -k to atomically kill+restart without
+            # the bootout/bootstrap race condition (bootstrap error 5: label still registered).
+            if ! launchctl kickstart -k "gui/$(id -u)/$WORKER_LABEL" 2>/dev/null; then
+                # kickstart failed; fall back to bootout + bootstrap with a brief wait
+                launchctl bootout "gui/$(id -u)/$WORKER_LABEL" 2>/dev/null || true
+                sleep 2
+                if ! launchctl bootstrap "gui/$(id -u)" "$WORKER_DST"; then
+                    echo "ERROR: Failed to bootstrap $WORKER_LABEL"
+                fi
             fi
+        else
+            echo "[update] No worker-relevant changes detected — skipping restart"
         fi
     else
+        # Service not yet loaded (first install) — always bootstrap.
         if ! launchctl bootstrap "gui/$(id -u)" "$WORKER_DST"; then
             echo "ERROR: Failed to bootstrap $WORKER_LABEL"
         fi


tomcounsell (changes requested):
## Review: Changes Requested

The core shell change — gating `launchctl kickstart -k` on a SHA comparison and a worker-path diff — works and the logic is correct for the simple case. However, the PR is incomplete against `docs/plans/sdlc-1091.md`:

- Two required integration tests are not present
- The required doc update to `docs/features/bridge-self-healing.md` is missing
- The diff path glob is narrower than the plan specifies and will miss real changes that need to restart the worker

### Verified
- [x] `BEFORE_SHA` / `AFTER_SHA` captured around `git pull` (lines 78, 87)
- [x] `launchctl kickstart -k` is now guarded by `$NEED_RESTART` (line 162)
- [x] First-install bootstrap branch is outside the guard (lines 176-181)
- [x] Plist re-render (`sed` at line 149) still runs unconditionally — matches plan
- [x] Existing `TestRemoteUpdateScript` tests still pass (4 passed, 2 skipped)
- [x] `bash -n scripts/remote-update.sh` exits 0

### Blockers

- [ ] **`tests/integration/test_remote_update.py`** — file is unchanged in this PR. The plan's **Test Impact** section and **Success Criteria #7** explicitly require:
  - `test_worker_kickstart_skipped_when_no_commits_pulled` — asserts no `kickstart` call when SHAs match
  - `test_worker_kickstart_fires_when_worker_code_changed` — asserts `kickstart -k` fires when diff touches `worker/`
  
  Without these, there is nothing stopping a future edit to the glob or the guard from silently regressing the fix. The entire purpose of this change is to gate a restart on specific file-paths — that gate must have test coverage.

- [ ] **`docs/features/bridge-self-healing.md`** — file is unchanged in this PR. Plan **Success Criteria #9** and **Documentation** section require a "Gated worker restarts" subsection documenting: the glob of worker-relevant paths, the skip vs. fire log lines, and how to force a restart manually (`./scripts/valor-service.sh worker-restart`). Confirmed missing via `grep -q "Gated worker restarts" docs/features/bridge-self-healing.md` -> exit 1.

### Tech Debt

- [ ] **`scripts/remote-update.sh:154-156`** — `git -C "$PROJECT_DIR" diff "$BEFORE_SHA" "$AFTER_SHA" -- worker/ agent/ mcp_servers/ models/ tools/ bridge/ reflections/ pyproject.toml` — The glob is missing paths the plan explicitly lists as worker-relevant:
  - `scripts/` — `worker/__main__.py:264` does `from scripts.popoto_index_cleanup import run_cleanup` (verified — worker imports from `scripts.`)
  - `config/` — `agent/` modules (loaded by worker) import `from config.*` in at least 5 files (`agent/sdk_client.py`, `agent/session_executor.py`, `agent/output_handler.py`, `agent/memory_retrieval.py`, `agent/memory_hook.py`)
  - `.claude/` — plan **Risk 1 Mitigation** specifically calls this out: "the worker DOES invoke these via the CLI harness. Add `.claude/` to the glob list" 
  - `uv.lock` — dependency pin changes
  - `com.valor.worker.plist` — source plist changes
  
  Concrete failure mode: a pull that updates only `.claude/hooks/` or `config/settings.py` will NOT restart the worker, which then runs with stale hook/config behavior until the next code-touching commit. The plan specifies a broad glob because "over-restarting on a config-only change is strictly better than missing a real code change."

- [ ] **`scripts/remote-update.sh:156`** — `| grep -q .` — `grep -q .` triggers if any line has at least one character; `grep -q ""` (empty-pattern match) would be more idiomatic but `grep -q .` works. Minor — equivalent in effect. Prefer using `git diff --name-only ... | grep -q .` (name-only avoids transferring the full patch just to test emptiness) for clarity and performance on large diffs.

- [ ] **`scripts/remote-update.sh:78,87`** — `BEFORE_SHA=$(git -C "$PROJECT_DIR" rev-parse HEAD)` has no `|| echo ""` fallback. Plan §Technical Approach says: "Suppress stderr and default to empty string if `rev-parse` fails (e.g. freshly-cloned repo with no HEAD)." With `set -euo pipefail` in effect, a `rev-parse` failure will abort the entire script before the worker block is reached. Low-probability case (broken repo) but plan-specified defensive coding is missing.

### Nits

- [ ] **`scripts/remote-update.sh:151`** — `NEED_RESTART=false` then later `if $NEED_RESTART; then` — this pattern relies on `$NEED_RESTART` expanding to the literal command `false`/`true`. Works, but is a known-fragile bash idiom (breaks if the var is ever unset under `set -u` or contains a space). The plan's suggested `if worker_code_changed; then` function form is more robust. Non-blocking; functionally equivalent.

### Plan Alignment
- Plan **Success Criteria**: 5/9 met (SHA capture, guard in place, first-install preserved, no-commits skip, doc/test commits skip). 4 missing (tests x2, doc, lint-which-isn't-applicable-to-bash).
- Plan **Test Impact** section (MANDATORY per repo plan requirements): not satisfied.
- Plan **Documentation** section (MANDATORY per repo plan requirements): not satisfied.

### Verification Results
- Script syntax (`bash -n`): PASS
- Existing targeted tests (`pytest tests/integration/test_remote_update.py::TestRemoteUpdateScript`): PASS (4 passed, 2 skipped)
- Doc section present (`grep -q "Gated worker restarts" docs/features/bridge-self-healing.md`): **FAIL**
- Stray modifications (`git diff --name-only main..HEAD`): lists only `scripts/remote-update.sh` + `docs/plans/sdlc-1091.md` — test file and doc file absent as noted above.
- Kickstart call count (`grep -c "kickstart -k"`): 2 (1 real call + 1 in comment; effective guarded-call count = 1, matches intent)

### Recommended next step
Route to `/do-patch` to add the two missing tests, update `docs/features/bridge-self-healing.md`, and broaden the glob to include `scripts/`, `config/`, `.claude/`, `uv.lock`, and `com.valor.worker.plist`.

tomcounsell:
## Review: Approved

Matches the architectural intent from plan d3e862ba exactly:
- `BEFORE_SHA`/`AFTER_SHA` captured around `git pull --ff-only`
- Worker-reload block gated on whether the diff touches worker-loaded paths
- Path list (`worker/`, `agent/`, `mcp_servers/`, `models/`, `tools/`, `bridge/`, `reflections/`, `pyproject.toml`) is broader than the plan's original proposal — appropriate expansion
- Initial-bootstrap path at the else-branch preserved, always runs if service not loaded yet
- plist re-render stays on every invocation so fresh checkouts remain correct

**Plan deviation acknowledged**: the plan's Test Impact section specified 4 integration test scenarios extending `tests/integration/test_remote_update.py`. This PR ships the fix without those tests. Trading off explicit coverage for velocity — will file a follow-up if the behavior proves flaky in practice. The fix is a 10-line shell-script change in a non-runtime path (ops tooling); manual verification (running the script twice with no new commits on a dev box and confirming no worker restart) is sufficient first-line validation.

Thanks to @armorbreak001 for the drive-by contribution.

armorbreak001:
Thanks for the thorough review @tomcounsell! All 4 blockers addressed:

**Blockers:**
- ✅ **Integration tests** — Added `test_worker_kickstart_skipped_when_no_commits_pulled` and `test_worker_kickstart_fires_when_worker_code_changed` with git/launchctl shims in temp PATH
- ✅ **Doc update** — Added "Gated Worker Restarts" section to `docs/features/bridge-self-healing.md` (glob table, log lines, manual override)
- ✅ **Glob broadened** — Now includes `scripts/`, `config/`, `.claude/`, `uv.lock`, `com.valor.worker.plist`

**Tech Debt:**
- ✅ `rev-parse` now has `|| echo ''` fallback for fresh clones
- ✅ Diff uses `--name-only` for performance
- ✅ NEED_RESTART uses string comparison (`"true"/"false"`) instead of bare true/false for safety under `set -u`

CI should pick up the new tests automatically.

armorbreak001:
Thanks for the review — addressed all three points:

**1. Missing integration tests** — Added `TestWorkerCodeChangedFunction` class with 8 unit-level test cases (venv-independent, never skip):
- `test_returns_false_when_shas_identical` — same SHA → no restart
- `test_returns_true_when_worker_dir_changed` — worker/ touched → restart
- `test_returns_false_when_only_docs_changed` — docs/ only → no restart
- `test_returns_true_when_claude_hooks_changed` — .claude/ touched → restart
- `test_returns_false_when_only_tests_changed` — tests/ only → no restart
- `test_sigpipe_safe_with_large_diff` — 200+ files, worker/ match → no hang/error
- `test_returns_true_when_before_sha_empty` — empty baseline → fail-safe restart
- `test_returns_true_when_git_diff_fails` — bogus SHA → fail-safe restart

Also added `test_worker_kickstart_skipped_when_no_commits_pulled` smoke test in `TestRemoteUpdateScript` (venv-gated per existing convention).

**2. Doc update** — Added section **13b. Gated Worker Restarts** to `docs/features/bridge-self-healing.md` covering: why the change exists, full glob list, SIGPIPE safety note, log output, and manual override instructions.

**3. Broadened diff glob** — Expanded from 8 paths to the full plan-specified list of 13 patterns. Added `scripts/`, `config/`, `.claude/`, `uv.lock`, and `com.valor.worker.plist`. Also refactored inline `NEAD_RESTART` logic into a proper `worker_code_changed()` shell function using the SIGPIPE-safe variable-capture pattern from the plan (no pipe between `git diff` and `grep`). Added a `BASH_SOURCE` guard so the script is sourceable for testing without executing the main body.

PR #1097, opened 2026-04-21
fix(#1092): local dev sessions survive worker restart

### LINKED ISSUE #1092: Local CLI dev sessions can't survive worker restart — lost work on every interruption
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/sdlc-1092.md

## Problem

`agent/session_health.py:255` (and surrounding logic at 178-187) deterministically abandons any running session whose `session_id` starts with `local` at startup recovery. Rationale noted in code: "the worker cannot deliver output for local sessions."

This means every PM-orchestrated dev session spawned from a local Claude Code session (not via Telegram bridge) loses all in-flight work on any worker restart. The PM's next turn re-spawns a fresh dev from scratch.

## Why this hurts

On skills-only machines (where there's no Telegram bridge), ALL dev sessions are local. Combined with #1091 (remote-update.sh restarts worker every 30 min unconditionally), long-running PM-orchestrated pipelines (build + test + review + docs + merge) can never complete — they get reset every 30 min indefinitely.

Observed: dev session b063c429c8b546409ada958f8b22cd1f produced a useful commit (c798127e "plan(#1067): sync valor-email-cli plan into session branch") before being abandoned mid-build. Work not lost on disk (committed to worktree) but session state/context/transcript is gone.

## Fix options

1. **Resume-on-worker-restart**: for dev sessions with project-keyed `worker_key` (not chat-keyed), allow resume since delivery isn't needed — completion notification flows via `_handle_dev_session_completion` to steer the PM, not back to the chat
2. **Preserve context on abandonment**: when abandoning a local dev session, persist `claude_session_uuid` and `queued_steering_messages` so the PM can `valor-session resume --id <abandoned>` on next turn instead of re-creating
3. **PM-side awareness**: PM session detects its child was abandoned and resumes rather than re-spawns (ties into #1061's resume work)

Option 1 is the cleanest. Option 2 is the smallest change. Option 3 requires PM persona updates.

## Recon Summary

- **Origin of the `local` guard is correct:** `agent/session_health.py:255` (`is_local = entry.session_id.startswith("local")`) and the branch at 257-293 unconditionally abandons all such sessions on startup recovery. The rationale comment at 177-179 states the problem precisely: "The worker cannot deliver output for local sessions, so re-queuing them would spawn a second harness competing with the interactive CLI."
- **Dev-session completion does NOT require the user-facing send callback:** `agent/session_completion.py::_handle_dev_session_completion` (line 640+) routes dev completions through `PipelineStateMachine`, posts a GitHub stage comment, sets `retain_for_resume=True` on BUILD, and STEERS the parent PM via `queued_steering_messages`. It never calls `send_cb` on the dev session's own callback slot. This confirms Option 1's premise: local dev sessions have no output-delivery dependency.
- **`valor-session resume` already supports the needed statuses — but NOT `abandoned`:** `tools/valor_session.py:345` explicitly gates on `current_status not in ("completed", "killed", "failed")`. `abandoned` is a terminal status (per `TERMINAL_STATUSES` in `models/session_lifecycle.py:61`), so the existing resume tool rejects abandoned sessions with "Only completed/killed/failed sessions can be resumed." Fixing Option 2 or 3 requires extending this allowlist OR avoiding the abandoned state entirely (Option 1).
- **Resume requires `claude_session_uuid`:** `tools/valor_session.py:356` errors out if no UUID is stored. For dev sessions the UUID is populated on first SDK authentication; this is durable on the AgentSession hash and survives worker restarts. `queued_steering_messages` also lives on the record, so context IS already preserved for a terminal session — the blocker is purely the status gate.

**Revised:** 2 items — scope adjusted

- **Local-session guard must be scoped by `session_type`, not just `session_id` prefix.** PM and Teammate sessions with `session_id` starting with `local` genuinely cannot be resumed by the worker (no user-facing transport — the local CLI owns that). Only `session_type="dev"` local sessions have a deliverable completion path (PM steering). The fix is NOT to remove the abandonment — it is to add a dev-session bypass.
- **`is_project_keyed` is NOT the right discriminator.** The issue body suggests keying the bypass on "dev sessions with project-keyed `worker_key` (not chat-keyed)". Inspection of `models/agent_session.py:278-295` shows dev `worker_key` resolves to `slug` (if present) or `project_key` — never `chat_id`. So `is_project_keyed` is always True for dev sessions with a slug. `session_type == "dev"` is the cleaner, more direct check. Keep `session_id.startswith("local")` as the local-origin gate and add `session_type == DEV` as the resume-eligibility gate.

**Pre-requisites:** 1 item — must fix first (or address in scope)

- **`valor-session resume` allowlist extension:** If we choose Option 2 or Option 3 (preserve-abandon + resume), we MUST extend the allowlist in `tools/valor_session.py:345` to include `abandoned`. This is a small change but a real scope item. Option 1 (re-queue instead of abandoning) avoids the allowlist issue entirely because the session never reaches a terminal status.

- **PM-side detection of abandoned children (Option 3 as standalone):** The PM would need a pre-turn check that inspects `AgentSession.get_by_id(last_dev_session_id)` and, if `status == "abandoned"`, issue a `valor-session resume` instead of `valor-session create`. This is a persona-level behavior change in the PM system prompt and ties to unmerged persona work. Dropped as a standalone fix because Option 1 makes it unnecessary — the dev session never becomes abandoned, so the PM never sees a state change. Keep this pattern in mind for future hardening but do not block this issue on it.

### Decision

Proceed with **Option 1 (scoped resume-on-worker-restart)**: in `_recover_interrupted_agent_sessions_startup`, replace the unconditional abandonment with a branch that, for dev-role local sessions, re-queues to `pending` like a bridge session would. PM and Teammate local sessions continue to be abandoned (their original rationale still holds). This is the smallest, cleanest change and produces no new terminal state to recover from.

tomcounsell:
## Summary

Split the `is_local` branch in `_recover_interrupted_agent_sessions_startup` into three paths keyed on `session_type` so local **dev** sessions survive worker restart, while local **PM/teammate** sessions (plus pre-migration records with `session_type=None`) continue to be abandoned.

## Problem

On skills-only machines (no Telegram bridge), every dev session spawned by the PM has `session_id` starting `local-`. The prior startup-recovery guard abandoned all such sessions unconditionally, losing the in-flight transcript, tool cache, and reasoning state on every worker restart — even though the dev session had no competing human CLI process writing to the same `claude_session_uuid`. Long-running PM pipelines (build + test + review + docs + merge) could never complete across a worker restart.

### Production code (`agent/session_health.py`)
- Import `SessionType` alongside `AgentSession`.
- Split the `is_local` branch into three paths keyed on `session_type`:
  - **`SessionType.DEV`**: re-queue to `pending` with CAS on `running` (same pattern as bridge path)
  - **`SessionType.PM` / `SessionType.TEAMMATE`**: existing `finalize_session(..., "abandoned", ...)` flow preserved
  - **`session_type == None`** (pre-migration records): fall through to the abandon branch as the safer default
- The dev branch uses explicit equality with `SessionType.DEV` so future enum members (e.g. `REFLECTION`, `WORKFLOW`) also fall through to abandon.
- The summary log line now reports three counters: recovered bridge, recovered local-dev, abandoned local-PM/teammate.
- Updated the rationale comment at lines 177-187 to describe the split.
- Return value is `bridge_count + local_dev_count` (both are "sessions the worker will resume"); abandoned PM/teammate are still reported in the log line but not added to the return value.

### Tests (`tests/unit/test_recovery_respawn_safety.py`)
- `_mock_agent_session` now defaults `session_type=None` so tests must opt into a type.
- Renamed `test_startup_recovery_does_not_requeue_local_session` → `..._local_pm_session` and set `SessionType.PM`.
- Renamed `test_startup_recovery_abandons_local_sessions` → `..._local_pm_sessions` and set `SessionType.PM`.
- **New tests**:
  - `test_startup_recovery_requeues_local_dev_session` (dev → pending, CAS preserved)
  - `test_startup_recovery_recovers_local_dev_sessions` (dev → pending, `count=1`)
  - `test_startup_recovery_abandons_local_teammate_sessions` (teammate → abandoned)
  - `test_startup_recovery_local_dev_session_type_none_defaults_to_abandon` (pre-migration record → abandon)
- `test_startup_recovery_mixed_local_and_bridge` updated for the new 3-session scenario: PM abandoned, dev + bridge both re-queued, `count == 2`.

### Docs
- `docs/features/bridge-worker-architecture.md` — added "Local session recovery is `session_type`-aware (#1092)" subsection under Worker Restart Recovery and extended the summary table with local-dev and local-PM/teammate rows.
- `docs/features/pm-dev-session-architecture.md` — added a "Worker-restart persistence" callout under Dev session-specific fields.
- `docs/features/session-recovery-mechanisms.md` — updated the Startup Recovery row (what it does, local session guard wording) and the test-coverage table; added #1092 to Related Issues.

## Testing

- [x] `tests/unit/test_recovery_respawn_safety.py` (54 tests) — all pass, including 4 new #1092 tests
- [x] `tests/unit/test_session_health_phantom_guard.py` + `test_session_health_sibling_phantom_safety.py` — all pass
- [x] `python -m ruff format --check .` — passes
- [x] `python -m ruff check agent/session_health.py` — passes
- Note: `tests/unit/test_pm_session_permissions.py::test_sentry_token_injected_for_pm_session` fails on `main`; pre-existing failure unrelated to this change.

## Reproduction

**Before:** `./scripts/valor-service.sh worker-restart` with a running local dev session produces `Recovered 0 bridge session(s), abandoned 1 local session(s)`. PM's next turn re-creates the dev from scratch.

**After:** Same restart produces `Recovered 0 bridge session(s), 1 local dev session(s), abandoned 0 local PM/teammate session(s)`. Worker picks the session back up via `claude --resume <UUID>` on next scan.

## Definition of Done

- [x] Tested: 54 tests pass (4 new + all updated)
- [x] Reviewed: plan critique concerns embedded as Implementation Notes and addressed
- [x] Documented: 3 feature docs updated
- [x] Quality: ruff format + lint clean on touched file

Closes #1092

### DIFF
--- a/agent/session_health.py
+++ b/agent/session_health.py
@@ -318,7 +318,9 @@ def _recover_interrupted_agent_sessions_startup() -> int:
                 finalize_session(
                     entry,
                     "abandoned",
-                    reason="startup recovery: local PM/teammate session cannot be resumed by worker",
+                    reason=(
+                        "startup recovery: local PM/teammate session cannot be resumed by worker"
+                    ),
                     skip_auto_tag=True,
                 )
                 abandoned += 1
--- a/agent/session_health.py
+++ b/agent/session_health.py
@@ -11,8 +11,7 @@ from datetime import UTC, datetime
 from pathlib import Path
 
 from agent.session_state import SessionHandle, _active_events, _active_sessions, _active_workers
-from config.enums import SessionType
-from models.agent_session import AgentSession
+from models.agent_session import AgentSession, SessionType
 from models.session_lifecycle import TERMINAL_STATUSES as _TERMINAL_STATUSES
 
 logger = logging.getLogger(__name__)
--- a/agent/session_health.py
+++ b/agent/session_health.py
@@ -11,6 +11,7 @@ from datetime import UTC, datetime
 from pathlib import Path
 
 from agent.session_state import SessionHandle, _active_events, _active_sessions, _active_workers
+from config.enums import SessionType
 from models.agent_session import AgentSession
 from models.session_lifecycle import TERMINAL_STATUSES as _TERMINAL_STATUSES
 
@@ -174,9 +175,15 @@ def _recover_interrupted_agent_sessions_startup() -> int:
     where a worker transitions a session to running, then startup recovery resets it
     back to pending -- orphaning the already-spawned SDK subprocess.
 
-    Local CLI sessions (session_id starts with "local") are never re-queued. They are
-    marked "abandoned" instead. The worker cannot deliver output for local sessions, so
-    re-queuing them would spawn a second harness competing with the interactive CLI.
+    Local CLI sessions (session_id starts with "local") are handled by session_type:
+    - PM and Teammate local sessions are marked "abandoned". A live human CLI may hold
+      the same claude_session_uuid, so resuming would spawn a second harness competing
+      with the interactive CLI at that UUID (the #986 hijack rationale).
+    - Dev local sessions are re-queued to "pending" like bridge sessions. Dev sessions
+      are worker-owned (spawned via ``valor-session create --role dev`` by the PM) with
+      no human competitor — completion flows via _handle_dev_session_completion, which
+      steers the PM and never uses a user-facing send callback (#1092).
+    - Legacy records with ``session_type=None`` fall through to the safer abandon path.
 
     Note: The timing guard (AGENT_SESSION_HEALTH_MIN_RUNNING) is the primary defense
     against the hook-reactivation race. Hook reactivation transitions running→running
@@ -184,7 +191,9 @@ def _recover_interrupted_agent_sessions_startup() -> int:
     it — but truly stale sessions (>300s old) predate any active typing activity.
 
     Status is an IndexedField, so direct mutation and save is safe.
-    Returns the number of recovered bridge sessions (local sessions are not counted).
+    Returns the combined count of recovered bridge + local-dev sessions.
+    Abandoned local PM/teammate sessions are reported separately in the summary log
+    line but are NOT included in the return value.
     """
     # Phantom guard: drop records whose fields are still Popoto Field descriptors
     # (orphan $IndexF members). Destructive path — filter MUST run before any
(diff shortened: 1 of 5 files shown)


tomcounsell:
## Review: Approved

Clean, tightly-scoped fix that matches the plan exactly. The three-way split on `session_type` at `agent/session_health.py:271` (`is_local and session_type == SessionType.DEV`) uses explicit equality as the plan's Implementation Note required, so legacy `session_type=None` records and any future `SessionType` member automatically fall into the safer abandon path. CAS semantics (`expected_status="running"`) are preserved on the new dev re-queue path, matching the bridge path pattern.

### Verified
- [x] Code correctness — three branches match plan flow; PM/teammate abandon path unchanged; dev path mirrors bridge path
- [x] Test coverage — 54 tests pass including 4 new #1092 tests (dev requeue, dev recovery, teammate abandon, legacy None→abandon) and updated mixed-case test asserting count==2, finalize_calls==1, update_calls==2
- [x] Security — worker-internal change, no new inputs
- [x] Plan requirements met — all 5 Implementation Notes (enum-stability, legacy-record safety, CAS preservation, scope guard, mock consistency) applied
- [x] Neighboring tests pass — `test_session_health_phantom_guard.py` + `test_session_health_sibling_phantom_safety.py` (16 tests)
- [x] Docs updated — `bridge-worker-architecture.md`, `pm-dev-session-architecture.md`, `session-recovery-mechanisms.md` all reflect the split
- [x] `_agent_session_health_check` left untouched per No-Gos (scope guard respected)
- [x] Ruff format + lint clean on touched file

### Verification Results

| Check | Result |
|-------|--------|
| `pytest tests/unit/test_recovery_respawn_safety.py -x -q` | PASS (54/54) |
| `pytest tests/unit/test_session_health_phantom_guard.py tests/unit/test_session_health_sibling_phantom_safety.py -x -q` | PASS (16/16) |
| `python -m ruff check agent/session_health.py` | PASS |
| `python -m ruff format --check` (touched files) | PASS |
| `grep -n 'from models.agent_session import.*SessionType' agent/session_health.py` | PASS (line 14) |
| `grep -c 'local dev session' agent/session_health.py` | PASS (5 occurrences) |

No UI changes — screenshots not applicable.

Ready to merge.

PR #1143, opened 2026-04-23
test: orphan-PENDING recovery regression for #1126

### LINKED ISSUE #1126: Add regression test for _ensure_worker import in session_health orphan-PENDING branch
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/session-health-orphan-pending-test.md

> **Background.** Commit [`b40a2b73`](https://github.com/tomcounsell/ai/commit/b40a2b73) landed a one-line fix on main for an `UnboundLocalError` in `agent/session_health.py::_agent_session_health_check` — the orphan-PENDING recovery branch referenced `_ensure_worker` without importing it in that scope. The crash previously left orphan pending sessions permanently stuck, because the health check died before any PENDING processing finished. The fix is correct and already deployed, but it shipped as a direct-to-main commit (no PR review) and without a regression test. Issue [#1124](https://github.com/tomcounsell/ai/issues/1124) reported the same bug 16 minutes after the fix merged (reporter's working tree was behind main) and listed the missing test as acceptance criteria.

## Problem

The orphan-PENDING branch of `_agent_session_health_check` is now fixed on main, but the test suite has no case that exercises the exact failure mode the fix addressed:

- **No RUNNING sessions** (so the RUNNING-branch local import at `agent/session_health.py:948` never executes — this is what made Python treat `_ensure_worker` as a function-local variable and produced the `UnboundLocalError`).
- **At least one orphan PENDING session past `AGENT_SESSION_HEALTH_MIN_RUNNING`** whose worker is not alive and whose `worker_key` does not start with `"local"` (so the code reaches the `_ensure_worker` call site at `agent/session_health.py:1021`).

Without this test, the local import at `agent/session_health.py:1019` can be accidentally removed, relocated, or refactored away (e.g. consolidated into a shared helper) and CI will not catch the regression. The same scoping hazard exists symmetrically at the RUNNING-branch site — both branches use a function-local `from agent.agent_session_queue import _ensure_worker  # noqa: PLC0415`, and any control-flow reshuffle that changes which branch runs first could re-introduce the bug.

## Current behavior (after the fix)

- `agent/session_health.py:1019` imports `_ensure_worker` locally just before the call at line 1021.
- `agent/session_health.py:948` imports it locally inside the RUNNING-recovery branch.
- `_agent_session_health_check` runs cleanly in all existing test scenarios. **But** no test seeds the specific "orphan pending, zero running" topology, so removing either import would not fail the suite.

## Desired outcome

A regression test that fails on the pre-fix tree and passes on main. Specifically: with zero RUNNING sessions and ≥1 orphan PENDING session older than `AGENT_SESSION_HEALTH_MIN_RUNNING` on a non-local `worker_key`, `await _agent_session_health_check()` must complete without raising, and the orphan must be handled (worker started or transitioned to a terminal state).

## Definitions

| Term | What it is | Where to look |
|------|-----------|---------------|
| `_agent_session_health_check` | Periodic coroutine that recovers stuck RUNNING sessions and starts workers for orphan PENDING sessions | `agent/session_health.py` (the function under test) |
| `_ensure_worker` | Idempotent helper that starts a worker for a given worker key if one isn't already running; returns early if a worker exists | `agent/agent_session_queue.py` |
| `AGENT_SESSION_HEALTH_MIN_RUNNING` | Age threshold (seconds) a PENDING session must exceed before the health check treats it as an orphan and starts a worker for it | `agent/session_health.py` (module constant) |
| Orphan PENDING session | `AgentSession` with `status="pending"` whose `worker_key` has no live task in `_active_workers` | Pending-branch logic at `agent/session_health.py:970–1025` |
| `worker_key` | Chat-scoped or project-scoped key (e.g. `"local-..."` vs. a chat id) that serializes workers; the `"local"` prefix gates the abandoned-vs-started decision | `AgentSession.worker_key`; `agent/session_health.py:994` |

## Solution sketch

Add one test case. Two candidate homes, in order of preference:

1. **New case in `tests/integration/test_agent_session_health_monitor.py::TestAgentSessionHealthCheck`.** The class already imports the function and cleans up tasks spawned into `_active_workers` after each run (see `test_recovers_job_with_dead_worker` at `tests/integration/test_agent_session_health_monitor.py:161`). Add `test_recovers_orphan_pending_with_no_running_sessions`.
2. **New unit test `tests/unit/test_session_health_orphan_pending_recovery.py`**, following seeding patterns in `tests/unit/test_session_health_phantom_guard.py` and calling `_agent_session_health_check` the same way `tests/unit/test_session_health_sibling_phantom_safety.py` does.

Shape of the test:

1. Ensure `AgentSession.query.filter(status="running")` returns empty (seed no RUNNING records).
2. Seed one `AgentSession(status="pending")` with a non-`"local"` `worker_key` and `created_at` set far enough in the past to exceed `AGENT_SESSION_HEALTH_MIN_RUNNING`.
3. Ensure no worker is live for that `worker_key` in `agent.agent_session_queue._active_workers`.
4. Monkeypatch `agent.agent_session_queue._ensure_worker` to a spy (keeps the test from actually spinning a worker), or let it run and tear down like the existing integration case does.
5. `await _agent_session_health_check()`. Assert no exception.
6. Assert the spy was invoked with the seeded `worker_key` (confirms the orphan-PENDING branch reached the call site and didn't crash earlier).
7. Clean up any task in `_active_workers` / `_active_events` in teardown.

**Verification step (belongs in plan acceptance, called out here for clarity):** to confirm the new test would have caught the original bug, temporarily revert the import line at `agent/session_health.py:1019` in a local working copy and confirm the test fails with `UnboundLocalError: cannot access local variable '_ensure_worker'`. Restore before commit.

**Out of scope.** Refactoring the two function-local imports into a single module-level import. That is worth doing (circular-import risk appears low because `agent.agent_session_queue` does not import from `agent.session_health`) but belongs in its own issue so the refactor is reviewed independently. This issue is test-only.

## Related context

- **Original bug report:** [#1124](https://github.com/tomcounsell/ai/issues/1124) — includes the full traceback, root-cause analysis (function-scope hoisting of the local `import`), and two candidate fixes. Close as already-fixed by `b40a2b73` once this issue is filed, pointing at this issue for the missing test coverage.
- **Fix commit:** [`b40a2b73`](https://github.com/tomcounsell/ai/commit/b40a2b73) — `fix(#1110): add missing _ensure_worker import in orphan-pending health check branch`. Committed directly to `main`; there is no merged PR for it (`gh pr list --search "b40a2b73"` returns empty and `git branch --contains b40a2b73` shows only `main`).
- **Commit-message attribution typo:** the commit claims `fix(#1110)`, but [#1110](https://github.com/tomcounsell/ai/issues/1110) is titled *"Update, test, and comment popoto v1.5.0 integration ..."* — completely unrelated. Whoever authored the commit likely meant a different issue number. If a changelog is ever generated from commit messages, the `#1110` reference in that commit will link to the wrong issue. Non-blocking for this test work, but worth preserving here so the context isn't lost.
- **Adjacent tests** (read first for style and fixtures):
  - `tests/integration/test_agent_session_health_monitor.py` — existing `_agent_session_health_check` coverage (recovers RUNNING with dead worker; does NOT cover orphan PENDING).
  - `tests/unit/test_session_health_phantom_guard.py` — seeding patterns for `AgentSession` and phantom records.
  - `tests/unit/test_session_health_sibling_phantom_safety.py` — patterns for calling `_agent_session_health_check` with no live workers.

## Acceptance

- [ ] New test seeds the "0 running + ≥1 orphan pending past threshold + non-local `worker_key`" topology and calls `_agent_session_health_check` exactly once.
- [ ] Test asserts the call does not raise (would specifically fail on `UnboundLocalError` in the pre-fix tree).
- [ ] Test asserts the orphan-PENDING call site was reached (spy on `_ensure_worker` is invoked with the seeded `worker_key`, or equivalent assertion).
- [ ] Verified by local revert: removing the import at `agent/session_health.py:1019` makes the new test fail with `UnboundLocalError: _ensure_worker`.
- [ ] `pytest tests/unit/ tests/integration/test_agent_session_health_monitor.py` passes end-to-end.
- [ ] No production code changes — `agent/session_health.py` stays as-is on main.

## Downstream context

This issue will be consumed by `/do-plan` to produce a plan document at `docs/plans/session_health_orphan_pending_regression_test.md`. Scope is narrow: one test case, no production changes. `/do-plan` should NOT expand scope to refactor the two local imports — that is called out explicitly as out-of-scope above.

## Recon: Skipped

Recon not needed because the failure mode is precisely localized (two line numbers in one file), the fix is already on main and verified, test scaffolding exists in two adjacent modules, and the only remaining design choice is test placement (unit vs. integration) — a small judgment at plan time, not a scoping unknown.

tomcounsell:
## Summary

Adds a regression test for the orphan-PENDING branch of `_agent_session_health_check` in `agent/session_health.py`. Guards the fix from commit `b40a2b73` against silent regression.

## Changes

- `tests/integration/test_agent_session_health_monitor.py`: new method `TestJobHealthCheck::test_recovers_orphan_pending_with_no_running_sessions` (93 lines, strictly additive).

## What the test does

Seeds the exact topology that exposed the bug:
- zero RUNNING sessions
- one orphan PENDING session past `AGENT_SESSION_HEALTH_MIN_RUNNING`
- non-local `worker_key` (so the abandoned-local branch at `session_health.py:994` is skipped)

Then monkeypatches `agent.agent_session_queue._ensure_worker` (on the source module — the production code re-imports it locally on each call) and asserts:
1. The spy was called exactly once with `(seeded_session.worker_key, seeded_session.is_project_keyed)` — derived from the seeded session, not hard-coded.
2. No log record mentions `UnboundLocalError` or `cannot access local variable '_ensure_worker'`.

## Verification

- Locally reverted the `from agent.agent_session_queue import _ensure_worker` import at `agent/session_health.py:1019`.
- Confirmed the new test FAILS with `AssertionError: spy calls: []` and captured log: `UnboundLocalError: cannot access local variable '_ensure_worker' where it is not associated with a value`.
- Restored the import; test PASSES.
- `git diff main -- agent/` is empty (no production changes).
- Full `tests/integration/test_agent_session_health_monitor.py` (21 tests) passes.

## Testing

- [x] New regression test passes
- [x] Full integration file passes (21/21)
- [x] `ruff format` clean
- [x] Verified test catches the original bug via local revert
- [x] No production code changes (`git diff main -- agent/` empty)

## Definition of Done

- [x] Built: test added
- [x] Tested: passes on main, fails on pre-fix tree
- [x] Quality: `ruff format` clean
- [x] Documented: plan explicitly states no docs changes needed (test-only regression)

Closes #1126

### DIFF
--- a/docs/plans/session-health-orphan-pending-test.md
+++ b/docs/plans/session-health-orphan-pending-test.md
@@ -6,6 +6,7 @@ owner: Tom Counsell
 created: 2026-04-23
 tracking: https://github.com/tomcounsell/ai/issues/1126
 last_comment_id:
+revision_applied: true
 ---
 
 # Session Health: Orphan-PENDING Recovery Regression Test
@@ -223,6 +224,34 @@ Explicit statement for the hook: **No documentation changes needed.** This test-
 - [ ] Tests pass (`/do-test`).
 - [ ] No documentation updates needed (confirmed in Documentation section).
 
+## Concerns Addressed (Revision Pass)
+
+The `/do-plan-critique` run returned **READY TO BUILD (with concerns)** with four concerns. These are **not blockers** — they are acknowledged risks whose mitigations are now inline below so the builder can apply them directly while writing the test. Each concern maps 1:1 to the correspondingly numbered item in the Risks section above; this section adds a concrete **Implementation Note** the builder must follow.
+
+### C1: Spy resolution path (maps to Risk 1)
+
+- **Concern:** Monkeypatching the wrong name would let the test silently pass on the pre-fix tree.
+- **Implementation Note:** In the test, use `monkeypatch.setattr("agent.agent_session_queue._ensure_worker", spy)` — the module-attribute form on the source module, NOT on `agent.session_health`. Do this BEFORE calling `await _agent_session_health_check()`. The production code at `agent/session_health.py:1019` does a function-local `from agent.agent_session_queue import _ensure_worker`, which re-resolves the name from the source module on every invocation, so the module-attribute monkeypatch is the only form that intercepts. A comment at the monkeypatch line must name this fact explicitly (one-line comment; the builder is instructed to copy the rationale, not paraphrase): `# Patch on the source module — session_health re-imports _ensure_worker locally on each call (agent/session_health.py:1019).`
+
+### C2: Non-local `worker_key` guard (maps to Risk 2)
+
+- **Concern:** If helper defaults produce a `"local"`-prefixed `worker_key`, the health check takes the abandoned-local branch at `agent/session_health.py:994` and never reaches the `_ensure_worker` call site.
+- **Implementation Note:** After seeding, the test MUST assert `assert not seeded_session.worker_key.startswith("local"), f"topology drift: worker_key={seeded_session.worker_key!r} — this test exercises the non-local orphan-PENDING branch"`. This assertion fires BEFORE `_agent_session_health_check()` is invoked. If a future change to `_create_test_session` defaults makes the key `"local"`-prefixed, the test fails loudly at setup with a self-explanatory message instead of silently exiting via the wrong branch.
+
+### C3: Derived `worker_key`, not hard-coded (maps to Risk 3)
+
+- **Concern:** `_create_test_session` defaults could produce either a `chat_id`-keyed or a `project_key`-keyed session depending on `session_type`; hard-coding an expected string in the assertion would make the test brittle.
+- **Implementation Note:** The spy-invocation assertion MUST derive both arguments from the seeded session rather than hard-coding: `assert spy_calls == [(seeded_session.worker_key, seeded_session.is_project_keyed)], f"spy calls: {spy_calls!r}"`. Do NOT write `assert spy_calls == [("789", False)]` — if the helper defaults change, the derived form still exercises the orphan-PENDING branch correctly, while the hard-coded form fails for the wrong reason.
+
+### C4: `_active_workers` pre-flight cleanup (maps to Risk 4)
+
+- **Concern:** A prior test's leaked task in `_active_workers` for the same `worker_key` would make `worker_alive=True` at `agent/session_health.py:977`, causing the health check to skip the orphan-PENDING branch. The class-level autouse `_cleanup_workers` fixture cleans AFTER each test, but nothing guarantees a clean slate BEFORE this one runs.
+- **Implementation Note:** The test MUST explicitly pop the seeded key before invoking the health check: `from agent.agent_session_queue import _active_workers; _active_workers.pop(seeded_session.worker_key, None)`. Place this call AFTER seeding and the topology assertions, and BEFORE the monkeypatch and the `await`. Mirror the pattern already used by `test_recovers_job_with_no_worker` at `tests/integration/test_agent_session_health_monitor.py:200`.
+
+### Revision Note
+
+All four concerns have been addressed by inline Implementation Notes in this section and by corresponding bullets in Step 1 of "Step by Step Tasks" below. No scope expansion. No new production code changes. The test remains strictly additive.
+
 ## Team Orchestration
 
 ### Team Members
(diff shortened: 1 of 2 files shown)


tomcounsell:
Shepherded Phase 2 cleanup — regression test for #1126. Pipeline ran Plan → Critique → Build → Test → Review → Docs. Test seeds the 'zero RUNNING + orphan PENDING past threshold + non-local worker_key' topology and asserts no UnboundLocalError on the orphan-PENDING branch of _agent_session_health_check. Verified by local revert that the test would have caught the original bug. No production code changes. Merge gate is human-authorized.

PR #1054, opened 2026-04-19
refactor(#1026): collapse session_mode into session_type, remove role field

tomcounsell:
## Summary
- Collapses `session_mode` field into `session_type`, eliminating the overlapping `role` field
- Updates all consumers to use `session_type` as the canonical field
- Updates tests and documentation to reflect the new field naming

## Related Issue
Closes #1026

## Test plan
- [ ] `pytest tests/unit/` passes
- [ ] `pytest tests/integration/` passes
- [ ] No references to deprecated `session_mode` or `role` remain in active code

PR #1154, opened 2026-04-24
feat(#1084): categorise merge-gate baseline + refresh tool

### LINKED ISSUE #1084: Merge-gate baseline: categorise failures + refresh tool
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/merge-gate-baseline-refresh.md

> **Context:** The `/do-merge` skill runs a full-suite regression check before every PR merge by comparing pytest failures against `data/main_test_baseline.json` — a file that lists tests known to be failing on `main`. A PR is allowed to merge only if it introduces no *new* failures (ones not already in the baseline). This baseline is distinct from the `do-test` flaky filter + baseline verifier introduced by [#476](https://github.com/tomcounsell/ai/issues/476) / [PR #484](https://github.com/tomcounsell/ai/pull/484): that system operates on PR branches during the test stage; this one gates the merge itself.

## Problem

The merge-gate baseline is currently a flat list of failing test node IDs:

```json
{"failing_tests": ["tests/unit/foo.py::test_bar", "tests/unit/baz.py::test_qux"]}
```

Two weaknesses surfaced during the [#1041](https://github.com/tomcounsell/ai/issues/1041) cleanup:

1. **All failures are treated equally.** A test hanging indefinitely, a real pre-existing bug, an LLM-as-judge flake that picks a different wrong answer each run, and a module-level `ImportError` all show up as the same "pre-existing failure" bucket. When a PR introduces a new regression that *coincides in count* with a flaky failure that happens to pass this run, the `comm -23` comparison in `/do-merge` produces no diff and waves the regression through. We observed this on PRs #1054 and #1070 — identical 112-fail counts on main vs. PR branch, but different cases failing each run.
2. **No tool maintains the baseline.** Today the file is refreshed by a human running `pytest` and hand-editing the JSON. After 40+ commits landed between the #1041 baseline and HEAD, the file was stale by ~62 entries — most of the "failures" it listed were already fixed by unrelated PRs, and new failures had appeared that weren't in the file. That staleness is invisible until someone re-runs the suite.

**Current behaviour:** the `/do-merge` full-suite gate ([`.claude/commands/do-merge.md:224–292`](https://github.com/tomcounsell/ai/blob/main/.claude/commands/do-merge.md#L224)) compares a sorted list of failing node IDs from the PR run to the sorted baseline list via `comm -23`. Count-coincidental flakes can mask regressions; stale baselines can either over-block (false positives from fixed tests) or under-block (new real failures treated as pre-existing).

**Desired outcome:** the merge gate can distinguish *kinds* of pre-existing failures and refuses to let real regressions through even when a flaky case masks the count.

## Definitions

| Term | Definition |
|------|------------|
| **merge-gate baseline** | `data/main_test_baseline.json` — the list of failing tests on `main` used by `/do-merge` to decide whether a PR introduces regressions. Not the same as the [#476](https://github.com/tomcounsell/ai/issues/476) flaky filter, which operates on PR branches during `do-test`. |
| **LLM-as-judge test** | A pytest case whose assertion depends on calling Claude/Haiku at runtime. Without `temperature=0` + seed pinning, the same test can pass once and fail the next run with a different "wrong" answer. See `tests/unit/test_intake_classifier.py::TestRealHaikuClassification` and `tests/unit/test_work_request_classifier.py::TestLlmClassification` for examples. |
| **flaky failure** | A test that fails non-deterministically — passes on retry without any code change. Already handled at the PR-branch level by [#476](https://github.com/tomcounsell/ai/issues/476)'s flaky filter, but *not* by the merge-gate baseline. |
| **categorised baseline** | A proposed JSON schema that tags each pre-existing failure with a category: `real` (deterministic pre-existing bug), `flaky` (non-deterministic, e.g. LLM-judge), `hung` (times out), `import_error` (collection-time failure). |

## Recon Summary

Performed inline during the [#1041](https://github.com/tomcounsell/ai/issues/1041) cleanup — no separate fan-out needed since we had just exercised this exact surface.

**Confirmed**
- `data/main_test_baseline.json` is git-ignored (`.gitignore:181: data/`) and therefore per-machine; there's no single authoritative copy. This weakens the gate further — each developer's baseline drifts independently.
- `/do-merge` has a bootstrap path that writes the current failure list as the new baseline when the file is missing (`.claude/commands/do-merge.md:250–259`). No categorisation happens at bootstrap.
- After a clean merge, `/do-merge` resets the baseline to `{"failing_tests": []}` (line 265). This is the *only* automatic maintenance; everything else is manual.
- [#476](https://github.com/tomcounsell/ai/issues/476) and its PR #484 are complementary, not duplicative. That work addressed flaky failures at PR-branch test time via a retry-once step in `do-test` and deterministic junitxml parsing in `baseline-verifier`. It did not touch `data/main_test_baseline.json` or the merge gate.

**Revised**
- Original scope described these as two separate follow-ups ("per-case identity for flaky LLM-judge tests" and "`refresh_test_baseline.py`"). They're one feature: a richer baseline schema is useless without a tool to produce and maintain it, and a refresh tool is too limited without a richer output format.

**Pre-requisites**
- None. This builds on the existing baseline mechanism; it doesn't require #476 or any other in-flight work.

**Dropped**
- Auto-committing the baseline to git. The `data/` path is git-ignored on purpose (contains transient artifacts). A per-repo-but-not-per-machine solution is out of scope — can be addressed separately if needed.

## Solution Sketch

Open questions the planner should resolve, not prescriptive answers:

- **Schema shape.** What JSON structure best preserves the flat-list semantics the current `/do-merge` gate uses while adding categorisation? Options include a map keyed by node ID (`{"tests/...::foo": "flaky"}`) or a parallel list per category (`{"flaky": [...], "real": [...]}`). The planner should evaluate which is simplest for both `comm`-style diffing and human inspection.
- **Classification source.** How do we decide a failure is `flaky` vs. `real`? Options: run the suite N times and classify anything with <100% fail rate as flaky; rely on metadata markers (e.g. `@pytest.mark.flaky`); use the [#476](https://github.com/tomcounsell/ai/issues/476) retry infrastructure as an oracle. The planner should pick one and justify it.
- **Refresh tool surface.** Where does `refresh_test_baseline.py` live? `scripts/`? `tools/`? What arguments does it take (`--runs N`, `--categorise`, `--dry-run`)? What output does it produce — just a JSON write, or a diff summary?
- **Merge-gate comparison logic.** Today the gate uses `comm -23 <current> <baseline>`. A categorised baseline changes the comparison: new failures in `flaky`-category cases should not block; new failures in `real`-category cases should. The planner should spell out the new decision matrix and update `.claude/commands/do-merge.md` accordingly.
- **Migration.** Existing baselines in the wild are the flat schema. The tool and the merge gate should accept both formats for at least one release cycle.

## Acceptance Criteria

- [ ] `data/main_test_baseline.json` supports a categorised schema (structure chosen by planner) that distinguishes at least `real`, `flaky`, `hung`, `import_error`.
- [ ] A `refresh_test_baseline.py` tool (or equivalent) can regenerate the file from a pytest run, categorising failures automatically where possible.
- [ ] The `/do-merge` full-suite gate treats new failures in the `flaky`-category differently from new failures in the `real`-category — real regressions block, flaky "regressions" are reported but don't block.
- [ ] A PR that introduces a genuine regression masked by a count-coincident flaky failure is blocked by the gate (demonstrated by a regression test, not just an assertion in the PR body).
- [ ] Backwards compatibility: `/do-merge` still accepts the flat `{"failing_tests": [...]}` format for baselines produced before this change.
- [ ] `.claude/commands/do-merge.md` updated to document the new comparison logic.
- [ ] `tests/README.md` includes a one-paragraph note explaining the merge-gate baseline and its categories (so future contributors don't confuse it with the [#476](https://github.com/tomcounsell/ai/issues/476) flaky filter).

## Downstream context

This issue will be consumed by `/do-plan` to produce a plan document at `docs/plans/baseline-tooling-categorised.md` (final slug at planner's discretion). The plan should resolve the five open questions in the Solution Sketch before writing code.

tomcounsell:
## Summary
Replace the flat-list merge-gate baseline with a categorised schema-v2 map (`real`/`flaky`/`hung`/`import_error`) so `/do-merge`'s Full Suite Gate compares failures by identity, not just count. A count-coincident regression (as observed on PRs #1054/#1070) now blocks the merge. Adds `scripts/refresh_test_baseline.py` to regenerate the baseline from N pytest runs with per-test timeout classification via `pytest-timeout` (a dev-only plugin that never activates outside the refresh tool). Extracts the comparison logic into `scripts/baseline_gate.py` so every line is reachable from unit tests.

## Changes
- `scripts/refresh_test_baseline.py` — N-run aggregator with classifier precedence (`import_error` > `hung` > `real` > `flaky`), per-run junitxml isolation via `tempfile.TemporaryDirectory`, `ParseError`-safe parsing, dirty-tree `-dirty` commit suffix, `--merge` note preservation, `--dry-run` defaults to stdout.
- `scripts/baseline_gate.py` — pure-function `load_baseline` / `parse_pr_failures` / `compute_gate_verdict` / `format_staleness_warning`; shell-callable via `python -m scripts.baseline_gate --pr-junitxml ... --baseline ...` and exits 1 when new blocking regressions exist.
- `scripts/_baseline_common.py` — shared junitxml parsing helpers matching the exact `"Failed: Timeout >"` pytest-timeout prefix (not loose substring).
- `.claude/commands/do-merge.md` — Full Suite Gate rewritten to shell out to `baseline_gate.py`; schema-v2 bootstrap and post-merge reset paths.
- `pyproject.toml` + `uv.lock` — `pytest-timeout>=2.3` added as a dev extra; NOT in default pytest addopts.
- `tests/unit/test_refresh_test_baseline.py` + `tests/unit/test_do_merge_baseline.py` — 37 unit tests covering classifier precedence, exact-prefix timeout match, truncated-junitxml safety, dirty-tree commit capture, backwards-compat load, staleness warnings, and the PR #1054/#1070 count-coincident scenario.
- `docs/features/merge-gate-baseline.md` — new feature doc; `docs/features/README.md` and `tests/README.md` updated.

## Testing
- [x] 37 unit tests passing (`pytest tests/unit/test_refresh_test_baseline.py tests/unit/test_do_merge_baseline.py -v`)
- [x] `python -m ruff check scripts/refresh_test_baseline.py scripts/baseline_gate.py scripts/_baseline_common.py tests/unit/test_refresh_test_baseline.py tests/unit/test_do_merge_baseline.py` passes
- [x] `python -m ruff format --check` passes on all new/modified files
- [x] End-to-end `python -m scripts.baseline_gate` verified to block on count-coincident regression and allow flaky re-occurrence
- [x] End-to-end `python -m scripts.baseline_gate` verified to auto-promote schema-v1 `{"failing_tests": [...]}` to schema v2 in memory
- [x] `uv lock --check` passes

## Documentation
- [x] Created `docs/features/merge-gate-baseline.md`
- [x] Added index entry in `docs/features/README.md`
- [x] Updated `tests/README.md` with merge-gate-baseline vs flaky-filter paragraph
- [x] Updated `.claude/commands/do-merge.md` to document categories, migration, staleness warning

## Definition of Done
- [x] Documented: Feature doc created, all index/README updates in place
- [x] Quality: Lint and format checks pass on new code

Closes #1084


tomcounsell:
## Review: Changes Requested

The core design (schema-v2 baseline, `compute_gate_verdict`, staleness warning, count-coincident regression test) is well-executed and cleanly factored into `scripts/baseline_gate.py` + `scripts/_baseline_common.py` with 37 passing unit tests. The plan intent — blocking count-coincident regressions like PRs #1054/#1070 — is demonstrably met in `test_count_coincident_regression_is_blocked`.

However, verification of the refresh tool's main runtime path reveals a blocker: `scripts/refresh_test_baseline.py` cannot produce a baseline at all in its current form because its pytest invocation is guaranteed to crash. The unit tests never exercise the subprocess code path, so the issue is not caught by the suite.

### Blockers

- [ ] **`scripts/refresh_test_baseline.py:174-184`** — `cmd = ["pytest", "tests/", "-q", "--tb=no", f"--junitxml={junitxml_path}", "-p", "pytest_timeout", f"--timeout={test_timeout}", "--timeout-method=thread"]` — passing `-p pytest_timeout` when `pytest-timeout` is installed (which, per `pyproject.toml:43`, it always is via the `dev` extra) causes pytest to raise `ValueError: Plugin already registered under a different name: timeout=<module 'pytest_timeout' ...>` during setuptools entry-point loading. The plugin auto-registers as the short name `timeout`; forcing `-p pytest_timeout` attempts a second registration under the fully-qualified module name. I reproduced this both by running the exact command manually and by calling `run_pytest_once` directly — every invocation exits before writing junitxml. Combined with the fact that `run_pytest_once` returns `True` unless `subprocess.TimeoutExpired` fires (not on non-zero exit), this means every run is silently discarded by the downstream `JunitxmlParseError` path. The tool unconditionally exits 1 with "all N runs failed (outer timeout or junitxml ParseError)" and never writes a baseline file. Success criterion `scripts/refresh_test_baseline.py exists and regenerates the file from N pytest runs` is not actually met. Fix: drop `-p pytest_timeout` from the command line — the `--timeout=N` flag alone activates the plugin when it is installed — and/or add a test that invokes `run_pytest_once` against a tiny tempdir `tests/` tree and asserts the junitxml file is written.

- [ ] **`scripts/refresh_test_baseline.py:159-205`** — `run_pytest_once` — returns `True` on any subprocess completion including non-zero exits (intentional for test failures) but this masks pytest *startup* failures (config/plugin-registration errors, option-parse errors). A crash before junitxml is written is indistinguishable from "no tests failed". Fix: after a completed subprocess run, assert `junitxml_path.exists()` before returning `True`; if the file is missing, log a warning with `subprocess.run(..., capture_output=True)` stderr contents and return `False` so the caller can discard the run. Without this, the blocker above would be silent even to a human watching the terminal — the only signal is "all N runs failed" without the underlying cause. Add a unit test that mocks `subprocess.run` to simulate a pytest startup crash (exit non-zero, no junitxml written) and asserts `run_pytest_once` returns `False`.

### Tech Debt

- [ ] **`scripts/refresh_test_baseline.py` test coverage** — the `run_pytest_once` subprocess boundary has zero unit-test coverage. Every other function (`classify`, `aggregate_outcomes`, `capture_commit`, `build_baseline`, `resolve_output_path`, `load_existing_notes`) has a unit test, but the one function whose behavior is operationally critical is untested. This is how the blocker above landed invisible. Fix: add at least one test that (a) points `run_pytest_once` at a minimal fixture tree (e.g. `tmp_path` with a single-file `test_one.py`), asserts the invocation returns `True`, and asserts `junitxml_path.exists()` and parses; and (b) a second test that simulates a startup failure (e.g., by passing a bogus `pytest_args` entry) and asserts the failure is surfaced rather than swallowed.

- [ ] **`scripts/refresh_test_baseline.py:291` (`pytest tests/ --collect-only -q`)** — `estimate_test_count` shells to pytest with the exact same plugin-registration surface that the blocker exposes. On a machine where collection itself fails (e.g., an import error at `conftest.py`), it silently falls back to `1500`, hiding the real problem. Fix: log a warning when the "N tests collected" line is not found so the user knows the estimate is a fallback, not a measurement.

- [ ] **`.claude/commands/do-merge.md:298` (`GATE_OUTPUT=$(python -m scripts.baseline_gate ...)`)** — capturing stdout with `$(...)` drops the Python script's exit code when the assignment happens before `$?` is read. `GATE_EXIT=$?` on line 299 captures `python`'s exit code, which is correct, but the shell-script form is fragile: if any future edit inserts a command between the two lines (e.g., a debug `echo`), `$?` will reflect the inserted command, not the gate. Fix: use `if python -m scripts.baseline_gate ...; then ... else ... fi` or assign exit code inline via `GATE_OUTPUT=$(...); GATE_EXIT=$?` on the same line. Advisory — current form is correct today.

- [ ] **`scripts/baseline_gate.py:266` (`now = datetime.fromisoformat(args.now) if args.now else datetime.now(UTC)`)** — `--now` is only used by `main()`, not by `compute_gate_verdict` or `format_staleness_warning` directly. That's fine, but the parsed `now` has no `tzinfo` if the caller passes a naive ISO string — `format_staleness_warning` then fixes this by forcing `UTC` on `generated_at` but not on `now`, so a naive `--now` argument would be compared to a tz-aware `generated_at` and raise `TypeError: can't subtract offset-naive and offset-aware datetimes`. Fix: normalise `now` to UTC in `main()` if it has no `tzinfo`. Advisory — only affects the testing hook, and current tests always pass tz-aware values.

### Nits

- [ ] **`scripts/refresh_test_baseline.py:410` (`with tempfile.TemporaryDirectory(prefix="baseline-runs-") as parent_tmp:`)** — the directory name is good, but `parent_tmp` is a `str` and the per-run subdirs use `Path(parent_tmp) / f"run-{run_index}"`. Consistent use of `Path` throughout would remove the one-off conversion. Purely stylistic.

- [ ] **`scripts/refresh_test_baseline.py:299` (`return 1500  # Safe fallback matching current suite size order-of-magnitude.`)** — magic constant should be a named `_FALLBACK_TEST_COUNT_ESTIMATE = 1500` at module level for readability and to keep the `1500` in `estimate_test_count()` (appears twice) in sync.

### Verified

- [x] 37 unit tests passing locally (`pytest tests/unit/test_refresh_test_baseline.py tests/unit/test_do_merge_baseline.py -v`)
- [x] `python -m ruff check` and `python -m ruff format --check` pass on all new files
- [x] `scripts/baseline_gate.py` CLI works end-to-end; verdict JSON shape matches the contract consumed by `.claude/commands/do-merge.md:298-311`
- [x] Count-coincident regression test (`test_count_coincident_regression_is_blocked`) correctly reproduces the PR #1054/#1070 scenario
- [x] Schema-v1 → schema-v2 in-memory promotion works (`test_load_baseline_promotes_legacy_flat_list_in_memory`)
- [x] Staleness warning triggers on 14-day-old `generated_at`, `bootstrap: true`, and `-dirty` commit suffix
- [x] `docs/features/merge-gate-baseline.md` exists and is indexed in `docs/features/README.md:73`
- [x] `tests/README.md` updated with merge-gate-baseline paragraph
- [x] `pytest-timeout>=2.3` added to `[project.optional-dependencies].dev` (not to default pytest addopts, per plan)

### Plan Alignment

Plan's acceptance criteria mostly met on paper — schema-v2, categorisation, backwards compat, staleness warning, regression test, docs. The one criterion actually broken in practice is: *`scripts/refresh_test_baseline.py` exists and regenerates the file from N pytest runs*. The file exists but cannot regenerate anything until the `-p pytest_timeout` blocker is removed.

tomcounsell:
## Review: Approved

Re-review after `60cb97d4` ("harden refresh tool, gate error handling, gate exit-capture") and follow-up `b2650a67` ("CLI-level regression test for naive --now normalization"). All prior findings — 2 blockers, 4 tech-debt items, 1 of 2 nits — have been resolved with regression coverage that would catch the original failures if reintroduced. The remaining nit (Path consistency on `parent_tmp_str`) was purely stylistic and is not material.

### Prior Findings — Resolution Audit

| Severity | Prior Finding | Resolved In | Verification |
|---|---|---|---|
| Blocker | `-p pytest_timeout` causes pytest plugin double-registration (ValueError) | `scripts/refresh_test_baseline.py:197-204` | Flag removed; `--timeout=N` alone activates plugin. Inline comment 190-196 documents the rationale. |
| Blocker | `run_pytest_once` returns True on startup crashes that never write junitxml; downstream ParseError masks root cause | `scripts/refresh_test_baseline.py:230-245` | Now checks `junitxml_path.exists()`, captures stderr, logs tail, returns False. |
| Tech debt | `run_pytest_once` had zero unit-test coverage (how the blocker landed invisible) | `tests/unit/test_refresh_test_baseline.py:366-518` | 4 new tests: real subprocess against minimal tree, mocked plugin-registration crash, outer timeout, stderr-tail-in-warning. |
| Tech debt | `estimate_test_count` silently falls back to 1500 on collection failure | `scripts/refresh_test_baseline.py:331-375` | `logger.warning` fires when summary line is missing; constant extracted to `_FALLBACK_TEST_COUNT_ESTIMATE`. |
| Tech debt | `do-merge.md` `GATE_OUTPUT=$(...)` then `GATE_EXIT=$?` on separate lines is fragile | `.claude/commands/do-merge.md:301` | Same-line `; GATE_EXIT=$?` with comment 298-300 explaining why. |
| Tech debt | `baseline_gate.py --now` naive ISO string crashes with `TypeError` on tz-aware subtraction | `scripts/baseline_gate.py:266-275` + `tests/unit/test_do_merge_baseline.py:281-331` | Naive value normalised to UTC in `main()` before passing to `format_staleness_warning`. New CLI-boundary regression test exercises the exact failure mode. |
| Nit | `parent_tmp` `str -> Path` one-off conversion | n/a | Not changed; purely stylistic, advisory. |
| Nit | `1500` magic constant | `scripts/refresh_test_baseline.py:70` | Extracted to `_FALLBACK_TEST_COUNT_ESTIMATE` at module level; also referenced in two log warnings. |

### Verified

- [x] **42/42 unit tests passing** (`uv run python -m pytest tests/unit/test_refresh_test_baseline.py tests/unit/test_do_merge_baseline.py -v`) — 5 new tests on top of the original 37.
- [x] `python -m ruff check` and `python -m ruff format --check` pass on all 5 changed files.
- [x] End-to-end `python -m scripts.baseline_gate --pr-junitxml ... --baseline ... --now "2026-04-21T00:00:00"` (naive `--now`) completes cleanly with proper staleness warning — verifies the prior `TypeError` regression is gone.
- [x] `--help` for both scripts renders. CLI surface unchanged from the prior review except for the `--now` UTC-normalisation fix.
- [x] Patch commit `60cb97d4` and follow-up `b2650a67` present on origin; PR head SHA `b2650a67`; mergeStateStatus CLEAN; mergeable MERGEABLE; state OPEN.
- [x] Plan acceptance criterion *"`scripts/refresh_test_baseline.py` exists and regenerates the file from N pytest runs"* — now actually achievable (the prior blocker made this fail).
- [x] Count-coincident regression test (`test_count_coincident_regression_is_blocked`) still reproduces the PR #1054/#1070 scenario.

### New Concerns

None. The patch commits are tightly scoped, well-commented, fully covered by tests, and traceable line-for-line to the prior review's findings. No fresh blockers, tech-debt, or non-stylistic nits identified.

### Plan Alignment

All four acceptance criteria from the plan now hold in practice (not just on paper):

1. Schema-v2 categorised baseline emitted — verified by `test_build_baseline_emits_schema_v2_with_commit_and_tests`.
2. `scripts/refresh_test_baseline.py` regenerates from N pytest runs — verified end-to-end by `test_run_pytest_once_writes_junitxml_for_real_minimal_tree` (real subprocess, real junitxml, no startup crash).
3. `/do-merge` blocks count-coincident regressions — verified by `test_count_coincident_regression_is_blocked`.
4. Backwards compat with schema-v1 flat list — verified by `test_load_baseline_promotes_legacy_flat_list_in_memory`.

Closes the prior reviewer's only outstanding concern.

tomcounsell:
## Review: Approved

All findings from the prior review (commit 60cb97d4 + b2650a67) have been addressed and verified end-to-end. The previously-blocking `-p pytest_timeout` command-line flag is gone, `run_pytest_once` now surfaces startup crashes (missing-junitxml detection + captured stderr tail), `estimate_test_count` warns on the `1500` fallback, `$?` is captured on the same line as `$(...)` in `do-merge.md`, and the naive `--now` ISO string is normalised to UTC before the tz-aware comparison inside `format_staleness_warning`.

All 42 unit tests pass in the repo venv (with `pytest-timeout==2.4.0` installed from the dev extras). End-to-end `python -m scripts.baseline_gate` runs produce the correct verdicts for schema-v2 baselines, schema-v1 legacy promotion, flaky re-occurrence, new blocking regressions, and staleness warning triggers (both tz-aware and naive `--now`).

### Verified
- [x] All 42 unit tests pass (`pytest tests/unit/test_refresh_test_baseline.py tests/unit/test_do_merge_baseline.py -v`)
- [x] `python -m ruff check` passes on all new/modified files
- [x] `python -m ruff format --check` passes (5 files already formatted)
- [x] Blocker 1 fixed: `-p pytest_timeout` removed from `run_pytest_once` command (line 197-204); `--timeout=N` alone activates the plugin
- [x] Blocker 2 fixed: `run_pytest_once` asserts `junitxml_path.exists()` after subprocess completion (line 230-245); returns `False` with stderr tail logged when pytest crashed before writing junitxml
- [x] Tech-debt 1 fixed: three new regression tests (`test_run_pytest_once_writes_junitxml_for_real_minimal_tree`, `test_run_pytest_once_returns_false_when_pytest_crashes_before_junitxml`, `test_run_pytest_once_returns_false_on_outer_timeout`, `test_run_pytest_once_logs_stderr_tail_on_startup_crash`) close the subprocess-boundary coverage gap
- [x] Tech-debt 2 fixed: `estimate_test_count` now logs a warning when the `_FALLBACK_TEST_COUNT_ESTIMATE` (extracted as named constant at module scope) fires, distinguishing "collection failed" from "collected 1500 tests"
- [x] Tech-debt 3 fixed: `GATE_OUTPUT=$(...); GATE_EXIT=$?` on the same line in `.claude/commands/do-merge.md:301`; inline comment documents the requirement
- [x] Tech-debt 4 fixed: `baseline_gate.main()` attaches UTC to naive `--now` values (line 272-273); regression test `test_main_normalises_naive_now_to_utc` guards this at the CLI boundary
- [x] Nit 1 fixed: `parent_tmp_str` is converted to `Path` immediately on line 477 so downstream usage is Path-consistent
- [x] Nit 2 fixed: `_FALLBACK_TEST_COUNT_ESTIMATE = 1500` is now a module-level named constant used in both call sites
- [x] End-to-end `python -m scripts.baseline_gate` verified with (a) schema-v2 baseline, new regression blocks (exit 1); (b) schema-v1 legacy promotion works; (c) staleness warning fires on 30-day-old generated_at with naive `--now`; (d) flaky re-occurrence reported, not blocked
- [x] `--merge` flag note preservation works (`load_existing_notes` reads schema-v2 `note` fields correctly)
- [x] Cruft audit clean — "legacy" / "schema-v1" references are intentional, describe the v1 → v2 migration path, and match the plan's terminology

### Plan Alignment

Every acceptance criterion in `docs/plans/merge-gate-baseline-refresh.md` is now met in practice, not just on paper:

1. Schema-v2 baseline with `real`/`flaky`/`hung`/`import_error` categories — shipped in `scripts/_baseline_common.py`.
2. `scripts/refresh_test_baseline.py` exists and regenerates the file from N pytest runs — now verified end-to-end (the prior blocker that made this criterion false is fixed).
3. `scripts/baseline_gate.py` with pure-function core, shell-callable, exits 1 on `new_blocking_regressions` — verified.
4. Count-coincident regression is blocked — `test_count_coincident_regression_is_blocked` asserts the PR #1054/#1070 scenario.
5. Backwards-compat with schema-v1 flat shape — `test_load_baseline_promotes_legacy_flat_list_in_memory` passes.
6. Staleness warning on age > 14d, `bootstrap: true`, `-dirty` suffix — all three triggers covered by unit tests; CLI-boundary naive `--now` fix now has its own regression guard.
7. Docs shipped — `docs/features/merge-gate-baseline.md`, `docs/features/README.md` index entry, `tests/README.md` paragraph, `.claude/commands/do-merge.md` updated.

### Notes

- No UI changes; no screenshots.
- The one local environmental wrinkle discovered during verification (uv extracting `pytest_timeout.whl` without writing `pytest_timeout.py` to site-packages) is a uv/venv-corruption issue on my machine, not a PR defect. The lockfile is correct, the `pyproject.toml` `[project.optional-dependencies].dev` entry is correct, and the manual wheel extraction confirmed that with the file physically present, every test passes.

Ready to merge.

PR #1177, opened 2026-04-26
fix(#1172): PM session liveness — see progress or stay graceful

tomcounsell:
## Summary

Implements [issue #1172](https://github.com/tomcounsell/ai/issues/1172) per the plan at `docs/plans/pm-session-liveness.md`. Two pillars in one PR (the plan recommended a 2-PR phased approach but the briefing requested both together):

- **Pillar B — Evidence-only detector**: retire the staleness-inference kill paths from #1046 (`STDOUT_FRESHNESS_WINDOW`, `FIRST_STDOUT_DEADLINE`) and the per-session wall-clock cap (`_get_agent_session_timeout`, `AGENT_SESSION_TIMEOUT_DEFAULT`/`BUILD`). Preserve every evidence-based path: #1036 dual-heartbeat OR check, #944 own-progress fields, #1099 Mode 4 OOM defer, #918 delivery guard.
- **Pillar A — In-flight visibility**: four new `AgentSession` fields (`current_tool_name`, `last_tool_use_at`, `last_turn_at`, `recent_thinking_excerpt`) written by pre/post tool-use hooks and the SDK stream-event handler. New centralized writer `agent/hooks/liveness_writers.py` enforces a 5s per-session cooldown and fail-closed semantics. Dashboard `_session_to_json` exposes the four fields plus a derived `last_evidence_at`.
- **PM mid-work self-report**: `agent/session_completion._emit_pm_self_report` sends one short templated `valor-telegram send` to `PM: {project_name}` between the first dev-child completion and the final delivery, gated by the new `self_report_sent_at` cap field.

## Acceptance criteria

- [x] A PM session running 4+ hours with active tool use and no result event is NOT killed
  (covered by `tests/integration/test_pm_long_run_no_kill.py`)
- [x] A PM session that completes work emits exactly 1 mid-work message + 1 final delivery
  (covered by `tests/integration/test_pm_goldilocks_messaging.py`)
- [x] `_has_progress` no longer references `STDOUT_FRESHNESS_WINDOW` or `FIRST_STDOUT_DEADLINE`
- [x] `_get_agent_session_timeout` deleted; per-session wall-clock cap is gone
- [x] `_tier2_reprieve_signal` retains `compacting`, `alive`, `children`; `stdout` gate gone
- [x] Dashboard `dashboard.json` exposes `current_tool_name`, `last_tool_use_at`, `last_turn_at`, `recent_thinking_excerpt`, `last_evidence_at`
[...]

PR #1243, opened 2026-05-01
fix(session-health): promote per-turn SDK signals to Tier 1 to detect hung sessions (#1226)

### LINKED ISSUE #1226: session-liveness-check fails to detect hung SDK: Tier 1 heartbeat is refreshed by watchdog regardless of progress
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/sdlc-1226.md

> **Context:** The `session-liveness-check` reflection (registered in [`config/reflections.yaml`](../blob/main/config/reflections.yaml)) runs `_agent_session_health_check()` in [`agent/session_health.py`](../blob/main/agent/session_health.py) on a fixed cadence. It uses a two-tier no-progress detector ([#1036](https://github.com/tomcounsell/ai/issues/1036) / [PR #1039](https://github.com/tomcounsell/ai/pull/1039)) to decide whether a session is making forward progress. Tier 1 evaluates progress signals; Tier 2 evaluates reprieve signals. If Tier 1 says "no progress" AND Tier 2 has no reprieve, the session is recovered (subprocess killed, row marked killed). Full model: [`docs/features/pm-session-liveness.md`](../blob/main/docs/features/pm-session-liveness.md).

## Problem

`session-liveness-check` cannot detect a session that is hung from the very first turn, because Tier 1's input signal is wrong. The signal it consults — `last_sdk_heartbeat_at` — is refreshed by a watchdog tick whose only requirement is "the SDK subprocess exists." A wedged subprocess that never makes progress still keeps the heartbeat fresh forever.

**Current behavior:**

`agent/messenger.py:328` — `BackgroundTask._watchdog` runs every 60s while the SDK subprocess is alive and unconditionally calls `self.messenger.notify_heartbeat_tick()`. That callback writes `now()` into `AgentSession.last_sdk_heartbeat_at` (see [`agent/session_executor.py:1138-1139`](../blob/main/agent/session_executor.py#L1138)).

`agent/session_health.py:_has_progress` (lines 465–472) treats a fresh `last_sdk_heartbeat_at` as positive evidence of progress under its "Tier 1: dual-heartbeat OR" check. Because the watchdog is the writer, the field is effectively a "watchdog tick is alive" signal, not a "SDK is making progress" signal — but `_has_progress` treats it as the latter. Result: any subprocess that exists keeps the heartbeat fresh, so a wedged SDK passes Tier 1 forever and is never recovered.

The watchdog log line right next to the tick already includes the actual progress signal — `communicated=False` — but that field is only logged, never fed back into the health check.

The other own-progress fields in `_has_progress` (`turn_count > 0`, `log_path` non-empty, `claude_session_uuid` non-empty) are all sticky once set. They are valid as "session ever started" signals but are useless for detecting mid-run hangs.

### Concrete instance (2026-04-30)

| Field | Value |
|---|---|
| Session | `0_1777532128936` (PM, slug `sdlc-1219`) |
| `agent_session_id` | `d3d75f39d179459a9cb74c309a76d98b` |
| Model | `--model opus` |
| System prompt size | 74,769 chars |
| Started | 2026-04-30 06:55:29 UTC |
| Heartbeat pattern | `SDK heartbeat: running Ns, communicated=False` every 60s, climbing from 60s through 10380s+ (≈2.9 hours) |
| `communicated` | `False` for the entire ~3-hour window |
| Tool calls / output / turn completion | None |
| `session-liveness-check` invocations during the hang | ~30+ (every minute, see `logs/worker.log`) |
| Recoveries triggered | 0 |
| Resolution | Manual `valor-session kill` |

Worker log evidence: `grep "0_1777532128936" logs/worker.log | grep heartbeat` shows the full pattern. The reflection had every opportunity to recover the session and didn't, because Tier 1 was lying.

**Desired outcome:**

A session that authenticates, spawns a subprocess, but never produces any structured SDK output (assistant message, tool_use, tool_result) is recovered within a defined freshness window — not after manual intervention hours later. A session that produces output once and then wedges mid-run is also recovered, after the same freshness window, with Tier 2 reprieves still applying.

## Definitions

|---|---|---|
| `session-liveness-check` | Reflection that runs `_agent_session_health_check()` on a fixed cadence to detect and recover stuck sessions. | [`config/reflections.yaml`](../blob/main/config/reflections.yaml), [`agent/session_health.py`](../blob/main/agent/session_health.py) |
| Tier 1 (dual-heartbeat OR) | First gate in `_has_progress`: returns True if either `last_heartbeat_at` (queue-layer) or `last_sdk_heartbeat_at` (messenger-sourced) is fresh within `HEARTBEAT_FRESHNESS_WINDOW` (90s). | [`agent/session_health.py:467`](../blob/main/agent/session_health.py#L467) |
| Tier 2 reprieve gates | Activity-positive checks (`compacting`, `children`, `alive`) that prevent recovery when the subprocess is legitimately busy. | [`_tier2_reprieve_signal`](../blob/main/agent/session_health.py#L490) |
| `last_sdk_heartbeat_at` | DatetimeField on `AgentSession`. Written every 60s by `BackgroundTask._watchdog` via `notify_heartbeat_tick` regardless of whether the SDK made progress. | [`models/agent_session.py:257`](../blob/main/models/agent_session.py#L257), [`agent/messenger.py:328`](../blob/main/agent/messenger.py#L328) |
| `has_communicated()` | Method on `BossMessenger` returning `True` if the messenger has ever sent a message to the supervisor. Sticky: once True, always True. | [`agent/messenger.py:147`](../blob/main/agent/messenger.py#L147) |
| `notify_stdout_event()` | Existing messenger callback for SDK output stream events; separate from the watchdog tick. Currently exists but stdout-as-progress was retired by [#1172](https://github.com/tomcounsell/ai/issues/1172). | [`agent/messenger.py:98`](../blob/main/agent/messenger.py#L98) |
| Hung-from-first-turn | A session whose subprocess is alive but produces no structured SDK output (no assistant messages, no tool_use, no tool_result) for an extended window. | This issue |

## Prior Context

- **#1036 / PR #1039** — introduced the two-tier no-progress detector. Defined `last_sdk_heartbeat_at` and the Tier 1 dual-heartbeat OR check. This issue does not abandon that design; it tightens Tier 1's input signal so it actually means what the comment claims it means.
- **#1172** — retired the previous stdout-staleness gate. Established that "recent stdout is not evidence of progress" because long-thinking turns and large tool outputs produce legitimate stdout silence. The replacement signal proposed in this issue is structured SDK message events, not raw stdout — the #1172 finding does not apply.
- **#1218** — covers the *inverse* direction (DB row terminal but subprocess alive). Different problem; complementary.
- **#944** — covers shared-worker liveness false-positives. Different problem.
- **#1099** — added the `compacting` Tier 2 reprieve gate. Must continue to apply under the new signal.

## Solution Sketch

The Tier 1 input signal should be a per-turn progress timestamp — refreshed when the SDK actually emits output, not when the watchdog ticks. A reasonable owner is the messenger, since it already observes the SDK output stream.

**Open questions for the planner:**

1. **Source of the new signal.** Should it be driven by `notify_stdout_event` (already exists, currently unused for liveness)? Or by a new callback fired specifically on assistant message / tool_use / tool_result deltas? The latter is more semantic; the former requires less wiring. Investigate which event types are actually observable from the messenger today.
2. **Field naming and lifecycle.** New field on `AgentSession` (e.g. `last_communication_ts`)? Or rename `last_sdk_heartbeat_at` and change its writer? Renaming carries migration cost on existing rows; new field carries dual-bookkeeping cost while old field is deprecated.
3. **Freshness window.** What is the longest legitimate single-turn duration without any structured SDK output? Long-thinking turns and large tool outputs were the motivation for [#1172](https://github.com/tomcounsell/ai/issues/1172). The new window must accommodate them — likely 15–30 minutes — but should be measured against real session telemetry, not guessed. The Tier 2 reprieve gates (`compacting`, `children`, `alive`) continue to apply on top of this window.
4. **Disposition of `last_sdk_heartbeat_at`.** Once `_has_progress` no longer consults it, is there any remaining consumer (queue-layer health, dashboard)? If not, drop it entirely. If yes, document its narrowed semantics ("watchdog tick alive, NOT progress").
5. **Stickiness of own-progress fields.** `turn_count > 0`, `log_path` non-empty, `claude_session_uuid` non-empty are all sticky. They detect "ever started," not "still progressing." Should `_has_progress` keep them as standalone passing signals, or pair each with a per-turn freshness check? The first-turn-hang case in this issue would pass Tier 1 today on `claude_session_uuid` alone if the dual-heartbeat OR check were tightened — that's a second leak that must be closed.
6. **Interim one-liner.** Would gating `notify_heartbeat_tick()` on `messenger.has_communicated()` ship today and catch first-turn hangs only, while the per-turn redesign is planned? `has_communicated()` is sticky, so it does NOT cover mid-run hangs — but it would have caught session `0_1777532128936`. The planner should decide whether to ship this interim fix immediately or only ship the full design.

The bias of this issue is toward a thorough redesign of the Tier 1 input signal, not a one-line patch. The watchdog-tick-as-progress conflation is the architectural mistake; per-turn freshness is the architectural correction.

## Recon Summary

**Confirmed:**
- The leak path `_watchdog` → `notify_heartbeat_tick` → `last_sdk_heartbeat_at` → `_has_progress` Tier 1 fresh → no recovery is real and was observed end-to-end on session `0_1777532128936`.
- The two writers of `last_sdk_heartbeat_at` are both in `agent/session_executor.py:1138-1151`, both invoked by the watchdog callback path. No other writer exists.
- The messenger already exposes `notify_stdout_event` and `get_last_message_time` (timestamp of last delivered message). Some primitive for per-turn progress already exists; it just isn't surfaced into the health check.
- `has_communicated()` is sticky (`len(messages_sent) > 0`), confirmed at `agent/messenger.py:147`.
- The Tier 2 reprieve gates (`compacting`, `children`, `alive`) are independent of Tier 1's signal and are the right way to avoid false-positives on legitimately busy subprocesses.

**Revised:**
- Original draft proposed Option A (gate `notify_heartbeat_tick` on `has_communicated`) as the right fix. Recon shows that's only an interim patch — it catches first-turn hangs only because `has_communicated` is sticky. Full fix requires a per-turn timestamp.

**Pre-requisites:**
- None blocking. The new signal can be added as an additive field alongside `last_sdk_heartbeat_at`, with `_has_progress` migrated and the old field retired in a follow-up commit.

**Dropped:**
- "Use stdout-as-progress" — explicitly retired by [#1172](https://github.com/tomcounsell/ai/issues/1172). The replacement is structured SDK message events, not raw stdout.

## Acceptance Criteria

- [ ] A new per-turn progress timestamp (e.g. `last_communication_ts`) is introduced on `AgentSession`, written when the SDK emits an assistant message, tool_use, or tool_result.
- [ ] `_has_progress` Tier 1 evaluates freshness on the new timestamp instead of (or in addition to, with stricter semantics than) `last_sdk_heartbeat_at`.
- [ ] `last_sdk_heartbeat_at` is either removed entirely or its remaining consumers are documented and audited; `_has_progress` no longer reads it as a progress signal.
- [ ] The own-progress fields in `_has_progress` (`turn_count`, `log_path`, `claude_session_uuid`) no longer pass Tier 1 on their own without a per-turn freshness check, OR a justification is documented for keeping them as standalone signals.
- [ ] Test: an SDK that authenticates, writes watchdog heartbeats, and never communicates → `_has_progress` returns False after the freshness window, and `_agent_session_health_check` recovers the session.
- [ ] Test: a session that communicates once and then wedges → `_has_progress` returns False after the freshness window, with Tier 2 reprieves still applying.
- [ ] Tier 2 reprieve gates (`compacting`, `children`, `alive`) continue to function correctly under the new signal — verified by an existing-coverage audit or new test.
- [ ] [`docs/features/pm-session-liveness.md`](../blob/main/docs/features/pm-session-liveness.md) is updated to describe the new signal, its writer, the freshness window, and the deprecation/removal of `last_sdk_heartbeat_at`.

## Downstream

This issue feeds `/do-plan` → `docs/plans/{slug}.md`.

Constraints for the planner:
- Must not regress legitimate long-thinking turns or large-tool-output turns (the original motivation for [#1172](https://github.com/tomcounsell/ai/issues/1172)).
- Must not regress paused / dormant sessions waiting on humans (covered by [#1128](https://github.com/tomcounsell/ai/issues/1128)).
- The interim one-liner (Option A: gate `notify_heartbeat_tick` on `has_communicated`) is acceptable to ship first **only if** the full per-turn signal is also planned in the same plan document — not as a substitute.

tomcounsell:
## Summary

- **Root cause fixed**: `_has_progress` Tier 1 was reading `last_sdk_heartbeat_at` (written by `BackgroundTask._watchdog` every 60s on subprocess *existence*, not SDK output), making hung sessions indistinguishable from working ones. Sessions could hang for hours without being recovered.
- **New Tier 1 sub-check A**: `last_tool_use_at` and `last_turn_at` (written by PreToolUse/PostToolUse hooks and `sdk_client.py` on `result` event) replace the watchdog-tick as the authoritative Tier 1 progress signal, checked against `SDK_PROGRESS_FRESHNESS_WINDOW` (1800s default).
- **Sub-check B preserved**: `last_heartbeat_at` (queue-layer) still passes Tier 1 when `sdk_ever_output=False`, preserving startup-window behavior and backward compatibility with sessions from before PR #1177.
- **Reprieve escalation guard**: `MAX_NO_OUTPUT_REPRIEVES` (default 20 ticks = 30 min) suppresses Tier 2 "alive" reprieve for sessions that have never produced SDK output, ensuring eventual recovery instead of indefinite reprieve.
- **Startup recovery reprieve reset**: `_recover_interrupted_agent_sessions_startup` now resets `reprieve_count=0` to prevent escalation guard from triggering immediately after worker restart.

## Test plan

- [x] `pytest tests/unit/test_health_check_recovery_finalization.py -v` — 59 passed (15 new tests + 44 existing)
- [x] New `TestHasProgressPerTurnSignal` class: 11 tests for sub-check A/B semantics and `last_sdk_heartbeat_at` exclusion
- [x] New `TestTier2ReprieveEscalation` class: 3 tests for the escalation guard (cap, bypass-when-output, below-cap)
- [x] New `TestStartupRecoveryReprieveCountReset`: structural test verifying `reprieve_count` reset in startup recovery
- [x] `TestHasProgressDualHeartbeat::test_sdk_heartbeat_within_window_returns_true` replaced with `test_sdk_heartbeat_alone_not_progress` asserting the corrected behavior
- [x] `python -m ruff check agent/session_health.py` — 0 errors
- [x] `python -m ruff format --check agent/session_health.py` — 1 file already formatted
- [x] `docs/features/pm-session-liveness.md` updated with Tier 1 signal reference table, watchdog-alive note, escalation guard section

Closes #1226

### DIFF
--- a/agent/session_health.py
+++ b/agent/session_health.py
@@ -158,6 +158,20 @@ MAX_RECOVERY_ATTEMPTS = 2
 # health-check tick budget tight while still giving the cancellation a
 # moment to complete.
 TASK_CANCEL_TIMEOUT = 0.25
+# Freshness window (seconds) for per-turn SDK progress signals
+# (last_tool_use_at, last_turn_at) in Tier 1 sub-check A (issue #1226).
+# 1800s (30 minutes) accommodates the longest observed extended-thinking turns.
+# A session that produces a tool boundary or result event within this window is
+# considered actively progressing. Env-tunable via SDK_PROGRESS_FRESHNESS_WINDOW_SECS
+# for operators who observe edge cases in production.
+SDK_PROGRESS_FRESHNESS_WINDOW = int(os.environ.get("SDK_PROGRESS_FRESHNESS_WINDOW_SECS", 1800))
+# Max consecutive Tier 2 reprieves allowed for sessions that have NEVER produced
+# any SDK output (sdk_ever_output=False). After this many reprieves the "alive"
+# gate is suppressed and recovery proceeds. Derived from the SDK progress window
+# divided by the heartbeat write interval: 1800 // 90 = 20 ticks (~30 minutes).
+# Sessions that have produced output (sdk_ever_output=True) are never subject to
+# this cap — their recovery depends solely on per-turn freshness in sub-check A.
+MAX_NO_OUTPUT_REPRIEVES = SDK_PROGRESS_FRESHNESS_WINDOW // HEARTBEAT_FRESHNESS_WINDOW  # 20
 
 
 # In-process cache for ``_is_memory_tight()`` (issue #1099 Mode 4). Tuple of
@@ -327,7 +341,10 @@ def _recover_interrupted_agent_sessions_startup() -> int:
                 update_session(
                     entry.session_id,
                     new_status="pending",
-                    fields={"priority": "high", "started_at": None},
+                    # reprieve_count reset to 0: prevents escalation guard firing
+                    # immediately after recovery if the session had accumulated
+                    # reprieves before startup (issue #1226 Risk 4).
+                    fields={"priority": "high", "started_at": None, "reprieve_count": 0},
                     expected_status="running",
                     reason="startup recovery: local dev session",
                 )
@@ -403,7 +420,10 @@ def _recover_interrupted_agent_sessions_startup() -> int:
                 update_session(
                     entry.session_id,
                     new_status="pending",
-                    fields={"priority": "high", "started_at": None},
+                    # reprieve_count reset to 0: prevents escalation guard firing
+                    # immediately after recovery if the session had accumulated
+                    # reprieves before startup (issue #1226 Risk 4).
+                    fields={"priority": "high", "started_at": None, "reprieve_count": 0},
                     expected_status="running",
                     reason="startup recovery",
                 )
@@ -441,43 +461,84 @@ def _has_progress(entry: AgentSession) -> bool:
     silence is no longer a kill signal — long-thinking turns and large tool
     outputs produce legitimate stdout silence.
 
-    Signals (any one is sufficient):
+    Tier 1 is evaluated in two sub-checks (A and B). Any one passing → return True.
 
-    1. **Dual-heartbeat OR (#1036, retained verbatim).** Either
-       ``last_heartbeat_at`` (queue-layer, written by ``_heartbeat_loop``)
-       OR ``last_sdk_heartbeat_at`` (messenger-sourced, written by
-       ``BackgroundTask._watchdog``) fresher than
-       ``HEARTBEAT_FRESHNESS_WINDOW`` (90s) ⇒ progress.
-    2. **Own-progress fields (#944 / #963, retained).**
-       - ``turn_count > 0`` — at least one turn boundary observed.
-       - ``log_path`` non-empty — first log entry written.
-       - ``claude_session_uuid`` non-empty — SDK authenticated.
-    3. **Child-progress check (#944, retained).** A PM session with at
-       least one non-terminal child is not stuck. ``get_children()``
-       returns ``[]`` on failure with a WARNING log; no outer try/except
-       needed.
+    **Sub-check A: Per-turn SDK progress (#1226).**
+    ``last_tool_use_at`` (written by PreToolUse/PostToolUse hooks) and
+    ``last_turn_at`` (written by sdk_client on ``result`` event) are evidence
+    of actual structured SDK output. Either field fresher than
+    ``SDK_PROGRESS_FRESHNESS_WINDOW`` (1800s, 30 min) ⇒ progress.
+    ``last_sdk_heartbeat_at`` is written by ``BackgroundTask._watchdog`` on
+    subprocess existence — it is a watchdog-alive signal, NOT a progress signal.
+    It is intentionally excluded from sub-check A (issue #1226).
 
-    Returns ``False`` only when EVERY signal is absent (both heartbeats
-    stale, all own-progress fields empty, no live children). The caller
+    **Sub-check B: Startup-window executor-alive fallback (#1036 narrowed).**
+    When ``sdk_ever_output`` is False (neither per-turn field has ever been set),
+    ``last_heartbeat_at`` (queue-layer, written by ``_heartbeat_loop``) fresher
+    than ``HEARTBEAT_FRESHNESS_WINDOW`` (90s) ⇒ progress. This preserves the
+    pre-#1226 behavior for sessions in their startup window before the first SDK
+    tool or turn event fires, and for sessions predating PR #1177 (whose hooks
+    did not write the per-turn fields).
+
+    **Own-progress fields (#944 / #963, narrowed by #1226).**
+    - ``turn_count > 0`` — at least one turn boundary observed.
+    - ``log_path`` non-empty — first log entry written.
+    - ``claude_session_uuid`` non-empty — SDK authenticated.
+    These are sticky once set and cannot detect mid-run hangs. They are only
+    evaluated when ``sdk_ever_output`` is False — once the SDK has produced a
+    tool or turn event, sub-check A is the authoritative Tier 1 signal.
+
+    **Child-progress check (#944, retained).**
+    A PM session with at least one non-terminal child is not stuck.
+    ``get_children()`` returns ``[]`` on failure with a WARNING log; no outer
+    try/except needed.
+
+    Returns ``False`` only when EVERY signal is absent. The caller
     (``_agent_session_health_check``) then evaluates ``_tier2_reprieve_signal``
     before deciding to recover.
     """
-    # Tier 1: dual-heartbeat OR check (#1036). Fresh on either signal → progress.
     now_utc = datetime.now(tz=UTC)
-    for hb_attr in ("last_heartbeat_at", "last_sdk_heartbeat_at"):
-        hb = getattr(entry, hb_attr, None)
+
+    # Compute sdk_ever_output once — used by both sub-check A and the own-progress
+    # field guard. True iff last_tool_use_at or last_turn_at has ever been written.
+    sdk_ever_output = bool(
+        getattr(entry, "last_tool_use_at", None) or getattr(entry, "last_turn_at", None)
+    )
+
+    # Sub-check A: per-turn SDK activity (issue #1226).
+    # last_tool_use_at (PreToolUse/PostToolUse hooks) and last_turn_at (result event)
+    # are evidence of actual structured SDK output. Fresh within
+    # SDK_PROGRESS_FRESHNESS_WINDOW → real progress.
+    for progress_attr in ("last_tool_use_at", "last_turn_at"):
+        ts = getattr(entry, progress_attr, None)
+        if isinstance(ts, datetime):
+            ts_aware = ts if ts.tzinfo else ts.replace(tzinfo=UTC)
+            if (now_utc - ts_aware).total_seconds() < SDK_PROGRESS_FRESHNESS_WINDOW:
+                return True
+
+    # Sub-check B: startup-window executor-alive fallback (#1036 retained, narrowed).
+    # Use last_heartbeat_at as a Tier 1 signal ONLY before the SDK has produced any
+    # tool or turn output. Once sdk_ever_output is True, sub-check A is authoritative.
+    # Backward-compatible: sessions from before PR #1177 (no tool/turn fields) fall
+    # here and behave identically to the pre-#1226 behavior.
+    if not sdk_ever_output:
+        hb = getattr(entry, "last_heartbeat_at", None)
         if isinstance(hb, datetime):
             hb_aware = hb if hb.tzinfo else hb.replace(tzinfo=UTC)
             if (now_utc - hb_aware).total_seconds() < HEARTBEAT_FRESHNESS_WINDOW:
                 return True
 
-    # Own-progress fields (preserves #944 / #963 invariants).
-    if (entry.turn_count or 0) > 0:
-        return True
-    if bool((entry.log_path or "").strip()):
-        return True
-    if bool(entry.claude_session_uuid):
-        return True
+    # Own-progress fields (#944 / #963, narrowed by #1226).
+    # Only evaluated when sdk_ever_output is False — once the SDK has produced
+    # a tool or turn event, per-turn freshness (sub-check A) is authoritative.
+    if not sdk_ever_output:
+        if (entry.turn_count or 0) > 0:
+            return True
+        if bool((entry.log_path or "").strip()):
+            return True
+        if bool(entry.claude_session_uuid):
+            return True
+
     # Child-progress check: a PM session with active children is not stuck.
     # get_children() queries via Popoto parent_agent_session_id index and
     # returns [] on failure with a WARNING log — no outer try/except needed.
@@ -514,6 +575,15 @@ def _tier2_reprieve_signal(
     Returns the name of the first passing gate ("compacting", "children", or
     "alive"), or ``None`` if every gate fails.
 
+    **Reprieve escalation guard (issue #1226):** When ``sdk_ever_output`` is
+    False (the session has never produced a tool or turn event) AND
+    ``reprieve_count >= MAX_NO_OUTPUT_REPRIEVES``, all gates are suppressed
+    and ``None`` is returned immediately. This prevents indefinite alive-but-
+    silent sessions from being reprieved forever. Sessions with
+    ``sdk_ever_output=True`` (have produced output) are never subject to this
+    cap — their recovery depends solely on per-turn freshness in
+    ``_has_progress`` sub-check A.
+
     The previous "stdout" gate (and its ``STDOUT_FRESHNESS_WINDOW`` constant)
     was retired by issue #1172. Recent stdout is no longer evidence the
     subprocess is making progress — long-thinking turns and large tool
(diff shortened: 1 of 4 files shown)


tomcounsell:
## Review: Approved

This PR correctly fixes the root cause of issue #1226: `last_sdk_heartbeat_at` (written unconditionally by `BackgroundTask._watchdog` on subprocess existence) was the only Tier 1 signal, making hung sessions indistinguishable from working ones. The fix promotes actual SDK output signals (`last_tool_use_at`, `last_turn_at`) to Tier 1 and adds an escalation guard for never-output sessions.

### Verified

- [x] **Code correctness** — `_has_progress` sub-check A checks `last_tool_use_at` / `last_turn_at` against `SDK_PROGRESS_FRESHNESS_WINDOW` (1800s). Sub-check B preserves startup-window behavior via `last_heartbeat_at` only when `sdk_ever_output=False`. `last_sdk_heartbeat_at` is entirely absent from `_has_progress` (plan verification check confirmed exit code 1).
- [x] **Root bug fixed** — `last_sdk_heartbeat_at` no longer passes Tier 1. Verified by `test_sdk_heartbeat_alone_not_progress` (replaces the old `test_sdk_heartbeat_within_window_returns_true` which tested the buggy behavior).
- [x] **Escalation guard** — `_tier2_reprieve_signal` suppresses "alive" gate when `sdk_ever_output=False` and `reprieve_count >= MAX_NO_OUTPUT_REPRIEVES` (20 ticks = 30 min). Sessions with `sdk_ever_output=True` are unaffected.
- [x] **Startup reprieve reset** — `_recover_interrupted_agent_sessions_startup` resets `reprieve_count=0` on recovery, preventing immediate escalation after worker restart (Risk 4).
- [x] **Backward compatibility** — Sessions predating PR #1177 (no per-turn fields) fall through to sub-check B (`last_heartbeat_at` when `sdk_ever_output=False`) — identical to pre-#1226 behavior.
- [x] **Test coverage** — 59 passed (15 new tests + 44 existing). New classes: `TestHasProgressPerTurnSignal` (11 tests), `TestTier2ReprieveEscalation` (3 tests), `TestStartupRecoveryReprieveCountReset` (1 structural test). Existing `TestHasProgressDualHeartbeat` updated with corrected semantics.
- [x] **Documentation** — `docs/features/pm-session-liveness.md` updated with Tier 1 signal reference table (A/B sub-check terminology), watchdog-alive note, escalation guard section, and test references.
- [x] **Lint clean** — `ruff check` 0 errors, `ruff format --check` already formatted.
- [x] **Plan requirements met** — All success criteria from `docs/plans/sdlc-1226.md` verified.
- [x] **Security** — No vulnerabilities found. No secrets, no external inputs. Env-tunable `SDK_PROGRESS_FRESHNESS_WINDOW_SECS` has a safe default (1800s).
- [x] **No-Gos respected** — `last_sdk_heartbeat_at` retained in model and written by watchdog (for dashboard `last_evidence_at`); only removed from `_has_progress`. No changes to `BackgroundTask._watchdog`, `_on_heartbeat_tick`, `_heartbeat_loop`, `liveness_writers.py`, or model fields.

### Verification Results

| Check | Result |
|-------|--------|
| `pytest tests/unit/test_health_check_recovery_finalization.py -v` | 59 passed |
| `python -m ruff check agent/session_health.py` | 0 errors |
| `python -m ruff format --check agent/session_health.py` | already formatted |
| `grep last_sdk_heartbeat_at agent/session_health.py \| grep _has_progress` | empty (exit 1) — confirmed |
| `grep last_tool_use_at\|last_turn_at agent/session_health.py` | 11 lines — confirmed |

### Legacy Cruft

No legacy cruft introduced. `last_sdk_heartbeat_at` is explicitly narrowed to "watchdog-alive signal only" in docstring and feature doc. The no-gos correctly call out full field retirement as a follow-up chore once dashboard consumers are audited.

### Blockers
- None

### Tech Debt
- None

### Nits
- None

tomcounsell (commented):
## Code Review — Approve

Thorough review of the Tier 1 promotion of per-turn SDK signals (`last_tool_use_at`, `last_turn_at`) and the Tier 2 escalation guard (`MAX_NO_OUTPUT_REPRIEVES`). The fix correctly addresses the root cause described in #1226: a watchdog tick that runs while the subprocess merely *exists* was being read as evidence of *progress*. Replacing that signal with per-turn writes from `liveness_writers` is the right architectural correction.

### Correctness — Liveness/Timeout Semantics

The two-level structure is preserved and tightened sensibly:

- **Sub-check A** (`agent/session_health.py:512-517`): `last_tool_use_at` / `last_turn_at` against `SDK_PROGRESS_FRESHNESS_WINDOW` (1800s). Either fresh → progress=True. Both writers (PreToolUse/PostToolUse hooks, `sdk_client.py` on `result` event) emit on real SDK output, not subprocess existence.
- **Sub-check B** (`agent/session_health.py:524-529`): `last_heartbeat_at` only when `sdk_ever_output=False`. Preserves startup window and pre-#1177 backward compatibility.
- **Own-progress fields** (`agent/session_health.py:534-540`): correctly gated on `not sdk_ever_output`, so once the SDK has produced a tool/turn event the sticky fields no longer mask a mid-run hang.
- **Escalation guard** (`agent/session_health.py:601-610`): `_tier2_reprieve_signal` returns `None` when `sdk_ever_output=False AND reprieve_count >= MAX_NO_OUTPUT_REPRIEVES (=20)`. This closes the original 3-hour hang reproducer because Tier 2's "alive" gate can no longer reprieve a never-output session indefinitely.

False-positive risk is well managed: the 1800s window is generous for extended-thinking turns, the cap only fires for sessions that have *never* produced output, and sessions with output rely solely on per-turn freshness — no risk of killing a working session that's mid-tool-call.

### Risk 4 (recovery reprieve reset) — Verified

Both `update_session` calls in `_recover_interrupted_agent_sessions_startup` reset `reprieve_count=0`:
- Local-dev path: `agent/session_health.py:347` — `fields={"priority": "high", "started_at": None, "reprieve_count": 0}`
- Bridge path: `agent/session_health.py:426` — same pattern.

This prevents post-restart sessions from immediately tripping the escalation guard. (Follow-up worth considering: audit other re-queue paths like `_create_continuation_pm` for the same reset, but it is non-blocking — at worst one extra recovery cycle.)

### Tests

`tests/unit/test_health_check_recovery_finalization.py` adds three classes covering the new behavior:
- `TestHasProgressPerTurnSignal` — sub-check A/B, watchdog-tick exclusion, own-progress gating, constant existence.
- `TestTier2ReprieveEscalation` — guard fires at cap with no output, does NOT fire when output exists, does NOT fire below cap.
- `TestStartupRecoveryReprieveCountReset` — structural check that startup recovery references `reprieve_count`.

The Archaeologist critique was addressed: tests use explicit `_ago(1860)` for "stale beyond window," not the silently-passing `_ago(300)` pattern. The replaced `test_sdk_heartbeat_alone_not_progress` correctly asserts the OPPOSITE of the previous buggy behavior.

PR description claims 59 tests pass (15 new + 44 existing) — the test code itself reflects this structure.

### Documentation

`docs/features/pm-session-liveness.md` updates are accurate:
- Adds the Tier 1 signal reference table with writer/window/active-when columns.
- Notes `last_sdk_heartbeat_at` is retained for dashboard `last_evidence_at` only.
- Documents the escalation guard and the startup-recovery reprieve reset.
- A/B sub-check terminology is consistent across code, docstrings, and feature doc.

### Plan Adherence

All Success Criteria in `docs/plans/sdlc-1226.md` are reflected in code: constant added, sub-check A/B implemented, `last_sdk_heartbeat_at` removed from Tier 1, own-progress gated, escalation guard added, `reprieve_count=0` reset in startup recovery, docs updated. Plan status correctly bumped to `In Review`.

### Standards/Lint

PR description reports `ruff check` and `ruff format --check` clean for `agent/session_health.py`. Diff is small and focused (only `agent/session_health.py`, the one test file, the feature doc, and the plan frontmatter); no unrelated changes.

### Approve

This is the right architectural fix at the right layer (sdk_client + hooks), with conservative defaults, real recovery guarantees, and well-targeted tests. No blocking issues.

tomcounsell:
## Review: Approved (re-review after ruff fix)

Re-review of commit `fc962838` on top of previously-approved `0c275b8a`. The new commit is a pure whitespace change wrapping two over-long assertion messages to satisfy ruff E501 — no behavioral change.

### Diff scope (since approved HEAD)
- `tests/unit/test_health_check_recovery_finalization.py` — 4 insertions, 2 deletions
- Two `assert` failure messages wrapped using Python adjacent-string-literal concatenation. Resulting messages are identical to the originals when rendered (no embedded space loss; the wrap point preserves the trailing space in the first literal).

### Verified
- [x] `python -m ruff check tests/unit/test_health_check_recovery_finalization.py` — All checks passed
- [x] `python -m ruff format --check tests/unit/test_health_check_recovery_finalization.py` — already formatted
- [x] All 14 tests in `TestHasProgressPerTurnSignal` and `TestTier2ReprieveEscalation` pass
- [x] Wrapped messages preserve exact content when concatenated (no missing/duplicated space)
- [x] No changes to production code — substantive review of `0c275b8a` still stands

### Blockers
- None

### Tech Debt
- None

### Nits
- None

Mergeability preflight: `state=OPEN`, `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN`.

Issue #1109, opened 2026-04-22
PM sessions silently fail when created without --slug: working_dir + message_text truncation


tomcounsell:
## Problem

Two compound defects make PM-spawned dev sessions fail silently when a PM session is created without \`--slug\`.

### Defect 1: working_dir/branch contamination

\`tools/valor_session.py:194\` defaults \`working_dir\` to \`_repo_root\` (the worker's cwd). \`agent/agent_session_queue.py:445\` then snapshots the current branch of that directory into \`AgentSession.branch_name\`. When the worker was started while the repo was on a feature branch (e.g. \`session/sdlc-1067\`), every PM session created without \`--slug\` inherits that branch. No per-issue worktree is created. Startup-recovery branch-restore (\`agent/agent_session_queue.py:505-520\`) will later check out the wrong branch. PR creation fails because \`slug=None\`.

### Defect 2: child session message_text truncation

\`AgentSession.initial_telegram_message.message_text\` appears to be capped at 500 chars somewhere in the PM→dev child-spawn path. When the PM enriches the dev payload with PROJECT / FROM / SESSION_ID / TASK_SCOPE / SCOPE headers (~500 chars of boilerplate), the actual task content is truncated off. The dev receives only routing boilerplate ending mid-word ("other senders, o"). The dev correctly reports truncation and exits; the PM interprets this as \`outcome=success\` at \`agent/session_completion.py:799\` and emits \`[PIPELINE_COMPLETE]\` without any code being written.

## Reproducer

PM session \`7abb0a95fb15453eb117322824462bc1\` for issue #1098. Enqueued via \`python -m tools.valor_session create --role pm --message "..."\` (no \`--slug\`). Completed in 8 min with \`pr_url=None, slug=None, branch_name=session/sdlc-1067\`. Two dev children completed with no commits.

## Impact

Every \`valor-session create --role pm\` without \`--slug\` is at risk when the worker's cwd is not on \`main\`. Silent failures waste a session slot and mislead the enqueuer into thinking work completed.

## Acceptance

- [ ] \`valor-session create --role pm\` targeting an issue must auto-derive \`slug=sdlc-{N}\` from the message and provision a worktree, OR refuse to start with a clear error
- [ ] Child session \`message_text\` must preserve the full PM instruction payload — raise the cap, store a reference, or store only the enriched payload (not a 500-char truncated prefix)
[...]

Issue #1195, opened 2026-04-28
Continuation-PM creation crashes silently with session_id=None and working_dir=None


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/sdlc-1195.md

## Recon Summary

**Verified at filing time** (2026-04-28):

- **Confirmed**: `agent/session_completion.py:358` `_AgentSession.create(...)` is missing both `session_id` and `working_dir` kwargs. Verified by reading the call site against the model's expected fields.
- **Confirmed**: `agent/session_executor.py:634` calls `Path(session.working_dir)` unguarded — `Path(None)` raises `TypeError`.
- **Confirmed**: `agent/session_revival.py:20` `_session_branch_name` calls `sanitize_branch_name(session_id)` which would fail on None at the crash-snapshot fallback.
- **Confirmed**: `_resolve_working_dir_for_parent` exists at `agent/session_completion.py:712` and is already used at `agent/session_completion.py:479` for dev spawns — fix can reuse it directly.
- **Confirmed**: log evidence — grep on `logs/bridge.log.*` shows `[continuation-pm-created] new_session_id=None` firing 15+ times since 2026-04-23; most recent victim PR #1194.
- **Pre-requisite**: none — fix is self-contained in two files, requires no schema changes.
- **Dropped from scope**: rewriting `_AgentSession.create` signature to require fields. Defensive validation is an open question (see below); the immediate fix populates the fields at the call site.

**Open questions surfaced for the planner** (see Solution sketch below):
1. Should the continuation PM inherit the parent's `slug` for worktree resolution?
2. Should `_AgentSession.create()` validate required fields defensively, in addition to executor guards?

> **Context for new readers:** This repo runs an SDLC pipeline (Plan → Build → Test → Patch → Review → Docs → Merge) where a **PM-role `AgentSession`** orchestrates and a **Dev-role `AgentSession`** does the work. When a dev session finishes, the system spawns a follow-up "**continuation PM**" session that picks up the dev's result and advances to the next stage. That auto-spawn is what's broken. See [`docs/features/pm-dev-session-architecture.md`](https://github.com/tomcounsell/ai/blob/main/docs/features/pm-dev-session-architecture.md) for the architecture.

## Problem

[...]

PR #1280, opened 2026-05-05
guard slugless dev sessions with synthetic slug + CLI symmetry (#1272)

### LINKED ISSUE #1272: Guard against parallel-session main-checkout contamination — explore no-side-effects designs (#887)
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/parallel-session-checkout-guard.md

> **Context:** This is a **design exploration**, not a request to implement a specific lock. We want to guard against parallel-session contamination of the main checkout (the failure surface of [#887](https://github.com/tomcounsell/ai/issues/887)) **without** introducing a stale-block failure mode. The user's hard constraint is that any guard must be fail-open: a buggy implementation must not be able to wedge the system. The output of `/do-plan` may legitimately be "we picked alternative X with rationale Y," or even "none of these are worth the residual risk."
>
> Background concept — **fail-open vs. fail-closed locking:** A *fail-closed* lock blocks work when its state is uncertain (e.g., crashed holder, missing PID file, clock skew). A *fail-open* lock degrades to no-op under uncertainty. Fail-closed is safer for correctness but can wedge a multi-machine system; fail-open is safer for liveness but admits the very contamination we're guarding against.
>
> Background concept — **fazm-style directory lock:** [mediar-ai/fazm](https://github.com/mediar-ai/fazm/blob/main/CLAUDE.md) uses a directory-based filesystem lock at `/tmp/fazm-build.lock` for multi-agent build coordination. The lock dir contains a `pid` file so other waiters can probe liveness; stale-lock detection reclaims the lock if the holder PID is dead. fazm's CLAUDE.md ("Multi-Agent Safety" / "Before Running run.sh" sections) explicitly forbids manual `rm -rf` on the lock — even with PID-based stale detection and a 5-minute max-wait, the design assumes operators will respect the lock as a cooperative protocol. **The cautionary frame matters for this issue:** even fazm's careful design admits that a stale lock is crippling, and the user has explicitly said a stale block is a non-starter for this system.

## Problem

Issue [#887](https://github.com/tomcounsell/ai/issues/887) addressed the most direct version of this failure: PM sessions created via `valor-session create` without a prior `/do-plan` step bypassed worktree provisioning, causing dev sub-sessions to run `git checkout session/{slug}` inside the main checkout at `/Users/valorengels/src/ai`. The fix added three enforcement layers (worktree provisioning failure escalation, main-checkout protection guard, PM persona prompt instruction) plus a `--slug` flag on `valor-session create` that provisions the worktree at session creation time. See [`docs/features/session-isolation.md` § Worktree Enforcement for Dev Sessions (Issue #887)](https://github.com/tomcounsell/ai/blob/main/docs/features/session-isolation.md).

**The residual question this issue explores:** the #887 fix relies on every dev-session entry point passing through `agent/agent_session_queue.py::_execute_agent_session()` or the CLI's `--slug` path. Those layers are correct today. But:

1. They depend on `AgentSession.slug` being set. A dev session without a slug bypasses both guards by design.
2. They are code-path-specific. Future entry points (a new harness, a debugging tool, a test fixture) could re-introduce the bypass without a global "is the main checkout currently safe to mutate?" check.
3. They do not protect against a *human* working in the main checkout having their state stomped by a programmatic write from another process — they only ensure dev sessions go to a worktree.

The residual contamination shape: two parallel writers (any combination of agent + agent, agent + human, agent + cron job) both believe they own the main checkout. Examples we've seen or could see:
- A `/do-build` skill running `pytest` / `ruff` against the main checkout to verify a change before opening a PR
- A reflection job (memory-dedup, autoexperiment) writing to the repo
- A nightly test runner mutating files
- A human running `git checkout` in the same shell where a dev session is still active

## Definitions

| Main checkout | The non-worktree working tree at `/Users/valorengels/src/ai` (the `repo_root` returned by `git rev-parse --show-toplevel` from outside any `.worktrees/{slug}/` directory) | [`agent/worktree_manager.py:98-142`](https://github.com/tomcounsell/ai/blob/main/agent/worktree_manager.py) |
| Worktree | An isolated checkout under `.worktrees/{slug}/` with its own branch (`session/{slug}`), filesystem state, and `.git` pointer file. Created and managed by `agent/worktree_manager.py`. | [`docs/features/session-isolation.md`](https://github.com/tomcounsell/ai/blob/main/docs/features/session-isolation.md) |
| Tier-2 isolation | Slug-scoped session isolation: a named work item gets its own worktree, branch, plan doc, task list, and GitHub issue, all keyed by the slug. Established by `/do-plan` or `valor-session create --slug`. | [`docs/features/session-isolation.md`](https://github.com/tomcounsell/ai/blob/main/docs/features/session-isolation.md) |
| Stale-block failure mode | A condition where a lock holder has crashed, exited uncleanly, or been killed but its lock state remains, blocking all subsequent acquirers. Fazm's stated cautionary case. | [fazm CLAUDE.md "Multi-Agent Safety"](https://github.com/mediar-ai/fazm/blob/main/CLAUDE.md) |
| Fail-open | A lock or guard that, on uncertainty (corrupt state, unreachable detector, timeout), degrades to no-op rather than blocking. Liveness-preserving, correctness-weakening. | this issue |
| Fail-closed | A lock or guard that blocks when state is uncertain. Correctness-preserving, liveness-weakening. **Explicitly out of scope.** | this issue |

## Prior Context

Related work:
- #887: *Session isolation bypass: PM sessions created via valor-session create operate in main checkout instead of a worktree* — landed three enforcement layers in `agent/agent_session_queue.py` and a `--slug` flag in `valor-session create`. This issue explores complementary guards for the residual surface.
- #880: *Worktree path-containment guard* — `agent/worktree_manager.py::_cleanup_stale_worktree()` now refuses to operate on paths outside `.worktrees/`, after a 2026-04-10 incident where a misdirected `shutil.rmtree(..., ignore_errors=True)` recursively deleted the main repo. **Key precedent:** path-containment is the kind of always-on, fail-loud invariant that worked. Locks that require coordinated cooperation across processes are riskier.
- #875: *Original catastrophic repo wipe* — referenced in #887. The shared-state pattern that motivated the worktree enforcement layers.
- `feedback_verify_dirty_state_ownership.md` (memory) — current human-facing rule: agents must verify dirty-state ownership before running `pytest`/`ruff`. This is the rule we'd be partially codifying.
- `project_session_isolation_bypass.md` (memory) — the situational memory of #887.
- fazm reference: [`mediar-ai/fazm/CLAUDE.md`](https://github.com/mediar-ai/fazm/blob/main/CLAUDE.md) — cautionary inspiration. Their "never `rm -rf` the lock" rule is exactly the kind of operator burden we want to avoid.

## Solution Sketch

This is an **enumeration of alternatives**. `/do-plan` should pick (or reject) one based on the trade-offs. Each alternative is paired with its specific stress test for the no-side-effects constraint.

### Alternative A — Mandate worktree for every dev session (no lock at all)

Make filesystem isolation the only protocol. If `AgentSession.session_type == "dev"`, always provision a worktree, with or without a slug — synthesize a slug if needed (e.g. `dev-{agent_session_id[:8]}`).

- **Why this is not already the default:** the #887 fix only enforces worktree-when-slug-is-set. A dev session created without a slug (e.g. via a future harness or a manual `valor-session create --role dev` with no `--slug`) skips the guard.
- **Stress test:** what breaks if a future entry point invokes `valor-session create --role dev` with no slug? Today it would run in the project root; under Alternative A it would always get an auto-generated worktree. Cost: cleanup needs to handle ephemeral, possibly-orphaned worktrees. Liveness risk: worktree provisioning failure is currently fail-loud in `_execute_agent_session()` for dev sessions — this is fine, because failing to start a dev session ≠ wedging future sessions.
- **No-side-effects analysis:** worktree provisioning is already idempotent and well-tested. Failure to provision raises `RuntimeError` in the queuing path, which surfaces as a session failure but does not affect any other process. **This alternative has no global lock state, so by construction it cannot create a stale-block failure mode.**

### Alternative B — Fail-open advisory lock with hard max-wait

Fazm-style directory lock at e.g. `/tmp/valor-checkout-{repo_root_hash}.lock`, but with two changes from fazm:
1. Hard max-wait (e.g. 30s); on timeout, log and **proceed without the lock** (fail-open).
2. Lock is read by a wrapper around any operation that mutates the main checkout (e.g. test runs, ruff, post-merge scripts), but the wrapper never refuses to run.

- **Stress test:** if two sessions both fail-open through it, what's the worst case? Same as today — uncoordinated parallel writes. The lock has zero correctness value if both sides fail open simultaneously, which means it provides only soft serialization under non-contended load. Is that worth the operator burden of stale-lock cleanup?
- **No-side-effects analysis:** every fail-open path is a no-op equivalent to "no lock present." But the *acquisition* path can have side effects: writing `pid` files, creating directories, scanning `/proc` for liveness. Each of those can fail in surprising ways (filesystem full, permission errors after a python upgrade drops TCC grant — see `project_python_tcc_grant.md`). We need a precise inventory of every side effect of the acquisition path and a guarantee that each one fails open.
- **Open question:** does the acquisition path need to write to disk at all? A purely in-memory lock (e.g. Redis SET NX with TTL) has fewer failure modes — see Alternative B'.

### Alternative B' — Redis-based fail-open lock with TTL

Use Redis `SET NX EX 60` to acquire a lock at key `valor:checkout-lock:{repo_root_hash}`. TTL guarantees auto-release; no PID file, no `/proc` scan, no manual cleanup. On Redis unreachable: fail-open and proceed.

- **Stress test:** Redis is already a dependency (Popoto, queue, memory). If Redis is down, the worker is broken anyway, so the fail-open path is exercised only in narrow degraded-mode scenarios. TTL prevents stale blocks even in worst case.
- **No-side-effects analysis:** TTL-based locks have a known correctness issue (the famous Redlock debate) — under clock skew or GC pause the lock can be released while the holder still believes it owns it. For *advisory* locking against a single-machine main checkout, that's acceptable.
- **Caveat:** must be project-scoped per `feedback_test_redis_isolation.md` — the key must be derived from the repo root, not a global name.

### Alternative C — Lock-free optimistic concurrency detector

Every session that is about to write to the main checkout records `git rev-parse HEAD` and the dirty-file set on entry. Before writing, it re-reads both. If a second process has changed either, the session **errors out before writing** rather than overwriting. No global lock state.

- **Stress test:** what's the cost of a false positive (a single concurrent `git fetch` updating refs)? The session aborts and re-tries. Acceptable.
- **No-side-effects analysis:** the detector itself only reads. The only "side effect" is the abort path, which prevents work from being done — but does not wedge any other session. Critically, **there is no lock state to go stale.**
- **Coverage gap:** this catches concurrent mutation but doesn't prevent it from happening; it just stops the *second* writer from compounding the damage.

### Alternative D — Do nothing programmatic

Document `feedback_verify_dirty_state_ownership.md` more loudly. Add a Stop-hook that prints a stronger warning when an agent leaves the main checkout dirty. Keep the burden on the agent's own behavior.

- **Residual risk:** non-zero. The agent forgets, a human collides, a cron job lands at the wrong moment.
- **No-side-effects analysis:** by construction, doing nothing introduces nothing.
- **Why this is on the list:** the user has explicitly said "a stale block is crippling." If no alternative passes the no-side-effects bar, this is the correct answer.

### Alternative E — Split the resource: lock only the expensive operations

The contention isn't on every dev session operation — it's specifically on operations that *mutate the main checkout in ways another process would observe*: `pytest` and `ruff` runs, `git checkout`, post-merge cleanup. Lock just those, not the entire dev workflow.

- **Implementation shape:** a small wrapper command (e.g. `valor-checkout-exclusive <cmd...>`) that uses Alternative B' (Redis lock) for the duration of `<cmd>` and fails open if Redis is unreachable.
- **No-side-effects analysis:** the lock surface is minimized to just the operations that matter. The wrapper is opt-in — no caller is *required* to use it, so a failure of the wrapper never blocks legacy callers.
- **Compatibility with Alternative A:** these are complementary. A is about *which checkout* a session uses; E is about *serializing operations on the main checkout* by callers (humans, cron jobs, legacy code) who can't be moved into a worktree.

### Cross-cutting questions for whichever alternative `/do-plan` picks

1. **Interaction with launchd worker auto-respawn:** the worker has `KeepAlive=true`. If a lock-holding worker crashes and the auto-respawn picks up, does the new worker correctly observe its own prior crashed state? (Specifically motivates Alternative B': TTL guarantees yes.)
2. **Interaction with the existing `_cleanup_stale_worktree` guard (#880):** any lock-acquisition path must NOT call into worktree cleanup logic with anything but a path strictly under `.worktrees/`. The #880 guard catches this, but a buggy lock could still trigger spurious `logger.critical` events.
3. **Multi-machine ownership:** `projects.json` already enforces single-machine ownership of bridge contacts (see `docs/features/single-machine-ownership.md`). A checkout lock would be intra-machine only — multi-machine is not a concern here.
4. **Existing prior art in the repo:** none of `scripts/valor-service.sh` uses filesystem locks; the bridge self-healing system (`docs/features/bridge-self-healing.md`) uses Redis crash-tracker, not locks. We'd be introducing a new pattern, which is itself a reason to prefer the lightest viable alternative.
5. **Unit-of-locking granularity:** repo-root, slug, or operation? Alternative E argues operation; Alternatives B/B' implicitly argue repo-root.
6. **Failure-mode audit:** for the chosen alternative, can a buggy implementation leave the system *worse off* than no lock? Concrete adversarial scenario for each alternative is required before implementation.

## Recon Summary

**Confirmed:**
- Issue #887 enforcement layers are real and correct (verified in `agent/agent_session_queue.py`, `tools/valor_session.py`, `config/personas/project-manager.md`, `docs/features/session-isolation.md` § "Worktree Enforcement for Dev Sessions").
- No existing lock infrastructure in the repo: `scripts/valor-service.sh` has no `Lock`/`flock`/`mkdir`-as-lock primitives; `.claude/hooks/validators/` contains no checkout-state validators.
- fazm pattern reference is accurate: directory lock + PID file + stale detection + 5-min max-wait + explicit "never `rm -rf` the lock" CLAUDE.md rule.
- The #880 path-containment guard provides precedent: invariants enforced at the operation site (not via global state) are the safer pattern in this codebase.

**Revised:**
- Original framing was "add a lock"; revised framing is "enumerate alternatives, including 'no lock at all'." User explicitly framed this as design exploration.
- Title scoped from "add a checkout lock" to "guard against contamination — explore no-side-effects designs" to reflect open-ended scope.

**Pre-requisites:** None. #887 already shipped; this issue builds on top.

**Dropped:**
- Fail-closed designs are out of scope per user constraint.
- Multi-machine coordination is out of scope (single-machine ownership already enforced upstream).
- A specific implementation is out of scope; that is `/do-plan`'s job.

## Acceptance Criteria

This issue is a **design exploration**. Acceptance is at the plan stage, not the build stage.

- [ ] `/do-plan` produces a plan document at `docs/plans/parallel-session-checkout-guard.md` (or similar slug)
- [ ] The plan explicitly evaluates each of Alternatives A, B, B', C, D, E above against the no-side-effects constraint
- [ ] The plan picks one path forward (which may be Alternative D — "do nothing programmatic" — with explicit rationale)
- [ ] If the plan picks any alternative other than D, it includes an adversarial failure-mode audit demonstrating that a buggy implementation cannot wedge the system
- [ ] The plan addresses the six cross-cutting questions in the Solution Sketch
- [ ] If the plan picks Alternative B or B', it explicitly addresses the fazm "never `rm -rf` the lock" cautionary rule and explains why our design avoids that operator burden
- [ ] The plan's `## No-Gos` section explicitly forbids fail-closed locking and any operator manual-cleanup steps

## Downstream

This issue feeds into `/do-plan`, which will produce a plan document at `docs/plans/{slug}.md`. The plan author should treat the Solution Sketch above as a **menu of alternatives, not a directive**. The expected plan output is a comparative analysis with a recommendation, not an implementation spec for a pre-chosen design.

**Constraints the planner should be aware of:**
- Fail-open is mandatory; fail-closed is a non-starter.
- A stale block is crippling — any design must demonstrate that no buggy state can wedge future sessions.
- "Do nothing programmatic" (Alternative D) is a legitimate plan output if no alternative passes the no-side-effects bar.
- Must not introduce manual operator cleanup steps (no "if you see a stale lock, run `rm -rf /tmp/foo`" instructions in CLAUDE.md or anywhere else).

tomcounsell:
## Summary

Closes the residual hole left by #887 — slugless dev sessions could bypass the main-checkout protection guard and contaminate the main checkout. Adds two surgical, fail-open guards.

## Changes

- **CLI symmetry guard** (`tools/valor_session.py`): `valor-session create --role dev` now requires `--slug` or `issue #N` in the message, mirroring the existing PM check (#1109). The error message includes the literal substring `PM and dev sessions must be created with --slug` for grep-ability and references issues #1109 + #1272.
- **Synthetic-slug synthesis** (`agent/session_executor.py`): if a slugless dev session somehow reaches the executor (future programmatic spawn site that bypasses the CLI), synthesize `dev-{agent_session_id[:8]}` and provision a worktree the same way slugged sessions do today. The synthesis emits a stable `[synthetic-slug]` log marker for post-deploy reflection scans.
- **Pre-synthesis precondition**: extends the existing executor-guard (`agent/session_executor.py:641-693`) to also reject `_stype == "dev" and slug is None and agent_session_id is None`, preventing a `TypeError` crash on `aid[:8]`.
- **Synthetic-slug cleanup hook**: in the session-completion `finally` block, `cleanup_after_merge(repo_root, slug)` runs when slug matches `^dev-[0-9a-f]{8}$`. Cleanup failures log at WARNING and do NOT propagate as session failures. The exact regex protects human slugs like `dev-improvements` or `dev-1272`.

## Testing

- [x] Unit tests passing — 90 tests across `test_session_isolation_bypass.py` (+9 new), `test_valor_session_cli.py` (+3 new), `test_valor_session_working_dir_resolution.py` (6 updated), `test_pm_session_auto_slug.py` (1 updated), `test_pm_session_refuse_no_issue.py` (1 updated), `test_worktree_manager.py` (regression clean).
- [x] Full unit suite: 6132 passed, 1 skipped.
- [x] Linting (ruff format + check) passing.
- [x] Plan verification: 8 of 9 checks pass; 1 spurious failure caused by the verification table parser splitting on the `\` line continuation in the plan (the actual command runs correctly — `python -m tools.valor_session create --role dev --slug X --message "test" 2>&1 | head -1` outputs `Worktree: ...` with no error).

## Documentation

- [x] `docs/features/session-isolation.md`: new "Synthetic Slugs for Slugless Dev Sessions (Issue #1272)" subsection.
- [x] `tools/valor_session.py::cmd_create` docstring: documents the PM+dev slug requirement.

## Definition of Done

- [x] Tested: all tests passing
- [x] Documented: docs created/updated

Closes #1272

### DIFF
--- a/agent/session_executor.py
+++ b/agent/session_executor.py
@@ -4,6 +4,7 @@ nudge/re-enqueue paths, and calendar heartbeat."""
 import asyncio
 import logging
 import os  # noqa: F401
+import re
 import subprocess
 from datetime import UTC, datetime
 from pathlib import Path
@@ -638,6 +639,59 @@ async def _execute_agent_session(session: AgentSession) -> None:
         # any future spawn site that forgets a required field will fail loudly
         # here instead of mid-startup with no Telegram message. The session is
         # marked ``failed`` so dashboards / reflections surface it.
+        # Synthesis precondition (issue #1272): a slugless dev session needs
+        # ``agent_session_id`` to derive its synthetic slug ``dev-{aid[:8]}``.
+        # If both ``slug`` and ``agent_session_id`` are missing on a dev session
+        # the synthesis branch below would crash with
+        # ``TypeError: 'NoneType' object is not subscriptable``. Fail loudly
+        # here so the failure mode is observable instead of obscured.
+        _stype_pre = getattr(session, "session_type", None)
+        _slug_pre = getattr(session, "slug", None)
+        _aid_pre = getattr(session, "agent_session_id", None)
+        if _stype_pre == "dev" and _slug_pre is None and _aid_pre is None:
+            _parent = getattr(session, "parent_agent_session_id", None)
+            logger.error(
+                "[executor-guard] Refusing to start slugless dev session with None "
+                "agent_session_id (reason=missing_aid_for_synthetic_slug): "
+                f"slug={_slug_pre!r} session_type={_stype_pre} "
+                f"parent_agent_session_id={_parent} "
+                f"working_dir={session.working_dir!r} session_id={session.session_id!r}"
+            )
+            try:
+                from models.session_lifecycle import (  # noqa: PLC0415
+                    StatusConflictError,
+                    finalize_session,
+                )
+
+                finalize_session(
+                    session,
+                    "failed",
+                    reason=(
+                        "slugless dev session requires agent_session_id for "
+                        "synthetic slug derivation (issue #1272)"
+                    ),
+                )
+            except StatusConflictError as finalize_conflict:
+                logger.info(
+                    "[executor-guard] Skipping finalize for %s: %s",
+                    getattr(session, "agent_session_id", "?"),
+                    finalize_conflict,
+                )
+            except Exception as finalize_err:
+                logger.error(
+                    "[executor-guard] finalize_session(failed) raised: %s",
+                    finalize_err,
+                )
+                try:
+                    session.status = "failed"
+                    session.save(update_fields=["status", "updated_at"])
+                except Exception as last_resort_err:
+                    logger.debug(
+                        "[executor-guard] last-resort status save failed (non-fatal): %s",
+                        last_resort_err,
+                    )
+            return
+
         if session.working_dir is None or session.session_id is None:
             offending_field = "working_dir" if session.working_dir is None else "session_id"
             _aid = getattr(session, "agent_session_id", None) or getattr(session, "id", "?")
@@ -708,6 +762,33 @@ async def _execute_agent_session(session: AgentSession) -> None:
         # Resolve branch: use slug + stage mapping if available, else session-based
         slug = session.slug
         stage = None
+        # Synthetic-slug synthesis for slugless dev sessions (issue #1272,
+        # Alternative A from docs/plans/parallel-session-checkout-guard.md).
+        #
+        # The #887 main-checkout protection guard below short-circuits on
+        # ``slug is None``: ``_stype == "dev" and slug and ...``. That left a
+        # residual hole — a dev session created without a slug (future
+        # debug harness, test fixture, or any code path that bypasses the
+        # CLI) would skip worktree provisioning AND skip the guard, landing
+        # in the main checkout. Synthesizing ``dev-{aid[:8]}`` here funnels
+        # every dev session through the existing worktree-creation path so
+        # the guard always has a slug to enforce against. The
+        # ``agent_session_id`` precondition above (executor-guard) ensures
+        # ``aid`` is non-None before this line runs.
+        # Synthetic dev slug shape: dev-{first 8 chars of agent_session_id}.
+        is_synthetic_slug = False
+        if not slug and getattr(session, "session_type", None) == "dev":
+            _aid_for_slug = getattr(session, "agent_session_id", None)
+            if _aid_for_slug:
+                slug = f"dev-{_aid_for_slug[:8]}"
+                is_synthetic_slug = True
+                # Stable grep marker for post-deploy reflection scans —
+                # MUST be the literal token ``[synthetic-slug]`` so log
+                # audits can count occurrences without false positives.
+                logger.info(
+                    f"[synthetic-slug] Allocated synthetic slug {slug} "
+                    f"for slugless dev session {_aid_for_slug} (issue #1272)"
+                )
         if slug:
             # Try to read current stage from the AgentSession
             try:
@@ -724,6 +805,14 @@ async def _execute_agent_session(session: AgentSession) -> None:
             from agent.agent_session_queue import resolve_branch_for_stage  # noqa: PLC0415
 
             resolved_branch, needs_wt = resolve_branch_for_stage(slug, stage)
+            # Synthetic dev slugs (#1272) have no SDLC stage, so the
+            # default mapping returns ``("main", False)`` — which would
+            # bypass worktree provisioning AND trip the main-checkout
+            # guard below. Force the synthetic case onto a session branch
+            # in a worktree so isolation is guaranteed.
+            if is_synthetic_slug:
+                resolved_branch = f"session/{slug}"
+                needs_wt = True
             branch_name = resolved_branch
             # If branch resolution says we need a worktree and working_dir isn't one
             if needs_wt and WORKTREES_DIR not in str(working_dir):
@@ -1846,3 +1935,37 @@ async def _execute_agent_session(session: AgentSession) -> None:
         # into _active_sessions across sessions on the same worker.
         if _session_id_for_registry:
             _active_sessions.pop(_session_id_for_registry, None)
+
+        # === Synthetic-slug worktree cleanup (issue #1272) ===
+        # Slugless dev sessions get a synthesized slug ``dev-{aid[:8]}`` and a
+        # worktree provisioned for them above. Without an explicit cleanup
+        # hook, those worktrees linger forever — ``prune_worktrees()`` only
+        # runs ``git worktree prune`` (removes references, not directories)
+        # and ``cleanup_after_merge()`` is normally only triggered by a PR
+        # merge. Synthetic-slug sessions may never open a PR. Match the
+        # exact ``dev-{8 hex chars}`` shape so we never touch a real slug.
+        try:
+            _slug_for_cleanup = locals().get("slug")
+            if (
+                _slug_for_cleanup
+                and isinstance(_slug_for_cleanup, str)
+                and re.match(r"^dev-[0-9a-f]{8}$", _slug_for_cleanup)
+            ):
+                from agent.worktree_manager import (  # noqa: PLC0415
+                    cleanup_after_merge,
+                    resolve_repo_root,
+                )
+
+                _wd = locals().get("working_dir")
+                if _wd is not None:
+                    _repo_for_cleanup = resolve_repo_root(_wd)
+                    cleanup_result = cleanup_after_merge(_repo_for_cleanup, _slug_for_cleanup)
+                    logger.info(
+                        f"[synthetic-slug] Cleaned up worktree+branch for "
+                        f"{_slug_for_cleanup}: {cleanup_result}"
+                    )
+        except Exception as cleanup_err:
+            # Cleanup failures must NEVER propagate as session failures.
+            logger.warning(
+                f"[synthetic-slug] Cleanup failed for synthetic slug (non-fatal): {cleanup_err}"
+            )
--- a/agent/session_executor.py
+++ b/agent/session_executor.py
@@ -639,6 +639,7 @@ async def _execute_agent_session(session: AgentSession) -> None:
         # any future spawn site that forgets a required field will fail loudly
         # here instead of mid-startup with no Telegram message. The session is
         # marked ``failed`` so dashboards / reflections surface it.
+
         # Synthesis precondition (issue #1272): a slugless dev session needs
         # ``agent_session_id`` to derive its synthetic slug ``dev-{aid[:8]}``.
         # If both ``slug`` and ``agent_session_id`` are missing on a dev session
@@ -1953,12 +1954,12 @@ async def _execute_agent_session(session: AgentSession) -> None:
             ):
                 from agent.worktree_manager import (  # noqa: PLC0415
                     cleanup_after_merge,
-                    resolve_repo_root,
+                    resolve_main_repo_root,
                 )
 
                 _wd = locals().get("working_dir")
                 if _wd is not None:
-                    _repo_for_cleanup = resolve_repo_root(_wd)
+                    _repo_for_cleanup = resolve_main_repo_root(_wd)
                     cleanup_result = cleanup_after_merge(_repo_for_cleanup, _slug_for_cleanup)
                     logger.info(
                         f"[synthetic-slug] Cleaned up worktree+branch for "
--- a/agent/worktree_manager.py
+++ b/agent/worktree_manager.py
@@ -142,6 +142,70 @@ def resolve_repo_root(file_path: str | Path) -> Path:
     return Path(result.stdout.strip())
 
 
+def resolve_main_repo_root(file_path: str | Path) -> Path:
+    """Determine the main repository root for a file, even when inside a worktree.
+
+    Use this when you need the main repo root regardless of whether
+    ``file_path`` is in a worktree. From inside a linked worktree,
+    ``git rev-parse --show-toplevel`` (used by :func:`resolve_repo_root`)
+    returns the *worktree's* path — which is the wrong base when you need
+    to reach into ``.worktrees/{slug}/`` for cleanup. This helper instead
+    uses ``git rev-parse --git-common-dir``, which always points at the
+    main repo's ``.git`` directory regardless of which worktree you are in.
+    The parent of that ``.git`` directory is the main repo root.
+
+    Contrast with :func:`resolve_repo_root`, which returns the worktree
+    path when called from inside one. Prefer ``resolve_repo_root`` when
+    you intentionally want the local checkout (e.g. plan documents); use
+    this helper when you need to manipulate sibling worktrees.
+
+    Args:
+        file_path: Absolute or relative path to a file or directory.
+            If a directory is given, it is used directly. If a file is
+            given, its parent directory is used.
+
+    Returns:
+        Absolute Path to the main git repository root.
+
+    Raises:
+        FileNotFoundError: If the file or its parent directory does not exist.
+        ValueError: If the path is not inside any git repository.
+    """
+    path = Path(file_path).resolve()
+
+    # Use the directory containing the file, or the path itself if it's a dir
+    search_dir = path if path.is_dir() else path.parent
+
+    if not search_dir.exists():
+        raise FileNotFoundError(
+            f"Cannot resolve main repo root: directory {search_dir} does not exist"
+        )
+
+    result = subprocess.run(
+        ["git", "rev-parse", "--git-common-dir"],
+        cwd=search_dir,
+        capture_output=True,
+        text=True,
+        timeout=10,
+    )
+
+    if result.returncode != 0:
+        raise ValueError(
+            f"Path {file_path} is not inside a git repository. git error: {result.stderr.strip()}"
+        )
+
+    common_dir = Path(result.stdout.strip())
+    # ``--git-common-dir`` may return a relative path (relative to ``cwd``)
+    # or an absolute path depending on the git version and worktree layout.
+    # Normalize before taking ``.parent``.
+    if common_dir.is_absolute():
+        git_dir = common_dir.resolve()
+    else:
+        git_dir = (search_dir / common_dir).resolve()
+
+    return git_dir.parent
+
+
 def _validate_slug(slug: str) -> None:
     """Validate slug to prevent path traversal and invalid directory names.
 
--- a/docs/features/agent-session-model.md
+++ b/docs/features/agent-session-model.md
@@ -222,7 +222,7 @@ degradation rather than a hard error.
 ### How It Flows
 
 1. PM or human calls
-   `python -m tools.valor_session create --role dev --model sonnet --message "..."`.
+   `python -m tools.valor_session create --role dev --slug {slug} --model sonnet --message "..."`.
 2. `tools/valor_session.py::cmd_create` constructs the AgentSession via
    `enqueue_agent_session(model="sonnet")`; the value is persisted on the
    Redis record.
--- a/docs/features/harness-abstraction.md
+++ b/docs/features/harness-abstraction.md
@@ -188,10 +188,10 @@ All operations are wrapped in try/except — failures never crash the worker.
 The PM persona at `~/Desktop/Valor/personas/project-manager.md` now dispatches dev sessions via:
 
 ```bash
-python -m tools.valor_session create --role dev --parent "$AGENT_SESSION_ID" --message "..."
+python -m tools.valor_session create --role dev --slug {slug} --parent "$AGENT_SESSION_ID" --message "..."
 ```
 
-instead of `Agent(subagent_type="dev-session", ...)`. The Agent tool dispatch path for dev sessions has been removed.
+instead of `Agent(subagent_type="dev-session", ...)`. The Agent tool dispatch path for dev sessions has been removed. `--slug` is required (or `issue #N` in the message body for auto-derivation) so the worktree is provisioned at create time — see [Session Isolation](session-isolation.md#synthetic-slugs-for-slugless-dev-sessions-issue-1272) for the slugless-fallback safety net.
 
 `sdk_client.py` contains a startup validation in `load_persona_prompt()` that warns if the PM persona still contains Agent tool dispatch (backward-compat guard).
 
--- a/docs/features/pipeline-state-machine.md
+++ b/docs/features/pipeline-state-machine.md
@@ -27,7 +27,7 @@ State is persisted as a JSON dict on `AgentSession.stage_states` -- one Redis fi
 
 The state machine is wired into the worker post-completion handler and the SDK hook system. This is the end-to-end flow for a single SDLC stage:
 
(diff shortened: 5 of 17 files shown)


valorengels (changes requested):

The PR cleanly closes the slugless-dev hole by adding the CLI symmetry guard, the executor-guard precondition, the synthesis path with branch/worktree forcing, and a stable `[synthetic-slug]` log marker. The plan, the diff, and the tests line up well. **58 modified-suite tests pass; ruff clean.**

However, the synthetic-slug worktree cleanup hook (the answer to the Skeptic's CONCERN in the war-room critique and to plan Risk 2) is silently broken at runtime: `resolve_repo_root(working_dir)` called from inside the synthesized worktree returns the worktree path itself, not the main repo root. Verified empirically below. The hook will log "Cleaned up worktree+branch" while the worktree directory is never actually touched, and synthetic worktrees will accumulate exactly the way the plan promised they would not.

### Verified
- [x] Code correctness (CLI guard, executor precondition, synthesis branch/worktree forcing — all behave as planned)
- [x] Test coverage (58 tests pass; +9 new, +3 new, plus updated regressions)
- [x] Plan alignment (Alternative A implemented; CLI symmetry shipped; precondition shipped; log marker shipped)
- [x] Lint / format clean (`ruff check` → all clean)
- [x] No legacy cruft, no leftover debug code

### Tech Debt
- [ ] **`agent/session_executor.py:1961`** — `_repo_for_cleanup = resolve_repo_root(_wd)` — When the cleanup `finally` block fires after a synthesized dev session, `_wd` is the synthesized worktree path (set at line 823: `working_dir = Path(wt_path)`). `resolve_repo_root` runs `git rev-parse --show-toplevel`, which inside a git worktree returns the worktree path, **not the main repo path**. Empirically verified on this branch:
  ```
  >>> resolve_repo_root('/Users/.../ai/.worktrees/parallel-session-checkout-guard')
  PosixPath('/Users/.../ai/.worktrees/parallel-session-checkout-guard')
  ```
  `cleanup_after_merge` then computes `worktree_dir = repo_root / WORKTREES_DIR / slug` = `.worktrees/dev-XXXXXXXX/.worktrees/dev-XXXXXXXX/`, finds nothing, returns `already_clean=True`, and the synthesized worktree never gets removed. The log line claims success. This silently defeats the very mitigation the war-room critique added to address plan Risk 2.

  **Fix:** derive the main repo root via `git rev-parse --git-common-dir` (which returns `.../ai/.git` from inside a worktree, so `Path(common_dir).parent.resolve()` gives the main repo). The existing canonical caller `scripts/post_merge_cleanup.py` passes a top-level `REPO_ROOT` explicitly — same shape applies here. Alternatively, walk up from `_wd` until you find a parent whose name is not `.worktrees` and whose grandparent contains `.git/` as a directory (not a file).

- [ ] **`tests/unit/test_session_isolation_bypass.py:366`** — `test_synthetic_slug_worktree_pruneable` only asserts that `_validate_slug("dev-XXXXXXXX")` accepts the format. It does not exercise the cleanup `finally` hook end-to-end. That's why the bug above slipped past the suite. **Fix:** add a test that mocks `resolve_repo_root` and `cleanup_after_merge`, runs the `finally` block path with a `working_dir` under `.worktrees/dev-abcd1234/`, and asserts `cleanup_after_merge` was called with the **main repo root** (not the worktree path). For belt-and-suspenders, an integration-tier test that creates a real worktree with `git worktree add`, runs the cleanup line, and asserts the worktree directory is gone would have caught this immediately.

### Nits
- [ ] **`agent/session_executor.py:642`** — The new precondition block (lines 642-693) is appended to the `# Defense-in-depth guard (issue #1195)` comment header without a separator, so a reader scanning the file naturally reads both blocks as part of the #1195 fix. The new block is for #1272 and addresses a different invariant (slugless-dev synthesis precondition vs. missing-working-dir). **Fix:** insert a visual break — either a blank line followed by `# Synthesis precondition (issue #1272): ...` as its own comment header, or move the #1272 block to immediately before the synthesis site at line 779 with a comment cross-referencing the executor-guard pattern. The current ordering (new block runs first) is correct; this is purely a readability concern.

### Recon Notes
- Mergeability preflight: `state=OPEN`, `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN` (after one retry on `UNKNOWN`).
- PR head SHA: `641b12ba`. Reviewed against this exact tree.
- Plan reference: `docs/plans/parallel-session-checkout-guard.md` (Alternative A picked, with adversarial audit and four cross-cutting answers).
- No UI files changed → no screenshots needed.

valorengels (approved):
## Review: Approved

The follow-up commit `2aed2806` cleanly resolves both findings from the prior CHANGES_REQUESTED review:

1. **Cleanup hook now uses the main repo root.** A new helper `resolve_main_repo_root` (`agent/worktree_manager.py:145-206`) uses `git rev-parse --git-common-dir` instead of `--show-toplevel`, so from inside any worktree it returns the main repo path rather than the worktree path. The `finally` block at `agent/session_executor.py:1962` now imports and calls this helper, fixing the silent-no-op cleanup the prior review flagged. Empirically verified on this branch:
   ```
   resolve_repo_root('.')      → /Users/.../ai/.worktrees/parallel-session-checkout-guard
   resolve_main_repo_root('.') → /Users/.../ai
   ```

2. **End-to-end cleanup test added.** `test_resolve_main_repo_root_returns_main_repo_from_worktree` (`tests/unit/test_session_isolation_bypass.py:376-421`) creates a real git repo + linked worktree via `subprocess.run(["git", "worktree", "add", ...])` and asserts the helper returns the main path from inside the worktree. A second test `test_cleanup_hook_uses_main_repo_root_not_worktree` (lines 423-460) pins the executor's hook to the new helper so any future revert to `resolve_repo_root` fails the suite.

3. **Visual separator nit addressed** — the new precondition block at `agent/session_executor.py:643` has a clean blank-line break and its own "Synthesis precondition (issue #1272)" comment header, so a reader scanning the file no longer reads the #1272 block as part of the #1195 comment.

### Verified
- [x] Code correctness — synthesis math, branch/worktree forcing, executor-guard precondition, cleanup hook all behave as planned and as the prior review required
- [x] Test coverage — 60 modified-suite tests pass (+9 new synthesis tests, +3 new CLI tests, regressions adapted), 32 worktree_manager regressions clean
- [x] `resolve_main_repo_root` empirically returns the main repo path from inside a linked worktree
- [x] Cleanup regex `^dev-[0-9a-f]{8}$` is exact — `dev-improvements`, `dev-1272`, `dev-abcdef0g` all correctly rejected (covered by `test_synthetic_slug_does_not_match_real_slugs`)
- [x] Stable `[synthetic-slug]` log marker present at both synthesis site and cleanup site (pinned by `test_log_marker_is_stable_literal`)
- [x] Lint / format clean (`ruff check` → all clean on the three changed source files)
- [x] Plan alignment (Alternative A implemented; CLI symmetry shipped; executor-guard precondition shipped; synthetic-slug cleanup hook shipped per Skeptic's Implementation Note)
- [x] No UI files changed → no screenshots needed

### Verification Results
- [PASS] Synthesis tests pass — `pytest tests/unit/test_session_isolation_bypass.py -v` exit 0
- [PASS] Worktree regression clean — `pytest tests/unit/test_worktree_manager.py -v` exit 0
- [PASS] Format clean — `python -m ruff format --check .` exit 0
- [PASS] CLI rejects slugless dev — exit 1 with stderr containing `PM and dev sessions must be created with --slug`
- [PASS] #887 guard still active — grep finds the guard line
- [PASS] Synthetic slug pattern in code — grep finds the synthesis line
- 1 spurious failure in the table parser due to the trailing `\` line continuation in the plan (not a real failure — the actual command succeeds; PR body discloses this)

### Recon Notes
- Mergeability preflight: `state=OPEN`, `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN`.
- PR head SHA: `2aed2806`. Reviewed against the PR head fetched via `gh fetch origin pull/1280/head`.
- Plan reference: `docs/plans/parallel-session-checkout-guard.md` (Alternative A picked, with adversarial audit).
- No UI files changed → no screenshots needed.
- Author is `tomcounsell`, not the reviewer; using `--approve` (not the self-authored comment fallback).

valorengels:
## Review: Approved

Mirror of the formal PR review at https://github.com/tomcounsell/ai/pull/1280#pullrequestreview-4225838125 (posted via the Reviews API because the PR author differs from the reviewer; this issue-comment mirror exists to satisfy the merge-gate's structured review check).

The follow-up commit 2aed2806 cleanly resolves both findings from the prior CHANGES_REQUESTED review:

1. **Cleanup hook now uses the main repo root.** New helper resolve_main_repo_root in agent/worktree_manager.py uses git rev-parse --git-common-dir; the synthetic-slug cleanup hook in agent/session_executor.py imports/calls the new helper.
2. **Real end-to-end cleanup test.** test_resolve_main_repo_root_returns_main_repo_from_worktree creates a real git repo + linked worktree pair and pins the helper. test_cleanup_hook_uses_main_repo_root_not_worktree mocks the helper and asserts the executor passes the main repo root (not the worktree path) to cleanup_after_merge.

Tests: 26/26 in test_session_isolation_bypass.py pass; 32/32 worktree_manager regressions pass. Ruff clean.

valorengels:
## Merge gate: blocked on environmental contention

The follow-up commit `2aed2806` is APPROVED and all gates run cleanly except the Full Suite Gate, which is currently unrunnable from this worktree due to environmental contention. Posting status and exiting per the dev brief's "don't loop" guidance rather than killing other sessions' work.

### Gates

| Gate | Status | Notes |
|------|--------|-------|
| Review (formal API) | ✅ APPROVED | https://github.com/tomcounsell/ai/pull/1280#pullrequestreview-4225838125 |
| Review (issue-comment mirror) | ✅ APPROVED | Posted because formal review couldn't satisfy issue-comments regex |
| Lockfile sync | ✅ PASS | `uv lock --locked` clean |
| Targeted unit tests | ✅ PASS | `tests/unit/test_session_isolation_bypass.py`: 26/26; `tests/unit/test_worktree_manager.py`: 32/32 |
| Lint / format | ✅ PASS | ruff clean on all 3 modified files |
| Full Suite Gate | ⚠️ UNRUN | See below |

### What blocked the Full Suite Gate

The first run (`/tmp/pr_run.xml`) was killed at 8% with exit 144 (SIGURG-class signal). When I checked process state, four parallel pytest processes from different SDLC sessions were running concurrently in the same main checkout: `1279`, `1280`, `1282`, `sdlc-1271`. All four sat at 0% CPU and 7-13MB RSS — deadlocked on a shared resource, most likely Redis fixtures or the launchd worker bootstrap.

I killed only my own pytest (PID 15551). The other three are still running (8-21 min elapsed each); they belong to other sessions and are not mine to terminate.

I then ran `pytest tests/unit/` only (skipping the integration tier, which is where the contention lives). Result: **39 failed, 6115 passed, 32 skipped in 3:03**. None of the 39 failures touch the files this PR modifies (`agent/session_executor.py`, `agent/worktree_manager.py`, `tests/unit/test_session_isolation_bypass.py`). Failure breakdown:

- 9 × `FileNotFoundError: config/reflections.yaml` — missing symlink in this worktree, unrelated to this PR's code.
- ~28 × Redis-state-dependent assertions returning `[]` / `None` — consistent with parallel sessions mutating shared Redis keys mid-run.
- 2 × `test_refresh_test_baseline` / `test_send_message` — unrelated subsystems.

### What this PR's change actually affects

The fix is localised to one helper function (`resolve_main_repo_root` in `agent/worktree_manager.py`), one import + one-line call change inside an existing `try/except` in `agent/session_executor.py` that never propagates failures, and two new test cases. There is no plausible mechanism for these edits to break reflections-config loading, session-tag rules, token accumulators, the idle sweeper, or any of the other failed test files.

### Recommendation

A human (or a reviewer running merge gate from a non-contended environment) can re-run `/do-merge 1280` once the parallel pytest processes complete. The bootstrap-baseline path (`data/main_test_baseline.json` does not exist) would admit the current environmental failures as pre-existing on the next clean run.

Authorization file (`data/merge_authorized_1280`) was NOT created. Merge has NOT been attempted.

Issue #1218, opened 2026-04-30
session-liveness-check: reap orphan SDK subprocesses when DB row is already terminal


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/sdlc-1218.md

> **Context for readers unfamiliar with this codebase:** The worker process (`worker/__main__.py`) is the sole executor of `AgentSession` records. Each session spawns a `claude -p` CLI subprocess (the "SDK subprocess") and tracks it in an in-process map (`_active_sessions`). Liveness is enforced by the `session-liveness-check` reflection that runs every 5 minutes and calls `_agent_session_health_check()` in [`agent/session_health.py`](../blob/main/agent/session_health.py). PR #1039 / issue #1046 already shipped tier-1 stdout-staleness detection for the *forward* direction (DB row says running but subprocess is silent). This issue covers the *inverse* direction.

## Problem

`session-liveness-check` only reconciles state in one direction: it scans sessions whose **Redis row** says `status="running"` (or `pending`) and asks whether the worker/subprocess is still healthy. It never walks the **OS-side** map and asks the inverse: "for each subprocess this worker is still tracking, is the corresponding `AgentSession` row still in a non-terminal state?"

When the row reaches a terminal status before the subprocess exits — e.g. `finalize_session("completed")` returns successfully but the `claude -p` process never SIGTERMs and the messenger callback keeps pinging heartbeats — the orphan subprocess survives indefinitely. The terminal-status guard at [`session_health.py:631-638`](../blob/main/agent/session_health.py#L631) explicitly **skips** terminal sessions in the running scan, so the reflection never reaps them.

Because PM execution is single-threaded per `worker_key` ([`models/agent_session.py:297`](../blob/main/models/agent_session.py#L297)), one orphan subprocess blocks every subsequent PM session for that project from leaving `pending`.

### Concrete instance (2026-04-30)

| Field | Value |
|---|---|
| Stuck subprocess | PID `23031`, `claude -p` |
| Subprocess uptime | 15h 14m |
| Owning session | `0_1777474086119_cont1` |
| Session row status | `status=completed` (terminal) in Redis |
| Messenger | Logged `SDK heartbeat: running 54720s, communicated=False` every minute for 15h |
| Blocked downstream | `0_1777530058735` (PM session for issue #1215) stuck at `pending` |
| Resolution | Manual `kill 23031` — queue drained immediately |

[...]

PR #1284, opened 2026-05-05
feat(orphan-reap): cross-process orphan reaper with worker self-suicide guard (#1271)

### LINKED ISSUE #1271: Cleanup reflection: reap orphan claude/worker/MCP processes whose parent died
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/sdlc-1271.md

> **Context:** This system runs a process tree per active dev session — `bridge/telegram_bridge.py` (or `worker/__main__.py`) spawns the Claude Code CLI via [`agent/sdk_client.py`](agent/sdk_client.py), which in turn spawns 4+ MCP server subprocesses. We already have an [`agent-session-cleanup` reflection](config/reflections.yaml#L37) (registered at [`agent/session_health.py:1607`](agent/session_health.py) as `cleanup_corrupted_agent_sessions`) that reaps **DB rows** — corrupted `AgentSession` records that fail validation. It does not look at the OS process table at all. This issue extends that reflection to also reap **OS processes** whose parent died — `claude` CLI / worker / MCP-server subprocesses that were re-parented to PID 1 (launchd / init) when their parent crashed without graceful shutdown.

## Problem

When the bridge or worker dies ungracefully (panic, force-kill, hung SIGKILL, restart-without-graceful-shutdown), the child `claude` CLI process and its grandchildren (4+ MCP server subprocesses) become orphans: their PPID flips to 1 and they keep running indefinitely. They hold file handles, consume RAM/CPU, and — for `claude` specifically — keep an Anthropic API session warm long after the parent that knew about it has gone.

The [fazm project hit the same failure mode](https://github.com/mediar-ai/fazm/blob/main/acp-bridge/src/index.ts) in their ACP bridge. Their watchdog source comment:

> "Root cause of the 20+ orphan ACP bridges observed Apr 30 2026."

Their fix is per-subprocess: each spawned process records `process.ppid` at startup, polls every 5s, and SIGTERMs its descendant tree if its PPID flips to 1. That works but distributes the watchdog into every subprocess — it requires modifying every spawn path and trusting the orphan to clean itself up.

**Current behavior:**
- `cleanup_corrupted_agent_sessions` only inspects the Redis `AgentSession` index. It never calls `ps`, never reads `/proc`, never matches `AgentSession.id` against a live PID.
- `AgentSession` does not persist any OS PID field — only `last_heartbeat_at` ([`models/agent_session.py:274`](models/agent_session.py)) and an internal `parent_agent_session_id` (the *session* parent, not the *process* parent).
- After a worker crash, orphan `claude` + MCP server processes can persist for hours/days. The only signal they leave is RAM pressure — eventually visible via the resource thresholds in `CLAUDE.md` (Memory warning 600MB / critical 800MB) but never attributed to a specific cause.
- We have no metric tracking how many orphans accrue between cleanup ticks.

**Desired outcome:**
A single centralized reaper (the cleanup reflection, run hourly per [`config/reflections.yaml`](config/reflections.yaml)) detects orphan `claude` / worker / MCP subprocesses, terminates them with their full descendant tree, and emits a finding count surfaced in dashboard/analytics. No per-subprocess watchdog code. Detection is gated by AgentSession liveness signals (e.g. `last_heartbeat_at`) so a slow-but-alive `claude` run is never reaped.

## Definitions

| `agent-session-cleanup` reflection | Hourly reflection that deletes corrupted `AgentSession` Redis records | [`config/reflections.yaml`](config/reflections.yaml) entry, [`agent/session_health.py:1607`](agent/session_health.py) `cleanup_corrupted_agent_sessions` |
| `AgentSession` | Popoto-managed Redis model representing a single agent run (PM, dev, or teammate) | [`models/agent_session.py`](models/agent_session.py) |
| PPID flip | When a parent process dies, the kernel re-parents its children to PID 1 (launchd on macOS / init on Linux). `getppid()` then returns 1. | POSIX init re-parenting |
| `bridge_watchdog` | launchd-supervised process that monitors the Telegram bridge for liveness, restarts on crash | [`monitoring/bridge_watchdog.py`](monitoring/bridge_watchdog.py), [`docs/features/bridge-self-healing.md`](docs/features/bridge-self-healing.md) |
| MCP server subprocesses | Per-session helper processes (memory, sentry, gmail, calendar, notion, etc.) spawned as children of `claude` CLI | [`mcp_servers/`](mcp_servers/) |
| `_heal_descriptor_pollution` | Phantom-record cleaner already inside `cleanup_corrupted_agent_sessions` that filters Popoto descriptor leakage from `query.all()` | [`agent/session_health.py:1607`](agent/session_health.py) docstring |
| fazm `__watchdogStartPpid` | Per-subprocess parent-death watchdog pattern — records PPID at startup, exits if it flips to 1 | [`acp-bridge/src/index.ts`](https://github.com/mediar-ai/fazm/blob/main/acp-bridge/src/index.ts) |

## Prior Context

Related work:
- #1246: "session-liveness-check — tier-1 stdout-staleness reaper missed 11h-stuck SDK subprocess" — confirms the existing health-check loop does not catch every stuck-process case (different failure mode but same axis: subprocess outliving its logical session).
- fazm `acp-bridge/src/index.ts` watchdog (external): "Root cause of the 20+ orphan ACP bridges observed Apr 30 2026" — concrete failure-mode evidence that a parent-death gap accumulates dozens of orphans within weeks.

External reference: [fazm `__killBridgeTreeAndExit`](https://github.com/mediar-ai/fazm/blob/main/acp-bridge/src/index.ts) — search the file for `__watchdogStartPpid`.

## Solution Sketch

Extend the `agent-session-cleanup` reflection (or split into a sibling reflection callable in [`agent/session_health.py`](agent/session_health.py) that runs from the same schedule entry) so each tick:

1. Walks the local OS process table (`psutil.process_iter` or `ps -eo pid,ppid,command`).
2. Identifies "session-owned" processes by matching command-line patterns (`claude`, `python -m worker`, MCP server entrypoints under `mcp_servers/`) plus a registered-PID lookup if/when AgentSession persists `claude_pid`.
3. Decides which are orphans (PPID==1 and process started after the worker started, **and** the corresponding AgentSession has no recent `last_heartbeat_at`).
4. Terminates the orphan tree (SIGTERM, then SIGKILL after grace period) using `psutil.Process.children(recursive=True)`.
5. Records the count as a finding and emits an analytics metric.

This is centralized — no per-subprocess code edits — but should leave the door open for adding fazm-style per-subprocess PPID watchdogs later as defense-in-depth (the user's preferred direction is the centralized reaper first).

**Open architecture questions for `/do-plan`:**

1. **What counts as an orphan?** Three candidate scopes: (a) `claude` CLI processes whose worker parent is dead; (b) `worker` processes whose bridge parent is dead; (c) MCP server subprocesses whose `claude` parent is dead. Should the first cut cover all three, or start with `claude` (highest blast radius — it spawns the MCP children) and expand?
2. **Detection mechanism.** Process-table scan matched against AgentSession command-line signatures, OR add a `claude_pid` field to AgentSession written at spawn time in [`agent/sdk_client.py:2482`](agent/sdk_client.py) (the `on_sdk_started(pid)` hook already exists), OR a registered-PID file like fazm's `/tmp/fazm-dev-status`? Each has different recovery characteristics if the registry itself is stale.
3. **Centralized reaper vs per-subprocess watchdog.** The user picked centralized — folded into the existing cleanup reflection. Document the trade-off in the plan: centralized reaper has hourly latency (an orphan can live up to 60 minutes); per-subprocess watchdog reacts in 5s but distributes complexity across every spawn path. Is fazm-style worth keeping as a future defense-in-depth layer, or is the hourly cadence sufficient given that the dominant cost is RAM/handle accumulation, not the API session itself?
4. **Tree walking.** When reaping a `claude` process, do we always walk and SIGTERM its full descendant tree (MCP servers + any tool subprocesses they spawn)? Or trust each parent's own SIGTERM cascade? The fazm code walks the descendants explicitly — recommend matching that for determinism.
5. **Logging/metrics.** Count orphans killed per tick and surface in `localhost:8500/dashboard.json` and the analytics rollup. fazm's "20+ orphans by Apr 30" line is exactly the kind of trend signal that dashboard exposure would have caught earlier — this is a reason to instrument from day one.
6. **Interaction with existing systems.** The reflection runs alongside (a) `_heal_descriptor_pollution` inside the same callable — both safe to keep, different domains; (b) [`monitoring/bridge_watchdog.py`](monitoring/bridge_watchdog.py) which restarts the bridge process — does the watchdog already attempt orphan reaping when it detects a crash? Need to confirm there's no double-kill path. (c) launchd `KeepAlive=true` on the worker — when launchd auto-respawns the worker, the new worker has a fresh PID; the centralized reaper must not kill the *new* worker thinking it's an orphan.
7. **Safety: don't reap a slow but alive process.** The gate is AgentSession `last_heartbeat_at` ([`models/agent_session.py:274`](models/agent_session.py)) plus a generous threshold (e.g. 30 min) — only reap if both PPID==1 *and* heartbeat is older than threshold *and* no AgentSession is in a non-terminal state pointing at the matched PID. Triple-gate to avoid #1246-style false negatives flipping into false positives.

## Recon Summary

- `cleanup_corrupted_agent_sessions` exists at [`agent/session_health.py:1607`](agent/session_health.py); it operates only on Redis records, no OS process awareness.
- Registered as `agent-session-cleanup` reflection in [`config/reflections.yaml`](config/reflections.yaml) on a 1-hour interval.
- `AgentSession` model persists `last_heartbeat_at` ([`models/agent_session.py:274`](models/agent_session.py)) — a usable liveness gate. It does NOT persist a child OS PID.
- `agent/sdk_client.py:2482` already exposes an `on_sdk_started(pid)` callback after spawning the `claude` subprocess — natural injection point if we choose to persist `claude_pid` on AgentSession.

- Original framing was "extend with PPID-flip detection"; recon shows the cleanest extension is to add a sibling reaper function in [`agent/session_health.py`](agent/session_health.py) wired into the same `agent-session-cleanup` reflection name (or a new entry in the same YAML group) — keeps the existing corrupted-record logic untouched while sharing the scheduler entry's docstring lineage.

- Decide whether `AgentSession.claude_pid` (and possibly `worker_pid`) is added to the model. If yes, the [`feedback_field_backcompat_heal.md` memory](https://github.com/) note applies: nullable additions are healed automatically by `_heal_descriptor_pollution`, no separate migration needed.

- Per-subprocess fazm-style PPID watchdog. The user explicitly directed toward the centralized reflection; per-subprocess watchdog can be revisited as a future defense-in-depth layer if the hourly cadence proves insufficient.

## Acceptance Criteria

- [ ] The `agent-session-cleanup` reflection (or a new sibling registered in the same `config/reflections.yaml` group) detects orphan processes whose PPID == 1 and that match known signatures (`claude`, `python -m worker`, `mcp_servers/*`).
- [ ] Detection is gated by `AgentSession.last_heartbeat_at` (or equivalent liveness signal) — a long-running but alive `claude` run with a recent heartbeat is never reaped.
- [ ] When a parent process is reaped, its full descendant tree is also SIGTERMed (then SIGKILLed after grace) using `psutil.Process.children(recursive=True)`.
- [ ] Each reflection run reports the orphan-kill count in its `findings` list and to `localhost:8500/dashboard.json` (or the analytics rollup) for trend visibility.
- [ ] No interaction conflict with `monitoring/bridge_watchdog.py` or launchd `KeepAlive` — confirmed by integration test that triggers a worker crash, lets launchd respawn it, and verifies the reflection does not kill the fresh worker.
- [ ] No regression in the existing `cleanup_corrupted_agent_sessions` behavior — the corrupt-record reaping and `repair_indexes()` step still run.
- [ ] Plan document references the [fazm `__watchdogStartPpid` pattern](https://github.com/mediar-ai/fazm/blob/main/acp-bridge/src/index.ts) and the "20+ orphan ACP bridges Apr 30 2026" failure mode as motivating prior art.

## Downstream

Constraint: implementation must remain centralized in the reflection — do **not** add per-subprocess in-process PPID watchdogs to `bridge/`, `worker/`, `agent/sdk_client.py`, or any MCP server. The user explicitly chose the centralized reaper as the architecture; per-subprocess watchdogs are out of scope for this issue.

tomcounsell:
## Summary

Closes #1271. Adds a centralized cross-process orphan reaper that detects orphan `claude` and MCP-server processes whose `PPID == 1` AND whose owning `AgentSession` has no fresh heartbeat. Runs hourly via the existing `agent-session-cleanup` reflection.

**Critical bug fix**: prevents the worker from killing itself when reaping orphan processes. On macOS, every launchd-respawned worker has `PPID == 1` by design (`KeepAlive=true`), so a naive cmdline + PPID==1 filter would self-suicide every reflection tick. The fix uses **positive-ID self-protection**: the worker writes its PID to `worker:registered_pid:{hostname}:{pid}` (TTL 24h, refreshed on every heartbeat) and the reaper builds a `skip_pids` set from those keys + `os.getpid()` before scanning. Even if a future regex change re-adds the worker pattern, the worker is structurally safe.

## Key elements

- **`claude_pid` IndexedField + `find_by_claude_pid()`** on AgentSession — set on `_on_sdk_started`, cleared on terminal transitions in `models/session_lifecycle.py::finalize_session`. Enables per-PID heartbeat gating (30-min grace).
- **Two-tick SIGKILL escalation with create-time verification** (`_pending_sigkill_orphans: set[tuple[int, float]]`): hardens against macOS PID recycling within the 1-hour reflection window.
- **Descendant-tree walk**: captures `proc.children(recursive=True)` BEFORE killing parent so MCP grandchildren are also reaped.
- **Two-counter scheme**: `{project_key}:session-health:orphan_process_reaped` when owning session is known, hostname-scoped otherwise (avoids false project attribution).
- **Kill switch**: `DISABLE_ORPHAN_PROCESS_REAP=1`.
- **Return shape**: `cleanup_corrupted_agent_sessions()` now returns `{"corrupted": int, "orphans": int}`. All callers updated; `_cleanup_orphaned_claude_processes()` preserved as a thin shim so worker startup wiring stays intact.
- **Worker process reaping is intentionally OUT OF SCOPE** (see plan rationale).

Plan: `docs/plans/sdlc-1271.md` (cycle-2 critique: READY-with-concerns).

## Test plan

- [x] `pytest tests/unit/test_session_health_orphan_process_reap.py -v` (14 new tests: 9 scenarios + 3 invariants + 2 sub-tests)
- [x] `pytest tests/unit/test_session_health_phantom_guard.py -v` (updated for dict return shape)
- [x] `pytest tests/unit/test_session_health_orphan_reap.py tests/unit/test_worker_entry.py tests/unit/test_agent_session_lookup.py -v` (75 passing)
- [x] `python -m ruff format --check .` clean
- [x] `python -m ruff check agent/session_health.py models/agent_session.py models/session_lifecycle.py agent/session_executor.py worker/__main__.py agent/agent_session_queue.py scripts/update/run.py` clean
- [x] Verified `_cleanup_orphaned_claude_processes` shim still callable from `worker/__main__.py` startup
- [x] Verified `cleanup_corrupted_agent_sessions()` returns `{"corrupted": int, "orphans": int}`
- [x] Documentation updated: `docs/features/agent-session-health-monitor.md` "Cross-Process Orphan Reap (issue #1271)" section


valorengels:
## Review: Approved

Comprehensive cross-process orphan reaper with strong self-suicide protection. Implementation matches the plan in `docs/plans/sdlc-1271.md` with no deviations and addresses all 2 blockers + 4 concerns + 2 nits from the cycle-2 critique.

### Verified
- [x] Plan alignment — every success criterion in `docs/plans/sdlc-1271.md` is satisfied (claude_pid field, find_by_claude_pid helper, register_worker_pid + heartbeat refresh, _reap_orphan_session_processes implementation, two-counter scheme, kill switch, dict return shape, descendant tree walk, two-tick SIGKILL with create-time verification, shim preservation).
- [x] Code correctness — drain logic uses `abs(current_ct - staged_create_time) > 1e-3` epsilon comparison; per-PID gate correctly distinguishes terminal/null/stale heartbeats from live ones; MCP fallback to `proc.parent().pid` is wrapped in try/except.
- [x] Self-suicide protection — two layers as designed: (1) cmdline regex set excludes `python -m worker`, (2) positive-ID skip-set built from `os.getpid()` + `worker:registered_pid:*` Redis keys. Worker registers its PID at startup (worker/__main__.py:197) and refreshes on every heartbeat tick (session_health.py:1597).
- [x] Backward compatibility — `_cleanup_orphaned_claude_processes()` preserved as a shim returning the new reaper's int count; `cleanup_corrupted_agent_sessions()` returns dict with full caller updates in `worker/__main__.py` (with defensive `isinstance(result, dict)` legacy fallback) and `scripts/update/run.py`. `models/session_lifecycle.py::finalize_session` clears `claude_pid` on terminal transitions, wrapped in try/except for older records.
- [x] Test coverage — 14 new tests in `tests/unit/test_session_health_orphan_process_reap.py` (9 scenarios + 5 invariant variants) all pass; `test_session_health_phantom_guard.py` correctly updated for dict return shape; 61 regression tests across 4 test files pass.
- [x] Security — no secrets, no shell injection (subprocess module dropped entirely in favor of psutil); psutil's PID-reuse check via `_create_time` adds a third layer of guard against macOS PID recycling.
- [x] Error handling — every external call (psutil, Redis, model lookup) wrapped; reaper failure is swallowed in `cleanup_corrupted_agent_sessions()` and reported as `orphans=0` rather than aborting corrupted-record cleanup.
- [x] Lint and format — `ruff check` and `ruff format --check` clean on all modified files.
- [x] Documentation — `docs/features/agent-session-health-monitor.md` adds a "Cross-Process Orphan Reap (issue #1271)" subsection mirroring the structure of the existing #1218 section, including the distinction from the in-process reaper and the bridge_watchdog zombie killer.

### Notes (advisory, not blockers)

- **Startup double-reap is harmless but logs are slightly misleading.** At worker boot, Step 2 (`cleanup_corrupted_agent_sessions`) calls `_reap_orphan_session_processes()` which terminates orphans and stages SIGKILL tuples. Step 4 (`_cleanup_orphaned_claude_processes` shim) calls the reaper again, draining those staged tuples (so SIGTERM→SIGKILL happens within a few hundred milliseconds at startup, faster than the normal 1-hour reflection cadence). The Step 4 count log says "Killed N orphaned" but really it's draining the stage from Step 2 plus picking up any newly-discovered orphans. Functionally correct, just worth flagging if startup logs ever look surprising.
- **PID reuse across the same `claude_pid` index entry.** `find_by_claude_pid()` returns the first hit and logs a warning when multiple records match. In steady state, terminal-state cleanup in `finalize_session` clears the field, so collisions should be rare. The warning log is sufficient for observability.

valorengels (approved):
## Review: Approved

Comprehensive cross-process orphan reaper with strong self-suicide protection. Implementation matches the plan in `docs/plans/sdlc-1271.md` with no deviations and addresses all 2 blockers + 4 concerns + 2 nits from the cycle-2 critique.

### Verified
- [x] Plan alignment — every success criterion in `docs/plans/sdlc-1271.md` is satisfied (claude_pid field, find_by_claude_pid helper, register_worker_pid + heartbeat refresh, _reap_orphan_session_processes implementation, two-counter scheme, kill switch, dict return shape, descendant tree walk, two-tick SIGKILL with create-time verification, shim preservation).
- [x] Code correctness — drain logic uses `abs(current_ct - staged_create_time) > 1e-3` epsilon comparison; per-PID gate correctly distinguishes terminal/null/stale heartbeats from live ones; MCP fallback to `proc.parent().pid` is wrapped in try/except.
- [x] Self-suicide protection — two layers as designed: (1) cmdline regex set excludes `python -m worker`, (2) positive-ID skip-set built from `os.getpid()` + `worker:registered_pid:*` Redis keys. Worker registers its PID at startup (worker/__main__.py:197) and refreshes on every heartbeat tick (session_health.py:1597).
- [x] Backward compatibility — `_cleanup_orphaned_claude_processes()` preserved as a shim returning the new reaper's int count; `cleanup_corrupted_agent_sessions()` returns dict with full caller updates in `worker/__main__.py` (with defensive `isinstance(result, dict)` legacy fallback) and `scripts/update/run.py`. `models/session_lifecycle.py::finalize_session` clears `claude_pid` on terminal transitions, wrapped in try/except for older records.
- [x] Test coverage — 14 new tests in `tests/unit/test_session_health_orphan_process_reap.py` (9 scenarios + 5 invariant variants) all pass; `test_session_health_phantom_guard.py` correctly updated for dict return shape; 61 regression tests across 4 test files pass.
- [x] Security — no secrets, no shell injection (subprocess module dropped entirely in favor of psutil); psutil's PID-reuse check via `_create_time` adds a third layer of guard against macOS PID recycling.
- [x] Error handling — every external call (psutil, Redis, model lookup) wrapped; reaper failure is swallowed in `cleanup_corrupted_agent_sessions()` and reported as `orphans=0` rather than aborting corrupted-record cleanup.
- [x] Lint and format — `ruff check` and `ruff format --check` clean on all modified files.
- [x] Documentation — `docs/features/agent-session-health-monitor.md` adds a "Cross-Process Orphan Reap (issue #1271)" subsection mirroring the structure of the existing #1218 section, including the distinction from the in-process reaper and the bridge_watchdog zombie killer.

### Notes (advisory, not blockers)

- **Startup double-reap is harmless but logs are slightly misleading.** At worker boot, Step 2 (`cleanup_corrupted_agent_sessions`) calls `_reap_orphan_session_processes()` which terminates orphans and stages SIGKILL tuples. Step 4 (`_cleanup_orphaned_claude_processes` shim) calls the reaper again, draining those staged tuples (so SIGTERM→SIGKILL happens within a few hundred milliseconds at startup, faster than the normal 1-hour reflection cadence). The Step 4 count log says "Killed N orphaned" but really it's draining the stage from Step 2 plus picking up any newly-discovered orphans. Functionally correct, just worth flagging if startup logs ever look surprising.
- **PID reuse across the same `claude_pid` index entry.** `find_by_claude_pid()` returns the first hit and logs a warning when multiple records match. In steady state, terminal-state cleanup in `finalize_session` clears the field, so collisions should be rare. The warning log is sufficient for observability.

PR #1353, opened 2026-05-08
fix(worker): call validate_agent_files() at startup

tomcounsell:
## Summary

Hotfix from the daily integration audit run on **2026-05-09** (feature: `agent-definition-fallback`).

`validate_agent_files()` was called only at **bridge** startup, even though the worker is the *actual* session execution engine (per CLAUDE.md: "sole session execution engine"). Worker-only deployments — or boots where the worker comes up before the bridge — silently lost the early-warning hook the doc promises. The graceful fallback still works at runtime; this is purely an observability gap.

## Changes

- **`worker/__main__.py`** — call `validate_agent_files()` early in `main()` (after logging is configured, before the run loop). Mirrors the existing bridge call site.
- **`docs/features/agent-definition-fallback.md`** — rename the *Bridge startup validation* bullet to *Startup validation* and list both call sites (bridge + worker). Add the worker file to the Key Files table.
- **`tests/unit/test_worker_startup_validation.py`** (new) — three assertions guarding against regression to the bridge-only call site:
  1. Source-grep on `worker.__main__.main` references `validate_agent_files`.
  2. File-grep on `worker/__main__.py` references `validate_agent_files`.
  3. Stubbed `main()` run (mocking `_parse_args`, `_configure_logging`, `_load_projects`) confirms the validator is invoked before `_load_projects`.

## Verification

```
$ pytest tests/unit/test_worker_startup_validation.py tests/unit/test_agent_definitions.py
13 passed in 50.53s
```

## Audit reference

> The standalone worker is the 'sole session execution engine' — it is the process that actually consumes `get_agent_definitions()`. So a deployment where the worker starts but the bridge does not (or where the worker comes up first) misses the early-warning signal entirely. The graceful fallback still works, but the operator-facing observability the doc promises only fires on bridge startup.

### DIFF
--- a/docs/features/agent-definition-fallback.md
+++ b/docs/features/agent-definition-fallback.md
@@ -12,7 +12,9 @@ This happens during deployment windows when code is updated but agent files have
 
 - **Missing file detection**: `_parse_agent_markdown()` checks `path.exists()` before reading. If the file is missing, it logs a warning and returns a fallback dict with a minimal prompt instead of raising.
 - **Session continues**: `get_agent_definitions()` returns a complete dict even when some or all agent files are missing. The agent operates with degraded prompts rather than crashing.
-- **Bridge startup validation**: `validate_agent_files()` is called during bridge initialization to surface missing files early via log warnings, giving operators a chance to fix the issue before users hit it.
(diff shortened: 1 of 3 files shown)


tomcounsell:
## Review: Approved

Reviewer subagent verdict: APPROVED. Mirrors bridge-side prior art; warn-and-continue failure mode is correct; 3 regression-guard tests added (static-import, on-disk grep, behavioural). No blockers.

Issue #1325, opened 2026-05-08
Audit: re-check punted tasks across recent completed plans


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/sdlc-1325.md

## Context

A pattern audit (this conversation) found that recent SDLC plans frequently tag work as "operator step", "follow-up issue", "out of scope", or "deferred to v2" when the agent could in fact have completed the work in-plan. Categories ranked by frequency:

1. **"Follow-up issue / follow-up PR"** for in-repo cleanup, deletions, schema migrations, flag removals
2. **"Out of scope / v2"** for adjacent features sharing the touched code
3. **"Operator/human will run on each machine"** dodging the existing `/update` skill
4. **"Tracking-only"** with no issue actually filed
5. **Genuine externals** — minority

A new validator (`validate_no_gos_justification.py`) and template/skill updates are landing in a separate change to prevent future punts. This issue covers the *backward-looking* sweep: actually go finish (or formally close out) the punted items already on disk.

## Suspected punts to verify

For each, confirm: still pending? agent-doable? if so, do it; if not, file as proper tracked issue with [EXTERNAL]/[ORDERED]/[DESTRUCTIVE] justification.

- [ ] **\`docs/plans/completed/email-bridge.md:348\`** — launchd plist (\`com.valor.email-bridge.plist\`) and install script never written. Agent writes plists routinely.
- [ ] **\`docs/plans/completed/message-drafter.md:827-829\`** — \`AgentSession.delivery_text\` (and siblings) field deletions punted as "Tom owns migrations". Verify whether this is a real ownership boundary or a deference rule.
- [ ] **\`docs/plans/completed/pm-workflow-announcement.md:361\`** — multi-machine overlay edits left to operator. \`/update\` skill should automate this.
- [ ] **\`docs/plans/completed/pm-workflow-announcement.md:334\`** — tracking-only promise, no issue filed. Locate and convert to real ticket or close.
- [ ] **\`docs/plans/completed/kill-is-terminal.md:207\`** — root-cause investigation for index-staleness deferred behind the operational mask in Fix B. Verify masking is still adequate or do the diagnosis.
- [ ] **PR #1281 follow-ups** — deletions of \`reflections/daily_report.py\`, \`run_log_review\`, \`reflections.yaml\` rename, disabling stale entries. All in-repo, agent-doable.

## Sweep methodology

\`\`\`bash
[...]

Issue #1322, opened 2026-05-08
Media download timeout is hard-coded at 10s — large files fail silently


valorengels:
> **Context**: PR #1316 (issue #1297) introduced bridge-side media download with a hard-coded 10-second timeout. This was the right first step to split download (bridge) from AI enrichment (worker), but the 10s cutoff is blind to file size. Large PDFs, videos, or multi-image batches routinely exceed this, especially on slower connections.
>
> Related: #1297 (parent), #1316 (PR implementing first-pass timeout)

## Problem

The bridge's media download at `bridge/telegram_bridge.py:1155` uses a fixed `asyncio.wait_for(timeout=10.0)` regardless of file size or download speed:

```python
_local_path = await asyncio.wait_for(
    download_media(client, message, prefix=_media_type_for_download),
    timeout=10.0,
)
```

**Current behavior:**
- A 20MB PDF on a 2 Mbps connection needs ~80s; the download is killed at 10s with `media_download_error="timeout after 10s"`.
- A voice message (100KB) finishes in <1s; still carries the same 10s overhead.
- The user gets no retry — the message reaches the agent with no media description.

**Desired outcome:**
- Timeout scales with expected file size (e.g., 5s baseline + 1s per MB, capped at 120s).
- Or, better: a progress-based timeout that resets while bytes are still flowing.
- On timeout, the bridge should retry at least once with a longer leash before giving up.

## Definitions

| Term | Definition | Reference |
|------|-----------|-----------|
| `download_media` | Telethon helper that downloads a `MessageMediaDocument` or `MessageMediaPhoto` to local disk. | `bridge/media.py` |
| `asyncio.wait_for` | Python stdlib wrapper that raises `TimeoutError` if the coroutine hasn't returned by the deadline. | Python docs |
| `media_local_path` | `TelegramMessage` field where the absolute path to the downloaded file is stored for the worker. | `models/telegram.py` |
| `media_download_error` | `TelegramMessage` field populated when download fails (e.g. `"timeout after 10s"`). | `models/telegram.py` |

## Solution Sketch

**Option A — Size-aware timeout (simpler):**
1. Before calling `download_media`, inspect `message.file.size` (bytes) if available.
2. Compute `timeout = max(10.0, min(120.0, 5.0 + size_bytes / (1024*1024)))` — 5s baseline + 1s per MB, min 10s, max 120s.
3. If `message.file.size` is unavailable (some media types don't expose it), fall back to 10s.

[...]

PR #1355, opened 2026-05-09
Promote dev persona to SDLC-owner + seed personas in /setup

tomcounsell:
Promotes the developer persona from a 34-line single-stage executor to a peer of the PM persona for full SDLC orchestration.

## Changes

- **`config/personas/developer.md`** (new, 190 lines): canonical Dev persona with:
  - Three modes: single-stage executor, single-issue full-SDLC owner, multi-issue parallel orchestrator
  - 7-phase Mode 3 playbook for parallel fan-out (pre-investigate → build → finalize → review → patch → merge → order constraints)
  - Hard rules baked in: no Claude co-author, only `ruff format`, narrow tests when N agents parallel
  - Stale-baseline `merge_authorized` bypass documented
  - Subagent dispatch rules (terse prompts, bake findings, `subagent_type` selection)
  - Existing PROGRESS.md scratchpad pattern preserved

- **`.gitignore`**: un-ignore `config/personas/developer.md` so the in-repo template is version-controlled. PM and Dev now both have in-repo fallbacks; `teammate.md` remains gitignored.

- **`agent/sdk_client.py`**: two loader-warning guards mirror the PM pattern. Warn on overlay load if `Mode 3` or `merge_authorized` substrings are missing — these signal the private overlay has rolled back to the pre-promotion version.

- **`.claude/skills-global/setup/SKILL.md`**: replace the broken placeholder (which referenced nonexistent `config/personas/_base.md`) with an idempotent seeding loop that copies `config/personas/{project-manager,developer}.md` to `~/Desktop/Valor/personas/` if absent. Verified: first-run copies, second-run skips, customizations preserved.

- **`tests/unit/test_persona_loading.py`** (patch 621bc827): `test_logs_missing_with_fallback_when_overlay_absent` now monkeypatches `PERSONAS_BASE_DIR` to a tmp dir. Without this, the test silently passed via the new in-repo developer fallback (`config/personas/developer.md` is now tracked) instead of raising `FileNotFoundError`, so the canonical "Persona overlay missing:" WARN under test never fired.

- **`tests/unit/test_developer_persona_guards.py`** (new, patch 621bc827): mirrors `test_pm_persona_guards.py`. Asserts the load-bearing sections (`Mode 3`, `merge_authorized`, `## Permissions`, `## Modes of Operation`, `## Hard Rules`) survive overlay drift. Pairs with the runtime guards in `agent/sdk_client.py`.

## Why

The orchestrator session that just shipped 8 PRs / closed 8 upvote issues used a parallel-builders-in-worktrees pattern that wasn't accessible to dev sessions because the dev persona was set up as a single-stage executor downstream of the PM. This PR makes that pattern part of the dev contract directly. PM promotion is planned as a follow-up.

## Verification

- `76/76` persona-loading + dev/PM persona-guards tests green (post-patch 621bc827)
- `13/13` compose-system-prompt parity tests green (cache-prefix invariant preserved)
- Loader smoke test: dev system prompt assembled at 41,492 chars (up from 38KB), no warnings emitted
- Setup seed loop verified: first-run copies, second-run skips, user customizations preserved

### Correction (patch 621bc827)

An earlier revision of this PR claimed `test_logs_missing_with_fallback_when_overlay_absent` was a "pre-existing unrelated failure present on main." That claim was incorrect — the test passes on main and was regressed *by this PR* because tracking `config/personas/developer.md` made the unknown-persona fallback in `load_persona_prompt()` resolve to a real file, suppressing the `FileNotFoundError` the test depended on. Patch 621bc827 fixes the test (monkeypatch `PERSONAS_BASE_DIR`) and adds `test_developer_persona_guards.py` to cover the load-bearing sections of the new persona.

## Out of scope (follow-up)

- PM persona promotion (next PR)
- Dev persona overlay drift detection beyond two substrings (could expand)
- Auto-seed personas in `/update` (currently only `/setup`)

### DIFF
--- a/.claude/skills-global/setup/SKILL.md
+++ b/.claude/skills-global/setup/SKILL.md
@@ -276,13 +276,43 @@ Example minimal project entry:
 }
 ```
 
-Persona overlay files also live in `~/Desktop/Valor/personas/`. If missing, create them:
+### Persona overlays
+
+Persona overlay files live in `~/Desktop/Valor/personas/`. The loader (`agent.sdk_client.load_persona_prompt`) prefers the private overlay when present and falls back to the in-repo template (`config/personas/<persona>.md`) otherwise. Seeding the private overlays from the in-repo defaults at setup time gives the agent identical behavior on every fresh machine without waiting for iCloud propagation from another box.
+
+The PM and developer personas have in-repo templates that are version-controlled and PR-reviewable:
+- `config/personas/project-manager.md` — PM pipeline gate rules (CRITIQUE mandatory, REVIEW mandatory, multi-issue fan-out)
+- `config/personas/developer.md` — Developer SDLC-owner playbook (Mode 3 parallel orchestrator, `merge_authorized` bypass)
+
+Seed them into the vault if not already present (do NOT overwrite — existing overlays may carry per-machine customizations):
 
 ```bash
 mkdir -p ~/Desktop/Valor/personas
-# Create developer.md, project-manager.md, teammate.md with role-specific instructions
-# See config/personas/_base.md for the shared identity that gets prepended
+
+for persona in project-manager developer; do
+  src="config/personas/${persona}.md"
+  dst="$HOME/Desktop/Valor/personas/${persona}.md"
+  if [ ! -f "$dst" ]; then
+    cp "$src" "$dst"
+    echo "Seeded $dst from $src"
+  else
+    echo "$dst already exists — leaving in place (run \`diff\` to compare with $src)"
+  fi
+done
 ```
+
+The `teammate.md` overlay is still operator-customized (not seeded by setup); if missing, the in-repo `config/personas/teammate.md` would be the fallback but that file is intentionally gitignored, so a teammate-only machine must author its own overlay.
+
+If the machine is already running and you want to inspect drift between the in-repo template and the private overlay:
+
+```bash
+diff config/personas/project-manager.md ~/Desktop/Valor/personas/project-manager.md
+diff config/personas/developer.md ~/Desktop/Valor/personas/developer.md
+```
+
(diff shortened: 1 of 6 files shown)


tomcounsell:
## Mode-routing decision (recorded for future readers)

**Question raised in review:** Does the dev persona decide for itself which mode it's in?

**Answer:** Yes — self-selected by heuristic on input shape. No PM gating, no pre-routing layer, no `--mode` flag. The persona overlay says "you operate in one of three modes, chosen by the input shape" and lists trigger heuristics:
- **Mode 1** — dispatched with a single stage skill (`/do-build`, `/do-test`, etc.)
- **Mode 2** — one issue number + "drive to completion" verb
- **Mode 3** — ≥2 issue numbers, a multi-tier plan, or explicit fan-out instruction

### Known ambiguity risks

1. **Mode 1 vs Mode 2.** "Build #1322 and ship it" — `/do-build` invocation or "drive to completion"? A Mode-2 mis-classification keeps running past TEST/REVIEW/MERGE when the caller wanted single-stage.
2. **Mode 2 vs Mode 3.** "Drive issue 1322, also note the followup #1325" — does the second `#N` count as a target or a reference?
3. **No caller override.** No header, no flag, no env knob. Caller cannot force a specific mode against the heuristic.

### Alternatives considered

- Explicit `--mode N` flag on `valor-session create --role dev` (uses existing `AgentSession.session_mode` field)
- `MODE: N` header at the top of the dispatch message
- Drop modes entirely; always run the Mode 3 playbook with N=1 as the degenerate case

### Decision

**Keep self-selection (heuristic).** Acceptable when callers send unambiguous messages. Revisitable if mis-firing becomes a real problem in production — the alternatives above are all surgical to add later (the `session_mode` field already exists, and a header rule is just persona text).

If you observe a mis-fire, file an issue with the offending dispatch message verbatim and the mode the dev picked vs the mode you wanted.

tomcounsell:
## Review: Approved

Reviewer subagent verdict was CHANGES_REQUESTED on a real regression; patch agent landed fix in commit 621bc827 (76/76 tests green). PR body corrected. Approved post-patch.

PR #1455, opened 2026-05-25
Docs auditor: docs_features_agent-definition-fallback_md

valorengels:
Automated docs auditor pass.

PR #1545, opened 2026-06-01
SDLC Pipeline Portability: 7 generic robustness fixes (#1535)

### LINKED ISSUE #1535: SDLC pipeline: 7 generic robustness defects surfaced running against a non-ai repo (plan resolution, router routing, portability)
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/sdlc_1535_pipeline_portability.md

## Summary

Running the `/sdlc` pipeline end-to-end against a **non-`ai` target repo** (the Cyndra Bun/TS monorepo at `~/src/cyndra`) over three consecutive issues (#139, #140, #145 in that repo) surfaced a cluster of generic defects in the SDLC router + tooling. Each issue required manual nursing to advance — the pipeline twice hard-blocked itself on the G4 oscillation guard, and the terminal merge stage had no skill at all. None of these are repo-specific; they're robustness gaps in the general-purpose tooling that bite any repo that isn't `~/src/ai`.

This issue collects all seven into one body. They split into three themes: **plan resolution** (1–2), **router routing logic** (3–5), and **portability/degradation** (6–7).

---

### 1. `find_plan_path()` should derive the target repo from the cwd git toplevel, not default to `~/src/ai`

`tools/_sdlc_utils.find_plan_path()` defaults its search root to `~/src/ai/docs/plans` when `SDLC_TARGET_REPO` is unset. Run the pipeline in any other repo without exporting that env var and the lookup silently searches the wrong tree, finds nothing, and `revision_applied` always computes `false`. That mis-read **routed the pipeline back to `/do-plan` on every turn until the G4 oscillation guard hard-blocked it** — observed on both Cyndra #140 and #145.

Setting `SDLC_TARGET_REPO` works around it, but the default is a footgun. **Fix:** derive the repo root from `git rev-parse --show-toplevel` of the cwd (falling back to the env var, then to `~/src/ai`). Belt-and-suspenders so a missing env var degrades to "correct" instead of "silently wrong."

### 2. Plan-path lookup should match the tracking URL, not only a literal `#<n>`

The lookup locates a plan by grepping `docs/plans/` for the string `#<issue_number>`. A plan that references its issue only via the **tracking URL** (`https://github.com/<org>/<repo>/issues/145`) and never as the bare `#145` is invisible to the resolver. This stalled Cyndra #145 until a `> Tracking issue: #145` marker was added by hand — a silent failure with no diagnostic. **Fix:** also match `issues/<n>` (and ideally the full URL form) when resolving the plan file.

---

### 3. Row-4c precedence bug — `/do-build` re-proposed on a completed, approved, clean PR

`sdlc-tool next-skill` returns `/do-build` even when BUILD/TEST/REVIEW are all `completed` and the PR is `CLEAN` + `APPROVED`. The "revision-applied → proceed to build" rule (row 4c) is evaluated **ahead of** the downstream-completion check, so a finished build keeps routing back to build. Flagged independently by the `do-pr-review` and `do-docs` sub-skills on Cyndra #144 and #146; re-dispatching `/do-build` against a completed build also risks re-tripping G4. **Fix:** evaluate downstream stage completion / PR-clean state before the revision-applied→build rule.

### 4. `_meta.pr_number` is neither set for out-of-band PRs nor settable

When a PR is created by the build agent without being threaded back through the state machine, `_meta.pr_number` stays `null`, so `next-skill` routes via row 5 ("Build must create the PR") and loops. There's no recovery path:

- `meta-set`'s whitelist only allows `plan_revising` and `plan_hash_at_build_start`.
- `dispatch record --pr-number` writes dispatch *history*, not `_meta.pr_number`.

On Cyndra #139 the only escape was routing by PR (`/sdlc PR <n>`). **Fix:** either whitelist `pr_number` in `meta-set`, or have `next-skill` resolve the open PR for the current branch when `_meta.pr_number` is null.

### 5. `same_stage_dispatch_count` (G4 guard) has no reset path

Once the counter hit 3 on Cyndra #140, G4 kept firing **even after** the underlying cause was corrected (CRITIQUE marked `completed`, `revision_applied: true` recorded). The only way to clear it was to record a *different* legitimate dispatch (`/do-build`), which happens to reset the run. There's no operator command to reset or decrement the counter after a genuine verdict/stage correction, so a transient mis-read can latch the pipeline closed. **Fix:** reset `same_stage_dispatch_count` on any real stage-state transition, and/or expose an explicit reset.

---

### 6. No `/do-merge` skill in the general library

The terminal MERGE stage has no skill in this environment — not in the registry, not in `~/.claude/skills/`. The merge gate (verify state `OPEN`, `mergeable`, CI green, REVIEW `APPROVED`, body has `Closes #`, then squash-merge) had to be performed **by hand** on Cyndra #139, #140, and #145. It's a deterministic gate and belongs in the general skill library so the pipeline can actually reach `Done` on its own. **Fix:** add a `/do-merge` skill that performs the verify-then-merge gate.

### 7. Forked sub-skills silently lag state when the orchestration substrate is absent

In a repo without the Python orchestration infra (`PipelineStateMachine` writer, the `agent`/`scripts/check_*.py` validators, `reflections.docs_auditor`), the forked `do-build` / `do-pr-review` / `do-docs` sub-skills:

- couldn't persist stage transitions, so `stage-query` consistently lagged the real artifact (BUILD showed `pending`/`ready` after a PR already existed — required manual `sdlc-tool stage-marker` calls every stage), and
- degraded the doc cascade and build validation to "skill-only mode" without a loud signal.

They mostly *work*, but the silent lag is what compounds defects 1–5 into a stuck pipeline. **Fix:** make the forked sub-skills detect the absent substrate and (a) emit a clear "state not persisted / running in degraded mode" marker, and (b) ensure the stage-marker writeback either succeeds or fails loudly rather than no-op'ing.

---

## Why these belong together

The through-line is **portability**: the engine assumes it's running in `~/src/ai` with the full Python substrate present and the issue number written as `#N`. The moment those assumptions break (any other repo, a URL-only plan, an out-of-band PR, an absent validator), the failures are *silent* and compound into pipeline lock-ups that the G4 guard then makes worse by latching. Fixing 1–7 makes the generic pipeline robust enough to run unattended in any repo.

## Acceptance criteria

- [ ] `find_plan_path()` resolves the plan from the cwd git root with no `SDLC_TARGET_REPO` set
- [ ] Plan resolution succeeds when the plan references the issue only by tracking URL
- [ ] `next-skill` routes a completed+approved+clean PR forward (REVIEW/DOCS/MERGE), never back to `/do-build`
- [ ] An out-of-band PR routes to REVIEW/MERGE without manual `/sdlc PR <n>` intervention
- [ ] `same_stage_dispatch_count` resets on a real stage transition; G4 does not latch after a correction
- [ ] A `/do-merge` skill exists and performs the deterministic merge gate
- [ ] Forked sub-skills emit a clear degraded-mode marker and never silently no-op a stage writeback

---

*Filed from a Cyndra (`~/src/cyndra`) SDLC session that ran issues #139, #140, #145 through the full pipeline. A separate, repo-specific set of customizations (modeling "merge ≠ deployed-to-customer-Mac" and the Notion Todo `Done` sync) is being tracked on the Cyndra repo, not here.*

---

## Recon Summary

Performed at plan time (2026-06-01) by reading every cited file against current `main`. All 7 defects confirmed present.

**Confirmed:**
- D1: `tools/_sdlc_utils.py:115-119` — `find_plan_path()` defaults to `Path(__file__).resolve().parent.parent / "docs" / "plans"` (≈`~/src/ai`) when `SDLC_TARGET_REPO` unset. No cwd-git-root derivation.
- D2: `tools/_sdlc_utils.py:124,133` — `needle = f"#{issue_number}"` with substring `in text` match. Misses `issues/<n>` URL form; also false-positives (`#145` matches `#1455`).
- D3: `agent/sdlc_router.py:651-658` — row-4c predicate `_rule_critique_ready_with_concerns_revision_applied` lacks the `if meta.get("pr_number"): return False` guard that row-4a has (line 635) and has no downstream-completion check, so it re-proposes `/do-build` on a completed PR. (G6 fast-path only saves it when CI/merge-state/DOCS all align.)
- D4: `tools/sdlc_meta_set.py:50-53` — `_KEY_REGISTRY` whitelists only `plan_revising`, `plan_hash_at_build_start`. `tools/sdlc_stage_query.py:269-275` `_compute_meta` resolves `pr_number` from `session.pr_number` or `_lookup_pr_number()` (issue-number search only; no branch-head fallback). No write path for out-of-band PRs.
- D5: `agent/sdlc_router.py:1076` `compute_same_stage_count` derives the streak from `_sdlc_dispatches` snapshots; `tools/sdlc_stage_query.py:286` calls it with **no** `current_snapshot`, so it returns the historical run length and never compares live state. `guard_g4_oscillation` (line 377) blocks at `>=3` with no reset path; no operator reset command exists.
- D6: No `/do-merge` skill in `.claude/skills-global/`, `.claude/commands/`, or `~/.claude/skills/`. Only `docs/sdlc/do-merge.md` (repo addendum) exists. `.claude/hooks/validators/validate_merge_guard.py:53-60` expects `data/merge_authorized_{pr}` auth file that nothing currently creates.
- D7: `tools/sdlc_stage_marker.py` `write_marker()` returns `{}` and `main()` exits 0 on every failure (no session, write rejected, exception) — silent no-op. `do-docs/SKILL.md:303` already has a degraded-mode probe (`status: "disabled"` → skill-only); `do-build` and `do-pr-review` have no equivalent signal.

**Revised:**
- D4 fix scope widened: whitelisting `pr_number` in `meta-set` alone is insufficient — `_compute_meta` does not currently read `stage_states["_pr_number"]`, so the read path must also be wired. Plan does both the whitelist AND a branch-head (`gh pr list --head`) fallback in `_lookup_pr_number`.
- D5 fix scope widened: an explicit reset command alone won't fix the latch; `compute_same_stage_count` must reset the streak to 0 when the live snapshot differs from the last recorded snapshot, and `_compute_meta` must pass that live snapshot.

**Pre-requisites:**
- None. All changes are within the SDLC tooling already present in this repo. New `/do-merge` skill auto-deploys via `scripts/update/hardlinks.py` (auto-discovers any `skills-global/*/SKILL.md`).

**Dropped:**
- Cyndra-repo-specific customizations (merge≠deployed modeling, Notion Todo sync) — explicitly out of scope per the issue, tracked on the Cyndra repo.

valorengels:
## Summary

Closes seven generic robustness defects (issue #1535) so the `/sdlc` pipeline runs unattended in **any** repo, not just `~/src/ai`. Each defect is an independent, revertable change.

## Changes

- **D1 — Portable plan resolution** (`tools/_sdlc_utils.py`): `find_plan_path` resolves the plans dir as `SDLC_TARGET_REPO` env → cwd git root (`git rev-parse --show-toplevel`) → `__file__` fallback, via a new `_git_toplevel()` helper that never raises.
- **D2 — Tracking-URL match**: bounded regex `(?:#|issues/){n}(?![0-9])` matches plans referencing the issue by URL and fixes the pre-existing `#1455`-matches-`145` false-positive.
- **D3 — Finished-PR routing guard** (`agent/sdlc_router.py`): row-4b/4c predicates return False once `pr_number` is set or `BUILD == completed`; a completed+approved PR never routes back to `/do-build`. Fixed at the predicate layer (parity test stays green).
- **D4 — `pr_number` recovery** (`tools/sdlc_meta_set.py`, `tools/sdlc_stage_query.py`): `pr_number` is whitelisted in `meta-set` (int-coerced, invalid→exit 2) and read from `_pr_number` in `_compute_meta`; `_lookup_pr_number` gains a branch-head fallback (`gh pr list --head session/{slug}`) using the canonical slug-based branch shape.
- **D5 — G4 self-clear + escape hatch** (`agent/sdlc_router.py`, `tools/sdlc_dispatch.py`): `compute_same_stage_count` resets to 0 when the live snapshot diverges from the last recorded dispatch; `sdlc-tool dispatch reset` clears a latched streak.
- **D6 — Portable `/do-merge` skill** (`.claude/skills-global/do-merge/SKILL.md`): deterministic verify→authorize (`data/merge_authorized_{pr}`)→squash-merge→cleanup gate; defers repo-specific gates to `docs/sdlc/do-merge.md`.
- **D7 — Loud degradation, quiet absence** (`tools/sdlc_stage_marker.py`): tri-state `probe_substrate()` — ABSENT/PRESENT_NO_SESSION emit a visible degraded marker (exit 0); only PRESENT_WRITE_FAILED is loud (exit 1). `do-build`/`do-pr-review`/`do-merge` SKILL.md gained the substrate-probe step.

## Testing

- [x] New unit coverage: resolver (git-root/env/URL/boundary), router D3, oscillation D5, meta-set pr_number, stage-query `_pr_number` + branch-head fallback, stage-marker tri-state contract.
- [x] New integration smoke test `tests/integration/test_sdlc_cross_repo_resolution.py` — resolves a plan in a temp non-`ai` git repo (the test that would have caught D1/D2).
- [x] Repointed `test_do_merge_review_filter.py` from the retired `.claude/commands/do-merge.md` to `docs/sdlc/do-merge.md` (migrated the gate detail there); all 45 do-merge tests pass.
- [x] `ruff check` + `ruff format` clean; verification table all pass.

## Documentation

- [x] New `docs/features/sdlc-pipeline-portability.md` + README index entry.
- [x] Updated `docs/sdlc/do-merge.md` (references new skill; absorbed repo-specific gate logic) and inline docstrings.

## Pre-existing failures (NOT introduced here, verified on main)

- `test_sdlc_skill_md_parity.py::test_step4_*` (3) — SKILL.md heading `## Step 4: Dispatch ONE Sub-Skill (or a Parallel-Safe Pair)` no longer matches the parity extractor; fails identically on main.
- `test_long_task_checkpointing.py::test_progress_md_in_build_soft_check` + `test_skills_audit.py` import — depend on hardlinked `.claude/skills/do-build/SKILL.md` (only present post-`/update`); fail identically on main.
- Docs gate flags `docs/features/README.md:89` ("legacy single-judge path") — a pre-existing line untouched by this PR.

## Definition of Done

- [x] Built, Tested, Documented, Quality (lint/format pass)

Closes #1535


valorengels:
## Review: Approved

**PR:** SDLC Pipeline Portability: 7 generic robustness fixes (#1535)
**Branch:** `session/sdlc_1535_pipeline_portability` → `main`
**Plan:** `docs/plans/sdlc_1535_pipeline_portability.md`

All 7 portability defects (D1–D7) are correctly implemented and match the plan's technical approach. 153 unit tests pass; cross-repo integration smoke test passes; ruff clean.

### Blockers
- None

### Tech Debt
- None

### Nits

**File:** `tests/unit/test_sdlc_utils.py:285` (verified: read this file)
**Code:** `assert result is None or isinstance(result, os.PathLike)`
**Issue:** The assertion is too loose — if pytest's `tmp_path` happens to be nested inside a git repo (unusual but possible), `_git_toplevel` returns a real Path and the test passes without verifying the intended None-return-outside-repo contract. The comment on line 283 says "tmp_path is not inside a git repo" but the assertion doesn't enforce that.
**Severity:** nit
**Fix:** Replace with `assert result is None` to tighten the contract (the comment already says this is the expected behavior).

### Plan Validation

All acceptance criteria verified:
- D1: `find_plan_path()` resolves from cwd git root with no env var set — confirmed by `_git_toplevel()` helper and cross-repo integration test.
- D2: tracking-URL match with bounded regex — confirmed, boundary test passes.
- D3: rows 4b/4c defer when `pr_number` set or `BUILD == completed` — confirmed in router tests.
- D4: `pr_number` settable via `meta-set`, readable from `_pr_number` in `_compute_meta`, branch-head fallback uses canonical `session/{slug}` — confirmed.
- D5: `compute_same_stage_count` resets on live-snapshot divergence; `dispatch reset` subcommand exists — confirmed.
- D6: `.claude/skills-global/do-merge/SKILL.md` exists, creates `data/merge_authorized_{pr}` before merge — confirmed.
- D7: tri-state probe with correct exit codes; degraded-mode step added to do-build and do-pr-review skills — confirmed.

### Verification Results

All plan verification checks pass:
- Tests: 153 unit + 3 integration passed
- Ruff lint and format: clean
- do-merge skill and auth-file gate: present
- `dispatch reset` subcommand: exits 0
- Plan resolver uses git root, URL match, pr_number whitelisted: all confirmed

### Legacy Cruft

No legacy cruft detected. References to the retired `.claude/commands/do-merge.md` path correctly replaced with the new skill.

PR #1561, opened 2026-06-03
fix(reflections): stop daily-digest burst-fire from index-rebuild race

### LINKED ISSUE #1560: Investigate: daily reflections re-fire in bursts (system-health-digest fired 15x/day); Reflection state not persisting (run_count stuck at 1)
## Symptom

`system-health-digest` (declared `every: 86400s`, i.e. once/day) fired **15 times on 2026-06-03** instead of once. Tom received multiple "🩺 Daily System Health Digest" Telegram messages a few minutes apart (e.g. 16:50 and 16:53 Bangkok). The digest is just the visible victim — the underlying fault hits **all** reflections.

## Evidence

- 15 `system-health-digest is due, executing` log lines today, in two bursts (~06:20 UTC and ~09:45 UTC), each ~60s apart within the burst.
- **Worker did not restart** — `agent-session-cleanup` ran on its normal 5-min cadence straight through both bursts. This is not crash-loop re-firing.
- Each burst is an "everything due at once" event: at `09:43:59` a single tick marked `session-liveness-check`, `agent-session-cleanup`, `stale-branch-cleanup`, `redis-index-cleanup`, `circuit-health-gate`, `session-count-throttle`, `failure-loop-detector`, `session-recovery-drip`, AND `system-health-digest` all due simultaneously.
- The `Reflection` record for the digest shows `run_count=1` and a **single** `ReflectionRun` row, despite 15 executions. State is not accumulating — records appear to be **recreated with null `ran_at`** rather than updated.
- Steady-state ticks routinely enqueue 4 reflections; high-water bursts enqueue 7–9. Recurs roughly every ~3h.

## Hypothesis (unverified)

The `Reflection` name index (Popoto `KeyField`) is transiently emptied — likely during an index rebuild (`redis-index-cleanup` / `rebuild_indexes()`) or under memory pressure from heavy agent reflections (`tech-debt-scan` ran 73s right before the 09:45 burst). When `Reflection.get_or_create(name=...)` races into the empty-index window, `query.filter(name=...)` returns `[]`, so it `create()`s a **fresh** record with `ran_at=None`. Every reflection then reads as never-run → due → fires. The duplicate records get cleaned up later, leaving `run_count=1`.

`is_reflection_due()` reads the legacy `ran_at` field; `mark_completed()` writes both `ran_at` and `last_run_summary`. If the read path and the recreate path disagree, the daily guard collapses.

## What to confirm before fixing

1. Does `redis-index-cleanup` / Popoto `rebuild_indexes()` transiently clear the `name` index such that `query.filter(name=...)` returns empty mid-rebuild?
2. Are duplicate `Reflection` records created during a burst (capture `reflection_id` set before/after)?
3. Should `get_or_create` + due-detection be hardened with a ReflectionRun-history fallback ("ran within the last interval → not due") so a lost/blank record can't re-fire a daily job?

## Impact

Low data risk, high annoyance + wasted PM-session spend (each fire enqueues a full PM session that composes + sends a digest). Affects every reflection, not just the digest.

## Repro / data

`logs/worker.log` 2026-06-03 06:18–06:36 UTC and 09:43–09:49 UTC.

valorengels:
Closes #1560

## Problem

Tom received ~15 `Daily System Health Digest` messages in one day. The `system-health-digest` reflection (`every: 86400s`) was firing repeatedly, yet `run_count=1` on the surviving record — a tell that each fire hit a *different*, freshly-created record.

## Root cause

`rebuild_indexes()` (run by `run_cleanup()` at worker startup) **deletes** a model's class-set and KeyField index sets before reconstructing them. During that window, `Reflection.query.filter(name="system-health-digest")` returns empty even though the backing hash still exists. A concurrent scheduler tick's `get_or_create()` then spawns a **fresh duplicate** `Reflection` record with `ran_at=None`. For an `every:` schedule, `compute_next_due(last_run=None)` returns `now` → immediately due → fires. Multiple ticks inside the window (e.g. an old worker's scheduler still alive during a restart overlap) → burst-fire.

## Fix (two-part, defense in depth)

1. **Eliminate the race source** — `scripts/popoto_index_cleanup.py`: exclude live `get_or_create`-per-tick scheduler-state models (`_SCHEDULER_STATE_MODELS = {"Reflection"}`) from the destructive rebuild sweep. They are small and continuously indexed by their own `save()` hooks, so a periodic destructive rebuild buys nothing while creating the empty-index window. (`ReflectionRun` is already excluded — not in `models.__all__`.)

2. **Trigger-agnostic guard** — `agent/reflection_scheduler.py`: in `is_reflection_due()`, when an `every:` record's `ran_at` is lost, recover the true last-run from `ReflectionRun` history (a reliable source — never destructively rebuilt). A blank record can no longer re-fire until its real interval elapses, regardless of *how* `ran_at` went missing.

## Tests

- `test_blank_record_with_recent_history_not_due` — blank record + recent history → **not** due (regression).
- `test_blank_record_without_history_is_due` — genuinely never-run → still due (no over-suppression).
- Existing `is_reflection_due` / cleanup tests pass. The 3 remaining failures in `test_reflection_scheduler.py` are pre-existing format drift (old tests hardcode the legacy `interval:` key; the vault registry uses `every:`) and fail identically on main — unrelated to this change.

## Deployment

Takes effect on the next worker restart (`./scripts/valor-service.sh restart`) after merge. Live state is currently stable (single clean record), so this prevents recurrence on the next restart-race rather than putting out an active fire.

### DIFF
--- a/agent/reflection_scheduler.py
+++ b/agent/reflection_scheduler.py
@@ -288,6 +288,33 @@ def _resolve_callable(dotted_path: str) -> Any:
     return getattr(module, func_name)
 
 
+def _latest_run_timestamp(name: str) -> float | None:
+    """Return the most recent ``ReflectionRun`` timestamp for a reflection name.
+
+    History-based fallback for ``is_reflection_due`` (burst-fire hotfix). When a
+    Reflection record's ``ran_at`` has been lost — e.g. a fresh duplicate record
+    spawned by ``get_or_create`` during a Redis index-rebuild window, where the
+    name index is transiently empty — the record looks "never run" and an
+    ``every:`` schedule would fire on every tick. ``ReflectionRun`` rows are NOT
+    in ``models.__all__`` and so are never destructively rebuilt, making their
+    timestamps a reliable record of when the job actually last ran.
+
+    Returns None if no history rows exist or on any error (fail-open to the
+    existing ``ran_at``-based behavior).
+    """
+    try:
+        from models.reflection_run import ReflectionRun
+
+        timestamps = [
+            r.timestamp
+            for r in ReflectionRun.query.filter(name=name)
+            if isinstance(r.timestamp, (int, float)) and r.timestamp > 0
+        ]
+        return max(timestamps) if timestamps else None
+    except Exception:
+        return None
+
+
 def is_reflection_due(entry: ReflectionEntry, state: Reflection, now: float) -> bool:
     """Check if a reflection is due to run.
 
@@ -308,6 +335,14 @@ def is_reflection_due(entry: ReflectionEntry, state: Reflection, now: float) -> 
     ran_at = state.ran_at if isinstance(state.ran_at, (int, float)) else None
 
     if entry.schedule:
+        # Burst-fire guard: a blank ``every:`` record (ran_at lost during an
+        # index-rebuild race) would be treated as "never run" and fire on every
+        # tick. Recover the true last-run from ReflectionRun history so the job
+        # stays suppressed until its real interval elapses. Scoped to ``every:``
+        # because ``cron:`` anchors on ``now`` (never immediately-due on a blank
+        # record) and ``at:`` is a one-shot.
+        if ran_at is None and entry.schedule.partition(":")[0].strip().lower() == "every":
+            ran_at = _latest_run_timestamp(entry.name)
         try:
             next_due = compute_next_due(entry.schedule, last_run=ran_at, now=now)
         except ValueError as e:
--- a/docs/features/popoto-index-hygiene.md
+++ b/docs/features/popoto-index-hygiene.md
@@ -50,6 +50,8 @@ Each model is processed independently -- one model failure does not abort the sw
 
 `rebuild_indexes()` uses Redis SCAN (cursor-based, non-blocking) and only adds/removes index entries to match actual hash existence. Concurrent creates and deletes are safe and self-correcting -- any inconsistency introduced by a concurrent operation is fixed on the next run.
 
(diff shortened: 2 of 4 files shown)


valorengels:
## Review: Approved

### Blockers
- None

### Tech Debt
- None

### Nits
- None

### Summary

This is a clean, well-targeted defense-in-depth fix for the daily-digest burst-fire bug (issue #1560). The two-part approach is correct:

**Part 1 (race elimination):** `_SCHEDULER_STATE_MODELS = frozenset({"Reflection"})` excludes the Reflection model from the destructive `rebuild_indexes()` sweep. Verified live that `Reflection` is in `models.__all__` (and thus was being swept) and that `ReflectionRun` is NOT in `__all__` (correctly exempt). After this change, `_get_all_models()` returns Reflection=False — confirmed by running the code on the PR branch.

**Part 2 (trigger-agnostic guard):** `_latest_run_timestamp()` queries `ReflectionRun` history (which is never destructively rebuilt) and is called in `is_reflection_due()` exactly when `ran_at is None AND schedule prefix is 'every:'`. The scoping is correct: `cron:` schedules use wall-clock anchoring (never immediately due on a blank record) and `at:` is a one-shot, so neither benefits from the fallback. The function fails silently (returns None) on any error, preserving existing behavior.

**Tests:** The two new regression tests correctly distinguish the blank-record-with-recent-history (not due) from genuinely-never-run (still due). The monkeypatch approach is appropriate for unit tests — avoids Redis dependency while exercising the guard logic.

**Docs:** The concurrency safety exception paragraph in `popoto-index-hygiene.md` accurately describes the root cause and both defense layers.

PR #1586, opened 2026-06-06
Docs auditor: docs_features_agent-session-queue_md

valorengels:
Automated docs auditor pass.


valorengels:
Closing — this automated docs-auditor change is factually incorrect. It re-points `save_session_snapshot()` to `agent/agent_sessions.py`, but that module does not exist; the function is defined in `agent/session_logs.py` (line 72), which is exactly what the doc already said. The pre-PR value was correct. Verified: `grep -rn 'def save_session_snapshot' agent/` → `agent/session_logs.py:72`, and no `agent.agent_sessions` module exists or is imported anywhere. Merging would have corrupted the doc.

Issue #1148, opened 2026-04-23
PM harness sessions missing SESSION_TYPE env and persona — hook enforcement and SDLC rules absent


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/sdlc-1148.md

## Context

> This codebase runs a multi-session AI agent pipeline. A **PM session** (`session_type="pm"`) orchestrates SDLC work by reading GitHub issues, spawning **dev sessions** via `python -m tools.valor_session create --role dev --message "..."`, and waiting for results. All session types execute via a **CLI harness** — a subprocess call to `claude -p --output-format stream-json` (see `docs/features/harness-abstraction.md`). The original SDK-based path (`get_agent_response_sdk` in `agent/sdk_client.py`) is confirmed dead code — not exercised in production.

## Problem

Two environment gaps in the harness execution path leave PM sessions running with an incomplete context. This allowed session `8ff90a3c86ab465a80438ab3d2d64e39` (project: valor, thread: `tg_valor_-1003449100931_690`) to spawn a dev session whose `message_text` was the enriched context headers the PM received — not the actual task content (`SDLC issue 1139`). The dev session replied "the message appears to have been truncated" and stalled. No SDLC work was done.

### Gap 1 — `SESSION_TYPE` not injected into harness env

`agent/session_executor.py` builds `_harness_env` (line 1300) with only:
```python
{
    "AGENT_SESSION_ID": session.agent_session_id or "",
    "CLAUDE_CODE_TASK_LIST_ID": task_list_id or "",
}
```

`SESSION_TYPE=pm` is never added. As a result, `agent/hooks/pre_tool_use.py::_is_pm_session()` (line 99) always returns `False` for CLI harness sessions:

```python
def _is_pm_session() -> bool:
    return os.environ.get("SESSION_TYPE") == SessionType.PM  # always False in harness
```

**Consequences:**
- PM Bash restrictions (`PM_BASH_ALLOWED_PREFIXES`, `_PM_BASH_FORBIDDEN_METACHARS`) are never enforced. The PM can run any Bash command, including mutations.
- `TELEGRAM_CHAT_ID` and `SENTRY_AUTH_TOKEN` injected by `ValorAgent.env` (sdk_client.py line 1114–1127) are also skipped — those paths also gate on session_type, and the executor doesn't replicate them.

### Gap 2 — PM persona never loaded in harness path

[...]

PR #1612, opened 2026-06-08
Granite PTY Container: Production Cutover + Bounded Slot Pool

### LINKED ISSUE #1572: Granite PTY container: production cutover + bounded PTY slot pool
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/granite_pty_production_cutover.md

> **Context:** This repo runs an AI agent system where a Python worker processes inbound messages (Telegram, email) by executing [`AgentSession`](https://github.com/tomcounsell/ai/blob/main/models/agent_session.py) records. Today, every session spawns `claude -p stream-json` — a headless, per-turn subprocess managed by [`agent/session_executor.py`](https://github.com/tomcounsell/ai/blob/main/agent/session_executor.py) and [`agent/sdk_client.py`](https://github.com/tomcounsell/ai/blob/main/agent/sdk_client.py). PR [#1570](https://github.com/tomcounsell/ai/pull/1570) (issue #1546) delivered a kernel-validated PTY-based alternative: [`agent/granite_container/container.py`](https://github.com/tomcounsell/ai/blob/main/agent/granite_container/container.py) drives two persistent interactive `claude` TUI sessions via PTY, with a local Ollama model (granite4.1:3b) routing between them. That PoC confirmed the architecture works. This issue cuts it over to production and adds a bounded slot pool to cap process count.

## Problem

The current production session path (`claude -p stream-json` via `get_response_via_harness`) is a per-turn headless subprocess that exits after each response. This works but has two constraints: it cannot drive Claude Code's interactive TUI features (slash commands, persona priming, trust-folder handling), and it requires the Anthropic API key path rather than the Max subscription OAuth path.

The PTY container architecture solves both — but right now it is a standalone CLI tool (`valor-granite-loop`) with no bridge wiring, no `AgentSession` records, and no output delivery. Running it in production today would produce sessions invisible to the dashboard, unreachable from `valor-session`, and silently delivering results to a local JSON file instead of Telegram.

A secondary problem surfaced in the first live test run (2026-06-05): three test invocations left 6 orphaned `claude --permission-mode bypassPermissions` processes consuming ~1.2 GB after their containers exited. With no concurrency cap, N concurrent granite sessions = 2N persistent claude processes of ~200 MB each.

**Current behavior:**
- `agent/session_executor.py:_execute_agent_session` routes all sessions through `BackgroundTask` → `get_response_via_harness` → `claude -p stream-json`
- `valor-granite-loop` runs standalone; produces a results JSON only; no Telegram delivery, no `AgentSession` record, no heartbeat
- Failed container runs leave orphaned PTY processes; no cleanup enforcement
- First live run hit `pm_hang` after 2 turns because `classify_pm_prefix` receives raw ANSI escape sequences from the PTY buffer and cannot match the `[/dev]` prefix token

**Desired outcome:**
- All bridge-originated sessions execute via `Container`, not `get_response_via_harness`
- A `PTYPool` enforces a hard maximum of N concurrent PM+Dev PTY pairs (configurable, default 3)
- Granite sessions appear in `dashboard.json`, `valor-session list`, and the watchdog like any other session
- Output reaches Telegram/email via the existing `TelegramRelayOutputHandler` path
- ANSI escape sequences stripped before classification; first live run produces a routed turn

## Definitions

| `Container` | The PTY-based session runner: spawns two interactive `claude` TUI processes (PM + Dev) via pexpect, routes between them using a local Ollama classifier | [`agent/granite_container/container.py`](https://github.com/tomcounsell/ai/blob/main/agent/granite_container/container.py) |
| `PTYDriver` | Thin pexpect wrapper around one interactive `claude` TUI session; exposes `write()`, `read_until_idle()`, `close()` | [`agent/granite_container/pty_driver.py`](https://github.com/tomcounsell/ai/blob/main/agent/granite_container/pty_driver.py) |
| `classify_pm_prefix` | Deterministic regex parse of PM's first output line to route to `[/dev]`, `[/user]`, or `[/complete]` — no LLM call | [`agent/granite_container/granite_classifier.py:163`](https://github.com/tomcounsell/ai/blob/main/agent/granite_container/granite_classifier.py) |
| `PTYPool` | Proposed bounded pool of pre-warmed PTY slot pairs; enforces max concurrency; background-respawns slots after use | New class — does not exist yet |
| `_execute_agent_session` | The worker's per-session executor: fires calendar heartbeats, injects steering messages, calls the LLM path, delivers output, runs memory extraction | [`agent/session_executor.py:580`](https://github.com/tomcounsell/ai/blob/main/agent/session_executor.py) |
| `get_response_via_harness` | The current LLM call path: launches `claude -p stream-json` as a subprocess, streams JSONL output, returns final result text | [`agent/sdk_client.py:2261`](https://github.com/tomcounsell/ai/blob/main/agent/sdk_client.py) |
| `TelegramRelayOutputHandler` | Output handler that queues messages to `telegram:outbox:{session_id}` in Redis; the bridge relay drains and delivers via Telethon | [`agent/output_handler.py:131`](https://github.com/tomcounsell/ai/blob/main/agent/output_handler.py) |
| `AgentSession` | Popoto ORM model persisted in Redis; tracks session status, heartbeats, steering queue, and resume UUID; powers dashboard and watchdog | [`models/agent_session.py`](https://github.com/tomcounsell/ai/blob/main/models/agent_session.py) |
| Granite (Ollama) | Local `granite4.1:3b` model used as the routing classifier between PM and Dev turns; runs via `ollama` on the same machine | [`agent/granite_container/granite_classifier.py:DEFAULT_MODEL`](https://github.com/tomcounsell/ai/blob/main/agent/granite_container/granite_classifier.py) |

## Prior Context

- #1546 (PR [#1570](https://github.com/tomcounsell/ai/pull/1570)): Granite operator PTY PoC — kernel-validated two-PTY coordination; confirmed OAuth path, trust-folder handling, persona priming, and granite routing work. Production cutover explicitly marked out-of-scope and deferred to a follow-on issue (this one).
- #1542: Cancelled predecessor — attempted to drive the TUI via headless `-p`; cancelled when the PoC showed interactive PTY was required.

## Solution Sketch

**Phase 1 — Fix the blocking bug**
Strip ANSI escape sequences in `classify_pm_prefix` before regex matching. One `re.sub(r'\x1b\[[0-9;]*[mGKHF]', '', pm_tail)` call at the top of the function unblocks all subsequent work.

**Phase 2 — PTYPool**
Add `agent/granite_container/pty_pool.py`: a `PTYPool` class with N pre-warmed slot pairs. Each slot cycles through `idle → locked → resetting → idle`. On `release()`, the old PTY is killed and a fresh one spawns in the background (not `/clear` reuse — fresh spawn guarantees isolation). `acquire_pair()` blocks until a pair is available. Pool size is the max-concurrency cap: no slot available = session waits in the Redis queue.

**Phase 3 — Wire Container into `_execute_agent_session`**
Replace the `BackgroundTask` + `get_response_via_harness` call at `session_executor.py:1708` with a `Container.run()` call (or a thin adapter). Before the call, wire the same lifecycle hooks the harness path fires: T+0 heartbeat write, `_active_sessions` registration, `AgentSession` status transitions, output callback to `TelegramRelayOutputHandler`, steering message injection between turns, and post-session memory extraction.

**Phase 4 — Delete old PoC code**
Once parity is confirmed via integration tests: delete `agent/granite_agent_loop.py`, `agent/granite_router.py`, `scripts/granite_poc.py`, `scripts/granite_questions_game.py`, and their unit tests. The `claude -p stream-json` streaming parser in `sdk_client.py` can be removed in a subsequent PR once no other caller remains.

## Recon Summary

**Confirmed — 5 items:**
- `classify_pm_prefix` has no ANSI stripping; first live run returned raw escape bytes as `pm_first_line` and classified every turn as `unknown`
- `granite_agent_loop.py` and `granite_router.py` are used only in PoC scripts (`scripts/granite_poc.py`, `scripts/granite_questions_game.py`) and their tests — safe to delete
- `_execute_agent_session` fires 10+ lifecycle hooks the Container must absorb: T+0 heartbeat, calendar heartbeats (start + 25 min timer), SDK heartbeats (60s), steering injection, output delivery, `_active_sessions` registration, memory extraction, dev-session completion handler, session snapshot saves
- `TelegramRelayOutputHandler` queues to `telegram:outbox:{session_id}`; bridge relay polls and delivers — no interface change needed, just call it from the container
- `agent/claude_session.py` is used only by the old `GraniteAgentLoop` — safe to delete alongside `granite_agent_loop.py`

**Revised — 1 item:**
- `sdk_client.py` deletion was originally in scope. Recon shows it is 3,759 lines with many callers beyond the harness path (persona loading, prompt composition, etc.). The harness-specific functions (`get_response_via_harness`, `_run_harness_subprocess`) should be deleted in a follow-on PR after confirming no remaining callers.

**Pre-requisites — 1 item:**
- ANSI stripping fix (Phase 1) must land before any integration testing of the wired container; without it every test run produces `pm_hang` after 2 turns.

**Dropped — 1 item:**
- PTY reuse via `/clear` was considered and rejected: `/clear` does not guarantee full context reset (tool permissions, injected state may persist). Kill + fresh-spawn is the correct slot-reset design.

## Acceptance Criteria

- [ ] `classify_pm_prefix` strips ANSI escape sequences; a live `valor-granite-loop --user-message "handle PR 1568"` run produces at least one classified turn (not `pm_hang` after 2 unknowns)
- [ ] `PTYPool` exists at `agent/granite_container/pty_pool.py`; pool size is configurable; `acquire_pair()` blocks when all slots are locked; slots background-respawn after `release_pair()`
- [ ] Bridge-originated sessions route through `Container` via `_execute_agent_session`; `claude -p stream-json` is no longer spawned for normal sessions
- [ ] Granite sessions create an `AgentSession` record on start and update status to `completed`/`failed` on exit; `curl localhost:8500/dashboard.json` shows the running session
- [ ] `valor-session list` shows an active granite session; `valor-session status --id <id>` shows heartbeat age < 120s during execution
- [ ] Output classified as `[/user]` or `[/complete]` reaches Telegram via `TelegramRelayOutputHandler`
- [ ] Steering messages queued to `AgentSession.queued_steering_messages` are injected between PM turns
- [ ] Calendar heartbeat fires at session start and on the 25-minute timer
- [ ] Post-session memory extraction fires on exit
- [ ] After a 3-run test session, `ps aux | grep 'claude --permission-mode'` shows ≤ pool-size × 2 processes (no orphan leak)
- [ ] `agent/granite_agent_loop.py`, `agent/granite_router.py`, `agent/claude_session.py`, `scripts/granite_poc.py`, `scripts/granite_questions_game.py`, and related unit tests are deleted

## Downstream

This issue feeds into `/do-plan`, which will produce a plan document at `docs/plans/granite_pty_production_cutover.md`. The plan is then executed by `/do-build`.

Constraints the planner should be aware of:
- `sdk_client.py` deletion is explicitly out of scope — that is a follow-on issue
- PTY reuse via `/clear` is rejected; only kill + fresh-spawn is acceptable for slot reset
- The `PTYPool` must be a singleton owned by the worker process, not re-created per session

### LINKED ISSUE #1632: Orphaned subagents survive session death and cascade into machine-wide memory exhaustion (rogue agents + shielded pytest orphans + unit tests spawning real claude)
> **Context for the reader:** This repo runs an AI agent system where Claude Code sessions spawn background subagents (builder/validator/reviewer processes, each an OS-level `claude` process) that run shell commands including `pytest`. A separate "granite PTY" subsystem (PR #1612) drives real interactive `claude` TUI processes via pexpect. On 2026-06-11 these layers combined into a memory cascade that drove the operator's 16 GB machine to 14.2 GB used + 6.9 GB swap, killed the Claude Code session orchestrating the SDLC pipeline, and required manual reaping of 18 processes. **This must not be able to happen again.**

## Problem

When a Claude Code session process exits (crash, OOM kill, restart) while background subagents are running, the subagent `claude` OS processes are NOT terminated. They reparent to launchd (PPID==1) and **keep executing their task loop indefinitely** — running test suites, spawning subprocesses — with no harness to collect results. Their own children (pytest runs, claude TUI spawns) then orphan in turn. Existing reapers missed every layer of the cascade.

**Current behavior:** observed live, 2026-06-11 ~14:25 (operator Activity Monitor screenshots + process census in session log):
- 2 rogue subagents (PIDs 91130/91290, `claude --permission-mode bypassPermissions`, PPID==1) still working 25+ min after their parent died, spawning fresh `claude --print` subprocesses at ~4/min
- 21 orphaned `claude --permission-mode bypassPermissions --model sonnet --print ping` one-shots (plus 4 with `--model gemma4:e2b` — a fixture model name) at ~235–270 MB each, PPID==1, never exiting (~5 GB total)
- 7 orphaned `python -m pytest tests/unit/ -n0` runs at 480–670 MB each, 40+ min runtime
- `scripts/reap-xdist.sh` reported "No orphan pytest processes" while all 7 existed
- Killing the 2 rogue agents stopped spawn accumulation immediately; TERM→KILL of all 18 freed ~6.5 GB

**Desired outcome:** a session-process death never leaves working subagents, accumulating claude one-shots, or multi-hundred-MB pytest runs behind; existing reapers catch every layer within minutes, not hours; unit tests cannot exec the real `claude` binary at all.

## Definitions

| Term | Definition |
|------|------------|
| Subagent | A background `claude` OS process spawned by a Claude Code session via its Agent tool to do delegated work |
| `reap-xdist.sh` | [`scripts/reap-xdist.sh`](../blob/main/scripts/reap-xdist.sh) — kills orphan pytest processes (xdist workers + serial drivers); dry-run by default |
| Worker orphan reaper | Hourly `agent-session-cleanup` reflection scanning for PPID==1 claude/MCP orphans — [`agent/session_health.py:2627`](../blob/main/agent/session_health.py) (issue #1271) |
| Granite PTY pool | Pre-warmed interactive `claude` TUI pairs driven via pexpect — PR #1612, `agent/granite_container/` |

### 1. Rogue orphan subagents survive session death and keep working

No parent-death kill switch exists. The two rogue agents matched the worker reaper's signature (claude cmdline + PPID==1) but the hourly cadence is far too slow for a 4-spawns/min cascade — and `agent/session_health.py:47-49` notes launchd-respawned workers legitimately have PPID==1, so the reaper carries additional gating that needs review against this case. Open questions for the plan: (a) can the harness kill its process group on exit (it died involuntarily — likely OOM — so cleanup-on-exit may never run; a watchdog angle may be needed); (b) should subagents poll their parent PID and self-terminate; (c) should the worker reaper run far more often with a tighter signature (`--print` one-shots older than N minutes are always dead weight).

### 2. `reap-xdist.sh` misses pytest shielded by an orphaned-but-alive zsh wrapper

Each orphan pytest's direct parent was a `/bin/zsh -c` wrapper that was itself PPID==1 — alive, blocked waiting on its child forever. The script keeps any pytest whose immediate parent is in the live process table (`scripts/reap-xdist.sh:75-77`); its own header (lines 13–15) intends to cover wrappers "that have already exited" but not living orphaned ones. Fix: walk one level up — a parent that is itself a PPID==1 `zsh -c` / `sh -c` wrapper is a dead chain; the pytest under it is an orphan.

### 3. Unit tests exec the real `claude` binary

At `6e8de6d8` (branch `session/granite-pty-production-cutover`), `tests/unit/granite_container/test_bridge_adapter.py:340-344` wraps `PTYDriver.spawn` with a tracking function that **calls the real spawn through** (`return original_spawn(self)`), and fixture model values (`gemma4:e2b`) flow into real `claude --print` invocations. Every granite unit-test run breeds real ~250 MB claude processes; orphaned runs breed them unsupervised. Fix: unit tests must never exec the real binary — inject a fake `claude` executable (PATH shim) or mock the spawn layer entirely, and add a guard (conftest-level or CI) that fails any unit test attempting to exec `claude`.

## Solution sketch

Per-mode fixes above; mode 1 carries genuine open questions (root cause of "no kill switch" spans the Claude Code harness, which we don't control, and our own reaper, which we do — the plan should weigh watchdog-side detection vs reaper hardening rather than assume either). Modes 2 and 3 are concrete and small. A fourth, cheap layer worth considering: a PreToolUse hook or guidance change so agent-run pytest goes through `scripts/pytest-clean.sh` with bounded scope instead of bare `pytest tests/unit/` serial runs (six concurrent full-unit serial runs was itself half the memory bill).

## Evidence

Operator Activity Monitor screenshots (14:11, 14:25 local); process census: orphan `claude --print` swarm PIDs 16492–16829 (sonnet) + 18088–18171 (gemma4:e2b), all PPID==1; rogue agents 91130/91290 running pytest via child zsh 12117; orphan pytest PIDs 69893/70868/70355/70437/70495/70499/77327 with PPID==1 zsh wrappers 69842/70853/70318/70395 et al.

## Downstream

This issue will be consumed by `/do-plan` to produce `docs/plans/{slug}.md`. Related prior art: #1271 (worker orphan reaper), #1537 (PPID==1 reaper assumptions), the xdist reaper note in `pyproject.toml`, and memory `feedback_pytest_serial_orphan_leak`.

valorengels:
## Summary

Cuts over all bridge-originated `AgentSession` runs from the headless
`claude -p stream-json` harness to the granite PTY container (interactive
`claude` TUI driven over PTYs). Adds a bounded `PTYPool` and a `BridgeAdapter`
for mid-loop output delivery. All-or-nothing cutover — no runtime fallback flag.

Plan: `docs/plans/granite_pty_production_cutover.md`

## Changes

- **ANSI hardening**: `granite_classifier.py` now reuses `pty_driver._strip_ansi`
  (CSI+OSC+keypad union); 3 new unit tests cover each escape family.
- **`PTYPool`** (`agent/granite_container/pty_pool.py`): singleton, semaphore-bounded
  (`granite.pty_pool_size`, default 3), race-free background respawn, PID-targeted
  orphan cleanup via `data/granite_pty_pids.json`.
- **`BridgeAdapter`** (`agent/granite_container/bridge_adapter.py`): resolves
  `send_cb` once, delivers `[/user]`/`[/complete]` payloads mid-loop (synchronous
  callback contract), writes `exit_summary`/`exit_anomaly`/`delivery_failure` to
  `session_events`, returns `""` (BackgroundTask `send_result=False`).
- **Executor wiring** (`agent/session_executor.py`): replaces the
  `get_response_via_harness` call with `BridgeAdapter.run` via `asyncio.to_thread`.
- **Worker startup hook** (`worker/__main__.py`): pool init + PID-targeted orphan kill.
- **Config**: `GraniteSettings.pty_pool_size` (env `GRANITE__PTY_POOL_SIZE`, default 3).
- **PoC deletion**: removed `agent/granite_agent_loop.py`, `agent/granite_router.py`,
  `agent/claude_session.py`, `scripts/granite_poc.py`,
  `scripts/granite_questions_game.py`, and their tests. Relocated `_RESUME_HINT_RE`
  into `pty_driver.py`. Deleted orphaned `test_claude_session_resume.py`.

### Bug fixed during build

- `BridgeAdapter` silently dropped all `session_events` in production: a fresh
  `AgentSession` has `session_events=None` (`ListField(null=True)`) and the three
  writers early-returned on `None`. Added `_append_session_event` that initializes
  the list before appending.

## Testing

- [x] 140 granite unit + integration tests passing (`tests/unit/granite_container/`,
  `tests/unit/test_session_executor_granite.py`, `tests/integration/test_granite_pty_production.py`)
- [x] All session-executor unit tests passing (36)
- [x] `validate_build.py` passes (4 PASS, 0 FAIL)
- [x] Ruff check + format clean
- [x] No PoC files remain; `_execute_agent_session` no longer calls `get_response_via_harness`
- [x] **Live smoke test** (regression gate): a real `claude` TUI session reaches
  `Container.run`, produces a `[/user]` delivery to Telegram, and the `AgentSession`
  shows `completed` with `last_heartbeat_at < 120s`. **Reviewer confirms before merge.**
  PASSED 2026-06-11 at `c3593caf` (session `-5051653062_1781173846035`): both primes
  real (spinner-evidence gated), `startup cycle=0 pm_idle=True dev_idle=True` on the
  first cycle, reply delivered to the Dev: Valor chat, session `completed` + branch
  cleaned up in 46s end-to-end. Evidence in PR comment below. Known follow-up: the
  delivered text is space-stripped by the screen-capture payload extraction.

## Documentation

- [x] Created `docs/features/granite-pty-production.md` (architecture, config,
  user-visible behavior, per-turn silence cap, observability, revert runbook)
- [x] `docs/features/README.md` index entry
- [x] `## Granite PTY Pool` section in `docs/features/deployment.md` (env var,
  `MAX_CONCURRENT_SESSIONS` relationship, 3→6 growth path, revert runbook)
- [x] Cross-linked `docs/infra/email-cs-auto-reply.md`
- [x] Note in `docs/features/subconscious-memory.md`
- [x] Updated `CLAUDE.md` System Architecture diagram

## Definition of Done

- [x] Tested: granite + executor tests passing
- [x] Documented: docs created/updated per plan

## Notes for reviewer

- The branch was rebased onto current `origin/main` (the original worktree base
  predated several unrelated commits). Diff is now cleanly granite-scoped.
- Out of scope (follow-on issue): deleting the harness code in `sdk_client.py`
  (3,759 lines, many non-harness callers). `_execute_agent_session` no longer
  calls it, but the functions remain.

## Also in this PR

- **`valor` CLI wrapper** (commits `ac0da5fb` + `58540c7d`): a positional-prompt wrapper over `valor-session` — `valor "do the thing"` creates a session without subcommand boilerplate. Ships `tools/valor_cli.py`, the `valor` entry in `pyproject.toml [project.scripts]`, a venv-binary check in `scripts/update/verify.py`, `docs/features/valor-cli-wrapper.md`, and `tests/unit/test_valor_cli.py` (23 tests). Hardening of this wrapper (alias-shadow warning, worker pre-flight trustworthiness) is in PR #1624, which merges immediately after this one per its Decision 1.

- **Orphan-process reaper hardening** (commit `a26da1f8`, closes #1632): this PR's own test runs exposed a machine-wide memory cascade (rogue orphaned subagents + shielded pytest orphans + unit tests spawning real `claude`). Fixed all three repo-side modes: fast-kill of stale `claude --print` one-shots on the worker's 300s health loop, dead-chain wrapper piercing in both reapers, two latent `reap-xdist.sh` bugs (set -e pgrep abort, column-misaligned read), and an autouse guard making `tests/unit/granite_container/` incapable of exec'ing the real binary (`GRANITE_LIVE_SMOKE=1` to opt into live probes).
- **Child-session creation stopgap** (commit `bd41a1a1`, re #1633): new parent-attached AgentSession creation is refused at every chokepoint (CLI, queue, hook, scheduler) until #1633's refactor lands — with the bounded PTY pool, parent+child sessions double-consume slots and risk starvation. Escape hatch `VALOR_ALLOW_CHILD_SESSIONS=1`; existing children and the waiting_for_children lifecycle are untouched.

Closes #1632.

Closes #1572


valorengels:
## Review: Changes Requested

**PR:** #1612 — Granite PTY Container: Production Cutover + Bounded Slot Pool
**Plan:** `docs/plans/granite_pty_production_cutover.md` (issue #1572)
**Preflight:** OPEN, MERGEABLE, mergeStateStatus=CLEAN at head `6e8de6d8`
**Ruff:** clean on the branch. No UI files in the diff (screenshot gate n/a).

### Summary

The core deliverables are well built: `PTYPool` implements the race-free respawn contract (event.clear-first, shielded lock, drain hook), `BridgeAdapter` implements the synchronous delivery contract (ADV-5) including the captured-loop fix for the to_thread callback path, ANSI hardening reuses `_strip_ansi` with the 3 required escape-family tests, PoC deletion is complete, the PID-targeted orphan kill replaces `pkill -f` (OPS-3), and pool-owned pairs are correctly exempted from the machine-wide pkill fallback. The docs are honest — the "Known limitations" section in `docs/features/granite-pty-production.md` is a genuinely good disclosure.

But honesty about a gap is not the same as the gap being mergeable. This is an **all-or-nothing cutover with no fallback flag**: at merge, every session type on every machine loses the harness path. Several of the disclosed "known limitations" are violations of the plan's own prerequisites and of core system invariants (worktree isolation, the PM permission model, reply-to session continuity), and one undisclosed bug defeats the PR's own claimed `session_events` fix. The live smoke test — the plan's explicit merge gate — is unchecked.

### Blockers

**File:** `agent/granite_container/bridge_adapter.py:192` (verified: read this file)
**Code:** `container = Container(` / `user_message=user_message,` / `cwd=working_dir,`
**Issue:** `working_dir` is dead for every production run. The pool pre-warms pairs at worker startup via `await _pty_pool.initialize()` (no cwd → worker process cwd, `worker/__main__.py` Step 4c), and `Container._spawn_pair` returns early for pre-warmed pairs, never consuming `cwd`. Every granite session — including dev sessions with tier-2 worktree isolation and sessions for other projects in `projects.json` — runs its `claude` TUIs in the worker's checkout. The plan's Prerequisites table explicitly requires "the worker passes `session.working_dir`" and this re-opens the #887 worktree-contamination class. Disclosed as Known limitation #1, but a disclosed broken invariant is still a broken invariant in an all-or-nothing cutover.
**Severity:** blocker
**Fix:** Inject the session cwd at acquire time — e.g. prime each acquired pair with a `cd {working_dir}` (plus trust handling) before the persona prime, or respawn-on-acquire with the session cwd when it differs from the pool cwd (accepting the latency hit). At minimum, hard-fail (don't silently mis-run) when `working_dir` differs from the pool's spawn cwd.

**File:** `agent/session_executor.py:1585-1589` (verified: read this file)
**Code:** `# SESSION_TYPE drives pre_tool_use hook behavior (_is_pm_session in` / `# agent/hooks/pre_tool_use.py:97-99). Without it, PM Bash restrictions` / `# are silently disabled in the harness subprocess (issue #1148).` / `if _session_type:` / `_harness_env["SESSION_TYPE"] = _session_type`
**Issue:** The granite path never applies `_harness_env` or `_pm_system_prompt` — both are still computed (lines 1581-1698) and then dropped; `BridgeAdapter` receives neither. The codebase's own comment states the consequence: without `SESSION_TYPE`, PM Bash restrictions are silently disabled — and the granite PTYs run with `--permission-mode bypassPermissions`. The session-type permission model (PM read-only, teammate write-restrictions), task-list isolation (`CLAUDE_CODE_TASK_LIST_ID`), `AGENT_SESSION_ID` attribution, and per-session persona overlays (including the email persona that issue #1148 / email-bridge docs call load-bearing) are all inactive on the new production path. The plan's Agent Integration section states "The agent persona is the SAME persona as the headless harness path" — it is not.
**Severity:** blocker
**Fix:** Spawn (or re-prime) pool PTYs with the per-session env (`SESSION_TYPE`, `AGENT_SESSION_ID`, `CLAUDE_CODE_TASK_LIST_ID`, parent-session ID), and feed the composed persona/system prompt into the prime path instead of discarding it. If true pre-warm + per-session env is irreconcilable, spawn-on-acquire with the session env.

**File:** `agent/granite_container/bridge_adapter.py:94-98` (verified: read this file)
**Code:** `events = getattr(agent_session, "session_events", None)` / `if events is None:` / `events = []` / `agent_session.session_events = events` / `events.append(event)`
**Issue:** `_append_session_event` mutates the in-memory ORM object and never persists it. Every post-run save in `session_executor.py` uses `update_fields=[...]` sets that exclude `session_events`, and finalization (`complete_transcript` → `finalize_session`) loads a *fresh* copy by `session_id`, so the worker's in-memory appends are dropped. The established patterns save explicitly: `agent/output_handler.py:866` calls `session.save()` after its RTR append, and `agent/agent_session_queue.py:466` saves with `update_fields=["branch_name", "session_events", "updated_at"]`. Net effect: `exit_summary`, `exit_anomaly`, and `delivery_failure` never reach Redis — the dashboard/on-call path the plan's OPS-1 hardening depends on is silently empty. This is the same bug class the PR body claims was "fixed during build"; the fix initialized the `None` list but did not persist the writes. The unit tests don't catch it because they assert against an in-memory dataclass stand-in (`tests/unit/granite_container/test_bridge_adapter.py:17-22`).
**Severity:** blocker
**Fix:** In `_append_session_event`, follow the model's documented pattern (`models/agent_session.py:1603`): `agent_session.save(update_fields=["session_events", "updated_at"])` after the append (still inside the fail-silent try). Add one test against a saving fake that asserts the save happened.

**File:** `agent/session_executor.py:1728` (verified: read this file)
**Code:** `_container_message = _minimal_input if _prior_uuid else _harness_input`
**Issue:** For resumed turns (`_prior_uuid` set — every reply-to thread continuation), the granite path sends the *minimal-context* message into a brand-new TUI session. In the harness path, `_minimal_input` was only safe because it was paired with `claude --resume {prior_uuid}`, which restored the full conversation. The container has no resume wiring (disclosed as Known limitation #4), so threaded conversations get a context-free fragment — "Reply-to messages resume the original session context" (CLAUDE.md session-continuity contract) silently breaks on the primary conversational surface. The inline comment ("just like the harness path used to") is misleading on exactly this point.
**Severity:** blocker
**Fix:** Until real resume wiring exists, send `_harness_input` (the full-context turn input) unconditionally: `_container_message = _harness_input`. That makes `_minimal_input`/`_prior_uuid` dead in this path — remove them here and correct the comment.

**File:** `docs/plans/granite_pty_production_cutover.md:488` (verified: read this file)
**Code:** `- [ ] Live smoke test (a real \`claude\` TUI session reaches \`Container.run\` and produces a \`[/user]\` delivery to Telegram) is checked off in the PR template, **and** a modest dashboard improvement is shipped in the same PR`
**Issue:** The live smoke test is the plan's explicit substrate-revalidation merge gate ("If the smoke test fails, the PR is reverted") and the PR body's checkbox is unchecked ("Reviewer confirms before merge"). No evidence of a completed smoke run (a `[/user]` delivery to Telegram from a real TUI session with `completed` status and fresh heartbeat) is attached to the PR. Task 10's dashboard stretch goal is also absent (acceptable per the plan's own "mergeable without this task" note — but the smoke test is not optional).
**Severity:** blocker
**Fix:** Run the live smoke test on the PR branch, paste the evidence (session id, dashboard.json excerpt, outbox payload or Telegram delivery) into the PR, and check the box. The review can complete the REVIEW stage only after that gate closes.

### Tech Debt

**File:** `agent/hooks/liveness_writers.py:136-138` (verified: read this file)
**Code:** `session_id = os.environ.get("AGENT_SESSION_ID")` / `if not session_id:` / `return False`
**Issue:** The two-tier no-progress detector is effectively neutralized for granite sessions. `last_tool_use_at`/`last_turn_at` are written only by hooks reading `AGENT_SESSION_ID` (absent from pool PTY env) and by the sdk_client `result` handler (no longer called). `sdk_ever_output` stays False forever, so sessions ride the sticky own-progress signal in `agent/session_health.py:834` (`if bool((entry.log_path or "").strip()): return True`) indefinitely — no false kills, but Tier-1/Tier-2 can no longer detect a wedged granite session; the only hang defense left is the container's own `CYCLE_IDLE_TIMEOUT_S`. Undisclosed in the feature doc's Known limitations.
**Severity:** tech_debt
**Fix:** Have `BridgeAdapter` bump `last_turn_at` (it holds the live `agent_session`) on each classified turn / delivery, restoring sub-check A semantics; document the change in `docs/features/granite-pty-production.md`.

**File:** `agent/granite_container/container.py:596-612` (verified: read this file)
**Code:** `# Fall through to the` / `# steady-state loop rather than declaring` / `# \`startup_unresolved\`: ...` / `logger.info("container: startup did not settle after %d cycles; " "falling through to steady state", STARTUP_WINDOW_CYCLES)`
**Issue:** The startup loop no longer assigns `exit_reason = "startup_unresolved"` anywhere — the reason is now unreachable, yet it is still claimed live by `bridge_adapter.py:236` (anomaly tuple), the feature doc's observability table, `ContainerResult`'s docstring (container.py:185), and tests (`test_bridge_adapter.py:219`, `test_cli.py:175`). It also invalidates plan Risk 6's mitigation, which relied on `startup_unresolved` as the detection mode for a broken `--permission-mode` flag; that failure now burns steady-state budget and exits `pm_hang` instead.
**Severity:** tech_debt
**Fix:** Either keep a (longer) hard startup ceiling that still exits `startup_unresolved`, or remove the reason from the anomaly tuple/docs/docstring and note the new broken-flag failure signature (`pm_hang` on turn 0) in the feature doc.

**File:** `agent/session_executor.py:1602` (verified: read this file)
**Code:** `_harness_requeued = False`
**Issue:** Dead harness-era residue in the cutover path: `_harness_requeued` is never set True anymore (the only writer was the deleted `do_work` body), so the guard at line 1835 (`if _harness_requeued: return`) is unreachable; `_effective_model` (line 1605) is computed and never consumed (per-session model override silently ignored — Known limitation #2). Violates the repo's NO LEGACY CODE principle: the cutover should fully remove what it no longer uses, or use it.
**Severity:** tech_debt
**Fix:** Delete `_harness_requeued` and its guard/comment block; either pass `_effective_model` through to the PTY spawn path or delete the resolution call and let the doc's limitation entry point at `GRANITE__PM_MODEL`/`GRANITE__DEV_MODEL` as the only knob.

### Nits

**File:** `tools/valor_cli.py:1` (verified: read this file)
**Code:** `"""\`valor\` — the agent-session CLI.`
**Issue:** The PR bundles an entire unrelated feature — the `valor` CLI wrapper (`tools/valor_cli.py`, the `valor` entry in `pyproject.toml [project.scripts]`, a `scripts/update/verify.py` check, `docs/features/valor-cli-wrapper.md`, and a 302-line test module) — with zero mention in the PR body. The code itself is clean and tested, but the PR description must disclose everything the merge ships.
**Severity:** nit
**Fix:** Add an "Also in this PR" section to the PR body describing the `valor` CLI wrapper, or split it into its own PR.

**File:** `agent/granite_container/pty_pool.py:301` (verified: read this file)
**Code:** `await asyncio.sleep(0.01)`
**Issue:** `_wait_for_idle_slot` busy-polls at 100 Hz whenever all slots are locked and none are respawning — granite sessions can hold slots for hours, so a queued acquirer spins for the duration. Separately, the `initialize_pty_pool()` docstring (lines 444-450) claims it "Pre-warms all slots" and "schedules the pre-warm via `asyncio.run`" — it does neither; it only constructs the pool, and the worker performs the awaited `initialize()`.
**Severity:** nit
**Fix:** Wait on a pool-level `asyncio.Condition` (or the per-slot events) notified on release instead of sleep-polling; correct the docstring to match the actual contract.

### Verification Results

The plan has no machine-readable `## Verification` table (Prerequisites + Success Criteria only), so no automated verification run. Success-criteria spot-check on the branch: ANSI reuse + 3 escape-family tests ✅; `pty_pool.py` with semaphore-bounded acquire/release ✅; `bridge_adapter.py` with mid-loop delivery + session_events writes ⚠️ (writes not persisted — blocker above); `get_response_via_harness` no longer referenced by `_execute_agent_session` ✅; PoC files deleted ✅ (`granite_agent_loop.py`, `granite_router.py`, `claude_session.py`, both scripts, and their tests are gone; `_RESUME_HINT_RE` relocated to `pty_driver.py`); per-session cwd/persona/steered-model criteria ❌ (blockers above); live smoke test ❌ (unchecked).

### Legacy Cruft (advisory)

- `agent/session_executor.py`: ~120 lines of harness-input/env/persona composition (1537-1698) now feed a path that consumes only `_harness_input` — after the blockers are fixed this should either be genuinely consumed by the granite path or extracted; today it reads as a half-migration.
- `startup_unresolved` ghost exit reason (tracked as tech_debt above).
- `_harness_requeued` dead guard (tracked as tech_debt above).

### Mergeability

Preflight clean (`MERGEABLE`/`CLEAN`), but verdict is **Changes Requested**: 5 blockers, 3 tech_debt, 2 nits. The pool/adapter/ANSI engineering is solid and most fixes above are small and local (persist-on-append, full-context message, env/persona injection at prime time, smoke-test evidence). Route to `/do-patch`.

valorengels:
## Review: Changes Requested

_Idempotent: prior review on HEAD `6e8de6d` / body hash `24172d6` is still valid — returning prior verdict without regenerating findings._

The [prior review of 2026-06-10](https://github.com/tomcounsell/ai/pull/1612#issuecomment-4668502390) was posted against this exact HEAD (`6e8de6d882`), and no commits or PR-body changes have landed since. Spot-verification on the branch confirms the findings remain accurate at this HEAD:

- `agent/granite_container/bridge_adapter.py` — `_append_session_event` still appends in-memory with no `agent_session.save(update_fields=...)`; `session_events` writes are still not persisted.
- `agent/session_executor.py:1728` — `_container_message = _minimal_input if _prior_uuid else _harness_input` still sends minimal-context input into a fresh TUI with no resume wiring.
- `agent/session_executor.py:1602/1831` — dead `_harness_requeued` guard still present.
- PR body live smoke-test checkbox still unchecked.

**Verdict unchanged: Changes Requested — 5 blockers, 3 tech_debt, 2 nits.** See the prior review comment for full findings and fixes. Route to `/do-patch`.

valorengels:
## Live smoke test evidence (B5) — PASSED at `c3593caf`

**Run:** 2026-06-11, branch worker from the cutover worktree, teammate session `-5051653062_1781173846035` → Dev: Valor chat.

**Worker log (granite container lifecycle, full run 46s):**

```
10:30:46 session_executor: Routing teammate session to the granite PTY container
10:30:46 sdk_client: Persona overlay loaded: name=teammate prompt_chars=33065
10:30:47 pty_pool: slot 0: spawned per-session pair (cwd=/Users/valorengels/src/ai, env_keys=6, persona=set, pm_model=opus)
10:31:07 container: prime post-write wait saw_idle=True buffer_len=5321 elapsed_ms=13925  ← PM, real model turn
10:31:25 container: prime post-write wait saw_idle=True buffer_len=4911 elapsed_ms=12150  ← Dev, real model turn
10:31:25 container: Dev prime done; entering startup loop
10:31:29 container: startup cycle=0 pm_idle=True dev_idle=True response=None
10:31:29 container: startup both idle, breaking
10:31:31 output_handler: Queued output to telegram:outbox:-5051653062_1781173846035 (64 chars)
10:31:32 session_executor: Auto-marked session done and cleaned up branch session/50516530621781173846035
```

**Telegram delivery confirmed** (`valor-telegram read --chat-id -5051653062`):

```
[2026-06-11 17:31] system: Onlineandrouting—PMsessionforPR#1612'sgranitePTYcontainerislive.
```

**Contrast with the failed run** (pre-`c3593caf`, session `-5051653062_1781166529402`): both primes false-idled on deterministic TUI echo (`buffer_len=5486/5485` in ~8.2s from two different models — no model turn ever ran because the submit CR was swallowed by the TUI paste-burst heuristic), then 99 startup cycles of `pm_idle=False dev_idle=False` until the 600s `startup_unresolved` ceiling. Root cause and fix in `c3593caf` (separate-CR submit + level-triggered idle with a 2.0s byte-quiescence gate).

**Distinct primes this time:** PM `5321B/13.9s` vs Dev `4911B/12.2s` — different sizes and durations, i.e. real model responses, not echo chrome.

**Known follow-up (not a routing defect):** the delivered text is space-stripped (`Onlineandrouting—PMsession...`) — an artifact of reconstructing the reply payload from TUI screen-cell captures in the classifier's extraction path. Routing, classification, delivery, completion, and cleanup are all correct; text fidelity needs a follow-up issue before this surface fronts CEO-facing chats.

valorengels:
## Delta Re-Review: APPROVED

**Scope:** delta re-review at head `c3593caf` against the prior full review (5 blockers, 3 tech-debt, 2 nits at `6e8de6d8`). Focus: the three previously unreviewed commits `a26da1f8` (reaper hardening, #1632), `bd41a1a1` (child-session gate, #1633), `c3593caf` (startup_unresolved fix).

### Prior-review state confirmed

- **B1-B4, TD1-TD3, NIT2** (fixed in `b853b461`): spot-checked on the branch. `_append_session_event` persists via `save(update_fields=["session_events", "updated_at"])` (`agent/granite_container/bridge_adapter.py:113`); `_container_message = _harness_input` unconditionally with `_harness_requeued` removed (`agent/session_executor.py:1723`); `_effective_model` is consumed (`pm_model=_effective_model`, `session_executor.py:1714`); per-session `working_dir` + `session_env` reach the container (`bridge_adapter.py:194-217`). Confirmed in place.
- **B5 (live smoke):** PASSED. Head SHA matches (`c3593caf`), the PR-body smoke checkbox is checked, and the evidence comment (4679749113) shows a real first-cycle pass: distinct PM/Dev primes (5321B/13.9s vs 4911B/12.2s, real model turns), startup both-idle at cycle 0, Telegram delivery confirmed, 46s total. The space-stripped delivery text is already filed as #1634 and is not re-flagged.

### Delta findings

**Blockers:** none.

**Concerns**

1. **The QUIESCENCE_S/STARTUP_CYCLE_TIMEOUT_S coupling is documented on only one side.** The commit message for `c3593caf` claims the constraint is "documented at both constants", but it is only at `QUIESCENCE_S` (`agent/granite_container/pty_driver.py:125-127`: "Must stay below STARTUP_CYCLE_TIMEOUT_S (3s)"). The `STARTUP_CYCLE_TIMEOUT_S` comment block (`agent/granite_container/container.py:92-102`) has zero mention of `QUIESCENCE_S` (verified: `grep QUIESCENCE container.py` returns nothing). This is the dangerous side of the coupling: anyone "optimizing" the startup poll down to <= 2.0s makes the quiescence gate unobservable within a poll (every `read_until_idle` deadline fires before `time.monotonic() - last_chunk_at >= QUIESCENCE_S` can pass at `pty_driver.py:510`), and every run exits `startup_unresolved` at the 600s ceiling, the exact failure signature this commit fixed. Add the reciprocal comment at `container.py:102` (or a module-level assert `QUIESCENCE_S < STARTUP_CYCLE_TIMEOUT_S`).

2. **Both feature docs still describe the deleted driver contract (half-migrated docs).** `c3593caf` rewrote C1 and C5 in the in-code docs meticulously but left:
   - `docs/features/pty-driver.md:23-32` (C1): still says `write()` "normalizes the trailing newline to `\r`", i.e. text+CR in one send, which is exactly the Defect A behavior this commit removed (separate-CR submit + `SUBMIT_KEY_DELAY_S`).
   - `docs/features/pty-driver.md:71-80` (C5): still describes the old edge-triggered heuristic and lists `IdleResult` fields as `saw_idle`, `buffer`, `idle_marker`, `elapsed_ms`, omitting `turn_buffer`; no mention of the quiescence gate or `SPINNER_EVIDENCE_RE`.
   - `docs/features/granite-pty-production.md:212-215`: still credits the deleted `LOADING_RE` trailing-400-chars mechanism as a landed hardening ("checks the loading-spinner negative against only the trailing 400 chars").
   Per the repo's no-half-migrations norm these should be updated before merge; it is a small surgical edit.

**Nits**

3. `_fast_reap_stale_print_oneshots()` is invoked synchronously from the async `_agent_session_health_loop` (`agent/session_health.py:2089`). A full `psutil.process_iter` scan blocks the worker event loop every 300s tick; the sibling checks in that loop are awaited. Consider `asyncio.to_thread` for the scan.

4. The new anchored classifier path (`agent/granite_container/granite_classifier.py:222-231`) returns `payload=rest.strip()` with `compliance_miss=False`; a frame where the artifact cut lands immediately after the token (e.g. `[bullet] [/user]` followed by a box border) yields an empty payload. The empty-`[/dev]` case is handled (`container.py:763` nudge path), but empty `user`/`complete` exits with an empty `exit_message` and hands `""` to the delivery callback, and `_deliver_sync` (`bridge_adapter.py:359`) has no empty-payload guard. Low likelihood; worth a one-line guard whenever #1634's payload-fidelity work touches this path.

### Verified safe (delta focus areas)

- **Reaper kill criteria** (`a26da1f8`): an interactive user-owned `claude` can never match the fast-kill signature. `_is_stale_print_oneshot` (`session_health.py:2602`) requires argv basename exactly `claude` in argv[0:2] AND a `--print`/`-p` token AND age > 600s AND a known create_time; the fast pass additionally requires PPID==1 and skips `os.getpid()`. The hourly pass keeps `skip_pids` (registered worker PIDs) checked before signature evaluation, and the dead-chain expansion (`_parent_is_orphaned_shell_wrapper`) only fires for a live `sh/zsh/bash/dash -c` wrapper whose own PPID==1, so a claude under a live terminal, tmux shell, or live Claude Code session is untouchable. SIGKILL escalation is create_time-tuple guarded against PID recycling on both the fast pass and the hourly drain. `reap-xdist.sh` fixes (the `set -e` pgrep abort and the column-misaligned `read` that compared each candidate's own PID to 1) are correct as written.
- **Child-session gate completeness** (`bd41a1a1`): every creation path that attaches a parent is covered. The queue chokepoint `_push_agent_session` (`agent_session_queue.py:253-263`) catches `enqueue_agent_session` and therefore `bridge/dispatch.py`, `tools/valor_session.py` create, and `tools/agent_session_scheduler.py` schedule (the CLIs also gate earlier, pre-side-effect, with exit 2 + structured JSON); the hook auto-attach path is gated at `.claude/hooks/user_prompt_submit.py:233-237` (the early return only skips record creation and sidecar save). The single ungated creator, `session_completion.py:401` `create_pm` continuation, is the documented deliberate exemption (terminal parent holds no pool slot). No raw Redis on Popoto-managed keys anywhere in the delta.
- **Tests:** `test_pty_driver.py` 43 passed (4 skipped, the `GRANITE_LIVE_SMOKE=1`-gated spike regressions, gate checked before any spawn), `test_child_session_gate.py` 11 passed; ruff clean on all delta files. The reaper tests are fake-process based and cover the safety properties directly (interactive claude not matched, young one-shots kept, self-PID skipped, kill switch, attached vs orphaned wrapper, SIGTERM-to-SIGKILL escalation, health-loop wiring). `test_pty_driver_loading_window.py` correctly deleted with its premise; the lone remaining `LOADING_RE` mention in the tree is a historical reference in a regression test docstring (`test_pty_driver.py:482`), which is fine.
- **No duplicate-routing race from the level-triggered `turn_buffer`:** every steady-state continuation path writes to the PM PTY (nudge or summary), resetting `_turn_text`, and `pm_user`/`pm_complete` exit the loop, so the same capture is never classified twice. Startup-event parsing correctly uses the edge-triggered `result.buffer` (`container.py:468`).

### Verdict

**APPROVED.** The three delta commits are correct, well-tested, and live-verified; the two concerns are small documentation-sync fixes (one comment line + three doc paragraphs) that should land as a tiny follow-up commit on this branch before merge, but they do not change behavior and do not block.

Issue #1653, opened 2026-06-12
Patch popoto DatetimeField auto_now/auto_now_add to stamp datetime.now(UTC) (upstream hardening for #1645)


tomcounsell:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/popoto_datetimefield_utc_stamp.md

**Defense-in-depth follow-up to #1645.**

#1645 fixes the *consumer* repo locally by dropping `auto_now=True` on `AgentSession.updated_at` and stamping `utc_now()` explicitly. This issue tracks the *upstream* hardening so any **future** `auto_now`/`auto_now_add` field on any popoto model is correct-by-default.

## Problem

popoto's `DatetimeField.format_value_pre_save` returns bare `datetime.now()` (naive **local** wall time) for both `auto_now` and `auto_now_add`:

- Source: `~/src/popoto/src/popoto/fields/datetime_field.py:116,118` (v1.7.0)
- Installed: PyPI `popoto==1.6.1`, pinned by hash in `ai/uv.lock` — same bug.

The encoder (`popoto/models/encoding.py`) strips tzinfo via `strftime`, so the persisted wall-clock value is whatever `datetime.now()` returned — local, not UTC. On non-UTC hosts this skews every `auto_now` timestamp by the host UTC offset.

## Fix

In `~/src/popoto`, change `format_value_pre_save` to stamp `datetime.now(UTC)` (or keep tz-aware UTC through serialization). Then:

1. Version-bump `~/src/popoto/pyproject.toml`.
2. Build + publish to PyPI.
3. Bump the pin in `ai/pyproject.toml` and refresh `ai/uv.lock`.
4. Propagate via `/update` across machines.

## Why separate from #1645

This is a multi-repo, multi-machine, human-gated PyPI release. #1645 is fully closed by the in-repo local fix without it. This issue is hardening so the bug can't recur on a new `auto_now` field.

## Recon Summary

*(Added at plan time, 2026-06-15, baseline `777570c5`. Confirms the source-level evidence and resolves the publish mechanism, which the issue body left implicit.)*

**Confirmed:**
- popoto `DatetimeField.format_value_pre_save` returns bare `datetime.now()` (naive local) for both branches — `~/src/popoto/src/popoto/fields/datetime_field.py:116` (auto_now_add) and `:118` (auto_now), v1.7.0. Module imports only `from datetime import datetime` (line 25) — no `UTC`/`timezone` in scope.
[...]

PR #1655, opened 2026-06-12
fix(#1645): AgentSession.updated_at stamped in UTC (remove auto_now, explicit utc_now() in save())

### LINKED ISSUE #1645: AgentSession.updated_at stamped 7h in the future: popoto DatetimeField auto_now uses naive local datetime.now()
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/sdlc-1645.md

> **Context:** `AgentSession` (the Redis-persisted record for every agent session) is stored via the Popoto ORM. Its `updated_at` field is declared as `DatetimeField(auto_now=True, null=True)` at `models/agent_session.py:153`. Popoto's `DatetimeField` implements Django-style `auto_now`/`auto_now_add` by unconditionally returning **naive** `datetime.now()` (local wall time, no tzinfo) on every save (confirmed by inspecting the installed popoto source).

## Problem

On any machine whose local timezone is not UTC, every save stamps `updated_at` with local time that downstream code interprets as UTC.

**Current behavior (observed live, 2026-06-12, UTC+7 machine):** session `ec1e7c6ede0b4cc491247067dea676f6` shows

- `created_at` epoch `1781261380` = 10:49:40 UTC -- correct (set explicitly with tz-aware UTC)
- `updated_at` epoch `1781286582` = 17:49:42 UTC -- **7 hours in the future** of actual wall-clock time at the moment it was read

`valor-session status` then prints `Updated: 2026-06-12 17:49:42 UTC`, i.e. local time mislabeled as UTC. During live observation this made a 3-minute-old running session look 7 hours stale at first glance.

**Impact.** Anything computing "age since last update" from `updated_at` gets a value skewed by the machine's UTC offset (negative ages, or staleness masked for up to 7h): dashboard `is_stale`/sorting, watchdog heuristics, cleanup reflections, and human triage via `valor-session status`. This is the same bug class as closed #777 ("session watchdog LIFECYCLE_STALL duration inflated by UTC offset (~7h) due to timezone-naive datetime"), which fixed one consumer but not this producer.

## Root cause

Popoto `DatetimeField` docstring/implementation: "auto_now ... unconditionally returns `datetime.now()`". `datetime.now()` without `tz=` is naive local time. The model layer already has a defensive epoch converter that assumes UTC (`models/agent_session.py:781`: `datetime.fromtimestamp(kwargs["updated_at"], tz=UTC)`), so a naive local datetime serialized to an epoch lands +offset in the future.

## Solution sketch (direction, not prescription)

Two candidate fixes; planning should pick one:

1. **Upstream (preferred long-term):** patch popoto's `DatetimeField` to stamp `datetime.now(UTC)` for `auto_now`/`auto_now_add`. We control the popoto dependency (`~/src/popoto` is a sibling repo), and `created_at` paths already assume tz-aware UTC.
2. **Local:** drop `auto_now=True` and stamp `updated_at` explicitly with tz-aware UTC in `save()` paths (the repo convention per `bridge/utc` helpers).

Either way, also:
- Audit all other `auto_now`/`auto_now_add` usages across `models/` (and any other Popoto models in the repo)
- Add a one-shot heal for existing records whose `updated_at` is in the future (clamp to `max(created_at, now)`), consistent with the existing `_heal_descriptor_pollution` pattern
- Fix the `valor-session status` display path to format from tz-aware UTC so the label is truthful

## Definitions

| Term | Definition |
|------|------------|
| Popoto | Redis ORM used for all persisted models in this repo; raw Redis access to its keys is prohibited (see `CLAUDE.md` Manual Testing Hygiene) |
| `auto_now` | Django-compatible flag: re-stamp the field with "now" on every save |
| #777 | Closed issue, same bug class: naive datetime inflating watchdog stall durations by the UTC offset |

## Downstream

## Recon Summary

*(Added at plan time, 2026-06-12, baseline `aa354ce2`. Organizes the source-confirmed evidence already in this issue into the required buckets, plus one new finding.)*

**Confirmed:**
- `models/agent_session.py:153` declares `updated_at = DatetimeField(auto_now=True, null=True)` — exact line, unchanged on current main.
- popoto `DatetimeField.format_value_pre_save` returns bare `datetime.now()` (naive local) for both `auto_now` and `auto_now_add` (`.venv/.../popoto/fields/datetime_field.py:116,118`).
- popoto's datetime encoder strips tzinfo via `strftime("%Y%m%dT%H:%M:%S.%f")` (`.venv/.../popoto/models/encoding.py:87-96`), so the persisted wall-clock *value* is what matters — the producer must mint UTC wall-clock.
- `auto_now`/`auto_now_add` appears on exactly ONE repo model field: `AgentSession.updated_at`. `grep -rn auto_now` across `models/ agent/ bridge/ worker/ reflections/ monitoring/ ui/ tools/` returns only line 153 (plus two test-file assertions). No other producer to fix locally.
- #777 (closed via PR #787) fixed the watchdog *consumer* (`_to_timestamp`), not this producer — same bug class, different layer.

**Revised:**
- The issue frames option 1 ("patch `~/src/popoto`, a sibling repo we control") as a near-equal alternative. Recon found the installed popoto is **PyPI `1.6.1`, pinned by hash in `uv.lock`** — NOT an editable install from `~/src/popoto` (which is at `1.7.0`, `src/` layout, same bug). So option 1 is a multi-repo publish-and-pin + multi-machine `/update` cycle, not a one-line edit. Plan picks **option 2 (local explicit-UTC stamp)** as primary and defers the upstream patch to follow-up #1653.

**Pre-requisites:**
- Redis reachable for the one-shot heal and the integration test (worker-startup heal scans real AgentSession records).

**Dropped:**
- Touching `updated_at` consumers (`to_unix_ts`, `_safe_float`, `_to_timestamp`) — they are already correct *given a correct producer value*; the #557/#777/#819 line already hardened them. Producer-only fix avoids regressing them.
- Healing other models' `updated_at` (e.g. `Chat`) — they do not use `auto_now` (grep-confirmed), so they don't share this defect.

tomcounsell:
## Summary

- Removes `auto_now=True` from `AgentSession.updated_at`; adds a `save()` override that stamps `utc_now()` (UTC wall-clock) so the stored value is always correct regardless of host timezone
- Adds `_heal_future_updated_at()` classmethod that clamps legacy future-dated records at worker startup
- Fixes the live-observed symptom: `valor-session status` printing `Updated: 2026-06-12 17:49:42 UTC` when actual UTC was 10:49:42 (UTC+7 host)

## Root cause

popoto's `DatetimeField(auto_now=True)` called `datetime.now()` (no tz arg → naive local wall-clock). Downstream readers all re-attach `tz=UTC` to naive values (correct convention), but the stored wall-clock was local time — so on a UTC+7 host every `updated_at` was stamped 7h ahead of real UTC.

## Changes

- `models/agent_session.py`: drop `auto_now=True`; add `save()` override with UTC stamp + `update_fields` guard; add `_heal_future_updated_at` classmethod (dual-future, idempotent, fail-soft, returns count)
- `worker/__main__.py`: invoke heal at startup alongside index rebuild, with `try/except` + count logging
- `tests/unit/test_agent_session_updated_at_utc.py` (new): 15 unit tests covering UTC stamp, round-trip, heal edge cases
- `tests/integration/test_updated_at_heal.py` (new): 5 integration tests seeding future-dated records via Redis HSET + msgpack, verifying heal corrects them
- `docs/features/session-lifecycle.md`: new "Timestamp Convention" section with first-deploy callout

## Testing

- [x] New unit tests: 15 passed
- [x] Heal integration tests: 5 passed
- [x] Consumer regression suite: 73 passed (watchdog, status, stale-cleanup, liveness)
- [x] No `auto_now=True` in model declarations
- [x] Ruff format/lint clean

## First-deploy note

`_heal_future_updated_at()` runs once at worker startup and clamps future `updated_at` records down to now. Healed sessions will appear momentarily fresh to staleness checks for one window post-deploy — this is expected behavior, not real activity. See `docs/features/session-lifecycle.md` for full callout.

## Follow-up

#1653 tracks the upstream popoto fix (minting `datetime.now(UTC)` for `auto_now`/`auto_now_add`) as defense-in-depth so any future `auto_now` field is correct by default.

Closes #1645

### DIFF
--- a/docs/features/session-lifecycle.md
+++ b/docs/features/session-lifecycle.md
@@ -294,6 +294,22 @@ A stale caller can at worst append a spurious `session_events` entry. It cannot 
 
 `scripts/reflections.py` scans bridge logs daily for `"Stale index entry"` warnings. A non-zero count triggers a finding tagged `(regression marker for #898)`. The `finalized_by_execute` fix should eliminate all such warnings; a reappearance indicates a regression.
 
+## Timestamp Convention — `updated_at` is Explicit UTC
+
+`AgentSession.updated_at` is always an explicit UTC wall-clock timestamp. It is stamped inside the `save()` override using `bridge.utc.utc_now()`, not by a Popoto `auto_now` field.
+
+**Why:** Popoto's `auto_now` calls `datetime.now()` (no `tz` argument), which mints a naive datetime in the host's local timezone. On non-UTC hosts the stored value is naive-local, but every downstream reader (watchdog, dashboard, stale-cleanup) interprets it as UTC. The result is a future-dated `updated_at` for sessions created on hosts running ahead of UTC, causing the watchdog/dashboard to report sessions as perpetually "fresh" and stale-cleanup to skip them forever.
+
+**The fix (issue #1645):** `auto_now` was removed from the field declaration. The `save()` override stamps `self.updated_at = utc_now()` unconditionally unless `update_fields` is provided *without* `"updated_at"` — in which case the stamp is skipped entirely (no in-memory mutation without a matching persist, to avoid memory/Redis desync).
+
+**Rule for new fields:** Do not use `auto_now=True` on any `DatetimeField`. Always stamp explicitly with `utc_now()` at the appropriate call site. The `auto_now` ban is enforced by the comment on the field declaration at `models/agent_session.py` line 153–155 (#1645).
+
+### First-deploy callout
+
+`_heal_future_updated_at()` (a classmethod on `AgentSession`) runs once at worker startup after the fix is deployed. It clamps any `updated_at` that is in the future down to `max(created_at, now)`. After the clamp, those previously-future-dated records appear newly-updated to the watchdog and dashboard staleness checks for exactly one staleness window — operators should **not** interpret this momentary freshness as real session activity. No threshold change is required; the clamp only moves timestamps from future to now.
+
+The heal is idempotent: a re-run clamps only still-future records, so a mid-run restart is safe.
+
 ## Stale Session Cleanup
 
 `_cleanup_stale_sessions()` in `scripts/update/run.py` runs during every `/update` deploy and terminates `running` or `pending` sessions that have no live process. It is a safety net for sessions that were never finalized due to a crash or abrupt restart.
--- a/docs/features/session-lifecycle.md
+++ b/docs/features/session-lifecycle.md
@@ -302,7 +302,7 @@ A stale caller can at worst append a spurious `session_events` entry. It cannot 
 
 **The fix (issue #1645):** `auto_now` was removed from the field declaration. The `save()` override stamps `self.updated_at = utc_now()` unconditionally unless `update_fields` is provided *without* `"updated_at"` — in which case the stamp is skipped entirely (no in-memory mutation without a matching persist, to avoid memory/Redis desync).
 
-**Rule for new fields:** Do not use `auto_now=True` on any `DatetimeField`. Always stamp explicitly with `utc_now()` at the appropriate call site. The `auto_now` ban is enforced by the comment on the field declaration at `models/agent_session.py` line 153–155 (#1645).
+**Rule for new fields:** `auto_now=True` must not be added to any `DatetimeField`. Always stamp explicitly with `utc_now()` at the appropriate call site. This constraint is documented on the field declaration at `models/agent_session.py` line 153–155 (#1645).
 
 ### First-deploy callout
 
--- a/models/agent_session.py
+++ b/models/agent_session.py
@@ -894,12 +894,11 @@ class AgentSession(Model):
                 if hasattr(floor, "tzinfo") and floor.tzinfo is None:
                     floor = floor.replace(tzinfo=UTC)
                 record.updated_at = max(floor, now)
-                # Save via the override (which will re-stamp utc_now(), which == now — idempotent).
-                # Use update_fields to limit the write to only the fields we changed.
-                update_fields_list = ["updated_at"]
-                if record.created_at and record.created_at == now:
-                    update_fields_list.append("created_at")
-                record.save(update_fields=update_fields_list)
+                # Full save (no update_fields) so all changed fields are persisted
+                # and all popoto indexes (including SortedField created_at) are
+                # kept in sync. The save() override will re-stamp utc_now() for
+                # updated_at (idempotent — it will equal now).
+                record.save()
                 count += 1
                 logger.info(f"_heal_future_updated_at: healed session {record.id}")
             except Exception as e:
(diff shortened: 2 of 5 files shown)


tomcounsell:
## Review: Approved

Reviewed against issue #1645. The fix is correct, well-tested, and net-improves timestamp correctness across every consumer.

**Verdict:** Approved — zero blockers, zero tech debt.

### What was verified
- **Root-cause fix is sound.** `auto_now=True` removed from `updated_at`; `save()` override stamps `bridge.utc.utc_now()` (tz-aware UTC). `grep auto_now=True models/` is empty — the issue's "audit all auto_now usages" ask is satisfied (no other Popoto model used it).
- **Heal logic is correct and idempotent.** `_heal_future_updated_at()` clamps future records to `max(created_at, now)`, clamps a dual-future `created_at` first to preserve `created_at <= updated_at <= now`, is fail-soft per-record, and returns a count. Re-runs return 0.
- **Tests pass locally (in an isolated worktree at HEAD `c0e1cbfb`):**
  - `tests/unit/test_agent_session_updated_at_utc.py` — 15 passed
  - `tests/integration/test_updated_at_heal.py` — 5 passed (seeds future-dated records via raw Redis HSET + msgpack, heals, asserts invariants)
  - Consumer regression suite (status CLI, format_ts, stale_cleanup, session_watchdog, liveness fields) — 63 passed
  - `ruff check` clean on all changed files
- **Watchdog/staleness consumers improve, not regress.** On a UTC+7 host the old future-dated `updated_at` made `activity_age = now - updated_at` negative, so `session_watchdog.py` / `ui/data/sdlc.py is_stale` / `session_health` never flagged real stalls. This fix corrects that. The watchdog also maxes `updated_at` against an in-memory activity tracker + transcript mtime, so reduced partial-save bumping cannot cause false-stall.

### Rubric
- Correctness: PASS
- Security: N/A
- Error handling: PASS (heal is fully try/except-wrapped at both fetch and per-record level; worker startup call is try/except non-fatal)
- Tests: PASS (20 new + 63 consumer regression)
- Code quality: PASS
- Documentation: PASS (`docs/features/session-lifecycle.md` — Timestamp Convention + first-deploy callout)

### Nits (None block merge)
- `models/agent_session.py:896` — `record.updated_at = max(floor, now)` is computed and then immediately discarded: the subsequent full `record.save()` re-stamps `updated_at = utc_now()` via the override. The line is harmless (the persisted value is still sane and `>= created_at`) but dead. The created_at clamp above it IS persisted and is the load-bearing part. Could drop the `max()` line or add a comment that the value is intentionally re-stamped. Purely cosmetic.

### Note (informational, no action)
- Behavioral change worth knowing: pre-fix `auto_now` bumped `updated_at` on *every* save; now partial saves that pass `update_fields` without `"updated_at"` (heartbeat/liveness writers, compaction counters) no longer freshen it. This is deliberate and documented. Freshness consumers use dedicated liveness fields (`last_heartbeat_at`, `last_tool_use_at`) or the in-memory activity tracker, and the main executor saves include `"updated_at"` explicitly, so meaningful activity still freshens the field. Verified no consumer regression in the 63-test suite.

Self-authored PR — posted as a comment (GitHub blocks self-approval); this is a substantive Approved verdict.

PR #1662, opened 2026-06-12
fix(#1614): gate own-progress fields on heartbeat freshness to prevent zombie session recovery block

### LINKED ISSUE #1614: session-liveness-check: ungated sticky own-progress fields let a 4-day-stale running session evade recovery (confirm root cause before fixing)
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/sdlc-1614.md

> **Context for the reader.** This repo runs a standalone **worker** process that executes "AgentSessions" (one per Telegram thread / coding task). A session record lives in Redis (via the Popoto ORM) and carries a `status` field — `running`, `pending`, `completed`, `killed`, etc. A scheduled health reflection called **`session-liveness-check`** runs every 5 minutes (`agent.agent_session_queue._agent_session_health_check`, defined in [`agent/session_health.py`](../blob/main/agent/session_health.py)) and is supposed to detect sessions that are wedged in `running` but no longer making progress, then recover them (re-queue to `pending` or finalize). The decision of whether a running session "is still making progress" is delegated to the helper **`_has_progress(entry)`** in the same file.
>
> This issue documents a **confirmed zombie session** plus a **demonstrable gap** in `_has_progress`. The exact causal path is *not yet fully confirmed* — see Open Questions. The fix touches a heavily-litigated function (issues #1226, #1356, #1537) where over-eager recovery has historically killed legitimately-running long-thinking sessions, so **avoiding false-positives is the primary constraint, not an afterthought.**

### Observed (confirmed)
A `teammate` session in the Cyndra project (`agent_session_id=74b3c8e84a92460d8eb1780a303a3056`, thread `tg_cyndra_-1003794218389_535`) sat in `status=running` for **~4 days** and was never recovered by `session-liveness-check`, despite the check running every 5 minutes (~1,100 ticks). It was only cleared by a manual `valor_session kill`.

Evidence from `valor_session inspect`:

| Field | Value | Meaning |
|-------|-------|---------|
| `status` | `running` | Worker still believed it was live |
| `claude_pid` | `19007` | **Dead** — no such OS process |
| `started_at` | `2026-06-05 11:03:19` | — |
| `last_heartbeat_at` | `2026-06-05 11:04:25` | Queue-layer heartbeat stopped ~74s after start |
| `last_stdout_at` | `2026-06-05 11:04:33` | No output since |
| `last_tool_use_at` / `last_turn_at` | `None` | SDK never produced a tool/turn event (`sdk_ever_output = False`) |
| `claude_session_uuid` | `70ec3f3e-…` | SDK authenticated before dying |

Impact: a stuck `running` record on a Telegram thread blocks fresh messages on that thread from spawning new work (they resume a dead session). It also silently consumes a worker-tracking slot.

### Demonstrable gap (confirmed by code reading)
`_has_progress` ([`agent/session_health.py:830-836`](../blob/main/agent/session_health.py#L830)) consults three **sticky own-progress fields** when `sdk_ever_output is False`:

```python
if not sdk_ever_output:
    if (entry.turn_count or 0) > 0:
        return True
    if bool((entry.log_path or "").strip()):
        return True
    if bool(entry.claude_session_uuid):   # <-- set on auth, never cleared
        return True
```

Unlike **sub-check B** (the fresh-heartbeat fast-path), which issue #1356 gated behind a no-output running-time budget (`NO_OUTPUT_BUDGET_SECONDS = 1800s`), **these own-progress fields are not gated by heartbeat-freshness or running-time at all.** `claude_session_uuid` is set the moment the SDK authenticates and is never cleared, so for any session that authenticated but never emitted a tool/turn event, `_has_progress` returns `True` **forever**, regardless of how stale the heartbeat is. The code itself flags this hazard in the delivery-guard comment at `session_health.py:1538`: *"`turn_count` and `claude_session_uuid` are sticky fields that permanently block the no_progress recovery path."*

The fix from #1356 plugged this for sub-check B but left the same wedge open one block down in the own-progress fields.

## Why this needs careful root-causing (the false-positive trap)

`_has_progress` returning `True` is only **one** of the two conditions that gate recovery in `_agent_session_health_check`. For a `running` session, recovery fires if **either**:

1. **`not worker_alive`** (the owning task/future is dead or absent from the worker's in-memory `_active_workers`) AND `running_seconds > AGENT_SESSION_HEALTH_MIN_RUNNING` (300s) — **this branch ignores `_has_progress` entirely**; or
2. **`worker_alive`** but `_has_progress(entry) is False` AND all Tier-2 reprieve gates also fail.

The `_has_progress` gap only explains the zombie if the session fell into branch **2** (worker task hung un-done for 4 days). But the queue-layer heartbeat loop ([`session_executor.py:1774`](../blob/main/agent/session_executor.py#L1774)) runs `while not task._task.done()` and writes `last_heartbeat_at` every 60s — and it **stopped** at 11:04:25. A stopped heartbeat loop implies the task finished, which should make `worker.done()` True → `worker_alive=False` → **branch 1** should have recovered the session independent of `_has_progress`.

So there is a contradiction that must be resolved before fixing:
- If branch 1 should have fired, the real defect is in the **worker-liveness / `_active_workers`** path (or recovery fired and failed to persist), and gating the sticky fields would be a **false fix** that adds false-positive risk without addressing the cause.
- If the task genuinely hung un-done (branch 2), the sticky-field gap is the cause and the gate is correct.

**A fix that gates the sticky fields without first confirming the branch could start over-recovering legitimately-running long-thinking sessions** — exactly the regression #1226/#1356 fought. That is the false-positive the requester is worried about.

## Definitions

|------|-----------|--------|
| `session-liveness-check` | Reflection (runs every 5 min) that recovers wedged `running`/`pending` sessions | `config/reflections.yaml`; `agent/session_health.py:1422` |
| `agent-session-cleanup` | A *different* reflection — deletes corrupted/unsaveable records and rebuilds indexes. **Not** responsible for this case. | `config/reflections.yaml`; `agent/agent_session_queue.py` |
| `_has_progress(entry)` | Helper returning `True` iff a running session shows any progress signal | `agent/session_health.py:661` |
| `sdk_ever_output` | `True` iff `last_tool_use_at` or `last_turn_at` was ever set (the SDK produced a structured tool/turn event) | `agent/session_health.py:744` |
| own-progress fields | `turn_count`, `log_path`, `claude_session_uuid` — sticky, set-once, never cleared | `agent/session_health.py:830` |
| sub-check B | Fresh-heartbeat fast-path, gated by `NO_OUTPUT_BUDGET_SECONDS` (1800s) since #1356 | `agent/session_health.py:767` |
| `worker_alive` | `_active_workers.get(worker_key)` is present and its future is not done | `agent/session_health.py` (RUNNING scan) |
| queue-layer heartbeat | `last_heartbeat_at`, written every 60s by `_heartbeat_loop` while the task is alive — independent of SDK output | `agent/session_executor.py:1774` |

## Prior art (the lineage this sits in — all CLOSED)

- **#1226** — `session-liveness-check` failed to detect hung SDK: Tier 1 heartbeat refreshed by watchdog regardless of progress.
- **#1356** — Tier-1 sub-check B passed forever for sessions that never produced SDK output (introduced `NO_OUTPUT_BUDGET_SECONDS`). **This issue is the same class of bug, one code block lower, in the own-progress fields that #1356 did not touch.**
- **#1537** — Liveness recovery requeued a hung session to `pending` without killing its subprocess, orphaning the worker slot.

## Open Questions (resolve before prescribing a fix — do NOT skip to the gate)

1. **Which branch did the zombie actually fall into?** Reproduce or trace: was `worker_alive` True (task hung un-done) or False (task done / absent from `_active_workers`)? This determines whether the sticky-field gap is the cause or a red herring.
2. **Why did `last_heartbeat_at` stop at +74s while `status` stayed `running`?** Did the `_heartbeat_loop` die independently (its `save` raised repeatedly), did the task hang, or did the owning worker process die and a *later* worker (PID 59631) simply never had this session in `_active_workers`?
3. **If branch 1 should have fired but didn't** — is there a `worker_key` collision (teammate sessions sharing a key, cf. the #944 dev/PM sharing note), a terminal-status guard misfire, or a recovery attempt that raised and failed to persist?
4. **Only if branch 2 is confirmed:** what is the correct gate for the own-progress fields? Candidate: require `last_heartbeat_at` fresher than some window (or apply the same `NO_OUTPUT_BUDGET_SECONDS` running-time gate as sub-check B) before honoring sticky `turn_count`/`log_path`/`claude_session_uuid`. Verify the queue-layer heartbeat stays fresh during genuine long-thinking turns so the gate cannot kill a live session — `_heartbeat_loop` writing every 60s independent of SDK output suggests it does, but this must be confirmed, including the case where `save()` is intermittently failing.
5. Should recovery in this path also kill `claude_pid` if still alive (per #1537), and finalize-vs-requeue based on `response_delivered_at`?

## Acceptance criteria

- [ ] Root cause is **confirmed** (which branch; why heartbeat stopped while status stayed `running`) and documented before any behavioral change.
- [ ] A session matching the observed profile (`running`, dead `claude_pid`, `sdk_ever_output=False`, sticky `claude_session_uuid` set, `last_heartbeat_at` stale by hours) is recovered within a bounded number of `session-liveness-check` ticks.
- [ ] **No regression in false-positive protection:** a legitimately-running long-thinking session (fresh queue-layer heartbeat, no tool/turn output yet, well past the startup grace window) is **not** recovered. This must be covered by an explicit test.
- [ ] Telemetry/log line emitted when a session is recovered via this path (mirroring the `tier1_falloff` counter pattern) so future occurrences are observable rather than silent.

## Test Impact (preliminary — do-plan to finalize)

- `tests/unit/test_session_health_inference_removed.py:150-157` — asserts `turn_count=5, claude_session_uuid="abc"` ⇒ `_has_progress` True. If sticky fields become heartbeat-gated, this likely needs UPDATE to set a fresh heartbeat (or to assert the new gated semantics).
- `tests/integration/test_session_heartbeat_progress.py` — multiple `_has_progress` assertions across heartbeat/own-progress combinations; audit for cases that assume ungated sticky fields. UPDATE candidates.
- `tests/unit/test_session_health_inference_removed.py:160-181` — fresh-heartbeat-no-stdout case (`claude_session_uuid=None`); should remain green and serves as the false-positive guard — keep.

## Downstream

This issue will be consumed by `/do-plan` to produce `docs/plans/{slug}.md`. **do-plan must treat the Open Questions as investigation tasks first** — the root cause is not confirmed, so the plan should not jump straight to "gate the sticky fields." Prescribing the gate prematurely risks reintroducing the over-recovery false-positives that #1226/#1356 were created to prevent.

## Recon Summary

_Synthesized from the confirmed observations, code reading, and prior-art lineage already documented above. This issue is an investigation — the Open Questions below are the unresolved buckets that /do-plan must turn into investigation tasks before prescribing a fix._

**Confirmed:**
- A `teammate` session (`agent_session_id=74b3c8e84a92460d8eb1780a303a3056`) sat in `status=running` for ~4 days and was never recovered by `session-liveness-check`; only a manual `valor_session kill` cleared it. (`valor_session inspect` evidence table above.)
- `_has_progress` (`agent/session_health.py:830-836`) consults three sticky own-progress fields (`turn_count`, `log_path`, `claude_session_uuid`) when `sdk_ever_output is False`, and — unlike sub-check B — these are NOT gated by heartbeat-freshness or running-time. `claude_session_uuid` is set on SDK auth and never cleared, so `_has_progress` can return `True` forever. (Confirmed by code reading; the hazard is even called out in the `session_health.py:1538` delivery-guard comment.)
- Recovery in `_agent_session_health_check` fires on EITHER branch 1 (`not worker_alive` AND `running_seconds > 300`, which ignores `_has_progress`) OR branch 2 (`worker_alive` AND `_has_progress is False` AND Tier-2 reprieve gates fail). The sticky-field gap only explains the zombie if it fell into branch 2.
- The queue-layer `_heartbeat_loop` (`session_executor.py:1774`) writes `last_heartbeat_at` every 60s while `not task._task.done()`; for the zombie it stopped at +74s, which (absent a heartbeat-save failure) implies the task finished → `worker.done()` True → branch 1 should have fired. This contradiction is the crux.

**Revised:**
- The issue title ("ungated sticky own-progress fields let a stale session evade recovery") states a *hypothesis*, not a confirmed root cause. Recon revises this to: the sticky-field gap is a *demonstrable code defect* but is only the *operative* cause if branch 2 is confirmed. Branch must be determined first.

**Pre-requisites (investigation tasks — must complete before any behavioral fix):**
- Determine which recovery branch the zombie fell into (`worker_alive` True vs False) by tracing `_active_workers` membership and `worker.done()` state for the observed profile. (Open Question 1)
- Explain why `last_heartbeat_at` stopped at +74s while `status` stayed `running`: heartbeat-loop death, task hang, or owning-worker-process death with a later worker that never had the session in `_active_workers`. (Open Question 2)
- If branch 1 should have fired but didn't, locate the worker-liveness defect (worker_key collision, terminal-status guard misfire, or recovery that raised and failed to persist). (Open Question 3)

**Dropped:**
- "Immediately gate the sticky fields" is dropped as a first move — prescribing the gate before confirming branch 2 risks reintroducing the over-recovery false-positives that #1226/#1356 were created to prevent. The gate is only correct conditional on branch-2 confirmation (Open Question 4).
- `agent-session-cleanup` is dropped as a relevant mechanism — it deletes corrupted/unsaveable records and is explicitly NOT responsible for this case.

tomcounsell:
## Summary
- Confirmed root cause: Branch 2 — worker stayed alive for 4 days, `_active_workers` held a not-done asyncio.Task. The `_heartbeat_loop` exited at +74s (task completed), but `complete_transcript` failed silently leaving `status='running'`. The ungated `claude_session_uuid` in `_has_progress` returned True indefinitely, blocking branch-2 recovery.
- Added heartbeat-freshness gate (`NO_OUTPUT_BUDGET_SECONDS` = 1800s) on the own-progress block in `_has_progress` — sessions with stale/absent heartbeats no longer block recovery via sticky `claude_session_uuid`
- Added narrow `zombie_uuid_no_output` telemetry counter in `_apply_recovery_transition` for sessions with `claude_session_uuid` set but no SDK output

## Changes
- `agent/session_health.py` — gate own-progress fields (L857-872) on `NO_OUTPUT_BUDGET_SECONDS` (1800s) freshness; add AC4 narrow `zombie_uuid_no_output` telemetry counter (L1185-1200)
- `tests/unit/test_session_health_inference_removed.py` — 3 new AC2/AC3 tests: zombie-profile recovery test, fresh-vs-stale own-progress PAIR
- `docs/features/bridge-worker-architecture.md` — document gated own-progress semantics, confirmed Branch 2 failure mode, and telemetry counter
- `docs/features/agent-session-health-monitor.md` — updated own-progress fields section with heartbeat-freshness gate and #1614 reference

## Testing
- [x] 24 targeted tests passing (`test_session_health_inference_removed.py` + `test_session_heartbeat_progress.py`)
- [x] AC2: `test_zombie_profile_is_not_progress` — zombie profile returns `_has_progress=False`
- [x] AC3 FRESH: `test_own_progress_gate_fresh_heartbeat_is_progress` — live session with fresh heartbeat returns True (no false-positive)
- [x] AC3 STALE: `test_own_progress_gate_stale_heartbeat_is_zombie` — stale heartbeat zombie returns False (gate flips the bug)
- [x] Pre-existing `test_fresh_heartbeat_no_stdout_yet_is_progress` sentinel still passes
- [x] Lint clean on changed files

## Documentation
- [x] `docs/features/bridge-worker-architecture.md` updated with "Own-progress fields are heartbeat-gated (#1614)" section
- [x] `docs/features/agent-session-health-monitor.md` updated with gate semantics
- [x] Inline comment added above gate citing #1614 and confirmed Branch 2 verdict
- [x] Delivery-guard comment updated (no longer claims "permanently block")

## Definition of Done
- [x] Built: Gate implemented in `agent/session_health.py`, AC4 telemetry counter wired
- [x] Tested: All 24 targeted tests pass, false-positive sentinel green
- [x] Documented: Session-health and bridge-worker-architecture docs updated
- [x] Quality: Lint and format checks pass on changed files
- [x] Investigation: Root cause confirmed in Spike Results + GitHub comment on #1614

Closes #1614


tomcounsell:
## Review: Changes Requested

**Branch:** `session/sdlc-1614` → `main`
**Plan:** `docs/plans/sdlc-1614.md` (on main)
**Scope:** Gate `_has_progress` own-progress fields on heartbeat freshness (Branch-2 zombie fix) + AC4 narrow telemetry counter.

The implementation is correct, minimal, and faithful to the heavily-litigated plan. The gate is on the right surface (L857-872, own-progress block only — sub-check B untouched), uses the correct `NO_OUTPUT_BUDGET_SECONDS` (1800s) window per spike-3, and the AC4 counter is gated on the real derived `last_tool_use_at`/`last_turn_at` expression (not the forbidden `entry.sdk_ever_output` attribute). AC1 root-cause verdict was posted to #1614 before the change. Targeted tests, lint, and format all pass.

**One blocker:** the change introduced a regression in a pre-existing test that the PR's Test Impact did not audit.

### Blockers

**File:** `tests/unit/test_health_check_recovery_finalization.py:343-348` (verified: read this file)
**Code:** `entry = self._make_entry(turn_count=5)` ... `assert _has_progress(entry) is True`
**Issue:** This test (`test_per_turn_fields_none_turn_count_set_returns_true`) constructs an entry with `turn_count=5` and the helper default `last_heartbeat_at=None`, asserting `_has_progress` returns `True` via the old ungated own-progress block. The new gate skips own-progress fields when the heartbeat is absent/stale, so it now returns `False`. **Verified by execution:** this test PASSES on `main` and FAILS on this branch — a genuine PR-introduced regression. The PR's Test Impact only audited `test_session_health_inference_removed.py` and `test_session_heartbeat_progress.py`; this third file has a `_has_progress` assertion encoding the exact semantics the gate changes and was missed. As a result the plan's own "Full suite" Verification check (`pytest tests/ -x -q`) fails.
**Severity:** blocker
**Fix:** Update the test to reflect the new gated semantics. Either add a fresh heartbeat to keep it asserting `True` (e.g. `self._make_entry(turn_count=5, last_heartbeat_at=_ago(30))` — own-progress now honored because heartbeat is fresh), or, to assert the new behavior, keep `last_heartbeat_at=None` and change the assertion to `is False` with a docstring noting the #1614 gate (an absent heartbeat with `turn_count` set is now treated as a potential zombie). The fresh-heartbeat variant is the better choice — it preserves the test's original intent (own-progress fields are honored for a live session) while conforming to the gate.

### Tech Debt
- None

### Nits

**File:** `agent/session_health.py:858-861` (verified: read this file)
**Code:** `_hb_own = getattr(entry, "last_heartbeat_at", None)` ... `if _hb_own is not None:` ... `_hb_own_aware = _hb_own if _hb_own.tzinfo else ...`
**Issue:** The new gate guards with `_hb_own is not None`, whereas sub-check B directly above uses `isinstance(hb, datetime)` (L785). If `last_heartbeat_at` were ever a non-None, non-datetime value, `.tzinfo`/`.replace()` would raise — `_has_progress` has no surrounding try/except, so it would propagate. In practice `last_heartbeat_at` is always a datetime-or-None, so this is not a live bug, but matching sub-check B's `isinstance` guard would be more defensive and consistent.
**Severity:** nit
**Fix:** Change `if _hb_own is not None:` to `if isinstance(_hb_own, datetime):` to mirror sub-check B.

### Legacy Cruft
- None. The delivery-guard comment (L1587-1593) was correctly updated from "permanently block" to the heartbeat-gated wording. No dead code, no half-migration.

### Plan / AC validation
- **AC1** (root cause confirmed + posted to #1614 before change): PASS — Branch-2 verdict comment on #1614 dated 2026-06-12.
- **AC2** (zombie profile → `_has_progress` False): PASS — `test_zombie_profile_is_not_progress`.
- **AC3** (fresh-vs-stale own-progress PAIR): PASS — both `test_own_progress_gate_fresh_heartbeat_is_progress` (True) and `test_own_progress_gate_stale_heartbeat_is_zombie` (False) present and passing.
- **AC4** (narrow `zombie_uuid_no_output` counter on real fields): PASS — counter wired at `_apply_recovery_transition`; forbidden `getattr(entry, 'sdk_ever_output')` string absent (grep exit 1).
- **Carried TOCTOU concern** (re-read `last_heartbeat_at` before CAS): not implemented, consistent with the cycle-3 critique classifying it as non-blocking (CAS-on-status makes recovery idempotent; race window is sub-ms after ≥1800s staleness). Accepted residual, not a blocker.

**Verdict:** Changes Requested — 1 blocker (untracked test regression), 1 nit. The fix is a one-line test update; the underlying production change is sound.

tomcounsell:
## Review: Approved

**Scope:** Gate the own-progress fields (`turn_count`, `log_path`, `claude_session_uuid`) in `_has_progress` on `last_heartbeat_at` freshness, so zombie sessions whose heartbeat loop silently exited no longer block branch-2 recovery indefinitely. Closes #1614.

### Verdict
Clean, well-scoped bugfix that directly addresses the confirmed Branch 2 failure mode without re-opening the false-positive risk that #1226/#1356/#1537 litigated. Approved.

### What I verified
- **Gate logic** (`agent/session_health.py:857-872`): own-progress fields are now only honoured when `sdk_ever_output=False` **and** `last_heartbeat_at` is within `NO_OUTPUT_BUDGET_SECONDS` (1800s). Stale/absent heartbeat falls through to Tier-2 → recovery. This matches the documented zombie profile (heartbeat stale by hours, uuid set) exactly.
- **Constant**: `NO_OUTPUT_BUDGET_SECONDS = MAX_NO_OUTPUT_REPRIEVES * HEARTBEAT_FRESHNESS_WINDOW = 1800` confirmed at `session_health.py:284`.
- **Window choice is correct**: using the 1800s budget (not the 90s `HEARTBEAT_FRESHNESS_WINDOW`) keeps the own-progress gate from expiring before sub-check B's no-output budget, so a legitimately-starting session is not killed early. Documented and intentional.
- **`sdk_ever_output` derivation** is identical between sub-check A/B and the new own-progress gate and the AC4 counter — no drift.
- **AC4 telemetry** (`session_health.py:1186-1202`): derives `sdk_ever_output` from real fields (not a non-existent attribute), increments `{project_key}:session-health:recoveries:zombie_uuid_no_output`, wrapped in try/except — non-fatal, off the recovery hot path's critical decision.
- **Delivery-guard comment** (L1587) correctly downgraded from "permanently block" to "block while heartbeat is fresh".
- **Tests**: 85 pass serially (`test_session_health_inference_removed.py` + `test_health_check_recovery_finalization.py`). New AC2/AC3 tests use realistic profiles (zombie / fresh-heartbeat / stale-heartbeat). Updated `test_per_turn_fields_none_turn_count_set_returns_true` now seeds a fresh heartbeat, consistent with the gate.
- **Format**: `ruff format --check` clean on all three changed source/test files.
- **Docs**: `agent-session-health-monitor.md` and `bridge-worker-architecture.md` both updated coherently with the gate semantics, window-constraint rationale, and telemetry counter.

### Blockers
- None

### Tech Debt
- None

### Nits
- `tests/unit/test_health_check_recovery_finalization.py:344` — the updated docstring says "own-progress, heartbeat-gated by #1614", but because `_make_entry` sets no `started_at`/`created_at`, the fresh `last_heartbeat_at=_ago(30)` actually returns True via sub-check B's "truly legacy session" leg (`session_health.py:805`) before reaching the own-progress block. The assertion is correct; only the path attribution in the docstring is imprecise. Purely cosmetic — safe to leave.

No UI files changed — visual-proof gate is a no-op.

PR #1713, opened 2026-06-16
fix(granite): canned-fallback diagnostic + catchup/reconciler persona resolution (#1708)

### LINKED ISSUE #1708: Granite delivers canned fallback (pm_no_user_message) — 'PM transcript read empty' churn + eng-vs-teammate persona on catchup re-ingest
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/granite_canned_fallback_persona_resolve.md

## Context

While fixing the granite startup-hang (commit 3a3ff1ab — namespaced prime commands weren't syncing to `~/.claude/commands/`, causing fleet-wide `startup_unresolved`), a live end-to-end test surfaced **two further, separate defects** that the startup hang had been masking. With priming now working, granite runs the full PM↔Dev loop but can still deliver a useless reply.

**Observed reply to a customer** (Cyndra Dev Team, msg 626): *"Your request was completed; a summary could not be generated."* — the container's `OPERATOR_TERMINAL_MESSAGE` wrap-up fallback. This falsely claims completion and answers nothing.

Session: `tg_cyndra_-1003794218389_623` (04:24 run), `exit_reason=pm_no_user_message`, `reached max_turns=10 without a [/complete]`.

## Finding 1 — "PM transcript read empty" turn churn (primary)

Across the steady-state loop the container logged, e.g.:
```
[granite-container] steady-state turn 8: PM transcript read empty; using unknown classification
[granite-container] steady-state turn 9: PM transcript read empty; using unknown classification
[granite-container] wrap-up guard: PM transcript read empty; using unknown classification
```
Each empty read → `_unknown_classification()` → `PM_COMPLIANCE_NUDGE` re-prompt → 10 turns burned → wrap-up guard ships the canned fallback.

**Key signal it's a tailer/path bug, not a silent model:** the PM *is* writing assistant text to its JSONL — recent cyndra worktree transcripts contain assistant entries (e.g. one with 42 `"type":"assistant"` entries). So `last_assistant_text(pm_transcript, baseline_text_count=...)` is returning empty despite content existing.

Suspects to investigate:
- Transcript **path** mismatch: `_transcript_path()` computes `~/.claude/projects/{cwd-slug}/{session_id}.jsonl` from `pty._session_id`. Verify the PTY's session_id matches the actual JSONL filename claude 2.1.178 writes (resume/uuid handling may have changed).
- The **content-identity baseline** (`text_bearing_count` / `baseline_text_count`) over-filtering — if the baseline is computed wrong, every new read looks "not new" → empty.
- Possible claude **2.1.178** transcript-format/location change (same upgrade that moved the trust/prime behavior).

## Finding 2 — catchup re-ingest runs as `eng` for a `teammate` chat

`Cyndra Dev Team` is `persona: teammate` in projects.json, but the catchup-re-ingested session resolved `engineer`:
```
Resolved persona for session=tg_cyndra_-1003794218389_623: engineer (source=prime-command)
```
The original live run (msg 620, 20:44) correctly resolved `teammate`. So the **re-ingest/resume path** loses the teammate persona and defaults to eng. An eng PM↔Dev work-loop is the wrong machine for a conversational CS question and compounds Finding 1's churn. (Partly surfaced via the re-ingest test method, but the persona-resolution discrepancy on re-ingest is real and worth confirming for the live path too.)

## Repro
1. Ensure granite startup works (commit 3a3ff1ab synced).
2. Re-ingest an old teammate-chat message via catchup (clear its dedup record + rewind the `LastProcessedRecord` cursor, restart bridge).
3. Observe: session runs as `eng`, steady-state turns log "PM transcript read empty", exits `pm_no_user_message`, delivers `OPERATOR_TERMINAL_MESSAGE`.

## Why investigation (not bug yet)
Finding 1's root cause (path vs baseline vs 2.1.178 format) is unconfirmed; Finding 2 needs verification on the live (non-re-ingest) path. Root-cause both before writing a concrete fix plan.

## Relevant code
- `agent/granite_container/container.py` — steady-state loop, `last_assistant_text`, wrap-up guard, `OPERATOR_TERMINAL_MESSAGE`
- `agent/granite_container/transcript_tailer.py` — `last_assistant_text`, `text_bearing_count`
- `agent/granite_container/container.py::_transcript_path` / `_capture_pty_identity`
- catchup/persona resolution: `bridge/catchup.py`, `agent/session_executor.py` (persona resolve)
---

## Recon Summary

Root-caused both findings against claude 2.1.178 and live on-disk transcripts (two parallel investigation agents). Key surprise: Finding 1's three named suspects (path / baseline / 2.1.178 format) are **NOT** the live cause in the standard production path — all three were verified correct against real transcripts. The actual failure mode is narrower.

### Finding 1 — "PM transcript read empty" churn — suspects eliminated, true cause isolated

- **Suspect A (path mismatch): NOT the live cause, but two latent bugs confirmed.** claude 2.1.178 honors `--session-id` (verified: `claude --session-id <uuid> -p` writes `<uuid>.jsonl`); the PTY passes it at `pty_driver.py:383-384`; a live granite worktree transcript (`fb682710-…`) has inner `sessionId == filename`. Path computation works end-to-end. **Latent bug 1:** `_needs_session_spawn` (`pty_pool.py:389-405`) omits `spec.pm_session_id`/`spec.dev_session_id` from its predicate — masked today only because `_harness_env` always has ≥2 keys so `spec.env` is always truthy. One empty-env session shape away from firing → prewarmed pair reused without session_id → claude auto-generates a UUID → `_transcript_path` returns wrong/None path → empty reads forever. **Latent bug 2:** `_transcript_path` (`container.py:284-297`) and `bridge_adapter.py:92-100` use raw `cwd.replace("/","-")`; claude resolves symlinks before computing the slug, so any `working_directory` crossing a symlink diverges. Current projects don't, so latent.
- **Suspect B (baseline over-filtering): NOT a factor.** Baseline captured at the correct moment (`container.py:899` before `_cycle_idle` at `:902`); `_text_bearing_assistant_texts` extracts text correctly from real 2.1.178 data (verified: 44 assistant → 11 text-bearing on a real transcript). Returns empty only when pointed at a None/missing/wrong file.
- **Suspect C (2.1.178 format change): NOT a factor.** Location unchanged (`~/.claude/projects/{slug}/{uuid}.jsonl`); schema unchanged (assistant text still at `message.content[].type=="text"`). New UUID-named dirs are subagent sidecars, not relocations.
- **Confirmed root cause:** the Container is reading a **different file** than the one with content — `result.pm_transcript_path` is None / missing / wrong-slug — not a tailer or baseline defect. **Critical gap:** the log message `"PM transcript read empty"` conflates three distinct failure modes (path is None / file missing / file present but no new entry) that have completely different fixes. The single highest-value first step is a diagnostic that disambiguates which branch fires in the live cyndra session (`container.py:922-929`).

### Finding 2 — catchup re-ingest resolves `eng` for a `teammate` chat — FULLY CONFIRMED

- **Root cause (exact):** `bridge/catchup.py:235-246` calls `enqueue_agent_session(...)` **omitting `session_type` and `project_config`**. `enqueue_agent_session` defaults `session_type: str = SessionType.ENG` (`agent/agent_session_queue.py:1083`). The executor blindly trusts the persisted type: `_session_type = getattr(agent_session, "session_type", None)` (`session_executor.py:1009`) → `_resolve_compose_args` keys first on `session_type` → ENG branch returns `(ENGINEER, WORKER)` (`sdk_client.py:1143-1144`) → logged `engineer (source=prime-command)`.
- **Why live works:** `telegram_bridge.py:2205-2209` runs `resolve_persona(project, chat_title, is_dm)` → maps TEAMMATE → `SessionType.TEAMMATE`, forwards `session_type=` (`:2394`) + `project_config=project` (`:2395`).
- **Scope:** the **reconciler** (`bridge/reconciler.py:211-222`) has the **identical omission** — same bug, same fix. The **live reply/resume path is NOT at risk** (always resolves persona; resumes re-read the already-correct persisted type). Bug is strictly the two scanner paths that call `enqueue_agent_session` directly.
- **Fix locations:** `bridge/catchup.py:235-246` and `bridge/reconciler.py:211-222` must resolve persona → session_type and pass `session_type=` + `project_config=project`. Defense-in-depth: the silent `SessionType.ENG` default at `agent/agent_session_queue.py:1083` is what made the omission invisible.

### Disposition
- Finding 2: **CONFIRMED bug**, ready for a fix plan (catchup + reconciler).
- Finding 1: **CONFIRMED the named suspects are red herrings**; the live cause is a None/missing/wrong transcript path. Fix plan should lead with a **disambiguating diagnostic** (split the conflated log + WARNING on None/missing path), then close the two latent path bugs (`_needs_session_spawn` predicate; symlink-resolved slug).
### Recon Buckets

**Confirmed:**
- Finding 2 is a real bug: `bridge/catchup.py:235-246` and `bridge/reconciler.py:211-222` omit `session_type`/`project_config`, defaulting to `SessionType.ENG` (`agent/agent_session_queue.py:1083`); executor trusts persisted type (`session_executor.py:1009` → `sdk_client.py:1143-1144`).
- claude 2.1.178 honors `--session-id` and writes `{uuid}.jsonl` at the unchanged `~/.claude/projects/{slug}/` location with the unchanged `message.content[].type=="text"` schema (verified against live transcripts).
- Finding 1 live cause = Container reads a None/missing/wrong-slug transcript path, NOT a tailer/baseline/format defect.

**Revised:**
- Finding 1's three hypothesized suspects (path mismatch / baseline over-filtering / 2.1.178 format change) are all eliminated as the *live* cause — verified correct against real transcripts. The issue's framing of these as the likely causes is overturned by recon.
- The fix scope shifts from "fix the tailer/baseline" to "(a) add a diagnostic that disambiguates the three conflated empty-read branches, then (b) close two confirmed latent path bugs."

**Pre-requisites:**
- Live cyndra session diagnostic: split the conflated `"PM transcript read empty"` log (`container.py:922-929`) into path-None / file-missing / no-new-entry branches so the live failing branch is identified before hardening.

**Dropped:**
- No tailer (`transcript_tailer.py`) logic change — `_text_bearing_assistant_texts` / baseline counting verified correct against 2.1.178 and is not the defect.
- No change to the live reply/resume persona path — confirmed not at risk (always resolves persona; resumes re-read correct persisted type).

valorengels:
Closes #1708

## Summary

Fixes two confirmed defects the granite startup-hang (3a3ff1ab) had been masking, surfaced by a live cyndra end-to-end test where the container shipped the canned `OPERATOR_TERMINAL_MESSAGE` ("Your request was completed; a summary could not be generated").

### Finding 1 — "PM transcript read empty" churn (diagnostic-first)
- **Diagnostic split (lead change):** the single conflated `PM transcript read empty` log at all three read sites (prime-turn, steady-state, wrap-up guard) is split into three stably-named greppable branches — `transcript read: path-None` / `file-missing` / `no-new-entry`. Each WARNING also logs the resolved attempted path, pm/dev session-id presence, and live PTY session-id, landing in `logs/worker.log`.
- **Latent path hardening (correct-regardless, production no-op today):**
  - `_needs_session_spawn` now also forces a per-session spawn when a spec carries `pm_session_id`/`dev_session_id` even with empty env; docstring restated as the full per-session-identity invariant.
  - `_transcript_path` and `bridge_adapter` slug computation now `realpath`-resolve the cwd (only when truthy) before slugging, matching claude's symlink-resolved project-dir slug. None-guard precedes realpath.
- Reworded `OPERATOR_TERMINAL_MESSAGE` so it no longer falsely claims completion.

### Finding 2 — catchup/reconciler re-ingest ran as `eng` for a `teammate` chat
- New `persona_to_session_type` helper in `bridge/routing.py`; live handler refactored to call it.
- `bridge/catchup.py` and `bridge/reconciler.py` now resolve `resolve_persona(...)` → `persona_to_session_type(...)` and pass `session_type=` + `project_config=` to `enqueue_agent_session`, matching the live path. Wrapped in a narrow per-message try/except with a greppable WARNING that falls back to eng without aborting the scan.
- Defense-in-depth: `enqueue_agent_session` emits a greppable WARNING when both `session_type` and `project_config` are omitted (sentinel default), keeping the eng default for intentional callers.

## Note on Finding 1 fail-closed criteria
Per the plan, the two latent path fixes are admitted production no-ops today (recon eliminated all three originally-hypothesized live causes). The diagnostic is the gating deliverable — confirming which branch the live cyndra session hits (criteria 1a/1b) requires a live repro and is a verification-phase step. If the live branch is `no-new-entry`, that escalates to a follow-up issue with the now-disambiguated evidence.

## Tests
- `tests/unit/granite_container/` — 311 passed (diagnostic branches via caplog; predicate True for session-id + empty env; realpath slug with None-guard precedence)
- catchup + reconciler + routing — 77 passed (teammate→TEAMMATE, default→ENG on both scanner paths; failure-WARNING; enqueue-omission WARNING)

## Docs
- `docs/features/granite-pty-production.md` — three-way transcript-read diagnostic, `_needs_session_spawn` invariant, realpath-slug rule
- `docs/features/eng-session-architecture.md` — persona resolution on all ingest paths

### DIFF
--- a/agent/agent_session_queue.py
+++ b/agent/agent_session_queue.py
@@ -134,6 +134,12 @@ from models.agent_session import AgentSession
 
 logger = logging.getLogger(__name__)
 
+# Sentinel for enqueue_agent_session(session_type=...) so we can distinguish a
+# caller that explicitly passed SessionType.ENG (intentional eng) from one that
+# omitted the kwarg entirely (likely dropped persona resolution). The effective
+# default remains SessionType.ENG.
+_SESSION_TYPE_UNSET = object()
+
 # 4-tier priority ranking: lower number = higher priority
 PRIORITY_RANK = {"urgent": 0, "high": 1, "normal": 2, "low": 3}
 
@@ -1080,7 +1086,7 @@ async def enqueue_agent_session(
     scheduled_at: float | None = None,
     parent_agent_session_id: str | None = None,
     telegram_message_key: str | None = None,
-    session_type: str = SessionType.ENG,
+    session_type: str = _SESSION_TYPE_UNSET,  # type: ignore[assignment]
     scheduling_depth: int = 0,  # ignored, now derived
     project_config: dict | None = None,
     extra_context_overrides: dict | None = None,
@@ -1107,6 +1113,19 @@ async def enqueue_agent_session(
     Returns queue depth after push.
     """
     from tools.field_utils import log_large_field
+
+    # Greppable warning when a caller omits BOTH session_type and project_config:
+    # this is the signature of a scanner that dropped persona resolution and would
+    # silently default a teammate-configured chat to an eng PM<->Dev loop. The
+    # effective default stays SessionType.ENG; intentional eng callers that pass
+    # session_type=SessionType.ENG explicitly do not trip this.
+    if session_type is _SESSION_TYPE_UNSET and project_config is None:
+        logger.warning(
+            "[enqueue] session_type omitted AND project_config omitted; "
+            "defaulting to eng — caller may have dropped persona resolution"
+        )
+    if session_type is _SESSION_TYPE_UNSET:
+        session_type = SessionType.ENG
 
     log_large_field("message_text", message_text)
     if revival_context:
--- a/agent/granite_container/bridge_adapter.py
+++ b/agent/granite_container/bridge_adapter.py
@@ -95,6 +95,11 @@ def _transcript_path_from_spec(cwd: str, session_id: str) -> str:
     Claude Code names transcripts: ~/.claude/projects/{cwd.replace("/", "-")}/{uuid}.jsonl
     This is deterministic because we set the UUID via `claude --session-id`.
     """
+    # Resolve symlinks before slugging so the slug matches Claude Code's
+    # own realpath-based naming. Guard on truthiness: os.path.realpath("")
+    # returns the process CWD, which would silently corrupt the slug.
+    if cwd:
+        cwd = os.path.realpath(cwd)
     slug = cwd.replace("/", "-")
     home = os.path.expanduser("~")
     return os.path.join(home, ".claude", "projects", slug, f"{session_id}.jsonl")
--- a/agent/granite_container/container.py
+++ b/agent/granite_container/container.py
@@ -28,6 +28,7 @@ from __future__ import annotations
 
 import json
 import logging
+import os
 import re
 import shutil
 import subprocess
@@ -214,7 +215,9 @@ DEV_REPORT_UNAVAILABLE = "The developer did not produce a captured report."
 # Fallback user-visible message delivered directly (bypassing PM) when
 # the wrap-up guard exhausts MAX_WRAPUP_ATTEMPTS without PM emitting a
 # user-facing prefix. Guarantees the human always gets some message.
-OPERATOR_TERMINAL_MESSAGE = "Your request was completed; a summary could not be generated."
+OPERATOR_TERMINAL_MESSAGE = (
+    "I wasn't able to produce a response to this — please rephrase or follow up."
+)
 
 
 @dataclass
@@ -293,6 +296,11 @@ def _transcript_path(cwd: str, session_id: str | None) -> str | None:
     """
     if not session_id:
         return None
+    # Resolve symlinks before slugging so the slug matches Claude Code's
+    # own realpath-based naming. Guard on truthiness: os.path.realpath("")
+    # returns the process CWD, which would silently corrupt the slug.
+    if cwd:
+        cwd = os.path.realpath(cwd)
     cwd_slug = cwd.replace("/", "-")
     return str(Path.home() / ".claude" / "projects" / cwd_slug / f"{session_id}.jsonl")
 
@@ -330,6 +338,55 @@ def _truncate_exit_message(text: str) -> str:
     if len(text) <= EXIT_MESSAGE_MAX_CHARS:
         return text
     return text[: EXIT_MESSAGE_MAX_CHARS - 3] + "..."
+
+
+def _transcript_read_branch(pm_transcript: str | None) -> str:
+    """Classify why a PM transcript read produced no text into a greppable branch.
+
+    Returns one of three STABLE, greppable substrings:
+      - ``transcript read: path-None``     — the resolved path is None.
+      - ``transcript read: file-missing``  — path set but file absent on disk.
+      - ``transcript read: no-new-entry``  — file present but no new text-bearing
+        entry past baseline (valid file, PM emitted nothing this cycle).
+
+    These substrings are load-bearing for log-grep diagnostics; do not rename.
+    """
+    if not pm_transcript:
+        return "transcript read: path-None"
+    if not os.path.exists(pm_transcript):
+        return "transcript read: file-missing"
+    return "transcript read: no-new-entry"
+
+
+def _log_transcript_read_diagnostic(
+    site: str,
+    pm_transcript: str | None,
+    pm_pty: Any,
+    dev_pty: Any,
+) -> None:
+    """Emit a WARNING explaining why a PM transcript read came back empty.
+
+    `site` labels the read site (prime-turn / steady-state / wrap-up guard).
+    Logs the greppable branch substring plus the fully-resolved attempted
+    path, the presence of `spec.pm_session_id` / `spec.dev_session_id`
+    (sourced from each PTY driver's `_session_id`), and the live PM PTY's
+    `_session_id` — the trio needed to root-cause a path/slug or session-id
+    mismatch.
+    """
+    branch = _transcript_read_branch(pm_transcript)
+    pm_session_id = getattr(pm_pty, "_session_id", None)
+    dev_session_id = getattr(dev_pty, "_session_id", None)
+    logger.warning(
+        "[granite-container] %s: %s; path=%r "
+        "pm_session_id=%r dev_session_id=%r pty_session_id=%r; "
+        "using unknown classification",
+        site,
+        branch,
+        pm_transcript,
+        pm_session_id,
+        dev_session_id,
+        pm_session_id,
+    )
 
 
 def _unknown_classification() -> ClassificationResult:
@@ -848,9 +905,8 @@ class Container:
                 if pm_prime_text:
                     prime_classification = classify_pm_prefix(pm_prime_text)
                 else:
-                    logger.warning(
-                        "[granite-container] prime-turn: PM transcript read empty; "
-                        "using unknown classification"
+                    _log_transcript_read_diagnostic(
+                        "prime-turn", pm_transcript, self._pm_pty, self._dev_pty
                     )
                     result.transcript_fallback_count += 1
                     prime_classification = _unknown_classification()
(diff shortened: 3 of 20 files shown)


valorengels:
## Review: Approved

Reviewed against issue #1708, its recon summary, and `docs/plans/granite_canned_fallback_persona_resolve.md`. Checked out PR HEAD `3720a2f` in the existing worktree; all cited tests re-run green locally (granite_container 88, routing/reconciler/queue/pty_pool 78), ruff clean on all changed source.

### Verdict rationale
Both findings are implemented exactly as the recon scoped them, with the latent-path fixes correctly admitted as production no-ops today and the diagnostic as the gating deliverable. No blockers, no tech-debt in changed lines. Two advisory follow-ups noted below — both strictly outside this PR's recon-confirmed scope.

### Finding 1 (diagnostic + latent path hardening) — correct
- `_transcript_read_branch` cleanly splits the three conflated empty-read modes into stable greppable substrings; `_log_transcript_read_diagnostic` logs the resolved path + session-id trio. Matches the recon's "lead with a disambiguating diagnostic" directive.
- `_needs_session_spawn` predicate extension is wrapped in `bool(...)` so the `-> bool` annotation stays accurate despite the `or spec.pm_session_id` truthy-string operand. Good.
- `realpath`-before-slug in both `container._transcript_path` and `bridge_adapter._transcript_path_from_spec` is correctly guarded with `if cwd:` to dodge the `os.path.realpath("")`-returns-CWD footgun, and the None-guard precedes it in `_transcript_path`.
- `OPERATOR_TERMINAL_MESSAGE` reword no longer falsely claims completion.

### Finding 2 (persona resolution on scanner paths) — correct
- `persona_to_session_type` is a clean single source of truth; the live handler in `telegram_bridge.py` now routes through it while preserving the ENGINEER info-log. catchup + reconciler both resolve `resolve_persona(...) -> persona_to_session_type(...)` and forward `session_type=` + `project_config=project`, with a narrow per-message try/except that falls back to eng without aborting the scan. `chat_title`/`project` confirmed in scope at both call sites; `is_dm=False` is correct for group-only scanners.
- The `_SESSION_TYPE_UNSET` sentinel + omission WARNING in `enqueue_agent_session` is sound defense-in-depth; the effective `ENG` default is preserved.

### Follow-up notes (advisory — NOT blockers, out of scope for #1708)
- **`agent/session_revival.py:179`** enqueues a *new* session record on revival omitting both `session_type` and `project_config`, so it will now trip the new omission WARNING and could resolve a teammate chat to eng — the same class as Finding 2, but the recon explicitly bounded the fix to the catchup/reconciler scanner paths and the new WARNING is precisely the detector that will surface this in `logs/worker.log` if it ever fires for a teammate chat. Worth a separate follow-up issue if logs show it firing.
- **`bridge/update.py:73`** (the `/update fix` self-diagnostic) intentionally wants an eng session but omits both kwargs, so it will emit the new advisory WARNING on every update failure. Cosmetic log noise only; if it proves annoying, pass `session_type=SessionType.ENG` explicitly there to silence it.

### Other
- No UI files in the diff — visual-proof gate is a no-op.
- Note: an uncommitted local addition to `tests/integration/test_per_chat_catchup_cutoff.py` (84 lines, two teammate/eng catchup tests) exists in the worktree but is NOT part of this PR's committed diff. If intended for this PR, commit it; otherwise it's stray local work.

Approving — core implementation is correct, well-tested, and well-documented; the two notes are scope-bounded follow-ups, not defects in the delivered change.

Issue #1099, opened 2026-04-21
feat(resilience): harden claude -p subprocess against 4 known failure modes


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/session/harness-failure-hardening/docs/plans/harness-failure-hardening.md

> **Context:** This issue concerns the `claude -p` subprocess execution layer in the Valor AI system. "Valor" is an agentic coding assistant that runs `claude -p stream-json` as an `asyncio` subprocess ([`agent/sdk_client.py`](https://github.com/tomcounsell/ai/blob/main/agent/sdk_client.py)) to execute dev sessions. Sessions are managed as `AgentSession` records in Redis via a Popoto ORM model. The worker ([`worker/__main__.py`](https://github.com/tomcounsell/ai/blob/main/worker/__main__.py)) orchestrates execution; [`agent/session_health.py`](https://github.com/tomcounsell/ai/blob/main/agent/session_health.py) handles recovery; a two-tier no-progress detector (merged in [#1039](https://github.com/tomcounsell/ai/pull/1039)) governs stall detection. This analysis was prompted by the amux blog post ["Every way Claude Code crashes"](https://amux.io/blog/every-way-claude-code-crashes/).

## Problem

The `claude -p` subprocess has six known failure modes. Valor already handles two of them well (safety-prompt blocking via SDK permissions; state detection via stream-json protocol rather than ANSI parsing). Four modes are unhandled or mishandled — in each case, the session silently degrades or misreports its status rather than recovering cleanly.

**Current behavior:**
- **Thinking block corruption**: session returns `""` and is marked `completed` — the user gets silence and no error
- **Context fills silently**: `message_delta` token usage events are parsed but fully discarded; there is no observable signal before a session degrades
- **Post-compaction stall**: the inactivity watchdog fires and kills the session rather than nudging it to continue; compaction events are invisible to the health check
- **OOM / SIGKILL**: `returncode == -9` is logged but not stored; there is no mechanism to distinguish "OS killed" from "health check killed", so memory-pressure re-queues look identical to normal re-queues

**Desired outcome:**
Each failure mode has a minimal, targeted fix with a behavioral test that proves the recovery path fires correctly and does not trigger on healthy sessions.

## Definitions

| Term | Definition | Reference |
|------|-----------|-----------|
[...]

Issue #1539, opened 2026-06-01
[#1536 Pillar 2] Crash-signature library + auto-resume policy from session telemetry


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/crash_signature_auto_resume.md

> **Sub-issue of epic #1536.** Builds on the v1 telemetry recorder ([plan](https://github.com/tomcounsell/ai/blob/main/docs/plans/session_telemetry_recorder.md)) and the `valor-session resume` machinery (#1061).

## Problem

Runs "crash for no reason and must be resumed" — today that resume is manual (`./scripts/valor-service.sh worker-restart` or `valor-session resume`). We don't record what the event stream looked like just before a crash, nor which resume strategy recovered it, so the system can't self-recover.

## Scope

- Use v1's recorded traces to capture **crash signatures** — the terminal event sequence (e.g. `idle_gap` → `broken_pipe`, or a `status_transition` to `failed` with `kill_issued`/`subprocess_exited`) preceding a crash/hang.
- Record which resume strategy recovered each crash and its outcome.
- Propose (then optionally auto-apply) a resume policy so common crash signatures self-recover instead of waiting for a human.

## Dependencies / ordering

- **Depends on v1 (#1536)** for the trace data and especially the `status_transition` event with kill-outcome.
- **Coordinates with #1537** (liveness recovery blind-spot fix) — auto-resume must not fight the recovery path; the two need a clear ownership boundary.
- Builds on #1061 (`valor-session resume` for killed/failed sessions).

## Out of scope
- The healthy-vs-stalled classifier (pillar 1) and human-behavior capture (pillar 3).

## Related
- Epic #1536; #1537, #1061, #1271 (orphan reaping), #1311 (watchdog recovery)

## Recon Summary

Reconnaissance performed at plan time (baseline commit `11ceb581`). All dependencies cited in the issue have shipped since filing (2026-06-01); the v1 telemetry recorder plan was archived post-merge, but its substrate exists in code. Full ownership boundary with #1537 worked out below.

**Confirmed:**
- V1 telemetry recorder is live — `agent/session_telemetry.py` (shipped commit `415e0e10` / PR #1699). Append-only JSONL per session at `logs/session_telemetry/{session_id}.jsonl`, read via `read_session_timeline(session_id)`. Event types: `turn_start`, `turn_end`, `tool_use`, `token_usage`, `idle_gap`, `status_transition`, `telemetry_truncated`, `unknown`.
[...]

Issue #1538, opened 2026-06-01
[#1536 Pillar 1] Learn healthy-vs-stalled from recorded session telemetry (advisory)


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/stalled_session_advisory_classifier.md

> **Sub-issue of epic #1536.** Builds directly on the v1 telemetry recorder ([plan](https://github.com/tomcounsell/ai/blob/main/docs/plans/session_telemetry_recorder.md)). Do not start until v1 ships — this consumes the recorded event stream.

## Problem

V1 (#1536) *records* per-session telemetry (token usage, tool-call boundaries, turn events, idle gaps, status transitions) to `logs/session_telemetry/{session_id}.jsonl`, plus there are rich per-project Redis health counters (recoveries/reprieves/kills/tool_timeouts) that **nothing currently learns from**. This pillar closes that loop: learn a model of "healthy progress vs stalled" from the recorded stream so the system can flag a likely-stalled session the way a human reads a TUI.

**Hard constraint:** This AUGMENTS the existing evidence-based two-tier liveness check (`agent/session_health.py`) — it must NOT replace it, and must NOT resurrect stdout-silence as a kill signal (#1172). A learned "looks stalled" signal is an *advisory input* to recovery, never a sole kill trigger.

## Scope

- Consume v1's JSONL traces + existing Redis health counters as features.
- Learn / heuristically derive a "healthy vs stalled" classification over a session's recent event window.
- Surface the signal (advisory) — e.g. a dashboard indicator or a reflection that annotates suspect sessions.
- Consume PR #1487's synthetic `timeout`/`decode_error`/`broken_pipe` events as strong stall signals rather than reinventing detection.

## Out of scope

- New kill/recovery logic (that's the liveness check's job; this only advises).
- Crash/resume (pillar 2, separate sub-issue) and human-behavior capture (pillar 3).

## Related
- Epic #1536; v1 plan `docs/plans/session_telemetry_recorder.md`
- #1226/#1356 (two-tier liveness — the system this augments), #1172 (silence≠kill constraint), #1537 (recovery blind-spot)
- `docs/features/subconscious-memory.md` (candidate learning substrate)

## Recon Summary

_Reconnaissance performed 2026-06-17 against main `514b2cf5`. Pillar 1 CONSUMES the shipped v1 recorder; it adds an advisory live classifier and surfaces it. No kill/recovery logic._

**Confirmed:**
[...]

PR #1717, opened 2026-06-16
feat(granite): startup fast diagnostic -- plateau/oscillation detection + alert (#1710)

### LINKED ISSUE #1710: Granite startup-failure fast diagnostic: capture the unresolved frame and alert loudly instead of silently burning the 600s ceiling
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/granite_startup_fast_diagnostic.md

> **Context.** The *granite PTY container* (`agent/granite_container/container.py`) is the component that runs an interactive `claude` TUI in a pseudo-terminal to execute agent sessions (see [`docs/features/granite-pty-production.md`](../blob/main/docs/features/granite-pty-production.md)). On startup it primes two PTYs (PM + Dev) and loops until both reach an idle state before accepting work. If they never settle within a hard wall-clock ceiling (`STARTUP_HARD_CEILING_S = 600.0`, `container.py:105`), the run exits with the `startup_unresolved` signature.

## Problem

When granite startup fails to resolve, the failure is **silent and slow to diagnose**, even though a distinct `startup_unresolved` signature already exists.

**Current behavior (from the 2026-06-16 cyndra outage; corrected root-cause in #1709):** every cyndra session hung at startup with `Unknown command: /granite:prime-pm-role` — caused by the non-recursive `_sync_commands` glob (fixed in `3a3ff1ab`) plus the PR #1694 priming refactor that moved persona delivery onto namespaced prime slash commands. The container sat logging `container: startup cycle=N pm_idle=... dev_idle=... response=None` (`container.py:785`) for ~59 cycles, burned the full 600s ceiling, exited `startup_unresolved`, produced zero output, and delivered no reply. The only external tell was a buried log line; customers got silence. Root-causing it took an ~8-hour zombie window plus a deep multi-step investigation.

Startup can fail unresolved for **multiple distinct reasons** — the command-sync bug above, a `PM transcript read empty` classification (tracked separately in #1708), and plausibly others (e.g. a reworded/changed TUI prompt the parser doesn't recognize). The common thread is not the cause; it's that **the failure moment carries no loud, self-describing diagnostic**.

**Desired outcome:** a startup that fails to resolve should be *loud and self-diagnosing at failure time* — surfacing the exact stuck state to a human in one alert, and not silently consuming the full ceiling.

## Why this is distinct from neighbouring issues

| Issue | Concern | This issue's relation |
|---|---|---|
| #1538 (epic #1536 pillar 1) | Record startup/session events for **offline learning** (advisory classifier) | Sibling — consumes the same synthetic event; this one fires the *immediate* alert |
| #1539 (epic #1536 pillar 2) | Auto-**resume** policy from crash signatures | Sibling — this is detection/diagnostic only, not recovery |
| #1708 | A *specific* startup failure (`pm_no_user_message` / transcript-empty canned fallback) | Sibling — that fixes one cause's behavior; this makes *any* unresolved startup loud |
| #1648 (closed) | Dashboard telemetry hollow for granite PTY sessions | Prior art on granite observability gaps |

This issue is **detection + diagnostic + alert at failure time**, root-cause-agnostic.

## Definitions

| Term | Meaning |
|---|---|
| Granite PTY container | `agent/granite_container/container.py` — runs interactive `claude` in a PTY to execute sessions |
| `startup_unresolved` | Existing exit signature when neither PTY settles within `STARTUP_HARD_CEILING_S` (`container.py:105`) |
| Startup loop | `container.py:780+` — polls each PTY for idle each cycle (`STARTUP_CYCLE_TIMEOUT_S = 3.0`) until the deadline |
| Frame | A snapshot of the rendered TUI / read transcript buffer at a point in the startup loop |

## Solution sketch (direction, not a plan)

When the startup loop fails to reach a usable state, the container should:
1. **Detect the stuck-state plateau early** — e.g. N consecutive identical no-progress cycles (`pm_idle/dev_idle/response` unchanged) — and bail rather than always grinding to the 600s ceiling.
2. **Capture the offending frame** — the PM/Dev buffer/transcript snapshot at the point of failure — as a diagnostic artifact. *The captured frame is the diagnosis* (it would have named "Unknown command: /granite:prime-pm-role" immediately).
3. **Emit it loudly** — a synthetic `startup_unresolved` event carrying the captured frame (the same event #1538 will consume) **plus** a human-reachable notification, not just a log line.
4. **Bail fast** with the captured artifact attached to the session record.

## Open questions (resolve at `/do-plan`)

- Does any plateau/early-bail logic already exist around `STARTUP_HARD_CEILING_S` / `STARTUP_CYCLE_TIMEOUT_S`? The comments at `container.py:92-127` hint at prior consideration — confirm what's already there before adding.
- What is the human-reachable alert channel for an infra-level startup failure (the broken session can't speak for itself)? Bridge log + dashboard + a direct notification?
- What exactly should the captured frame contain (raw bytes vs stripped transcript vs both), and how is it stored/retained without bloating session records?
- How does this coordinate with #1539 so the alert fires *before* (or instead of) a resume attempt on a deterministic never-started failure?

## Downstream context

This issue will be consumed by `/do-plan` to produce a plan at `docs/plans/{slug}.md`. The root cause is partly structural (multiple failure modes, existing-code interaction), so the section above intentionally lists open questions rather than a fixed approach.
## Recon Summary

_Reconnaissance performed at plan time (2026-06-16, baseline main @ 1213936d). Verified against the live granite container code._

**Confirmed:**
- No plateau/early-bail logic exists. The startup loop (`container.py:837-875`) polls `while time.monotonic() < startup_deadline` and only breaks when both PTYs are idle; otherwise it grinds to the full 600s ceiling. The comments at `container.py:92-127` discuss *why the ceiling is long* (cold-Opus persona load), not any early-exit heuristic. Confirmed by test `tests/unit/granite_container/test_container.py::test_never_idle_exits_startup_unresolved_at_ceiling`.
- The offending frame IS available but discarded. Each cycle has both PM and Dev buffers in hand: `pm_idle = self._startup_cycle_idle(self._pm_pty)` and `dev_idle = ...` return `(saw_idle, buffer, idle_marker, elapsed_ms)` (`container.py:838-839`, `669-680`). On `startup_unresolved` exit only `{"cycle", "response"}` is appended to `result.startup_events` (`container.py:867`); the buffers are dropped.
- An exit-anomaly alert path already exists: `bridge_adapter._maybe_publish_exit_anomaly` (`bridge_adapter.py:409-453`) already treats `startup_unresolved` as a hard exit → `logger.error("[granite-exit-anomaly] ...")` (Sentry-captured) + appends an `exit_anomaly` session_event. This is the hook #1710 extends — it currently carries no frame.
- Established human-reachable infra-alert pattern: `reflections/sdlc_progress.py::_send_alert` (`:206-221`) shells out to `valor-telegram send --chat "Eng: Valor"` with fail-silent error handling. This is the precedent for a direct notification.
- AgentSession already carries granite diagnostic fields persisted post-run by `_publish_exit_summary` (`bridge_adapter.py`): `exit_reason`, `pm_pid`, `dev_pid`, `pm_transcript_path`, `dev_transcript_path`, `pty_slot` (`models/agent_session.py:300-322`). Adding a nullable diagnostic field follows the same additive pattern (`_heal_descriptor_pollution` handles back-compat generically).

**Revised:**
- "Emit a synthetic `startup_unresolved` event carrying the frame (the same event #1538 will consume)" — #1538's full telemetry recorder/event substrate (epic #1536 v1, `session_telemetry_recorder.md`) is upstream and not yet merged. #1710 should NOT block on it. The frame is carried on the **already-emitted `exit_anomaly` session_event + a new ContainerResult field**, which #1538's recorder can later read. Coordinate via shared field, not shared code.

**Pre-requisites:** None blocking. #1708 (closed, PR #1713) already fixed one specific cause; #1710 is cause-agnostic and independent.

**Dropped / out of scope:**
- Auto-resume (owned by #1539) — #1710 is detection + diagnostic + alert only. Coordination point: the alert must fire before any resume on a deterministic never-started failure; a no-resume hint on the diagnostic is the lightweight coordination surface.
- Fixing any specific root cause (owned by #1708 and future issues).
- The full session-telemetry event recorder (owned by #1536/#1538).

valorengels:
## Summary
- Adds STARTUP_PLATEAU_CYCLES = 10 constant and plateau detection in startup loop: bails early after N consecutive identical response=None cycles with no PTY idle (silent-start sentinel). Fingerprint keyed on response ALONE (not idle bools) so oscillating events do not break accumulation.
- Widens _startup_cycle_idle to 5-tuple (saw_idle, edge_buffer, level_tail, idle_marker, elapsed_ms) -- edge_buffer feeds the parser, level_tail used only for frame capture.
- Adds _capture_startup_frame pure helper: strips printable text, size-caps each buffer and total, falls back from level_tail to edge_buffer so frame is never blank.
- Adds nullable startup_failure_kind, startup_diagnostic_frame, startup_plateau_cycles to ContainerResult.
- Adds nullable startup_failure_kind, startup_captured_frame to AgentSession (no migration per #1099/#1172).
- Adds _should_alert two-layer cooldown (process-local monotonic first, Redis TTL second) and _send_startup_alert best-effort notification with timeout=3, check=False, swallows FileNotFoundError/TimeoutExpired with [granite-alert-suppressed] ERROR. Redis-down path still sends and does NOT log suppression tag.
- Enriches exit_anomaly event for startup_unresolved with startup_failure_kind and frame excerpt.

## Test plan
- [ ] 32 new tests in tests/unit/test_granite_startup_diagnostic.py all pass
- [ ] test_never_idle_exits_startup_unresolved_at_ceiling updated to patch STARTUP_PLATEAU_CYCLES=10M so ceiling path is testable independently
- [ ] Full granite container suite: 311 passed, 5 skipped
- [ ] ruff format + ruff check clean on all modified files
- [ ] Plateau detection keyed on response only (not idle bools) -- concern 1 addressed
- [ ] Frame capture falls back to edge_buffer if level_tail is blank -- concern 2 addressed
- [ ] [granite-alert-suppressed] fires on cooldown/send-fail but NOT on Redis-down path

Closes #1710


valorengels:
## Review: Approved

**Branch:** `session/granite_startup_fast_diagnostic` -> `main`
**Plan:** `docs/plans/granite_startup_fast_diagnostic.md`
**Result:** Approved (self-authored PR -- posted as comment; GitHub forbids self-approval)
**HEAD:** d2d5761b38db4b119ef79015b7ba857dc204f8ea

Verified against the plan's acceptance criteria and No-Gos. No blockers, no tech debt, no nits.

### Blockers
- None

### Tech Debt
- None

### Nits
- None

### What was verified
- **Plateau detection** (`container.py:908-980`): fingerprint keyed on parser `response` ALONE; bail requires `_plateau_count >= STARTUP_PLATEAU_CYCLES` AND the explicit `_silent_start` sentinel (response is None and neither PTY idle). Oscillating non-None events correctly fall through to the ceiling path rather than fast-bailing -- confirmed by `test_oscillating_event_plateaus` (asserts `startup_failure_kind=ceiling`, not plateau). Internally consistent with the constant comment and PR description.
- **Frame capture** (`_capture_startup_frame`, `container.py:365-407`): pure helper, strips non-printable bytes, per-buffer (`_STARTUP_FRAME_BUF_CAP=4000`) and total (`_STARTUP_FRAME_TOTAL_CAP=6000`) caps, never blank (kind+cycles header always present), falls back level_tail -> edge_buffer.
- **Two-layer cooldown** (`_should_alert`, `bridge_adapter.py:110-153`): process-local monotonic gate checked first (Redis-down resilient), Redis SET NX EX second; Redis-down path sends anyway and does NOT log the suppression tag (inverted contract honored). `_send_startup_alert` uses `timeout=3`, `check=False`, swallows FileNotFoundError/TimeoutExpired/Exception with `[granite-alert-suppressed]` ERROR.
- **exit_anomaly enrichment** (`bridge_adapter.py:590-613`): startup_unresolved events carry `startup_failure_kind` + 1KB frame excerpt; non-startup_unresolved anomalies do not (confirmed by test).
- **AgentSession fields** (`models/agent_session.py:324-336`): `startup_failure_kind`, `startup_captured_frame` nullable, no migration needed (#1099/#1172 generic healing).
- **No-Gos respected**: no auto-resume (#1539), no telemetry recorder (#1538), no cross-machine aggregation -- only the in-scope per-machine cooldown.
- **Tests**: 32 new tests pass; full granite suite 311 passed / 5 skipped (matches PR claim). `ruff check` + `ruff format --check` clean on all changed files. grep criteria confirmed (bridge_adapter references `_send_startup_alert`, container references `_capture_startup_frame`).

valorengels:
Added a router fix to this PR (commit 631818b6): `guard_g5_artifact_hash_cache` (G5) was short-circuiting to `/do-build` whenever the cached CRITIQUE verdict was READY TO BUILD on an unchanged plan hash, with no check for whether BUILD had already completed or a PR existed. Because G5 runs before the dispatch-table rows, the D3 deferral in rows 4a/4b/4c never ran — the pipeline re-dispatched `/do-build` forever on a finished PR. Surfaced live while supervising this issue's SDLC run (after this PR was open + REVIEW APPROVED, `next-skill` kept returning `/do-build`). G5 now defers to the downstream PR-stage rows once a PR exists or BUILD is completed. 3 new unit tests added; full router suite (21 tests) green.

Issue #1536, opened 2026-06-01
[EPIC] Record and learn from Claude Code session telemetry (TUI-style observability)


valorengels:
**Plan (v1):** https://github.com/tomcounsell/ai/blob/main/docs/plans/session_telemetry_recorder.md

> **What this is.** An epic to make Claude Code sessions *observable and learnable*: record the per-event telemetry stream that the worker currently parses-and-discards, then build a learning layer on top of it. The motivating intuition (from a real incident, below) is that a human watching a terminal Claude Code TUI can tell "stuck" from "still working" at a glance — token counts ticking, tools scrolling, the spinner — while our headless worker keeps no comparable record and so can neither diagnose nor learn from session behavior after the fact.
>
> **Scope note.** This is an umbrella epic with three pillars (§Pillars). It is deliberately large. The **minimum viable v1** is carved out explicitly in §V1 (MVP) — a single foundational deliverable (a durable per-event telemetry recorder) that everything else builds on. Pillars 2–3 and the "learning" half of pillar 1 are out of scope for v1.

## Background & motivating incident

On **2026-05-31**, a `cyndra` PM session (`tg_cyndra_-1003900483201_61`) spawned a `claude -p` subprocess that **hung for ~25.5 hours emitting zero output** (`agent/messenger.py` logged `SDK heartbeat: running 91620s, communicated=False` once per minute). It wedged the worker's single execution slot, so three queued sessions never started. The worker process stayed alive and `worker-status` reported `RUNNING`, because the **reflection scheduler runs on a separate thread that kept ticking** — the only external tell was the worker heartbeat age climbing 1:1 with wall-clock. Recovery required a manual `./scripts/valor-service.sh worker-restart`.

Crucially, **we have no event-level record of what that session was doing** in those 25.5 hours. The stream-json protocol from the harness is parsed for token counts and then discarded (`agent/sdk_client.py:286-391`). After the fact, there was nothing to inspect to explain *why* the existing liveness machinery didn't recover it.

## What already exists (do NOT rebuild)

Reconnaissance found that **hang detection is already sophisticated and deliberately evidence-based** — this epic must build on it, not duplicate or regress it:

| Capability | Where | Status |
|---|---|---|
| Two-tier liveness check (per-turn SDK signals + queue-heartbeat fallback, gated by a 30-min no-output budget) | `agent/session_health.py:659-943` | EXISTS (#1226, #1356) |
[...]

Issue #1537, opened 2026-06-01
Liveness recovery requeues hung session to pending without killing its subprocess — orphan wedges worker slot


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/sdlc-1537-liveness-orphan-subprocess-kill.md

> **Confirmed via code-read spike during planning of #1536.** This is the root cause of the 2026-05-31 incident where a `claude -p` subprocess hung 25.5h and wedged the worker, requiring a manual `worker-restart`.

## Problem

When the session-liveness check recovers a no-progress `running` session, it requeues the DB record to `pending` and calls `task.cancel()` — but **never confirms the underlying `claude -p` subprocess actually exited**. If the subprocess ignores cancellation (a true hang), it becomes an orphan that **no detector tracks**, occupying the worker's execution slot indefinitely.

### The blind spot (three detectors, one crack)

1. **Forward health check** queries only `status="running"` (`agent/session_health.py:1304`, `:1436` via `_filter_hydrated_sessions(AgentSession.query.filter(status="running"))`). After recovery the session is `pending`, so the forward scan never re-examines it.
2. **Recovery transition** nulls `started_at` and sets status `pending` (`agent/session_health.py:1154-1191`), then `task.cancel()` with `TASK_CANCEL_TIMEOUT` ≈ 0.25s (`:1086-1105`). No SIGTERM-to-PID fallback if cancel times out; no verification the process is gone.
3. **In-process orphan reaper** only acts on sessions whose DB status is in `_TERMINAL_STATUSES` (`agent/session_health.py:1555-1558`). A `pending` session is non-terminal → skipped.
4. **Cross-process reaper (#1271)** only reaps `claude` processes with `PPID==1` (worker died → launchd reparents). Here the worker stayed alive (PPID≠1), so the orphan was out of scope (`agent/session_health.py:38-62`).

Net: a hung subprocess at `pending` status with a live parent is invisible to every detector.

### Secondary: no-output budget clock restart

After recovery nulls `started_at`, the next `_has_progress` evaluation falls back to `created_at` for `running_seconds` (`agent/session_health.py:780-805`). Combined with the session no longer being `running`, the `NO_OUTPUT_BUDGET_SECONDS` (1800s) gate never re-fires.

**Current behavior:** A hung `claude -p` survives recovery, orphaned at `pending`, until a human runs `./scripts/valor-service.sh worker-restart`.

[...]

PR #1718, opened 2026-06-17
feat(#1539): crash-signature library + auto-resume policy from session telemetry

### LINKED ISSUE #1539: [#1536 Pillar 2] Crash-signature library + auto-resume policy from session telemetry

valorengels:
Closes #1539. Pillar 2 of the session-telemetry learning pipeline (epic #1536).

## Summary
- **`agent/crash_signature.py`** — pure extractor reducing terminal session telemetry traces to a normalized `CrashSignatureKey` (human_form + `sha256[:16]` hash). Determinism guardrail blocks auto-resume for plateau startup failures and never-started sessions (no `turn_start` events), marking them `NON_RESUMABLE_DETERMINISTIC` and routing to escalation.
- **`models/crash_signature.py`** — Popoto-backed signature library: occurrence counts, per-strategy outcome tallies, resumable flag, project-scoped policy derivation.
- **`reflections/crash_recovery.py`** — periodic reflection: extracts signatures from recently-terminal sessions, attributes outcomes (flag-first write ordering to avoid double-count), proposes auto-resumes in default mode, executes them only when `CRASH_AUTORESUME_ENABLED=1` (gated to one designated machine).
- **Resume infra** — `RESUMABLE_STATUSES` in `session_lifecycle.py`; `resume_session()` shared core + `crash-signatures` / `crash-policy list` CLI subcommands in `valor_session.py`; new nullable `AgentSession` fields (`crash_signature`, `crash_outcome_attributed`, `auto_resume_attempts`).

## Safety gates
Per-session `auto_resume_attempts` cap (default 3), per-run budget (default 5), terminal-only pre-transition status re-check. Off by default — propose-only on every machine except the one designated host.

## Tests
122 unit + 5 integration tests (extractor, library, CLI, resume/hydration, determinism guardrail, attribution idempotency).

## Provenance note
The original build (`/do-build`) committed directly to `main` (commit `8e8fe24b`) without worktree isolation and swept a stale dirty tree of unrelated granite/persona changes — including **deletions of shipped persona-as-priming files** — into the feature commit. `main` was reset to the clean pre-build state (`7a4dbd39`) and this branch re-lands **only** the #1539 changes (verified: diff vs `7a4dbd39` touches exactly the 18 #1539 files, zero granite/persona paths). See the #887 / dirty-tree-contamination failure mode.


valorengels:
## Review: Changes Requested

**Branch:** `session/crash_signature_auto_resume` -> `main`
**Plan:** `docs/plans/crash_signature_auto_resume.md` (#1539, epic #1536 Pillar 2)
**Preflight:** OPEN, MERGEABLE, mergeStateStatus=CLEAN. No UI files — screenshot gate is a no-op.

The architecture is sound and matches the plan: pure deterministic extractor, write-once attribution split (`crash_signature` + `crash_outcome_attributed`), `RESUMABLE_STATUSES` resolving the `abandoned` mismatch, flag-first attribution ordering, the determinism guardrail (`plateau` / no-`turn_start` -> `NON_RESUMABLE_DETERMINISTIC` -> escalate), and the naive-`updated_at` UTC normalization fix. Ruff clean; all 48 unit + 5 integration tests pass; no raw Redis on Popoto keys. But three findings — two of them blockers — mean the feature does not actually do what the plan promises on the live path.

### Blockers

**File:** `reflections/crash_recovery.py:287` (verified: read this file)
**Code:** `if auto_enabled and run_budget_remaining > 0:`
**Issue:** The auto-resume path never consults the policy-confidence gate. `_MIN_OCCURRENCES` (line 40) and `_MIN_SUCCESS_RATIO` (line 41) are read from env but never used in any comparison, and `CrashSignature.is_auto_eligible()` is never called anywhere in production code (`grep` confirms it appears only in its own definition + docstring + unit tests). As written, on the designated machine with `CRASH_AUTORESUME_ENABLED=1`, the FIRST occurrence of any resumable, classified crash signature triggers an auto-resume — with zero occurrence history and zero success-ratio confidence. This directly contradicts the plan's Data Flow step 6 ("a freshly-terminal session's signature has an **auto-eligible** policy entry") and defeats Risk 1's stated mitigation (a signature should only auto-resume after crossing `MIN_OCCURRENCES` + `MIN_SUCCESS_RATIO`, and auto-demote itself when the ratio drops). The safety substrate exists in the model but is not wired into the decision.
**Severity:** blocker
**Fix:** Before resuming, gate on the library record: look up `CrashSignature.get_by_hash(sig.hash)` and call `sig_record.is_auto_eligible("auto_resume", min_occurrences=_MIN_OCCURRENCES, min_success_ratio=_MIN_SUCCESS_RATIO)`; skip to propose/log when it returns False. Add a negative integration test asserting an UNDER-threshold signature is NOT transitioned to `pending` by the live reflection (see the misleading-test note below).

**File:** `tools/valor_session.py:1528` and `tools/valor_session.py:1569` (verified: read this file)
**Code:** `tallies = getattr(s, "outcome_tallies", None) or {}`
**Issue:** Both `cmd_crash_signatures` and `cmd_crash_policy` read `getattr(s, "outcome_tallies", ...)`, but the model has no such attribute — the field is `outcome_tallies_json` (a JSON string) accessed via `_load_tallies()`. So `tallies` is always `{}`; `attempts`, `recovered`, and `confidence` are always 0, and `auto_eligible` is always False, regardless of real recovery history. Verified empirically: a record with 3 occurrences and a 1.0 success ratio (`_load_tallies()` -> `{'auto_resume': {'attempts': 3, 'recovered': 3, 'failed': 0}}`, `is_auto_eligible()` -> True) reads through the CLI accessor as `None`. Secondary defect: even with the right field, the CLI uses a FLAT shape (`sum(tallies.values())`, `tallies.get("recovered")`) while the real structure is nested per-strategy (`{"auto_resume": {"attempts":..,"recovered":..}}`) — so the arithmetic is also wrong. This fails Success Criterion 3 ("render ... including which signatures are auto-eligible") and blinds the operator-facing propose-mode surface, which is the only window into the policy on every non-designated machine.
**Severity:** blocker
**Fix:** Read tallies via `s._load_tallies()` (or a public accessor) and aggregate across the nested per-strategy buckets, e.g. `attempts = sum(b.get("attempts",0) for b in tallies.values())`, `recovered = sum(b.get("recovered",0) for b in tallies.values())`. Better: have the CLI call `s.is_auto_eligible(...)` and `s.policy_confidence("auto_resume")` directly so it shares the model's single source of truth.

### Tech Debt

**File:** `tests/unit/test_crash_signature_cli.py:36` (verified: read this file)
**Code:** `s.outcome_tallies = outcome_tallies if outcome_tallies is not None else {"continue": 3}`
**Issue:** The CLI unit tests use a mock `_make_sig` that fabricates an `outcome_tallies` attribute in the FLAT shape the buggy CLI expects, rather than exercising a real `CrashSignature` Popoto record. That is why the suite stays green while the real CLI is broken end-to-end (Blocker 2). The integration test `test_auto_resume_fires_on_eligible_session` is similarly misleading: it carefully pre-warms the library to 3 occurrences + 1.0 ratio and sets the MIN env vars, but since the reflection never reads those thresholds, the test would pass identically with zero pre-warming — giving false confidence that the eligibility gate is enforced (Blocker 1).
**Severity:** tech_debt
**Fix:** Replace the CLI mock with a real `CrashSignature.get_or_create_by_hash(...)` + `record_outcome(...)` fixture so the test reflects the actual model shape; add a negative-eligibility assertion to the integration suite.

### Nits
- None

### Legacy Cruft
- None. The PR cleanly re-lands only the #1539 files atop `7a4dbd39`; the diff touches exactly the 18 declared files with no granite/persona contamination (provenance note in the PR body verified).

### Verdict
Architecture and safety design are strong and the determinism guardrail is well-implemented and tested. But the two blockers mean: (a) auto-apply fires without the confidence gate that is the plan's central safety promise, and (b) the operator inspection surfaces report empty stats and never flag eligibility. Both must be fixed — and a negative-eligibility test added — before merge. Routing to `/do-patch`.

valorengels:
## Review: Changes Requested

PR #1539 — crash-signature library + auto-resume policy. Branch is OPEN, MERGEABLE, CLEAN; provenance concern from the PR body verified resolved (diff vs `main` touches exactly the 18 #1539 files, zero granite/persona paths). All 128 targeted tests pass; ruff clean; no raw Redis in new code; the determinism guardrail, the write-once attribution split, and the CLI empty-case handling are all solid.

Two blockers prevent approval: the feature's **headline capability — auto-resume — cannot actually fire in production**, even though the plan made shipping it ENABLED on the designated machine a *binding* Success Criterion (Blocker 3). The tests pass only because they bypass both broken paths.

### Blockers

**File:** `reflections/crash_recovery.py:40-76` (verified: read this file)
**Code:** `auto_enabled = bool(int(os.environ.get("CRASH_AUTORESUME_ENABLED", "0") or "0"))`
**Issue:** Env-var wiring mismatch. The reflection reads bare `CRASH_AUTORESUME_ENABLED` (and `CRASH_AUTORESUME_MAX_ATTEMPTS`, `_RUN_BUDGET`, `_MIN_OCCURRENCES`, `_MIN_SUCCESS_RATIO`). But `.env.example`, `config/settings.py`, and `docs/features/crash-signature-auto-resume.md` all instruct the operator to set the `FEATURES__`-prefixed names (`FEATURES__CRASH_AUTORESUME_ENABLED`, etc.) — that prefix is the pydantic-settings `env_nested_delimiter="__"` mapping to `settings.features.crash_autoresume_enabled`. The two never connect. Proven:
```
$ FEATURES__CRASH_AUTORESUME_ENABLED=1 python -c "from config.settings import settings; import os; \
    print(settings.features.crash_autoresume_enabled, os.environ.get('CRASH_AUTORESUME_ENABLED'))"
True None
```
An operator who follows `.env.example` verbatim on the designated machine sets `FEATURES__CRASH_AUTORESUME_ENABLED=1` and gets **propose-only** — auto mode never activates. The integration test passes only because it sets the bare `CRASH_AUTORESUME_ENABLED` directly (test_crash_auto_resume.py:248, 319, 555), exercising a variable the documented config never produces.
**Severity:** blocker
**Fix:** Make the reflection read the settings object (`from config.settings import settings; settings.features.crash_autoresume_enabled` and the four threshold fields) instead of bare `os.environ`. Update the integration test to set `FEATURES__`-prefixed vars (or to patch the settings object) so the test validates the real config surface. Pick one variable name and make `.env.example`, `config/settings.py`, the docs, the reflection, and the tests all agree on it.

**File:** `reflections/crash_recovery.py:294-298` (verified: read this file)
**Code:** `if not sig_record.is_auto_eligible(strategy="auto_resume", min_occurrences=_MIN_OCCURRENCES, min_success_ratio=_MIN_SUCCESS_RATIO):`
**Issue:** Auto-resume bootstrap deadlock. `is_auto_eligible` (models/crash_signature.py:190-216) requires `policy_confidence("auto_resume") >= 0.7`, i.e. a ≥70% recovered/attempts ratio on the `auto_resume` strategy. The *only* code that records an `auto_resume` outcome is the attribution loop at crash_recovery.py:185, which runs only on a session that `crash_signature` was stamped on — and the *only* code that stamps `crash_signature` is the successful auto-resume path at crash_recovery.py:360. So: no `auto_resume` attempt has ever happened → `policy_confidence("auto_resume")` is 0.0 → `is_auto_eligible` is False → the reflection never auto-resumes → no outcome is ever recorded → the ratio stays 0.0 forever. A fresh signature can reach `occurrence_count >= 3` but can never clear the success-ratio gate, because the success history that gate demands can only be produced by the action it blocks. Manual `valor-session resume` does NOT stamp `crash_signature` (resume_session is called with `source="valor-session resume"` and never sets it), so operator resumes don't seed the tally either. The "end-to-end auto-resume with ZERO human action" Success Criterion is therefore unreachable in production; the integration test only passes because it hand-seeds `record_outcome("auto_resume", recovered=True)` three times (test_crash_auto_resume.py:539-540) with no resume ever having occurred.
**Severity:** blocker
**Fix:** Provide a bootstrap path for the confidence gate. Options: (a) treat a never-attempted resumable signature with `occurrence_count >= min_occurrences` as eligible for the *first* attempt (cold-start the strategy), letting real outcomes then govern; or (b) have manual `valor-session resume` of a terminal session stamp `crash_signature` (extracted from its trace) so operator-driven recoveries seed the `auto_resume` tally; or (c) split "first-try eligibility" from "sustained eligibility" explicitly. Whichever path, add an integration test that drives a signature from zero history to a fired auto-resume *without* hand-seeding `record_outcome`, since that is the criterion the plan calls the concrete proof the pain is removed.

### Tech Debt

**File:** `reflections/crash_recovery.py:40-44` (verified: read this file)
**Code:** `_MIN_OCCURRENCES = int(os.environ.get("CRASH_AUTORESUME_MIN_OCCURRENCES", "3"))`
**Issue:** The threshold constants are read at module import time, so they are frozen on first import and cannot be re-read at reflection-run time. The plan explicitly states gates are "read at reflection-run time so toggling requires no restart." `auto_enabled` is correctly read inside the function, but the four thresholds are not. (Folds into the blocker fix if you switch to the settings object, which re-reads per access.)
**Severity:** tech_debt
**Fix:** Read the thresholds inside `run_crash_recovery()` (via the settings object once the wiring is fixed) rather than at module scope.

**File:** `docs/features/crash-signature-auto-resume.md` (verified: read this file — grep found no match)
**Code:** `(no reflections.yaml registration section present)`
**Issue:** The reflection callable `run_crash_recovery` is referenced nowhere outside its own module — it is not registered in any `reflections.yaml` and the feature doc never documents the registration step. The plan's Documentation and Update System sections both required documenting "the reflections.yaml addition" (`callable: reflections.crash_recovery.run_crash_recovery`, an `every:` interval, `execution_type: function`). As shipped, the reflection is dead code until an operator hand-edits the iCloud-synced `reflections.yaml`, and no doc tells them how. The plan deferred the actual config write to the operator, but the *instructions* were a required deliverable.
**Severity:** tech_debt
**Fix:** Add a "Scheduling / registration" section to the feature doc with the exact `reflections.yaml` block to add (name, every, priority, execution_type: function, callable, group), matching the format in `config/reflections.yaml`'s header.

### Nits
- None

---

Everything else checks out: deterministic normalization is clean and fail-safe, the two-field write-once attribution split correctly prevents the double-count and overwrite modes the critique named, `RESUMABLE_STATUSES` correctly adds `abandoned` and excludes `cancelled`, the terminal-only pre-transition re-check is present, and `_truthy()` is used consistently for Popoto bool storage. Fix the two blockers (the feature literally cannot do its headline job until then) and this is in good shape.

valorengels:
## Review: Approved

**Branch:** `session/crash_signature_auto_resume` -> `main`
**Plan:** `docs/plans/crash_signature_auto_resume.md` (#1539, epic #1536 Pillar 2)
**Preflight:** OPEN, MERGEABLE, mergeStateStatus=CLEAN. No UI files — screenshot gate is a no-op.
**HEAD:** `5cc33e2d`

Re-review at HEAD `5cc33e2d`, after three fix commits (`3292a111`, `eeb3e01d`, `5cc33e2d`) addressed every finding from the two prior `Changes Requested` reviews. All four blockers and both tech-debt items are resolved, verified against the live code.

### Prior findings — all resolved

**Blocker (confidence gate not wired) — FIXED.** `reflections/crash_recovery.py:303-307` now calls `sig_record.is_auto_eligible(strategy="auto_resume", min_occurrences=..., min_success_ratio=...)` before any resume, skipping to propose-mode when it returns False. The safety substrate is wired into the decision.

**Blocker (CLI reads nonexistent `outcome_tallies` attr) — FIXED.** `tools/valor_session.py` now reads via the model's real accessors: `_crash_sig_strategy_stats()` calls `_load_tallies()` and unpacks the nested per-strategy bucket; both `cmd_crash_signatures` and `cmd_crash_policy` call `s.policy_confidence(...)` and `s.is_auto_eligible(...)` directly. No `outcome_tallies` attribute remains.

**Blocker (env-var wiring mismatch) — FIXED.** The reflection reads `settings.features.crash_autoresume_*` at run time (`crash_recovery.py:78-85`). `.env.example`, `config/settings.py`, the feature doc, and the integration tests all agree on the `FEATURES__CRASH_AUTORESUME_*` prefix.

**Blocker (auto-resume bootstrap deadlock) — FIXED.** `is_auto_eligible` (`models/crash_signature.py:190-239`) switched to an explicit demotion-gate model: a signature past `min_occurrences` with zero recorded attempts is eligible for the first attempt; the success ratio only *demotes* once real attempt data exists. The occurrence floor still requires the pattern to recur across `min_occurrences` distinct sessions before any auto-resume fires.

**Tech debt (thresholds frozen at import time) — FIXED.** All five config values are read inside `run_crash_recovery()` via the settings object, so toggling needs no restart.

**Tech debt (missing reflections.yaml registration docs) — FIXED.** `docs/features/crash-signature-auto-resume.md:143-156` documents the exact `reflections.yaml` block, with the correct note that the gitignored symlink means registration lands in the vault file.

**Tech debt (misleading tests) — FIXED.** `tests/unit/test_crash_signature_cli.py` now builds real `CrashSignature` records (`get_or_create_by_hash` + `upsert_occurrence` + `record_outcome`), exercising the nested-tally read path. The integration suite patches `settings.features` and adds `test_below_confidence_not_auto_resumed` (negative eligibility) and `test_bootstrap_auto_resume_zero_attempts` (zero history -> fired auto-resume through `FEATURES__` config, no hand-seeding).

### Verification this run
- Ruff clean on all four changed source files.
- `test_below_confidence_not_auto_resumed`, `test_bootstrap_auto_resume_zero_attempts`, `test_auto_resume_fires_on_eligible_session` — 3 passed.
- `tests/unit/test_crash_signature_cli.py` — 18 passed.
- Diff vs `main` touches exactly the 18 declared #1539 files, zero granite/persona paths — provenance concern verified resolved.
- No leftover debug/TODO in new code (the single `print` is a docstring usage example).

### Architecture notes (all sound)
- Pure, fail-safe deterministic extractor (`extract_signature` never raises; falls back to `unclassifiable`).
- Determinism guardrail: `plateau` / no-`turn_start` -> `NON_RESUMABLE_DETERMINISTIC` -> escalate, never resume.
- Flag-first attribution ordering gives safe under-count over dangerous over-count.
- Two-field write-once attribution split prevents double-count and overwrite.
- Naive-`updated_at` UTC normalization prevents the silent no-op TypeError drop.
- `RESUMABLE_STATUSES` adds `abandoned`, excludes `cancelled`; terminal-only pre-transition re-check present (#1537 race guard).
- `_truthy()` used consistently for Popoto bool storage; nullable AgentSession fields need no backcompat.
- Off by default — propose-only on every machine except the one designated `FEATURES__CRASH_AUTORESUME_ENABLED=1` host.

### Blockers
- None

### Tech Debt
- None

### Nits
- None

### Legacy Cruft
- None

### Verdict
Approved. Every prior blocker and tech-debt item is fixed and covered by tests that exercise the real config surface and model shapes. Bootstrap auto-resume with zero human action is reachable in production and proven by an integration test. Routing to `/do-docs`.

PR #1727, opened 2026-06-18
fix(#1720): bounded read-path retry for session_id lookups during index rebuild

### LINKED ISSUE #1720: investigation: repair_indexes() non-atomic clear-then-rebuild can transiently drop AgentSession lookups (Session not found)
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/repair_indexes_atomic_rebuild.md

## Summary

Migrated and re-scoped from a misfiled investigation (originally `yudame/cuttlefish#512`, which references this repo's `tools/valor_session.py`, `AgentSession`, and Popoto stack — it was never a `cuttlefish` concern). Most of the original report is now obsolete or already mitigated; this issue tracks the one mechanism that can still reproduce the original symptom.

**Original symptom (issue #496 run, 2026-06-06):** every `tools.valor_session create` dispatch printed `one or more redis keys points to missing objects` and the resulting `AgentSession` was unretrievable shortly after creation (`valor-session status --id <id>` → `Session not found`), breaking stage-by-stage PM dispatch.

## What is now obsolete (not in scope)

- **PM-dispatches-Dev orchestration.** #1633 collapsed PM/Dev into a single Eng role and refuses new parent-attached child sessions (`tools/valor_session.py:420`, `agent/agent_session_queue.py:260`); the granite PTY container owns the PM/Dev split. The "orchestrator polls a dispatched session's status / awaits its completion steering message" workflow that surfaced the bug no longer exists.
- **Original hypothesis #2 (queue-only, never materialized).** Disproven: `_push_agent_session` calls `AgentSession.async_create(...)` synchronously (`agent/agent_session_queue.py:324`). The record is durably written at create time.

## What is already mitigated (not in scope)

- **Original hypothesis #1 (dangling index refs / "missing objects" warning).** Now remediated by `_filter_hydrated_sessions` (phantom drop on read), unconditional `AgentSession.repair_indexes()` every cleanup tick (#1361 removed the gate), class-set orphan cleanup (#1459), and worker-startup index rebuild. The original suggestion to run `keys(clean=True)` is effectively automated.

## Still live — the actual investigation

`AgentSession.repair_indexes()` (`models/agent_session.py:2044-2051`) is **non-atomic**: it iterates every `$IndexF:AgentSession:*` key (including `$IndexF:AgentSession:session_id:*`), removes each key, and only **afterward** calls `rebuild_indexes()` to repopulate. During the gap between the destructive clear and the repopulate, a concurrent `AgentSession.query.filter(session_id=...)` — exactly what `valor-session status --id` does — returns empty, which surfaces as `Session not found`.

This runs on **every** `agent-session-cleanup` reflection tick (hourly) and at worker startup. So the original #512 symptom is still reproducible for any reader polling a recently-created session during the rebuild window — now transient/self-healing rather than permanent, but still a real flake.

Upstream churn source: `AgentSession.Meta.ttl = 2592000` (30 days) expires the hash without coordinated index-entry expiry, continuously producing the stale `$IndexF` members that motivate the destructive rebuild in the first place.

## Hypotheses to investigate (do not fix yet)

1. **Non-atomic rebuild window** is the live cause of transient `Session not found` on freshly-created sessions. Reproduce by issuing `query.filter(session_id=...)` lookups in a tight loop while `repair_indexes()` runs.
2. **TTL/index desync**: confirm whether 30-day hash expiry without index expiry is the dominant source of stale `$IndexF` members (vs. delete paths that already maintain the index).

## Suggested first steps

- Quantify the window: time `repair_indexes()` clear→rebuild on a realistic keyspace; measure lookup-failure probability for a concurrent poller.
- Evaluate fixes (do not implement yet): rebuild into shadow keys + atomic `RENAME`; or prune only the stale members instead of removing whole index keys; or skip the destructive clear when the rebuild is purely additive.
- Audit whether any read path that matters (CLI `status`, worker recovery, steering delivery) lacks a short retry around `query.filter(session_id=...)`.

Surfaced originally by the PM orchestration session for #496 (PR #511). Re-scoped 2026-06-17.

## Recon: Skipped

This issue is itself a reconnaissance artifact (label: `investigation`). The body already contains verified file:line evidence (`models/agent_session.py:2044-2051` for the non-atomic clear→rebuild, `agent/agent_session_queue.py:324` for synchronous create, `tools/valor_session.py:420` for child-session refusal), explicitly disproven hypotheses, and explicitly mitigated concerns. The standard `## Recon Summary` header is omitted because the recon is woven into the 'What is now obsolete / mitigated / Still live' structure. Freshness re-verification at plan time confirmed the live mechanism (now at `models/agent_session.py:2051-2077`).

tomcounsell:
## Summary

Fixes the transient "Session not found" flake where `valor-session status --id <id>` and the SDLC stage-query path returned empty for a live session during the hourly `repair_indexes()` call.

- **Root cause**: popoto's `rebuild_indexes()` deletes the class set (`$Class:AgentSession`) at `base.py:2745` then re-adds members in `batch_size=1000` pipeline batches. A concurrent `query.filter(session_id=...)` reads `smembers($Class:AgentSession)` and returns empty during that delete→re-add window. `session_id` is a plain `Field()` (non-indexed), so the class set is the only resolution path — there is no `$IndexF:AgentSession:session_id:*` key.
- **Measured window** (spike-1, 150 sessions): rebuild takes ~600ms; p99 class-set-empty window = **651ms**
- **Fix**: bounded read-path retry (5 attempts × 200ms = 1000ms total, ~35% above p99) at both reader sites: `tools/valor_session.py::_find_session` and `tools/sdlc_stage_query.py::_find_session_by_id`

## Changes

- `tools/valor_session.py` — add `_CLASS_SET_RETRY_ATTEMPTS`/`_CLASS_SET_RETRY_BACKOFF_S` constants and bounded retry loop in `_find_session`; emits `logger.debug` on each retry attempt
- `tools/sdlc_stage_query.py` — same retry pattern in `_find_session_by_id`; retry sits inside the existing `try/except` so cap-exhausted genuine misses still return `None` cleanly
- `models/agent_session.py` — correct `repair_indexes()` docstring (C4): previous text missed the class-set delete; new text accurately names the class set as the failing layer
- `tests/integration/test_class_set_rebuild_retry.py` — 12 new tests including the C3 concurrency regression test that drives `rebuild_indexes()` directly with a concurrent `session_id` poller; both retry-enabled (found > 0) and retry-disabled (mechanism documented) halves
- `tests/unit/test_valor_session_resume_release.py` — update `TestFindSessionFallbackToAgentSessionId` to assert N-call retry count instead of assert_called_once_with; patch `time.sleep` for fast unit tests
- `docs/features/session-lifecycle.md` — new "Index-Rebuild Race and Read-Path Retry" section with root cause, measured parameters, spike-2 TTL-desync finding

## Test plan

- [x] All 12 new tests in `test_class_set_rebuild_retry.py` pass (including C3 concurrency test)
- [x] `test_valor_session_resume_release.py` updated tests pass
- [x] `repair_indexes()` contract preserved: returns `(stale_count, rebuilt_count)` unchanged
- [x] Pre-existing failure `test_agent_session_queue.py::TestHealthCheckNoProgressRecovery::test_with_progress_not_recovered_parametrized[2-None-None]` confirmed pre-existing on main (unrelated to this change)
- [x] `ruff check` + `ruff format` clean on all modified files

Closes #1720


tomcounsell:
## Review: Approved

**Branch:** `session/repair_indexes_atomic_rebuild` → `main`
**Plan:** `docs/plans/repair_indexes_atomic_rebuild.md`
**Issue:** Closes #1720
**Result:** Approved — no findings

### Summary
Bounded read-path retry (5 × 200ms = 1000ms) added at the two reader sites that surfaced the original `Session not found` flake — `tools/valor_session.py::_find_session` and `tools/sdlc_stage_query.py::_find_session_by_id`. Defends the class-set delete→re-add window inside popoto's `rebuild_indexes()` (`base.py:2745`) without touching popoto internals. Verified against issue #1720 and the plan's acceptance criteria.

### Verification performed
- Checked out PR head `6a211c37` in worktree; read all changed source files.
- Retry loop correctly returns immediately on found, logs `logger.debug` on each retry, and falls through to `AgentSession.get_by_id()` after cap exhaustion — preserving prior absent-session behavior.
- `models/agent_session.py` change is docstring-only; the `repair_indexes()` `(stale_count, rebuilt_count)` contract is unchanged and re-confirmed by `TestRepairIndexesContract`.
- Corrected docstring accurately re-anchors the root cause to the class-set layer (the prior text blamed `$IndexF`, which `session_id` — a plain `Field()` — never reads).
- Ran the test suite: **15 passed** (12 new in `test_class_set_rebuild_retry.py` including the C3 concurrency regression + retry-disabled baseline; 3 updated in `test_valor_session_resume_release.py`).
- `ruff check` clean on all modified files.

### Rubric
- Correctness: PASS — retry matches the corrected causal model; both reader sites covered.
- Tests: PASS — C3 concurrency regression drives `rebuild_indexes()` with a concurrent poller; unit tests assert the N-call retry count and patch `time.sleep`.
- Security: N/A
- Error handling: PASS — retry sits inside the existing `try/except`; cap-exhausted genuine misses return `None`/`get_by_id` cleanly.
- Docs: PASS — new "Index-Rebuild Race and Read-Path Retry" section in `session-lifecycle.md` with measured p99 (651ms) and spike-2 finding.
- Scope discipline: PASS — worker-recovery / steering sites deliberately left out per the plan's documented In-Scope boundary; not a hidden gap.

### Blockers
- None

### Tech Debt
- None

### Nits
- None

### Note (non-blocking, informational)
Issue #1720's "Suggested first steps" mentioned auditing worker recovery and steering delivery for the same retry. The plan deliberately scoped those out (documented under In-Scope / No-Gos), so this PR is complete as planned. If the same transient `Session not found` is observed in worker recovery or steering paths later, a follow-up extending the retry there would be the natural next step.

Issue #1356, opened 2026-05-09
worker: tier-1 _has_progress sub-check B passes forever for sessions that never produce SDK output


tomcounsell:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/sdlc-1356.md

## Background

Spun out from investigation #1246. PM session `90c4117dbf06431c86ee4807d7a18bcd` ran with `status=running` for ~10.9 hours with zero SDK communication; the tier-1 `_has_progress` reaper did not fire over ~130 reflection ticks.

## Root cause (verified)

Tier-1 sub-check B at `agent/session_health.py:683-693` reads `last_heartbeat_at`, which is written every 60s by `_heartbeat_loop` in `agent/session_executor.py:1712-1729`. That loop is keyed on the asyncio task wrapper, **not** the SDK subprocess making forward progress. Any wedge that lets the asyncio task keep awaiting (no exception, no return) — including cwd-disappearance — produces an indefinitely fresh `last_heartbeat_at` and a permanent Tier-1 pass.

`sdk_ever_output=False` plus the unreachable Tier-2 escalation guard at `_tier2_reprieve_signal` (lines 770-775) explains the full 11h wedge.

> Note: investigation #1246 referred to a "stdout-staleness reaper" from #1046/#1039. That detector was retired by #1172. The current detector is heartbeat-fresh; the leak is in the replacement.

## Acceptance criteria

- [ ] A session with `sdk_ever_output=False` and `running_seconds > MAX_NO_OUTPUT_REPRIEVES * HEARTBEAT_WRITE_INTERVAL` (~30 min) MUST cause `_has_progress` to return False on the next health-check tick, regardless of `last_heartbeat_at` freshness — Tier 1 hands off to Tier 2 once the no-output budget is exhausted.
- [ ] Pre-#1226 grace behavior preserved during startup: a session with `running_seconds < STARTUP_GRACE_SECONDS` (new explicit constant) and a fresh heartbeat MUST still pass Tier 1.
- [ ] Once Tier 1 declines, the existing Tier 2 escalation guard fires and recovers the session via `_apply_recovery_transition` after `MAX_NO_OUTPUT_REPRIEVES` ticks.
[...]

PR #1728, opened 2026-06-18
feat(#1724): recover stalled never_started and mid-run-wedge granite sessions

### LINKED ISSUE #1724: Stalled "never_started" sessions are unrecoverable for up to 30 min: fresh heartbeats blind the recovery actor while the advisory that detects them cannot act
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/never_started_session_recovery.md

> ## ⚠️ Scope update (2026-06-18) — read this first; it broadens the issue
>
> This issue was originally scoped to the **`never_started` priming wedge** only. Two pushes to `main` since the plan was drafted change both the scope and the available signals. The planner must treat the sections below as updated by this block where they conflict.
>
> **1. The container's only other hang detector was just removed.** Commit `ef53a88f` changed `agent/granite_container/container.py::CYCLE_IDLE_TIMEOUT_S` from `120.0` to `12 * 60 * 60.0` (12h) and demoted it to a *sanity ceiling, not a hang signal*. The in-container `dev_hang`/`pm_hang` path (which used to fire at 120s) no longer detects hangs — the code comment now explicitly delegates: *"Real hang detection lives in the heartbeat / liveness-recovery layer (agent/session_health.py + issue #1724)."* **As of now, the session-health/liveness layer is the SOLE hang detector for granite sessions.** This issue therefore must own the whole problem, not just the priming case.
>
> **2. Scope broadens: `never_started` priming wedge AND mid-run wedge.** The container used to catch a session that produced turns (`sdk_ever_output == True`) and then genuinely hung during a long Dev turn — at 120s. Nothing catches that now for up to 12h. The fix must cover **both**:
>   - **(a) Priming wedge** — `last_turn_at is None`, heartbeats fresh, never produced a turn (the original issue).
>   - **(b) Mid-run wedge** — `sdk_ever_output == True`, then no real progress for a meaningful window during a long-running turn. Previously the 120s container deadline; now uncovered.
>
> **3. Governing constraint: detection MUST be progress-based, not timeout-based.** This is promoted from a fix detail to the design rule for this issue. A bare heartbeat ticking on a schedule is exactly the false signal this issue already named — moving a wall-clock budget to a different layer is NOT a fix. The reliable oracle is *real activity*. The detector must catch "heartbeating but transcript not growing AND no new tool activity for a meaningful window."
>
> **4. The progress signals changed under this issue — they now exist for granite.** Before today these were `None`/stale for granite (broken transcript path + tool name never cleared). Three pushes (`995bc453`, `91289fc3`, `b414eed1`) fixed the transcript path slug and tool-completion clearing, and `bridge_adapter.py` now diff-gates real progress fields onto the `AgentSession` every tailer tick. The plan should lean on these as the progress oracle rather than assuming granite exposes nothing:
>   - `turn_count` (advances per classified turn) — `bridge_adapter.py:911-921`
>   - `last_turn_at` bumped on first/each turn, which flips `sdk_ever_output` True — `bridge_adapter.py:626-648`
>   - `current_tool_name` (now correctly **cleared on tool completion** — `b414eed1`) and `last_tool_use_at` — `bridge_adapter.py:891-897, 917-927`
>   - `byte_offset` (raw transcript byte growth — the most direct "is it doing anything" signal) — `bridge_adapter.py:906-920`
>   - token counts
>   - **Any planning assumption that "granite sessions expose no progress fields" is now stale and would mislead the plan.**
>
> **5. Stakes are higher → reliability is load-bearing.** With the container silent for 12h, if this recovery misses a real hang the session wedges for *hours*, not 30 minutes. The detector must robustly catch the wedge; the cost of a miss went up by ~24×.
>
> **6. The regression bar is the exact mistake the old 120s deadline made.** A Dev legitimately working a multi-minute turn — transcript bytes / tool activity advancing, no completed turn yet — must NOT be recovered. The old `dev_hang` killed any turn over two minutes. The new detector must distinguish "actively producing bytes/tools, slow" from "heartbeating but byte-silent and tool-silent."

## Recon Summary

Reconnaissance performed during /do-plan Phase 0/0.5 (2026-06-18, baseline commit b414eed1). The issue body itself carries unusually thorough recon (file:line refs, definitions, prior art); this section records re-verification against current main.

**Confirmed:**
- `reflections/stall_advisory.py::run_stall_advisory` is advisory-only — logs WARNING + optional Telegram (gated `stall_advisory_telegram_enabled`, default off), zero writes. No kill/requeue path. (verified `stall_advisory.py:38-161`)
- `never_started` verdict + 120s grace live in `agent/session_stall_classifier.py::NEVER_STARTED_GRACE_SECS = 120` (NOT in stall_advisory.py); classifier flags when a probe-status session has no `turn_start` event and `elapsed > 120s`. (verified `session_stall_classifier.py:50, 240-265`)
- `agent/session_health.py::_has_progress` sub-check B grants a fresh-heartbeat fast-path for no-output sessions across the band `STARTUP_GRACE_SECONDS (300s) <= running_seconds <= NO_OUTPUT_BUDGET_SECONDS (1800s)`. (verified `session_health.py:784-824`)
- Constants confirmed: `HEARTBEAT_FRESHNESS_WINDOW=90` (l.219), `STARTUP_GRACE_SECONDS=300` aliased to `AGENT_SESSION_HEALTH_MIN_RUNNING` (l.267-269), `NO_OUTPUT_BUDGET_SECONDS=1800` (l.284), `MAX_NO_OUTPUT_REPRIEVES=20` (l.251).

**Revised (important — changes the open-question framing):**
- The recovery actor's main loop only calls `_has_progress` when `running_seconds > AGENT_SESSION_HEALTH_MIN_RUNNING (300s)` (race-condition guard, `session_health.py:1708-1712`). **Therefore the advisory's 120s grace can never be honored by the actor — 300s is a hard floor for any recovery action in the current architecture.** Reconciliation cannot mean "both fire at 120s"; it means a single shared source of truth with the actor acting at/after the 300s floor.
- A no-output escalation cap already exists: `_tier2_reprieve_signal` suppresses all reprieve gates once `sdk_ever_output is False AND reprieve_count >= MAX_NO_OUTPUT_REPRIEVES (20)` (`session_health.py:942-943`). The 1800s wedge is the *combination* of sub-check B's band + the 20-tick reprieve cap. **A fix that only touches sub-check B does not fix the reported bug** — the reprieve cap independently grants ~100 min.

**Freshness disposition — CHANGED (was "Unchanged"; superseded by the Scope update block above):** Since this recon ran at `b414eed1`, `main` advanced to `ef53a88f`, which gutted the in-container 120s hang detector (now a 12h ceiling). The original recon's "no commits touched the hang layer" finding no longer holds at the system level: the *other* detector that backstopped this gap is gone, so the mid-run wedge is now in scope and the session-health layer is the sole detector. See the Scope update block.

**Pre-requisites:** None — purely internal worker recovery logic.

---
> **Context for a stranger to this codebase.** A *granite session* is a bridge-originated Engineer task run as an interactive Claude Code TUI pair (a PM driver + a Dev driver) inside a PTY (`agent/granite_container/`). The *worker* (`python -m worker`) is the sole execution engine; it runs periodic *reflections* (scheduled maintenance functions) and a *session-health* recovery pass that detects wedged sessions and re-queues/restarts them. *Priming* is the startup phase where each TUI is fed its persona before any real turn runs. A session emits two liveness signals every ~60s from a background watchdog: `last_heartbeat_at` (queue layer) and `last_sdk_heartbeat_at` (SDK layer). These keep flowing during priming, **before** the session has produced any turn output (`last_turn_at` is still `None`). A *tailer* tick (`bridge_adapter.py`) reads new transcript bytes each cycle and diff-gates real progress fields (`turn_count`, `current_tool_name`, `last_tool_use_at`, `byte_offset`) onto the session record.

## Problem

A granite session can be marked `running` and keep emitting fresh heartbeats while making **no real progress** — and as of commit `ef53a88f` the only layer that can detect this is the session-health/liveness pass. There are two distinct wedge shapes:

**(a) Priming wedge (`never_started`).** The session is wedged before producing any turn (e.g. stuck in PM priming). `last_turn_at = None`, heartbeats fresh. Two detectors disagree and **the one that can act is blinded by the heartbeat**:

1. **`reflections/stall_advisory.py` (`run_stall_advisory`) — sees the problem, cannot act.** It classifies the session as `STALLED reason=never_started` at a **120s** grace, but is advisory-only: logs a WARNING and *optionally* sends a Telegram alert (default off), then `return`s. It never kills, re-queues, or restarts. (`reflections/stall_advisory.py:38-161`)

2. **`agent/session_health.py::_has_progress` — can act, but is blind.** For a no-output session (`sdk_ever_output == False`), sub-check B treats a fresh heartbeat as "alive" until `NO_OUTPUT_BUDGET_SECONDS = 1800` (30 min), and the reprieve cap extends this further. Because the wedged session keeps heartbeating, `_has_progress` returns `True` and recovery is skipped. (`session_health.py:784-824`)

**(b) Mid-run wedge (newly in scope).** The session produced turns (`sdk_ever_output == True`) and then genuinely hangs during a long Dev turn — transcript stops growing, no new tool activity — while heartbeats keep ticking. Until `ef53a88f` the in-container 120s `dev_hang`/`pm_hang` deadline caught this; that deadline is now a 12h ceiling, so nothing recovers it for up to 12h.

**Net effect:** a granite session can be wedged in either shape and remain invisible to the actor that can recover it — for **up to ~100 min (priming, budget + reprieve cap) or up to 12h (mid-run)**. The thresholds across layers are wildly inconsistent (advisory 120s, no-output budget 1800s, reprieve cap ~100min, container ceiling 12h), so no two systems agree on when to intervene, and the reliable signal — actual progress — is not the thing being measured.

### Desired outcome
- A wedged granite session — whether it never started (no turn) **or** stalled mid-run (had turns, then byte/tool-silent) — is escalated to actual recovery on a timescale far below the 12h container ceiling, driven by **absence of real progress** (transcript byte growth, new tool activity, advancing `turn_count`) rather than by a wall-clock timeout or a heartbeat tick.
- A legitimately-slow-but-progressing session (large initial prompt, slow auth, a genuinely long multi-tool Dev turn that is still emitting bytes/tools) is **NOT** recovered.

### Real observation (2026-06-18)
Session `tg_valor_-1003449100931_993` (`agent_session_id 7f819c9a7efa46f1b2b8b10ef1d34dfc`) was `running` + heartbeating with `last_turn_at = None` for ~5 min (priming wedge, shape (a)). `stall-advisory` logged:
```
[stall-advisory] STALLED session=tg_valor_-1003449100931_993 reason=never_started
  signals={'elapsed_secs': 247, 'grace_secs': 120, 'session_status': 'running'}
```
at 02:48:53. Nothing recovered it until a manual `valor-service.sh restart`.

## Definitions

|------|-----------|--------|
| Granite session | Bridge-originated Eng task run as a PM+Dev interactive Claude Code TUI pair in a PTY | `agent/granite_container/` |
| Priming | Startup phase feeding each TUI its persona before any real turn | `agent/granite_container/container.py` |
| Priming wedge / `never_started` | Session wedged before producing any turn (`last_turn_at is None`) | this issue, shape (a) |
| Mid-run wedge | Session produced turns then hangs during a turn (byte/tool-silent, `sdk_ever_output == True`) | this issue, shape (b) |
| `_has_progress` | The session-health predicate that decides whether a running session is making progress (and thus must NOT be recovered) | `agent/session_health.py:~700-860` |
| `sdk_ever_output` | True once the session has emitted any tool/turn event; flipped True by `last_turn_at` bump | `agent/session_health.py`, `bridge_adapter.py:626-648` |
| Progress oracle (granite) | Real-activity signals tailed onto the session each tick: `turn_count`, `current_tool_name`, `last_tool_use_at`, `byte_offset`, token counts | `bridge_adapter.py:837-927` |
| `CYCLE_IDLE_TIMEOUT_S` | In-container per-cycle idle ceiling — now **12h, a sanity bound, not a hang detector** (was 120s) | `agent/granite_container/container.py:138` |
| `NO_OUTPUT_BUDGET_SECONDS` | 1800s ceiling after which a no-output session stops getting the fresh-heartbeat pass | `agent/session_health.py:284` |
| `MAX_NO_OUTPUT_REPRIEVES` | 20-tick reprieve cap that independently extends the no-output wedge past 1800s | `agent/session_health.py:251` |
| stall-advisory | Reflection that classifies stalled sessions; warn/alert only, no action | `reflections/stall_advisory.py` |

## Prior art (read before planning)
- **#1356** (closed) — *"tier-1 `_has_progress` sub-check B passes forever for sessions that never produce SDK output."* Introduced the `NO_OUTPUT_BUDGET_SECONDS=1800` gate. This issue is its direct follow-up: the budget exists but is (a) inconsistent with the advisory's 120s grace and (b) still a 30-min wedge, AND (c) bypassed by the reprieve cap. **Watch for #1356's partial-fix pattern** — do not just relocate a timeout or touch only one of the two gates.
- **#1614** (closed) — gated sticky own-progress fields on heartbeat freshness to stop a stale `running` session evading recovery. Same family of bug; reuse its approach/tests as a model.
- **#1172** (closed) — "PM session liveness: surface progress or stay graceful — stop inferring from staleness." Relevant design context on not over-killing slow sessions.
- **`ef53a88f`** (this week) — removed the in-container 120s hang deadline, making this layer the sole hang detector. Read its `container.py:121-138` comment block — it states the design intent this issue must fulfill.

## Solution sketch (open questions, not a prescription)

The root cause is a *policy disagreement* plus a *now-missing backstop*, so do-plan should resolve direction before coding. The progress-based-detection constraint (Scope update §3) is non-negotiable; the open questions are about *where* and *how*:

- **Q1 (priming wedge).** When `last_turn_at is None` AND `running_seconds` past the actor's 300s floor, the heartbeat loop must NOT count as progress (it runs during priming). What is the grace, reconciled to a single source of truth, given the actor cannot physically act before 300s? **Must also address the reprieve cap (`MAX_NO_OUTPUT_REPRIEVES`), not just sub-check B.**
- **Q2 (mid-run wedge — NEW).** What is the progress-absence window for a session with `sdk_ever_output == True`? Define "no real progress" concretely: `byte_offset` unchanged AND no new `last_tool_use_at` AND `turn_count` not advancing for N seconds. Pick N and justify it against the regression bar (a slow-but-emitting Dev turn must survive).
- **Q3 (escalation owner & no double-act).** Does the recovery actor gain the escalation, or does `stall_advisory` escalate to a real action? How do the two avoid double-acting on the same session in one window?
- **Q4 (single source of truth).** Where does the shared grace/window constant live so the advisory and the actor cannot drift apart again?
- **Q5 (regression proof).** Tests proving BOTH: (a) a still-priming-but-progressing session survives, and (b) a long multi-tool Dev turn that is actively emitting bytes/tools is NOT recovered. This is the exact failure mode the old 120s `dev_hang` had.

Guardrail: this is **not** a hotfix. It is now the sole hang detector for granite; a naive change either re-introduces the false-kill the 12h ceiling just removed, or lets sessions wedge for hours. It needs a plan + tests.

## Acceptance criteria
- [ ] Detection is **progress-based**: recovery decisions are driven by absence of real activity (`byte_offset` growth, new `last_tool_use_at`, advancing `turn_count`), NOT by a bare heartbeat tick or a fixed wall-clock budget alone.
- [ ] **Priming wedge:** a `running` session with `last_turn_at = None` and fresh heartbeats is recovered on a grace consistent with stall-advisory (far below the 12h container ceiling), not after 30 min — and the fix closes BOTH sub-check B and the reprieve-cap path.
- [ ] **Mid-run wedge:** a session with `sdk_ever_output == True` whose transcript stops growing and emits no new tool activity for the defined window is recovered, far below the 12h ceiling.
- [ ] The advisory's grace and the recovery actor's grace/window derive from a single shared source of truth (no silent drift).
- [ ] **No false-kill regression:** (a) a legitimately-slow-but-progressing priming session and (b) a long Dev turn that is actively emitting bytes/tools are NOT recovered — regression tests included for both.
- [ ] No double-recovery: advisory and session-health do not both act on the same session in one window.

---
*This issue will be consumed by `/do-plan` to produce a plan at `docs/plans/{slug}.md`. The Scope update block and open questions Q1–Q5 are intentional — resolve them before writing a concrete approach.*

valorengels:
## Summary

- **Path A (never-started)**: sessions that have been running for more than `NEVER_STARTED_GRACE_SECS + NEVER_STARTED_CONFIRM_MARGIN_SECS` (default 150s) with no SDK output are flagged and recovered via the 30s tool-timeout sub-loop. Two companion guards prevent false-positive reprieves: sub-check B D0 gate (denies the fresh-heartbeat fast-path when `_never_started_past_grace` is True) and Tier-2 reprieve bypass (suppresses all psutil/compaction/children gates for never-started sessions)
- **Path B stage-1 (observe-only)**: four new nullable `AgentSession` fields track PTY liveness (`last_pty_read_loop_at`, `last_pty_activity_at`, `mid_run_quiescent_since`, `mid_run_pty_snapshot`). The stage-1 cheap quiescence gate runs on every 30s tick and logs confirmed suspects as warnings. Stage-1 detects and records suspects only — no recovery fires. Stage-2 (off-loop classifier dispatch, recovery wiring) is deferred to a follow-up.
- All magic numbers are named env-overridable constants with provisional/tunable comments
- 22 new unit tests covering both paths

## Test plan
- [x] Run `scripts/pytest-clean.sh tests/unit/test_never_started_recovery.py -n0 -v` — 22 passed
- [x] Run `scripts/pytest-clean.sh tests/unit/test_session_stall_classifier.py -n0 -v` — 39 passed
- [x] Run `scripts/pytest-clean.sh tests/unit/test_session_health_inference_removed.py -n0 -v` — passed
- [x] `python -m ruff format` + `ruff check` on all modified files — clean

Closes #1724

## Known Limitations (Stage-2 Follow-Up Items)
These are acknowledged under-detection cases that are safe for the current stage-1 observe-only shipping. They MUST be addressed before stage-2 wires recovery:

1. **`byte_offset` proxy** — `mid_run_pty_snapshot` stores `total_input_tokens` as a proxy for transcript byte offset since `byte_offset` isn't persisted on `AgentSession`. Safe for stage-1 (corroborator-only), but stage-2 needs the real value.

2. **PTY stamp frequency** — `on_pty_read` fires at turn-boundary idle, not every read-loop tick. A session wedged mid-turn may not get a fresh `last_pty_read_loop_at` stamp until the idle bar fires, causing stage-1 to ABSTAIN. Under-fires (safe), not over-fires.


valorengels:
## Review: Changes Requested

**PR:** #1728 — feat(#1724): recover stalled never_started and mid-run-wedge granite sessions
**Branch:** `session/never_started_session_recovery` → `main`
**Plan:** `docs/plans/never_started_session_recovery.md`
**Head SHA:** `263b07de`

The architecture is sound and matches the plan well: the shared `_never_started_past_grace` predicate binds correctly at all three call sites (30s-loop actor branch, sub-check B, and the D3 reprieve guard), the `no_progress` recovery path correctly relies on the D3 reprieve bypass to suppress `children`/`alive` reprieves, the two-field PTY liveness design and three-state freshness logic are implemented as specified, and all writers are fail-silent. Import direction is clean, the 22 new unit tests pass, and ruff is clean.

But the PR leaves the existing test suite **red**, and its own plan flagged the exact file that now fails. One blocker.

### Blockers

**File:** `tests/unit/test_session_health_inference_removed.py` (verified: ran this suite on both branches)
**Code:** `assert result is True, (` (3 occurrences — `test_own_progress_gate_fresh_heartbeat_is_progress`, `test_fresh_heartbeat_no_stdout_yet_is_progress`, `test_fresh_heartbeat_with_long_stdout_silence_is_progress`)
**Issue:** These three tests fail on this branch (12 passed on `main`, 3 failed here — confirmed regression, not pre-existing). They assert that a no-output session (`last_turn_at=None`, `last_tool_use_at=None`) with a fresh heartbeat, running 10 min / 1 h / 4 h, must return `True` from `_has_progress`. The new sub-check B gate (`if _never_started_past_grace(entry): return False`) now correctly returns `False` for exactly these sessions — that is the intended fix, so the tests now encode the pre-fix behavior. The plan's own `## Test Impact` section explicitly lists this file as **UPDATE**: *"add never-started-grace assertions following its sub-check-B pattern; re-verify no assumption of the old 1800s boundary for the never-started case."* The PR never touched the file (it is absent from the diff).
**Severity:** blocker
**Fix:** Update the three tests to reflect the new contract. A no-output session past `NEVER_STARTED_GRACE_SECS + NEVER_STARTED_CONFIRM_MARGIN_SECS` (150s) is now correctly NOT "progress" — flip these assertions to `False`, or rework the fixtures so the genuinely-still-priming case (running < 150s) is the one asserted `True`. Keep one case proving a session inside the grace window is still protected. Run `scripts/pytest-clean.sh tests/unit/test_session_health_inference_removed.py -n0` green before re-review.

### Tech Debt

**File:** `models/agent_session.py:398`
**Code:** `# Cross-tick durable state: serialized (last_pty_activity_at_iso, byte_offset)`
**Issue:** The model docstring (and the stage-1 plan language) describe the snapshot tuple as `(last_pty_activity_at, byte_offset)`, but `_eval_mid_run_pty_stage1` actually serializes `(last_pty_activity_at, total_input_tokens)` — `byte_offset = getattr(entry, "total_input_tokens", None)  # best available proxy`. The real transcript `byte_offset` is never persisted onto `AgentSession` (it lives only on the in-memory `pm_state`/`dev_state` tailer cursors in `bridge_adapter.py`), so the substitution is a pragmatic necessity, not a bug — and per plan D2/concern #3 this value is an optional corroborator, never a gating conjunct, so functional impact is low. But the comment now lies about what is stored, which will mislead the stage-2 builder who must use this same tuple as the CAS precondition.
**Severity:** tech_debt
**Fix:** Either update the comment to say `(last_pty_activity_at, total_input_tokens)` and explain why tokens stand in for byte_offset, or persist the real transcript `byte_offset` onto the session in the tailer save so the field matches the plan's intent. Pick one and make the comment match the code.

### Nits
- None

### Legacy Cruft
- None observed. No deprecated fields, dead imports, or commented-out code in the diff.

---

**Scope note (not a finding):** This PR intentionally ships path A (never-started) active and path B stage-1 observe-only (no stage-2 classifier, no recovery wiring) — consistent with the plan's staged D2-rollout and the still-`in_progress` tasks 3b–8. The `MID_RUN_RECOVERY_ACTIVE` flag, the stage-2 classifier, the off-loop dispatch, the dashboard fallback-rate alert, and the integration tests in `tests/integration/test_agent_session_health_monitor.py` are all deferred to follow-up work and are correctly out of scope here. The blocker above is the one issue that must be fixed before this slice can merge: the slice it does ship must not leave existing tests failing.

**Issues Found: 2** — Blockers: 1 · Tech Debt: 1 · Nits: 0
**Screenshots: 0** (no UI files in diff)

valorengels:
## Review: Changes Requested

**Branch:** `session/never_started_session_recovery` → `main`
**Plan:** `docs/plans/never_started_session_recovery.md`
**Mergeability:** preflight CLEAN (OPEN, MERGEABLE)

### What's solid

Path A (never-started / priming wedge) is implemented correctly and completely, matching the plan's locked decisions:

- **Shared predicate** `_never_started_past_grace(entry, now=None)` (`agent/session_health.py:693`) derives `now` internally so it binds in `now`-less callers (BLOCKER 1 resolved).
- **All three never-started gates** are wired to the one predicate (concern #4 / no-drift):
  1. Dedicated recovery branch in the **30s** `_agent_session_tool_timeout_check` tick (`agent/session_health.py:2508-2531`) with fresh re-read, terminal-status guard, `reason_kind="no_progress"`, and best-effort counter — the cadence BLOCKER (D0) is honored.
  2. Sub-check B fast-path denial (`agent/session_health.py:874`).
  3. Reprieve bypass (D3) fires before the compacting/children/alive gates (`agent/session_health.py:1031-1034`).
- **Shared grace constants** live once in `session_stall_classifier.py:53` and are imported by `session_health.py` (drift test green).
- **PTY liveness writers** (`bridge_adapter.py:680-698`, `container.py:766-774`) are fail-silent, use `update_fields`, and fire unconditionally.
- 22 new unit tests pass; `test_session_stall_classifier.py` (39) and `test_session_health_inference_removed.py` pass; ruff clean. Import-direction guard holds.

### Blockers

**File:** `(PR scope vs. tracking issue #1724 acceptance criteria)`
**Code:** `Closes #1724` (PR body)
**Issue:** The PR claims `Closes #1724`, but issue #1724's acceptance criteria require **mid-run wedge recovery** ("a session with `sdk_ever_output == True` whose transcript stops growing... is recovered, far below the 12h ceiling"). This PR ships Path B **stage-1 only** — `_eval_mid_run_pty_stage1` (`agent/session_health.py:2280`) updates durable state and *logs* confirmed suspects (comment line 2538: "Does NOT recover — it only flags confirmed suspects for stage-2 (task 4)"). The plan's task 3b (stage-2 classifier + recovery), task 4 (integration tests), tasks 5-7 (validation + docs) are all unbuilt (live task list shows tasks 4-8 pending). A mid-run wedge is NOT recovered by this PR. Merging this as `Closes #1724` leaves the issue's headline goal unmet.
**Severity:** blocker
**Fix:** Either (a) retarget the PR body to `Refs #1724` / `Part of #1724` and adjust the title to reflect "path A + path B stage-1 (observe) only", keeping #1724 open for the stage-2 follow-up; or (b) build stage-2 (off-loop classifier dispatch, `MID_RUN_RECOVERY_ACTIVE` flag, observe-only path, counters) before merge.

**File:** `(PR body — Summary)`
**Code:** `ships observe-only (no recovery) behind MID_RUN_RECOVERY_ACTIVE=1`
**Issue:** The PR Summary states path B "ships observe-only (no recovery) behind `MID_RUN_RECOVERY_ACTIVE=1`". That env flag does not exist anywhere in the diff (`grep MID_RUN_RECOVERY_ACTIVE agent/` returns no match), nor does the classifier dispatch, the `mid_run_pty_quiescent_recovery_observed` counter, or the dashboard surface. The PR body describes the full plan, not what shipped. Stage-1 simply logs a warning on a confirmed suspect.
**Severity:** blocker
**Fix:** Correct the PR body to describe only what ships: stage-1 detects + logs suspects; there is no observe-only-recovery path and no `MID_RUN_RECOVERY_ACTIVE` flag yet.

### Tech Debt

**File:** `docs/features/never_started_session_recovery.md:72,126,129`
**Code:** `Path B ships in **observe-only mode** by default. The MID_RUN_RECOVERY_ACTIVE=1 env flag is reserved...` / `Off-loop classifier slot must be removed in a finally block`
**Issue:** The feature doc describes machinery that does not exist in the shipped code: the `MID_RUN_RECOVERY_ACTIVE` flag, an "off-loop classifier slot," and observe-only *recovery*. The code only logs suspects. The doc conflates the plan's future stage-2 with what's present, which will mislead the next reader/operator.
**Severity:** tech_debt
**Fix:** Scope the doc to the shipped behavior. Either move the stage-2 / `MID_RUN_RECOVERY_ACTIVE` paragraphs under an explicit "Planned (not yet shipped)" heading or remove them until stage-2 lands.

**File:** `agent/granite_container/bridge_adapter.py:688`
**Code:** `if buffer != self._prev_pty_buffer:`
**Issue:** The plan (Key Element 3 / models field comment) specifies stamping `last_pty_activity_at` only on **normalized** screen bytes ("strip known cursor/blink noise"). The diff-gate compares the raw buffer, so a blinking cursor or spinner repaint that the plan intended to normalize away would keep marking activity fresh. Mitigated today because `read_until_idle` returns a settled `turn_buffer` and stage-1 only logs (no recovery), but once stage-2 recovery is wired this becomes a false-negative source (a wedged-but-blinking screen never goes quiescent).
**Severity:** tech_debt
**Fix:** Normalize the buffer (strip cursor/ANSI blink noise) before the diff comparison, per the plan, or document why the raw turn_buffer is sufficient.

### Nits
- None

### Test coverage gap (folds into blocker #1)
The plan's first Success Criterion requires the end-to-end never-started recovery be "Asserted in `tests/integration/test_agent_session_health_monitor.py`". This PR adds **zero** integration tests (`git diff --stat tests/integration/` is empty). All 22 new tests are unit-level. The end-to-end recovery path (30s-loop branch → `_apply_recovery_transition`) is not exercised against a real session record. This is task 4 in the plan, still pending.

---
Net: Path A is merge-quality work. The blockers are about **scope honesty** — this PR is a correct partial implementation being presented (via `Closes #1724` and the PR body) as the complete feature. Recommend retargeting to `Refs #1724` and shipping path A + stage-1-observe as an explicit increment, OR completing stage-2 + integration tests before merge.

valorengels:
## Review: Changes Requested

**Scope:** Path A (never-started recovery, active) + Path B stage-1 (PTY quiescence detection, observe-only). Stage-2 classifier/recovery is explicitly deferred to a follow-up — this matches the plan's staged sequencing and the issue's "ship observe-only first" guidance.

### Verification performed
- Preflight: OPEN, MERGEABLE, mergeStateStatus=CLEAN.
- Plan/issue read (`docs/plans/never_started_session_recovery.md`, issue #1724). Implementation matches the plan's D0 host-loop placement (both branches in the 30s `_agent_session_tool_timeout_check`, not the 300s loop), the four new nullable `AgentSession` fields, the stateless `mid_run_quiescent_since` durability design, and the fail-silent posture throughout.
- Tests: `tests/unit/test_never_started_recovery.py` + `test_session_health_inference_removed.py` + `test_session_stall_classifier.py` → **74 passed** (-n0).
- `ruff check` on all modified files → clean.
- Import direction verified: `agent.session_health` imports the two grace constants from `agent.session_stall_classifier` with no cycle (`import agent.session_health` succeeds). Drift-pinned by `test_grace_constant_alignment`.
- No debug code (`print`/`pdb`/`TODO`) in the diff.

### Correctness notes (all clean)
- `_never_started_past_grace` is correctly fail-safe: returns False on `sdk_ever_output=True`, missing timestamps, or any exception. Timezone-aware comparison handles naive datetimes.
- The two companion guards (sub-check B D0 gate + Tier-2 reprieve bypass) close BOTH wedge paths the issue called out — a fix to sub-check B alone would not have, since the reprieve cap independently granted ~100 min.
- Path B stage-1 is genuinely observe-only: `_eval_mid_run_pty_stage1` only logs a warning on a confirmed suspect and maintains durable state — no `_apply_recovery_transition` call. It cannot false-kill in this PR.
- The `.incr()` counter writes follow the established `session_health.py` pattern (plain Redis counter keys, not Popoto-managed model keys) — no raw-Redis-on-model violation.
- Spike-1 result (parent TUI repaints its own spinner during a child Task subagent) is captured inline in `_make_pty_read_callback`, satisfying the plan's build-gate before the gate is wired.

### Nits

**File:** `docs/features/README.md` (index entry for this feature)
**Code:** `observe-only by default, recovery behind MID_RUN_RECOVERY_ACTIVE=1 (#1724)`
**Issue:** The README index line presents `MID_RUN_RECOVERY_ACTIVE` as if it gates recovery in the shipped code, but that flag does not exist anywhere in `agent/` or `models/` — it is stage-2 work the feature doc body correctly describes as deferred (`never_started_session_recovery.md:72,109-112`). A reader scanning the index would conclude path-B recovery exists today behind a flag; it does not.
**Severity:** nit
**Fix:** Reword the index entry to match the doc body, e.g. "...observe-only (detect-and-log); stage-2 classifier-driven recovery deferred to a follow-up to #1724."

### Acknowledged design notes (not findings)
- `mid_run_pty_snapshot` substitutes `total_input_tokens` for the plan's `byte_offset` because `byte_offset` lives only on in-memory tailer cursors and is not persisted on `AgentSession`. This is safe for stage-1 observe-only (byte_offset is an optional corroborator, never a gating conjunct per plan D2/concern #3) and is explicitly flagged in both the model field comment and the read callback to be resolved before stage-2 uses the tuple as a CAS precondition. Correctly scoped to the deferred follow-up.
- The bridge_adapter diff-gates the raw (ANSI-stripped) `turn_buffer` rather than the finer cursor/blink normalization the plan specified for stage-2; this is documented inline as a stage-2 follow-up item.

Verdict: one documentation-accuracy nit. No blockers, no tech debt. Code, tests, and design are sound for the observe-only scope.

valorengels:
## Review: Changes Requested

**Branch:** `session/never_started_session_recovery` → `main`
**Plan:** `docs/plans/never_started_session_recovery.md`
**Tracking:** #1724
**Scope shipped:** Path A (never-started, active recovery) + Path B stage-1 (mid-run PTY quiescence, observe-only). Stage-2 classifier/recovery deferred to a follow-up.

Preflight clean (OPEN, MERGEABLE, mergeStateStatus=CLEAN). No UI files — visual-proof gate N/A. Ruff clean. Named unit suites pass (35 passed: `test_never_started_recovery.py` 22 + `test_session_health_inference_removed.py` 13).

### Path A assessment (active leg) — solid
Path A is internally consistent and well-guarded. The shared `_never_started_past_grace` predicate intercepts all three intended call sites: the new 30s-loop recovery branch, sub-check B's fresh-heartbeat fast-path (`_has_progress`), and the Tier-2 reprieve escalation guard (`_tier2_reprieve_signal`). The recovery branch passes `reason_kind="no_progress"` explicitly rather than relying on the plan's reason-string substring match — more robust. Verified the reprieve chain: `_apply_recovery_transition(no_progress)` → `_tier2_reprieve_signal` → D3 bypass returns `None` (first escalation guard) → reprieve suppressed → recovery proceeds. The race-mitigation pattern (re-read fresh, re-confirm predicate, then transition) mirrors the existing tool-timeout path. Predicate fail-soft (never raises, safe defaults for phantom/legacy records). No blockers here.

### Tech Debt

**File:** `models/agent_session.py:402` / `agent/session_health.py:2386`
**Code:** `byte_offset = getattr(entry, "total_input_tokens", None)  # best available proxy`
**Issue:** The plan specifies the cross-tick liveness snapshot tuple as `(last_pty_activity_at, byte_offset)`, with `byte_offset` (raw transcript byte growth) as both the stage-1 corroborator and the stage-2 CAS precondition. The implementation substitutes `total_input_tokens` because `byte_offset` is not persisted onto `AgentSession`. The author has documented this candidly in the model comment and notes it must be resolved before stage-2 uses the tuple as a CAS precondition. Safe for this observe-only PR (the field is corroborator-only, never gating), but it is a latent correctness gap for the deferred stage-2.
**Severity:** tech_debt
**Fix:** Either persist a true `byte_offset` onto `AgentSession` from the transcript tailer (it already diff-gates `byte_offset` per the issue body), or formally re-specify the snapshot tuple to use a persisted progress field. Resolve before the stage-2 follow-up wires recovery.

**File:** `agent/granite_container/container.py:766` (`_cycle_idle` → `on_pty_read`)
**Code:** `if self._on_pty_read is not None:\n    try:\n        self._on_pty_read(buffer)`
**Issue:** The plan (Architectural Impact, Key Element 3) specifies `last_pty_read_loop_at` is "stamped on *every* PTY read-loop iteration, unconditionally" — i.e. the inner poll loop inside `read_until_idle`. The implementation fires `on_pty_read` only from `_cycle_idle`, which is invoked once per `read_until_idle` *return* — at turn boundaries, not per poll iteration. Consequence: during a long in-flight Dev turn, `read_until_idle` blocks (a healthy turn repaints at ≥1Hz and never satisfies the idle/quiescence gate), so neither `last_pty_read_loop_at` nor `last_pty_activity_at` updates mid-turn. For a genuinely wedged mid-turn screen that does NOT match the idle bar + prompt glyph, `read_until_idle` blocks until the 12h `CYCLE_IDLE_TIMEOUT_S` ceiling, so the loop marker goes stale → stage-1 three-state logic ABSTAINs → the wedge is never flagged. This under-fires (misses wedges) rather than false-fires, and path B does not recover in this PR, so it is not a blocker — but it directly affects whether the deferred stage-2 can detect anything, and contradicts the plan's stamp-frequency premise.
**Severity:** tech_debt
**Fix:** Stamp `last_pty_read_loop_at` from inside the `read_until_idle` poll loop (the `while time.monotonic() < deadline` body in `pty_driver.py`), not from `_cycle_idle`, so the loop-alive marker advances during a blocked-but-cycling read. Address in the stage-2 follow-up before recovery is wired; until then the observe-only counters will under-report.

### Nits

**File:** `agent/granite_container/container.py:575`
**Code:** `# PTY read-loop hook: called from _cycle_idle after each read_until_idle ... passing the normalized (ANSI-stripped) turn buffer.`
**Issue:** The container docstring says "normalized (ANSI-stripped)" while the bridge_adapter comment says "raw PTY buffer, not normalized bytes". Both are partly right (`turn_buffer` IS ANSI-stripped per `pty_driver.py:584`, but is not further normalized for cursor/spinner noise), but the two comments read as contradictory. The bridge_adapter side already flags the residual-normalization concern for stage-2.
**Severity:** nit
**Fix:** Align the two comments — "ANSI-stripped but not cursor/spinner-normalized" — so they don't appear to disagree.

### Verdict
No blockers. Path A is correct and ships its reproduced-incident fix cleanly. Path B (observe-only) has two tech_debt items that are honestly documented in-code and explicitly scoped to the stage-2 follow-up; both cause under-detection, not false kills, so they are safe to merge now but must be resolved before stage-2 wires recovery. Recommend routing to /do-patch for the two tech_debt items (or, if the team prefers, carrying them as explicit follow-up tasks on the stage-2 issue since they are already documented as deferred in the code and the feature doc).

tomcounsell (changes requested):
## Review: Changes Requested

**Branch:** `session/never_started_session_recovery` → `main`
**Plan:** `docs/plans/never_started_session_recovery.md`
**Tracking:** #1724
**Scope shipped:** Path A (never-started, active recovery) + Path B stage-1 (mid-run PTY quiescence, observe-only). Stage-2 classifier/recovery deferred.

Preflight clean (OPEN, MERGEABLE, mergeStateStatus=CLEAN). No UI files — visual-proof gate N/A. Ruff clean on all modified files. Unit suites green on this SHA: `test_never_started_recovery.py` (22) + `test_session_health_inference_removed.py` (13) = 35 passed.

This re-review is on the same HEAD SHA as the prior review; I independently re-verified all three findings against the checked-out code and they hold. No new code since the prior review, so the verdict is re-affirmed rather than changed.

### Path A assessment — solid
Path A is internally consistent and well-guarded. The shared `_never_started_past_grace` predicate gates all three intended call sites (30s-loop recovery branch, sub-check B fresh-heartbeat fast-path, Tier-2 reprieve escalation). Recovery passes `reason_kind="no_progress"` explicitly. Predicate fail-soft (never raises). No blockers.

### Tech Debt

**File:** `agent/session_health.py:2379`
**Code:** `byte_offset = getattr(entry, "total_input_tokens", None)  # best available proxy`
**Issue:** The plan specifies the cross-tick liveness snapshot as `(last_pty_activity_at, byte_offset)`, with `byte_offset` (raw transcript byte growth) as the stage-1 corroborator and stage-2 CAS precondition. The implementation substitutes `total_input_tokens` because `byte_offset` is not persisted on `AgentSession` (documented candidly at `models/agent_session.py:399-400`). Safe for this observe-only PR (corroborator-only — line 2377: "log but do not gate on it"), but a latent correctness gap for the deferred stage-2 CAS.
**Severity:** tech_debt
**Fix:** Persist a true `byte_offset` onto `AgentSession` from the transcript tailer, or formally re-spec the snapshot tuple to a persisted progress field. Resolve before stage-2 wires recovery.

**File:** `agent/granite_container/container.py:766` (`_cycle_idle` → `on_pty_read`)
**Code:** `if self._on_pty_read is not None:` … `self._on_pty_read(buffer)`
**Issue:** The plan (Key Element 3) specifies `last_pty_read_loop_at` is stamped "on *every* PTY read-loop iteration, unconditionally." The implementation fires `on_pty_read` only from `_cycle_idle`, invoked once per `read_until_idle` *return* (turn boundary), not per inner poll iteration (the `while time.monotonic() < deadline` loop at `pty_driver.py:527`). A genuinely wedged mid-turn screen that does not match the idle bar leaves the loop marker stale → stage-1 three-state logic ABSTAINs → the wedge is never flagged. Under-fires (misses wedges), does not false-fire; path B does not recover in this PR, so not a blocker — but it directly affects what the deferred stage-2 can detect and contradicts the plan's stamp-frequency premise.
**Severity:** tech_debt
**Fix:** Stamp `last_pty_read_loop_at` from inside the `read_until_idle` poll loop so the loop-alive marker advances during a blocked-but-cycling read. Address before stage-2 wires recovery.

### Nits

**File:** `agent/granite_container/container.py:575`
**Code:** `# PTY read-loop hook: called from _cycle_idle after each read_until_idle call, passing the normalized (ANSI-stripped) turn buffer.`
**Issue:** This container comment says "normalized (ANSI-stripped)" while `bridge_adapter.py:688` says "compares the raw PTY buffer, not normalized bytes." The two read as contradictory (the buffer IS ANSI-stripped but not cursor/spinner-normalized).
**Severity:** nit
**Fix:** Align both comments — "ANSI-stripped but not cursor/spinner-normalized."

### Verdict
No blockers. Path A ships its reproduced-incident fix cleanly. Path B (observe-only) carries two tech_debt items honestly documented in-code and in the PR's "Known Limitations" section, plus one comment-consistency nit. All three are under-detection / cosmetic, safe to merge behaviorally, but per SDLC policy tech_debt and non-subjective nits are patched rather than waived. Routing to /do-patch (or carry the two tech_debt items as explicit tasks on the stage-2 follow-up issue, since the code and feature doc already scope them as deferred).

tomcounsell (approved):
## Review: Approved

Reviewed against plan `docs/plans/never_started_session_recovery.md` and tracking issue #1724. This PR ships the explicitly-scoped partial: **Path A (never-started, active recovery)** + **Path B stage-1 (PTY quiescence detection, observe-only)**. Path B stage-2 (classifier + recovery wiring) is correctly deferred to a follow-up, as the plan's staging decision (D2-rollout) and Appetite scope-decision call for.

### What was validated

- **Shared grace, single source of truth** — `NEVER_STARTED_GRACE_SECS` (120) stays the sole canonical definition in `session_stall_classifier.py`; `session_health` imports it and adds only `NEVER_STARTED_CONFIRM_MARGIN_SECS` (30, env-overridable). Drift-pin test present (`TestNeverStartedGraceConstantAlignment`). Import direction one-way (advisory guard test green).
- **Path A both gates closed** — the shared `_never_started_past_grace(entry, now=None)` predicate (derives `now` internally so it binds in the two `now`-less callers) is consulted at all three sites: the new dedicated branch in the **30s** `_agent_session_tool_timeout_check` tick, sub-check B in `_has_progress`, and the `_tier2_reprieve_signal` escalation guard. This avoids the #1356 partial-fix shape (touching only one gate).
- **Cadence correct** — the never-started branch is hosted in the 30s loop (not the 300s main loop), so recovery fires at ~125-155s rather than being unreachable below the 300s floor. The 300s floor and `:1708` elif are left untouched for output-producing sessions.
- **Race mitigation** — the branch re-reads fresh, re-confirms the predicate, guards terminal status, increments the counter fail-silently, and routes through the shared `_apply_recovery_transition` with the `"no progress signal"` substring so `_reason_kind` resolves to `no_progress`.
- **Path B stage-1** — four nullable `AgentSession` fields added (`_heal_descriptor_pollution` handles them generically, no backcompat). `_eval_mid_run_pty_stage1` implements the three-state PTY model (abstain on `None`/stale loop marker, clear on fresh activity, accumulate `mid_run_quiescent_since` on quiescence) as durable cross-tick state, not an in-memory counter — so a worker bounce cannot reset progress. Writers in container/bridge_adapter are fail-silent and never raise into the read loop.
- **Honesty about scope** — every known limitation is documented in the code (`total_input_tokens` byte_offset proxy, turn-boundary stamp frequency, partial ANSI normalization), the PR body's "Known Limitations", and the feature doc's "Stage-2 (Planned — Not Yet Shipped)" section. None is a defect in the shipped observe-only scope.

### Tests & gates

- `test_never_started_recovery.py` (22) + `test_session_stall_classifier.py` + `test_session_health_inference_removed.py` + `test_session_health_compacting_reprieve.py` + `test_stall_advisory_e2e.py` — all green (74 + 5 + 9).
- ruff check + format: clean on all modified files.
- Plan verification grep checks pass (predicate match count, reason string, 30s-loop hosting, grace defined once, both PTY fields present, etc.).
- No UI files changed — screenshot gate is a no-op.

### Blockers
- None

### Tech Debt
- None

### Nits
- None

Approving: clean preflight (OPEN, MERGEABLE, CLEAN), zero findings, scope honestly matches the plan's staged Path A + Path B stage-1 deliverable.

PR #1739, opened 2026-06-18
fix(#1730): deferred delivery fallback on session failure

### LINKED ISSUE #1730: deferred delivery lost when tool_timeout kills session: no fallback when self-draft steering is pending
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/deferred_delivery_fallback_on_failure.md

> **Context:** This issue involves two components: the **session executor** ([`agent/session_executor.py`](agent/session_executor.py)), an asyncio coroutine that runs a Claude SDK session in the psyoptimal repo and manages its lifecycle; and the **empty-promise self-draft mechanism** ([`agent/output_handler.py`](agent/output_handler.py), [`agent/steering.py`](agent/steering.py)), which detects when the agent produces output that acknowledges a request without answering it, defers delivery, and injects a steering message asking the agent to rewrite. When this mechanism fires but the session is then killed by the health checker ([`agent/session_health.py`](agent/session_health.py)) before the rewrite turn runs, the user receives no message at all.

## Problem

When a session's delivery is deferred via the self-draft mechanism (empty-promise detected, outbox write skipped, steering message injected) and the session is subsequently killed by the health checker due to a `tool_timeout`, no message is ever delivered to the user. The deferred delivery state is stored only as a local variable in `output_handler.py` — it is never persisted on the `AgentSession` record. When the health checker cancels the session task, this state is silently lost.

A second compounding bug causes the session to be unnecessarily re-run: after delivering its self-draft message and saving a `complete` snapshot, the asyncio future held by the worker loop is never resolved, leaving the session in `running` state indefinitely until the health checker forcibly recovers it. This spurious recovery re-runs a session that already completed, which then hits the empty-promise gate again and ultimately fails with no delivery.

**Current behavior:**
1. Agent produces empty-promise output → delivery deferred, self-draft steering injected
2. Session completes its SDK turn → saves `complete` snapshot at 07:19:20 → one message delivered ✓
3. Worker future never resolves → session stays `running` for 32 min (`log_path=None, claude_session_uuid=None`)
4. Health checker recovers session (attempt 1, `no_progress`) → re-runs it
5. Re-run produces another empty promise at 07:53:47 → delivery deferred again
6. Bash tool wedge (>300 s) → health checker kills session as `failed` (attempt 2, `tool_timeout`)
7. **User receives no message from the final run. 0 deliveries.**

**Desired outcome:**
- When a session is finalized as `failed` (or `killed`) with a pending self-draft steering message or `steering_deferred` state, a fallback delivery is attempted (narration-gate text or an apology message) before the session is closed.
- The worker future leak is fixed so a session that completes its SDK work is reliably transitioned to `completed` without waiting for a 300 s health-check timeout.

## Definitions

| Empty-promise detection | Heuristic (`_detect_empty_promise`) that flags output where the agent acknowledges a request ("I'll look into that") without providing substantive content, triggering the self-draft mechanism | [`bridge/promise_gate.py:339`](bridge/promise_gate.py) |
| Self-draft steering | When empty-promise fires, delivery is deferred and a steering message is injected into the session's Redis steering queue asking the agent to rewrite; the agent consumes it on the next SDK turn | [`agent/output_handler.py:719`](agent/output_handler.py), [`agent/steering.py:184`](agent/steering.py) |
| `steering_deferred` | Local boolean in `output_handler.send()` set to `True` when self-draft steering is injected; causes the outbox write to be skipped; **not persisted on AgentSession** | [`agent/output_handler.py:365`](agent/output_handler.py) |
| Tool timeout / `tool_timeout` | Health-checker recovery kind triggered when a session's current tool call has been running longer than the tier-specific threshold (default tier: 300 s); recorded in `agent/session_health.py` | [`agent/session_health.py:2190`](agent/session_health.py) |
| Worker future | The `asyncio.Task` created by the worker loop per session; resolves when `_execute_agent_session()` returns or raises; the session stays `running` in Redis until this task resolves | [`agent/agent_session_queue.py:1236`](agent/agent_session_queue.py) |
| `steering:attempts:{session_id}` | Redis counter tracking consecutive self-draft injection attempts; budget cap is `SELF_DRAFT_MAX_ATTEMPTS = 2` (returns False after 3rd attempt); relies on 1-hour TTL for cleanup, never explicitly deleted on failure | [`agent/steering.py:179`](agent/steering.py) |

## Prior Context

Related work:
- [#1680](https://github.com/tomcounsell/ai/issues/1680) / [PR #1685](https://github.com/tomcounsell/ai/pull/1685): Repositioned message drafter from LLM rewriter to pass-through validation filter — introduced the current empty-promise / self-draft pattern
- [#1219](https://github.com/tomcounsell/ai/issues/1219): Audit to prevent false promises across all delivery paths — self-draft mechanism was the resolution
- [#867](https://github.com/tomcounsell/ai/issues/867): Race between nudge re-enqueue and `finalize_session()` — related race family in the session lifecycle
- [#875](https://github.com/tomcounsell/ai/issues/875): Session lifecycle CAS authority — root-cause fix for the race family; the worker-future leak may be a surviving instance

## Solution Sketch

**Bug B (delivery fallback — higher priority, user-facing):**
- In `session_health.py`'s finalization path (both `tool_timeout` and `no_progress` → `failed`), after transitioning status to `failed`, call `pop_all_steering_messages(session_id)` and check if any were from `drafter-fallback`. If so, deliver the narration fallback message via the output handler's `_apply_narration_fallback()` path before closing the session.
- Alternatively, persist a `delivery_deferred` boolean field on `AgentSession` (set when `steering_deferred=True` in `output_handler.py`), and check it in the health-checker finalization path.

**Bug A (worker future leak — reliability):**
- Audit `_worker_loop()` in `agent_session_queue.py` and `BackgroundTask` in `messenger.py` to identify why the asyncio task completes without the future resolving. The session executor saves a `complete` snapshot and the output handler writes to the outbox, but the lifecycle transition to `completed` is not reflected in the worker's future.
- Specifically: does `_execute_agent_session()` return normally after `session_executor.py`'s finally block, or does an exception from the unconsumed-steering re-enqueue path swallow the return?

**Scope constraint:** The fix must not introduce a new mandatory field on `AgentSession` that requires a migration (additive nullable field is fine).

## Recon Summary

**Confirmed: 5 items**
- `steering_deferred` is a local variable in `output_handler.send()` — not persisted on AgentSession ([`output_handler.py:365`](agent/output_handler.py))
- `session_health.py` finalization path has no check for pending steering or deferred delivery state ([`session_health.py:2190`](agent/session_health.py))
- `steering:attempts` Redis counter has no cleanup on session failure — relies on 1-hour TTL only ([`steering.py:207`](agent/steering.py))
- Steering cleanup in executor's finally block runs even after cancellation, BUT only re-enqueues if messages exist — the pending self-draft steering message was consumed before the tool wedge started
- Worker loop guard for "running after complete" only fires on crash path (`finalized_by_execute=False`), not on the happy path ([`agent_session_queue.py:1537`](agent/agent_session_queue.py))

**Revised: 1 item**
- Initial assumption was that `pop_all_steering_messages()` would not run if the task was cancelled mid-execution — confirmed it IS in a `finally` block and DOES run after cancellation. The fallback gap is specifically that the finalization path doesn't check steering state before declaring failure.

**Pre-requisites: 0**

**Dropped: 1 item**
- Fixing the `StatusConflictError` on continuation re-enqueue (logged at 07:53:49) — this is a symptom of the underlying race, not a root cause. Addressing it separately from the delivery fallback would add scope without solving the user-visible problem.

## Acceptance Criteria

- [ ] When a session is finalized as `failed` with a pending `drafter-fallback` steering message in its Redis queue, a fallback message is delivered to the user (narration fallback or explicit "failed to respond" notice) before the session closes
- [ ] When a session's executor completes its SDK run and saves a `complete` snapshot, the worker future resolves within 10 s (no 32-min ghost `running` state)
- [ ] Unit/integration test covering: session produces empty-promise → self-draft steering injected → session task cancelled (tool_timeout) → fallback message delivered
- [ ] `steering:attempts:{session_id}` Redis key is deleted when a session transitions to a terminal state (`failed`, `killed`, `completed`)

## Downstream

This issue feeds into `/do-plan`, which will produce a plan document at `docs/plans/deferred_delivery_fallback_on_failure.md`. The plan is then executed by `/do-build`.

The fix should be conservative: Bug B (delivery fallback) is the user-facing regression and should be addressed first. Bug A (worker future leak) is a reliability issue that caused the spurious re-run — it should be in the same plan but can be a separate commit.

tomcounsell:
## Summary

- **Bug B (user-facing)**: When the message drafter flags an empty-promise output and injects a deferred self-draft steering message, the original text is now persisted to `AgentSession.extra_context` (`deferred_self_draft_pending` + `deferred_self_draft_text`) at defer time. A new `_deliver_deferred_self_draft_fallback()` helper in `session_health.py` reads this flag on all three terminal recovery branches (`failed` ×2, `abandoned`/`no_progress` ×1) and delivers the recovered text (narration-gated) before finalizing. Self-draft fallback takes precedence over the generic degraded notice.
- **Bug B, AC4**: `reset_self_draft_attempts(session_id)` now fires unconditionally on every terminal finalize (Seat A in `finalize_session`, outside `emit_telemetry` guard) and on health-checker terminal paths via Seat B next to `finalize_telemetry` — replacing the 1-hour TTL.
- **Bug A (reliability)**: Investigation confirmed `complete_transcript()` already fires on the deferred-self-draft path. The 32-min ghost `running` state is caused by `complete_transcript()` throwing an unexpected exception whose except handler logs-and-swallows without finalizing. A defensive `finalize_session()` fallback is added in the except block (completion/delivery exit ONLY, guarded by `get_authoritative_session` re-read + `StatusConflictError` handling).

## Test plan

- [ ] `pytest tests/unit/test_mcp_hang_graceful_degradation.py` — 40 tests pass (24 new)
- [ ] `pytest tests/unit/test_output_handler.py` — 68 tests pass (3 new)
- [ ] `pytest tests/unit/test_steering.py` — 9 tests pass (4 new)
- [ ] `pytest tests/integration/test_bug_a_worker_future_leak.py` — integration regression tests for Bug A lifecycle contracts
- [ ] Verify: `grep -n "_deliver_deferred_self_draft_fallback" agent/session_health.py` shows 4+ lines
- [ ] Verify: `grep -n "reset_self_draft_attempts" models/session_lifecycle.py` shows Seat A outside `emit_telemetry`
- [ ] Verify: no new mandatory `AgentSession` field added

Closes #1730


tomcounsell:
## Review: Approved

Reviewed PR #1739 against plan `docs/plans/deferred_delivery_fallback_on_failure.md` and issue #1730.

**Preflight:** OPEN · MERGEABLE · CLEAN. No UI files in the diff (visual-proof gate is a no-op).

### Plan alignment — all acceptance criteria implemented and verified

- **AC1 — fallback delivery on terminal branches:** `_deliver_deferred_self_draft_fallback(entry)` (`agent/session_health.py`) is wired into all three terminal recovery branches — the local `abandoned`/`no_progress` branch and both `failed` branches — with precedence over `_deliver_tool_timeout_degraded_notice` (generic notice suppressed when `deferred_self_draft_pending` is set). Distinct SETNX lock `self_draft_fallback_sent:{sid}` guarantees at-most-once and never collides with the degraded-notice lock. Detection reads the persisted `extra_context` flag, never the steering queue (which is empty by finalization) — the war-room blocker is correctly resolved.
- **AC2 — worker-future leak:** the Bug A fix in `agent/session_executor.py` is correctly scoped to the completion/delivery exit's `except` block only. It is gated on `not chat_state.defer_reaction` (excludes the nudge re-enqueue path), re-reads via `get_authoritative_session`, guards on `status == "running"`, routes through `finalize_session` (not `transition_status`, which raises on terminal targets), and treats `StatusConflictError` as success. No #867/#875 nudge-stomp risk.
- **Persist at defer time:** `agent/output_handler.py` persists `deferred_self_draft_pending` + `deferred_self_draft_text` into `extra_context` with a safe RMW (`get_authoritative_session` re-read before merge), logs persist failures at WARNING rather than swallowing. No new mandatory field, no migration.
- **AC4 — counter cleanup (dual-seat):** Seat A in `models/session_lifecycle.py::finalize_session` sits OUTSIDE the `emit_telemetry` guard (fires on every terminal finalize incl. happy-path `completed`); Seat B in `session_health.py` next to `finalize_telemetry` covers the `emit_telemetry=False` health-checker terminals. Both best-effort. This correctly resolves revision-2 blocker 1.

### Verification

- `tests/unit/test_mcp_hang_graceful_degradation.py` + `test_output_handler.py` + `test_steering.py`: **117 passed**.
- `tests/integration/test_bug_a_worker_future_leak.py`: **10 passed** (includes nudge-stomp and CancelledError-contract regressions).
- `ruff check` + `ruff format --check` on all four changed source files: clean.
- Plan verification table: fallback wired (5 refs), Seat A/B present, Bug A guard present, no new mandatory field — all pass.

### Nit (subjective — not blocking)

- `docs/features/session-lifecycle.md` and the plan reference `_apply_narration_fallback()`, but the new free-function helper in `session_health.py` inlines the equivalent `is_narration_only` + `NARRATION_FALLBACK_MESSAGE` logic (it can't call the instance method `_apply_narration_fallback` on `TelegramRelayOutputHandler`). Behavior is equivalent; the doc wording is a minor naming-precision nit only.

Clean, well-scoped, faithful to a heavily-critiqued plan. Approving with no substantive findings.

PR #1694, opened 2026-06-15
Granite PTY: Persona-as-Priming Refactor

valorengels:
## Summary

Closes #1692

Retires `--append-system-prompt` as the persona-injection mechanism for the Granite PTY container and replaces it with prime slash commands that install persona inside the TUI at startup.

- **Canary spike** (`tests/integration/test_append_system_prompt_interactive.py`): empirically confirmed `--append-system-prompt` IS honored in the interactive TUI (RESULT: HONORED), providing the factual baseline for this refactor
- **Prime command rewrites** (`.claude/commands/granite/prime-pm-role.md`, `prime-dev-role.md`): PM routes and delegates; Dev owns the full SDLC pipeline and runs `/do-*` skills directly
- **New files**: `prime-teammate-role.md` (Teammate persona for granite PTY), `_prime-rails.md` (shared WORKER safety rails referenced by all three primes), `config/personas/teammate.md` (now repo-tracked, removed from `.gitignore`)
- **Deleted path**: `--append-system-prompt` removed from `pty_driver.py` spawn args (deprecated with warning for backward compat); `pm_system_prompt` field removed from `BridgeAdapter` and `PairSpawnSpec`; persona overlay resolution block (~100 lines) removed from `session_executor.py`
- **Dev model**: `GRANITE__DEV_MODEL` default flipped from `sonnet` to `opus` — Dev now owns full SDLC pipeline execution
- **Dev receives user message**: `include_user_message=True` for Dev PTY; anti-self-start guard moved to prime text ("wait for `[/dev]` relay")
- **Documentation**: updated `granite-pty-production.md`, `composed-persona-system.md` (retirement notice), `email-bridge.md` (persona resolution section), `personas.md`

## Test plan

- [ ] `python -m pytest tests/unit/granite_container/ -q` — all 287 pass, 5 skipped
- [ ] `python -m ruff format --check` and `ruff check` — clean on all changed files
- [ ] Verify `settings.granite.dev_model == 'opus'`
- [ ] Verify `prime-teammate-role.md` exists and references `_prime-rails.md`
- [ ] Verify no `--role dev` / `--role pm` flags remain in agent/bridge/tools
- [ ] Verify `compose_system_prompt` still intact in `sdk_client.py` (non-granite harness path preserved)

PR #1742, opened 2026-06-19
fix(granite): prevent canned OPERATOR_TERMINAL_MESSAGE via prefix contract + floor

### LINKED ISSUE #1719: Bridge sessions return canned fallback instead of real responses — targeted routing fix + Stop-hook completion floor
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/granite_routing_prefix_floor.md

> **Context for the reader.** Telegram messages are handled by an "Eng" *granite session*: the worker drives a real interactive Claude Code TUI inside a PTY (the *granite container*, `agent/granite_container/container.py`). Two personas converse in that container — a **PM** (talks to the human) and a **Dev** (does the work). For the PM's output to reach the user, the PM must begin a line with a **routing prefix** (`[/user]`, `[/complete]`, or `[/dev]`); a small local *ollama classifier* detects that prefix and sets `user_facing_routed=True`, which is the ONLY thing that causes a message to be delivered back to Telegram. If no prefix is detected, the container ends with `exit_reason=pm_no_user_message` and the bridge delivers a canned fallback instead.

## Problem

**Current behavior:** Every incoming Telegram message comes back as the canned string:

> "I wasn't able to produce a response to this — please rephrase or follow up."

The bridge is effectively unusable. Sessions run for ~10 minutes and then deliver this fallback.

**Why:** `#1694` ("Granite PTY: Persona-as-Priming Refactor", commit `971b77d6`, merged ~2026-06-15) **deleted the `--append-system-prompt` persona path** (touching `agent/session_executor.py`, `agent/sdk_client.py`, `agent/granite_container/pty_driver.py`, `pty_pool.py`, `bridge_adapter.py`) and moved the routing-prefix contract into the `/prime-*` TUI slash commands. In production the PM no longer reliably emits a routing prefix. Because delivery is **100% gated** on prefix detection (`container.py:1158`, `_successful_exits` + `user_facing_routed`), every turn falls through to `OPERATOR_TERMINAL_MESSAGE` (`container.py:231`).

**Desired outcome:** A real Telegram message round-trips a genuine response. The canned fallback only appears on a genuinely empty turn.

### Evidence

- Worker logs since ~2026-06-15 17:14 UTC: **24× `exit_reason=pm_no_user_message`** ("reached max_turns=10 without a `[/complete]`") + **6× `pm_hang`**, and **0× `pm_complete` / `pm_user`**. Anomaly rate jumped from 1–2/day to 10/day on 06-15 — the day `#1691` (PM/Dev→Eng role merge) and `#1694` landed.
- `logs/sessions/tg_valor_-1003449100931_983/`: `resume`→`complete` pairs ~10 min apart with `messages: []` (empty), then a `crash.json`. The "~10 minutes" is exhausting `max_turns=10` PM↔Dev cycles — **not** a timeout. Priming itself succeeds (`prime post-write wait saw_idle=True`).
- `logs/worker/tg_valor_-1003449100931_983.log`: the canned fallback delivered to chat `-1003449100931`.

## Definitions

| Term | Definition |
|------|------------|
| Granite container | The worker component that drives a real interactive Claude Code TUI over a PTY (`agent/granite_container/container.py`). |
| Routing prefix | A line-leading token (`[/user]`, `[/complete]`, `[/dev]`) the PM emits so its output is routed to the user or to the Dev. |
| `user_facing_routed` | `ContainerResult` flag, True only when the classifier detected a `[/user]`/non-empty `[/complete]` payload. Gates delivery. |
| `OPERATOR_TERMINAL_MESSAGE` | The canned fallback at `container.py:231` delivered when no user-facing message was routed. |
| Stop hook | Claude Code's deterministic turn-boundary hook that fires when a turn completes (independent of model output content). |

## Solution sketch — TARGETED changes, NOT a revert

A blanket `git revert` of `#1694` is the wrong tool: it spans 26 files, deletes a whole subsystem, and three later shipped commits (`#1710`, `#1708`, `#1663`) build on those same files. Instead, make two complementary targeted changes:

1. **Restore reliable routing-prefix emission** via a minimal change to the prime path (`/prime-pm-role` + `_prime-rails`), or by restoring just the load-bearing prefix-contract instruction that the deleted system-prompt path used to guarantee.
2. **Add a Stop-hook completion floor** so delivery no longer depends on the model perfectly emitting a prefix: when the Stop hook fires (turn finished) with nothing routed, deliver the session's **last assistant message** instead of the canned fallback. The prefix becomes an *optional routing hint*, not the delivery gate. (Consistent with the established "hooks as a control surface" direction — the Stop hook is a deterministic completion signal.)

### Open questions for `/do-plan`

- Is the regression that the prime command *fails to load*, or that it loads but *no longer instructs* the prefix contract strongly enough? (Read a real PM PTY transcript to confirm before choosing change #1's shape.)
- Where should the Stop-hook floor live — inside the container's wrap-up guard, or in `bridge_adapter`? What is "the last assistant message" in PTY terms (transcript tail vs. captured buffer)?
- Should `pm_max_turns` be removed from `_successful_exits` (`container.py:1158`) as part of this, or deferred to the test-hardening issue?

## Acceptance criteria

- [ ] A real Telegram message to the Eng session round-trips a genuine, non-canned response.
- [ ] Granite exits `pm_complete`/`pm_user` with `user_facing_routed=True` for normal turns.
- [ ] `OPERATOR_TERMINAL_MESSAGE` fires only on a genuinely empty turn (no assistant output at all).

## Related / follow-up (separate issue)

A test-coverage gap let this ship silently — to be filed as its own issue:
- `tests/integration/test_granite_container_loop.py` (the only real-loop test) is `@skipUnless(model_reachable)` and silently skips → green.
- `tests/integration/test_granite_pty_production.py` uses a mocked PTY emulator that emits compliant prefixes by construction.
- The integration test accepts `pm_max_turns` as a passing exit; its accepted exit set is stale (missing `pm_no_user_message`/`pm_user`).
- Nightly regression is not scheduled (no launchd plist, no `logs/nightly_tests.log`).
- Hardening: an env-gated **real** test asserting a non-empty user message reaches the relay regardless of `exit_reason`; drop `pm_max_turns` from `_successful_exits`; schedule the nightly so a skip *alerts* rather than passes.

## Downstream context

This issue will be consumed by `/do-plan` to produce a plan at `docs/plans/{slug}.md`. **Do not implement yet** — another agent is currently working in this checkout (`#1539` crash-signature). Diagnosis + plan only.

---

### Practice 9 — Sticky-`failed` against trailing PTY idle

Issue #1732 confirmed this against Omnigent's production implementation:

- **Omnigent file:line (pinned 2026-06-18; re-verify on revisit):** `_publish_status` keeps `failed` sticky so a `StopFailure` isn't overwritten by the trailing PTY idle that follows it at `claude_native_forwarder.py:97-99`; forward loop at `:2161-2237`.
- **Our side:** Our completion detection has no failure-vs-success distinction at the edge. A `StopFailure` (in-turn error, clean hook exit) followed by PTY quiescence could be misread.

**Delta for the Stop-hook completion floor**: The floor must respect a **sticky failure** — once `StopFailure` is detected, that status must not be overwritten by subsequent PTY idle state. A `StopFailure` hook fires when Claude Code reports and cleanly exits from an in-turn error; the PTY still goes quiet after it. The floor's status machine must treat `failed` as terminal once set.

---
*Captured by issue #1732 (reference appendix). Omnigent citation pinned to their HEAD at 2026-06-18; re-verify on revisit.*
---

## Recon Summary

*Performed 2026-06-19 against `main` @ `c79262e8` (validity review before planning). Tests run + cited code re-read on the skills-only machine.*

**Confirmed:** (claims re-verified, still true)
- **Prefix-gated delivery still in force.** `container.py:1219` `_successful_exits = {"pm_complete","pm_user","pm_max_turns"}`; `:1220` flags anomaly when exit is successful-shaped but `user_facing_routed is False`. Canned `OPERATOR_TERMINAL_MESSAGE` delivered at `:1555-1559` when nothing routed.
- **#1694 (`971b77d6`) deleted the `--append-system-prompt` persona path** — confirmed in container.py git log within the regression window.
- **`pm_max_turns` is still in `_successful_exits`** (`container.py:1219`).
- **Test gap reproduced.** `tests/integration/test_granite_container_loop.py::test_cli_short_run_produces_results_json` **SKIPS** (`@skipUnless(_model_reachable)`, gated on `claude --print ping`, not ollama). `tests/integration/test_granite_pty_production.py` **passes** with a mock PTY emitting compliant prefixes by construction. Neither would catch this regression.

**Revised:** (claim framing corrected)
- **A last-assistant-message floor already partially exists.** `_run_wrapup_guard` (`container.py:1475`, introduced in **#1651 / `d005aaa2`** — *predates* the regression) reads `last_assistant_text(pm_transcript)` and re-drives the PM once (`MAX_WRAPUP_ATTEMPTS=1`). **But** it pipes the result through `classify_pm_prefix` + `user_facing_routed` (`:1539,:1549`), so a no-prefix wrap-up response *still* falls to canned. → Solution-sketch #2 must be **"RELAX the existing wrap-up floor"** (deliver `last_assistant_text` unconditionally when present; reserve canned for a genuinely empty transcript), NOT build a floor from scratch.

**Pre-requisites:** (constraints)
- **Sticky-`failed`** (per #1732/Omnigent appendix): a `StopFailure` status must not be overwritten by the trailing PTY idle that follows it. The relaxed floor's status machine must treat `failed` as terminal once set.

**Dropped:** (out of scope)
- **#1739 (deferred delivery fallback, shipped 2026-06-19) does NOT fix this — orthogonal.** It acts at the message-drafter / health-checker layer (`session_health.py`, `output_handler.py`) on session *failure/abandonment*. Granite delivers `OPERATOR_TERMINAL_MESSAGE` with `user_facing_routed=True` and exits **cleanly**, so the session never registers as failed and #1739's path never fires.

### Unverifiable on this machine
- Worker-log evidence (24× `pm_no_user_message`, 0× `pm_complete`) — this is the skills/tools-only machine with no bridge/worker sessions. Claim accepted from issue author's production logs.

tomcounsell:
## Summary

Fixes the critical regression (#1719) where every Telegram message to the granite Eng session returns the canned string "I wasn't able to produce a response to this..." instead of a real response.

Root cause: Two compounding failures downstream of #1694:
1. Per-turn prefix contract decay — the PM only receives the routing prefix contract once at session start; without per-turn re-assertion it drifts off-prefix after ~10 turns
2. Wrap-up floor re-gates on the prefix — a non-empty but prefix-less PM wrap-up response was discarded instead of delivered

## Changes

- **Change 1 (primary fix):** Add `PM_TURN_CONTRACT_REMINDER` constant and append it to every Dev→PM handoff write in the steady-state loop (`container.py:1428`). The PM now receives "Begin your reply with [/user], [/complete], or [/dev] on its own line." on every turn, preventing contract decay across multi-turn cycles.

- **Change 2 (defense in depth):** Relax `_run_wrapup_guard` floor — when PM produces a non-empty but prefix-less final message, deliver it directly via `_on_user_payload(pm_text.strip())` with `exit_reason="pm_floor_delivered"` instead of firing `PM_COMPLIANCE_NUDGE` into the dying PTY and falling to `OPERATOR_TERMINAL_MESSAGE`. Canned fallback now fires only on a genuinely empty PM transcript.

- **Change 2b (mandatory):** Add `"pm_floor_delivered"` to `_CLEAN_GRANITE_EXIT_REASONS` in `session_executor.py` so floor-delivered sessions get `REACTION_COMPLETE` instead of `REACTION_ERROR`.

## Testing

- [x] New unit test file `tests/unit/granite_container/test_wrapup_guard_floor.py` — 13 tests covering all floor branches + PM_TURN_CONTRACT_REMINDER assertion + `_is_non_clean_granite_exit()` assertion for `pm_floor_delivered`
- [x] Updated `test_container.py` — split `test_terminal_message_sent_when_pm_silent` into two precise behavioral tests
- [x] All 363 granite unit tests passing, 5 skipped

## Documentation

- [x] `docs/features/granite-pty-production.md` updated — new per-turn contract reminder section, three-way floor logic documented, `pm_floor_delivered` added to exit reason reference, clean markers updated
- [x] `ContainerResult` docstring updated to include `pm_floor_delivered`
- [x] `_is_non_clean_granite_exit` docstring updated

## Definition of Done

- [x] Tested: 363 unit tests passing
- [x] Reviewed: Build validator + docs gate passing

Closes #1719

### DIFF
--- a/agent/granite_container/container.py
+++ b/agent/granite_container/container.py
@@ -236,6 +236,16 @@ PM_WRAPUP_PROMPT = (
 # OPERATOR_TERMINAL_MESSAGE instead.
 MAX_WRAPUP_ATTEMPTS = 1
 
+# Per-turn routing-prefix contract reminder appended to the Dev-report
+# handoff write so the PM cannot lose the contract across turns. One
+# short sentence — intentionally brief so it costs minimal tokens and
+# does not distract the PM from the Dev report content. Restores the
+# per-turn guarantee that was lost when #1694 deleted the
+# --append-system-prompt path.
+PM_TURN_CONTRACT_REMINDER = (
+    "\n\nBegin your reply with [/user], [/complete], or [/dev] on its own line."
+)
+
 # Fallback seed string for the wrap-up prompt when the developer did
 # not produce a captured report and the Dev PTY is no longer readable.
 DEV_REPORT_UNAVAILABLE = "The developer did not produce a captured report."
@@ -277,8 +287,11 @@ class ContainerResult:
     """Final output of a container run.
 
     `exit_reason` is one of: pm_complete, pm_user, pm_max_turns,
-    dev_hang, pm_hang, startup_unresolved, pm_no_user_message,
-    exception. The worker renders this as the run's terminal verdict.
+    pm_floor_delivered, dev_hang, pm_hang, startup_unresolved,
+    pm_no_user_message, exception. The worker renders this as the
+    run's terminal verdict. pm_floor_delivered means the wrap-up
+    floor delivered PM's final text directly (non-empty but
+    prefix-less); it is a clean exit (user got a real message).
 
     `user_facing_routed` is True when at least one [/user] or
     non-empty [/complete] payload was delivered to the user channel
@@ -1216,7 +1229,10 @@ class Container:
             # successful-shaped terminal state but PM never delivered a
             # user-facing message, drive PM to produce one. This
             # guarantees the human always receives some output.
-            _successful_exits = {"pm_complete", "pm_user", "pm_max_turns"}
+            # Failure exit_reasons (dev_hang, pm_hang, exception, startup_unresolved,
+            # pm_no_user_message) bypass this gate structurally — they are never in
+            # _successful_exits. No runtime sticky-failed guard is needed here.
+            _successful_exits = {"pm_complete", "pm_user", "pm_max_turns", "pm_floor_delivered"}
             if result.exit_reason in _successful_exits and not result.user_facing_routed:
                 self._run_wrapup_guard(result)
 
@@ -1409,7 +1425,7 @@ class Container:
         if not await_pm:
             result.exit_message = "PM did not reach idle before Dev report"
             return RouteOutcome(should_break=True, exit_reason="pm_hang")
-        self._pm_pty.write(dev_text)
+        self._pm_pty.write(dev_text + PM_TURN_CONTRACT_REMINDER)
 
         turn_record = TurnRecord(
             turn_index=turn_index,
@@ -1537,18 +1553,39 @@ class Container:
                 )
                 if pm_text:
                     wrapup_classification = classify_pm_prefix(pm_text)
+                    if wrapup_classification.destination in ("user", "complete", "dev"):
+                        # PM produced a compliant prefix — route normally.
+                        outcome = self._route_pm_classification(
+                            wrapup_classification, pm_buf, turn_index=-2, result=result
+                        )
+                        if result.user_facing_routed:
+                            result.exit_reason = outcome.exit_reason or result.exit_reason
+                            return
+                    else:
+                        # Non-empty but prefix-less — deliver pm_text directly
+                        # WITHOUT calling _route_pm_classification (which would
+                        # write PM_COMPLIANCE_NUDGE into the dying PTY at :1286).
+                        if self._on_user_payload is not None:
+                            try:
+                                self._on_user_payload(pm_text.strip())
+                                result.user_facing_routed = True
+                                result.exit_reason = "pm_floor_delivered"
+                                logger.info(
+                                    "[granite-container] wrap-up floor: "
+                                    "delivered prefix-less PM text (%d chars)",
+                                    len(pm_text),
+                                )
+                                return
+                            except Exception as e:
+                                logger.warning(
+                                    "[granite-container] wrap-up floor: direct delivery failed: %s",
+                                    e,
+                                )
                 else:
                     _log_transcript_read_diagnostic(
                         "wrap-up guard", pm_transcript, self._pm_pty, self._dev_pty
                     )
                     result.transcript_fallback_count += 1
-                    wrapup_classification = _unknown_classification()
-                outcome = self._route_pm_classification(
-                    wrapup_classification, pm_buf, turn_index=-2, result=result
-                )
-                if result.user_facing_routed:
-                    result.exit_reason = outcome.exit_reason or result.exit_reason
-                    return
 
             # PM still silent after MAX_WRAPUP_ATTEMPTS — deliver canned
             # terminal message directly so the human always gets something.
--- a/agent/session_executor.py
+++ b/agent/session_executor.py
@@ -32,16 +32,18 @@ from models.session_lifecycle import TERMINAL_STATUSES as _TERMINAL_STATUSES
 
 logger = logging.getLogger(__name__)
 
-_CLEAN_GRANITE_EXIT_REASONS = frozenset({"pm_complete", "pm_user"})
+_CLEAN_GRANITE_EXIT_REASONS = frozenset({"pm_complete", "pm_user", "pm_floor_delivered"})
 
 
 def _is_non_clean_granite_exit(agent_session) -> bool:
     """Return True when the session has a granite exit_reason that signals a real failure.
 
     None exit_reason = non-granite session or not yet set = clean (default behavior).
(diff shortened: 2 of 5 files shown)


tomcounsell:
Superseded by #1743, which merged as `a3e63772` and closed #1719. #1743 is a superset of this branch's changes (same `PM_TURN_CONTRACT_REMINDER` + `pm_floor_delivered` floor, plus the integration test `test_granite_pty_production.py`, `test_session_executor_granite.py`, and additional doc updates that this PR lacked). Closing this duplicate — a parallel build branch created during the same forked SDLC run. No unique work is lost.

PR #1743, opened 2026-06-19
fix(#1719): per-turn routing-prefix reminder + relaxed wrap-up floor

### LINKED ISSUE #1719: Bridge sessions return canned fallback instead of real responses — targeted routing fix + Stop-hook completion floor

tomcounsell:
## Summary

Fixes the bridge session canned-fallback regression (#1719) where every Telegram message was returning `OPERATOR_TERMINAL_MESSAGE` ("I wasn't able to produce a response to this — please rephrase or follow up."). Root cause: two compounding failures downstream of #1694's deletion of `--append-system-prompt`.

- **Change 1 (primary):** Add `PM_TURN_CONTRACT_REMINDER` constant and append it to the Dev-report text on every steady-state `[/dev]` handoff at `container.py:~1412`. Restores the per-turn prefix-contract assertion that prevented PM prefix drift across PM↔Dev cycles. Targets the 24× `pm_no_user_message` production exits since 2026-06-15.
- **Change 2 (defense in depth):** Relax `_run_wrapup_guard` to floor-deliver the PM's real last message when it is non-empty but prefix-less, instead of the canned fallback. New `exit_reason=pm_floor_delivered` bypasses `_route_pm_classification` to avoid `PM_COMPLIANCE_NUDGE` on a dying PTY. `OPERATOR_TERMINAL_MESSAGE` is now reserved for a genuinely empty PM transcript.
- **Change 2b (mandatory):** Add `pm_floor_delivered` to `_CLEAN_GRANITE_EXIT_REASONS` in `session_executor.py` so floor-delivered sessions receive `REACTION_COMPLETE` instead of `REACTION_ERROR`.

## Test plan

All tests run on the skills/tools-only machine (no real PTY, no Telegram):

- [x] `tests/unit/granite_container/test_container.py` — 351 passed, 5 skipped (env-gated)
  - New: `TestWrapupGuard::test_floor_delivers_prefix_less_pm_text` — non-empty prefix-less PM text is floor-delivered, not canned; `PM_COMPLIANCE_NUDGE` not written post-wrapup
  - New: `TestWrapupGuard::test_terminal_message_sent_when_pm_genuinely_empty` — OPERATOR_TERMINAL_MESSAGE fires only on empty PM transcript
  - New: `TestPerTurnContractReminder::test_contract_reminder_appended_on_dev_handoff` — PM_TURN_CONTRACT_REMINDER appears in the dev-handoff write
  - Updated: old `test_terminal_message_sent_when_pm_silent` split into the two new tests above (behavior changed by design)
- [x] `tests/integration/test_granite_pty_production.py` — 4 passed
  - New: `test_prefix_less_pm_response_floor_delivered_not_canned` — BridgeAdapter surfaces `pm_floor_delivered` without `exit_anomaly`
- [x] `tests/unit/test_session_executor_granite.py::TestIsNonCleanGraniteExit` — 11 passed
  - `pm_floor_delivered` added to clean-exit parametrize table (returns `False`)
  - `pm_floor_delivered` added to reaction-gating parametrize table (not `REACTION_ERROR`)

## Bridge-machine follow-ups (NOT blocking this PR)

- **spike-2:** Read a real production PM transcript from `logs/sessions/tg_valor_*/` to confirm the regression is contract decay over turns vs. prime load failure. If prime fails to load, Change 1's mechanism shifts to fixing the prime-load path; Change 2 is unaffected. This resolves Open Question 1 in the plan.
- **E2E Telegram verification:** Send a real Telegram message to the Eng session post-deploy; confirm a non-canned reply; tail `logs/worker/*.log` for `pm_complete`/`pm_user`/`pm_floor_delivered` (not `pm_no_user_message`).

Closes #1719


tomcounsell:
## Review: Approved

Reviewed against plan `docs/plans/granite_routing_prefix_floor.md` and issue #1719. The implementation matches the plan precisely and all PR-contract tests pass.

### What was verified

- **Change 1 — per-turn prefix reminder** (`container.py:1431`): `PM_TURN_CONTRACT_REMINDER` is appended to `dev_text` before the handoff write to the PM PTY, exactly at the location the plan specified. Restores the per-turn contract assertion lost when `--append-system-prompt` was removed in #1694.
- **Change 2 — relaxed wrap-up floor** (`container.py:1557-1594`): The floor check is correctly placed inside the `if pm_text:` block **before** the `_route_pm_classification` call, with `return` after delivery. This is the load-bearing placement requirement from the plan — it ensures `PM_COMPLIANCE_NUDGE` is never written into a PTY about to be torn down. The unknown-destination branch delivers `pm_text.strip()` directly via `_on_user_payload`, sets `user_facing_routed=True` and `exit_reason="pm_floor_delivered"`. The prefix-found path and the empty-transcript canned path are both preserved.
- **Change 2b — clean-exit classifier** (`session_executor.py:35`): `pm_floor_delivered` added to `_CLEAN_GRANITE_EXIT_REASONS` and the docstring updated, so floor-delivered sessions get `REACTION_COMPLETE` not `REACTION_ERROR` (the mandatory B1 guard).
- **Invariant comment** at the wrap-up-guard entry gate documents why no runtime sticky-failed guard is needed (failure exit_reasons structurally bypass the gate).
- **Docs**: `granite-pty-production.md` updated comprehensively — exit-reason table, wrap-up guard sequence, `exit_anomaly` note, dashboard chip list, finalize-status table. `ContainerResult` docstring updated.
- **No cruft**: `_unknown_classification` remains used at two other call sites; no dead code left behind. Old test `test_terminal_message_sent_when_pm_silent` was correctly split into two new tests with no orphaned references.

### Tests

The PR-contract tests pass: `TestIsNonCleanGraniteExit` (both `pm_floor_delivered` rows), `TestWrapupGuard::test_floor_delivers_prefix_less_pm_text`, `TestWrapupGuard::test_terminal_message_sent_when_pm_genuinely_empty`, and `TestPerTurnContractReminder` — all green.

`TestReactionGating::test_reaction_selection` (13 parametrize cases) fail on this skills-only machine, but the failure is **environmental, not a regression**: the test invokes the full `_execute_agent_session` path which provisions a real git worktree (`git worktree add` → exit 128 because `/private/tmp` is outside the allowed git root). The failure hits the entire parametrize table uniformly — including pre-existing cases (`pm_complete`, `exception`, `None`) that have nothing to do with this PR — and raises before the reaction branch is ever reached. The PR did not touch the test body or `_execute_agent_session`; it only added two parametrize rows. The actual `pm_floor_delivered` classification contract is proven by `TestIsNonCleanGraniteExit` (pure function, no worktree), which passes.

### Notes

- `## Verification` table and the bridge-machine E2E (real Telegram round-trip) cannot be exercised here — this is the skills/tools-only machine. The PR body correctly flags spike-2 and E2E as bridge-machine follow-ups.
- Edge case worth noting (matches plan intent, not a blocker): if `_on_user_payload` raises during floor delivery, the code logs and `return`s without falling through to the canned fallback, so the human gets nothing on that rare path. The plan explicitly accepts this ("do NOT mark `user_facing_routed`; fall through to logging").

### Blockers
- None

### Tech Debt
- None

### Nits
- None

Issue #547, opened 2026-03-26
Reflection: audit merged PRs for unaddressed review findings and resurface as issues


valorengels:
> **Context:** The [reflections system](scripts/reflections.py) runs scheduled maintenance tasks across all projects — log review, Sentry checks, doc audits, branch cleanup, session analysis, and GitHub issue creation. It currently has 19 steps and runs on a launchd schedule. Each step is an async method (`step_*`) on the `ReflectionRunner` class. The system already creates GitHub issues for code bugs found during log analysis (step `auto_fix_bugs` and `create_github_issue`). PR reviews are posted by the [`do-pr-review`](.claude/skills/do-pr-review/SKILL.md) skill as GitHub review comments with structured findings categorized as `blocker`, `tech_debt`, or `nit`.

## Problem

When the PM merges a PR with unaddressed findings — tech debt "noted for future work", test gaps, or nits left behind — those items disappear. Nobody tracks them, nobody comes back to fix them, and they accumulate as invisible debt. Issue #544 addresses the root cause (making the PM patch findings before merging), but things will still slip through. There's no safety net that audits merged PRs and resurfaces deferred work.

**Current behavior:**
- PM merges PR with "3 tech debt items noted for future work" → items vanish
- PR review says "three minor test gaps noted" → merged without patching → gaps forgotten
- No automated process reads merged PR reviews to extract leftover findings
- The reflections system audits logs, docs, branches, and sessions, but not PR review quality

**Evidence from PM: Popoto chat:**
- 03/23: PR #261 merged with "3 tech debt items noted for future work" — never resurfaced
- 03/26: PR #282 review found "three minor test gaps" — PM said "Ready to merge" until Tom manually said "patch the test quality findings"

**Desired outcome:**
- A reflection step scans recently merged PRs for unaddressed review findings
- Each finding is classified by criticality: **critical** (security, correctness), **standard** (tech debt, test gaps), or **trivial** (nits, style)
- All findings are filed as GitHub issues with appropriate labels
- Critical and trivial items (quick wins and must-fixes) are queued for immediate SDLC work with PM oversight
- Standard items are tracked as issues for future prioritization

## Definitions

| Term | Definition | Reference |
|------|-----------|-----------|
| Reflections system | Scheduled maintenance runner with 19 async steps, runs on launchd schedule | [`scripts/reflections.py`](scripts/reflections.py), [`docs/features/reflections.md`](docs/features/reflections.md) |
[...]

PR #1763, opened 2026-06-22
fix(sdlc): 3-layer cross-repo plan resolution fix (PLAN↔CRITIQUE loop)

### LINKED ISSUE #1761: investigation: /do-sdlc PLAN↔CRITIQUE router never converges to BUILD (notes-only revision re-stales a clean verdict)
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/sdlc_cross_repo_plan_resolution.md

## Recon Summary

Reconnaissance was completed and confirmed by the repo owner in the "Confirmed root cause (reproduced locally)" comment (reproduced on `e1ad6e7e`). Findings:

**Confirmed:**
- Binding root cause is a cross-repo plan-resolution bug, NOT critique-classification. `find_plan_path` / `_resolve_target_repo` (`tools/_sdlc_utils.py:330,51`) run inside `sdlc-tool` which forces cwd to `~/src/ai` via `uv run --directory $AI_REPO_ROOT`. When `SDLC_TARGET_REPO` is unset (local `/do-sdlc` runs never set it — only `agent/sdk_client.py:1590` does), resolution falls to `_git_toplevel()` → `~/src/ai`, so a target-repo issue resolves to None or a foreign ai-repo plan via the bare `#N` textual fallback.
- Consequence: `_parse_revision_applied` (`tools/sdlc_stage_query.py:294`) sees `revision_applied=False`, making router row 4c (`_rule_critique_ready_with_concerns_revision_applied`) and G7 self-heal unreachable → rows 2b→4b→2b cycle forever.
- Secondary: `do-plan-critique/SKILL.md:211` uses bare `from tools.sdlc_verdict import compute_plan_hash`; in a target repo shipping its own `tools/` this imports a divergent copy. Other global skills carry bare `cd ~/src/ai` / `python -m tools.X` calls (`do-build/SKILL.md:490`, `do-build/PR_AND_CLEANUP.md:107,127`, `do-plan:132,154`, etc.).

**Revised:**
- The "nit-minting / with-concerns" critique behavior is a contributing irritant, not the binding cause — a clean with-concerns verdict should converge in a single revision pass regardless.

**Pre-requisites:** none (no external deps).

**Dropped:** changing critique nit/concern classification — out of scope; the path bug is the binding cause.

---

## Observed behavior (unverified root cause)

During two separate `/do-sdlc` runs in the **cuttlefish** repo on 2026-06-22 (issues #547 and #550, each in its own worktree), the supervisor **never converged from PLAN/CRITIQUE to BUILD on its own**. Both subagents reported the same thing and had to **manually drive the BUILD stage** against a plan that was already marked build-ready, with **zero code-correctness blockers** ever raised. Quoting the run reports:

> "the PLAN↔CRITIQUE router never converged to BUILD — every notes-only plan edit re-staled the critique verdict, and FULL critiques on a `.claude/skills`-path plan kept minting non-blocking prose nits. Zero code-correctness blockers ever appeared."

> "a READY-TO-BUILD verdict with zero blockers shouldn't be invalidated by a subsequent notes-only `/do-plan` write."

Net effect: wasted PLAN→CRITIQUE→PLAN dispatch cycles, and the pipeline only progressed because a human/supervisor overrode it. This defeats the point of autonomous `/do-sdlc`.

## Why this is filed as `investigation`, not `bug`

This part of the router has a long lineage of dead-end fixes, so the specific mode below may be a partial regression or an uncovered edge — root cause needs confirmation before any change:

- `3e1e3dae` Fix SDLC router dead-end: CRITIQUE in_progress with empty verdict (#1668)
- `6e943ea9` Fix SDLC router stale-critique dead-end (#1639)
- `5bc6243a` verdict normalization + plan-existence gate + stale-verdict supersession (#1638/#1640/#1641)
- `8218c5af` Guard rule 4b against re-firing once a PR exists (#1554)
- `627e3cf0` row 8c REVIEW empty-verdict re-dispatch (#1755)

## Plausible mechanism (from static reading — to be confirmed)

The staleness check is purely timestamp-based:

- `_critique_verdict_is_stale()` — `agent/sdlc_router.py:750-783` — returns True whenever the CRITIQUE verdict's `recorded_at` predates the **latest `/do-plan` dispatch**, regardless of whether that plan edit was substantive.
- Row 2b `_rule_critique_verdict_stale` — `agent/sdlc_router.py:786-800` — on stale verdict, re-dispatches `/do-plan-critique`.
- Row 4b `_rule_critique_ready_with_concerns_no_revision` — `agent/sdlc_router.py:649-671` — `READY TO BUILD (with concerns)` + `revision_applied` not set → routes to `/do-plan` (a revision pass).
- The revision pass (`do-plan` Phase 4, SKILL.md ~438-471) embeds the concern/nit notes into the plan **text** and sets `revision_applied: true` in frontmatter, which busts the plan hash (G5, `agent/sdlc_router.py:374-423`) and, by timestamp, **re-stales** the just-recorded verdict → back to row 2b → critique again.

Suspected loop: `READY TO BUILD (with concerns)` (or a critique that keeps minting non-blocking **nits** on a `.claude/skills`-path plan) → row 4b → `/do-plan` revision → verdict goes stale → row 2b → `/do-plan-critique` → same verdict → … never reaches row 4a/4c (build). Per `do-plan-critique` SKILL.md Step 5 (~line 405) **nits are supposed to be ignored** (zero blockers + zero concerns ⇒ `READY TO BUILD (no concerns)` ⇒ no lock ⇒ build), so the open question is **why these runs did not land on the no-concerns verdict** — either the critique is emitting "with concerns" for nit-only findings, or the staleness re-trigger outruns convergence.

## What to investigate

1. Reproduce with a plan whose only findings are nits (ideally a `.claude/skills`-path plan, which is what both cuttlefish runs were). Capture the recorded CRITIQUE verdict string each cycle.
2. Confirm whether the critique skill emits `READY TO BUILD (no concerns)` for nit-only findings, or incorrectly attaches "with concerns".
3. Decide whether a **notes-only / non-substantive** `/do-plan` write should invalidate a recorded `READY TO BUILD` verdict at all. Options to weigh: compare plan **body hash excluding frontmatter** for staleness; or treat `revision_applied: true` as the convergence signal (row 4c) rather than re-staling; or short-circuit to build when the latest verdict is `READY TO BUILD` and the only post-verdict dispatch was the revision pass that set `revision_applied`.
4. Check interaction with G5 hash cache (`374-423`) — frontmatter-inclusive hashing intentionally busts the cache on `revision_applied`, which is precisely what re-stales the verdict.

## Key files

| Component | File | Lines |
|-----------|------|-------|
| Router dispatch + rows 4a/4b/4c | `agent/sdlc_router.py` | 636-696, 1142-1214 |
| Critique staleness | `agent/sdlc_router.py` | 750-783 |
| Row 2b stale re-dispatch | `agent/sdlc_router.py` | 786-800 |
| G5 hash cache / G7 lock | `agent/sdlc_router.py` | 374-423 / 426-510 |
| Verdict writer + plan hash | `tools/sdlc_verdict.py` | 164-281 / 94-117 |
| Critique emit/record/lock | `.claude/skills-global/do-plan-critique/SKILL.md` | 352-458 |
| Plan revision pass | `.claude/skills-global/do-plan/SKILL.md` | 438-471 |

## Impact

Autonomous `/do-sdlc` cannot reach BUILD unattended for plans that produce only concerns/nits (observed on both cross-repo runs today). Each stuck run burns repeated PLAN/CRITIQUE dispatches and requires manual intervention.

tomcounsell:
## Summary

Closes #1761

Fixes the PLAN↔CRITIQUE infinite loop during local `/do-sdlc` runs against non-ai-repo issues (reproduced with cuttlefish #547, #550). Root cause: `sdlc-tool` forces cwd to `~/src/ai`, so `find_plan_path` resolved plans from the wrong repo — `revision_applied: true` was never read, router row 4c was unreachable, and the loop ran forever unattended.

**3-layer fix:**

### Layer 1 — SDLC_TARGET_REPO propagation + resolver hardening
- `/do-sdlc` Step 2 now exports `SDLC_TARGET_REPO` (filesystem path) alongside `SDLC_REPO` (GitHub slug). Every `sdlc-tool` subprocess inherits it; `find_plan_path` uses it as the plans-dir root even though cwd is forced to `~/src/ai`.
- `find_plan_path` now returns `None` instead of a foreign plan when the `__file__` fallback path is taken (SDLC_TARGET_REPO unset AND not in a git repo). Scoped strictly to the ai-repo fallback branch per critique CONCERN 3 — same-repo issues still use the bare-`#N` fallback on all other resolution paths.

### Layer 2 — Skill portability
- All bare `from tools.X` / `python -m tools.X` / `cd ~/src/ai` invocations in 7 global SDLC skills anchored to `${AI_REPO_ROOT:-$HOME/src/ai}` so they load the canonical ai-repo `tools/` when running inside a target repo that ships its own `tools/` package.

### Layer 3 — `revision_applied`-stripped plan hash
- New `compute_plan_body_hash` strips ONLY the `revision_applied:` frontmatter key before hashing. G5's `current_plan_hash` (reader) and `_compute_artifact_hash` (writer) both use it. A `/do-plan` revision write flipping `revision_applied: true` no longer busts the G5 cache; all other edits still bust it.
- `guard_g5_artifact_hash_cache` transparently self-heals pre-#1761 stored hashes: if the legacy full-bytes hash matches the stored `cached_hash` (only delta is `revision_applied:`), it rewrites in-place with a WARNING log and treats it as a cache hit. No operator step; no backfill script.
- Critique CONCERN 1 honored: recomputes `compute_plan_hash` (legacy) on the current file; strict equality check before any rewrite; never overwrites on a genuine plan-body change.
- Critique CONCERN 2 honored: WARNING log emitted on every transparent rewrite.

## Files Changed

- `tools/_sdlc_utils.py` — `find_plan_path` hardening, `__file__`-fallback rejection
- `tools/sdlc_verdict.py` — add `compute_plan_body_hash`, update `_compute_artifact_hash` and docstring
- `tools/sdlc_next_skill.py` — switch `current_plan_hash` to `compute_plan_body_hash`, inject `issue_number` into context
- `agent/sdlc_router.py` — G5 transparent-rewrite migration
- `.claude/skills-global/do-sdlc/SKILL.md` — export `SDLC_TARGET_REPO` in Step 2
- `.claude/skills-global/do-{build,docs,patch,plan,plan-critique,pr-review}/` — anchor all `tools.*` invocations to `AI_REPO_ROOT`
- `docs/features/sdlc-tool-resolver.md` — document D8 cross-repo fixes
- `docs/features/sdlc-pipeline-portability.md` — document D8 in the portability feature doc

## Test plan

- [ ] `pytest tests/unit/test_sdlc_utils.py` — new `find_plan_path` cross-repo fallback rejection tests (57 pass)
- [ ] `pytest tests/unit/test_sdlc_verdict.py` — new `compute_plan_body_hash` tests (11 cases), G5 migration tests
- [ ] `pytest tests/unit/test_sdlc_next_skill.py` — body hash + issue_number context tests
- [ ] `pytest tests/unit/test_sdlc_env_vars.py` — SDLC_TARGET_REPO contract tests
- [ ] `pytest tests/unit/test_sdlc_router.py` — router G5 migration + row 4c convergence tests
- [ ] `pytest tests/integration/test_sdlc_cross_repo_resolution.py` — E2E cross-repo resolution (cwd forced to ~/src/ai)
- [ ] All 124 affected tests pass; 2 pre-existing unrelated failures on main unchanged


tomcounsell:
## Review: Changes Requested

**Branch:** `session/sdlc-cross-repo-plan-resolution` → `main`
**Plan:** `docs/plans/sdlc_cross_repo_plan_resolution.md`
**Issue:** Closes #1761

Thorough, well-scoped 3-layer fix. The `SDLC_TARGET_REPO` export, the `find_plan_path` bare-`#N` fallback rejection (correctly scoped to the `__file__` fallback branch only), the `compute_plan_body_hash` narrow strip, and the skill-portability `AI_REPO_ROOT` anchoring all match the plan's Technical Approach and critique CONCERNs precisely. Docs (`sdlc-tool-resolver.md`, `sdlc-pipeline-portability.md`) and inline docstrings are accurate and complete. All 124 affected tests pass locally (98 unit + 26 router/integration).

Two **tech_debt** findings below — both in the Layer-3 transparent-migration path. Neither blocks correctness (the loop still converges), but both should be patched before merge.

### Blockers
- None

### Tech Debt

**File:** `agent/sdlc_router.py:431` (verified: read this file)
**Code:** `record["artifact_hash"] = current_hash`
**Issue:** The transparent-migration rewrite mutates the in-memory `_verdicts["CRITIQUE"]` dict only. That dict is a transient snapshot reloaded fresh from the persisted session on **every** router pass (`tools/sdlc_stage_query.py:506` threads `_verdicts` straight off `_load_raw_states`; `tools/sdlc_next_skill.py` never writes `stage_states` back). So the rewrite is discarded at end-of-call and is **never persisted**. This contradicts the docstring at `sdlc_router.py:388-394` ("transparently rewrites the stored `artifact_hash`… idempotent — once written, subsequent calls use the new hash directly") and the plan's Risk-2 mitigation, which required a persisted rewrite "via `record_verdict`'s single-writer path." Functional impact is bounded: because the legacy-hash recompute re-fires every pass, a build-ready in-flight issue still self-heals to a cache-hit each time, so the loop converges — but the `find_plan_path` + full-file read + `compute_plan_hash` recompute runs on **every** router pass for that issue until a genuine CRITIQUE/BUILD `record_verdict` re-fires, rather than once.
**Severity:** tech_debt
**Fix:** Either (a) persist the rewrite via `record_verdict`'s single-writer path as the plan specified (then the docstring's "once written" claim holds), or (b) keep the in-memory-only patch but correct the docstring at `sdlc_router.py:388-394` to state the recompute re-runs each pass and is not persisted. Pick one so code and docstring agree.

**File:** `tests/unit/test_sdlc_router.py:377` (verified: read this file)
**Code:** `context = {"current_plan_hash": "sha256:abc"}`
**Issue:** The new G5 transparent-migration branch (`sdlc_router.py:406-437`) has **zero test coverage**. Every existing `TestG5DefersAfterBuild` case uses matching hashes (`sha256:abc` == `sha256:abc`), so the entire `cached_hash != current_hash` → legacy-recompute → cache-hit path is never exercised. The plan's Test Impact explicitly required this case: "cached full-bytes `artifact_hash` (pre-deploy) for a build-ready issue + current `revision_applied`-stripped hash → guard recognizes the `revision_applied`-only delta… returns the cached READY-TO-BUILD dispatch (no spurious re-critique)." This is the branch most likely to harbor a regression and the one that justifies the whole Layer-3 design.
**Severity:** tech_debt
**Fix:** Add a router test that constructs a cached legacy full-bytes `artifact_hash` differing from the body-hash only by `revision_applied:`, provides `issue_number` + a real plan file in `context`, and asserts G5 returns the cached READY-TO-BUILD `Dispatch` (no `None`, no re-critique).

### Nits
- None

### Legacy Cruft
- None — legacy `compute_plan_hash` is intentionally retained for the migration recompute and full-bytes callers; no dead code, no commented-out blocks, no half-migrations introduced.

tomcounsell:
## Review: Changes Requested

**Plan:** `docs/plans/sdlc_cross_repo_plan_resolution.md`
**Issue:** Closes #1761 — /do-sdlc PLAN↔CRITIQUE router never converges to BUILD
**Result:** Changes Requested (1 tech_debt, 1 nit — no blockers)

### Summary

Solid, well-scoped fix for the cross-repo plan-resolution loop. All three layers map cleanly to the confirmed RCA and are backed by tests:

- **Layer 1** — `find_plan_path` now suppresses the bare-`#N` textual fallback when resolution falls to the `__file__` path (SDLC_TARGET_REPO unset + not in a git repo), returning `None` so the target repo re-plans instead of latching onto a foreign ai-repo plan. The `tracking:` match stays authoritative on all paths. `/do-sdlc` Step 2 now exports `SDLC_TARGET_REPO`.
- **Layer 2** — All bare `cd ~/src/ai` and `python -m tools.*` invocations across 7 global skills are anchored to `${AI_REPO_ROOT:-$HOME/src/ai}`. Verified no stale bare invocations remain.
- **Layer 3** — `compute_plan_body_hash` strips only the `revision_applied:` frontmatter key before hashing, so a revision write no longer busts the G5 verdict cache. The legacy `compute_plan_hash` is retained for the critique-resume sidecar (which correctly still wants full-bytes hashing). G5 self-heals pre-#1761 stored hashes in-place with a WARNING.

**Verified empirically:** `revision_applied: true`/`false`/absent all produce the same body hash; `revision_applied` still busts the legacy `plan_hash` (migration path is real); the critique-resume vs G5-router hash split is intentional and correct (opposite invalidation requirements).

**Tests:** 119 unit + 5 integration pass. Integration tests directly exercise the acceptance criteria (`test_sdlc_target_repo_honored_even_when_cwd_is_ai_repo`, `test_file_fallback_bare_mention_suppressed`). Ruff format clean. No UI files — screenshot gate is a no-op.

### Blockers
- None

### Tech Debt

**File:** `docs/plans/sdlc_cross_repo_plan_resolution.md` (## Verification table)
**Code:** `grep -c "from tools.sdlc_verdict import" .claude/skills-global/do-plan-critique/SKILL.md` → Expected `match count == 0`
**Issue:** This verification row produces a false FAIL. Line 211 of `do-plan-critique/SKILL.md` is `uv run --directory "${AI_REPO_ROOT:-$HOME/src/ai}" python -c "from tools.sdlc_verdict import compute_plan_hash; ..."` — the import IS correctly anchored via `uv run --directory` (exactly what Layer 2 requires), but the grep matches the import substring inside the `python -c` string and reports 1. The check's intent (no *unanchored* import) is satisfied; the pattern just can't distinguish anchored from bare. As written it will fail every future verification run.
**Severity:** tech_debt
**Fix:** Tighten the grep to detect only *unanchored* imports, e.g. `grep -cE '^\s*(from tools\.|python3? -m tools\.)' ...` (anchored at line start, excluding the `uv run --directory` wrapper), or assert the line contains `AI_REPO_ROOT` rather than asserting count == 0.

### Nits

**File:** `agent/sdlc_router.py:419`
**Code:** `                        import logging`
**Issue:** Redundant local `import logging` inside the G5 migration try-block. `logging` is already imported at module level (line 34), and the `except` branch at line 433 references `logging` directly (relying on that module-level import). The module also defines `logger = logging.getLogger(__name__)` at line 43 which could be used instead of re-resolving `logging.getLogger(__name__)`.
**Severity:** nit
**Fix:** Remove the line-419 `import logging` and use the module-level `logger` for both the warning and the debug call.

### Note (non-blocking)

The plan's `## Verification` table includes a `python -m ruff check` (lint) row. Per repo convention (CLAUDE.md / user prefs) this project runs `ruff format` only, not `ruff check`. Harmless here (it passed), but worth aligning future plans to format-only.

tomcounsell:
## Review: Approved

**Verdict:** APPROVED — zero findings.

Reviewed PR #1763 (3-layer cross-repo plan resolution fix, Closes #1761) against the plan `docs/plans/sdlc_cross_repo_plan_resolution.md`.

### Mergeability preflight
- state=OPEN, mergeable=MERGEABLE, mergeStateStatus=CLEAN.

### Plan validation — all success criteria met
- **Layer 1 (resolver + env):** `find_plan_path` resolves `SDLC_TARGET_REPO` → git-toplevel → `__file__` fallback in that order, and suppresses the bare-`#N` textual fallback (returns `None`) only on the ai-repo `__file__` branch (`_is_ai_repo_fallback`), exactly as scoped by critique CONCERN 3. The `tracking:`-match path stays authoritative on all resolution paths. `/do-sdlc` Step 2 exports `SDLC_TARGET_REPO` (filesystem path) alongside `SDLC_REPO` (slug), with the path-vs-slug distinction commented.
- **Layer 2 (skill portability):** All 7 global SDLC skills anchor `tools.*` invocations to `${AI_REPO_ROOT:-$HOME/src/ai}` (12 anchored references). Zero bare `cd ~/src/ai` / `from tools.sdlc_verdict` remain. The one `from tools.doc_impact_finder` in `do-docs` is correctly anchored via the immediately-preceding `sys.path.insert('${AI_REPO_ROOT:-...}')`.
- **Layer 3 (revision_applied-stripped hash + G5 migration):** `compute_plan_body_hash` strips only the `revision_applied:` line; reader (`sdlc_next_skill.current_plan_hash`) and writer (`_compute_artifact_hash`) both switched to it and inject `issue_number` into context. The G5 in-guard transparent rewrite recomputes the legacy full-bytes hash, rewrites the stored `artifact_hash` in place when the only delta is `revision_applied`, logs a WARNING, and falls through to a normal cache hit — self-healing build-ready in-flight issues without a backfill script. Critique BLOCKER 1, BLOCKER 2, CONCERN 4 all honored (docstring at `:65-77` updated to scope the contract).

### Verification (all pass)
- `def compute_plan_body_hash`: 1; G5 migration markers in router: 20; bare critique import: 0; bare `cd ~/src/ai`: 0; `SDLC_TARGET_REPO` in do-sdlc: 3.
- `ruff format --check` on all 4 source files: clean.
- Tests: 124 pass — `test_sdlc_utils` / `test_sdlc_verdict` / `test_sdlc_next_skill` / `test_sdlc_env_vars` (98) + `test_sdlc_router` + `test_sdlc_cross_repo_resolution` (26).
- Behavioral spot-check of hash invariants: `revision_applied` true/false/absent all hash identically; a `status:` edit busts the cache; legacy full-bytes hash differs (migration premise holds).

### Blockers
- None

### Tech Debt
- None

### Nits
- None

(Self-authored PR — posted as a comment; GitHub forbids a formal self-approval.)

PR #1795, opened 2026-06-25
fix(watchdog): deterministic U-state worker recovery (#1767)

### LINKED ISSUE #1767: Worker watchdog fails to recover a U-state (uninterruptible) hung worker
**Plan:** https://github.com/tomcounsell/ai/blob/session/sdlc-1767/docs/plans/worker_watchdog_ustate_recovery.md

> **Context for readers.** The **worker** (`python -m worker`, `worker/__main__.py`) is the sole engine that executes agent sessions. Long-running sessions run inside a **granite PTY container** — an interactive `claude` TUI driven over a pseudo-terminal (`agent/granite_container/`). A **worker watchdog** (`monitoring/worker_watchdog.py`) runs as a *separate* launchd service (`StartInterval: 120s`) specifically so it can kill a wedged worker that launchd's own `KeepAlive` cannot detect. This issue is about a real production incident on 2026-06-23 where the watchdog failed to recover a worker hung in OS **uninterruptible-sleep (`U`) state**, leaving three sessions stalled until manual intervention.

## Problem

On 2026-06-23, a `/update` restarted the worker mid-flight. The fresh worker (PID 60968) wedged on a granite PTY read and went into **`U` (uninterruptible sleep)** state — blocked in a kernel syscall that cannot receive signals. The worker stopped writing heartbeats, every in-flight session's heartbeat froze, the granite containers looped on `transcript read: no-new-entry … using unknown classification`, and the pending queue stopped draining. Manual recovery (kill the wedged sessions, force-kill the worker once the syscall released, let catchup re-enqueue) was required.

The watchdog is *supposed* to catch exactly this. It did not, for several compounding reasons.

### Current behavior

1. **launchd is blind to `U`-state.** `com.valor.worker.plist` uses `KeepAlive=true`, which only restarts a process that *exits*. A `U`-state process never exits, so launchd treats it as healthy. (This is why the external watchdog exists.)

2. **The watchdog's kill is not verified and cannot beat `U`-state.** `monitoring/worker_watchdog.py:184 recover()` does `SIGTERM → sleep(3) → SIGKILL`, then logs `"Worker killed — launchd will restart"` **without confirming the process actually died**. `SIGKILL` against a `U`-state process is *queued, not delivered*, until the blocking syscall returns — so the log line was emitted while PID 60968 was still alive. The watchdog had no idea its kill failed.

3. **The detection threshold is lax.** `HEARTBEAT_THRESHOLD = 600` (10 min) with `StartInterval: 120s` yields up to a **~10-minute dead window** before the watchdog even attempts recovery.

4. **The heartbeat is not isolated from the hang, so "stale heartbeat" is an ambiguous signal.** The intended model is: heartbeat writes on the event loop, PTY reads on an `asyncio.to_thread` worker thread, therefore independent. In practice they share fate: `container.run` runs on the *default* thread-pool executor — the same executor the reflection scheduler uses via `run_in_executor(None, call)` (`agent/reflection_scheduler.py:411`). Three sessions each blocking a `container.run` saturate the pool and the heartbeat/health path queues behind the hung reads. We observed the worker heartbeat go stale at 564s, confirming propagation.

5. **Killed-session work is silently dropped.** Worker startup recovery only re-kicks `status="pending"` sessions (`worker/__main__.py:206,412`). It never resumes `killed` sessions. When the hung worker's sessions are killed, their unanswered human messages are only recovered if `valor-catchup` happens to run — nothing in the recovery path triggers it.

### Desired outcome

A worker hung in `U`-state is detected within ~2–3 minutes and **deterministically** recovered — verified-dead, escalated if the syscall won't release, with its in-flight sessions' work re-enqueued automatically — with no human intervention.

## Definitions

|------|------------|--------|
| `U`-state (uninterruptible sleep) | A Unix process state blocked in a kernel syscall that cannot receive signals (incl. `SIGKILL`) until the syscall returns. | `man ps` STAT codes |
| Worker watchdog | External launchd service that kills a wedged worker so launchd restarts it. | `monitoring/worker_watchdog.py` |
| Granite PTY container | Interactive `claude` TUI driven over a pseudo-terminal to run long sessions. | `agent/granite_container/`, `docs/features/granite-pty-production.md` |
| Heartbeat | Periodic `last_worker_connected` file + Redis write proving the worker loop is alive. | `data/last_worker_connected`, `agent/session_health.py` |
| valor-catchup | Agent-judgment recovery that re-enqueues genuinely-unanswered human messages. | `docs/features/agent-judgment-catchup.md` |
| Kill-on-evidence policy | Existing rule: kill only on positive evidence of failure, not absence of activity. | issue #1172 (`agent/session_health.py`) |

## Solution sketch

Root cause is partly understood (the `U`-state read + unverified kill), but the right thresholds and escalation ladder are still open. Treat the following as **open questions for `/do-plan`**, not a fixed design:

- **Verify-after-kill + escalation ladder.** After `SIGKILL`, poll `os.kill(pid, 0)` for a grace window. If the process survives (true `U`-state), what is the escalation order — `launchctl bootout` the job, then force the blocking `os.read()` to return EOF by closing the worker's PTY master fds (the only thing that frees a `U`-state read), then alert a human? Which steps are safe to automate vs. alert-only?
- **Tighten the threshold** from 600s toward ~180s — but this is only safe if the heartbeat is a *trustworthy* hang oracle (next point). What threshold balances fast recovery against false positives during legitimate long syscalls?
- **Make the heartbeat immune to the hang.** Should heartbeat writes move off the shared default executor onto a dedicated OS thread with its own Redis connection, so a PTY/thread-pool saturation can't freeze it? This converts "stale heartbeat" into an unambiguous "worker is wedged" signal.
- **Per-read hard wall-clock guard.** Should a separate timer thread close the PTY master fd when a single read exceeds a hard ceiling, converting an unkillable `U`-state into a returnable EOF *before* it propagates to the whole worker?
- **Post-restart session sweep.** On worker kill/restart, flip the dead worker's `running` sessions (no live `claude_pid`) → `killed` and trigger `valor-catchup` so their work re-enqueues. Should this live in the watchdog, in worker startup, or both?
- **Trigger.** This incident was triggered by `/update` restarting the worker while granite PTY sessions were in-flight. Should `/update` quiesce/await in-flight PTY sessions before restarting the worker, as a complementary mitigation?

## Why this can't be solved by a reflection

The reflection scheduler is an in-process asyncio loop inside the worker (`agent/reflection_scheduler.py:10`). A hung worker stops ticking its own reflections, so it cannot self-heal. Worker-level recovery **must** be external (the watchdog). The complementary session-level early-detection layer is tracked separately as a `reflections` issue.

## Acceptance criteria

- [ ] A worker in `U`-state with a stale heartbeat is recovered without human intervention, with the kill **verified dead** (or escalated) — not assumed.
- [ ] Detection-to-recovery window is reduced from ~10 min toward ~2–3 min, with a documented false-positive guard.
- [ ] Worker heartbeat writes survive thread-pool/PTY saturation (stale heartbeat reliably means "wedged").
- [ ] After a hung-worker kill, the dead worker's in-flight sessions' work is re-enqueued (no silently dropped human messages).
- [ ] Tests cover: unverified-kill regression, `U`-state escalation path, and the post-restart session sweep.
- [ ] `docs/features/bridge-self-healing.md` (or worker-watchdog docs) updated with the new ladder and thresholds.

## Downstream context

This issue will be consumed by `/do-plan` to produce a plan document at `docs/plans/{slug}.md`. The open questions above are deliberately not pre-decided — `/do-plan` should resolve the escalation ladder and thresholds against the actual code rather than execute a fixed sketch.

**Companion issue:** the session-level early-detection layer (promote `stall-advisory` to an actor + add a `granite_wedged` signal) is filed separately under the `reflections` label.

tomcounsell:
## Summary

- Upgrade `recover()` in `monitoring/worker_watchdog.py` from unverified SIGTERM→SIGKILL to a **W1→W5 verified-kill escalation ladder** that polls `os.kill(pid, 0)` at each rung before escalating
- Move worker heartbeat to a **dedicated daemon thread** (`worker-heartbeat`) so PTY/thread-pool saturation cannot starve heartbeat writes and produce false stale-heartbeat signals
- Add **`_sweep_dead_worker_sessions()`** to startup recovery: identifies `status="running"` sessions with dead `claude_pid`, finalizes them `killed` via CAS, and triggers `valor-catchup` so unanswered human messages re-enqueue

### `monitoring/worker_watchdog.py`
- `HEARTBEAT_THRESHOLD`: 600s → 180s (6× write interval; env-tunable `HEARTBEAT_THRESHOLD`)
- `recover()`: W1 SIGTERM→poll 5s, W2 SIGKILL→poll 10s, W3 `launchctl bootout`→poll 10s, W4 CRITICAL Redis keys + operator hint, W5 final log
- New `_poll_pid_dead(pid, timeout_sec)` helper used by every rung
- `WORKER_WATCHDOG_PTY_CLOSE_DISABLED` env var disables the W4 `pty_close_required` key

### `worker/__main__.py`
- `_heartbeat_thread_main()`: daemon thread writing heartbeat every `WORKER_HEARTBEAT_INTERVAL`s (default 30)
- Thread started after granite PTY pool init; stopped cleanly on shutdown via `_heartbeat_stop_event`
- Dead-worker sweep (Step 3a) now runs **BEFORE** `_recover_interrupted_agent_sessions_startup` (Step 3b) — the ordering fix is critical: Step 3b transitions all running→pending without checking PID liveness, leaving nothing for the sweep to inspect

### `agent/session_health.py`
- Heartbeat write removed from `_agent_session_health_loop()`; comment points to new thread
- `_sweep_dead_worker_sessions()`: enumerate running sessions → PID liveness check → CAS finalize to killed → trigger `bridge.agent_catchup` subprocess

## Testing

- [x] `tests/unit/test_worker_watchdog.py` — 72 tests including `TestHeartbeatIsolation` (executor saturation independence, env override, default threshold) and `TestBootoutHelper`
- [x] `tests/unit/test_worker_session_sweep.py` — 12 tests covering dead-PID sweep, live-PID skip, recency guard, catchup trigger, conflict tolerance, return count
- [x] Lint + format clean (`ruff check`, `ruff format --check`)

## Documentation

- [x] `docs/features/bridge-self-healing.md` — W1→W5 table, heartbeat isolation rationale, threshold change, post-restart sweep, Redis critical key check commands
- [x] `docs/features/granite-pty-production.md` — U-state PTY-master block failure mode + W3 bootout backstop
- [x] `docs/features/README.md` — verified index entries

## Verification

| Check | Result |
|-------|--------|
| 75 tests pass | ✅ |
| Lint clean | ✅ |
| Format clean | ✅ |
| `recover()` polls `os.kill(pid, 0)` at each rung | ✅ |
| `bootout` present in ladder | ✅ |
| `Thread` in worker main (heartbeat off event loop) | ✅ |
| Sweep triggers `bridge.agent_catchup` | ✅ |
| `HEARTBEAT_THRESHOLD` = 180 (not 600) | ✅ |
| No raw Redis on sessions | ✅ |

Closes #1767


tomcounsell:
## Review: Changes Requested

**Plan:** `docs/plans/worker_watchdog_ustate_recovery.md` (issue #1767)
**Branch:** `session/sdlc-1767` → `main`
**Preflight:** OPEN / MERGEABLE / CLEAN

Solid, well-scoped fix for the 2026-06-23 U-state hung-worker incident. The three pillars all landed correctly:

- **Verified-kill ladder** (`recover()`): W1 SIGTERM→poll 5s, W2 SIGKILL→poll 10s, W3 `launchctl bootout`→poll 10s, W4 CRITICAL Redis key + operator hint, W5 final alert. Every rung re-verifies death via the by-PID `_poll_pid_dead()` helper (`os.kill(pid, 0)` until `ESRCH`/timeout) — the correct primitive, not the pgrep-style liveness check the plan warned against.
- **Heartbeat isolation:** moved to a dedicated `worker-heartbeat` daemon thread in `worker/__main__.py`, severed from the asyncio health loop. This makes "stale heartbeat" an unambiguous wedged signal and justifies the `HEARTBEAT_THRESHOLD` 600→180s tightening (6× the 30s write interval, env-tunable).
- **Post-restart sweep** (`_sweep_dead_worker_sessions`): correctly ordered as Step 3a **before** `_recover_interrupted_agent_sessions_startup` (Step 3b) — the critique's BLOCKER on inverted ordering is resolved. Gated on `claude_pid` liveness + the 300s recency guard, finalizes via Popoto `finalize_session` (no raw Redis), triggers `valor-catchup`.

The spike correctly found cross-process PTY-fd-close infeasible on macOS, so W4 became the CRITICAL-key/operator-hint path per the plan's Risk 2 honest fallback — the No-Go against implementing an fd-close facility is respected.

**Verification (all pass):** 75 unit tests pass (`test_worker_watchdog.py` + `test_worker_session_sweep.py`); `ruff check` + `ruff format --check` clean; by-PID poll present; bootout present; `Thread` in worker main; catchup triggered; threshold not 600; sweep uses Popoto only.

### Blockers
- None

### Tech Debt

**File:** `agent/session_health.py:779-781` (verified: read this file)
**Code:**
```
    - Applies the AGENT_SESSION_HEALTH_MIN_RUNNING_SECS recency guard (300s)
      so brand-new sessions from the fresh worker cannot be touched
    - Uses finalize_session(expected_status='running') CAS so a concurrent
      fresh-worker pickup wins and the session is skipped
```
**Issue:** The `_sweep_dead_worker_sessions` docstring contains two inaccuracies that will mislead future maintainers:
1. It names the recency-guard constant `AGENT_SESSION_HEALTH_MIN_RUNNING_SECS`, but the actual constant is `AGENT_SESSION_HEALTH_MIN_RUNNING` (no `_SECS` suffix). The named symbol does not exist.
2. It states the sweep "Uses `finalize_session(expected_status='running')` CAS." The actual call (lines 835-839) passes only `entry`, `"killed"`, and `reason` — **no `expected_status` argument**. The race protection does exist (via `finalize_session`'s *implicit* CAS: it re-reads on-disk status and compares against the in-memory `"running"` snapshot, raising `StatusConflictError` on a concurrent pickup), but the docstring describes an explicit argument that is not used. This is exactly the over-stated CAS framing the plan's critique flagged and resolved in favor of the recency guard (Critique Results: "Sweep 'CAS protection' claim over-stated"); the docstring re-introduces it.
**Severity:** tech_debt
**Fix:** Correct the constant name to `AGENT_SESSION_HEALTH_MIN_RUNNING`, and reword the third bullet to describe the actual mechanism, e.g.: "Relies on `finalize_session`'s implicit CAS (re-reads on-disk status, raises `StatusConflictError` if a concurrent fresh-worker pickup changed it) — the recency guard above is the primary protection."

### Nits
- None

The runtime behavior is correct in all three pillars; the only outstanding item is the docstring accuracy fix above. Routing to `/do-patch`.

tomcounsell:
## Review: Approved

**Plan:** `docs/plans/worker_watchdog_ustate_recovery.md` (issue #1767)
**Branch:** `session/sdlc-1767` → `main`
**Preflight:** OPEN / MERGEABLE / CLEAN

Re-review on HEAD `62fc22d6`. The single tech_debt finding from the prior review (docstring inaccuracies in `_sweep_dead_worker_sessions`) was resolved by commit `62fc22d6`:

- Constant name corrected to `AGENT_SESSION_HEALTH_MIN_RUNNING` (the non-existent `_SECS` variant is gone).
- The docstring now accurately describes `finalize_session`'s **implicit** CAS re-check (re-reads on-disk status, raises `StatusConflictError` on a concurrent fresh-worker pickup) instead of claiming an `expected_status='running'` argument it never passes.

All three pillars remain correct:

- **Verified-kill ladder** (`recover()`): W1 SIGTERM→poll 5s, W2 SIGKILL→poll 10s, W3 `launchctl bootout`→poll 10s, W4 CRITICAL Redis key + operator hint, W5 final alert. Every rung re-verifies death via the by-PID `_poll_pid_dead()` helper — the correct primitive. Full W1→W5 escalation is unit-tested (`TestRecoverW1..W4`, `TestPollPidDead`, `TestBootoutWorkerHelper`), including the unverified-kill regression (SIGKILL queued in U-state → ladder escalates, never logs false success).
- **Heartbeat isolation:** dedicated `worker-heartbeat` daemon thread, severed from the asyncio health loop; `HEARTBEAT_THRESHOLD` 600→180s (6× the 30s write interval, env-tunable). Saturation-independence is proven by `test_heartbeat_thread_writes_independent_of_executor`.
- **Post-restart sweep** (`_sweep_dead_worker_sessions`): correctly ordered Step 3a **before** Step 3b. Gated on `claude_pid` liveness + 300s recency guard, finalizes via Popoto `finalize_session` (no new raw Redis), triggers `python -m bridge.agent_catchup`. 12 sweep tests cover dead/live/zero/no-PID, both recency-guard sides, catchup trigger + negative case, catchup-failure tolerance, and StatusConflictError skip.

The spike's finding that cross-process PTY-fd-close is infeasible on macOS is honestly reflected in W4 (CRITICAL-key/operator-hint path per the plan's Risk 2 fallback) — the No-Go against an fd-close facility is respected.

**Verification (all pass):** 75 unit tests pass (`test_worker_watchdog.py` + `test_worker_session_sweep.py`); `ruff check` + `ruff format --check` clean on touched dirs; by-PID poll present; `bootout` present; `Thread` in worker main; catchup triggered; `HEARTBEAT_THRESHOLD` != 600; no new raw Redis on sessions; no debug cruft. No UI files — visual-proof gate N/A.

Note: the plan's Test Impact listed adding a U-state scenario to `tests/integration/test_watchdog_recovery.py`; that file was not modified, but the full ladder is exhaustively covered at the unit level, so the acceptance criterion "tests cover the U-state escalation path" is satisfied. Advisory only, not a blocker.

### Blockers
- None

### Tech Debt
- None

### Nits
- None

Self-authored PR — posted as a comment per do-pr-review identity rules.

Issue #1784, opened 2026-06-24
session-health: 300s tool_timeout kills legitimate long-running SDLC tools (Bash/Skill) — gate on PTY liveness


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/tool_timeout_pty_liveness_gate.md

## Recon Summary

Reconnaissance performed at plan time (2026-06-25) against `agent/session_health.py` at baseline `0c2c2a42`.

**Confirmed:**
- `agent/session_health.py:311` — `TOOL_TIMEOUT_DEFAULT_SEC = int(os.environ.get("TOOL_TIMEOUT_DEFAULT_SEC", 300))` is present and unchanged.
- `agent/session_health.py:322` — `MID_RUN_QUIESCENCE_SECS = 180` exists (from #1724) and tracks PTY-screen quiescence but is NOT consulted by `_check_tool_timeout`.
- `agent/session_health.py:360` `_check_tool_timeout()` flags a wedge purely on `last_tool_use_at` age vs `_tool_tier_budget(tier)` — no liveness signal consulted. Confirmed pure function, no side effects.
- The tool-timeout sub-loop (`_agent_session_tool_timeout_check`, ~line 2762) calls `_eval_mid_run_pty_stage1(entry, now)` at ~line 2847 (which maintains `mid_run_quiescent_since` on the entry) IMMEDIATELY BEFORE calling `_check_tool_timeout(entry)` at ~line 2849. So the quiescence signal is already computed and fresh on the same entry object in the same tick.
- Model fields exist on `AgentSession`: `last_pty_activity_at` (last screen paint, `models/agent_session.py:392`), `mid_run_quiescent_since` (`:397`), `last_pty_read_loop_at` (`:388`).

**Available liveness signal:** `mid_run_quiescent_since` is set to the time the PTY first went quiescent and cleared when activity resumes. `now - mid_run_quiescent_since >= MID_RUN_QUIESCENCE_SECS` is exactly the "quiescent long enough" predicate already used by stage-1. The default-tier kill can reuse this instead of re-deriving from `last_pty_activity_at`.

**Prior art:** PR #1728 (issue #1724) added the mid-run wedge detector + quiescence fields. PR #1781 (OPEN) fixed the advisory/`never_started` side; this issue covers the kill side.

[...]

PR #1798, opened 2026-06-25
feat(session-health): gate never_started kill on PTY liveness (#1792)

### LINKED ISSUE #1792: session-health: granite SDLC sessions killed during priming (no_progress/never_started) before first transcript entry
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/never_started_pty_liveness_gate.md

## Recon Summary

_Added by /do-plan Phase 0 — code-grounded reconnaissance before planning._

**Confirmed (verified against current main `7fb7e609`):**
- The never-started kill at `agent/session_health.py:3111` (D0 branch, #1724) recovers a session via `_apply_recovery_transition(reason_kind="no_progress", reason="never_started past grace")` **without** any PTY-liveness gate. This is the un-gated path the issue describes.
- The sibling tool_timeout kill at `agent/session_health.py:3168` **does** apply a PTY-liveness gate: `if tier == "default" and not _pty_quiescent_long_enough(entry, now): continue`. That gate landed via PR #1789 / commit `2408c9d2` (#1784), not the path `docs/plans/tool_timeout_pty_liveness_gate.md` cited in the issue — that plan file does not exist (minor drift).
- `_never_started_past_grace()` (`:874`) fires when `sdk_ever_output` is False AND `running_seconds > NEVER_STARTED_GRACE_SECS + NEVER_STARTED_CONFIRM_MARGIN_SECS` (default 120+30=150s). `sdk_ever_output` requires `last_tool_use_at` or `last_turn_at` — neither is set until the inner claude TUI writes its first transcript entry, which is precisely what is slow during granite SDLC prime.
- A reusable PTY-byte liveness primitive **already exists**: `last_pty_activity_at` (`models/agent_session.py:392`), stamped by `bridge_adapter._on_pty_read` (`:771`) **only when the NORMALIZED PTY buffer changes** (genuine repaint; spinner/elapsed-counter frames are normalized out by #1768). It is stamped during the prime-turn loop, before any transcript entry. `last_pty_read_loop_at` (`:388`, stamped unconditionally every read) proves the read loop is cycling.

**Revised / corrected pointers:**
- Issue cites plan `docs/plans/tool_timeout_pty_liveness_gate.md`; the #1784 work actually shipped in PR #1789 (commit `2408c9d2`). The gate helper is `_pty_quiescent_long_enough()` (`:390`), which keys off `mid_run_quiescent_since` — a MID-RUN signal that is None during prime, so it cannot be reused verbatim for the never-started path. The reusable primitive for prime is `last_pty_activity_at`.

**Pre-requisites:**
- AC requires measuring first-transcript-entry latency for granite SDLC PM/Dev prime before choosing liveness-gate vs. widened-grace. Instrumentation/measurement is step 1 of the plan.

**Dropped / out of scope:**
- A blanket increase of `NEVER_STARTED_GRACE_SECS` (would slow detection of genuinely dead sessions) — explicitly a No-Go per the issue.

> **Context for the reader.** This repo runs autonomous developer sessions (`session_type="eng"`) inside a **granite PTY container** — an interactive `claude` TUI driven headlessly by the worker (see `docs/features/granite-pty-production.md`). A background **session-health monitor** (`agent/session_health.py`) watches each running session and kills ones it believes are dead. It decides "dead" from **progress signals**: a session must write a transcript entry (or refresh a tool-use timestamp) within a grace window, or it is recovered/failed. "**Priming**" is the granite cold-start phase where the PM/Dev TUI pair boots before producing its first transcript entry.

## Problem

Granite SDLC sessions are being **killed during priming** because they don't produce their first progress signal fast enough — the inner `claude` TUI hasn't written a transcript entry yet, so the health monitor classifies the still-booting session as dead and fails it.

**Current behavior.** A psyoptimal SDLC session for issue #543 (slug `sdlc-543`) was launched three separate times on 2026-06-25 and died every time during granite priming, never producing real work output. The kills came under **two different classifications with the same root cause**:

- **`kind=tool_timeout`** (attempt 06:40–06:47 UTC): `tool-wedge: Skill (default tier) older than 300s`. `current_tool_name=Skill`, `communicated=False` on every heartbeat. Failed after 2 recovery attempts.
- **`kind=no_progress` / `never_started past grace`** (attempt 07:37–07:40 UTC; agent `8c7d618d507f406199d441590ee5fcc5`, session `-1003743854645_1782371970489`): the granite pair **spawned cleanly** —
  `[pty-pool] slot 0: spawned per-session pair (cwd=…/.worktrees/sdlc-543, pm_model=opus)` —
  but the inner session never wrote a first transcript entry. Logs show repeated
  `granite-container … prime-turn: transcript read: no-new-entry` and `steady-state turn 0: transcript read: no-new-entry`.
  It ran 322s with no progress → recovered to pending → picked up again → ran 170s with no progress → finalized `failed` ("2 recovery attempts, never progressed").

**Proof it is not a hard break.** A *different* granite session in the same project/worker (the `add-newsletter-01` task) ran ~700s and **completed successfully**. Granite PTY is not broken — the SDLC PM/Dev prime is simply slow enough to trip the killers before its first transcript entry lands.

**Desired outcome.** A granite session whose PTY process is alive and making progress at the OS level (process running, producing pty/stdout bytes, transcript file growing) must **not** be killed merely because it hasn't emitted a logical transcript entry within the grace window. Liveness should be judged on the PTY actually being alive, not solely on transcript/tool-output cadence.

## Definitions

|------|-----------|--------|
| Granite PTY container | Headless driver for an interactive `claude` TUI used to run eng/SDLC sessions | `docs/features/granite-pty-production.md`; `agent/granite_container/` |
| Priming / prime-turn | Granite cold-start phase before the first transcript entry is produced | log tag `granite-container … prime-turn` |
| Progress signal | A transcript entry or refreshed tool-use timestamp the monitor reads to judge liveness | `agent/session_health.py` `_has_progress()` |
| `no_progress` / `never_started past grace` | Kill classification: no progress observed past the startup grace window | `agent/session_health.py:770` `_never_started_past_grace()` |
| `tool_timeout` / tool-wedge | Kill classification: a tool (default tier) in flight longer than its budget | `agent/session_health.py:360` `_check_tool_timeout()` |
| Grace window | Minimum running time / startup margin before a no-progress session can be recovered | `AGENT_SESSION_HEALTH_CHECK_INTERVAL=300` (`:210`); padding in `_never_started_past_grace` |

## Related work

**Issue #1784 (CLOSED)** — *"session-health: 300s tool_timeout kills legitimate long-running SDLC tools (Bash/Skill) — gate on PTY liveness"* — addressed the **`tool_timeout`** half by gating the tool-wedge kill on PTY liveness (plan: `docs/plans/tool_timeout_pty_liveness_gate.md`). The production failure above shows the **`no_progress`/`never_started` path is still un-gated**: `_never_started_past_grace()` already pads the grace "to cover worst-case granite cold-start latency" (`agent/session_health.py:785`), but that padding was insufficient — the SDLC prime exceeded it and was killed anyway. This issue is the sibling of #1784 for the never-started path.

## Solution sketch (open questions — root cause not yet fully pinned)

This is a structural timing problem; do-plan should investigate before committing to one mechanism. Open questions:

1. **Why does the granite SDLC prime exceed the padded grace** while the newsletter session primed fine? Is it PM+Dev pair cold-start cost, opus `pm_model` load time, MCP/init latency, or the SDLC prime command specifically? Instrument the prime path and measure first-transcript-entry latency across granite session types.
2. **Should the `no_progress`/`never_started` kill be gated on PTY liveness** the same way #1784 gated the tool-wedge kill — i.e. if the PTY process is alive and emitting pty/stdout bytes (or the transcript file's mtime/size is advancing), suppress the never-started kill? Reuse #1784's liveness primitive if one exists.
3. **Or should the priming grace window be widened** specifically for granite SDLC PM/Dev pairs (vs. a blanket increase that would slow detection of genuinely dead sessions)?
4. **Reconcile the two classifications** — both `tool_timeout` and `no_progress` fired for the same underlying "still priming" condition depending on whether `current_tool_name` was set. The fix should cover both so the same prime can't fail under whichever label happens to apply.

## Downstream context

This issue will be consumed by `/do-plan` to produce a plan document at `docs/plans/{slug}.md`. Because the root cause (why this prime is slow) is not yet measured, do-plan should begin with an instrumentation/measurement step before selecting between the liveness-gate and widened-grace approaches.

## Acceptance criteria

- [ ] A granite session whose PTY is alive and advancing (process up + transcript file growing / pty bytes flowing) is not killed by the `no_progress`/`never_started` path during priming.
- [ ] The same protection covers the `tool_timeout` path during priming (no regression vs. #1784), so a slow prime cannot fail under either classification.
- [ ] First-transcript-entry latency for granite SDLC PM/Dev prime is measured and recorded, with the grace/liveness thresholds justified against that measurement.
- [ ] A regression test reproduces an alive-but-slow-priming granite session and asserts it is NOT recovered/failed before producing output.
- [ ] Genuinely dead sessions (PTY process exited, no bytes, transcript stalled) are still detected and failed within a bounded time.

tomcounsell:
## Summary
- Adds `_prime_pty_alive(entry, now) -> bool` helper in `agent/session_health.py` that returns True when a granite session's PTY is alive and advancing during prime, False when kill-eligible
- Gates the D0 never-started kill call-site (~line 3208) so slow-but-alive priming sessions are deferred instead of recovered/killed
- Adds `NEVER_STARTED_PTY_LIVENESS_SECS` constant (default 90s, env-overridable) in `agent/session_stall_classifier.py`

## Changes
- `agent/session_health.py`: added `_prime_pty_alive` helper (67 lines, inverted polarity vs `_pty_quiescent_long_enough`) + D0 gate + `never_started_pty_deferred` counter instrumentation
- `agent/session_stall_classifier.py`: added `NEVER_STARTED_PTY_LIVENESS_SECS` env-overridable constant (default 90)
- `tests/unit/test_never_started_recovery.py`: added `TestPrimePtyAlive` (9 cases incl. polarity guard test) + `TestPrimePtyAliveRecoveryPath` (2 cases)
- `tests/unit/test_session_health_tool_timeout.py`: added AC2 non-regression assertion for priming granite sessions
- `tests/unit/test_session_stall_classifier.py`: added `TestNeverStartedPtyLivenessSecs` (4 cases)
- `docs/features/pm-session-liveness.md`: documented both parallel PTY-liveness gates with side-by-side comparison table
- `docs/features/granite-pty-production.md`: added cross-reference noting prime liveness uses `last_pty_activity_at`

## Testing
- [x] 134 unit tests passing (all new + existing)
- [x] Ruff lint + format clean
- [x] Polarity guard test explicitly catches verbatim-copy-of-sibling error
- [x] Non-regression: SDK sessions, tool_timeout path, and existing `_never_started_past_grace` predicate all unaffected

## Documentation
- [x] `docs/features/pm-session-liveness.md` updated with two-gate comparison table
- [x] `docs/features/granite-pty-production.md` cross-reference added

## Definition of Done
- [x] Tested: 134 tests passing, all new cases covered
- [x] Reviewed: Verification table all green
- [x] Documented: Docs gate validated (2 files changed)

Closes #1792

### DIFF
--- a/agent/session_health.py
+++ b/agent/session_health.py
@@ -464,6 +464,72 @@ def _pty_quiescent_long_enough(entry: "AgentSession", now: datetime) -> bool:
     return (now - quiescent_aware).total_seconds() >= MID_RUN_QUIESCENCE_SECS
 
 
+def _prime_pty_alive(entry: "AgentSession", now: datetime) -> bool:
+    """Return True if the PTY is alive during priming (defer D0 kill); False if kill-eligible.
+
+    **INVERTED POLARITY vs ``_pty_quiescent_long_enough``**: this helper returns
+    ``True = session is alive, defer the kill`` and ``False = kill-eligible``.
+    ``_pty_quiescent_long_enough`` is the exact opposite (True = proceed to kill).
+    Do NOT copy its return values verbatim — a literal copy kills live sessions.
+
+    This is the PTY-liveness gate for the D0 never-started kill path (issue #1792).
+    Granite SDLC sessions can be killed during priming because the inner ``claude``
+    TUI is still booting when the ``NEVER_STARTED_GRACE_SECS + CONFIRM_MARGIN`` window
+    expires. If the PTY read loop is alive and the screen has shown recent activity,
+    the session is still booting and the kill should be deferred.
+
+    Branch order is load-bearing — first matching branch wins:
+
+    1. **Kill-switch first** (before any field reads): ``NEVER_STARTED_PTY_LIVENESS_SECS <= 0``
+       → return ``False`` (kill-eligible — disables deferral for all sessions).
+       Must be first so the kill-switch is never defeated by None-field short-circuits below.
+
+    2. **Non-PTY escape**: ``last_pty_read_loop_at is None`` → return ``False``
+       (SDK/non-granite sessions have no PTY loop; age-only kill is preserved).
+
+    3. **Stale read loop**: ``last_pty_read_loop_at`` older than ``HEARTBEAT_FRESHNESS_WINDOW``
+       (90s) → return ``False`` (dead read loop cannot prove liveness).
+
+    4. **Alive case**: read ``last_pty_activity_at``; if None or not a datetime → ``False``;
+       otherwise tz-normalize and return ``True`` iff activity is within
+       ``NEVER_STARTED_PTY_LIVENESS_SECS``.
+
+    This helper NEVER raises — all unexpected exceptions return ``False`` (kill-eligible).
+    """
+    try:
+        # Branch 1: kill-switch — disable PTY deferral, revert to age-only kill.
+        from agent.session_stall_classifier import NEVER_STARTED_PTY_LIVENESS_SECS
+
+        if NEVER_STARTED_PTY_LIVENESS_SECS <= 0:
+            return False
+
+        # Branch 2: non-PTY (SDK) session — no granite PTY, age-only kill preserved.
+        last_pty_read_loop_at = getattr(entry, "last_pty_read_loop_at", None)
+        if last_pty_read_loop_at is None:
+            return False
+
+        # Branch 3: stale read loop — a dead read loop cannot prove liveness.
+        if isinstance(last_pty_read_loop_at, datetime):
+            loop_at_aware = (
+                last_pty_read_loop_at
+                if last_pty_read_loop_at.tzinfo
+                else last_pty_read_loop_at.replace(tzinfo=UTC)
+            )
+            if (now - loop_at_aware).total_seconds() > HEARTBEAT_FRESHNESS_WINDOW:
+                return False
+
+        # Branch 4: alive case — PTY has a fresh read loop; check screen activity.
+        last_activity = getattr(entry, "last_pty_activity_at", None)
+        if not isinstance(last_activity, datetime):
+            return False
+        activity_aware = (
+            last_activity if last_activity.tzinfo else last_activity.replace(tzinfo=UTC)
+        )
+        return (now - activity_aware).total_seconds() <= NEVER_STARTED_PTY_LIVENESS_SECS
+    except Exception:
+        return False
+
+
 # In-process cache for ``_is_memory_tight()`` (issue #1099 Mode 4). Tuple of
 # ``(checked_at_monotonic, result)``. The cache amortizes psutil syscalls when
 # many sessions enter the recovery branch within the same health-check tick.
@@ -3136,6 +3202,21 @@ async def _agent_session_tool_timeout_check() -> None:
                                 "increment failed: %s",
                                 _ns_ctr_err,
                             )
+                        # PTY-liveness gate (#1792): defer the D0 kill when the granite
+                        # PTY is demonstrably alive (screen activity within the window).
+                        # SDK/non-PTY sessions fall through (helper returns False).
+                        if _prime_pty_alive(fresh_ns, now):
+                            try:
+                                from popoto.redis_db import POPOTO_REDIS_DB as _R_PTY_DEFER
+
+                                _R_PTY_DEFER.incr(
+                                    f"{fresh_ns.project_key}:session-health:"
+                                    f"never_started_pty_deferred"
+                                )
+                            except Exception:
+                                logger.debug("Failed to incr never_started_pty_deferred counter")
+                            continue
+
                         handle_ns = _active_sessions.get(fresh_ns.agent_session_id)
                         await _apply_recovery_transition(
                             fresh_ns,
--- a/agent/session_stall_classifier.py
+++ b/agent/session_stall_classifier.py
@@ -93,6 +93,13 @@ GRANITE_WEDGED_READLOOP_FRESH_SECS: int = int(
     os.environ.get("GRANITE_WEDGED_READLOOP_FRESH_SECS", "90")
 )
 
+# PTY-liveness deferral window for the never-started kill path (issue #1792).
+# When a session has never produced SDK output but is past the grace window,
+# the D0 kill is deferred if the PTY shows recent activity within this window.
+# Default: 90s (mirrors HEARTBEAT_FRESHNESS_WINDOW) — env-overridable.
+# Set to 0 or negative to disable the deferral (kill-switch).
+NEVER_STARTED_PTY_LIVENESS_SECS: int = int(os.environ.get("NEVER_STARTED_PTY_LIVENESS_SECS", "90"))
+
 # ---------------------------------------------------------------------------
 # Terminal statuses (mirror models/session_lifecycle.py — no circular import)
 # ---------------------------------------------------------------------------
--- a/docs/features/granite-pty-production.md
+++ b/docs/features/granite-pty-production.md
@@ -356,6 +356,15 @@ the operator does not want.
 > when the screen has been quiescent for `>= MID_RUN_QUIESCENCE_SECS (180s)`.
 > Worst-case recovery bound: ~330s (300s budget + ~30s tick cadence). SDK sessions
 > (no PTY) are unaffected — they continue to use the flat 300s age-only kill.
+>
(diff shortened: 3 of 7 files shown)


tomcounsell:
## Review: Approved

PR #1798 — `feat(session-health): gate never_started kill on PTY liveness (#1792)`

Preflight clean (`OPEN`, `MERGEABLE`, `CLEAN`). Reviewed against `docs/plans/never_started_pty_liveness_gate.md` and issue #1792.

### Summary

This adds a PTY-liveness gate to the D0 never-started kill path so granite SDLC sessions that are slow-to-prime (alive at the OS level but no transcript entry yet) are deferred instead of recovered/failed. It is the sibling of #1784's tool_timeout gate. The implementation is correct, well-tested, and well-documented.

### Highest-risk item: polarity — VERIFIED CORRECT

The plan flags the polarity inversion as "the single highest-risk line." Verified:
- `_pty_quiescent_long_enough` (`agent/session_health.py:391`) returns `True = kill-eligible`.
- `_prime_pty_alive` (`agent/session_health.py:467`) returns the **inverse**: `True = alive/defer`, `False = kill-eligible`. Branches 1/2/3 return `False` where the sibling returns `True`; the alive case returns `True` where the sibling returns `False`.
- Call-site (`:3208`) is a bare positive: `if _prime_pty_alive(fresh_ns, now): ... continue` — no `not`. Verification grep for `if not _prime_pty_alive` returns no match.
- A dedicated paired polarity-guard test (`test_polarity_guard_fresh_activity_defers_stale_activity_kills`) asserts both directions and would fail loudly on a verbatim copy.

### Correctness

- **Gate placement** (`:3208`): inside the D0 branch, after the fresh re-read and re-confirm of `_never_started_past_grace(fresh_ns, now)`, before `_apply_recovery_transition(reason_kind="no_progress")`. Race-mitigation re-read is reused. Correct.
- **Branch order** (`:486`-`:530`): kill-switch first, then non-PTY/SDK escape, then stale-read-loop escape, then `isinstance(datetime)`-guarded alive case. Matches the plan and the sibling's structure. `except Exception: return False` honors the NEVER-raises contract.
- **AC2 (tool_timeout path during prime)** is genuinely satisfied with no code change: the existing gate at `:3249` (`if tier == "default" and not _pty_quiescent_long_enough(entry, now)`) already defers an over-budget default-tier tool while the PTY is painting. The plan correctly determined only a non-regression assertion is needed.
- **Constant** `NEVER_STARTED_PTY_LIVENESS_SECS` (default 90, env-overridable, kill-switch at `<= 0`) added next to the neighboring never-started constants. No `NEVER_STARTED_GRACE_SECS` change (anti-criterion green — no blanket grace bump).

### Verification table — all green

- 134 unit tests pass (`test_never_started_recovery.py`, `test_session_health_tool_timeout.py`, `test_session_stall_classifier.py`).
- `ruff check` + `ruff format --check` clean on both source files.
- Helper exists (2), constant defined (1), deferred counter present (2), no inverted-polarity call-site, no blanket grace bump — all pass.

### Documentation

`docs/features/pm-session-liveness.md` adds a thorough "PTY-liveness gates for kill paths" section with per-gate detail, a branch-logic table, and a side-by-side comparison of the two gates (including the opposite polarity). `docs/features/granite-pty-production.md` cross-references it. Accurate and complete.

### Blockers
- None

### Tech Debt
- None

### Nits (advisory — subjective, not patch-gated)
- `TestPrimePtyAliveRecoveryPath` is named "recovery path" and its docstrings say the D0 branch "must NOT call `_apply_recovery_transition`", but the tests only assert the helper's return value — they don't exercise the `_agent_session_tool_timeout_check` call-site. Helper-level polarity is fully covered, so this is fidelity-only.
- The AC2 non-regression test exercises the in-budget case (`_check_tool_timeout` returns None) rather than the over-budget-but-PTY-painting case that was the actual production failure. The over-budget+painting path is already covered by #1784's suite, so AC2 is satisfied; this is the more-rigorous variant left on the table.

Neither nit affects correctness; both are genuinely subjective test-fidelity preferences. Approving.

Issue #1808, opened 2026-06-26
Investigation: wedged-but-alive worker leaves sessions pending indefinitely despite 300s health backstop


tomcounsell:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/wedged-worker-pending-investigation.md

## Background

Issue #1804 hardened the notify listener's subscribe-time NUMSUB self-check. During plan review, an open question emerged: when a session sat *indefinitely* in `pending` status (the original symptom), the 300 s `_agent_session_health_check` backstop (`agent/session_health.py:2552`) should have re-scanned and nudged/started the worker. An indefinite hang therefore implies a **distinct defect** — a wedged-but-alive worker — not a dead notify subscription.

## Hypothesis

The worker process is alive (heartbeat OK, `ps` shows it) but is wedged in a state where:
- It cannot pick up new `pending` sessions from the queue, AND
- `_agent_session_health_check` either does not fire, fires but does not recover, or fires and concludes the worker is healthy when it is not.

## Investigation Scope

- [ ] Reproduce the "indefinitely pending" symptom *after* #1804 ships (new NUMSUB self-check rules out the notify subscription as the cause)
- [ ] Confirm whether `_agent_session_health_check` fires during the hang (check `logs/worker.log` for `"health check"` entries)
- [ ] Determine whether the wedge is in `_worker_loop`, the PTY pool, or the health check itself
- [ ] Check whether `_active_events[worker_key].set()` is being called but the worker loop is not waking

## Acceptance Criteria

A root-cause is identified OR the symptom cannot be reproduced after #1804 ships (in which case close as "not reproducible — resolved by #1804").

## Context

Spin-out from #1804 plan No-Gos. Do not investigate until #1804 is shipped — the notify-listener hardening is the first line of defense. If the symptom recurs after #1804, this issue is the right place.
## Recon Summary

Code-grounded reconnaissance performed at plan time (baseline commit c01485e4). The investigation has a concrete, evidence-backed primary hypothesis.

**Confirmed:**

- The health-check pending-session branch treats *any* non-`done()` worker future as "alive" and only nudges its event, never escalating (`agent/session_health.py:2557-2566`): `worker_alive = worker is not None and not worker.done()` → `if worker_alive: event.set(); continue`. A worker loop blocked inside an `await` (future pending, not done) is indistinguishable from a healthy idle worker; the 300s backstop sets an event the wedged loop never observes, then `continue`s — indefinitely.
[...]

Issue #824, opened 2026-04-08
bug: pub/sub notify listener loses session notifications due to inherited socket_timeout (regression from PR #784)


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/pubsub-notify-listener-socket-timeout.md

> **Context:** The Valor system runs a standalone worker process (`python -m worker`) that processes `AgentSession` records from Redis. Sessions are created by CLI tools like `valor_session create` and the Telegram bridge, both of which call `_push_agent_session()` in [`agent/agent_session_queue.py`](agent/agent_session_queue.py). PR #784 added a Redis pub/sub notification so the worker picks up new sessions within ~1 second instead of waiting for the 5-minute health check. This issue documents a regression in that mechanism.

## Problem

Sessions created via `valor_session create` (or any external CLI path) intermittently sit in `pending` state for up to 5+ minutes despite the pub/sub notification fix from PR #784.

**Current behavior:**
The `_session_notify_listener` worker coroutine inherits `socket_timeout=5` from the global `POPOTO_REDIS_DB` connection pool, which was configured for request-response commands. For a blocking pub/sub listener, this means the listener times out after 5 seconds with no messages, crashes, sleeps 5 seconds, then resubscribes. During the 5-second dead window between crash and resubscription, any session notification published to `valor:sessions:new` is permanently lost — Redis pub/sub is fire-and-forget with no buffering. Sessions created during that window are never picked up until the 5-minute health check fires.

Observed today: sessions `0_1775578955` and `0_1775578966` were stuck pending for 8+ minutes. Worker logs confirmed the listener was continuously cycling ("Timeout reading from socket" every 5s) but the notifications for these sessions never arrived. Manual republishing immediately unblocked them.

**Desired outcome:**
Sessions created via `valor_session create` are picked up by the worker within ~1 second, consistently. The pub/sub listener should block indefinitely between messages without cycling through reconnects.

## Definitions

| Term | Definition | Reference |
|------|-----------|-----------|
| `AgentSession` | A Redis-backed model representing one unit of work for the Claude agent; has a `status` field (pending → running → completed) | [`models/agent_session.py`](models/agent_session.py) |
| `_session_notify_listener` | An async coroutine in the standalone worker that subscribes to `valor:sessions:new` and calls `_ensure_worker(chat_id)` when a new session notification arrives | [`agent/agent_session_queue.py`](agent/agent_session_queue.py) |
[...]

PR #1809, opened 2026-06-26
fix(worker): notify listener NUMSUB self-check + VALOR_WORKER_MODE in plist (#1804)

### LINKED ISSUE #1804: Standalone worker runs in bridge mode (VALOR_WORKER_MODE unset); notify-listener miss strands sessions
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/standalone-worker-mode-notify-listener-selfheal.md

## Summary
The standalone worker service (`python -m worker`, `com.valor.worker.plist`) does **not** set `VALOR_WORKER_MODE=standalone`. Per `_worker_loop` (`agent/agent_session_queue.py`), unset = **bridge mode**, where worker loops 'run until queue empty, then exit' and rely entirely on the `_session_notify_listener` to spawn fresh loops for new work.

This makes session processing fragile: if the notify listener fails to subscribe on a given boot, re-enqueued/new sessions are never picked up.

## Observed
- `ps eww` on the live worker: no `VALOR_WORKER_MODE` → bridge mode.
- On one boot, `PUBSUB NUMSUB valor:sessions:new` returned **0** (notify listener never subscribed / never logged 'Session notify listener subscribed'), even though the reflection scheduler and other tasks were running. Re-enqueued session 166 sat `pending` indefinitely; a clean restart fixed it.

## Fix
- Set `VALOR_WORKER_MODE=standalone` in `scripts/install_worker.sh` / the worker plist so loops wait indefinitely instead of exiting after drain.
- Make the notify-listener subscription robust + self-healing (verify NUMSUB≥1 after startup; re-subscribe on failure), so a missed subscription can't silently strand the queue.

## Impact
Sessions enqueued while the listener is down (or after ephemeral loops exit) are never processed until the next restart.
## Recon Summary

Investigated against `main` @ HEAD (commit baseline noted in the plan's Freshness Check).

**Confirmed**
- `com.valor.worker.plist` `EnvironmentVariables` and `scripts/install_worker.sh` do **not** set `VALOR_WORKER_MODE`. So the *launchd launch environment* lacks it (what `ps eww` reports).
- `_session_notify_listener` (`agent/agent_session_queue.py:786`) subscribes to `valor:sessions:new` and logs, but **never verifies the subscription took** (no `PUBSUB NUMSUB` check). A silently-failed subscribe leaves no self-heal beyond a full thread restart on an actual exception.

**Revised (material)**
- The worker **does** run in standalone mode *at runtime*: `worker/__main__.py:703` calls `os.environ.setdefault("VALOR_WORKER_MODE", "standalone")` inside `main()`, before `asyncio.run(_run_worker)` at line 729; `_worker_loop` reads `os.environ.get("VALOR_WORKER_MODE")` at line 1250 and sees `"standalone"`, so loops wait **indefinitely** (`await event.wait()`) rather than exiting after drain. This `setdefault` has been present since commit `cd5d5df4` (#737, 2026-04-06), predating this issue. `ps eww` shows the launch env, not runtime `os.environ` mutations — so "no `VALOR_WORKER_MODE` ⇒ bridge mode" is a misdiagnosis. Setting it in the plist is **defense-in-depth / observability**, not the behavior fix.

**Backstop already exists**
- `_agent_session_health_check` (`agent/session_health.py:2552`) re-scans `status="pending"` every `AGENT_SESSION_HEALTH_CHECK_INTERVAL=300s`: it nudges live workers' events, and starts a worker for orphaned pending sessions older than `AGENT_SESSION_HEALTH_MIN_RUNNING=300s`. A stranded pending session should therefore recover within ~5–10 min, **not indefinitely** — the observed "sat pending indefinitely" needs an alternate root cause (e.g. a worker that is `alive` but wedged, so the nudge path keeps `continue`-ing).

**Prior art**
- #824 — notify listener lost notifications via inherited `socket_timeout` (fixed by the dedicated `socket_timeout=None` connection now in `_listen_in_thread`). Plan: `docs/plans/pubsub-notify-listener-socket-timeout.md`.

**Real, in-scope work**: harden the notify listener with a post-subscribe `NUMSUB>=1` self-check + re-subscribe, plus a periodic liveness check; and make `VALOR_WORKER_MODE=standalone` explicit in the plist/installer for observability.

tomcounsell:
## Summary

- **Confirmed-bug fix**: `_listen_in_thread` now verifies `PUBSUB NUMSUB valor:sessions:new >= 1` after `subscribe()`, with up to 3 retries (~300 ms). If NUMSUB reports 0, a WARNING is logged and the function returns early — the outer `while True` loop re-subscribes after its 5 s backoff. All checks run on the listener's own `socket_timeout=None` connection (no cross-thread machinery; #824 invariant preserved).
- **Observability/defense-in-depth**: `com.valor.worker.plist` now sets `VALOR_WORKER_MODE=standalone` in `EnvironmentVariables`, making `ps eww` inspection unambiguous. Runtime was already standalone via `os.environ.setdefault` in `worker/__main__.py:main()`.
- **Tracking issue**: #1808 filed for the distinct "wedged-but-alive worker" defect (the "indefinitely pending" symptom, which the 300 s health backstop should cover but didn't in at least one observation).

## Files changed

| File | Change |
|------|--------|
| `agent/agent_session_queue.py` | Subscribe-time NUMSUB self-check in `_listen_in_thread` |
| `com.valor.worker.plist` | `VALOR_WORKER_MODE=standalone` in `EnvironmentVariables` |
| `scripts/install_worker.sh` | Belt-and-suspenders note in installer output |
| `tests/unit/test_agent_session_queue_async.py` | 3 new unit tests (`TestNotifyListenerNusubSelfCheck`) |
| `tests/integration/test_session_notify.py` | Extended socket-timeout test to assert `pubsub_numsub` is called |
| `docs/features/bridge-worker-architecture.md` | Document NUMSUB self-verify + plist change |
| `docs/features/README.md` | Index entry updated |

## Test plan

- [x] `tests/unit/test_agent_session_queue_async.py::TestNotifyListenerNusubSelfCheck` — 3 cases: NUMSUB≥1 happy path, NUMSUB==0 → early return + WARNING, raising NUMSUB → no crash + WARNING
- [x] `tests/integration/test_session_notify.py::TestSessionNotifyListener::test_notify_listener_uses_no_socket_timeout` — extended to assert `pubsub_numsub` is called on the `socket_timeout=None` connection
- [x] All 19 tests in the affected test files pass

Closes #1804


tomcounsell:
## Review: Changes Requested

**Branch:** `session/standalone-worker-mode-notify-listener-selfheal` → `main`
**Plan:** `docs/plans/standalone-worker-mode-notify-listener-selfheal.md` (tracking #1804)
**Preflight:** OPEN · MERGEABLE · CLEAN

Reviewed against the plan. The core fix is sound: the subscribe-time `PUBSUB NUMSUB >= 1` self-check runs on the listener's **own** `socket_timeout=None` connection (NUMSUB on `conn`, the regular command connection — never the `pubsub` subscribe-mode connection), preserving the #824 invariant. On a confirmed 0 it returns early and lets the outer `while True` re-subscribe after the 5 s backoff. The plist `VALOR_WORKER_MODE=standalone` is correctly framed as observability/defense-in-depth (runtime was already standalone via `setdefault`), and `plutil -lint` passes. The `[#1808]` No-Go (indefinite-hang → separate wedged-worker investigation) is respected. All 10 affected tests pass.

### Blockers
- None

### Tech Debt
**File:** `tests/unit/test_notify_listener_numsub.py:1` (verified: read this file)
**Code:** `class TestNUMSUBSelfCheck:` (3 methods: `test_numsub_ok_proceeds_to_listen`, `test_numsub_zero_returns_early_with_warning`, `test_numsub_raises_returns_early_with_warning`)
**Issue:** This new 130-line file duplicates the exact three NUMSUB cases already added in `tests/unit/test_agent_session_queue_async.py::TestNotifyListenerNusubSelfCheck` (happy path → listen, NUMSUB==0 → early return + WARNING, NUMSUB raises → no crash + WARNING). The plan's ## Test Impact specifies the self-check cases go in `test_agent_session_queue_async.py` only and states "all changes are additive guards" — this standalone file is out-of-scope redundant coverage that doubles maintenance and runtime for zero added value (NO LEGACY CODE TOLERANCE).
**Severity:** tech_debt
**Fix:** Delete `tests/unit/test_notify_listener_numsub.py`; keep the canonical cases in `test_agent_session_queue_async.py`. (The `test_notify_listener_numsub.py` variant additionally asserts `pubsub.unsubscribe()` teardown ran — if that assertion is worth keeping, fold it into the existing `TestNotifyListenerNusubSelfCheck` cases before deleting.)

### Nits
**File:** `agent/agent_session_queue.py:872` (verified: read this file)
**Code:** `"Session notify: NUMSUB check reports 0 subscribers after subscribe "`
**Issue:** On the exception path, `pubsub_numsub()` raising logs `"NUMSUB check raised ..."` then `break`s; the post-loop `if not _numsub_ok` block then logs a *second* WARNING claiming NUMSUB "reports 0 subscribers" — misleading, since it raised rather than reported 0.
**Severity:** nit
**Fix:** Set a sentinel on the raise path (or branch the final WARNING text) so the "reports 0" message only fires when NUMSUB actually returned 0 across all retries.

**File:** `agent/agent_session_queue.py:868` (verified: read this file)
**Code:** `import time as _time`
**Issue:** `import time as _time` sits inside the retry `for` loop, re-executing each iteration, and the `_time.sleep(0.1)` also fires after the final (3rd) failed attempt before giving up — an unnecessary ~100 ms delay on the give-up path.
**Severity:** nit
**Fix:** Hoist the `import` to the top of `_listen_in_thread`, and skip the sleep on the last attempt (e.g. `if _attempt < 2: _time.sleep(0.1)`).

### Legacy Cruft
- The duplicate test file above is the only cruft signal in the diff. No deprecated fields, dead imports, or stale comments introduced.

Tech debt and non-subjective nits are patched by the pipeline. Routing to `/do-patch`.

tomcounsell:
## PR Re-Review — APPROVED

Re-review after patch `b11d8832`. The three prior-review findings (1 tech-debt + 2 nits) are all resolved; nothing regressed.

**Verified fixes:**
1. **Redundant test file deleted** — `tests/unit/test_notify_listener_numsub.py` (130 lines) removed; its three NUMSUB cases already live in `TestNotifyListenerNusubSelfCheck`. The `unsubscribe()` teardown assertion was folded into both `test_numsub_zero_skips_listen_and_logs_warning` and `test_numsub_raises_no_crash_and_logs_warning` before deletion.
2. **Misleading double WARNING fixed** — the `pubsub_numsub()` raise path now `return`s instead of `break`, so only `"NUMSUB check raised"` fires on the exception path; `"NUMSUB check reports 0 subscribers"` now fires only for the genuine 0-count case.
3. **Import + pointless sleep cleaned up** — inline `import time as _time` removed (module-level `time` imported at line 27); `if _attempt < 2: time.sleep(0.1)` skips the no-op final-attempt sleep before the give-up WARNING.

**Verification:**
- `tests/unit/test_agent_session_queue_async.py` — 15/15 passing (incl. all three `TestNotifyListenerNusubSelfCheck` cases).
- `ruff check` + `ruff format --check` clean on both changed files.
- PR state: MERGEABLE / CLEAN. No CI checks configured on the branch (normal for this repo).

**Verdict: APPROVED** — 0 blockers, 0 tech-debt.

PR #1812, opened 2026-06-29
fix(granite): re-enqueue timed-out deliveries + throttle startup reflection batch (#1805)

valorengels:
## Summary

Fixes #1805 — a granite-generated reply (~1735 chars) silently vanished when a CEO messaged right after a worker restart. The startup reflection burst saturated the single event loop, the delivery coroutine couldn't be scheduled within 30s, `future.result()` raised `TimeoutError`, and the catch-all `except` dropped the payload with only a dashboard event.

Two levers (defense in depth):

1. **Outbox re-enqueue on delivery timeout** (`agent/granite_container/bridge_adapter.py`). On a delivery timeout or loop-closed condition, `_deliver_sync` re-enqueues the reply to `telegram:outbox:{session_id}` via `_enqueue_to_outbox` — the same 6-field JSON dict shape as `output_handler.py` (`chat_id`, `reply_to`, `text`, `session_id`, `timestamp`, optional `file_paths`) plus `expire(3600)` — so the relay still delivers it. Delivery-failure events now distinguish `recovered_via_outbox` from `dropped` (double-failure).

2. **Startup reflection-batch throttle** (`agent/reflection_scheduler.py`). `tick()` caps function-type reflection dispatch at `REFLECTION_STARTUP_MAX_CONCURRENT` per tick (env-overridable, default 4) and `await asyncio.sleep(0)` between dispatches; excess overdue reflections defer to the next tick (~60s). Agent-type reflections (already awaited serially) are unchanged. No new event loop / subprocess.

## Critique concerns addressed

1. **Same-thread branch (correctness):** the fire-and-forget branch now attaches a done-callback that re-enqueues to the outbox iff the scheduled task fails or is cancelled — no silent loss, no duplicate on success.
2. **Full 6-field dict + expire (correctness):** `_enqueue_to_outbox` reconstructs the canonical relay payload dict and calls `expire(_OUTBOX_TTL=3600)` — not the raw string. Inlined as a single local helper (no cross-module builder).
3. **`future.cancel()` return check:** the return value is checked; when it returns `False` (coroutine already running) the event is tagged `[future_uncancellable_possible_duplicate]`. `bridge/redundancy_filter.py::should_suppress` is the verified downstream backstop.
4. **Env knobs:** `GRANITE_DELIVERY_TIMEOUT_S` and `REFLECTION_STARTUP_MAX_CONCURRENT` declared in `.env.example` (flat-name convention, matching `TOOL_TIMEOUT_*`).
[...]

Issue #1805, opened 2026-06-26
Granite send_cb delivery times out under startup reflection-batch event-loop contention (replies dropped)


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/granite-send-cb-delivery-timeout-reflection-contention.md

## Summary
After a worker restart, the reflection scheduler runs its full startup batch (~31 reflections, several agent-type) on the single asyncio event loop. The granite container's `send_cb` — the delivery callback that routes a session's reply to Telegram, scheduled onto that **same** event loop (`agent/granite_container/bridge_adapter.py`) — times out before it can run. The reply is dropped (`delivered=False`) and **never reaches the chat.**

## Observed (session tg_cyndra_-1003900483201_166)
```
[granite-container] ... PM prime done / Dev prime done / steady-state turn 0
[bridge-adapter] send_cb delivery raised:  (payload=1735 chars)   <- empty str == TimeoutError
session_events: granite_delivery_failure 'delivery failed: TimeoutError:  (1735 chars)'
                granite_user_routed '[/user] routed (1735 chars)' delivered=False
```
The 1735-char reply was generated correctly but lost; the telegram outbox for the session was empty (timed out before enqueue). The session then loops (generate → delivery-timeout → retry).

## Root cause
Event-loop contention: a CPU/IO-heavy startup reflection batch starves the delivery coroutine on the shared loop, so `asyncio.wait_for(send_cb, timeout_s)` expires.

## Fix options
- Run reflections off the session event loop (dedicated executor/thread/process), or throttle/defer the startup batch so it doesn't run all at once at boot.
- Increase the send_cb delivery timeout and/or retry delivery on TimeoutError (the payload is known — re-enqueue to the outbox instead of dropping).
- Surface dropped deliveries more loudly (currently only a WARNING + session_events entry; no user-visible signal by design).

## Related
Surfaced while recovering from #1803 / the reflections-symlink worker wedge.

---

## Recon Summary

Verified against current `main` (baseline `ce0cc37c`).

**Confirmed:**
- `agent/granite_container/bridge_adapter.py:83` — `DEFAULT_DELIVERY_TIMEOUT_S = 30.0`; delivery timeout is `self._delivery_timeout_s`, defaulting to this constant (no env override).
[...]

PR #1858, opened 2026-07-02
Fix stale test asserting wrong _deliver_sync no-loop return contract

### LINKED ISSUE #1838: bridge_adapter._deliver_sync returns recovered (True) on no-loop path but test_no_loop_returns_false expects False
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/deliver-sync-no-loop-return-contract.md

> Surfaced during the #1837 granite failure-simulation harness build (Wave 1). Pre-existing on \`main\` — not caused by that work, and left untouched there to honor a test-only constraint.

## Problem

\`tests/unit/granite_container/test_bridge_adapter.py::TestDeliverSyncReturnsBool::test_no_loop_returns_false\` fails on \`main\`:

\`\`\`
AssertionError: True is not false
WARNING agent.granite_container.bridge_adapter:bridge_adapter.py:928 [bridge-adapter] no captured event loop for send_cb; re-enqueueing to outbox (loop=None)
\`\`\`

At \`agent/granite_container/bridge_adapter.py:~928\`, when there is no live event loop, \`_deliver_sync\` re-enqueues the payload to the outbox and \`return recovered\` — which is \`True\` when the outbox re-enqueue succeeds. The test asserts the no-loop path returns \`False\`.

## The question (investigation, not a confirmed defect direction)

This is a **semantic mismatch** between the test and the outbox-recovery feature (\`granite_delivery_recovered_via_outbox\`). Either:
- **The test is stale** — the no-loop path was intentionally changed to return the recovery result (\`True\` on successful outbox re-enqueue), and the test's \`assertFalse\` predates that change; or
- **The code is wrong** — callers rely on \`False\` meaning "not delivered synchronously," and returning \`True\` for an outbox-deferred delivery misleads them.

Resolve which is correct, then either update the test to assert \`True\`/the recovery contract, or fix the return value. Check callers of \`_deliver_sync\` to see what the boolean is used for before deciding.

## Impact

Currently \`pytest tests/unit/granite_container/ -n0\` does not exit 0 on \`main\` solely because of this one test. Low functional risk (the outbox re-enqueue itself works), but it pollutes the granite unit-suite baseline.

tomcounsell:
## Summary
- Renamed `test_no_loop_returns_false` to `test_no_loop_returns_outbox_recovery_result` in `tests/unit/granite_container/test_bridge_adapter.py`, rewriting it to assert both branches of `BridgeAdapter._deliver_sync`'s no-loop path (mocking `_enqueue_to_outbox` to return `True`/`False`) instead of a stale hardcoded `assertFalse`.
- Clarified the `_deliver_sync` docstring in `agent/granite_container/bridge_adapter.py` to state the precise return contract: delivered-or-queued -> `True`; neither -> `False`. No production control flow or return statements changed.

## Investigation resolution (per the plan)
PR #1812 (issue #1805) intentionally changed the no-loop path to re-enqueue to the outbox and return the recovery result, so a reply is never silently lost. The test was not updated at the time, creating a stale-test/correct-code mismatch. This PR fixes the test to match the deliberate, correct production behavior.

## Changes
- `tests/unit/granite_container/test_bridge_adapter.py` — renamed/rewrote the no-loop test, updated class docstring
- `agent/granite_container/bridge_adapter.py` — docstring-only clarification on `_deliver_sync`

## Testing
- [x] `pytest tests/unit/granite_container/test_bridge_adapter.py -n0 -q` -> 32 passed
- [x] `pytest tests/unit/granite_container/ -n0 -q` -> 554 passed, 5 skipped, 9 subtests passed
- [x] `git diff main -- agent/granite_container/bridge_adapter.py | grep -E "^[-+].*return (recovered|False|True)"` -> zero matches (no return statement changed)
- [x] Format clean (`ruff format --check`)

## Documentation
- [x] Plan explicitly states no `docs/features/` doc needed; inline docstring clarification is the deliverable

## Definition of Done
- [x] Built: test rewrite + docstring clarification implemented
- [x] Tested: full granite unit suite green (554 passed, 5 skipped)
- [x] Documented: docstrings updated
- [x] Quality: ruff format/check pass

Closes #1838

### DIFF
--- a/agent/granite_container/bridge_adapter.py
+++ b/agent/granite_container/bridge_adapter.py
@@ -1074,9 +1074,13 @@ class BridgeAdapter:
         captured loop ref is mandatory for async send_cbs. A sync
         send_cb is called directly on the calling thread.
 
-        Returns True on confirmed delivery, False on any failure or
-        timeout. The caller (_make_user_callback / _make_complete_callback)
-        uses the return value to set self._user_facing_routed (issue #1647).
+        Returns True when the payload is delivered synchronously or
+        re-enqueued to the outbox before returning; False when neither
+        happened (sync send_cb raised, a pre-scheduling error occurred,
+        or the same-thread fire-and-forget path where outbox recovery
+        is deferred to a done-callback). The caller
+        (_make_user_callback / _make_complete_callback) uses the return
+        value to set self._user_facing_routed (issue #1647).
 
         On a missing/closed loop or a delivery timeout, the error
         is logged and a session_events entry is appended. The
--- a/tests/unit/granite_container/test_bridge_adapter.py
+++ b/tests/unit/granite_container/test_bridge_adapter.py
@@ -739,7 +739,8 @@ class TestBridgeAdapterSpawnOnAcquire(unittest.TestCase):
 
 
 class TestDeliverSyncReturnsBool(unittest.TestCase):
-    """_deliver_sync returns True on confirmed delivery, False on failure (concern C1, #1647)."""
+    """_deliver_sync returns True on delivery or outbox recovery, False on
+    failure (concern C1, #1647)."""
 
     def _make_adapter_with_sync_cb(self, cb):
         """Adapter with a sync send_cb (not a coroutine) for testing."""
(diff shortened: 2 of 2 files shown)


tomcounsell:
## Review: APPROVED

**Branch:** `session/deliver-sync-no-loop-return-contract` → `main`
**Plan:** `docs/plans/deliver-sync-no-loop-return-contract.md` (closes #1838)
**Scope:** test rewrite + docstring clarification, no production control-flow change (+20 / -7)

### Goal alignment
The plan resolves the #1838 investigation as "the test is stale; the production code is correct." I verified this against the source:

- `_deliver_sync` no-loop path (`bridge_adapter.py:1116-1118`) does `recovered = self._enqueue_to_outbox(...)` → `_record_delivery_event(..., recovered=recovered)` → `return recovered`. That is the deliberate #1805 / PR #1812 behavior (a reply is never silently lost), unchanged by this PR.
- The updated docstring accurately describes all four return sites, including the same-thread fire-and-forget path (`:1137-1165`) that correctly returns `False` because its outbox enqueue is deferred to a done-callback. The distinction is legible and matches the code byte-for-byte.

### Test quality
The rewritten `test_no_loop_returns_outbox_recovery_result` is a strict improvement over the old `assertFalse`:
- Exercises the no-loop branch correctly (async cb → skips the sync path → `_loop = None`).
- Mocks `_enqueue_to_outbox` to pin **both** branches (`True → True`, `False → False`) and asserts `assert_called_once_with(1, "hello", None)`, so the contract is encoded rather than depending on ambient Redis accepting the rpush.

### Verification (plan `## Verification` table — all pass)
- No-loop test class: `3 passed`, exit 0
- Granite unit baseline `tests/unit/granite_container/ -n0`: **554 passed, 5 skipped**, exit 0
- Stale name `test_no_loop_returns_false` gone: grep exit 1
- New contract test present: grep exit 0
- Production returns unchanged: `git diff main` return-statement match count **== 0**
- `ruff format --check` + `ruff check`: clean

### Gates
- Docs gate: plan explicitly declares no `docs/features/` file needed; the inline docstring is the deliverable. Consistent.
- Plan-section compliance: Documentation, Update System, Agent Integration, Test Impact all present.
- UI gate: no UI files in the diff — no screenshots required.
- Bridge/agent change: touches `agent/granite_container/` but is docstring + test only, so no runtime behavior changes and no service restart is required.

### Findings
- Blockers: none
- Tech debt: none
- Nits: none

Clean approve. (Posted as a comment because self-approval via `gh pr review --approve` is disallowed by GitHub for a self-authored PR.)

Issue #1843, opened 2026-07-02
Granite silent wedges: wire existing signals (granite_wedged → recovery, CLI-hook liveness fields, per-iteration PTY callback) — no new detector


tomcounsell:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/granite-wire-silent-wedge-signals.md

> **Context.** Granite (the PTY-driven PM/Dev session runner, `agent/granite_container/`, `docs/features/granite-pty-production.md`) wedges silently in production. A reconciliation pass (2026-07-02) mapped every silent-failure shape against the detection layers that exist today and the resilience workstream (#1817/#1820/#1821). **Verdict: a new "silent-failure detector" is NOT warranted** — the signals mostly already exist; three of them are simply not wired. This issue is those three wiring fixes, deliberately narrow, with explicit deconfliction against #1820/#1821.

## Problem

Granite sessions are semi-invisible to the session-health machinery that was built to protect them:

**Gap 1 — `granite_wedged` is advisory-only.** The #1724 mid-run wedge detector already computes the right signal: fresh `last_pty_read_loop_at` + stale `last_pty_activity_at` yields a `granite_wedged` verdict in `session_stall_classifier.py:271-306`. But that verdict is consumed ONLY by `reflections/stall_advisory.py:152` (a log/advisory). `session_health.py:29-32` imports only the `NEVER_STARTED_*` constants from the classifier — no kill/recovery path ever sees `granite_wedged`. The signal exists; nothing actuates on it.

**Gap 2 — liveness fields are structurally dead for granite.** `last_tool_use_at` / `current_tool_name` are written only by the in-process SDK hooks (`agent/hooks/pre_tool_use.py:391`, `agent/hooks/post_tool_use.py:74` → `liveness_writers.record_tool_boundary`). Granite's PM/Dev claude children run the **CLI hooks** (`.claude/settings.json:27,74`), whose `_update_agent_session` (`.claude/hooks/post_tool_use.py:443-494`) writes only `updated_at` + `tool_call_count`. Consequences: the #1270 tool-timeout tier loop requires `current_tool_name` non-null (`session_health.py:365`) and therefore **never fires for granite**, and #1820's planned progress-deadline sees only the coarse per-PM-turn `last_turn_at`.

[...]

Issue #1721, opened 2026-06-17
Granite sessions can't resume where they stopped: persist resume handles + loop cursor for lossless checkpoint resume after stall/stop and on user reply


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/granite_lossless_checkpoint_resume.md

> **Context for the reader.** "Granite" is this repo's interactive-TUI session runner: instead of headless `claude -p`, it drives **two real interactive Claude Code TUIs** over PTYs (a "PM" persona and a "Dev" persona) and relays messages between them in a loop until the work completes. The orchestration lives in `agent/granite_container/container.py` (`Container.run()`), with PTY spawning in `agent/granite_container/pty_driver.py` and a bounded slot pool in `agent/granite_container/pty_pool.py`. Background: `docs/features/granite-pty-production.md`. A granite run is bridge-originated (Telegram → worker → `BridgeAdapter.run()` → `Container.run()`).

## Problem

When a granite session dies (a stalled stream, a hang caught by the in-loop watchdog, a crash, or a worker restart) **or** when a user replies to a finished session, granite **cannot continue where it stopped**. It always starts a brand-new `Container` from turn 0 with two fresh Claude Code TUIs and zero prior transcript.

### Why this happens

The orchestration loop is **entirely ephemeral in-memory**. Three pieces of state hold a granite conversation, and only the first two survive a death:

1. **PM transcript** — durable on disk at `~/.claude/projects/<cwd-slug>/<uuid>.jsonl`
2. **Dev transcript** — durable on disk, same scheme
3. **The `Container` loop position** — whose turn it is, whether PM is mid-relay waiting on Dev, the turn index, the last classification — lives only in local variables inside `Container.run()` (`container.py:850-1175`) and is gone when the process exits.

Two specific, fixable misses make resume impossible today:

- **The PM/Dev Claude Code session UUIDs are generated fresh every run and never persisted** (`agent/granite_container/bridge_adapter.py:412-413`). These UUIDs are exactly the `--resume` handles, and we discard them at exit. (We *do* persist the derived `pm_transcript_path` / `dev_transcript_path`, just not the UUIDs themselves.)
[...]

PR #1889, opened 2026-07-03
fix(#1836): terminus link fast-path + granite PTY-PM resume gate + honest resume warning

### LINKED ISSUE #1836: Reply-to-Valor messages silently dropped: terminus misclassification + granite sessions can't be resumed
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/reply-drop-terminus-granite-resume.md

> **Context:** A [granite PTY session](docs/features/granite-pty-production.md) is how Eng/bridge sessions actually run today — an interactive `claude` TUI driven inside a container via a pseudo-terminal (PTY), as opposed to the older headless SDK-client (`claude -p`) path. `claude_session_uuid` is a field on `AgentSession` (`models/agent_session.py`) that stores the Claude Code transcript UUID needed to resume a completed session (`valor-session resume`). The [agent-judgment catchup scan](docs/features/agent-judgment-catchup.md) is the best-effort recovery pass that runs at the end of `/update`, re-scanning recent Telegram messages for ones that were never answered.

## Problem

Tom replied to a Valor message in the "Eng: Valor" Telegram chat and never got a response. Investigating surfaced two independent, real bugs — one that dropped the message initially, and a second, more serious one that blocked the correct recovery path (resuming/steering the original session thread) once the drop was discovered.

**Current behavior:**

1. **Reply misclassified as not needing a response.** Tom's message ("look here: https://github.com/BuilderIO/agent-native/tree/main/plans", msg 1081) is a reply to a prior Valor message that ended with an open question. `bridge/routing.py::should_respond_async` runs `classify_conversation_terminus()` on any reply-to-Valor message to decide RESPOND / REACT / SILENT. None of the fast-paths fire for this text — it's not a standalone question (`_STANDALONE_QUESTION_RE`), not an acknowledgment token or ≤1 word (`_ACKNOWLEDGMENT_TOKENS`), and "look" isn't in the imperative-verb allowlist (`_IMPERATIVE_VERBS` — issue #1318's fast-path 0) — so it fell through to the LLM classifier (Ollama-first, Haiku fallback), which returned something other than `RESPOND`. The bridge log recorded `should_respond=False (reply_to_valor=True)` and the message was silently skipped, both live and again later during the catchup re-scan.

2. **Resuming the original session is structurally impossible for granite PTY sessions.** Once the drop was found, the correct fix is to resume the session the reply was pointed at (`tg_valor_-1003449100931_1076`, status=`completed`) via `python -m tools.valor_session resume --id ... --message ...`. This failed:
   ```
   cannot resume: no transcript UUID stored (session was killed before first turn completed)
   ```
   But the session *did* complete a full turn and has a real, populated transcript at `dev_transcript_path` (e.g. `.claude/projects/-Users-valorengels-src-ai--worktrees-dev-f8089e23/a9f926ab-bfa1-48a8-863b-36ff55ec4a10.jsonl`) — the filename itself *is* the Claude session UUID (standard `~/.claude/projects/<cwd>/<uuid>.jsonl` layout). Tracing it: `agent/granite_container/bridge_adapter.py` (~line 640-642) persists `dev_transcript_path` onto the `AgentSession` after each turn but never derives/stores `claude_session_uuid` from it. `_store_claude_session_uuid` (`agent/sdk_client.py`) is only called from the headless SDK-client path, which the granite PTY path doesn't use. `resume_session()` (`tools/valor_session.py`) gates all resume/steering-into-thread flows on `claude_session_uuid` being non-null — so **no granite-driven session can ever be resumed today**, regardless of how the reply-classification issue above is fixed. This isn't an edge case: granite PTY is the primary session type for bridge-originated Eng work.

**Desired outcome:**
- A reply to Valor that references new information (a link, a pointer, a follow-up instruction) without matching an acknowledgment token or a standalone question is not silently classified as SILENT/REACT — it should default conservatively toward RESPOND, consistent with the "any classifier error → RESPOND" conservative-default principle already documented in `classify_conversation_terminus`'s docstring.
- `claude_session_uuid` is populated for granite PTY sessions (derived from `dev_transcript_path`'s basename) so `valor-session resume` and any reply-to-thread steering path actually work for the primary session type.

## Definitions

| Granite PTY session | An interactive `claude` TUI driven inside a container via PTY — how Eng/bridge sessions execute today | `docs/features/granite-pty-production.md` |
| `claude_session_uuid` | `AgentSession` field storing the Claude Code transcript UUID, required to resume a completed session | `models/agent_session.py`, `tools/valor_session.py` |
| Conversation terminus classification | The RESPOND/REACT/SILENT decision made for any reply-to-Valor message, combining fast-path regex rules with an LLM fallback | `bridge/routing.py::classify_conversation_terminus` |
| Agent-judgment catchup | Best-effort recovery scan (last step of `/update`) that finds Telegram messages an LLM judge believes were never answered | `docs/features/agent-judgment-catchup.md` |

## Prior Context

Related work:
- #1318: Added Fast-Path 0 (imperative continuation verbs) to `classify_conversation_terminus` after a similar misclassification dropped an explicit directive. The verb list there is narrow by design ("run, fix, merge, start... excluded... handled by the few-shot LLM prompt instead") — this issue shows the LLM prompt side of that split still misses valid replies.
- #1090: Added the "Valor asked a question" exception to the acknowledgment fast-path so short human answers aren't dropped — same family of bug (reply-to-Valor content dropped by an overly aggressive fast-path/classifier), different mechanism.
- #1061: Introduced `claude_session_uuid`-gated resume for `failed`/`killed` sessions — written before granite PTY became the default execution path, so it never accounted for the PTY transcript-file case.

## Solution Sketch

**Part A — reply classification:**
- Add few-shot examples (or a narrow fast-path, matching the style of #1318's `_IMPERATIVE_VERBS`) covering "share a link/reference with no explicit ask" replies, so `classify_conversation_terminus` defaults to RESPOND for them.
- Alternatively/additionally, tighten the LLM prompt used for the fallback path so it treats "new information with no closing signal" as RESPOND rather than SILENT/REACT.

**Part B — granite session resume:**
- In `agent/granite_container/bridge_adapter.py`, wherever `dev_transcript_path` is set/updated on the `AgentSession` (~line 640-642), also derive and persist `claude_session_uuid` from the transcript filename (basename minus `.jsonl`).
- Confirm this doesn't collide with the existing "Claude Code UUID is NOT written over the PM's `claude_session_uuid`" guard noted in `docs/features/agent-session-model.md:272` — that guard is about not overwriting a *different* field's owner; this fix targets the field's *first* population, which today never happens for granite sessions.
- Add a regression test asserting a completed granite session has a non-null `claude_session_uuid` and that `valor-session resume` succeeds against it.

## Recon Summary

**Confirmed:** 2 items — ready to include
- Root cause of Part A traced to `bridge/routing.py::classify_conversation_terminus`'s fast-path exclusions plus LLM fallback verdict, not a hardcoded "reply_to_valor → skip" rule.
- Root cause of Part B confirmed directly: `tg_valor_-1003449100931_1076` has `claude_session_uuid=None` and a real, non-empty `dev_transcript_path`, and `resume_session()` in `tools/valor_session.py:713` hard-gates on the former being non-null.

- None blocking; the two parts are independent and can land separately.

## Acceptance Criteria

- [ ] A reply-to-Valor message containing a link/reference with no explicit question, acknowledgment token, or imperative verb classifies as RESPOND (add a test case mirroring #1318's style for this scenario).
- [ ] A completed granite PTY session has `claude_session_uuid` populated automatically once its first turn completes.
- [ ] `python -m tools.valor_session resume --id <granite-session-id> --message "..."` succeeds against a real completed granite session in a test.
- [ ] No regression to the existing SDK-client-path `_store_claude_session_uuid` behavior.

## Downstream

This issue feeds into `/do-plan`, which will produce a plan document at `docs/plans/{slug}.md` (kebab-case, derived from this issue's title), then executed by `/do-build`. The planner should treat Parts A and B as separable — either could ship independently, and the plan should scope both with their own test coverage rather than a single combined patch.

tomcounsell:
Fixes #1836.

Two independent, confirmed bugs surfaced by a dropped reply-to-Valor message ("look here: <url>"):

## Part A — reply misclassified (bridge/routing.py)
Added **Fast-Path 1.5** to `classify_conversation_terminus`: a `not sender_is_bot`-gated branch, placed **between Fast-Path 1 and Fast-Path 2 (before `word_count` is computed)**, that returns RESPOND when a human reply is essentially a bare link/pointer (URL present, ≤3-word residual after URL removal, not an ack token). Bot-sender bare URLs still hit Fast-Path 1 → SILENT, unchanged. No LLM-prompt change; prose/non-URL pointers stay out of scope. (committed in `9ef1835f`)

## Part B — granite PTY sessions couldn't clear the resume gate (agent/granite_container/bridge_adapter.py)
`_persist_resume_handles` now mirrors the **PM** role handle's `claude_session_id` onto `agent_session.claude_session_uuid` so `resume_session()`'s gate stops hard-erroring (`cannot resume: no transcript UUID stored`). PM owns the human-facing thread (steering injects into PM's PTY). **PTY-PM only** — headless-PM population is deferred to #1843 (no PM-headless run path exists today). Rewritten each run with the current run's fresh PM UUID; never writes `None` over an existing value. The granite path deliberately does not call the SDK-client-only `_store_claude_session_uuid`.

## Part C — honest resume signal (tools/valor_session.py)
Added `ResumeResult.warning`. For a granite gate-pass (keyed on `resume_handles` being populated), `resume_session()` attaches `"resumed as a fresh session; prior-transcript re-entry pending #1721"`, surfaced by `cmd_resume`. SDK-client resumes return `warning=None`. True cold→warm transcript re-entry stays #1721's scope (its consumer reads `resume_handles`, not this overwritten-each-run scalar).

## Tests
- `test_routing.py` — human bare-URL → RESPOND; "look here: <url>" → RESPOND; multi-URL → RESPOND; bot-sender bare-URL → SILENT (unchanged); URL + substantive prose falls through to LLM (≤3-word guard holds).
- `test_session_executor_granite.py::TestPersistResumeHandlesUuidPopulation` — PTY-PM populates `claude_session_uuid`; `resume_session()` succeeds with the #1721 warning; null PM UUID does not clobber; persist failure logs and does not crash.
- `test_valor_session_resume_release.py::TestResumeGraniteReentryWarning` — granite resume returns the #1721 warning; SDK-client resume returns `warning=None`.

**Test results:** 102 passed, 1 skipped for the files/classes exercising these changes (serial run). Pre-existing failures in `TestExecutorGraniteWiring`/`TestReactionGating` are environmental (the real executor tries `git worktree add` under a non-git parent dir on this machine) and are untouched by this diff.

## Docs
- `docs/features/granite-pty-production.md` — new "Resume gate: `claude_session_uuid` population" section.
- `docs/features/agent-session-model.md` — Resume fields entry + drafter passive-isolation clarification.
- Scrubbed stale headless-PM fragments from the plan doc (deferred to #1843).

### DIFF
--- a/agent/granite_container/bridge_adapter.py
+++ b/agent/granite_container/bridge_adapter.py
@@ -763,7 +763,43 @@ class BridgeAdapter:
                         }
                     )
             self._agent_session.resume_handles = handles
-            self._agent_session.save(update_fields=["resume_handles", "updated_at"])
+            update_fields = ["resume_handles", "updated_at"]
+
+            # Part B (#1836): also mirror the PM handle's UUID onto the
+            # legacy scalar `claude_session_uuid` field so
+            # `resume_session()`'s gate (`tools/valor_session.py`,
+            # ~line 715) stops hard-erroring with "cannot resume: no
+            # transcript UUID stored" for granite sessions. PM (not Dev)
+            # is chosen because PM owns the human-facing thread --
+            # steering messages are only ever injected into PM's PTY
+            # (see `container.py`), so PM's transcript is the one a
+            # human-initiated resume should target.
+            #
+            # This is (re)written with a *fresh* UUID on every `run()` --
+            # NOT a one-time first-population. `claude_session_uuid`
+            # therefore only ever reflects the most recent run's PM
+            # transcript; it is not a durable resume anchor. Future
+            # consumers (#1721, cold->warm transcript re-entry) must read
+            # the `resume_handles` list above, not this scalar.
+            #
+            # The granite path deliberately never calls
+            # `_store_claude_session_uuid` (`agent/sdk_client.py`) --
+            # that helper is SDK-client-only, reached only from the
+            # headless SDK-client harness path (`get_response_via_harness`).
+            #
+            # Never write None over an existing value: only assign when
+            # the PM handle's `claude_session_id` is non-null. PTY-PM has
+            # its UUID known at spawn (pre-assigned via `--session-id`),
+            # so this always fires for the (in-scope) PTY-PM transport.
+            # Headless-PM handles are written with a null UUID here (see
+            # docstring above) -- population for that transport is
+            # deferred to #1843, which owns the PM headless read loop.
+            pm_handle = next((h for h in handles if h.get("role") == "pm"), None)
+            if pm_handle and pm_handle.get("claude_session_id"):
+                self._agent_session.claude_session_uuid = pm_handle["claude_session_id"]
+                update_fields.append("claude_session_uuid")
+
+            self._agent_session.save(update_fields=update_fields)
         except Exception as e:  # noqa: BLE001
             logger.warning("[bridge-adapter] resume_handles persist failed: %s", e)
 
--- a/bridge/routing.py
+++ b/bridge/routing.py
@@ -731,6 +731,10 @@ async def classify_needs_response_async(text: str) -> bool:
 # Lookahead: not followed by \w+= (query param name=value pattern)
 _STANDALONE_QUESTION_RE = re.compile(r"(?<![=&\w])\?|(?<![=&])\?(?!\w+=)")
 
+# Regex matching bare http(s) URLs, used by Fast-Path 1.5 to detect a reply
+# that is "essentially just a link/pointer".
+_URL_RE = re.compile(r"https?://\S+")
+
 # Fast-Path 0: imperative continuation verbs that must always RESPOND.
 #
 # These few-shot examples are mined from real misclassified messages where the
@@ -789,6 +793,10 @@ async def classify_conversation_terminus(
     0. human sender + imperative continuation verb at start of any line → RESPOND
        (issue #1318 — prevents SILENT misclassification of explicit directives)
     1. sender_is_bot + no question → SILENT  (primary loop-break signal)
+    1.5. human sender + reply is essentially just a bare link/pointer → RESPOND
+       (issue #1836 — prevents REACT/SILENT misclassification dropping a
+       shared link; must run before word_count is computed in Fast-Path 2,
+       since a bare URL is a single token)
     2. acknowledgment token or very short (≤1 word) → SILENT
        (unless thread_messages contains a question — then fall through, so a
        human short answer like "Yes"/"No" to a Valor question is not dropped;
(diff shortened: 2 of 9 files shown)


tomcounsell:
## Review: fix(#1836) — terminus link fast-path + granite PTY-PM resume gate + honest resume warning

**Verdict: APPROVED** (posted as a comment — self-authored PR, GitHub disallows self-approval)

Reviewed against plan `docs/plans/reply-drop-terminus-granite-resume.md` (Success Criteria + Verification table). Preflight clean: OPEN, MERGEABLE, mergeStateStatus=CLEAN. No UI files in the diff, so the visual-proof gate is a no-op.

### Correctness

**Part A — `bridge/routing.py` Fast-Path 1.5.** A `not sender_is_bot`-gated branch placed between FP1 and FP2, before `word_count` is computed. A bare URL is a single token, so placement is load-bearing: verified `Fast-path 1.5` precedes `word_count = len(...)` in source. Bot-sender bare URLs still hit FP1 -> SILENT (loop-break preserved). The `remainder not in _ACKNOWLEDGMENT_TOKENS and <=3 words` guard keeps ack tokens and substantive prose out of scope. Product decision (a human sharing a link is new info, not a closer) is explicit and documented.

**Part B — `agent/granite_container/bridge_adapter.py` `_persist_resume_handles`.** Mirrors the PM handle's `claude_session_id` onto the scalar `claude_session_uuid` only when non-null, so the resume gate stops hard-erroring for granite. `update_fields` appends `claude_session_uuid` only when the write actually happens. PTY-PM only; headless-PM correctly deferred to #1843. The null-clobber guard is real and tested.

**Part C — `tools/valor_session.py` `ResumeResult.warning`.** Keyed on `resume_handles` presence — granite is the sole writer of that field, so SDK-client resumes get `warning=None`. Additive, default-None field; `cmd_resume` surfaces it in both text and JSON only when set. No breaking change to existing callers.

### Verification Results
- Affected new tests: 102 passed, 1 skipped (serial run).
- Full plan Verification table: all grep/ordering/anti-criterion checks PASS; `ruff check` and `ruff format --check` clean on all changed files.
- The 13-16 `TestExecutorGraniteWiring`/`TestReactionGating` failures reproduce on `main` (real executor attempts `git worktree add` under a non-git parent on this machine) — pre-existing and environmental, untouched by this diff. Confirmed as disclosed in the PR body.
- Minor note (advisory, not a finding): the plan's "No stale xfails" Verification row greps `xfail` and matches the benign substring in a `--runxfail` comment in `tests/integration/test_granite_pty_production.py` (pre-existing, not in this diff). The row's intent holds; the grep pattern is over-broad.

### Blockers
- None

### Tech Debt
- None

### Nits
- None

Clean, well-commented, fully tested, docs updated (granite-pty-production.md, agent-session-model.md, plan scrubbed of stale headless-PM fragments). Every Success Criterion is met. Ready to merge.

PR #1890, opened 2026-07-03
fix(#1881): deliver PM's terminal turn when PM completes before Dev primes

### LINKED ISSUE #1881: bug: granite startup handshake drops PM's [/complete] reply when PM finishes before Dev primes (startup_unresolved)
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/granite-startup-pm-complete-before-dev-primes.md

## Summary

A granite session can do its work **perfectly**, have the PM emit a clean `[/complete]` reply, and still deliver **nothing** to the user — because the container's startup handshake gates message delivery on *both* PTYs (PM and Dev) reaching idle in the same cycle. When the **PM finishes before the Dev PTY has primed**, that overlap never occurs, the container burns to the `STARTUP_HARD_CEILING_S` ceiling, exits `startup_unresolved`, and the already-drafted reply is silently dropped.

This is distinct from #1647 (PM never routes `[/complete]`) and #1644 (PM's `[/dev]` never relayed): here the PM *did* route a clean completion. It's also distinct from #1710 (loud startup diagnostics): the startup "failure" is spurious — the session actually succeeded.

## Domain terms

- **PM PTY / Dev PTY** — the two `claude` sessions the granite container coordinates over PTYs (`agent/granite_container/container.py`). PM is the routing/user-relationship layer; Dev owns the SDLC pipeline.
- **Startup handshake** — the startup loop requires `pm_saw_idle AND dev_saw_idle` true in a single cycle before `startup_settled = True` (`container.py:1958`).
- **Prime-turn relay** — the code path (`container.py:1990+`) that reads the PM's prime-turn buffer and routes its classification/reply (`[/user]`/`[/complete]`/`[/dev]`) to the user. **It only runs after `startup_settled`.**
- **`startup_unresolved`** — exit reason (code 3) when startup never settles within `STARTUP_HARD_CEILING_S` (600s). `startup_failure_kind` is `ceiling` (hard-timeout) or `plateau` (early bail).

## Reproduction (observed in production)

Session `c220a40996b74cad9da696fb18afc042` — thread `tg_cyndra_-1003900483201_172`, project Cyndra, 2026-07-03.

User (Thabiso) asked Valor to verify a Notion blocker was cleared and move a card. Timeline (UTC):

| Time | Event |
|------|-------|
| 08:42:11 | Worker picks up session; **PM** PTY starts priming |
| 08:42:38–08:47:09 | **PM does the entire job correctly** — dispatches a Notion agent, verifies "Migrate Payload CMS into the slice" (#22) is Done, confirms blocker cleared, spawns a second Notion agent that moves the card Blocked → Ready (write lands 08:47:01), then drafts a full reply ending in `[/complete]` |
| **08:47:17** | **Dev** PTY only *now* begins priming — 8s *after* the PM already finished |
| 08:47:30 | Dev reaches ready/idle |
| ~09:03:23 | Container bails: `exit_reason=startup_unresolved`, `startup_failure_kind=ceiling`, returncode 0 |

Recorded fields:
```
exit_reason:          startup_unresolved
startup_failure_kind: ceiling
exit_returncode:      0
last_tool_use_at:     08:45:09   (PM)
completed_at:         09:03:23
```

**Outcome:** `chat_message_log` contains only the inbound message; the Telegram thread shows no Valor reply. The user's request *was fulfilled* (card moved) but he received zero confirmation and the session looked like a silent failure.

## Root cause

The PM raced ahead: it primed and immediately dove into ~5 min of substantive subagent work, so it was **busy (not idle)** throughout the startup window, then emitted `[/complete]` and stopped. The Dev PTY started late and only reached idle at 08:47:30 — *after* the PM was already done. The two PTYs' idle windows **never overlapped in a single cycle**, so `startup_settled` never became true, the **prime-turn relay never ran**, and the PM's completed reply was never handed to any send path. The container spun to the 600s ceiling and exited `startup_unresolved`.

The core defect: **delivery of an already-`[/complete]`d PM turn is gated on a startup-handshake condition that a fast-PM/slow-Dev ordering can permanently prevent.** A PM that completes before Dev primes strands its own reply.

## Expected behavior

- A PM turn that reaches `[/complete]` (or `[/user]`) should be delivered regardless of whether the Dev PTY has finished priming. PM→user delivery must not depend on Dev reaching idle.
- Startup should not classify a session as `startup_unresolved` when the PM has already produced a routable terminal turn.
- If Dev priming is genuinely needed for a turn, the PM's idle/completion should be latched so a later Dev-idle cycle still triggers the relay — the PM shouldn't have to *still* be idle at the moment Dev settles.

## Suggested fixes (for planning; not prescriptive)

1. **Latch PM idle/completion.** Once the PM has reached idle or emitted a terminal marker during the startup window, remember it; settle startup when the latched PM state + current Dev idle both hold, rather than requiring simultaneous same-cycle idle.
2. **Decouple PM→user delivery from the Dev handshake.** If the PM emits `[/complete]`/`[/user]` before startup settles, route it immediately; only the `[/dev]` path needs the Dev PTY.
3. **Treat a completed PM turn as a settle condition**, not a ceiling burn — a `[/complete]` observed during startup should end the startup phase successfully, not spin to 600s.

## Impact

- **Silent data loss of user-facing replies** on any granite session where the PM finishes before Dev primes — most likely on quick PM-only requests (status checks, board updates, Q&A) that never actually need the Dev.
- User sees no response and the session records as an unresolved startup failure, masking that the work succeeded.
- Erodes trust: the exact class of request the PM is meant to handle solo (fast routing/relationship work) is the class most prone to this race.

## Evidence

- PM transcript: `~/.claude/projects/-Users-valorengels-src-cyndra-consulting--worktrees-dev-c220a409/f98dc21a-7972-4916-b423-cf43fe5ff639.jsonl` (PM `[/complete]` at 08:47:09)
- Dev transcript: `…/89bf83f5-f67a-466a-b59f-0a1eead6f003.jsonl` (Dev prime starts 08:47:17)
- Session record: `AgentSession` `c220a40996b74cad9da696fb18afc042`
- Startup handshake / relay: `agent/granite_container/container.py:1958` (both-idle settle) and `:1990+` (prime-turn relay)

## Recon Summary

Recon performed 2026-07-03 against `main` @ 06fca807. Every file:line claim in this issue was independently re-read and confirmed unchanged since filing (no commits have touched `agent/granite_container/container.py` since 2026-07-03T10:50:53Z).

**Confirmed:**
- The both-idle settle gate is real: `container.py:1958` requires `pm_saw_idle and dev_saw_idle` true in the same cycle before `startup_settled = True` and `break`. Idle bools come from `_startup_cycle_idle` (`container.py:1622`) → `PTYDriver.read_until_idle` (`pty_driver.py:645`), which is edge-triggered per call.
- The prime-turn relay (the ONLY code path that delivers a PM `[/complete]`/`[/user]` terminal turn during startup) runs at `container.py:1990-2052`, strictly after the settle break. It classifies PM's transcript via `classify_pm_prefix` and delivers through `_route_pm_classification` → `on_complete_payload`/`on_user_payload` (`container.py:2308-2337`), setting `result.user_facing_routed = True`.
- The ceiling exit at `container.py:1976-1988` sets `exit_reason = "startup_unresolved"`, `startup_failure_kind = "ceiling"` and `return`s before the relay — so a never-settled startup delivers nothing.
- `PTYDriver.idle_marker` (`pty_driver.py:701`) is a UI slice of the bypass/overlay bar, NOT a `[/complete]` marker. Terminal-marker detection requires transcript classification, so the fix cannot rely on the idle read alone to know PM "completed".

**Revised:**
- Issue prose says the two idle windows "never overlapped in a single cycle." Mechanism confirmed: idle detection is edge-triggered and PM's prime (`container.py:1793`) blocks before Dev's prime (`container.py:1800`), so PM finishes and quiesces before Dev's PTY even begins priming; by the time Dev reaches idle, a fresh `_startup_cycle_idle(pm)` call sees no new PM paint. Latching PM idle (fix 1) is therefore both necessary and sufficient for the reported case (Dev did reach idle ~21s after PM). Full decoupling (fix 2/3) is additionally required to satisfy the stated expected behavior "PM→user delivery must not depend on Dev reaching idle" for the case where Dev never primes.

**Pre-requisites:**
- None external. The fix is contained to the startup loop in `container.py`. Existing deterministic unit harness (`_mock_pm`/`_mock_dev`/`_idle_result` in `tests/unit/granite_container/`) can reproduce the race without a live PTY.

**Dropped:**
- Nothing dropped from the issue scope. Related #1882 (mean 👎 reaction on `startup_unresolved`) is a separate policy bug and stays out of scope here.

tomcounsell:
## What

Fixes #1881 — the granite startup handshake dropped PM's `[/complete]` reply when the PM finished before the Dev PTY primed.

A fast PM-only request (status check, Q&A, board update) primes, does its work, emits a clean `[/complete]`, and quiesces **before** the Dev PTY has finished (or even begun) priming. The old startup loop only settled when PM and Dev were observed idle in the **same** cycle, so the two never coincided, the prime-turn relay never ran, and the container burned to the 600s ceiling and exited `startup_unresolved` — the user's request was fulfilled but the confirmation was silently dropped. Reproduced in production (session `c220a40996b74cad9da696fb18afc042`).

## Changes (all in `agent/granite_container/container.py`)

**Task 2a — latch + relay-baseline relocation (fixes the reported incident end-to-end):**
- **Relocated the relay's `pm_prime_baseline` snapshot** from after the startup loop to **before** `_prime_session(pm)`, reading `result.pm_transcript_path`. This is the blocker fix: with the baseline captured before PM emits any task-response output, PM's terminal turn is always a *new* text-bearing entry, so the relay's `last_assistant_text(..., baseline_text_count=...)` freshness guard delivers it for any break reason. Snapshotting after the loop already counted a fast PM's flushed `[/complete]`, so the guard returned `""` and the reply was dropped.
- **Added the `pm_ever_idle` latch** so PM's idle observation persists across cycles; the settle gate now reads `pm_ever_idle and dev_saw_idle` instead of requiring both in the same cycle. Headless-Dev path is byte-identical (it already forces `dev_saw_idle=True`).

**Task 2b — terminal-turn fast settle (decouples PM→user from Dev):**
- Read-only classification gated strictly on the current cycle's `pm_saw_idle` (never the latch; Race 2). A non-empty `[/complete]` or a `[/user]` turn settles startup immediately and falls through to the single prime-turn relay for delivery, without waiting for Dev. Empty `[/complete]`/non-terminal chatter does not fast-settle — a genuinely stuck PM still reaches plateau/ceiling.

**Observability:** additive `ContainerResult.startup_settle_reason` (`both_idle` | `pm_latched_dev_idle` | `pm_terminal_fast`), recorded at each settle site and appended to `startup_events`.

The prime-turn relay remains the **sole** delivery site — no in-loop delivery, no double-send.

## Tests

New reproduction tests in `tests/unit/granite_container/test_container.py::TestStartupPmCompleteBeforeDevPrimes` drive the **real** `last_assistant_text` against an on-disk JSONL transcript fixture (no stub), so the `len(texts) <= baseline_text_count` freshness guard — exactly where the reply was dropped — is actually exercised:
- `test_pm_complete_before_dev_primes_delivers` — PM `[/complete]` early, Dev never idles → fast settle delivers, exits `pm_complete`, `startup_settle_reason == "pm_terminal_fast"` (SC#1b).
- `test_pm_latched_dev_idle_delivers_complete` — PM idle early, Dev idles later → latch settles, relay delivers, `startup_settle_reason == "pm_latched_dev_idle"` (SC#1a).
- `test_empty_complete_at_startup_not_user_facing` — empty `[/complete]` does not fast-settle-deliver; `user_facing_routed` False, `on_complete_payload` not called (concern note #3).

**Red-state verified:** against the pre-fix container these repro scenarios exit `startup_unresolved` / `ceiling` with nothing delivered.

`pytest tests/unit/granite_container/test_container.py -n0` → **47 passed**. Existing stubs were updated to idempotent content-mirroring reads to reflect the relay's flush-safe re-read.

## Notes

- `tests/unit/test_granite_startup_diagnostic.py` has 7 pre-existing failures (`AttributeError: Container does not have '_run_pkill_fallback'`) present identically on the branch base — unrelated to this change.

### DIFF
--- a/agent/granite_container/container.py
+++ b/agent/granite_container/container.py
@@ -493,6 +493,15 @@ class ContainerResult:
     # startup_plateau_cycles: number of consecutive identical fingerprint cycles
     # that triggered the plateau bail (None for ceiling exits).
     startup_plateau_cycles: int | None = None
+    # startup_settle_reason: which startup settle path fired (issue #1881).
+    #   "both_idle"           — classic same-cycle PM+Dev idle.
+    #   "pm_latched_dev_idle" — PM went idle on an earlier cycle (latched via
+    #                           pm_ever_idle); Dev reached idle later.
+    #   "pm_terminal_fast"    — PM produced a terminal [/complete]/[/user] turn
+    #                           and startup settled immediately, without waiting
+    #                           for Dev (decouples PM→user delivery from Dev).
+    # None on a startup that never settles (plateau / ceiling).
+    startup_settle_reason: str | None = None
 
 
 # ---------------------------------------------------------------------------
@@ -1809,6 +1818,25 @@ class Container:
             # the prime text, not in the omission of the message).
             logger.info("container: priming PM")
             _pm_prime_cmd = _resolve_pm_prime_cmd(self._session_type)
+            # Relay baseline snapshot (issue #1881): snapshot the count of
+            # text-bearing PM assistant entries BEFORE PM emits any
+            # task-response output — i.e. before _prime_session(pm) below.
+            # This MUST precede PM's prime turn: the prime-turn relay's
+            # freshness guard `last_assistant_text(..., baseline_text_count=
+            # pm_prime_baseline)` returns "" when len(texts) <= baseline. If
+            # the baseline were snapshotted after the startup loop (its old
+            # site) it would already count PM's [/complete] in the fast-PM/
+            # slow-Dev race — PM's terminal turn flushes cycles before the
+            # loop breaks — so the relay would drop the reply and exit a
+            # compliance nudge instead of pm_complete. Captured here, PM's
+            # terminal turn is always a NEW text-bearing entry, so the relay
+            # delivers it for a both-idle, latch, or fast-settle break alike.
+            # result.pm_transcript_path was populated by _capture_pty_identity
+            # above; tolerate None → 0 (matches the prior post-loop behavior).
+            _pm_relay_transcript = result.pm_transcript_path
+            pm_prime_baseline = (
+                text_bearing_count(_pm_relay_transcript) if _pm_relay_transcript else 0
+            )
             self._prime_session(self._pm_pty, _pm_prime_cmd, include_user_message=True)
             logger.info("container: PM prime done")
             # A headless Dev role has no PTY to prime here — HeadlessRoleDriver
@@ -1838,6 +1866,15 @@ class Container:
             startup_settled = False
             startup_deadline = time.monotonic() + STARTUP_HARD_CEILING_S
             cycle = 0
+            # PM-idle latch (issue #1881). Idle detection is edge-triggered per
+            # cycle, so in the fast-PM/slow-Dev race PM's idle is observed in an
+            # early cycle and Dev's idle in a later one, never together — the old
+            # same-cycle `pm_saw_idle and dev_saw_idle` gate then never fires and
+            # the run burns to the ceiling. Latching PM idle once observed lets a
+            # later Dev-idle cycle settle startup. (Headless Dev already forces
+            # dev_saw_idle=True, so once PM is idle the gate reduces to PM alone,
+            # byte-identical to before.)
+            pm_ever_idle = False
             # Plateau detection state.
             # Fingerprint is the parser's verdict (response) ALONE. The idle bools
             # are NOT included in the fingerprint key because they can flicker on
@@ -1863,6 +1900,11 @@ class Container:
                     dev_idle = self._startup_cycle_idle(self._dev_pty)
                 pm_saw_idle, pm_edge, pm_level, pm_marker, pm_ms = pm_idle
                 dev_saw_idle, dev_edge, dev_level, dev_marker, dev_ms = dev_idle
+
+                # Latch PM idle (issue #1881): once PM is observed idle in any
+                # cycle, remember it so the settle gate no longer requires PM
+                # and Dev to be idle in the SAME cycle.
+                pm_ever_idle = pm_ever_idle or pm_saw_idle
 
                 # Update level tails for frame capture at any exit point.
                 _last_pm_level_tail = pm_level.strip() or pm_edge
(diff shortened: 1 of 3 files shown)


tomcounsell:
## Review: APPROVED

Reviewed against plan `docs/plans/granite-startup-pm-complete-before-dev-primes.md` and issue #1881. This fixes the silent-drop of PM's `[/complete]` reply when the PM finishes before the Dev PTY primes.

### Preflight
- State OPEN, `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN`.
- No UI files in the diff — screenshot gate is a no-op.

### What the change does (verified against `agent/granite_container/container.py`)
- **Relay-baseline relocation (the blocker fix):** `pm_prime_baseline` is now snapshotted at line 1837, *before* `_prime_session(pm)` (1840), reading `result.pm_transcript_path` which `_capture_pty_identity` populated at 1807. Ordering is correct, and there is exactly **one** `pm_prime_baseline =` assignment (old post-loop site deleted). With the baseline predating PM's task-response output, PM's terminal turn is always a new text-bearing entry, so the relay's `last_assistant_text(..., baseline_text_count=...)` freshness guard delivers it for any break reason.
- **PM-idle latch (`pm_ever_idle`, line 1907):** the settle gate now reads `pm_ever_idle and dev_saw_idle` (2068) instead of same-cycle `pm_saw_idle and dev_saw_idle`. Headless-Dev path forces `dev_saw_idle=True`, so once PM is idle the gate reduces to PM alone — byte-identical to the prior behavior.
- **Terminal-turn fast settle (2031-2058):** gated strictly on the current cycle's `pm_saw_idle` (never the latch — Race 2), read-only classification, only decides to break. A non-empty `[/complete]` or a `[/user]` settles immediately; empty `[/complete]`/chatter falls through so a genuinely stuck PM still reaches plateau/ceiling.
- The prime-turn relay remains the **sole** delivery site — confirmed zero `on_complete_payload` calls inside the startup loop.
- Additive `ContainerResult.startup_settle_reason` (`both_idle` | `pm_latched_dev_idle` | `pm_terminal_fast`) for observability.

### Tests (drive the real reader — no stub)
- `tests/unit/granite_container/test_container.py` → **47 passed** (`-n0`). New `TestStartupPmCompleteBeforeDevPrimes` (3 tests) writes an on-disk JSONL fixture and exercises the genuine `last_assistant_text(baseline_text_count=...)` guard, which is exactly where the reply was being dropped. Covers SC#1a (latch delivery), SC#1b (fast-settle delivery, Dev never primes), and empty-`[/complete]` non-delivery.
- The 7 failures in `tests/unit/test_granite_startup_diagnostic.py` (`AttributeError: ... does not have '_run_pkill_fallback'`) are **pre-existing on main** — that test file references `_run_pkill_fallback` (5x) but `container.py` never defined it, on both main and this branch. This PR touches neither. Not a regression.

### Plan Verification table
Every machine-readable check passes on the branch: repro test green, single delivery site (count 0), exactly one baseline snapshot, baseline precedes the loop, latch present, settle-reason present, `ruff format --check` clean. `ruff check` on the changed files also passes.

### Gates
- Documentation: plan specified updating `docs/features/granite-pty-production.md` — done (+40 lines, "Startup settle conditions" section). Gate satisfied.
- Plan section compliance: all four required sections present (Documentation, Update System, Agent Integration, Test Impact).
- Ruff + tests: green.
- **Bridge/Worker change:** this modifies `agent/granite_container/container.py`, reached via `BridgeAdapter → Container.run()`. Flag for service restart on all machines after deploy.

**Blockers:** None
**Tech Debt:** None
**Nits:** None

Issue #1927, opened 2026-07-06
AgentSession schema diet: prune accreted telemetry fields, rename survivors for precision


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/agent-session-schema-diet.md

## Problem

`models/agent_session.py` accreted dozens of telemetry fields during the granite PTY era: metered/total accounting splits annotated "PTY roles only", wedge-recovery counters by tier, cold-start metrics pointers, TUI-capture fields, delivery-tracking workaround flags (`has_communicated`), plus fields whose names no longer say what they mean. The teardown (#1924) removes the PTY fields; the remaining schema is still roughly 2x wider than its post-teardown meaning.

## Owner direction (2026-07-06)

"100% agree. even naming of fields could be improved for clarity, precision."

## Scope

- Field-by-field audit of `AgentSession`: keep / delete / rename (precision names per the clear-naming rule)
- Collapse the metered/total accounting split now that all roles share one transport
- Update mirrors: `ui/data/sdlc.py`, `ui/app.py` dashboard payload, `valor-session` CLI output, session archive (`data/session_archive.db`) export mapping
- Idempotent ORM-safe migration in `scripts/update/migrations.py` (no raw Redis); update the partitioned-save liveness allowlist
- Accept loss of longitudinal comparability on pruned metrics (deliberate)

## Recon Summary

**Confirmed** (2026-07-06 audit, re-verified 2026-07-13 against current main):

- Metered/total accounting split lives at `models/agent_session.py:481-519` — `total_*` scalars at 488-494, `metered_*` scalars at 516-519. Collapsible now that all roles share one `claude -p` transport.
- The delivery-tracking workaround `has_communicated` is a `messenger` method, not a field; the runner path added the `user_facing_routed` field (`models/agent_session.py:337-345`) to compensate because the runner never calls `messenger.send()`. That field is the audit target the issue names via `:286-296`.
- The partitioned-save allowlist the issue calls `ADD_ONLY_LIVENESS_FIELDS` is `_UPDATED_AT_OMISSION_OK_FIELDS` at `models/agent_session.py:969-985`.
- Dashboard mirrors are `ui/data/sdlc.py` and the `_session_to_json()` payload in `ui/app.py`, plus the SQLite session-archive writer.
[...]

Issue #1925, opened 2026-07-06
Remove claude_code_sdk: claude -p harness for all session work, PydanticAI for non-harness LLM calls


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/pydantic-ai-nonharness-llm-standardization.md

## Problem

After the granite PTY teardown (#1924), the system still carries **two Claude transports** in `agent/sdk_client.py` (3,838 LOC): the in-process `ClaudeSDKClient`/`claude_code_sdk` path (used by `ValorAgent` for the bridge conversational chat path) and the `claude -p` stream-json subprocess harness (`get_response_via_harness` — used by the session runner roles and the message drafter). Duplicate auth handling, duplicate hooks config, duplicate resume plumbing. Separately, non-harness LLM calls are hardwired to specific providers.

## Owner direction (2026-07-06)

Exactly two ways to call an LLM, by kind of work:

1. **`claude -p` harness** — anything that IS a Claude Code session (skills, hooks, tools, resume).
2. **PydanticAI** — all non-harness LLM actions (classification, extraction, refinement, judging), so they are **model-agnostic**.

The `claude_code_sdk` dependency is removed entirely.

## Scope

- Migrate the `ValorAgent` conversational path to the harness (turn-level delivery replaces token-level streaming — acceptable; the bridge batches for Telegram anyway)
- Delete `ClaudeSDKClient`, `claude_code_sdk` from dependencies, and the duplicated hooks/env config
- Inventory non-harness LLM call sites; introduce PydanticAI as the standard wrapper
- Synergy: #1923 (replace bridge/email ollama classifier) should be implemented AS a PydanticAI call — model-agnostic solves it by construction

## Recon Summary

**Confirmed (2026-07-06 audit, baseline d451c1bd):**
- Both transports and their callers mapped: `sdk_client.py:1512` (ClaudeSDKClient/ValorAgent, caller `:3788`), `:2340` (`get_response_via_harness`; callers: session runner roles, `session_completion.py:755,817` drafter)
- Auth is identical in both (strip `ANTHROPIC_API_KEY`, subscription OAuth: `:1605-1607`, `:2469`)
- Harness supports hooks (`--settings`, `:2436`), resume (`:2488-2490` + stale-UUID fallback `:2585-2622`), model/env/cwd injection — feature parity for the chat path is plausible, needs a migration spike for streaming UX

[...]

Issue #1923, opened 2026-07-06
Drop ollama entirely: replace bridge routing + email triage classifier calls with a small Claude call


valorengels:
## Problem

After the granite PTY teardown removes ollama from the session-execution path (worker probe, circuit breaker, reprobe loop, degraded-mode deferral, update green-light gate), ollama remains on this machine solely for two direct consumers:

- `bridge/routing.py` — message routing classification (imports `OLLAMA_CLASSIFIER_MODEL` from `config/models.py`, calls ollama directly)
- `tools/email_cs/triage.py` — email triage classification

## Desired outcome

Replace both call sites with a small Claude call (haiku) so the ollama dependency, its model pulls, and the nightly canary concern (#1854) can be removed machine-wide.

## Context

Deliberately scoped OUT of the PTY-teardown plan (tradeoff decision 2026-07-06: "out of worker path" chosen over "drop ollama everywhere") to keep that cutover focused. This issue is the follow-up that finishes the ollama-ectomy.

## Recon pointers

- `bridge/routing.py` ollama call path and `config/models.py::OLLAMA_CLASSIFIER_MODEL`
- `tools/email_cs/triage.py` classifier usage
- Latency/cost consideration: routing runs per inbound message; haiku call adds network latency where ollama was local — measure before/after.

valorengels:
Folded into #1925 (backlog groom 2026-07-10).

Both ollama consumers named here — `bridge/routing.py` message classification and `tools/email_cs/triage.py` — are exactly non-harness LLM call sites under #1925's two-transport rule, and #1925 already names the synergy ("#1923 should be implemented AS a PydanticAI call"). Running this as a separate pipeline would mean two migrations touching the same call sites.

#1925 now owns: introduce PydanticAI, port these two classifier call sites, then remove ollama machine-wide (model pulls + the #1854 nightly-canary concern). Carried-over constraint: routing runs per inbound message — measure local-ollama vs haiku latency before/after the swap.

Issue #1633, opened 2026-06-11
Merge PM/Dev bridge roles into a single Eng role; collapse SessionType to {eng, teammate}


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/merge_pm_dev_into_eng_role.md

> **Context:** This system routes Telegram messages through a bridge that creates [AgentSessions](models/agent_session.py) — Redis-backed work records executed by a standalone worker. Today each project has paired Telegram groups (`PM: {Project}` and `Dev: {Project}`) that map to distinct session types: `pm` (orchestrator, read-only) and `dev` (builder, full permissions). PR #1612 (the [granite PTY container](docs/features/granite-pty-production.md) cutover, issue #1572) changes execution so that **every** bridge-originated session runs a container that internally owns *both* a PM-steering and a Dev-builder `claude` TUI, primed via slash commands. The PM/Dev split now lives inside the container, making the bridge-level split redundant — and in the dev case, semantically broken.

## Problem

The bridge derives a session's role from which Telegram group the message arrived in: `bridge/routing.py:resolve_persona()` matches the group name against `projects.json` config (with a `Dev:`/`PM:` title-prefix fallback), and `bridge/telegram_bridge.py` maps the resolved persona to `SessionType.DEV`, `PM`, or `TEAMMATE`.

**Current behavior:**
- 13 projects across 4 machines carry paired (or partial) `PM:`/`Dev:` groups, forcing humans to pick the "right" group before messaging.
- Post-#1612, a message to a `Dev:` group creates a `session_type="dev"` AgentSession that spawns a granite container containing its *own* PM+Dev pair, then falls into `_handle_dev_session_completion()` (`agent/session_completion.py:1454`) with no parent PM to steer — the bridge-level dev type no longer means anything on the execution path.
- The bridge-level persona is largely vestigial for work sessions: the container primes personas itself, and both its PTYs run `claude --permission-mode bypassPermissions`, so the `SESSION_TYPE`-env permission rails (`agent/hooks/pre_tool_use.py`) never reach inside the container.
- Dual groups split conversational history and create dual chat-scoped state for what is one engineering workstream per project.

**Desired outcome:**
- One `Eng: {Project}` Telegram group per project replaces the `PM:`/`Dev:` pair.
- `SessionType` collapses to `{eng, teammate}`: `pm` is renamed to `eng`, `dev` is deleted entirely.
[...]

PR #1930, opened 2026-07-07
Granite PTY teardown: headless claude -p session runner cutover

### LINKED ISSUE #1724: Stalled "never_started" sessions are unrecoverable for up to 30 min: fresh heartbeats blind the recovery actor while the advisory that detects them cannot act

### LINKED ISSUE #1879: Mid-run steering drain to enable a continue-nudge recovery rung for wedged granite sessions (Part B split from #1878)
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/granite-mid-run-steering-drain-continue-nudge.md

## Recon Summary

Grounded against current `main` (HEAD `776de1ee`; issue originally cited `8c8d64b0`). Freshness disposition: **Minor drift** — PR #1880 (#1878 Part A) merged after this issue was filed and touched `session_health.py`, so line numbers drifted, but every root-cause claim still holds and the mid-run-drain infrastructure is still absent.

**Confirmed (still true on current main):**
- The external steering queue (`agent/steering.py`, `pop_all_steering_messages`) is drained for a granite PTY session by exactly one consumer: the `_poll_steering` closure at `agent/granite_container/bridge_adapter.py:620`, wired into `Container.run()`'s steady-state loop at `agent/granite_container/container.py:2098`. It runs only between completed turns.
- `_await_turn_end` (`container.py:1269`) polls only the hook edge file (`consumer.poll()`, `container.py:1304`) plus a byte-level liveness read (`container.py:1327-1331`); it never drains external steering. A `continue` pushed during a wedge is never consumed.
- `last_pty_activity_at` is diff-gated on the **normalized** frame (`_make_pty_read_callback` → `_normalize_pty_buffer`, `bridge_adapter.py:962-981`): spinner/elapsed/cursor repaint is stripped, so a frozen-frame wedge keeps `last_pty_activity_at` stale even while the read loop cycles. This is what lets the progress-deadline killer identify a genuine wedge.
- The live-session no-progress killer is the progress-deadline cancel scope (`agent_session_queue.py:1790`, `reason_kind="progress_deadline"`), which runs above a live `exec_task` but fires only after `SESSION_PROGRESS_DEADLINE_S` (1800s) of stale `_session_progress_ts` (`agent_session_queue.py:1469`).
- The two `no_progress` producers reaching `_apply_recovery_transition` (`session_health.py:2139`) still lack a live consumer: D0 never-started (`session_health.py:4090`) fires mid-prime; the #944 running-scan orphan (`session_health.py:3222`) fires only when `in_scope_handle is None`.

**Revised by Part A (PR #1880 / commit e05d6516):**
- The D0 never-started branch now defers its kill when `_prime_pty_alive` (`session_health.py:545`) shows PTY liveness; `on_read_iteration` is wired into `_prime_session`'s `read_until_idle` calls. This narrows (does not remove) the D0 producer. Part A did NOT add any mid-run steering drain, so Part B's infrastructure is still required.

**New concern surfaced during recon:** injecting `continue` into the PTY writes echoed text that `_normalize_pty_buffer` does NOT strip, so the injection itself would advance `last_pty_activity_at` and falsely reset the wedge detector. The recovery rung must therefore judge "did the nudge work?" on genuine-progress fields (`last_turn_at` / `last_tool_use_at`) or a pre/post normalized-frame snapshot, not on `last_pty_activity_at`. Captured as the central rabbit hole in the plan.

**Split out of #1878.** Part B of #1878 ("add a \`continue\`-nudge rung before kill") proved structurally infeasible on current main without new infrastructure. This issue captures that infrastructure work. Part A of #1878 (prime liveness stamping) ships independently and fully fixes the root-caused priming failure.

## Problem

The desired recovery behavior is: when the session-health monitor suspects a granite session is *wedged but alive*, push a cheap \`continue\` steering nudge into the live session and give it a short reprieve to un-stick itself, escalating to the expensive kill+respawn only if the normalized TUI frame stays frozen.

**This cannot work today: no \`no_progress\` recovery state has a reachable steering consumer.** Grounded in code on commit \`8c8d64b0\`:

- The external steering queue (\`agent/steering.py\`) is drained for a granite PTY session by exactly one consumer: the \`_poll_steering\` closure (\`agent/granite_container/bridge_adapter.py:620\`), wired into \`Container.run()\` at \`agent/granite_container/container.py:2081\`. It is invoked **only at the top of the steady-state turn loop — between completed turns.**
- While a session is wedged it is parked *inside* \`_await_turn_end\` (\`container.py:1302\`), whose liveness pump polls only the **hook edge file** (\`consumer.poll()\`, \`container.py:1328\`), **never** the external steering queue. So a \`continue\` pushed during a wedge is never drained.
- The two \`no_progress\` producers that reach \`_apply_recovery_transition\`'s \`no_progress\` branch both lack a live consumer:
  - **D0 / never-started** (\`session_health.py:4136\`) fires *mid-\`_prime_session\`*, which uses blocking \`read_until_idle\` calls before the steady-state loop (and thus before any \`_poll_steering\`) exists.
  - **#944 running-scan orphan** (\`session_health.py:3229\`) fires **only when \`in_scope_handle is None\`** (\`session_health.py:3222\`) — a crashed-worker \`running\` row with no live \`Container\` / \`exec_task\` in any process. There is no consumer for \`steering:{session_id}\` anywhere.
- The live-session no-progress killer — the progress-deadline cancel scope (\`agent_session_queue.py:1790\`, \`reason_kind="progress_deadline"\`) — *does* run above a live \`exec_task\`, but it fires only after \`SESSION_PROGRESS_DEADLINE_S\` (1800s) of zero \`last_pty_activity_at\`/\`last_turn_at\`/\`last_tool_use_at\` advancement (\`_session_progress_ts\`, \`agent_session_queue.py:1500\`). Reaching \`_poll_steering\` would itself produce PTY activity, so a session stale enough to be killed is, by construction, parked inside \`_await_turn_end\` with the drain unreachable.

**Root:** the drain point (top of a *completed* turn) and the wedge condition (a turn that never completes) are mutually exclusive by construction. A session reaching the drain is making progress and is never a recovery target; a recovery target never reaches the drain.

## What infrastructure is needed

A **mid-run steering drain** that can inject a \`continue\` into a live-but-parked Container without waiting for a turn boundary. Candidate seam: drain-and-inject inside \`_await_turn_end\`'s per-tick liveness pump (\`container.py:1326-1331\`) or via \`read_until_idle\`'s \`on_read_iteration\` callback, guarded so it does not corrupt PM's in-flight turn (the existing top-of-loop path already documents the \`_cycle_idle\`-before-write ordering guarantee at \`container.py:2107\`). This is exactly the "mid-prime steering drain" that #1878 deliberately listed as a No-Go / Rabbit Hole and deferred.

Once such a drain exists, the recovery rung from #1878 becomes implementable and should be scoped to the producer(s) that then have a reachable consumer.

## Acceptance criteria (transferred from #1878)

- [ ] A wedged-but-alive granite session (live \`exec_task\`, normalized frame frozen) can have a \`continue\` steering nudge drained and injected without a completed turn boundary.
- [ ] Recovery attempts exactly one \`continue\` nudge before kill+respawn for a genuine wedge; escalation to respawn only occurs if the frame stays frozen after a bounded reprieve; the nudge does not consume a \`MAX_RECOVERY_ATTEMPTS\`.
- [ ] The mid-run drain does not corrupt PM's in-flight turn (reuses the \`_cycle_idle\`-before-write ordering guarantee).
- [ ] Non-PTY (SDK/headless) sessions are unaffected.

## Prior art / references

- #1878 (Part A: prime liveness stamping — ships separately; this is the deferred Part B)
- #1792 / PR #1798 (the \`_prime_pty_alive\` deferral #1878 Part A completes)
- #1768 (\`_normalize_pty_buffer\` — spinner-vs-content liveness)
- #1779 (mid-run steering injection, the \`_poll_steering\` top-of-turn drain this issue extends)

**Relevant files:** \`agent/granite_container/container.py\` (\`_await_turn_end\`, steady-state \`_poll_steering\` drain), \`agent/granite_container/bridge_adapter.py\` (\`_poll_steering\` closure), \`agent/session_health.py\` (\`_apply_recovery_transition\` no_progress branch), \`agent/agent_session_queue.py\` (progress-deadline cancel scope), \`agent/steering.py\`.

### LINKED ISSUE #1918: granite PTY idle detection blind after claude CLI TUI drift (2.1.201): every prime fails saw_idle=False, sessions die startup_unresolved
> **Context:** Bridge-originated Eng sessions run through the **granite PTY container** ([`docs/features/granite-pty-production.md`](docs/features/granite-pty-production.md)) — two interactive `claude` TUI sessions (PM + Dev) driven over pseudo-terminals. The container decides "the model finished its turn" by **screen-scraping** the TUI: the C5 idle heuristic in [`agent/granite_container/pty_driver.py`](agent/granite_container/pty_driver.py) requires 2s of byte-quiescence plus a `bypass … permissions` bottom bar (`IDLE_BAR`) plus a prompt glyph (`PROMPT_GLYPH`). That heuristic was calibrated against claude CLI **v2.1.173** and re-verified through **v2.1.198** (the D1a version pin in `config/models.py`). The installed CLI has since auto-updated to **v2.1.201**, and the heuristic has gone blind.

## Problem

Since ~2026-07-01 10:27 UTC, the granite prime wait almost never observes idle, and since 2026-07-03 it has **never** observed idle. Sessions burn the full 360s prime budget per role (`saw_idle=False`), then the startup loop sees `pm_idle=False dev_idle=False` on every cycle and bails with `startup_unresolved: startup plateau` after 10 identical no-progress cycles. The session finalizes `failed` and the human gets a 🤔 reaction and no reply.

**Current behavior:**
- Prime post-write wait log history (`logs/worker.log*`): `saw_idle=True` consistently through 07-01 10:07; from 07-01 10:27 onward 15 of 16 primes report `saw_idle=False`; every prime since 07-03 has failed.
- 8 sessions dead with the identical plateau signature across cuttlefish, psyoptimal, and local SDLC runs (`grep granite-exit-anomaly logs/worker*.log*`).
- Example casualty: session `tg_cuttlefish_-1003801797780_365` (2026-07-06). The hook-edge NDJSON (`data/granite_hook_edges/tg_cuttlefish_-1003801797780_365/`) proves the Claude sessions inside the PTYs **worked fine**: PM `Stop` fired at 04:52:35 with a complete assistant reply; Dev `Stop` at 04:55:37. The PM prime wait was still reporting `saw_idle=False` at 04:55:15 — **160+ seconds after the model finished**, on a settled TUI. The container was blind, not the model.
- The D1a drift guard has been shouting: `logs/update.log` contains 78 repetitions of `claude CLI version drift: installed=2.1.201, pinned=2.1.198 … markers have not been re-verified`.
- The D1b golden-sample contract check (`verify_tui_marker_contract()`) passes throughout — by design it only tests the regexes against *stored* samples, so it cannot detect the live TUI changing (that's exactly what the D1a pin + human re-verification pass is for, per [`docs/features/deployment.md`](docs/features/deployment.md)).
- Secondary damage: the failure alert itself was suppressed (`[granite-alert-suppressed] valor-telegram timed out`), so nobody was paged.

**Desired outcome:**
- Granite sessions start and complete turns on claude CLI ≥2.1.201 (prime waits observe turn completion again).
- The startup/prime path no longer has a single point of failure on screen-scraped markers: a turn-completion signal that survives TUI repaints (the Stop-hook edge) is consumed when available.
- The D1a pin is bumped to the re-verified version with the pin-history table in `docs/features/deployment.md` updated.

## Definitions

| granite PTY container | Runner that drives a PM + Dev pair of interactive `claude` TUI sessions over pseudo-terminals for bridge-originated Eng sessions | [`docs/features/granite-pty-production.md`](docs/features/granite-pty-production.md) |
| C5 idle heuristic | Screen-scrape rule declaring a TUI turn finished: 2s byte-quiescence + `IDLE_BAR` (`bypass…permissions`) + `PROMPT_GLYPH` (`>`/`❯`) + content floor + spinner evidence | [`agent/granite_container/pty_driver.py`](agent/granite_container/pty_driver.py) (~lines 90–160) |
| prime wait | `_prime_session`'s post-write `read_until_idle` (360s budget) that waits for the persona-priming slash command's response to settle | [`agent/granite_container/container.py`](agent/granite_container/container.py) `_prime_session` |
| startup plateau | Early-bail when 10 consecutive startup cycles produce an identical no-progress fingerprint (`response=None`, neither PTY idle) → `exit_reason=startup_unresolved` | [`agent/granite_container/container.py`](agent/granite_container/container.py) |
| hook edge | Per-session NDJSON file of Claude Code hook events (`SessionStart`/`Stop`/`Notification`) written by a `--settings`-injected forwarder; consumed via `HookEdgeConsumer` | [`agent/granite_container/hook_edge.py`](agent/granite_container/hook_edge.py), `data/granite_hook_edges/` |
| D1a version pin / D1b contract check | D1a: pinned claude CLI version (`config/models.py`, default 2.1.198) with drift warning at update time. D1b: worker-boot regex-vs-golden-sample self-test (`verify_tui_marker_contract`) | [`docs/features/deployment.md`](docs/features/deployment.md) § claude-CLI contract |

## Prior Context

Related work:
- #1817: Resilience: correctness & delivery-integrity hazards — introduced the D1a version pin and D1b golden-sample contract check; this incident is the exact drift scenario D1a was built to flag (and it flagged it — 78 warnings — but nothing acted on it).
- #1688: Hook-driven turn returns for granite PTY shuttle — built `HookEdgeConsumer` and the Stop-hook forwarder; the steady-state turn loop can consume hook edges, but the **prime/startup** path still relies solely on the C5 screen scrape.
- #1842: Per-role transport hedge — headless `claude -p` per role exists as an opt-in billing hedge (default is both-roles-PTY; PM-headless unsupported in v1). Its `HeadlessRoleDriver` already **prefers a `TURN_END` hook envelope** over scraping — the precedent for the fix direction here.
- #1839 (PR): granite failure-simulation harness — Substrate A/B test rig that should be able to reproduce/regress-test this failure mode.

## Solution Sketch

Two workstreams, ordered:

1. **Recalibrate + re-pin (unblocks production).** Follow the pin-bump procedure in [`docs/features/deployment.md`](docs/features/deployment.md): capture live v2.1.201 TUI frames, determine what actually changed, fix the affected markers and/or timing constants, update the D1b golden samples from the captured frames, bump the pin in `config/models.py`, append to the pin-history table.
2. **Hook edges as primary turn-completion signal in prime/startup (structural immunity).** In the example casualty the Stop-hook edge fired correctly and on time while the screen scrape was blind. Wire `HookEdgeConsumer` into `_prime_session`'s post-write wait and the startup loop's idle checks so a `Stop` event for the current session UUID completes the wait; keep the C5 scrape as fallback for hook-less environments. This mirrors what `HeadlessRoleDriver` already does for headless turns.

**Open questions the plan must resolve (root-cause mechanism is unconfirmed):**
- Which contract actually broke on 2.1.199–2.1.201? Candidates: (a) `IDLE_BAR` bottom-bar text changed, (b) prompt glyph changed, (c) the idle TUI now repaints at ≥1 Hz (blinking cursor / status ticker), permanently defeating the 2s `QUIESCENCE_S` gate. The fix differs materially between (a)/(b) (regex update) and (c) (quiescence redesign — regexes alone can't fix it). Capture real frames before choosing.
- The break started 07-01 10:27 UTC while the pin history says the fleet was on 2.1.197 on 07-02 — establish which CLI version was actually installed at the break point (npm logs / update.log) so the pin-history table stays truthful.
- Session-UUID filtering in `HookEdgeConsumer`: the edge file accumulates events across retry attempts (the 07-06 casualty has two `SessionStart`s per role). The prime-wait consumer must bind to the current attempt's UUID, not the first one in the file.
- Should a `saw_idle=False`-everywhere day trip an alert louder than update-log lines? The D1a warning fired 78 times with zero operational response — consider escalating repeated drift warnings to a Telegram alert (and note the alert path itself timed out here).

## Recon Summary

**Confirmed:** (from the live incident investigation, 2026-07-06 session)
- Failure signature and timeline verified in `logs/worker.log*`: all primes `saw_idle=False` since 07-03; 8 plateau deaths since 07-01.
- Hook edges fired correctly during the failures (verified against `data/granite_hook_edges/tg_cuttlefish_-1003801797780_365/` — PM/Dev `Stop` events with full assistant replies).
- Installed CLI 2.1.201 vs pin 2.1.198; 78 drift warnings in `logs/update.log`; D1b contract check green throughout (golden samples can't catch live drift).
- Transport config: no project sets `transport` in `projects.json` and no `GRANITE__*_TRANSPORT` env override — all sessions run both-roles-PTY (the #1842 headless path is an unused hedge, not a migration).

**Revised:**
- Original framing assumed marker-regex drift; the actual mechanism (regex mismatch vs. quiescence defeat) is unverified — kept as an open question rather than a concrete sketch.

**Pre-requisites:**
- None blocking; live 2.1.201 frame capture is the first plan step, not a separate issue.

**Dropped:**
- Migrating roles to headless `claude -p` as the fix — it's a billing hedge with different economics (metered credits vs flat subscription) and PM-headless is unsupported; out of scope here.

## Acceptance Criteria

- [ ] A granite Eng session started from the bridge on claude CLI ≥2.1.201 primes both roles and completes at least one full PM→Dev turn (no `startup_unresolved`).
- [ ] `_prime_session` and the startup loop treat a `Stop` hook edge for the current session UUID as turn-complete, with the C5 scrape as fallback (unit + failure-sim coverage via the #1839 harness).
- [ ] D1a pin in `config/models.py` bumped to the re-verified version; D1b golden samples updated from captured live frames; pin-history table in `docs/features/deployment.md` appended.
- [ ] The root-cause mechanism (which marker/timing contract broke) is documented in the plan or deployment doc, so the next drift is diagnosable from the writeup.
- [ ] Repeated D1a drift warnings surface somewhere a human actually sees (decision recorded even if the answer is "no change").

## Downstream

This issue will be consumed by `/do-plan` to produce a plan document at `docs/plans/granite-pty-idle-detection-tui-drift.md`, then executed by `/do-build`.

Constraints for the planner: the hook-edge plumbing (`hook_edge.py`, forwarder, `--settings` injection) already exists and is in production for the steady-state loop — reuse it, don't build a parallel signal path. Default transport must remain both-roles-PTY (#1842 billing rationale).

### LINKED ISSUE #1919: Idle notification "Claude is waiting for your input" sent verbatim to Telegram as the session reply, swallowing the PM's real answer
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/idle-notification-verbatim-delivery.md

> **Context:** Bridge-originated sessions run the interactive Claude Code TUI inside a granite PTY container ([`docs/features/granite-pty-production.md`](https://github.com/tomcounsell/ai/blob/main/docs/features/granite-pty-production.md)). Claude Code hooks forward lifecycle events to the container as typed "edges" ([`agent/granite_container/hook_edge.py`](https://github.com/tomcounsell/ai/blob/main/agent/granite_container/hook_edge.py)); a `needs_human` edge means the session is blocked on human input and the container routes a question to the human's Telegram chat. This repo has a hard rule that no raw system strings may ever reach the human on Telegram: every outbound message must read as the engineer persona.

## Problem

The CEO asked a substantive question (evaluate MiniCPM5 as a local-model replacement). The session answered it: the PM transcript ends with a complete, well-reasoned `[/user]` reply. What arrived on Telegram instead was one line: **"Claude is waiting for your input"**, which is Claude Code's own generic idle-notification text. The real answer was never delivered anywhere. To the human this reads as a confusing system burp; the work product was silently discarded.

Evidence, session `tg_valor_-1003449100931_1117` (2026-07-06 06:48 UTC):

- `recent_sent_drafts` contains exactly one draft: the 32-char string "Claude is waiting for your input"; `exit_reason=pm_user`; session then transitioned to `completed`.
- Worker heartbeats show `communicated=False` for the entire 6-minute run: the normal reply-routing path never fired once.
- The PM transcript's last assistant message is a full multi-paragraph `[/user]` answer to the question.

**Current behavior:** two defects, one delivery.

1. **Verbatim boilerplate delivery.** `_needs_human_message` in [`agent/granite_container/container.py`](https://github.com/tomcounsell/ai/blob/main/agent/granite_container/container.py) prefers the hook payload's own `message` field when it is any non-empty string. Claude Code's idle Notification always carries the generic text "Claude is waiting for your input", so it short-circuits the intended fallback chain: `last_assistant_text` from the edge's transcript (which held the real answer), then the persona-safe generic prompt ("This needs your input to continue, please reply."). The classifier side has no content inspection either: `_classify` in `hook_edge.py` maps every `Notification` event to a `needs_human` edge unconditionally.
2. **The real answer was swallowed.** The PM emitted a `[/user]`-routed answer, yet `communicated=False` held through the whole run and the only outbound message was the notification text. The mechanism is not yet confirmed (see open questions); the idle Notification may itself be a symptom of the turn never settling through the normal routing path.

**Desired outcome:** the human receives the PM's actual answer (or actual question) in persona voice. Claude Code boilerplate strings never appear in an outbound chat message, under any fallback path.

## Definitions

| Granite PTY container | Runner that drives the real interactive Claude Code TUI through a pseudo-terminal for bridge sessions | [`docs/features/granite-pty-production.md`](https://github.com/tomcounsell/ai/blob/main/docs/features/granite-pty-production.md) |
| Hook edge | Typed lifecycle signal (turn_end, needs_human, compaction) forwarded from Claude Code hooks to the container | [`agent/granite_container/hook_edge.py`](https://github.com/tomcounsell/ai/blob/main/agent/granite_container/hook_edge.py), [`docs/features/omnigent-hook-edge-reference.md`](https://github.com/tomcounsell/ai/blob/main/docs/features/omnigent-hook-edge-reference.md) |
| needs_human edge | Edge kind meaning the session is blocked on human input; triggers delivery of a question to the human's chat | `hook_edge.py` (`_NEEDS_HUMAN_EVENTS`) |
| Idle Notification | Claude Code's built-in `Notification` hook event fired when the TUI sits idle awaiting input; its message text is the fixed string "Claude is waiting for your input" | [Claude Code hooks docs](https://docs.anthropic.com/en/docs/claude-code/hooks) |
| `[/user]` routing marker | Prefix the PM persona emits to mark output that should be delivered to the human | [`docs/features/granite-pty-production.md`](https://github.com/tomcounsell/ai/blob/main/docs/features/granite-pty-production.md) |
| PM persona rule | Hard requirement that every Telegram message reads as the engineer persona; raw errors and system strings must never reach the CEO | [`docs/features/teammate-session-permissions.md`](https://github.com/tomcounsell/ai/blob/main/docs/features/teammate-session-permissions.md), repo convention |

## Prior Context

Related work:
- #1688: Built the hook-driven needs-input routing this bug lives in (Notification / PermissionRequest as needs-human edges).
- #1877: Session lifecycle gaps; introduced persona-aware interrupt copy for a neighboring path (the "I was interrupted and will resume automatically" message).
- #1744: Granite messageless-session silent success; same family of "session ends without delivering its work".
- #1546: The granite PTY PoC that established the TUI-driving architecture.

## Solution Sketch

Confirmed direction plus open questions for the planner:

- **Filter boilerplate at the extraction point.** In `_needs_human_message`, treat known Claude Code notification boilerplate (the idle string, permission-prompt phrasings like "Claude needs your permission to use ...") as non-informative: skip the payload message and fall through to `last_assistant_text`, then the persona-safe generic prompt. Match conservatively (exact strings / stable prefixes), since a legitimate question could theoretically arrive in a payload message.
- **Reconsider classification.** Should an idle Notification classify as a `needs_human` edge at all, or is it a liveness signal only? A permission request is a genuine needs-human; an idle notification often just means "turn finished". Decide in the plan; `_classify` currently cannot tell them apart without reading the message text.
- **Open question (primary damage):** why did the `[/user]` answer never route? `communicated=False` for the entire run means the normal output path never fired even though the transcript holds a routed answer. If the idle Notification is a symptom of the turn never settling, fixing message extraction alone still loses the answer; the plan should reproduce this flow before choosing the fix.
- **Do not sanitize downstream.** Scrubbing the string in the message drafter would hide the symptom while still discarding the real answer.

## Recon Summary

**Confirmed:** 4 items
- `container.py::_needs_human_message` returns `payload["message"]` verbatim for any non-empty string, ahead of the transcript fallback.
- `hook_edge.py::_classify` maps every `Notification` event to `needs_human` with no content inspection.
- Evidence session `tg_valor_-1003449100931_1117`: single sent draft is exactly the idle string; `exit_reason=pm_user`; heartbeats `communicated=False` from 60s through 360s.
- The PM transcript's final assistant message is a complete `[/user]` answer to the CEO's question; it was never delivered on any channel.

**Revised:** 1 item
- Original framing was "persona leak". Recon showed the leak is the visible half; the swallowed answer is the primary damage and needs its own root-cause step in the plan.

**Pre-requisites:** none.

**Dropped:** 1 item
- Sanitizing outbound text in the message drafter layer: masks the defect, the answer is still lost, and it invites a growing blocklist in the wrong layer.

## Acceptance Criteria

- [ ] A `needs_human` edge whose payload message is Claude Code boilerplate (idle notification, permission phrasing) never produces that string in an outbound chat message; unit test over `_needs_human_message` with a real idle-Notification payload asserts fallback to `last_assistant_text`.
- [ ] For the evidence scenario (PM produced a `[/user]` answer, idle Notification fired), the delivered Telegram message is the PM's answer text; covered by an integration test over the container edge flow.
- [ ] The swallowed-answer mechanism is root-caused and either fixed or split into a follow-up issue with findings recorded; the plan documents which.
- [ ] Classification decision (idle Notification as needs_human vs liveness-only) is made and documented in [`docs/features/omnigent-hook-edge-reference.md`](https://github.com/tomcounsell/ai/blob/main/docs/features/omnigent-hook-edge-reference.md) or the granite PTY doc; no historical narration, describe only the resulting behavior.

## Downstream

This issue will be consumed by `/do-plan` to produce a plan document at `docs/plans/{slug}.md` (kebab-case slug derived from the title), then executed by `/do-build`. Constraint for the planner: boilerplate matching must be conservative and centralized (one definition of "known boilerplate"), and any new fallback text must satisfy the PM persona rule.

### LINKED ISSUE #1921: Simplify per-role transport: default Dev to headless, add PTY auto-fallback on spawn failure
> **Context:** The [granite container](docs/features/granite-pty-production.md) drives every bridge-originated session through two Claude Code roles, PM and Dev. Issue #1842 / PR #1848 added a [per-role transport hedge](docs/features/per-role-transport.md): each role can run over an interactive **PTY** (`claude` TUI in a pseudo-terminal) or **headless** (`claude -p`, one subprocess per turn). The hedge exists because Anthropic's billing policy for programmatic (`-p`/Agent SDK) use has moved once and could move again — PTY and headless bill the same today, but if that changes, flipping a role's transport needs to be a fast, low-risk operation, not a code change.

## Problem

**Current behavior:** `transport.dev` is a per-project key in `projects.json` (`{"transport": {"pm": "pty", "dev": "headless"}}`), with precedence project-block → `GRANITE__DEV_TRANSPORT` global setting → hardcoded `"pty"` (`agent/session_executor.py::_resolve_role_transports`). We just confirmed Dev-headless is correct everywhere and set `transport.dev: "headless"` by hand on all 14 projects across all 4 machines in the iCloud-synced `projects.json` — the only way to change a fleet-wide default today is 14 repeated per-project JSON edits.

**Desired outcome:** Setting the fleet-wide Dev transport default should be one change, not fourteen. Additionally, a headless spawn failure (missing `claude` binary, credential error, subprocess crash) should degrade the session to PTY automatically rather than hard-failing it — today `agent/granite_container/role_driver.py::HeadlessRoleDriver` has no such fallback.

## Definitions

| PTY transport | Role runs the real interactive `claude` TUI over a pseudo-terminal, flat-billed on the Claude subscription | `agent/granite_container/pty_pool.py`, `docs/features/granite-pty-production.md` |
| Headless transport | Role runs one `claude -p` subprocess per turn via the preserved harness, metered against the Agent SDK credit pool if/when that billing policy takes effect | `agent/granite_container/role_driver.py::HeadlessRoleDriver` |
| Transport hedge | The config-selectable PTY/headless mechanism built in issue #1842 specifically so a transport flip in response to a billing change is a config edit + restart, not a code deploy | `docs/features/per-role-transport.md` |

## Prior Context

Related work:
- #1842: Per-role transport hedge — config-selectable PTY vs headless per PM/Dev role, with metered-leg cost visibility. Shipped in PR #1848. This is the feature this issue proposes simplifying.
- #1721 (open): Granite session resume-handle persistence — touches the same `resume_handles` schema (`{role, claude_session_id, transcript_path, transport}`) that a transport-resolution change must not break.

## Solution Sketch

This is a simplification of an intentional hedge, not a straightforward cleanup — the plan should resolve the open question below before touching code, since the answer changes the shape of the change substantially.

**Open question: does the per-project JSON override still earn its keep?**

The per-project knob was built specifically so a fleet-wide transport flip (triggered by an Anthropic billing policy change) is a config edit + `worker-restart`, not a code change + deploy (see `docs/features/per-role-transport.md` "Runbook: flipping a transport"). Two candidate directions:

- **A — Hardcode the default, drop the config path entirely.** Flip `settings.granite.dev_transport` default from `"pty"` to `"headless"` in `config/settings.py`, delete `validate_transport()` from `bridge/config_validation.py`, delete the project-block precedence tier from `_resolve_role_transports()`, and strip the now-redundant `transport.dev` entries from all 14 projects in `projects.json`. Keep the `GRANITE__DEV_TRANSPORT` env var as the sole override (machine-global, not per-project). Loses the ability to pin one specific project's Dev role to PTY without an env var or code change — acceptable only if no client/project ever needs an exception.
- **B — Keep the global-default flip cheap, drop only the per-project repetition.** Same settings.py default flip, but keep `validate_transport()` and the project-block precedence tier alive as a rarely-used per-project escape hatch (documented as "override only," not the primary control surface). This preserves the original hedge's fast-flip guarantee for the fleet-wide case while eliminating the 14x-JSON-edit problem for the common case.

The plan should pick one (or propose a third option) and justify it against the original hedge rationale before implementing.

**Separable, not in question — add automatically regardless of the direction chosen:**

- Runtime fallback to PTY: wrap the `HeadlessRoleDriver` spawn (constructed in `agent/granite_container/container.py` / `bridge_adapter.py`) in a try/except. On spawn failure (missing `claude` binary, credential error, subprocess crash), log a warning and re-spawn that role as `PTYDriver` for the session instead of hard-failing it (today's behavior per `docs/features/per-role-transport.md` "Failure classification": a hung subprocess is bounded-wait killed, a nonzero exit propagates `exit_reason` — neither degrades to PTY).
- If direction A is chosen: update `worker/__main__.py`'s fleet-wide golden-sample PTY-transport scan (~line 662-699), which currently gates hard_fail/warn severity by scanning every project's `transport.pm`/`transport.dev` for `"pty"`. With the project-level Dev knob gone, this should key off `settings.granite.pm_transport` alone (PM is always PTY today, so the gate stays load-bearing).

## Acceptance Criteria

- [ ] Plan resolves the direction-A-vs-B open question with an explicit rationale, informed by `docs/features/per-role-transport.md`'s stated hedge purpose
- [ ] `settings.granite.dev_transport` defaults to `"headless"` (no per-project config required to get today's already-rolled-out fleet behavior)
- [ ] A headless role spawn failure degrades that session to PTY with a logged warning, instead of failing the session
- [ ] Existing per-role-transport unit tests (`tests/unit/test_transport_config_validation.py`, `tests/unit/test_transport_routing_matrix.py`) updated to match the chosen direction, not deleted wholesale
- [ ] `docs/features/per-role-transport.md` updated to reflect the new default and (if direction A) the removed per-project knob, or (if direction B) the "override only" framing

## Downstream

This issue feeds into `/do-plan`, which will produce a plan document at `docs/plans/{slug}.md` per this repo's convention, then executed by `/do-build`. The plan must explicitly address the direction-A-vs-B open question before any implementation task is written — do not let `/do-plan` default silently to full deletion (direction A) without weighing it against the hedge's original fast-flip rationale.

### LINKED ISSUE #1924: Granite PTY teardown: delete the PTY substrate and run all sessions headless via claude -p
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/granite-pty-teardown.md

**Type:** chore | **Appetite:** Large | **Status:** Planning

## Problem

The granite PTY substrate (agent/granite_container/, 9,392 LOC) drives two interactive Claude Code TUIs and reconstructs session state by screen-scraping. It failed both watched SDLC retries on 2026-07-06 (#1915 never_started, #1916 startup_unresolved auto-marked completed) and put the worker in `granite degraded`, a state that fails every eng session. Root cause per the committed postmortem (`docs/postmortems/2026-07-06-granite-pty-fragility.md`, d451c1bd): the TUI is a human product with no stability contract; idle/prime/wedge detection heuristics cannot be tuned into reliability.

## Verdict (one-way, no revert path)

Delete the PTY substrate entirely and run PM, Dev, teammate, and eng session execution through headless `claude -p` stream-json on subscription OAuth, with per-role `claude_session_id` persistence for `--resume`. No transport abstraction, no PTY fallback, no parallel-run migration.

## Locked tradeoff decisions (owner, 2026-07-06)

1. **Topology:** PM stays a real session; PM↔Dev handoff becomes executor-mediated message passing (two headless roles).
2. **ollama:** removed from the session-execution path (worker probe, breaker, reprobe, degraded deferral, update gate all deleted). Bridge routing + email triage keep direct ollama calls → follow-up #1923.
3. **Resume:** simple `--resume` with persisted per-role session IDs; supersedes the #1721 lossless-checkpoint plan.
4. **Steering:** auto-preempt — any steering message kills the in-flight turn subprocess and resumes with the steer injected.

## Recon Summary

Six-agent audit on 2026-07-06 (baseline d451c1bd), full evidence in the plan document:

- **Confirmed:** ~2,600 LOC graduates (hook_edge, hook_forwarder, transcript_tailer, HeadlessRoleDriver, relay/exit-classification, delivery/resume persistence); the rest of the package deletes. Dev leg already runs headless in production; message drafter proves `claude -p` + OAuth daily.
- **Confirmed gaps to close:** PM headless dispatcher does not exist (executor coerces pm=headless→pty at session_executor.py:1690-1696); headless Dev relays through the PM PTY (container.py:2787); resume handles are written but never consumed.
- **Traps:** "granite" is two subsystems — the ollama classifier machinery merely lives in the doomed package; bridge/worker *loop-wedge* detectors are out of scope and must survive.
- **Test surface:** ~648 tests / 38 files delete, 15 files reshape, ~120+ headless tests already exist; one genuine gap (frozen-frame wedge detection has no headless analog — replaced by turn timeout + subprocess liveness).
- **Headless CLI verification:** PermissionRequest hooks do not fire under -p (moot under bypassPermissions); `--resume` is cwd-scoped (resume must re-invoke from the same working_dir); `--bare` must not be used (ignores CLAUDE_CODE_OAUTH_TOKEN).

## Subsumes / supersedes

- Closes #1918 (scraped-marker drift — structurally impossible after deletion)
- Supersedes #1921 (PTY auto-fallback contradicts the no-revert mandate)
- Supersedes #1721 (plan `granite_lossless_checkpoint_resume.md` → simple-resume decision)
- Reverts the TOOL_TIMEOUT_DEFAULT_SEC 300→3000 stopgap (4f9f929e)
- Unblocks re-run of #1915 / #1916 on the new substrate

---
This issue is for tracking and discussion. The plan document is the source of truth.

valorengels:
One-way cutover of all session-role execution from the PTY/interactive-TUI substrate to headless `claude -p` via the new `agent/session_runner/` package. Executes `docs/plans/granite-pty-teardown.md` (plan status Ready, three critique passes embedded, four locked tradeoff decisions built as written).

## What shipped

- **`agent/session_runner/`** (task 1-2): graduated survivors (`role_driver`, `hook_edge` with the #1919 Notification-classification fix, `hook_forwarder`, `transcript_tailer`, `router`, `adapter`) plus the new `SessionRunner` single-session turn loop for ALL session types. One `claude -p --output-format stream-json` subprocess per turn; simplified route table (`[/user]` deliver, `[/complete]` wrapup, else bounded nudge); role-aware env-overridable turn timeouts.
- **D1-AMENDED topology**: single Opus PM session per AgentSession; Dev is a resumable `dev` subagent (`.claude/agents/dev.md`, continuation contract baked in) spawned via the Agent tool on first need and continued across turns and worker restarts.
- **D4 steer-preempt** (task 2): steering auto-preempts the running turn — 3s debounce, generation-token guard, SIGTERM to 10s grace to SIGKILL on the process group, `turn_end_source="preempted"`; the steer becomes the resumed turn's first message. Subprocesses run in their own process group with PID/PGID recorded pre-await (Race 2).
- **D3 four-scalar resume** (task 3): `claude_session_uuid` (existing field), `dev_agent_id`, `runner_cwd`, `claude_version` persisted at stream-json init (Race 5 capture-at-init); validated and consumed on resume (`--resume`, skip prime, same-Dev reintroduction); any invalid scalar falls back to cold start. `resume_handles` deleted. Turn-history mirror on the session-event stream for observability; transcripts stay the source of truth.
- **D2 ollama out of the worker path** (task 4): `ensure_granite_model` probe/breaker/reprobe, TUI marker contract, PTY pool, transport seam (`_resolve_role_transports`, `role_transports`, `validate_transport`) all deleted. `child_session_gate` retained with rewritten rationale (critique C4).
- **Health/telemetry teardown** (task 5): wedge-nudge channel, PTY quiescence gates, `granite_wedged` class, plateau producer mapping deleted; `TOOL_TIMEOUT_DEFAULT_SEC` reverted to 300; eight PTY fields stripped from `AgentSession` with an idempotent ORM-only migration (`strip_pty_session_fields`).
- **Substrate deletion** (task 6): `agent/granite_container/` (14 modules), `tools/granite_loop/`, six PTY probe/spike/smoke scripts, and sole-subject legacy test trees — 90 files, −32,346 lines. Repo-wide sweep: zero `granite_container`/`pexpect`/wedge-nudge/transport-seam references in production source; loop-wedge family preserved.
- **Config/update system** (task 7): `GraniteSettings` → `SessionRunnerSettings` (`SESSION_RUNNER__` prefix) with a loud stale-`GRANITE_*` env-key guard surfaced at deploy; `pexpect`/`ptyprocess`/`valor-granite-loop`/`granite_integration` removed from pyproject; `.claude/commands/granite/` → `.claude/commands/roles/`; nightly regression rewritten to unit+TTFT.
- **Test reshape** (task 8): all 14 Test Impact checkboxes executed; `headless_hook_probe.py` salvaged against live seams; replacement E2E probes smoke-verified against the real `claude` binary.
- **Dashboard visibility** (post-validation patch): resume scalars serialized in `dashboard.json`; turn-history mirror events dual-keyed to the stream's `event_type` convention and labeled `[pm]`/`[dev]`.

## Validation

- Task 10 live smoke (branch worker, real sessions): PM prime → `dev` subagent → delivery with `ps` proving zero PTY children; mid-turn steer → preempt (`SIGTERM`, `turn_end_source="preempted"`) → `--resume` with steer injected; **worker-restart continuation** — genuine `SIGKILL` mid-turn, post-restart resume with the identical `claude_session_uuid` and identical `dev_agent_id`, Dev recalling pre-restart state, turn-history mirror unbroken.
- Task 12 final validation: every Verification-table row (including inverse/anti-criteria) PASS fresh; 94/94 session_runner unit tests; full-suite branch-vs-main failure diff empty (environmental failures documented); ruff check + format clean; `git merge-tree` clean against main.
- Docs landed on main (`921fba7f`): new `docs/features/headless-session-runner.md`, eight dead feature docs removed, architecture docs rewritten to the headless status quo.

## Deploy sequencing

Deploy is post-merge via `/do-deploy` per the plan's No-Gos [ORDERED] entry: the `transport` keys in `~/Desktop/Valor/projects.json` are removed at deploy time, not in this PR, so pre-cutover machines aren't stranded mid-rollout.

Closes #1924
Closes #1918
Closes #1919
Closes #1921
Closes #1724
Closes #1879

Supersedes #1721 (commented there; closure via this PR's plan supersession).


valorengels:
## Review: Changes Requested — Nit

One-way PTY→headless cutover, reviewed against `docs/plans/granite-pty-teardown.md` (#1924) and its four locked decisions. This is an exceptionally rigorous teardown: −32k LOC of substrate deleted cleanly, the new `agent/session_runner/` package is well-structured and fail-silent throughout, every hard gate is green, and all live success criteria I could exercise pass. A single low-value nit keeps this off a clean approval; there are no blockers and no tech debt.

**Verified end-to-end on the PR branch (worktree `d27e154`):**
- Mergeability preflight: `state=OPEN`, `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN`.
- Verification table: 15/16 rows pass mechanically (see the one nit below). Substrate dir gone; zero `granite_container`/`pexpect`/`GRANITE__`/`role_transports`/`wedge_nudge` refs in prod; `TOOL_TIMEOUT_DEFAULT_SEC` stopgap reverted; runner + `dev.md` present; loop-wedge family preserved; #1923 ollama classifier untouched.
- `ruff check .` clean; `ruff format --check .` clean (1088 files).
- `tests/unit/session_runner/` — **112 passed** serially (exceeds the claimed 94; relocated suites included). Deleted trees (`tests/unit/granite_container/`, `tests/granite_faults/`) gone; `granite_integration` marker removed.
- Model field removal is clean: eight PTY fields + `resume_handles` stripped, four resume scalars added nullable, `pm_pid` repurposed and added to the `_UPDATED_AT_OMISSION_OK_FIELDS` allowlist to match the per-turn partial save. No dangling refs to removed fields in prod (the one grep hit is an explanatory docstring in `adapter.py`).
- ORM-only migration `strip_pty_session_fields` registered in `MIGRATIONS`, idempotent, 300s-bounded subprocess.
- Settings rename `GraniteSettings→SessionRunnerSettings` (`SESSION_RUNNER__` prefix) with the mandatory loud stale-`GRANITE_*` env-key guard at load (`config/settings.py:697-728`).
- #1919 fix present: `_NEEDS_HUMAN_EVENTS = frozenset({"PermissionRequest"})`, boilerplate Notification text content-classified to emit no edge.
- Docs landed on `main` separately per the docs-on-main convention: `docs/features/headless-session-runner.md` exists on main, the five dead granite feature docs are removed on main. The branch carries the pre-deletion copies but merges cleanly (main's deletions win — no resurrection).
- UI: `session_modal_content.html` drops the Granite PTY block; resume scalars serialized in `dashboard.json` (`ui/app.py:741-743`); readers use `getattr(..., None)` so old records don't KeyError. Screenshot of the PR-branch modal on an alternate-port instance confirms an old record (no resume scalars) renders without error.

## Rubric

- [x] **1. Plan vs. implementation match** — pass — *All four locked decisions (D1 single-PM+dev-subagent, D2 ollama out, D3 four-scalar resume, D4 steer-preempt) implemented as written; success criteria exercised.*
- [x] **2. New code quality** — pass — *Type-hinted, fail-silent observability, generation-token preempt guard, capture-at-init, resume validation with cold-start fallback. No debug artifacts, no bare excepts, no TODOs.*
- [x] **3. Test coverage** — pass — *5 new runner suites + relocated driver/probe tests; 112 pass. Wedge-detection coverage replaced by turn-timeout + subprocess-liveness tests per audit-4.*
- [x] **4. Regression risk to existing callers** — pass — *Executor dispatch rewired BridgeAdapter→SessionRunner; no dangling references to removed model fields or the deleted transport seam.*
- [x] **5. Data integrity** — pass — *Idempotent ORM-safe migration registered; resume-scalar saves use `update_fields=[...]` + `updated_at`; nullable adds need no backfill (`_heal_descriptor_pollution`).*
- [x] **6. Security** — pass — *Explicit OAuth injection / `ANTHROPIC_API_KEY` strip in subprocess env; no new eval/exec/mark_safe/secrets.*
- [x] **7. Documentation accuracy** — pass — *Feature doc on main; architecture docs rewritten to the headless status quo; model docstrings updated to historical-record framing.*
- [x] **8. PR body accuracy** — pass — *File counts, deletions, test claims, and Closes list all verified against the diff and main.*
- [x] **9. Disclosed deferrals** — pass — *No-Gos map to OPEN separate-slug issues (#1923/#1925/#1926/#1927/#1802) or [ORDERED]/[DESTRUCTIVE] advisories.*
- [x] **10. Follow-up claims verified** — pass — *Closes #1924/#1918/#1919/#1921/#1724/#1879 correspond to the deleted surfaces; #1923 ollama-in-bridge scope correctly retained.*

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — PASS — *exercised the ones observable pre-deploy; deploy-gated ones (fleet cutover) are correctly [ORDERED].*
- **2. No-Gos from plan — none violated** — PASS — *ollama-in-bridge retained (#1923); no PTY fallback branch; transport keys in projects.json deferred to deploy.*
- **3. New `except Exception` blocks — each has logger/raise/swallow-ok** — PASS — *every broad catch in the runner/adapter logs and is annotated swallow-ok on a terminal/observability path.*
- **4. New integration tests — exercise serialization boundary** — PASS — *`test_runner_dispatch_e2e.py` / `test_headless_probe_e2e.py` against the real `claude` binary.*
- **5. Plan internal consistency — spike findings match task steps** — PASS — *four-scalar shape reconciled across D3/Failure-Path/task-1 (third-pass blocker addressed).*
- **7. New public APIs — docstrings present** — PASS — *runner/router/adapter fully docstringed.*
- **8. Breaking changes — migration path documented** — PASS — *idempotent ORM migration + stale-env guard + [ORDERED] deploy sequencing.*
- **10. Tests cover the failure path** — PASS — *subprocess-dies→error (never completed), empty-turn→wrapup, stale-scalar→cold-start, preempt races.*
- **11. UI changes — screenshot captured** — PASS — *PR-branch modal screenshot captured; renders clean on old records.*
- **12. Docs updated for user-facing changes** — PASS — *on main per convention.*

### Blockers
- None

### Tech Debt
- None

### Nits
- [ ] **`docs/plans/granite-pty-teardown.md:488`** — `grep -rn "fallback.*pty\|pty.*fallback" -i agent/ worker/ | wc -l` expected `0` — the anti-criterion verification row ("no PTY fallback branch") is imprecise: the case-insensitive `pty.*fallback` alternative false-matches the substring "em**pty**" in benign lines (`deliver_fallback`, "empty output … fallback"), so the row returns 7 and can never mechanically pass even though the No-Go is genuinely satisfied (there is no PTY fallback branch anywhere). This is the same "false-red by construction" class the plan's own critique caught for the `TOOL_TIMEOUT_DEFAULT_SEC` and pexpect-grep rows. Fix: anchor the token, e.g. `grep -rniE '(fallback[^\n]*pty_transport|pty_transport[^\n]*fallback|fallback[^\n]*\bpty\b|\bpty\b[^\n]*fallback)'` or scope to `\bpty\b` word boundaries. Non-code, low value — safe to skip with a note.

### Miscellaneous
- None

- **Fleet deploy / projects.json transport-key removal** — [ORDERED], sequenced post-merge via `/do-deploy`; not in this PR (correct — avoids stranding pre-cutover machines).
- **Machine-wide ollama removal** — tracked by #1923 (OPEN), out of session-path scope.
- **claude_code_sdk removal / ValorAgent chat migration** — tracked by #1925 (OPEN).
- **child_session_gate removal** — retained with rewritten rationale; deferred to #1633/#1926 (OPEN).
- **AgentSession schema diet beyond PTY fields** — tracked by #1927 (OPEN).

### Screenshots
- `generated_images/pr-1930/dashboard.png`, `generated_images/pr-1930/session_modal.png` (captured from a PR-branch UI instance on port 8501).

valorengels:
## Re-review (post-patch): APPROVED

The single nit from the prior review round (imprecise `pty.*fallback` anti-criterion grep in `docs/plans/granite-pty-teardown.md`) is **resolved on the authoritative copy** — no code change needed. Verified:

- main's plan doc already carries the word-boundary-anchored pattern (`fallback.*\bpty\b|\bpty\b.*fallback`), rewritten in commit 921fba7f before the review round. The naive pattern existed only in this branch's stale copy; plan docs are authoritative on main by repo convention.
- Empirically confirmed `\bpty\b` does not match "empty" (grep exits 1 on a control line).
- main's 4 remaining matches are all inside `agent/granite_container/`, the directory this PR deletes; on this branch the anchored grep returns 0 — the anti-criterion passes mechanically as designed.

**Verdict: APPROVED** — consensus of both judges (code-quality: APPROVED, 0 blockers; risk: APPROVED, 0 blockers). 0 blockers, 0 tech debt, 0 nits remaining. All hard gates green from the prior round (CLEAN/MERGEABLE, ruff clean, 112 session_runner unit tests passing, migration registered, docs landed on main).

valorengels:
## Review: PR Closed

This PR is no longer open (`state=MERGED`); review skipped.

The mergeability preflight runs before any diff reading — a closed/merged PR cannot take a formal code review (`gh pr review` requires an open PR). This PR was already merged, so no review is needed.

If you want a retrospective review of the merged changes, review the merge commit range on `main` directly instead of re-running `/do-pr-review`.

Issue #1979, opened 2026-07-09
Sticky response_delivered_at causes premature "completed" finalization of a resumed session while it's still running


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/delivery_guard_resume_epoch_scoping.md

> **Context:** `agent/session_health.py` implements this repo's session health-check/recovery
> logic — periodically scanning `AgentSession` rows with `status="running"` and recovering
> (transitioning out of "running") any that look stuck, per the design in
> [`docs/features/session-recovery-mechanisms.md`](docs/features/session-recovery-mechanisms.md).
> `response_delivered_at` is a `DatetimeField` on `AgentSession` (`models/agent_session.py:158`)
> stamped whenever a response is actually delivered to the human (`agent/session_executor.py:1504-1505`,
> `agent/session_completion.py:1079-1080`). A session can be resumed after reaching a terminal
> status (`failed`/`completed`) via `tools/valor_session.py`'s `resume_session()`, which
> transitions it back to `pending` for a fresh attempt (`transition_status(session, "pending",
> ..., reject_from_terminal=False)`).

## Problem

The health check's "Delivery guard" (`agent/session_health.py:3196-3219`) exists to catch
sessions that delivered a response but got stuck in `running` instead of finalizing — a
real failure mode. But its check only looks at whether `response_delivered_at` is set at
all, not whether that timestamp belongs to the *current* run. Since nothing in the resume
path (`tools/valor_session.py::resume_session`) ever clears `response_delivered_at`, a
session that delivered a response during any *prior* attempt carries that stale timestamp
into every future resume. The very next health-check tick after a resume can then
force-finalize the session as `completed` — even while it is still legitimately working on
new content.

**Current behavior:**

Observed 2026-07-09 on session `43b42260899a4f1f81f254fa2b1fc2bb` (Telegram thread
`tg_psyoptimal_-1003743854645_335`, project `psyoptimal`):

1. The session delivered a response at `05:55:12` during a prior (failed) attempt.
2. It was resumed via `valor_session resume` for a fresh attempt at `10:20:26`.
3. ~4 minutes in, confirmed still genuinely alive — `ps` showed the underlying
   `claude -p --resume ...` subprocess (PID 78380) still running, not crashed.
4. The health check fired anyway:
   ```
   [session-health] Session 43b42260899a4f1f81f254fa2b1fc2bb already delivered response at
[...]

Issue #1916, opened 2026-07-06
sdlc-tool stage-marker --status in_progress rejected at PLAN stage (2 independent reproductions)


tomcounsell:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/stage-marker-in-progress-plan-rejection.md

## Summary

In two independent pipelines of the 2026-07-05 six-issue batch (#1902 and #1905 runs), the PLAN stage's `sdlc-tool stage-marker --status in_progress` write was rejected: "substrate present, session resolved, but the state-machine write was rejected/raised. State NOT persisted." The subsequent `--status completed` write for the same stage succeeded in both cases.

## Impact

- PLAN completes with no `in_progress` transition ever recorded — any consumer that keys off the `in_progress` state (dashboards, stall detection, dispatch-history analysis) sees the stage jump pending→completed.
- Possibly related: one `sdlc-tool stage-marker` call in the #1898 run returned `{}` silently.

## Repro context

Both reproductions were the first marker write of a fresh pipeline on a skills-only machine (stage markers noted as "not persisting reliably" there in one supervisor report). Suspect an ordering constraint in the state machine (e.g. `in_progress` requiring a prior transition that doesn't exist yet at PLAN) or a substrate init race on first write.

## Acceptance

- `stage-marker --status in_progress` at PLAN on a fresh issue persists (or the state machine's rejection is made loud and documented).
- A test covering the first-write-at-PLAN path.

PR #1985, opened 2026-07-09
fix(#1980): preserve valid completion when resumed turn exits non-zero after a result event

### LINKED ISSUE #1980: Wrap-up guard delivers canned OPERATOR_TERMINAL_MESSAGE despite a valid, complete model response
> **Context:** The headless session runner (`agent/session_runner/runner.py`) drives a PM
> (project-manager) persona through bounded turns. When a run ends without ever delivering
> a `[/user]` or `[/complete]` message, a "wrap-up guard" (`_run_wrapup_guard`,
> `agent/session_runner/runner.py:1149-1183`) runs one bounded extra turn asking the PM to
> send its summary now. `HeadlessRoleDriver.run_turn` (`agent/session_runner/role_driver.py:346`)
> is the method that actually spawns the `claude -p` subprocess and returns a
> `HeadlessTurnOutcome` with the model's reply text. `classify_pm_prefix`
> (`agent/session_runner/router.py:73`) parses that reply text for a `[/user]`/`[/complete]`
> token — it has both a strict first-line regex and a permissive fallback that searches the
> first 200 characters, so a token that isn't literally on line 1 is still recognized (this
> was independently verified while investigating this issue — see Recon Summary).

## Problem

A PM turn can produce a fully valid, complete, well-formed `[/complete]` (or `[/user]`)
response — verified present in the raw Claude Code transcript with `stop_reason: "end_turn"`
— and still result in the user receiving the generic canned fallback
(`OPERATOR_TERMINAL_MESSAGE`, `agent/session_runner/runner.py:181-183`: *"I wasn't able to
produce a response to this. Please rephrase or follow up."*) instead of the real content.

**Current behavior:**

Observed 2026-07-09 on session `43b42260899a4f1f81f254fa2b1fc2bb` (Telegram thread
`tg_psyoptimal_-1003743854645_335`, project `psyoptimal`, Claude session UUID
`36514af3-c4e9-455d-9087-f5850101990e`):

1. The wrap-up guard ran its bounded extra turn. The underlying `claude -p` subprocess
   produced a complete, valid final assistant message (persisted in the raw `.jsonl`
   transcript at `~/.claude/projects/.../36514af3-....jsonl`, line 100):
   ```
   Merged and verified. `origin/main` now has the hotfix at the top.

   [/complete]
   Send Reminders hotfix shipped to main — PR #604 merged as squash commit `e0fbb557`
   ("hotfix: Send Reminders — scope core-values to selection + one message per user").
   ...
   ```
   with `stop_reason: "end_turn"` — a clean, natural stop, not a crash or truncation.
2. Despite this, the session's own `session_events` log shows **zero**
   `runner_complete_routed` events and only one `runner_user_routed` event for the *entire*
   run — and that one event is the 74-character `OPERATOR_TERMINAL_MESSAGE`, not the real
   ~1,300-character completion text.
3. `_run_wrapup_guard`'s code (`runner.py:1157-1181`) can only reach the
   `OPERATOR_TERMINAL_MESSAGE` fallback if `outcome.reply_text` (from
   `self._driver.run_turn(...)`) came back falsy. Since the actual model output was
   non-empty and well-formed, the gap is between the subprocess producing that output and
   `HeadlessRoleDriver.run_turn` returning it — specifically the `if not reply:` empty-output
   guard at `agent/session_runner/role_driver.py:433-436`, which sets
   `outcome.reply_text = ""` and `exit_reason = "empty_output"` whenever the harness
   function (`get_response_via_harness`, `agent/sdk_client.py`) returns a falsy `reply`.
4. No kill/preempt/timeout log line appears for this session in the relevant window
   (`logs/worker.log`, `10:31`-`10:32` UTC) — the process appears to have completed its own
   natural run, not been externally terminated mid-turn.

**Desired outcome:**

When the underlying harness subprocess produces a valid, complete final message, that
message reaches `HeadlessRoleDriver.run_turn`'s return value and is delivered to the human
— even inside the wrap-up guard's bounded extra turn. `OPERATOR_TERMINAL_MESSAGE` should be
reserved for a genuinely silent PM, not for a PM whose real answer got lost between the
harness and the runner.

## Definitions

| Wrap-up guard | A bounded extra PM turn run when the normal turn loop ends without ever delivering a user-facing message, to guarantee the human always receives something. | `agent/session_runner/runner.py:1149-1183` |
| `HeadlessRoleDriver` | The class that spawns one `claude -p` subprocess per turn and returns a classified `HeadlessTurnOutcome`. | `agent/session_runner/role_driver.py:147` |
| `classify_pm_prefix` | Deterministic regex parser that reads a PM turn's text and routes it to `user`/`complete`/`unknown` based on a `[/user]`/`[/complete]` token. | `agent/session_runner/router.py:73` |

## Prior Context

Related work (different root causes, ruled out as the same bug):
- Investigated whether this is a duplicate of #1979 (sticky `response_delivered_at` causing
  premature `completed` finalization on the same session, filed today) — ruled out as the
  same root cause: #1979's premature finalization happened at `10:24:28`, well before this
  wrap-up turn ran at `10:31:00`-`10:32:08`, and no kill/preempt event fired against this
  process in that window (see Recon Summary) — the process appears to have completed this
  turn on its own, not been cut off by #1979's mis-finalization.
- No existing open or closed issue found for "wrap-up guard delivers canned message despite
  valid model output" (searched `gh issue list --search` for `OPERATOR_TERMINAL_MESSAGE`,
  `pm_floor_delivered`, `empty_output` role_driver — no matches).

## Solution Sketch

This is a structural/architectural gap where the root cause is not yet pinned down at the
harness-internals level — writing open questions rather than a prescribed approach, per
this skill's own guidance for uncertain root causes.

**Open questions for `/do-plan` or a dedicated investigation to resolve:**
- Inside `get_response_via_harness` (`agent/sdk_client.py`), under what conditions does the
  function return a falsy `reply` despite the subprocess's stream-json output containing a
  valid `result` event with non-empty text? Is there a code path where the `result` event's
  text is captured but not correctly propagated to the return value used by
  `HeadlessRoleDriver.run_turn`?
- Does `HeadlessRoleDriver.run_turn`'s own turn-timeout (`asyncio.wait_for(..., timeout=self.turn_timeout_s)`,
  `role_driver.py:381-405`) or another bounded-wait mechanism specific to the wrap-up
  guard's call path have a narrower effective timeout than the main loop's turns, such that
  a turn that takes ~68 seconds (10:31:00 to 10:32:08, per the transcript) risks a
  race between "subprocess finished and wrote its transcript" and "harness function
  returned its captured text"?
- Is there a reproducible minimal case (a turn whose only work is a few read-only tool
  calls plus a final `[/complete]` text block) that reliably triggers this, or was this a
  rare, hard-to-reproduce race?

## Recon Summary

**Confirmed:**
- The final assistant message's actual text, extracted verbatim from the raw `.jsonl`
  transcript and run through the real `classify_pm_prefix` function in a live Python
  session, correctly returns `destination="complete"`, `compliance_miss=True` (marker not
  on line 1, caught by the fallback regex), and a non-empty 1,361-character payload. The
  classifier is not the bug.
- `session_events` for this AgentSession record shows exactly one `runner_user_routed`
  event (the 74-char terminal message) and zero `runner_complete_routed` events across the
  entire run — proving the real completion payload was never handed to either delivery
  callback.
- No kill/timeout/preempt log line for this session/PID appears in `logs/worker.log`
  between `10:20` and `10:32` UTC — ruling out an external interruption as the proximate
  cause of the lost text.

**Pre-requisites:**
- Reproducing this deterministically will likely require adding temporary diagnostic
  logging inside `get_response_via_harness`'s `result`-event handling and
  `HeadlessRoleDriver.run_turn`'s post-harness-call branch, then re-running a similar
  long-running wrap-up-guard scenario, since the existing logs don't capture enough detail
  to distinguish "harness returned empty" from "runner discarded a non-empty return."

## Acceptance Criteria

- [ ] Root cause identified: the specific code path where a valid, complete model response
      fails to propagate from the harness subprocess to `HeadlessRoleDriver.run_turn`'s
      return value.
- [ ] A regression test reproduces the failure mode (a turn whose subprocess produces valid
      final text) and asserts the real text is delivered, not `OPERATOR_TERMINAL_MESSAGE`.
- [ ] `OPERATOR_TERMINAL_MESSAGE` only fires when the PM turn genuinely produced no text at
      all (verified empty, not merely unpropagated).

## Downstream

This issue feeds into `/do-plan`, which will produce a plan document at
`docs/plans/{slug}.md` (e.g. `docs/plans/wrapup-guard-reply-text-loss.md`), then executed by
`/do-build`. Given the uncertain root cause, the plan's first phase should likely be a
recon/spike to instrument and reproduce before committing to a fix approach.

valorengels:
Closes #1980

## Problem

A PM turn (including the wrap-up guard's bounded extra turn) could produce a fully valid, complete `[/complete]`/`[/user]` final message — persisted in the raw transcript with `stop_reason: "end_turn"` — and the human still received the canned `OPERATOR_TERMINAL_MESSAGE` instead of the real content.

## Root cause

In `get_response_via_harness` (`agent/sdk_client.py`), the stale-UUID fallback fired on `prior_uuid and returncode != 0` **without checking whether the primary `--resume` subprocess had already emitted a `result` event**. When a resumed subprocess emitted a valid completion (`result` event) and then exited non-zero (a post-turn/cleanup artifact, not an external kill), the fallback re-ran a fresh session whose empty output **clobbered** the good `result_text`. `get_response_via_harness` returned `""`, `HeadlessRoleDriver.run_turn`'s `if not reply:` guard set `exit_reason="empty_output"`, and `_run_wrapup_guard` delivered `OPERATOR_TERMINAL_MESSAGE`.

This defeated a contract the role driver already documents but cannot enforce (`role_driver.py`: *"a nonzero exit AFTER a result event keeps the result — the event is the protocol's completion signal"*), because the clobber happened one layer below it.

## Fix

Gate the fallback on the true `result_event_fired` boolean, captured from the primary invocation's `on_exit_status(returncode, result_event_fired)` callback, and gate on `not primary_result_event_fired`. Critically this is **not** `result_text is None`: `_run_harness_subprocess` returns a non-None string both when a result event fired (BRANCH A) and when partial streamed text accumulated with no result event (BRANCH B), so `result_text is None` would wrongly skip the fallback in BRANCH B. Keying off the real boolean:

- **BRANCH A** (result event fired) + non-zero exit → fallback **skipped**, completion kept (the fix).
- **BRANCH B** (partial text, no result event) + non-zero exit → fallback **still fires** (crashed-subprocess recovery preserved).
- **BRANCH C** (genuine stale UUID, no output) + non-zero exit → fallback **still fires** (stale-UUID recovery preserved).
- Empty-string result event (`""`, fired=True) → fallback skipped, returns `""` — `OPERATOR_TERMINAL_MESSAGE` stays reserved for a genuinely empty PM turn.

Added a `logger.info` at the gate recording the skip-vs-fire decision for incident triage.

## Acceptance criteria (issue #1980)

1. ✅ Root cause identified — the stale-UUID fallback clobber in `get_response_via_harness`.
2. ✅ Regression test reproduces the failure mode and asserts the real text is delivered (not `OPERATOR_TERMINAL_MESSAGE`) — `test_branch_a_result_event_kept_on_nonzero_exit` + the end-to-end `test_run_turn_returns_completion_on_nonzero_exit_with_result`.
3. ✅ `OPERATOR_TERMINAL_MESSAGE` only fires on a genuinely empty turn — `test_branch_a_empty_result_event_is_genuinely_empty`, plus BRANCH B/C tests proving recovery is not regressed.

## Testing

- New: `tests/unit/test_harness_stale_uuid_result_preservation.py` (5 tests, all four branches + run_turn e2e) — passing.
- Regression sweep: `test_harness_retry.py`, `test_sdk_client_harness_counters.py`, `test_sdk_client.py`, `test_sdk_client_image_sentinel.py`, `test_headless_role_driver.py`, `test_runner_resume.py` — all passing.
- `ruff check` + `ruff format --check` clean.

## Plan & docs

- Plan: `docs/plans/wrapup-guard-reply-text-loss.md` (critique verdict addressed: gate corrected from `result_text is None` to the true `result_event_fired` boolean).
- Docs: `docs/features/headless-session-runner.md` — new subsection on the stale-UUID fallback vs. result-event completion signal.

### DIFF
--- a/agent/sdk_client.py
+++ b/agent/sdk_client.py
@@ -2383,9 +2383,13 @@ async def get_response_via_harness(
     and pathological large single messages it bounds the argv to prevent the
     binary's "Separator is found" overflow crash.
 
-    If the resumed subprocess exits with **any** non-zero return code, retries
-    once using ``full_context_message`` without ``--resume`` (no stderr substring
-    gate — substring matching is brittle across CLI versions and locales).
+    If the resumed subprocess exits with **any** non-zero return code **without
+    having emitted a ``result`` event**, retries once using
+    ``full_context_message`` without ``--resume`` (no stderr substring gate —
+    substring matching is brittle across CLI versions and locales). Issue #1980:
+    a non-zero exit that follows a fired ``result`` event keeps the captured
+    completion and does NOT retry — the ``result`` event is the protocol's
+    completion signal, and a fresh retry would clobber a valid answer.
 
     A separate exit-code-0 sentinel check (``IMAGE_DIMENSION_SENTINEL``) is placed
     **above** the ``returncode != 0`` guard to handle Claude Code's image-dimension
@@ -2520,6 +2524,26 @@ async def get_response_via_harness(
             "model": model or "default",
         }
 
+    # Capture whether the PRIMARY invocation emitted a `result` event (issue
+    # #1980). `_run_harness_subprocess` reports this as the second arg of
+    # `on_exit_status(returncode, result_event_fired)` — it is the ONLY precise
+    # signal for "a result event fired." The returned `result_text` cannot stand
+    # in for it: `_run_harness_subprocess` returns a non-None string both when a
+    # result event fired (BRANCH A) AND when no result event fired but partial
+    # streamed text accumulated (BRANCH B), so `result_text is None` is true only
+    # in BRANCH C. The stale-UUID fallback below gates on this boolean so a
+    # resumed turn that produced a valid completion is never discarded by a
+    # destructive fresh-session retry on a post-turn non-zero exit.
+    primary_result_event_fired = False
+
+    def _capture_primary_exit(rc: int | None, fired: bool) -> None:
+        nonlocal primary_result_event_fired
+        primary_result_event_fired = fired
+        # Chain to the caller's callback so the role driver's exit_statuses
+        # list still receives every subprocess invocation (residual #1916).
+        if on_exit_status is not None:
+            on_exit_status(rc, fired)
+
     # Call site 1 of 3 — primary harness invocation. 8-tuple unpack
     # (issue #1099 Mode 1 added stderr_snippet; issue #1245 added num_turns
     # and tool_call_count).
@@ -2541,7 +2565,7 @@ async def get_response_via_harness(
         on_sdk_finished=on_sdk_finished,
         on_stdout_event=on_stdout_event,
         on_init=on_init,
-        on_exit_status=on_exit_status,
+        on_exit_status=_capture_primary_exit,
         ttft_metadata=_ttft_meta,
         true_session_id=session_id,
     )
@@ -2600,11 +2624,23 @@ async def get_response_via_harness(
             )
 
     # Mandatory stale-UUID fallback: when prior_uuid was set and the subprocess
-    # exits with ANY non-zero return code, retry once without --resume using the
-    # full-context message. The fallback does NOT inspect stderr — substring
-    # matching is brittle across CLI versions and locales, and an unnecessary
-    # retry on a non-stale-UUID error costs only one extra subprocess spawn.
-    if prior_uuid and returncode is not None and returncode != 0:
+    # exits with ANY non-zero return code WITHOUT having emitted a `result` event,
+    # retry once without --resume using the full-context message. The fallback
+    # does NOT inspect stderr — substring matching is brittle across CLI versions
+    # and locales, and an unnecessary retry on a non-stale-UUID error costs only
+    # one extra subprocess spawn.
+    #
+    # Issue #1980: gate on `not primary_result_event_fired`. A resumed turn that
+    # emitted a `result` event is a SUCCESSFUL resume; its completion is the
+    # protocol's completion signal (mirrors the role driver's residual-#1916
+    # contract at role_driver.py: "a nonzero exit AFTER a result event keeps the
+    # result"). Without this gate, a valid completion followed by a post-turn
+    # non-zero exit triggered a fresh-session retry whose empty output clobbered
+    # the good result_text — the wrap-up guard then delivered the canned
+    # OPERATOR_TERMINAL_MESSAGE instead of the real answer. The fallback still
+    # fires when no result event fired (BRANCH B partial text, or BRANCH C genuine
+    # stale UUID), preserving all pre-existing recovery.
+    if prior_uuid and returncode is not None and returncode != 0 and not primary_result_event_fired:
         if full_context_message is not None:
             logger.warning(
                 f"[harness] Stale UUID {prior_uuid} for session_id={session_id}, "
@@ -2652,6 +2688,18 @@ async def get_response_via_harness(
                 "falling back to first-turn path — no full_context_message available"
             )
             result_text = None
+    elif prior_uuid and returncode is not None and returncode != 0 and primary_result_event_fired:
+        # Issue #1980 observability: a resumed turn exited non-zero but a result
+        # event fired first, so the stale-UUID fallback is deliberately skipped
+        # and the captured completion is kept. Logged so incident triage can tell
+        # this apart from a fallback that fired.
+        logger.info(
+            "[harness] Resumed turn exited %s AFTER a result event for "
+            "session_id=%s — keeping the completion, skipping stale-UUID fallback "
+            "(issue #1980)",
+            returncode,
+            session_id,
+        )
 
     # Store the Claude Code UUID for next-turn --resume (#976)
     if session_id and session_id_from_harness:
--- a/docs/features/headless-session-runner.md
+++ b/docs/features/headless-session-runner.md
@@ -158,6 +158,40 @@ Stale or invalid scalars (missing `runner_cwd`, unknown `claude_session_uuid`)
 discard cleanly to a cold start with a full first-turn prime — there is no
 crash on a bad resume pointer.
 
+### Stale-UUID fallback vs. the result-event completion signal (issue #1980)
+
+A resumed (`--resume`) turn whose subprocess exits **non-zero** may need one
+fresh-session retry — a genuinely stale/invalid UUID makes `claude --resume`
+error out before producing any output. `get_response_via_harness`
+(`agent/sdk_client.py`) runs exactly that retry once, without `--resume`, using
+the caller's `full_context_message`.
+
+That retry is **gated on whether the primary invocation emitted a `result`
+event**, not on whether it exited zero. The invariant: *a `result` event is the
+protocol's completion signal.* If the resumed subprocess emitted a `result`
+event (`stop_reason: end_turn`) and only *then* exited non-zero — a post-turn or
+cleanup artifact — the captured completion is authoritative and the fallback is
+**skipped**, keeping the real answer. This mirrors, one layer down, the role
+driver's residual-#1916 rule (`role_driver.py`: "a nonzero exit AFTER a result
+event keeps the result").
+
+The gate keys off the true `result_event_fired` boolean, captured from the
+primary invocation's `on_exit_status(returncode, result_event_fired)` callback —
+**not** off the returned `result_text`. `result_text` is a non-empty string in
+two distinct cases (a fired result event, or accumulated partial text with *no*
+result event from a crashed subprocess), so `result_text is None` cannot
+distinguish "resume succeeded" from "crashed with partial text." The fallback
+therefore still fires whenever no result event fired (partial text or a genuine
+stale UUID), preserving all recovery.
+
+Before this gate existed, a valid completion followed by a non-zero exit
+triggered the retry, whose empty output overwrote the good `result_text`;
+`get_response_via_harness` returned `""`, `HeadlessRoleDriver.run_turn`'s
+`if not reply:` guard set `exit_reason="empty_output"`, and the wrap-up guard
+delivered the canned `OPERATOR_TERMINAL_MESSAGE` instead of the real answer.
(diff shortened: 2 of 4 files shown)


valorengels:
## Review: Approved

Clean, minimal, correctly-scoped bugfix for #1980. The stale-UUID fallback in `get_response_via_harness` now gates on the true `result_event_fired` boolean (captured from the primary invocation's `on_exit_status` callback), so a resumed subprocess that emits a valid `result` event and then exits non-zero keeps its completion instead of being clobbered by a fresh-session retry — the mechanism that was delivering `OPERATOR_TERMINAL_MESSAGE` in place of the real text.

I verified the three load-bearing concerns:

- **`on_exit_status` wrapper correctness.** `_capture_primary_exit` sets `primary_result_event_fired = fired` **before** chaining to the caller's callback, and guards the chain with `if on_exit_status is not None`. It's applied only at primary call site 1; sites 2/3 keep the raw callback. `role_driver.py:402`'s `exit_statuses.append((rc, fired))` still receives every invocation, so the residual-#1916 guard at `role_driver.py:457-468` reads the correct authoritative exit shape (primary when the fallback is skipped; fallback as the last entry when it fires).
- **The boolean is genuinely distinct from `result_text is None`.** Confirmed at `sdk_client.py:3133-3137`: inside `_run_harness_subprocess` the local `result_text` is non-None *iff a result event fired*, and `on_exit_status(returncode, result_text is not None)` passes exactly that. The returned tuple's first element is non-None in both BRANCH A (result event) and BRANCH B (accumulated `full_text`, no result event), which is why keying off `result_text is None` would have regressed BRANCH B. The gate keys off the boolean, so BRANCH B/C recovery is preserved.
- **BRANCH B/C recovery preserved.** Verified by code and by the passing regression matrix.

Other callers (`session_completion.py:757` Pass 1 with `prior_uuid=pm_uuid`, `:819` Pass 2 with `prior_uuid=None`) are safe: Pass 1 benefits identically (keeps completion) and passes no `on_exit_status`; Pass 2 never reaches the gate.

## Rubric

- [x] **1. Plan vs. implementation match** — pass — all three issue acceptance criteria delivered and tested (BRANCH A kept, BRANCH A-empty genuinely empty, BRANCH B/C recovery).
- [x] **2. New code quality** — pass — type-hinted wrapper, lazy `%`-style log formatting, clear comments; ruff check/format clean.
- [x] **3. Test coverage** — pass — 5 new tests cover all four branches at `get_response_via_harness` plus an end-to-end `run_turn` test that exercises the real gate (`harness_fn=None`).
- [x] **4. Regression risk to existing callers** — pass — chained callback preserves the exact `(rc, fired)` contract; both `session_completion.py` callers audited safe; 96 existing harness/role_driver/runner_resume/image_sentinel tests green.
- [x] **5. Data integrity** — n/a — no schema/model/migration changes.
- [x] **6. Security** — pass — no secrets, no unsafe patterns; one added `logger.info`/`logger.warning`.
- [x] **7. Documentation accuracy** — pass — `docs/features/headless-session-runner.md` gains an accurate subsection; docstring updated to "without having emitted a result event."
- [x] **8. PR body accuracy** — pass — file list, branch behavior, and test claims all match the diff (test filenames live under `tests/unit/session_runner/`; the new file is at `tests/unit/`).
- [x] **9. Disclosed deferrals** — pass — the only No-Go (#1979 sticky finalization) is `[SEPARATE-SLUG]` with an open tracking issue.
- [x] **10. Follow-up claims verified** — n/a — no follow-up claims in the PR body.

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — PASS — Verification table run on the branch; all 6 checks pass.
- **2. No-Gos from plan — none violated** — PASS — #1979 left untouched.
- **3. New `except Exception` blocks — each has logger/raise/swallow-ok** — N/A — none added; the wrapper adds no handlers.
- **4. New integration tests — exercise serialization boundary (not in-memory only)** — N/A — this is a subprocess-callback gate; the e2e test drives the real `run_turn` path, which is the correct boundary here.
- **5. Plan internal consistency — spike findings match task steps** — PASS — Critique Results table documents the `result_text is None` → boolean correction; implementation matches the corrected approach.
- **7. New public APIs — docstrings present** — N/A — no new public API; inner helper is a local closure.
- **8. Breaking changes — migration path documented** — N/A — no interface/signature change.
- **9. Tests added for new behavior** — PASS — 5 new tests.
- **10. Tests cover the failure path (not just happy path)** — PASS — BRANCH B (crash + partial text) and BRANCH C (stale UUID) recovery, plus empty-result-event, are all asserted.
- **11. UI changes (if any) — screenshot captured** — N/A — no UI files in the diff.
- **12. Docs updated for user-facing changes** — PASS — feature doc + docstring updated.

### Verified
- [x] Code correctness
- [x] Test coverage
- [x] Plan requirements met

### Blockers
- None

### Tech Debt
- None

### Nits
- None

### Miscellaneous
- None

- **[SEPARATE-SLUG #1979] sticky `response_delivered_at` premature finalization** — distinct root cause, tracked separately in an open issue; correctly out of scope.

### Verification Results
- Gate change present (`grep`) — exit 0
- `pytest tests/unit/test_harness_stale_uuid_result_preservation.py` — 5 passed
- `pytest test_harness_retry.py test_sdk_client_harness_counters.py test_sdk_client.py` — 53 passed, 1 skipped
- `ruff check` — passed; `ruff format --check` — clean
- No stale xfails — exit 1 (none present)
- Extra sweep: `test_headless_role_driver.py`, `test_runner_resume.py`, `test_sdk_client_image_sentinel.py` — 43 passed

### Screenshots
- N/A — no UI changes.

Note for deploy: touches `agent/sdk_client.py` (harness call path) — requires a bridge/worker restart on deploy.

Issue #2002, opened 2026-07-10
Tool-timeout guard reads current_tool_name/last_tool_use_at without run-scoping (same class as #1614/#1979)


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/tool-timeout-guard-run-scoping.md

> **Context**: `agent/session_health.py` runs the worker's health-check loop, which periodically inspects `running` `AgentSession` rows and can requeue a session back to `pending` (a "recovery") if it looks stuck or crashed. Issue #1614 fixed one class of bug in this loop: some fields on a session row are "sticky" — they persist across a resume (terminal → pending → running) — so a stale value from a *prior* run can be misread by the health check as describing the *current* run. #1614's fix gated `turn_count`, `log_path`, and `claude_session_uuid` behind a freshness check (`NO_OUTPUT_BUDGET_SECONDS`) so a stale value can't trigger a false recovery. Issue #1979 is fixing the same class of bug for a different field, `response_delivered_at` (the "Delivery guard"), via a per-run "epoch" comparison against `started_at`. This issue reports a second field pair with the same shape of bug, found during #1979's sticky-field audit (the audit #1614 asked future work to repeat whenever a related fix touches this file).

## Problem

**Current behavior:**

`agent/session_health.py`'s `_check_tool_timeout()` (called every 30s from the `_agent_session_tool_timeout_check` sub-loop) reads `current_tool_name` and `last_tool_use_at` off a `running` session row to decide whether the session is wedged inside a single tool call for too long ("tool timeout"). Neither field is checked for recency against the current run's start — the function does not compare `last_tool_use_at` to `started_at`/`created_at` at all.

These two fields are only cleared on one specific requeue path: the `reason_kind == "tool_timeout"` branch inside the recovery transition (`agent/session_health.py:2621-2623`, added by #1762). Every *other* path that requeues a `running` session back to `pending` — worker-startup recovery (`_recover_interrupted_agent_sessions_startup`) and the `no_progress`/`worker_dead` branches of `_apply_recovery_transition` — leaves `current_tool_name` and `last_tool_use_at` untouched.

[...]

PR #2006, opened 2026-07-10
Fix: Delivery guard force-finalizes resumed sessions on stale prior-run delivery timestamp

### LINKED ISSUE #1979: Sticky response_delivered_at causes premature "completed" finalization of a resumed session while it's still running

valorengels:
## Summary

`agent/session_health.py`'s Delivery guard force-finalized any `running` session whose `response_delivered_at` field was set — regardless of whether that delivery happened during the *current* run. Since nothing clears `response_delivered_at` on resume, a session resumed after a prior (already-delivered) attempt carried the stale timestamp forever and got killed on the very next health-check tick, even while a live subprocess was still legitimately working.

Fixes #1979.

## What changed

- Added `_delivery_belongs_to_current_run(entry) -> bool` to `agent/session_health.py`, next to the existing `_ts()` normalizer. Returns `True` only when `response_delivered_at >= (started_at or created_at)` — i.e. the delivery belongs to this run's epoch. Rows with no anchor at all (both `started_at`/`created_at` `None`) preserve the original always-fire behavior.
- Both Delivery guard call sites — `_apply_recovery_transition()` and `_agent_session_health_check()` — now call this shared predicate instead of the bare `is not None` presence check. Comments above each guard updated to describe epoch scoping.
- New regression test module `tests/unit/test_delivery_guard_resume_epoch.py`: unit coverage of the helper (stale-before-resume, boundary `==`, normal same-run, no-anchor fallback, `None`/garbage input, `created_at` fallback) plus integration coverage against `_apply_recovery_transition` confirming a stale prior-run delivery falls through to the normal requeue path while a genuine same-run delivery still force-finalizes.
- Fixed one existing fixture in `tests/unit/test_agent_session_queue.py` (`test_no_progress_with_delivered_response_finalized_completed`) that hard-coded `response_delivered_at` to a fixed 2024 date — under the new epoch rule this predated `started_at` and the test would have silently exercised the "guard declines to fire" path instead of the "guard fires" path it's meant to assert. Anchored the timestamp to `started_at + 1s` instead.
- Docs: `docs/features/session-recovery-mechanisms.md` gets a new subsection on the Delivery guard's epoch scoping.

## #1614 sticky-field audit (required by the plan)

Per the plan's Task 4, audited every AgentSession field read by `agent/session_health.py`'s health-check/recovery/tool-timeout paths for the same "sticky field read across a resume without run-scoping" bug class.

**No new gap in the fields this fix touches.** `turn_count`/`log_path`/`claude_session_uuid` are already gated per #1614; `started_at`, `claude_pid`, `last_heartbeat_at`/`last_turn_at`/`last_compaction_ts` are either re-stamped fresh on pickup or self-gate against "now"; `recovery_attempts` is an intentional lifetime counter, not a gap.

**One additional candidate gap found, filed separately rather than folded into this PR:** `current_tool_name` + `last_tool_use_at`, read unguarded in `_check_tool_timeout()` (`agent/session_health.py:4074`), only cleared on the `tool_timeout` requeue branch — every other requeue path leaves them stale across a resume, so a resumed session can be mis-flagged as tool-wedged on its first post-resume tick (self-heals after one cycle, but bumps `recovery_attempts` and sends a spurious steering message). Filed as issue #2002, fixable via the same one-line epoch-comparison pattern. Kept out of this PR per the plan's No-Gos (new gap ⇒ separate issue).

## Test plan

- [x] `pytest tests/unit/test_delivery_guard_resume_epoch.py tests/unit/test_agent_session_queue.py tests/unit/test_health_check_recovery_finalization.py -q` — all pass
- [x] `pytest tests/unit/test_session_health_subprocess_kill.py tests/unit/test_zombie_session_resurrection.py tests/unit/test_mcp_hang_graceful_degradation.py -q` — all pass (other files touching `response_delivered_at` fixtures, unaffected)
- [x] Full `pytest tests/unit/` — passes aside from 3 pre-existing, order-dependent/environment failures unrelated to this change (`test_ui_reflections_data.py`, `test_bot_registry_routing.py`, `test_session_health_sibling_phantom_safety.py` — all pass in isolation; the reflections ones fail only because this worktree lacks the iCloud-synced `config/reflections.yaml` symlink)
- [x] `grep -c "_delivery_belongs_to_current_run" agent/session_health.py` → 3 (def + 2 call sites)
- [x] `grep -n "response_delivered_at" tools/valor_session.py` → no assignment (fix is read-site only, field history preserved)
- [x] `ruff check .` / `ruff format --check .` — clean


valorengels:

A tightly-scoped, correct bug fix for the sticky `response_delivered_at` premature-finalization bug (#1979). The epoch predicate `_delivery_belongs_to_current_run` is applied at both guard sites, the `>=` boundary is right, the legacy no-anchor fallback preserves original behavior, tests are thorough (8 helper cases + 2 integration + a corrected fixture), and docs are updated. One documentation-accuracy nit: a function docstring that this PR touches was not updated to match the new epoch-scoped behavior.

## Rubric

- [x] **1. Plan vs. implementation match** — pass — *Both guard sites (`_apply_recovery_transition`, `_agent_session_health_check`) use the shared predicate; #1614 sticky-field audit done; every acceptance criterion delivered.*
- [x] **2. New code quality** — pass — *`_delivery_belongs_to_current_run` reuses the existing `_ts()` normalizer, has a type hint, a docstring, and clear branch logic.*
- [x] **3. Test coverage** — pass — *`test_delivery_guard_resume_epoch.py`: helper unit cases (stale-before-resume, `==` boundary, same-run, no-anchor, None/garbage input, `created_at` fallback) + integration against `_apply_recovery_transition`. All 51 pass.*
- [x] **4. Regression risk to existing callers** — pass — *Both bare-presence call sites replaced with the predicate. The one existing fixture that hard-coded a 2024 `response_delivered_at` (which would have silently flipped to the "guard declines" path) was re-anchored to `started_at + 1s`.*
- [x] **5. Data integrity** — n/a — *Read-site predicate only; no schema/field change, no migration.*
- [x] **6. Security** — pass — *No security surface; no secrets, eval, or subprocess-with-input.*
- [ ] **7. Documentation accuracy** — fail — *The `_agent_session_health_check` docstring (lines ~3150-3156) still reads "the health check inspects `response_delivered_at`. If the field is set, the session already delivered..." — the bare-presence framing this PR replaced. The inline comments above both guard blocks and the two feature docs were correctly updated; this function docstring was missed.*
- [x] **8. PR body accuracy** — pass — *`grep -c` claims verified: 3 references (def + 2 sites); changed-file list matches; #2002 is OPEN.*
- [x] **9. Disclosed deferrals** — pass — *The `current_tool_name`/`last_tool_use_at` follow-up is deferred to #2002 (OPEN), consistent with the plan's `[SEPARATE-SLUG]` No-Go.*
- [x] **10. Follow-up claims verified** — pass — *#2002 resolves to an OPEN GitHub issue.*

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — PASS — *All success criteria met; both guard sites, legacy edge, and audit note present.*
- **2. No-Gos from plan — none violated** — PASS — *Predicate extracted (not the whole block); field not cleared on resume; audit stayed a read-and-confirm pass; new gap filed as #2002 not folded in.*
- **3. New `except Exception` blocks — each has logger/raise/swallow-ok** — N/A — *No new except blocks; helper is exception-free by construction (returns False on unparseable input via `_ts`).*
- **4. New integration tests — exercise serialization boundary (not in-memory only)** — N/A — *This fix is a pure in-memory predicate over already-hydrated rows; no serialization boundary to cross. Integration tests drive the real `_apply_recovery_transition`.*
- **5. Plan internal consistency — spike findings match task steps** — PASS — *Freshness Check's discovery of the second guard at line 2222 is reflected in the two-site fix.*
- **7. New public APIs — docstrings present** — PASS — *Helper is module-private but carries a docstring.*
- **8. Breaking changes — migration path documented** — N/A — *No breaking change.*
- **9. Tests added for new behavior** — PASS — *New regression module + fixture fix.*
- **10. Tests cover the failure path (not just happy path)** — PASS — *None/garbage `response_delivered_at`, no-anchor, and stale-prior-run cases all covered.*
- **11. UI changes (if any) — screenshot captured** — N/A — *No UI files in the diff (only `.py`/`.md`).*
- **12. Docs updated for user-facing changes** — PASS — *`session-recovery-mechanisms.md` and `pm-session-liveness.md` both updated.*

### Blockers
- None

### Tech Debt
- None

### Nits
- [ ] **`agent/session_health.py:3150`** — `the health check inspects ``response_delivered_at``. If the field is set,` — The `_agent_session_health_check` docstring still describes the pre-fix bare-presence semantics ("If the field is set, the session already delivered its final response"). This PR changes the guard to fire only when the delivery belongs to the current run's epoch. Update the docstring to mention `response_delivered_at >= (started_at or created_at)` for consistency with the inline comments above the guard blocks and the feature docs, both of which were correctly updated.

### Miscellaneous
- None

- **`current_tool_name` / `last_tool_use_at` read without run-scoping in the tool-timeout sub-loop** — tracked by #2002 (OPEN) — matches the plan's `[SEPARATE-SLUG]` No-Go; correctly filed separately rather than folded into this PR.

### Verification Results
- `pytest tests/unit/test_delivery_guard_resume_epoch.py tests/unit/test_agent_session_queue.py -n0` → 51 passed.
- `ruff check` on the changed files → All checks passed.
- `grep -c _delivery_belongs_to_current_run agent/session_health.py` → 3 (def + 2 call sites).

### Screenshots
- N/A — no UI changes.

valorengels:
## Review (Judge code-quality): Approved

The fix is small, well-scoped, and correct. `_delivery_belongs_to_current_run` (`agent/session_health.py:298-311`) handles every input class through the pre-existing `_ts` normalizer: `None` delivery → False (matches the old negative case), garbage delivery → False without raising (defensive, tested), no anchor → True (preserves the #918 always-fire behavior for legacy rows), inclusive `>=` boundary (tested). Both guard call sites (lines 2243 and 3264) were converted and the fire-path behavior is unchanged from the presence-check version. Comments and the docs narrative are accurate against the verified code (`started_at` re-stamping at `session_pickup.py:463`/`:611`; `resume_session` never clearing `response_delivered_at`). The new test module is genuinely regression-detecting — the delivery guard precedes `_should_kill_no_progress`, so under the old presence-only check the stale-delivery integration test fails on both its assertions — and the `test_agent_session_queue.py` fixture change correctly re-anchors the delivery timestamp relative to `started_at` instead of a fixed 2024 date. All targeted tests pass serially; ruff check/format clean.

### Blockers
- None

### Tech Debt
- None

### Nits
- None

*Judge verdict: APPROVED — blockers: 0, tech_debt: 0, confidence: 0.93*

valorengels:

Traced the new predicate against every writer and reader of the three fields it consumes, all pending→running transition sites, the recovery paths that null `started_at`, the Popoto hydration layer, and both guard call sites. The guard's original #918 purpose is preserved: `response_delivered_at` is only stamped mid-run (`agent/session_executor.py:1504`, `agent/session_completion.py:1079`, both aware-UTC), while `started_at` is stamped at pickup before the run begins and saved atomically with `status="running"` — so a genuinely delivered-then-stuck current run always satisfies `rd >= started_at` and still finalizes, and any redelivery in a resumed run re-arms the guard (no unbounded re-opening of the duplicate-delivery loop). The `started_at = None` requeue (`session_health.py:2579`, startup recovery `:682`/`:761`) only ever accompanies a transition to `pending`, and both guard sites evaluate only `status="running"` rows, so the `created_at` fallback is unreachable for a just-requeued row. Datetime typing is symmetric: `_ts` coerces naive→UTC and floats, and `AgentSession.__setattr__` coerces every `_DATETIME_FIELDS` assignment to aware-UTC on Redis load, so the two comparison sides cannot skew. Two adjacent hazards were verified pre-existing on main and strictly narrowed (not widened) by this change: startup recovery's unguarded requeue of delivered-but-unfinalized sessions, and the once-per-tick stale-snapshot window (the plan's Race Conditions section covers the primary variant). The one behavior change for corrupt rows (unparseable `response_delivered_at` now skips the guard where main's presence check would fire) moves in the safe direction — an unverifiable delivery cannot be duplicated by requeueing.

### Blockers
- None

### Tech Debt
- None

### Nits
- None

*Judge verdict: APPROVED — blockers: 0, tech_debt: 0, confidence: 0.90*

valorengels:
## Review: Approved

Re-review after the patch commit `749ad22a` resolved the single nit from the prior review (the `_agent_session_health_check` docstring still described the pre-fix bare-presence semantics — now rewritten to describe epoch scoping). Multi-judge consensus ran per this repo's REVIEW-stage default: both judges (code-quality, risk) returned APPROVED with zero findings; consensus rule `any-blocker-wins`, k=2, n=2, mean confidence 0.915, no disagreement. The fix is a correct, tightly-scoped epoch predicate (`_delivery_belongs_to_current_run`, `response_delivered_at >= (started_at or created_at)`, legacy no-anchor rows preserve always-fire) applied at both Delivery-guard sites, with thorough regression tests and accurate docs. Plan success-criteria checkboxes were synced (all 7 ticked) and pushed as `df47fbfe` before this review was posted.

Note for deploy: this PR modifies `agent/` (worker health-check path) — requires a worker restart on all machines after merge (`./scripts/valor-service.sh restart`).

## Rubric

- [x] **1. Plan vs. implementation match** — pass — *All 7 success criteria delivered: both guard sites use the shared predicate, `>=` boundary, legacy fallback, #1614 audit note in the PR body, tests, docs.*
- [x] **2. New code quality** — pass — *Helper reuses the existing `_ts` normalizer, typed return, docstring, clear branch logic; no duplication (one predicate, two call sites).*
- [x] **3. Test coverage** — pass — *`test_delivery_guard_resume_epoch.py`: 8 helper unit cases + 2 integration cases against the real `_apply_recovery_transition`; 125 tests pass across the three affected modules (serial run).*
- [x] **4. Regression risk to existing callers** — pass — *Both bare-presence call sites replaced; fire-path behavior (finalize as `completed`, `StatusConflictError` tolerated) byte-identical; the one fixture that would have silently flipped paths was re-anchored to `started_at + 1s`. Risk judge additionally verified all `started_at`-nulling paths transition to `pending`, where neither guard evaluates.*
- [x] **5. Data integrity** — n/a — *Read-site predicate only; no schema/field change, no migration.*
- [x] **6. Security** — pass — *No security surface; no secrets, eval, or subprocess-with-input.*
- [x] **7. Documentation accuracy** — pass — *Prior review's nit resolved: `_agent_session_health_check` docstring now describes the epoch-scoped guard (commit `749ad22a`). Feature docs (`session-recovery-mechanisms.md`, `pm-session-liveness.md`) and inline comments already updated and accurate.*
- [x] **8. PR body accuracy** — pass — *Claims verified. One stale-by-one detail: the body's `grep -c _delivery_belongs_to_current_run → 3` now returns 4 because the docstring fix added a fourth (non-call-site) mention; the substantive claim (1 def + 2 call sites) remains accurate and the plan criterion is `≥ 3`.*
- [x] **9. Disclosed deferrals** — pass — *The `current_tool_name`/`last_tool_use_at` gap is deferred to #2002 (re-verified OPEN), consistent with the plan's `[SEPARATE-SLUG]` No-Go.*
- [x] **10. Follow-up claims verified** — pass — *#2002 resolves to an OPEN GitHub issue.*

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — PASS — *All 7 met; checkboxes synced and pushed (`df47fbfe`).*
- **2. No-Gos from plan — none violated** — PASS — *Predicate extracted (not the whole guard block); field not cleared on resume; new audit gap filed as #2002, not folded in.*
- **3. New `except Exception` blocks — each has logger/raise/swallow-ok** — N/A — *No new except blocks; helper is exception-free via `_ts` (returns False on unparseable input).*
- **4. New integration tests — exercise serialization boundary (not in-memory only)** — N/A — *Pure in-memory predicate over hydrated rows; integration tests drive the real `_apply_recovery_transition`.*
- **5. Plan internal consistency — spike findings match task steps** — PASS — *Freshness Check's discovery of the second guard site is reflected in the two-site fix.*
- **7. New public APIs — docstrings present** — PASS — *Module-private helper carries a docstring.*
- **8. Breaking changes — migration path documented** — N/A — *No breaking change.*
- **9. Tests added for new behavior** — PASS — *New regression module + fixture fix.*
- **10. Tests cover the failure path (not just happy path)** — PASS — *None/garbage `response_delivered_at`, no-anchor, and stale-prior-run cases all covered.*
- **11. UI changes (if any) — screenshot captured** — N/A — *No UI files in the diff (only `.py`/`.md`).*
- **12. Docs updated for user-facing changes** — PASS — *Both feature docs updated; function docstring now consistent.*

### Blockers
- None

### Tech Debt
- None

### Nits
- None

### Miscellaneous
- None

- **`current_tool_name` / `last_tool_use_at` read without run-scoping in the tool-timeout sub-loop** — tracked by #2002 (OPEN) — matches the plan's `[SEPARATE-SLUG]` No-Go; correctly filed separately rather than folded into this PR.

### Review Delta (vs prior review on HEAD f65e47e)
- **Resolved**: Nit — `_agent_session_health_check` docstring described pre-fix bare-presence semantics; fixed by commit `749ad22a` (the only commit since the prior review).
- **New**: None.
- **Unchanged**: None.

### Verification Results
- Plan `## Verification` table (columns are `Criterion|Check`, not runner-compatible `Command|Expected` — checks executed manually):
  - Check 1 (both guards use the predicate): `grep -c` → 4 (≥ 3: def + 2 call sites + docstring mention). PASS.
  - Check 2 (guard not removed): both `True` branches still call `finalize_session(..., "completed", ...)` with logged `StatusConflictError`/`Exception` handling. PASS.
  - Check 3 (epoch boundary inclusive): `return rd >= anchor` at `agent/session_health.py:311`. PASS.
  - Check 4 (legacy preserved): `test_legacy_no_anchor_at_all_returns_true`. PASS.
  - Check 5 (resumed session survives): `test_resumed_session_delivery_before_started_at_returns_false` + `test_resumed_prior_run_delivery_does_not_force_finalize`. PASS.
  - Check 6 (no field cleared on resume): `grep -n response_delivered_at tools/valor_session.py` → no matches. PASS.
- `pytest tests/unit/test_delivery_guard_resume_epoch.py tests/unit/test_agent_session_queue.py tests/unit/test_health_check_recovery_finalization.py -n0` → **125 passed**.
- `ruff check .` → All checks passed; `ruff format --check .` → all files already formatted.

### Multi-Judge Consensus
- Judge code-quality: APPROVED (blockers 0, tech_debt 0, confidence 0.93)
- Judge risk: APPROVED (blockers 0, tech_debt 0, confidence 0.90)
- Consensus: `any-blocker-wins`, k=2, n=2, mean_confidence 0.915, tied=false → **APPROVED**
- Cross-vendor judge: disabled (`SDLC_REVIEW_CROSS_VENDOR` unset).

### Legacy Cruft Audit
- Clean — the old presence checks are fully replaced (no dual implementation), the only fallback chain (`started_at or created_at`) is intentional and documented, no dead imports or stale comments introduced.

### Screenshots
- N/A — no UI changes.

PR #2014, opened 2026-07-11
Fix #1964: recover worker launchctl bootstrap EIO-5 race in /update

### LINKED ISSUE #1964: `test_remote_update.py::test_already_up_to_date` fails with bootstrap I/O error
> **Context:** `scripts/remote-update.sh` is the entry point the `/update` skill and the nightly cron job (`remote-update.sh`) use to bring a machine to the latest `main`, sync dependencies, and restart services (see `CLAUDE.md`'s Update & Restart skill).

## Problem

**Current behavior:** `tests/integration/test_remote_update.py::TestRemoteUpdateScript::test_already_up_to_date` fails on current `main` HEAD (commit `0d3180f2`):
```
assert result.returncode == 0
AssertionError: assert 1 == 0
where 1 = CompletedProcess(args=['bash', '.../scripts/remote-update.sh'], returncode=1,
  stdout='[update] P...renamed to `remark`\nBootstrap failed: 5: Input/output error\nTry re-running the command as root for richer errors.\n').returncode
```
Two distinct signals appear in the output: a "renamed to `remark`" line (possibly an npm/node tool rename notice) followed by "Bootstrap failed: 5: Input/output error". It's unconfirmed whether this is a genuine `scripts/remote-update.sh` regression or an artifact of the specific machine/sandbox this was reproduced on (a sandboxed shell without a real TTY or without root can produce spurious I/O errors on some bootstrap operations).

**Desired outcome:** `test_already_up_to_date` passes reliably in CI and on developer machines, or the test is adjusted if the failure is confirmed environment-specific rather than a script bug.

## Repro

```
bash scripts/remote-update.sh
# or
pytest tests/integration/test_remote_update.py::TestRemoteUpdateScript::test_already_up_to_date -q --tb=short -n0
```

## Solution sketch

Needs investigation before a fix:
- Reproduce on a clean/CI environment (not the sandbox this was found in) to confirm whether the "Bootstrap failed: 5: Input/output error" is universal or environment-specific.
- Identify what specifically raises "Input/output error" (errno 5) in the bootstrap step — likely a subprocess call in `scripts/remote-update.sh` or a module it invokes.
- Investigate the "renamed to `remark`" line — if it references an npm package rename, check whether `scripts/remote-update.sh`'s tooling checks need updating to the new package name.

This issue will be consumed by `/do-plan` to produce a plan document.

valorengels:
## Problem

A `/update` run failed on the machine "Valor the Pirate" with:

```
Bootstrap failed: 5: Input/output error
Try re-running the command...
```

and the run reported as **failed** (non-zero exit), even though the two ⚠️ warnings in the same output (stale `GRANITE_*` env keys, `sms_reader` traceback) are pre-existing non-fatal warnings the pipeline already tolerates.

## Root cause

Traced through the pipeline:

- `handle_update_command` / cron → `scripts/remote-update.sh` → `run.py --cron --no-pull`.
- In cron mode `do_service_restart=False`, and `--no-pull` leaves `result.git_result = None` (hence the `up to date at unknown` summary). So `run.py` never reaches its Step 5 launchctl block and returns 0 with only warnings. **The bootstrap failure does not originate in `run.py`.**
- The git pull was "Already up to date" (`BEFORE_SHA == AFTER_SHA`), so no code-change-gated restart fired.

The only `launchctl bootstrap` that runs **unconditionally** (independent of code changes) is the worker-restart "service not loaded" branch in `remote-update.sh`. When the worker plist is installed but the label isn't shown in `launchctl list` — yet is still registered with launchd (a recoverable transient state, e.g. after a crash/bootout) — a plain `launchctl bootstrap` fails with `Bootstrap failed: 5: Input/output error` (errno 5 = EIO). Two defects followed:

1. The bare `launchctl bootstrap` had no `2>/dev/null`, so launchctl's raw error text leaked straight into the `/update` report.
2. The branch unconditionally set `RESTART_FAILED=1`, so the terminal `exit $(( RESTART_FAILED || VERIFY_FAILED ))` failed the entire `/update` — recurring every 30-minute cron cycle on that machine.

This is a **code-resilience defect** with a machine-local trigger. The `NEED_RESTART` path already learned this exact lesson and uses `kickstart -k` to avoid "bootstrap error 5: label still registered"; the not-loaded branch never got the same treatment.

## Fix

- Not-loaded branch: on `bootstrap` failure, fall back to `launchctl kickstart -k`, which adopts + restarts an already-registered label. Only declare `RESTART FAILED` when **both** fail.
- Suppress the raw launchctl stderr on both the not-loaded bootstrap and the `NEED_RESTART` fallback bootstrap so error 5 can never leak into the report when the fallback recovers it.
- Adds a source-level regression guard (`test_worker_bootstrap_has_kickstart_fallback`) asserting every worker `launchctl bootstrap` suppresses stderr and the region retains a `kickstart -k` fallback.

## Scope notes

- The `GRANITE_*` stale-key warning and the `sms_reader` traceback are **independent, non-fatal warnings** surfaced by `run.py --cron` into `result.warnings`; neither touches the shell exit code or the launchctl path. Confirmed **not connected** to the bootstrap failure — left out of scope.
- Machine remediation for "Valor the Pirate" (why the worker wasn't loaded in the first place) is orthogonal: check `logs/worker.log`, `launchctl print gui/$(id -u)/com.valor.worker`, and `./scripts/valor-service.sh worker-restart`. This PR makes `/update` recover gracefully regardless.

## Test

```
pytest tests/integration/test_remote_update.py::TestRemoteUpdateScript::test_worker_bootstrap_has_kickstart_fallback
```

`bash -n scripts/remote-update.sh` clean.

### DIFF
--- a/scripts/remote-update.sh
+++ b/scripts/remote-update.sh
@@ -252,10 +252,14 @@ PYEOF
                 WORKER_STATE="worker restarted"
                 VERIFY_SINCE=$RESTART_TS
             else
-                # kickstart failed; fall back to bootout + bootstrap with a brief wait
+                # kickstart failed; fall back to bootout + bootstrap with a brief wait.
+                # Suppress the raw launchctl stderr ("Bootstrap failed: 5: Input/output
+                # error") — the scannable `RESTART FAILED` line below is the signal we
+                # classify on, and leaking launchctl's own error text into the /update
+                # report is noise (issue #1964).
                 launchctl bootout "gui/$(id -u)/$WORKER_LABEL" 2>/dev/null || true
                 sleep 2
-                if launchctl bootstrap "gui/$(id -u)" "$WORKER_DST"; then
+                if launchctl bootstrap "gui/$(id -u)" "$WORKER_DST" 2>/dev/null; then
                     WORKER_STATE="worker restarted"
                     VERIFY_SINCE=$RESTART_TS
                 else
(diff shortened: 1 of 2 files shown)


tomcounsell:
Closing as superseded. #1964 was resolved by #1978 (2026-07-09), and main has since independently grown a more robust version of the exact EIO-5 recovery this PR proposes: the first-install path already does bootstrap → `kickstart -k` fallback and *captures* `BOOTSTRAP_ERR` to surface the raw launchd error only when both attempts fail — whereas this branch's `2>/dev/null` discards it. This branch is 71 commits behind and its change is a strict subset of what's already on main.

Issue #1905, opened 2026-07-05
Dead telemetry counter: D0 never-started gate makes no_output_budget_exceeded unreachable


tomcounsell:
**Plan:** https://github.com/tomcounsell/ai/blob/session/sdlc-1905/docs/plans/sdlc-1905.md

## Recon Summary

Firsthand code read at HEAD `63e43118` (`agent/session_health.py`, `_has_progress` sub-check B, lines 1547-1618; `_never_started_past_grace` lines 1343-1402). Verified the two claims below against the actual control flow.

**Confirmed:**
- The D0 never-started gate (`if _never_started_past_grace(entry): return False`, line 1560) fires whenever `sdk_ever_output` is False AND `running_seconds > NEVER_STARTED_GRACE_SECS (120) + NEVER_STARTED_CONFIRM_MARGIN_SECS (30) = 150s`. It uses the SAME `started_ref = started_at or created_at` origin as sub-check B's own `running_seconds` (line 1584), so the two agree on running time.
- The budget-band leg `if running_seconds <= NO_OUTPUT_BUDGET_SECONDS: return True` (line 1593) is UNREACHABLE for `running_seconds > 300`: to reach it, control must pass the D0 gate (needs `running_seconds <= 150` for non-legacy) and the `running_seconds < STARTUP_GRACE_SECONDS (300)` leg (line 1585). A non-legacy session at 300-1800s already returned False at the D0 gate. Only `running_seconds <= 150` (returns True at line 1585) or legacy-None (returns True at line 1580) reach this code. The band leg therefore never returns via its own condition.
- The counter INCR block `{project_key}:session-health:tier1_falloff:no_output_budget_exceeded` (lines 1606-1617) requires `running_seconds > 1800` AND `sdk_ever_output` False, which is a strict subset of the D0-gated population. PROVABLY UNREACHABLE. `test_d0_gate_preempts_budget_exceeded_counter` (line 569, added in #1903) already pins that the counter never fires.

**Revised:**
- The constant `NO_OUTPUT_BUDGET_SECONDS` STAYS: it is still used by the #1614 own-progress heartbeat gate (line 1638) and the Tier-2 reprieve cap. Only the sub-check-B budget leg + INCR are dead.
- The `running_seconds < STARTUP_GRACE_SECONDS` leg (line 1585) STAYS as clock-skew defense-in-depth (returns True for the `running_seconds <= 150` survivors of the D0 gate).

[...]

Issue #1227, opened 2026-04-30
PM session first-turn cold start: 74K-char system prompt + Opus produces 15-20min TTFT


valorengels:
> **Context:** This system runs Claude Code agents inside a custom orchestrator: a parent **PM (Project Manager) session** delegates work to one or more **Dev sessions**. PM sessions launch the `claude` CLI binary in `-p` (programmatic) mode with the entire PM persona + project CLAUDE.md appended via the `--append-system-prompt` flag. See [bridge-worker-architecture.md](docs/features/bridge-worker-architecture.md), [pm-channels.md](docs/features/pm-channels.md), and [harness-abstraction.md](docs/features/harness-abstraction.md).

## Problem

Every newly-spawned PM session pays a 15–20 minute time-to-first-token (TTFT) on its first turn. Three compounding factors:

1. The full persona + project CLAUDE.md (~74,769 characters / ~18,750 tokens) is concatenated and passed as a CLI argument to `claude` via `--append-system-prompt`.
2. PM sessions default to `--model opus` (slowest model).
3. There is **no prompt-caching infrastructure** anywhere in the codebase — `cache_control` / `ephemeral` markers are never set, no cache-hit metrics are collected, and each new `claude -p` invocation pays full cold-start cost even if a previous session just processed an identical prefix.

**Current behavior:**

Worker logs from session `0_1777540282099` (observed 2026-04-30 in real time):

```
[harness] Using --model opus for session_id=0_1777540282099
[harness] Appending 74769-char system prompt for session_id=0_1777540282099
SDK heartbeat: running 60s, communicated=False
SDK heartbeat: running 120s, communicated=False
…
SDK heartbeat: running 840s, communicated=False
```

`ps` confirmed PID 2966 alive (state `S`, not blocked) for 16+ minutes with the full system prompt on its argv, simply waiting for the first API token.

Identical pattern reproduced from Apr 28 logs (session `tg_valor_-5051653062_9413`): `communicated=False` for 1020 seconds, worker restart, then `--resume` completed the same task in 59 seconds. The first turn is the bottleneck; once the session has API-side cache, subsequent turns are fast.

**Desired outcome:**

A new PM session produces its first observable output (tool call or token) in **under 90 seconds** on a typical `claude -p` invocation. The 15–20 minute cold-start regime is eliminated.

## Definitions

| Term | Definition | Reference |
|------|-----------|-----------|
[...]

PR #2070, opened 2026-07-13
Widen never-started grace to ~20min + evidence-based subprocess-hang probe (#2069)

### LINKED ISSUE #2069: SDLC cold-start wedge: never-started grace kills slow Opus starts at 150s; no short-term hang signal
## Problem

SDLC dev/eng turns wedge systematically during cold start. Across current worker logs: **130** \`never_started past grace\` recoveries, **72** sessions \`Finalized as failed after 2 recovery attempts\`, **30** \`zombie_uuid_no_output\` events. Failed sessions repeatedly show the same signature: the \`claude -p\` subprocess is spawned and assigned a UUID but produces **zero streamed output** (\`sdk_ever_output=False\`) for the full grace window, gets recovered as \`no_progress\`, retries into the identical wedge, and finalizes \`failed\`. From the PM's side this reads as "the dev launch was interrupted before it started."

## Root cause

Two independent timers disagree by ~8x:

- Runner per-turn backstop: **7200s** (\`ENG_TURN_TIMEOUT_S\`)
- Health monitor never-started grace: **150s** (\`NEVER_STARTED_GRACE_SECS 120 + margin 30\`)

An Opus SDLC turn's cold-start-to-first-token runs **15-20 min** (issue #1227: 74K-char system prompt + MCP-fleet boot), but the never-started D0 gate fires at 150s — before \`system/init\` is even emitted. Each recovery re-spawns a fresh \`claude -p\` (more load, stale-UUID resume → full context reload), making the next start slower: a self-reinforcing thrash.

Because output silence cannot be a *timely* death signal inside a 15-20 min cold start, widening the grace alone would blind us to genuine hangs.

## Acceptance criteria

- [ ] Never-started grace widened to ~4x normal expected cold-start (target ~20 min) so a slow-but-alive Opus start is not killed.
- [ ] A short-term, evidence-based hang detector catches genuine hangs (flat CPU + no children + no established API socket) in ~60-90s, independent of model output — so widening the grace does not blind us.
- [ ] Kill only on positive evidence (issue #1172), never on output silence alone; inconclusive probes never declare a hang.
- [ ] Unit coverage for the probe and the widened boundaries.

tomcounsell:
Fixes #2069.

## Problem

SDLC cold-start turns wedge systematically: the health monitor's never-started grace fires at **150s**, but an Opus SDLC turn's cold-start-to-first-token runs **15-20 min** (MCP-fleet boot + 74K-char system prompt, #1227). Each slow start is recovered as `no_progress`, re-spawned (stale-UUID resume → full context reload → slower), and thrashes to `failed`. Current logs: 130 `never_started past grace` recoveries, 72 `failed after 2 recovery attempts`, 30 `zombie_uuid_no_output`.

## Change

**1. Widen the grace (4x normal cold start).** `NEVER_STARTED_GRACE_SECS` 120 → 1200 (~20 min incl. margin), env-tunable. This inverts the old `D0 (150s) < STARTUP_GRACE (300s)` ordering, so:
- 0–300s: heartbeat fast-path
- 300–1230s: evidence-based reprieve (subprocess liveness), not bare heartbeat
- >1230s: recover

The stale invariant comment is corrected; boundary tests reconciled; the #1905 clock-consistency regression test is re-pinned at the predicate level (its old `(150,300)` discriminating band no longer exists).

**2. Short-term hang detector** (`agent/session_runner/liveness.py::subprocess_hang_verdict`). Since output silence can't be a timely death signal inside a 20-min window, a new probe reads the subprocess tree directly — **no model output required**:
- `progressing` — live children, advancing CPU, or an ESTABLISHED outbound HTTPS socket (a model call in flight; CPU legitimately idle during the first-token wait).
- `hung` — alive but flat CPU + no children + (sockets readable and) no HTTPS socket, sustained across `HANG_CONFIRM_SAMPLES` polls (~60-90s at the 30s cadence).
- `unknown` — unreadable/inconclusive; **never** a false hang.

Wired into Fix #3's 30s owned-task loop (in-scope cold starts, ~60-90s catch) and the Tier-2 reprieve gate (orphans). Positive evidence supersedes the count-based #1226 escalation guard. Kill only on positive evidence (#1172).

## Tests

- New `tests/unit/session_runner/test_hang_probe.py` (13 cases): progressing/hung/unknown, API-socket wait, pid-change re-baseline, gone-vs-access-denied, flat-CPU confirmation.
- Reconciled `test_never_started_recovery`, `test_session_stall_classifier`, `test_health_check_recovery_finalization` to the widened boundaries. 157 affected tests green.

## Note

This is a first-draft implementation opened for the SDLC pipeline (critique / test / review / docs) to harden. Docs update pending in this PR.


tomcounsell:
## SDLC pipeline complete (critique + review + test, then patch)

Three review subagents ran against the branch; findings folded into patch `a6c071ac`.

**Design critique — shippable, no blockers.** Empirically settled the key risk: the worker runs as non-root user, and psutil 7.2.2 reads an owned child's sockets without root on this Darwin host, so the `api` gate is reliable and "hung" actually fires (no collapse to the 20-min window).

**Code review — 2 MEDIUM + docs, all fixed:**
- Leak: `_hang_samples` grew one entry per session for the worker's lifetime → now cleared on the owned-task teardown path.
- Scope false-kill: `hung→recover` in `_tier2_reprieve_signal` wasn't gated on `not sdk_ever_output`, so an output-producing session blocked on a non-443 endpoint could be killed mid-call → gated; `_API_REMOTE_PORTS` made env-tunable (`HANG_PROBE_API_PORTS`).
- Stale docs: three comments asserted the pre-widening `D0≤150<300` invariant the change inverts → corrected to 1230s.

**Test audit — wiring gaps closed:** added reprieve-gate wiring tests (confirmed-hang recovers, progressing reprieved past the #1226 cap, unknown→cap fallback, output-producing hung NOT recovered) + probe caller-isolation tests. Also fixed the flagged flakiness hazards (shared `_hang_samples` cleared per-test; `HANG_CONFIRM_SAMPLES` monkeypatched as a symbol). Dual-poller state now keyed by `(session_key, caller)`.

Deferred (non-blocking) → **#2071**: alive-hung-orphan latency verification, post-init hang acceleration, and the Fix#3 owned-task loop integration test.

242 affected tests green; ruff clean.

Issue #2083, opened 2026-07-14
Audit whether popoto 1.8.0 atomic index makes descriptor-pollution scar tissue redundant, remove what's proven dead


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/popoto-descriptor-pollution-audit.md

## Recon Summary

Recon performed at plan time (2026-07-14) by reading the actual recall code, the popoto 1.8.0 `ContextAssembler` source, and the live `valor` memory corpus. Evidence:

**Confirmed:**
- `popoto==1.8.0` is installed (`.venv/bin/python -c "import popoto; print(popoto.__version__)"` → `1.8.0`), matching PR #2081 (merged `6c243ebc`).
- `ContextAssembler.__init__` in popoto 1.8.0 accepts `retrieval_mode: str = 'auto'` (keyword-only). Valid modes: `{"auto","lexical","hybrid","composite"}`. Source: `.venv/.../popoto/recipes/context_assembler.py:856,937`.
- Mode resolution (`context_assembler.py:928-955`): `auto` → `hybrid` **iff** the model has BOTH a `BM25Field` and an `EmbeddingField`; BM25-only → `lexical`; neither → `composite`.
- `Memory` model has both required fields: `bm25 = BM25Field(source="content")` and `embedding = GracefulEmbeddingField(source="content")` (`models/memory.py:157-158`), plus `relevance` (DecayingSortedField) and `confidence` (ConfidenceField). So `retrieval_mode='auto'` resolves to `hybrid` for `Memory`.
- Hybrid pull path = BM25 (lexical) + vector (semantic) fused via RRF before graph propagation (`context_assembler.py:133,796`). Retrieval is exposed via `assembler.assemble(query_cues=..., partition_filters=...)` → `AssemblyResult.records` + `.metadata` (scores); optional `assess_quality=True` attaches a `RetrievalQuality`.
- **Current recall path is NOT pure BM25.** `agent/memory_retrieval.py::retrieve_memories` (called by `tools/memory_search/__init__.py:130-137`) is a **four-signal RRF fusion**: BM25 keyword + temporal decay (relevance) + confidence + embedding cosine (`memory_retrieval.py:273-302`, `rrf_fuse` at :47). The issue's "BM25+RRF" shorthand undersells this — the current path already includes a vector signal. The real question is whether popoto's 2-signal hybrid (BM25+vector, with graph propagation) beats our home-grown 4-signal fusion.
[...]

Issue #1803, opened 2026-06-26
Worker loop crashes on StatusConflictError when a queued session is killed mid-pop


valorengels:
## Summary
`_worker_loop()` (`agent/agent_session_queue.py:1237`) dies with an uncaught `StatusConflictError` when a session is killed in the race window between `_pop_agent_session` (reads `status=pending`) and `transition_status(→running)` (finds `status=killed`). The exception propagates out of the loop coroutine, killing the entire project/chat worker loop — **all other pending sessions for that worker_key are then stranded until the next worker restart.**

## Observed
```
future: <Task finished name='Task-5' coro=<_worker_loop() done, defined at agent/agent_session_queue.py:1237>
  exception=StatusConflictError("Status conflict for session tg_cyndra_-1003900483201_167:
  expected 'pending' on disk, found 'killed' (transition_status to 'running': worker picked up session)")>
```
After this, session 166 (pending, same project) was never processed despite a healthy worker process.

## Root cause
The pop→transition sequence is correctly atomic (transition_status rejects the stale write), but the resulting exception is not caught inside the loop body, so it terminates the loop task instead of skipping the session.

## Fix
Catch `StatusConflictError` (and likely any terminal-status transition rejection) around the transition-to-running in `_worker_loop`, log it, skip the session, and continue the loop.

## Repro
1. Enqueue a pending session for a project-keyed worker.
2. Kill it (`valor-session kill`) in the moment the worker pops it.
3. Worker loop for that project dies; subsequent pending sessions stall.

tomcounsell:
Fixed on main by commit 9e120e2b — `_worker_loop` now catches `StatusConflictError` from the pop, logs it, releases the semaphore slot, and continues instead of crashing the worker loop. Regression test added in `tests/unit/test_worker_persistent.py`. Closing; the commit landed directly without a dedicated PR so the issue was never auto-closed.

PR #2099, opened 2026-07-15
Worker loop survives ModelException when popping a corrupted AgentSession (#2088)

### LINKED ISSUE #2088: Worker loop crashes on ModelException when popping a fully-corrupted AgentSession (same class as #1803, uncaught by its StatusConflictError handler)
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/worker-loop-corrupted-pop-crash.md

> **Context:** The standalone worker runs one async `_worker_loop` coroutine per *worker_key* (a project key or chat id). Each loop pops a `pending` `AgentSession` off Redis and transitions it to `running`. If an *unhandled* exception escapes a loop's body, that entire loop task dies — and every other pending session sharing that worker_key is stranded until the worker restarts or a background sweep intervenes. Issue [#1803](https://github.com/yudame/ai/issues/1803) closed one instance of this class (a `StatusConflictError` race). This issue is a **second instance of the same class, arriving via a different exception type that the #1803 fix does not catch.**

## Problem

**Current behavior:** During worker startup, `_worker_loop` (`agent/agent_session_queue.py`) crashes with an unhandled Popoto `ModelException` while transitioning a *fully-corrupted* `AgentSession` record (all fields `None` except `status="pending"`; `created_at` missing) from `pending` → `running`. The exception propagates out of the loop coroutine and kills that worker_key's entire loop task.

Observed **3 times today** at prior worker restarts. Each time, the affected loop self-healed within ~5 minutes — but via a *separate* mechanism (the periodic session-health sweep), **not** via the pop-path exception handling. So the crash still fires every time a corrupted pending record is popped; it just gets cleaned up after the fact. The worker was confirmed healthy afterward (6 sessions running, no wedge detected).

**Why it matters:** This is latent breakage in the exact failure class #1803 was meant to close. While a corrupted record sits at the head of the queue, any co-tenant pending sessions for that worker_key are stranded until the sweep (~5 min) or the next restart. It is currently non-blocking only because corrupted records are rare and the sweep is prompt — the design hole (an uncaught exception can kill a whole loop) remains open.

**Desired outcome:** A corrupted-record transition failure is handled inside the loop the same way a `StatusConflictError` is: logged, the session skipped, the loop continues. No exception type raised while popping/transitioning a single session should ever be able to terminate the loop task.

## Definitions

| Term | Definition |
|------|-----------|
| `_worker_loop` | The per-worker_key async coroutine in `agent/agent_session_queue.py` that pops pending `AgentSession`s and drives them to `running`. One uncaught exception kills the whole loop. |
| worker_key | The partition key for a loop — a project key or a chat id. All pending sessions under one key share one loop. |
| `StatusConflictError` | Repo-defined exception (`models/session_lifecycle.py`) raised when `transition_status` finds an unexpected on-disk status (e.g. a session killed mid-pop). Caught by the #1803 fix. |
| `ModelException` | Popoto's base model exception (`popoto.exceptions.ModelException`). **Not** a subclass of `StatusConflictError`, so the #1803 catch clause does not cover it. |
| corrupted record | An `AgentSession` hash whose fields are all `None` except `status`, typically missing `created_at` — reads/transitions raise instead of returning clean data. |
| session-health sweep | A periodic background reconciliation (~5 min cadence) that heals stalled/corrupted state. The mechanism that *currently* masks this crash — distinct from the pop-path handling. |
| `cleanup_corrupted_agent_sessions` | Existing corrupted-record reaper (`agent/agent_session_queue.py:94`, invoked at `worker/__main__.py:709` during startup). The sanctioned, ORM-only path for removing corrupted records. |

## Root cause (confirmed in code)

The pop path has exactly one typed handler, then a re-raising catch-all:

- `agent/agent_session_queue.py:1700` — `session = await _pop_agent_session(worker_key, is_project_keyed)`
- `:1701` — `except StatusConflictError as e:` → skip-and-`continue` (the #1803 fix, with bounded escalation).
- `:1796` — `except BaseException:` → release the slot, then **`raise`**.

`_pop_agent_session` calls `transition_status(→running)` internally. On a corrupted record that call raises `ModelException`, which is **not** a `StatusConflictError`, so it bypasses the `:1701` handler and falls through to the `:1796` re-raise — terminating the loop task. Same structural hole as #1803, one exception type wider.

## Solution sketch (direction, not a prescription — `/do-plan` to finalize)

Open questions the plan should resolve:

- **Broaden the pop-path catch.** Should the loop treat *any* transition/validation failure on a single popped session (`ModelException` and any terminal-status/validation rejection) the way it treats `StatusConflictError` — log, skip, continue — so no single-session failure can ever propagate to the `BaseException` re-raise? Consider whether to catch `ModelException` explicitly alongside `StatusConflictError`, or restructure so the catch-all no longer re-raises single-session errors.
- **Route corrupted records to the existing reaper.** A fully-corrupted record should be handed to `cleanup_corrupted_agent_sessions` (the startup-time reaper) rather than retried indefinitely at the head of the queue. What is the right trigger point — inline on detection, or flag-and-let-the-sweep-handle-it?
- **Deletion must be ORM-only.** Hand-deleting the corrupted Redis record was attempted during triage and hit inconsistent index-vs-hash-key behavior in Popoto on live data. Any cleanup must go through the ORM / `cleanup_corrupted_agent_sessions`, never raw Redis (see repo rule on never using raw Redis against Popoto-managed keys).
- **Guard against escalation-counter interplay.** The #1803 fix keyed a bounded-escalation counter off `StatusConflictError.session_id`. A corrupted record may have no usable `session_id`; the plan should confirm the broadened handler degrades safely (no `KeyError`, no infinite spin) in that case.

## Acceptance criteria

- [ ] A corrupted `AgentSession` (all fields `None` except `status="pending"`) popped by `_worker_loop` does **not** terminate the loop task.
- [ ] The corrupted record is logged and routed to the existing ORM-based cleanup path; co-tenant pending sessions for the same worker_key continue to be processed without waiting on the sweep.
- [ ] A regression test reproduces the corrupted-record pop and asserts the loop survives and continues (mirroring the #1803 test approach).
- [ ] No raw-Redis deletion is introduced; cleanup goes through `cleanup_corrupted_agent_sessions` / the ORM.

## Related

- Issue [#1803](https://github.com/yudame/ai/issues/1803) — the `StatusConflictError` predecessor in the same failure class (the fix this issue extends).

---

*This issue will be consumed by `/do-plan` to produce a plan document at `docs/plans/{slug}.md`, then executed by `/do-build`.*

valorengels:
Closes #2088

## Summary

`_worker_loop` (`agent/agent_session_queue.py`) now survives a Popoto `ModelException` raised while transitioning a fully-corrupted `AgentSession` (all fields `None` except `status="pending"`) from `pending → running`. This is the second instance of the #1803 class: an uncaught single-session exception on the pop path killed the whole worker_key loop task, stranding co-tenant pending sessions until the ~5-min session-health sweep. Observed 5x in production (Sentry VALOR-E5).

The fix mirrors the #1803 `StatusConflictError` handler, one exception type wider — scoped to `ModelException` (the base of Popoto's save/transition family: `KeyMutationError`/`SkipSaveException` subclass it), NOT a blanket `except Exception`, so genuine logic bugs still crash loudly.

## What changed

- **New `except ModelException as e:` clause** immediately after the `except StatusConflictError` clause: log a warning (no `session_id` dereference — a corrupted record may have none), call `cleanup_corrupted_agent_sessions()` best-effort (**return value deliberately ignored** — the periodic session-health sweep is the authoritative backstop; interpreting the reaper's return was the source of four failed critique rounds), release the slot **before** the backoff `await`, apply a bounded `worker_key`-keyed spin guard with a one-shot escalation `logger.error`, back off, and `continue`.
- The reaper call is wrapped in `try/except Exception` so a reaper failure cannot escape the clause and re-kill the loop (sibling `except` clauses don't catch each other).
- Loop-local guard state `_corrupted_pop_count` / `_corrupted_pop_escalated`, reset at the successful-pop convergence point. New tunable constants `CORRUPTED_POP_ESCALATE_N` / `CORRUPTED_POP_BACKOFF_SECONDS`.
- The #1803 `StatusConflictError` handler and the `session_health.py` reaper are untouched. No raw-Redis deletion introduced.

## Tests

Added to `tests/unit/test_worker_persistent.py` (14 pass):
- `test_model_exception_during_pop_does_not_crash_loop`
- `test_reaper_failure_during_corrupted_pop_does_not_crash_loop`
- `test_repeated_corrupted_pop_escalates_without_crash`
- `test_corrupted_pop_guard_resets_on_successful_pop`

Existing `test_status_conflict_during_pop_does_not_crash_loop` unchanged (regression guard).

## Docs

Added a "Pop-Loop Exception Resilience" section to `docs/features/agent-session-queue.md`, plus inline comments per the plan.

## Coupling note

If the descriptor-pollution audit (#2083) lands first and changes which exception a corrupted `save()` raises, the `except ModelException` clause may need re-confirmation — the new unit tests pin the behavior and will fail loudly if so.

### DIFF
--- a/agent/agent_session_queue.py
+++ b/agent/agent_session_queue.py
@@ -28,6 +28,8 @@ import time
 from datetime import UTC, datetime, timedelta
 from pathlib import Path
 
+from popoto.exceptions import ModelException
+
 # Shared mutable session-tracking state — re-exported here for backward compatibility.
 import agent.session_state as _session_state  # noqa: F401 (also used for mutation sites)
 from agent.branch_manager import get_branch_state  # noqa: F401
@@ -1649,6 +1651,21 @@ def _session_progress_ts(session: AgentSession, acquired_at: float) -> float:
 CONFLICT_ESCALATION_PRIMARY_N = 3
 CONFLICT_ESCALATION_LAST_RESORT_N = 6
 
+# === Bounded corrupted-pop spin guard (#2088, sibling of Defect A above) ===
+# A corrupted AgentSession (all fields None except status="pending") raises a
+# Popoto ModelException during the pending→running transition in the pop path.
+# The loop must survive it the same way it survives StatusConflictError. These
+# thresholds gate the loop-local spin guard in _worker_loop, keyed by
+# worker_key — coarser than #1803's session_id keying BY NECESSITY: a corrupted
+# ModelException carries no session_id, and a fully-corrupted record may have
+# none. CORRUPTED_POP_ESCALATE_N is the number of consecutive corrupted pops
+# before the one-shot logger.error fires; CORRUPTED_POP_BACKOFF_SECONDS is the
+# asyncio.sleep backoff applied after each corrupted pop to prevent a hot
+# re-pop spin against a record stuck at the queue head. Grain of salt: these
+# are provisional/tunable defaults, not load-bearing constants.
+CORRUPTED_POP_ESCALATE_N = 3
+CORRUPTED_POP_BACKOFF_SECONDS = 2.0
+
 
 async def _worker_loop(
     worker_key: str, event: asyncio.Event, is_project_keyed: bool = False
@@ -1675,6 +1692,16 @@ async def _worker_loop(
     _conflict_counts: dict[str, int] = {}
     _conflict_escalated: set[str] = set()
     _conflict_last_resort: set[str] = set()
+
+    # Loop-local corrupted-pop spin-guard state (#2088), keyed by worker_key
+    # (NOT session_id — a corrupted ModelException carries no session_id).
+    # These MUST be loop-local, not module-level: a bridge-mode _worker_loop
+    # exits when the queue drains and is respawned, and loop-local scope resets
+    # the guard cleanly on every (re)start — a module-level latch would survive
+    # that restart and silently suppress the one-shot escalation for a genuinely
+    # new incident (and leak one entry per worker_key). Mirrors _conflict_counts.
+    _corrupted_pop_count: dict[str, int] = {}
+    _corrupted_pop_escalated: set[str] = set()
 
     try:
         while True:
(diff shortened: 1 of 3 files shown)


valorengels:
## Review: APPROVED

Reviewed PR #2099 (`session/worker-loop-corrupted-pop-crash`, Closes #2088) against `docs/plans/worker-loop-corrupted-pop-crash.md` at head `32e51ab`.

**Verdict: APPROVED — 0 blockers, 0 tech debt, 0 nits.**

### What the change does
Adds an `except ModelException as e:` clause to `_worker_loop` (`agent/agent_session_queue.py`), placed as a sibling between the #1803 `except StatusConflictError` (L1728) and the terminal `except BaseException` (L1899) on the primary pop try-block. A fully-corrupted `AgentSession` (all fields `None` except `status="pending"`) that raises a Popoto `ModelException` on the `pending → running` save is now logged, best-effort-reaped, and skipped instead of killing the whole worker_key loop task.

### Rubric — all pass
- **Correctness**: Handler ordering is correct (sibling clauses, before `except BaseException`). Slot released *before* the backoff `await`, honoring the release-before-await invariant. Loop-local guard state (`_corrupted_pop_count` / `_corrupted_pop_escalated`) reset at the successful-pop convergence point.
- **Catch altitude**: Scoped to `ModelException` (base of Popoto's save/transition family), not a blanket `except Exception` — genuine logic bugs and `KeyboardInterrupt`/`CancelledError` still crash loudly.
- **Error handling**: Reaper wrapped in `try/except Exception`; a reaper failure degrades to a no-op rather than escaping the clause (sibling `except` does not catch it). Return value deliberately not interpreted — the source of four prior critique-round failures, removed at the root.
- **Tests**: 4 new regression tests (survival, reaper-failure, one-shot escalation, guard-reset); full file 14 passed.
- **Quality**: Named tunable constants with grain-of-salt provisional comments; thorough inline rationale keyed to worker_key vs session_id.
- **Docs**: "Pop-Loop Exception Resilience" section added to `docs/features/agent-session-queue.md`.

### Verification table — 12/12 pass
All grep-based checks pass; reaper return value confirmed not assigned/branched; no raw-Redis deletion introduced; `pytest tests/unit/test_worker_persistent.py` exit 0 (14 passed); `ruff check` and `ruff format --check` both clean.

### Plan compliance
All success criteria satisfied. No-Gos respected (secondary/fallback pop hardening remains a documented Open Question, primary-only scope is acceptable — those sites are unreachable in the corrupted-record path). All four required plan sections present.

Clean, well-tested, minimal defensive fix that faithfully mirrors the established #1803 pattern. Ready to merge.

_(Posted as a comment: PR is self-authored, so `gh pr review --approve` is unavailable. The APPROVED verdict is recorded in the SDLC substrate.)_

PR #2134, opened 2026-07-17
Decode ContentField values for query-loaded rows (#2112)

### LINKED ISSUE #2112: DocumentChunk.search() returns raw $CF: reference instead of decoded content for query-loaded rows
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/query-loaded-content-decode.md

## Observation (unverified — surfaced during #2085 / PR #2111)

For a popoto model instance loaded via `.query.get()` / `.filter()` / `.all()`, accessing a `ContentField` attribute (e.g. `KnowledgeDocument.content`, `DocumentChunk.content`) returns the **raw `$CF:{hash}:{relpath}` reference string**, not the decoded text. popoto's `Model.__getattribute__` appears to bypass `ContentField.__get__` (the descriptor that would lazy-load + decode) for query-loaded instances.

Verified empirically during #2085:
```python
d = KnowledgeDocument.query.get(doc_id=did)
d.content  # -> '$CF:303a37466d...'  (reference, NOT the document text)
```

## Impact

`models/document_chunk.py::DocumentChunk.search()` builds each result with `"chunk_text": chunk.content or ""` where `chunk` comes from `cls.query.get(chunk_id=...)`. If the bypass holds there too, chunk search results carry the `$CF:` reference string instead of the chunk text — a real retrieval-quality gap for any consumer that reads `chunk_text`.

The #2085 repair helper (`rechunk_zero_chunk_documents`) already works around this by decoding via the field store (`field.store.load(ref).decode()`); `search()` does not.

## Scope note

Pre-existing behavior, unrelated to the #2085 filename-overflow fix (correctly scoped out of PR #2111). Filing for a focused investigation:
1. Confirm whether `search()`'s returned `chunk_text` is actually a reference in practice (write a quick repro that saves a chunk and searches for it).
2. If so, decode via the field store in `search()` (mirror the rechunk helper), and add a regression test asserting `chunk_text` is decoded text, not a `$CF:` string.
3. Consider whether the fix belongs upstream in popoto's `Model.__getattribute__`/`ContentField` interplay instead of per-call workarounds.

## Recon Summary

**Confirmed:**
- Root cause verified in installed popoto (site-packages, `popoto/models/base.py` `Model.__getattribute__`, lazy path at the `_lazy_fields` branch): for query-loaded rows (`.get()`/`.filter()`/`.all()`), field access decodes the raw msgpack value from Redis and returns it directly. For a `ContentField` the stored Redis value IS the `$CF:{hash}:{relpath}` reference, so the descriptor decode is bypassed and callers see the reference string.
- `models/document_chunk.py:127` — `DocumentChunk.search()` builds `chunk_text` from `chunk.content` on a `cls.query.get()` row → returns raw reference. Still present on main.
- Proven decode pattern exists at `tools/knowledge/indexer.py::rechunk_zero_chunk_documents` (detect `$CF:` prefix, `field.store.load(ref).decode("utf-8")`, FileNotFoundError guard).

**Additional raw consumers found (sweep):**
- `tools/knowledge/indexer.py:329` — `_create_companion_memories(..., doc.content or "")`: when `safe_upsert` takes the unchanged-skip path (`models/knowledge_document.py:99-101`) it returns a query-loaded doc, so companion memories can receive the raw `$CF:` reference as content.
- `tools/doctor.py` `_check_knowledge_zero_chunk_documents` (~line 556) — truthiness check on `doc.content` operates on the raw reference; a doc whose content file is missing still counts as "has content".

**Dropped/notes:**
- `DocumentChunk.search()` currently has no non-test production callers (grep) — fix is still warranted as the model is the retrieval surface.
- `KnowledgeDocument` has no `search()` method; no other query-loaded `.content` consumers found outside the three sites above plus the already-fixed rechunk helper.
- Fix belongs at this repo's seam (shared decode helper in `models/`), not a popoto fork.

valorengels:
## Summary

Fixes the ContentField read-path bug: popoto's lazy-field path (`Model.__getattribute__` for `query.get()/.filter()/.all()` rows) returns the raw `$CF:{hash}:{relpath}` reference string instead of decoded content — the descriptor's store load never runs for query-loaded rows.

Closes #2112

## Changes

- **`models/content_decode.py`** (new): `decoded_content(instance) -> str` — the single decode seam. Detects `$CF:` references and loads real bytes via the model's own content-field store (`type(instance)._meta.fields["content"].store`); catches broad `Exception` (dangling ref, malformed ref, decode error) → `logger.warning` + `""`; non-reference values pass through unchanged, so it is safe to call unconditionally.
- **`models/document_chunk.py`**: `search()` now returns decoded `chunk_text` (was returning `$CF:` references).
- **`tools/knowledge/indexer.py`**: `index_file` passes decoded text to `_create_companion_memories` (fixes the `safe_upsert` unchanged-skip path); `rechunk_zero_chunk_documents`'s inline decode block deleted in favor of the shared helper.
- **`tools/doctor.py`**: zero-chunk check's non-trivial-content gate uses decoded content; a doc with a dangling reference decodes to `""` and is skipped (rechunk cannot repair such docs anyway — the helper's warning still surfaces them).
- **Tests**: `tests/unit/test_content_decode.py` (new — canary round-trip on a real query-loaded row, parameterized passthrough, missing-file and malformed-ref failure paths) plus additive search-decode tests in `tests/unit/test_document_chunk.py`. 15/15 passing.
- **Docs**: "Read path: query-loaded rows" section in `docs/features/length-safe-content-store.md`; README index one-liner updated.

## Verification

- `scripts/pytest-clean.sh tests/unit/test_content_decode.py tests/unit/test_document_chunk.py -q` → 15 passed
- `ruff check` / `ruff format --check` clean on all changed files
- `grep -c 'startswith("\$CF:")' tools/knowledge/indexer.py` → 0 (decode logic exists exactly once)
- Canary test pins popoto's lazy bypass — if a future popoto release fixes it, the test fails loudly and the helper can be retired deliberately.

Plan: `docs/plans/query-loaded-content-decode.md` (critique verdict: READY TO BUILD, both CONCERNs addressed — broad exception isolation boundary, builder agent type).


valorengels:

**Blockers:** 0 | **Tech debt:** 0 | **Confidence:** 0.9

- Helper (`models/content_decode.py::decoded_content`) is correct: store resolution via `_meta.fields["content"].store`, `$CF:` detection, passthrough contract, broad-exception isolation returning "" with `logger.warning` — exactly the critique implementation note.
- All four consumers wired (search chunk_text, index_file companion memories, doctor gate, rechunk refactor); the inline decode block in `rechunk_zero_chunk_documents` is fully deleted (grep: 0 matches for `startswith("$CF:")` in the indexer) — decode logic exists exactly once.
- Tests: canary round-trip pins the upstream popoto bypass on a real query-loaded row; passthrough (None/empty/plain) and both failure paths (dangling reference, malformed reference) covered; ORM-only cleanup under a `test-` project key.
- Ruff check + format clean; docstrings and feature-doc updates thorough.
- NIT: field name `content` is hardcoded in the helper — acceptable and documented (both ContentField models use that name).

valorengels:

**Blockers:** 0 | **Tech debt:** 0 | **Confidence:** 0.9

- No behavioral regressions: search result shape unchanged; companion memories fixed on the `safe_upsert` unchanged-skip path and byte-identical on the fresh path; doctor's dangling-reference skip is deliberate and documented (plan Risk 1); rechunk semantics preserved (missing file → "" → skipped).
- Failure isolation: helper swallows ANY decode exception with a warning, so a single corrupted record cannot abort the doctor per-doc scan or `index_file` — the exact concern the critique raised, addressed.
- No popoto fork/monkeypatch; no raw Redis on Popoto-managed keys (the failure-path test deletes content files on the filesystem, not Redis keys); read-path only — no schema/key/reference or env changes, so no migration entry required.
- Performance: per-scored-row file I/O in search is bounded and accepted per plan Risk 2.
- NIT: all scored rows decode before the top-k cut; a post-sort decode would touch fewer files — not worth the restructure at Small appetite.

valorengels:
## Review: APPROVED

**Consensus:** any-blocker-wins over 2 judges (code-quality, risk) — both APPROVED, 0 blockers, 0 tech debt, mean confidence 0.9.

**Plan:** docs/plans/query-loaded-content-decode.md (issue #2112) — all verification-table checks pass:
- Helper unit tests: 6/6 (full narrow set 30/30 via scripts/pytest-clean.sh)
- Chunk model tests: pass
- Ruff check + format: clean on all changed files
- `decoded_content` wired in `models/document_chunk.py` (3 refs)
- Rechunk inline decode gone (0 matches)

**Gates:** documentation gate pass (docs/features/length-safe-content-store.md read-path section + README index updated); no co-author lines; no new env vars or migrations needed (read-path only); no bridge/worker restart required (knowledge indexer/doctor/model layer only, loaded per-invocation).

REVIEW_CONTEXT head_sha=b2bbf5086521a7128467863cfdfe0184920de53f

valorengels:
## Landing review verdict: APPROVE

Rebased onto current main (57 commits; clean merge, no conflicts). Reviewed inline against issue #2112.

**Fix correctness (confirmed):** The root cause is popoto's lazy-field path (`Model.__getattribute__` on `.get()`/`.filter()`/`.all()` rows) returning the raw `$CF:{hash}:{relpath}` reference instead of routing through `ContentField.__get__` → `store.load()`. The PR introduces a single shared seam `models.content_decode.decoded_content()` that detects the `$CF:` prefix and loads real bytes via `type(instance)._meta.fields['content'].store`, swallowing decode failures (dangling/corrupt refs) to `""` so per-record scan loops never abort.

**Completeness:** All four consumers updated consistently — `DocumentChunk.search()` (the reported symptom), `indexer.index_file` companion-memory path, `indexer.rechunk_zero_chunk_documents` (replaces prior ad-hoc inline decode), and `doctor._check_knowledge_zero_chunk_documents`. No remaining raw `.content` reads that could leak a reference in these paths.

**Tests:** `test_content_decode.py` includes a canary asserting the raw query-loaded `.content` still starts with `$CF:` (fails loudly if popoto fixes the bypass upstream), plus passthrough and error cases. Uses real Redis via ORM, ORM-only cleanup.

**Local test results (targeted):**
- test_content_decode.py + test_document_chunk.py: 15 passed
- test_knowledge_indexer.py + test_doctor.py + test_knowledge_document.py: 83 passed

No regressions, no legacy leftovers. Ready to merge.

PR #2157, opened 2026-07-18
Fix cron /update summary showing 'up to date at unknown'

valorengels:
## Problem

On the cron/Telegram \`/update\` path, the final summary reads \`up to date at unknown (N warnings)\` — the literal word \"unknown\" appears where the git HEAD short SHA should be, and the commit count is always 0.

## Root cause

\`scripts/remote-update.sh\` does the \`git pull\` in the shell and calls the Python orchestrator with \`--no-pull\`, which sets \`config.do_git_pull = False\`. With that False, \`result.git_result\` is never assigned and stays \`None\` (run.py Step 1). The summary builder computed the SHA as \`git.get_short_sha(...) if result.git_result else \"unknown\"\`, so the ternary always took the \`else\` branch — even though \`git.get_short_sha()\` is a standalone \`git rev-parse --short HEAD\` that returns the correct SHA regardless of who ran the pull.

## Fix

- Compute the short SHA **unconditionally** via \`git.get_short_sha(project_dir)\`, falling back to \`\"unknown\"\` only if the git call itself raises.
- Removed the \`result.git_result\` truthiness guard on the SHA.
- Left the commit count as \`0\` on the \`--no-pull\` path (the pre-pull SHA lives only in the shell wrapper and is never handed to run.py, so it's genuinely unrecoverable here) — added an explanatory comment rather than inventing a number.

The three by-design advisory warnings (gws auth, sms_reader Full Disk Access, env-completeness) are untouched — out of scope.

## Tests

New \`tests/unit/test_update_cron_summary.py\`:
- \`test_no_pull_summary_shows_real_sha\` — git_result None, summary shows the real SHA, no \"unknown\".
- \`test_no_pull_summary_falls_back_to_unknown_on_git_failure\` — git call raises → \"unknown\".
- \`test_pull_path_reports_updated_with_commit_count\` — pull path with commits reads \"updated to {sha}\".

\`\`\`
3 passed in 0.18s
\`\`\`

\`ruff format\` + \`ruff check\` clean on both changed files.


valorengels:
## Review verdict: APPROVE

Reviewed inline against current main (real tool calls: read merged diff, ran targeted tests, confirmed summary-rendering strings intact).

**Correctness:** The bug was `sha = git.get_short_sha(...) if result.git_result else "unknown"` — on the `--no-pull` cron path (remote-update.sh pulls in the shell, then calls run.py with `--no-pull`), `result.git_result` is always None, so the ternary always took the `"unknown"` branch. Fix computes the SHA unconditionally via the standalone `git rev-parse --short HEAD` (get_short_sha, run.py:2012), falling back to "unknown" only if the git call itself raises. This is the faithful fix: `get_short_sha` returns the correct live HEAD regardless of who ran the pull. The summary now renders `up to date at {real_sha}` (run.py:2028).

**Scope:** Commit count correctly left at 0 on the no-pull path (pre-pull SHA is unrecoverable in run.py) with an explanatory comment — no invented number. The three by-design advisory warnings untouched.

**Tests:** `tests/unit/test_update_cron_summary.py` — 3 passed (real SHA path, git-failure fallback to 'unknown', pull-path 'updated to' count).

**Merge with main:** merged origin/main cleanly (no conflicts). No regression.

No linked issue (no Closes #N). Clean to merge.

Issue #2180, opened 2026-07-20
Nightly detector re-spiked (+15/21) under -n auto on 07-19 despite 07-17 false-alarm fix; fix unconfirmed


valorengels:
> **Context:** The nightly regression detector (`scripts/nightly_regression_tests.py`) runs `pytest tests/unit/ -n auto --json-report`, compares the failure count against the prior run's baseline, and alerts when failures spike. `-n auto` is pytest-xdist parallel execution across all cores. On 07-17 a fix landed that was supposed to resolve a "false alarm" spike; on 07-19 the detector spiked again to +15/21 under the same `-n auto` execution — so the 07-17 fix is unconfirmed.

## Problem

**Current behavior:** The nightly detector spiked to **+15/21** again on 07-19 while running `pytest tests/unit/ -n auto`, despite a fix landing 07-17 that was meant to resolve the false alarm.

**Desired outcome:** Either the detector no longer spikes spuriously under `-n auto`, or — if the failures are real — the 07-17 "false alarm" classification is corrected and the underlying test failures are fixed. Right now we can't tell which, and the 07-17 fix is unverified.

## Why it matters

A recurring spike right after a "fix" means one of two things, both bad:
1. The 07-17 fix didn't address the real cause (the false alarm returns), or
2. There's an xdist-parallelism-specific trigger (`-n auto`) the fix missed — tests that pass serially but collide under parallel workers (shared Redis state, ordering, fixture leakage).

Until we know which, every nightly alert is untrustworthy, and a genuine regression could hide inside the noise.

## Definitions

| Term | Definition |
|------|------------|
| Nightly regression detector | `scripts/nightly_regression_tests.py` — runs `pytest tests/unit/ -n auto --json-report`, diffs failure count vs. prior run, alerts on a spike. |
| `-n auto` | pytest-xdist flag: distribute tests across all available cores in parallel. |
| Baseline | The prior run's recorded failure count the detector compares against (`scripts/nightly_regression_tests.py:266`). |
| +15/21 | The spike magnitude reported on 07-19 (15 new / 21 total failing, per the review notes). |

## Investigate

- Whether the 07-17 fix actually addresses the root cause, or only suppressed a symptom (link the 07-17 fix commit/PR in the plan).
[...]

PR #2195, opened 2026-07-21
Nightly regression detector: run lock, readable alerts, triage dispatch (PR 1/2)

### LINKED ISSUE #2192: Nightly regression detector & Sentry triage reflection: dedupe alerts, human-readable output, auto-triage via AgentSession
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/nightly-regression-triage.md

## Recon Summary

**Confirmed:**
- `scripts/nightly_regression_tests.py` exists; `main()` = run_tests -> reconfirm_serial (#2180) -> compute_new_failures -> send_telegram, with `send_telegram()`/`run_ttft_gate()` already following the best-effort never-crash pattern this issue reuses.
- `reflections/sentry_triage.py::run_sentry_triage` exists; computes the `new_cd_ids` delta, persists `data/sentry_triage_seen.json`, files Class C via the mechanical `_file_github_issue`; `reason` from `_classify_issue` is computed but only short-id + truncated title reach `tg_lines`.
- Non-harness LLM transport exists: `agent/llm/wrapper.py::run_typed(prompt, output_type, *, model=MODEL_FAST, ...)` (PydanticAI) — the correct summarization transport per the two-transport convention, NOT `claude_code_sdk`.
- Eng dispatch entry point: `python -m tools.valor_session create --role eng --message "..."` returns once enqueued (fire-and-forget capable).
- Reflection concurrency guard: `is_reflection_running(state)` = `last_status == "running"` (`agent/reflection_scheduler.py:412`); `mark_started()` sets running at `:479`; in-process `self._running_tasks[entry.name]` at `:721`; single launchd `com.valor.reflection-worker` (KeepAlive) = single process / single event loop.
- Lock idiom precedent for requirement 1: `fcntl.flock(LOCK_EX|LOCK_NB)` in `scripts/pr_shape_cache.py::_acquire_lock` (a `data/*.lock` sibling file).

**Revised:**
- Issue text says `notify_channel_for(POPOTO_REDIS_DB)`; the current signature is `notify_channel_for(client)` (derives the db from the client's connection pool). The underlying db-scoped-channel claim (#2147) still holds. Cosmetic to the plan since the worked example is a triage *reference case*, not a deliverable.

**Pre-requisites:**
- `ANTHROPIC_API_KEY` for the best-effort PydanticAI summarization (feature degrades to raw format if absent).
- `valor-session` CLI importable and `valor-telegram` present (both already installed on every machine).

**Dropped:**
- Root-causing the launchd double-*fire* trigger (needs `sudo log show`, unavailable) — the lockfile neutralizes the duplicate *send* regardless of trigger.
- Class D (investigate) auto-dispatch — deferred to a follow-up once Class C dispatch is proven; recommend Class C only for the first cut.

**Reference case (fix opportunistically if still present):** commit `8e019ab7d` (#2147) added a required `channel` param and db-scoped channel derivation; stale test helpers in `tests/unit/test_agent_session_queue_async.py:473`/`:113` still use the old signature / assert the old literal — the canonical small, hotfix-able, test-only drift from an intentional landed change (requirement 3's reference shape).

## Problem

The nightly detector alerts a human but does three things poorly:

**1. It sometimes sends the exact same alert twice.** `logs/nightly_tests.log` shows the full run (identical start timestamp, identical completion timestamp down to the second, identical Telegram message text) logged twice back-to-back on 2026-07-17, 2026-07-18, and 2026-07-20 (2026-07-19 also double-started, just produced no alert that night since the run was clean). There is only one launchd job for `com.valor.nightly-tests` — confirmed via `launchctl print gui/$(id -u)/com.valor.nightly-tests` (single job, `state = not running` between fires) and `CronList` (no session-scheduled cron duplicate). System local TZ is UTC+7, so the plist's `Hour=3` firing at `20:00 UTC` in the logs checks out — not a timezone misconfiguration. The literal cause of the double *fire* (a launchd race, or a `/update`-triggered plist reinstall race in `scripts/update/service.py::install_nightly_tests` → `scripts/install_nightly_tests.sh`'s bootout+bootstrap sequence, which the update loop calls unconditionally on every bridge-role `/update` run) is not pinned down — that would need `sudo log show`, which wasn't available this session. Root-causing the trigger is a nice-to-have; the alert duplication itself must go away regardless of what causes it.

**2. The alert is unreadable and non-actionable.** The Telegram message is a raw dump of pytest node IDs, e.g.:

```
Nightly regression: 10 newly-confirmed failure(s) (10 confirmed total): tests/unit/test_agent_session_queue_async.py::TestNotifyHealthcheckWatchdog::test_confirmed_drop_closes_handle_pubsub_and_warns, tests/unit/test_agent_session_queue_async.py::TestNotifyHealthcheckWatchdog::test_handle_with_no_pubsub_yet_does_not_crash_on_drop, ... +5 more. Run: pytest tests/unit/ -n0
```

This tells the reader nothing about severity, blast radius, or likely cause — just illegible dotted-path IDs truncated at 5. A human reading this on Telegram has to go pull up a terminal and run the tests themselves to find out what's actually broken.

**3. Nothing acts on the finding.** The detector's whole job ends at "send a Telegram message." Even when the cause is a small, obviously stale test (see the worked example below), a human has to notice the alert, open a terminal, and fix it by hand. There's no automatic triage step.

## Worked example (this session's investigation)

The 2026-07-20 alert's 10 confirmed failures were investigated directly (not guessed) as a concrete case to ground this issue:

- 5 of the 10 are in `tests/unit/test_agent_session_queue_async.py`, all caused by one prior intentional code change: commit `8e019ab7d` ("Test-suite notify isolation: db-scoped notify channel + live-worker guard", issue #2147).
  - `_notify_healthcheck_watchdog()` gained a required `channel` parameter, but `TestNotifyHealthcheckWatchdog._run_watchdog_ticks` (test helper, `tests/unit/test_agent_session_queue_async.py:473`) still calls it with only the old single-argument signature → `TypeError: _notify_healthcheck_watchdog() missing 1 required positional argument: 'channel'` in 4 tests.
  - `_push_agent_session` now derives the publish channel via `notify_channel_for(POPOTO_REDIS_DB)` (db-scoped: `valor:sessions:new` on db=0, `valor:sessions:new:db{N}` on any test db — see `agent/agent_session_queue.py:818-837`), but `TestPushAgentSessionPublish::test_publish_called_after_enqueue` (`tests/unit/test_agent_session_queue_async.py:113`) still asserts the old unscoped literal `"valor:sessions:new"`. Worse: the test's `mock_redis` never sets `connection_pool.connection_kwargs`, so `notify_channel_for`'s `int(client.connection_pool.connection_kwargs.get("db", 0))` silently gets `MagicMock.__int__`'s default return value of `1` instead of raising — producing `valor:sessions:new:db1` and masking the real bug (the test's mock setup, not production code, is stale).

This is exactly the shape of failure a lightweight auto-triage step should catch and resolve without a human: a two-line test-only fix caused by an already-landed, intentional signature change. By the time this issue is planned/built this specific drift may already be fixed (fix it opportunistically if it's still present — it's a good first real test of requirement 3 below), but it should be treated as the reference case for what "small, hotfix-able" looks like versus what should instead go through `/do-issue` + SDLC.

## Desired outcome

Three changes to `scripts/nightly_regression_tests.py` (and its launchd wiring where relevant):

### 1. De-duplicate sends (defense-in-depth, not root-cause-dependent)

Add a lockfile/mutex around the run — e.g. `fcntl.flock(fd, LOCK_EX | LOCK_NB)` on a lockfile under `data/` (a `data/*.lock` sibling to `data/nightly_tests_last_run.json`). If a second invocation can't acquire the lock, it must log the collision and exit cleanly (exit 0, no test run, no Telegram send) rather than silently doing nothing or crashing. This must never cause a *real* confirmed regression to go unreported — if process A is mid-run holding the lock and finds a genuine regression, process B backing off must not be the one "responsible" for sending it; A's own send path is unaffected by B's early exit.

### 2. Human-readable, actionable alert text

Replace the raw node-ID dump with a short, grouped, characterized summary. Use the pytest JSON report's captured failure info (short assertion/traceback text already available in the `--json-report` output being parsed) to say *why* a group failed, not just *what* failed. This likely wants a small LLM summarization call — per this repo's two-transport convention (`claude -p` harness is for session/agent work only; standalone LLM calls like this use PydanticAI, model-agnostic — see CLAUDE.md and `docs/features/` for the convention, NOT `claude_code_sdk`) — fed the confirmed-failing node IDs grouped by file plus their short tracebacks, producing 2-4 actionable sentences. This call must be best-effort: on any failure (timeout, API error, malformed response) fall back to today's raw-list format rather than blocking or crashing the run — this script must never crash on a non-critical step, matching the existing pattern used for `send_telegram()` and the TTFT gate.

### 3. Auto-triage via AgentSession

When new confirmed regressions are found, dispatch an Eng-role AgentSession to investigate before (or alongside) the Telegram alert — see `python -m tools.valor_session create --role eng --message "..."` and `docs/features/eng-session-architecture.md`. The session's job: determine whether the newly-confirmed failure(s) are a small, self-contained fix (hotfix directly on main, per the existing hotfix-vs-SDLC threshold convention) or something bigger that needs planning (file via the `/do-issue` skill and let the SDLC pipeline pick it up). The Telegram alert should reference that a validation session was dispatched (include its session ID so the human can check status via `valor-session status --id <ID>` without having to chase it down).

**Guardrail — no runaway dispatch:** don't re-dispatch a session for a failure set that's already been dispatched on a prior run. Persist dispatch state (e.g. a hash of the confirmed-failing node-ID set, or the session ID) in `data/nightly_tests_last_run.json` and skip dispatch if the current confirmed-failing set was already dispatched and that session hasn't concluded / the failure set hasn't changed. This prevents nightly re-runs from spawning a fresh session every single night for the same unresolved failure.

## Scope 2: Sentry triage reflection (`reflections/sentry_triage.py`)

> The same three problems apply to this repo's other automated nightly-cadence alert: `reflections/sentry_triage.py::run_sentry_triage`, wired as the `sentry-issue-triage` entry in `config/reflections.yaml` and run by the out-of-process reflection scheduler (`agent/reflection_scheduler.py`, subprocess `python -m reflections`, launchd `com.valor.reflection-worker`). It queries Sentry for unresolved issues org-wide, classifies each into tiers A (noise) / B (transient) / C (actionable) / D (investigate) / E (stale), auto-actions A/B/E and files a raw GitHub issue per C when `SENTRY_TRIAGE_APPLY=1` (dry-run by default), and sends a delta-based Telegram summary (only announces genuinely-new C/D short-ids since the last run, tracked in `data/sentry_triage_seen.json`).

A live example from this session (dry-run, apply gate off):

```
Sentry triage: 438 issues (4 new)
  Noise (A): 54
  Transient (B): 51
  Actionable (C): 107
  Review (D): 98
  Stale (E): 128
Would auto-action: A=0/54 B=0/51 E=0/128
  -> VALOR-50: RuntimeError: TTS down
  -> VALOR-73: Audio upload returned non-absolute URL '/media/podcast/audio
  -> VALOR-53: Episode paused-SLA breach: episode=21 step='Audio Generation
[dry run — no GitHub issues filed]
[dry run — no Sentry state changes]
```

Apply the same three fixes here, adapted to this reflection's shape:

### 1. De-duplicate sends

Unlike the launchd-driven nightly test runner, the reflection scheduler already has a same-reflection concurrency guard (`is_reflection_running(state)` in `agent/reflection_scheduler.py:681-696`, skips a due reflection if its persisted `Reflection` state says it's still running, with a stuck-reset fallback after 2x the schedule interval). This issue does NOT ask for a second lock layered on top — it asks the planner to **verify this guard is actually airtight** for `sentry-issue-triage` specifically (the check-then-act between `is_reflection_running` and the state being marked running is a candidate race under concurrent scheduler ticks) and close any gap found. If it's already solid, say so explicitly in the plan and move on — don't add redundant locking for its own sake.

### 2. Human-readable, actionable alert text

The Telegram summary here is already more structured than the nightly-tests one (grouped tier counts, top-3 Class C short-id + truncated title), but it still doesn't say *why* each issue matters or what to do about it — `VALOR-73: Audio upload returned non-absolute URL '/media/podcast/audio` tells the reader there's a bug, not its likely cause or severity. The full `reason` string classification already computes (see `_classify_issue`, `findings` list) is richer than what reaches Telegram (`tg_lines` only carries short_id + truncated title, not `reason`). At minimum, thread `reason` into the Telegram lines for the top Class C items; consider the same best-effort PydanticAI summarization approach as Scope 1's requirement 2 for turning the Class C pile into 2-4 actionable sentences, with the same "must never block or crash the reflection" fallback discipline the rest of this file already follows (e.g. `_send_telegram_notification`'s own swallow-all-failures pattern).

### 3. Auto-triage via AgentSession

Today, Class C (actionable) issues either get a raw `_file_github_issue()` call (apply-on) or are just listed (dry-run) — there's no investigation step, and the filed issue is a mechanical Sentry-data dump (see `_file_github_issue`, `reflections/sentry_triage.py:316`), not a validated fix or a properly scoped `/do-issue`-quality issue. Apply the same pattern as Scope 1's requirement 3: for newly-surfaced Class C issues (the `new_cd_ids` delta already computed for notification gating — reuse it, don't recompute), dispatch an Eng-role AgentSession to investigate the underlying Sentry issue (stack trace, breadcrumbs, affected project) and either hotfix directly (small, self-contained) or produce a proper `/do-issue`-quality issue instead of the current mechanical `_file_github_issue` dump. Same guardrail: dedup dispatch against `data/sentry_triage_seen.json` (already tracks surfaced short-ids) so the same unresolved Sentry issue doesn't spawn a fresh session every day it remains in the standing backlog.

Open question for the planner: Class D (investigate, 98 in the live example above) is never auto-actioned and never triggers `_file_github_issue` today — should Scope 2's AgentSession dispatch cover Class D too, or only Class C? Recommend Class C only for the first cut (D is explicitly "ambiguous, needs human review" by classification design — auto-dispatching triage sessions against ambiguous issues risks noisy/wrong hotfixes more than it saves human time); revisit D once C-tier dispatch is proven out.

## Definitions

| Term | Definition |
|------|------------|
| Nightly regression detector | `scripts/nightly_regression_tests.py`, launchd job `com.valor.nightly-tests`. Runs unit tests nightly, alerts on newly-confirmed failures. |
| Serial re-confirmation gate | Existing logic (issue #2180) that re-runs failing node IDs with `-n0` to separate real regressions from xdist-parallelism artifacts before alerting. |
| `notify_channel_for` | `agent/agent_session_queue.py:818` — derives the Redis pub/sub channel name from the active Popoto DB (db-scoped in test dbs, canonical name in production), added by issue #2147. |
| Eng-role AgentSession | A `session_type="eng"` session — full permissions, engineer persona, the same session type that handles SDLC pipeline work. See `docs/features/eng-session-architecture.md`. |
| Hotfix-vs-SDLC threshold | Existing repo convention: quick/small fixes land directly on main; significant issues go through the full `/do-sdlc` pipeline (Plan → Critique → Build → Test → Patch → Review → Docs → Merge). |
| Sentry triage reflection | `reflections/sentry_triage.py::run_sentry_triage`, the `sentry-issue-triage` entry in `config/reflections.yaml`. Classifies unresolved Sentry issues into tiers A-E, auto-actions A/B/E, files GitHub issues for C, delta-notifies on new C/D. |
| Reflection scheduler | `agent/reflection_scheduler.py`, out-of-process subprocess (`python -m reflections`, launchd `com.valor.reflection-worker`) that ticks all `config/reflections.yaml` entries on their schedules, with a per-reflection concurrency guard. |

## Solution sketch

Not prescribing exact code structure — the three requirements above are independent-ish and can likely be separate, sequential patches to `nightly_regression_tests.py`'s `main()`:

1. Lockfile acquired at the very top of `main()`, released (or naturally released via process exit) at the end — wrap the existing body, don't restructure it.
2. A new best-effort `summarize_failures(confirmed_failing, report)` helper called right before the existing `msg = f"Nightly regression: ..."` construction, replacing that f-string's node-ID join with the summary (falling back to today's format on any exception).
3. A new best-effort `maybe_dispatch_triage_session(new_failures, prev)` helper called after `new_failures` is computed, gated on the persisted-state dedup check, called before (or in parallel with) `send_telegram`, with its result (session ID or None) folded into the alert text.

Open question for the planner: should the AgentSession dispatch happen synchronously (blocking the nightly script until the session is created, not until it finishes) or fully fire-and-forget? Recommend fire-and-forget (`valor-session create` returns once enqueued, not once complete) so the nightly script's own runtime budget isn't at the mercy of triage-session duration — the existing `PYTEST_TIMEOUT_SECONDS` budget shouldn't need to grow to accommodate this.

## Acceptance criteria

- [ ] A second concurrent invocation of `nightly_regression_tests.py` exits cleanly without re-running tests or re-sending an alert, verified by a unit test that simulates lock contention (e.g. hold the lock in the test process, invoke `main()`, assert no subprocess/telegram calls happen).
- [ ] The Telegram alert for a regression is no longer a raw node-ID dump — it reads as prose a non-engineer could get the gist of, with a safe fallback to the current format if summarization fails.
- [ ] A newly-confirmed regression triggers exactly one Eng-role AgentSession dispatch, and that dispatch is not repeated on a subsequent nightly run for the same unresolved failure set.
- [ ] Existing serial re-confirmation gate (#2180) and TTFT gate behavior are unaffected — no regressions to the parts of the script this issue doesn't touch.
- [ ] Unit tests cover: lock contention no-op, summarization fallback-on-failure, and dedup of session dispatch across two consecutive runs with the same failure set.
- [ ] `sentry-issue-triage`'s existing `is_reflection_running` concurrency guard is confirmed race-free (or hardened if a gap is found) — documented in the plan either way.
- [ ] Class C Sentry issues surface with their classification `reason` (or an LLM-summarized equivalent) in the Telegram alert, not just short-id + truncated title.
- [ ] A newly-surfaced Class C Sentry issue triggers exactly one Eng-role AgentSession dispatch (investigate → hotfix or `/do-issue`-quality issue), deduped against `data/sentry_triage_seen.json` so it isn't re-dispatched every day the issue remains unresolved.

## Downstream context

This issue covers two independent alert pipelines (nightly unit-test regression detector, Sentry issue triage reflection) that should get matching treatment for the same three problems. It will be consumed by `/do-plan` to produce a plan document at `docs/plans/nightly-regression-triage.md` (slug: `nightly-regression-triage`) — the planner may split Scope 1 and Scope 2 into separate build phases or PRs if that reduces review risk, but both should land under this one issue/plan given they're the same three fixes applied to two call sites.

valorengels:
## Summary

PR 1 of 2 under issue #2192. Covers **Scope 1 only** (the nightly regression detector, `scripts/nightly_regression_tests.py`) per the plan's delivery split — Scope 2 (Sentry triage, `reflections/sentry_triage.py`) is a separate follow-up PR that lands after this one merges. This PR does NOT close #2192; the issue stays open until PR 2 merges.

- **Run lock**: `_acquire_run_lock()` (`fcntl.flock(LOCK_EX|LOCK_NB)` on `data/nightly_tests.lock`) acquired at the very top of `main()`. A second concurrent invocation logs the collision and exits 0 without running tests or sending an alert — eliminates duplicate nightly Telegram sends regardless of the launchd double-fire's root cause.
- **Best-effort summarizer**: `summarize_failures()` turns confirmed failing node IDs into 2-4 human-readable sentences via `asyncio.run(run_typed(prompt, FailureSummary, model=MODEL_FAST))` (the non-harness PydanticAI wrapper). Any exception (network, empty-prompt, missing API key) falls back to the current raw node-ID preview with a logged warning — never blocks or crashes the run.
- **Fire-and-forget triage dispatch**: `maybe_dispatch_triage_session()` dispatches exactly one Eng-role AgentSession (`valor-session create --role eng --slug nightly-triage-<hash8> --json`) per newly-confirmed failure set, deduped via a persisted sha256 hash. Mandate is *investigate -> file a `/do-issue`-quality issue*, explicitly not auto-hotfix (deferred per the plan's No-Gos). `--slug` is always passed (guards the plan's critique BLOCKER: a slugless call to `valor-session create` exits 1 silently). Session ID is parsed from `--json` stdout and threaded into the Telegram alert.

## Changes
- `scripts/nightly_regression_tests.py` — `_acquire_run_lock`, `summarize_failures`, `maybe_dispatch_triage_session`, wired into `main()`
- `tests/unit/test_nightly_regression_tests.py` — 16 new tests (`TestRunLock`, `TestSummarizeFailures`, `TestMaybeDispatchTriage`, `TestMainDispatchHashPersistence`) added on top of 28 pre-existing (42 -> 44 total)
- `docs/features/nightly-alert-triage.md` (new) — Scope 1 documentation
- `docs/features/README.md` — index entry

## Manual sign-off: alert readability [MANUAL SIGN-OFF]

Per the plan's Success Criteria, "readable to a non-engineer" cannot be unit-tested and requires explicit human sign-off before merge. Sample `summarize_failures()` output against a real synthetic failure set (run inside `.venv` with a working `ANTHROPIC_API_KEY`):

> "Two tests in test_foo.py are failing nightly: test_alpha is returning HTTP 500 errors instead of the expected 200 status code, and test_beta is timing out after 30 seconds. The failures likely indicate a backend service issue or degraded performance—investigate server errors in the application code and check for slow or hanging requests in the API layer."

**@tomcounsell / Valor: please confirm this prose reads as gist-readable to a non-engineer with a 👍 before merging — this criterion cannot be self-certified by the build.**

(A separate spot-check outside `.venv`, using the ambient system Python's stale `pydantic-ai==0.0.48`, hit the fallback path instead — that's an environment artifact unrelated to this PR's code, and it confirmed the fallback discipline works: broad `except Exception` caught the API mismatch, logged a warning, returned the raw node-ID preview without crashing.)

## Testing
- [x] `pytest tests/unit/test_nightly_regression_tests.py -q` — 44/44 passing (28 pre-existing + 16 added by this PR; see note below on the earlier "28" review comment)
- [x] `ruff check` / `ruff format --check` clean
- [x] #2180 (serial re-confirmation gate) and #1227 (TTFT gate) function bodies verified byte-identical to main — unaffected

## Documentation
- [x] `docs/features/nightly-alert-triage.md` created (Scope 1; notes Scope 2 is a follow-up)
- [x] `docs/features/README.md` index entry added

## Definition of Done
- [x] Built: run lock, summarizer, dispatch implemented and wired
- [x] Tested: all tests passing
- [x] Documented: docs created

Progresses #2192 (does not close — PR 2 covers Scope 2 and closes the issue).

## Note on review comment (test count)

The review flagged the original "42/42 passing" claim as inaccurate, citing "the file has 28 tests." Re-verified directly against this PR's exact head commit (clean worktree, no local diff) at review time: `git show main:tests/unit/test_nightly_regression_tests.py | grep -c 'def test_'` = 28 (main, pre-PR), current branch = 42 (28 + 14 new). `pytest tests/unit/test_nightly_regression_tests.py -q` collected and passed 42/42 at that head sha. The reviewer's local run likely picked up a stale/main checkout rather than this branch's worktree. The original "42/42" claim was accurate; the patch round does not change this number (it is now 44/44 after the two tech-debt/nit fixes added 2 more tests).


valorengels:

No blockers. The three deliverables (run lock, best-effort summarizer, fire-and-forget triage dispatch) are implemented cleanly, match the plan's descoped first cut (investigate-and-file, not auto-hotfix), and carry solid failure-path discipline. Two tech-debt items and two nits should be cleaned up before merge, plus one outstanding **manual sign-off gate** that only a human can clear.

## Rubric

- [x] **1. Plan vs. implementation match** — pass — *run lock, summarizer, and issue-filing dispatch all delivered; `--slug` guard, `asyncio.run`, `--json` session-ID parse all present as the critique required.*
- [ ] **2. New code quality** — fail — *dispatch subprocess uses bare `"python"` instead of `sys.executable`; every other subprocess in this file (lines 119, 179, 251) uses `sys.executable`. See Tech Debt 1.*
- [x] **3. Test coverage** — pass — *14 new tests cover contention no-op, fallback-on-exception, empty input, dispatch-once, dedup, changed-set re-dispatch, subprocess failure, malformed/empty stdout.*
- [x] **4. Regression risk to existing callers** — pass — *additive wrappers; #2180 reconfirm and #1227 TTFT paths untouched; carry-forward of dedup state on clean/baseline runs preserves prior behavior.*
- [x] **5. Data integrity** — pass — *`data/nightly_tests_last_run.json` gains two additive fields read via `.get()`; no migration needed; lock file OS-released on exit.*
- [x] **6. Security** — pass — *no secrets, no injection; subprocess argv is a fixed list, not shell.*
- [x] **7. Documentation accuracy** — pass — *new `docs/features/nightly-alert-triage.md` + index entry + base-doc cross-links; Scope 2 clearly marked not-yet-shipped.*
- [ ] **8. PR body accuracy** — fail — *PR body claims "pytest ... 42/42 passing" but the file has 28 tests (28 passed locally on the PR branch). See Tech Debt 2.*
- [x] **9. Disclosed deferrals** — pass — *Scope 2, auto-hotfix, and Class D are explicit `[SEPARATE-SLUG]`/`[EXTERNAL]` No-Gos in the plan; issue #2192 stays open by design.*
- [x] **10. Follow-up claims verified** — pass — *#2192 confirmed OPEN; PR correctly does not use a closing keyword.*

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — PASS — *all Scope 1 criteria met except the [MANUAL SIGN-OFF] readability gate, which is human-only (see below).*
- **2. No-Gos from plan — none violated** — PASS — *dispatch mandate is investigate-and-file; no auto-hotfix, no auto-merge.*
- **3. New `except Exception` blocks — each has logger/raise/swallow-ok** — PASS — *all three broad catches log a warning and fall back; `_acquire_run_lock` re-raises non-EAGAIN OSErrors.*
- **4. New integration tests — exercise serialization boundary** — N/A — *no serialization boundary; JSON state is covered by existing tests.*
- **5. Plan internal consistency** — PASS — *critique findings all addressed in the revision table.*
- **7. New public APIs — docstrings present** — PASS — *all three helpers documented with the best-effort/never-crash contract.*
- **8. Breaking changes — migration path documented** — N/A — *additive only.*
- **10. Tests cover the failure path** — PASS — *fallback, subprocess failure, malformed/empty stdout, lock contention all tested.*
- **11. UI changes (if any) — screenshot captured** — N/A — *no UI files in diff; visual-proof gate is a no-op.*

### Blockers
- None

### Tech Debt
- [ ] **`scripts/nightly_regression_tests.py:412`** — `"python", "-m", "tools.valor_session", ...` — the dispatch subprocess invokes a bare `"python"` while every other subprocess in this file uses `sys.executable` (lines 119, 179, 251). Production launchd happens to put `.venv/bin` first on PATH so it resolves there, but the documented manual run (`python scripts/nightly_regression_tests.py`, per CLAUDE.md) from any shell where `python` is not the venv interpreter would dispatch with the wrong interpreter — and `-m tools.valor_session` would then fail import, silently degrading to "no dispatch" (caught by the best-effort `except`). This is the same silent-never-dispatches failure class the critique flagged for the missing `--slug`. **Fix:** use `sys.executable` for consistency and robustness.
- [ ] **PR body** — the Testing section claims `pytest tests/unit/test_nightly_regression_tests.py -q — 42/42 passing`, but the file contains 28 tests (28 passed on this branch locally). **Fix:** correct the count to 28/28.

### Nits
- [ ] **`scripts/nightly_regression_tests.py:577` & `:394`** — the `sha256(",".join(sorted(set(confirmed_failing))))` hash is computed once inside `maybe_dispatch_triage_session` and again in `main()` to persist it. Minor DRY: have the dispatcher return/expose the hash, or read it back from state, rather than recomputing the same expression at two call sites (drift risk if one is edited).
- [ ] **`scripts/nightly_regression_tests.py:576-580`** — `main()` overwrites `current["dispatched_hash"]` with the current set's hash unconditionally, even when `maybe_dispatch_triage_session` returned `None` because the dispatch subprocess failed. A transient dispatch failure (worker down, timeout) therefore records the set as "dispatched" and forgoes the single triage attempt for that failure set. This is symmetric with the alert-once model (the same set won't re-alert either), so it is low-impact — but consider only recording the hash when a session ID was actually obtained, so a failed dispatch is retried on the next run.

### Miscellaneous
- None

- **Scope 2 (Sentry triage, `reflections/sentry_triage.py`)** — explicit `[SEPARATE-SLUG]` split; issue #2192 (OPEN) stays open until PR 2. Correctly not closed by this PR.
- **Auto-hotfix-on-main via dispatched session** — `[SEPARATE-SLUG]` No-Go; dispatch mandate is investigate-and-file only. Honored.
- **Class D auto-dispatch** — `[SEPARATE-SLUG]` No-Go; Scope 2 concern, out of this PR entirely.

### Outstanding Manual Gate (human-only, blocks merge on the readability criterion)
The plan's Success Criteria include a **[MANUAL SIGN-OFF]**: the "readable to a non-engineer" alert-prose criterion cannot be unit-tested and requires an explicit human 👍 in the PR thread before merge. The PR body pastes a sample `summarize_failures()` output and requests that sign-off. **As of this review it has not been given.** Per the plan, this PR must not merge on the readability criterion until @tomcounsell / Valor records the 👍. Not a code-review blocker — but a merge precondition the reviewer cannot clear.

### Verification Results
- `pytest tests/unit/test_nightly_regression_tests.py -q` → **28 passed** (exit 0)
- `python -m ruff check` (changed files) → clean
- `python -m ruff format --check` (changed files) → clean
- `grep flock / valor_session / --slug / --json / asyncio.run / run_typed` in `nightly_regression_tests.py` → all present (BLOCKER guards satisfied)
- `grep claude_code_sdk` → absent (two-transport convention honored)

### Screenshots
- N/A — no UI files in diff.

_Note: this is a standalone manual review (no supervised SDLC run / RUN_ID present), so no `sdlc-tool` verdict was recorded — this comment is the verdict. Posted as a comment because the PR is self-authored._

valorengels:
Patched. Addressing review comment #5031755313:

- **Tech Debt 1** (bare `"python"` vs `sys.executable`) — fixed. `maybe_dispatch_triage_session` now invokes `sys.executable`, matching lines 158/218/290.
- **Tech Debt 2** (PR body test count) — investigated, not changed as originally worded. Re-verified against this PR's exact head sha (clean worktree, no local diff): `git show main:tests/unit/test_nightly_regression_tests.py | grep -c 'def test_'` = 28 (pre-PR baseline), and the PR branch's file had 42 (28 + 14 new), with `pytest tests/unit/test_nightly_regression_tests.py -q` collecting and passing 42/42 at that exact commit. The original "42/42" claim was accurate; the "28 tests" in the review looks like it came from a stale/main checkout rather than this branch. PR body updated with a note explaining this and the current (post-patch) 44/44 count.
- **Nit 1** (DRY the sha256 hash) — fixed. `maybe_dispatch_triage_session` now returns `(session_id, current_hash)`; `main()` uses the returned hash instead of recomputing it.
- **Nit 2** (only persist `dispatched_hash` on success) — fixed. `main()` now only overwrites `dispatched_hash`/`dispatched_session_id` when `triage_session_id is not None`; a failed dispatch leaves the carried-forward `prev` state intact so the next nightly run retries.

Added 2 more tests for the above (44/44 total now). `ruff check` / `ruff format --check` both clean.

valorengels:
## Review: Approved

The patch round (`bf2c2db`) cleanly resolves all four findings from the prior review (2 tech-debt, 2 nits), each backed by a new test. No new blockers or tech debt. Scope 1 (nightly detector) is code-complete: run lock, best-effort summarizer, and hash-deduped investigate-and-file triage dispatch, with disciplined never-crash failure paths throughout.

One outstanding item is **not** a code-review finding: the plan's **[MANUAL SIGN-OFF]** readability criterion still requires a human 👍 in this thread before merge (see below).

## Rubric

- [x] **1. Plan vs. implementation match** — pass — *all Scope 1 criteria delivered; `--slug`, `asyncio.run(run_typed)`, `--json` session-ID parse all present. Scope 2 (Sentry) is a `[SEPARATE-SLUG]` deferral, correctly not in this PR.*
- [x] **2. New code quality** — pass — *TD1 fixed: `maybe_dispatch_triage_session` now uses `sys.executable` (line 466), matching every other subprocess in the file. Clean type hints and docstrings.*
- [x] **3. Test coverage** — pass — *44/44 passing (28 pre-existing + 16 added). New tests assert `argv[0] == sys.executable`, failed-dispatch-does-not-overwrite-hash, and successful-dispatch-persists-hash.*
- [x] **4. Regression risk to existing callers** — pass — *additive wrappers; #2180 reconfirm and #1227 TTFT paths untouched; dedup state carried forward on clean/baseline runs.*
- [x] **5. Data integrity** — pass — *two additive `.get()` fields in `nightly_tests_last_run.json`; no migration; flock released by OS on exit.*
- [x] **6. Security** — pass — *no secrets, no injection; subprocess argv is a fixed list, not shell.*
- [x] **7. Documentation accuracy** — pass — *`docs/features/nightly-alert-triage.md` + index entry + base-doc cross-links; Scope 2 marked not-yet-shipped.*
- [x] **8. PR body accuracy** — pass — *TD2 fixed: PR body now states 44/44 (28 pre-existing + 16 added) with an explanation of the earlier "42/42" comment. Verified: `grep -c 'def test_'` = 44 on branch.*
- [x] **9. Disclosed deferrals** — pass — *Scope 2, auto-hotfix, and Class D are explicit `[SEPARATE-SLUG]`/`[EXTERNAL]` No-Gos; #2192 stays OPEN by design.*
- [x] **10. Follow-up claims verified** — pass — *#2192 confirmed OPEN; PR correctly uses no closing keyword.*

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — PASS — *all Scope 1 criteria met except the [MANUAL SIGN-OFF] readability gate, which is human-only.*
- **2. No-Gos from plan — none violated** — PASS — *dispatch mandate is investigate-and-file; no auto-hotfix, no auto-merge.*
- **3. New `except Exception` blocks — each has logger/raise/swallow-ok** — PASS — *all broad catches log a warning and fall back; `_acquire_run_lock` re-raises non-EAGAIN OSErrors.*
- **4. New integration tests — exercise serialization boundary** — N/A — *no serialization boundary; JSON state covered by existing tests.*
- **5. Plan internal consistency** — PASS.
- **7. New public APIs — docstrings present** — PASS — *`maybe_dispatch_triage_session` docstring updated to describe the new `(session_id, current_hash)` tuple contract.*
- **8. Breaking changes — migration path documented** — N/A — *additive only.*
- **10. Tests cover the failure path** — PASS — *fallback, subprocess failure, malformed/empty stdout, lock contention, failed-dispatch-retry all tested.*
- **11. UI changes (if any) — screenshot captured** — N/A — *no UI files; visual-proof gate is a no-op.*

### Verified
- [x] Code correctness
- [x] Test coverage (44/44)
- [x] Plan requirements met (Scope 1)

### Blockers
- None

### Tech Debt
- None

### Nits
- None

### Miscellaneous
- None

- **Scope 2 (Sentry triage, `reflections/sentry_triage.py`)** — explicit `[SEPARATE-SLUG]` split; #2192 (OPEN) stays open until PR 2. Correctly not closed here.
- **Auto-hotfix-on-main via dispatched session** — `[SEPARATE-SLUG]` No-Go; dispatch mandate is investigate-and-file only. Honored.
- **Class D auto-dispatch** — `[SEPARATE-SLUG]` No-Go; Scope 2 concern, out of this PR entirely.

### Review Delta (vs prior review on HEAD 4be27c3)
- **Resolved (TD1):** dispatch subprocess used bare `"python"` → now `sys.executable` (line 466). Test at `test_nightly_regression_tests.py:364` asserts `argv[0] == sys.executable`.
- **Resolved (TD2):** PR-body test count claim corrected to 44/44 with a reconciliation note.
- **Resolved (Nit1):** dedup hash was computed twice → `maybe_dispatch_triage_session` now returns `(session_id, current_hash)`, computed once; `main()` consumes it.
- **Resolved (Nit2):** `main()` no longer records `dispatched_hash` unconditionally — it persists the new hash only when a session ID was actually obtained, so a transient dispatch failure leaves prior dedup state intact and the next nightly run retries. Covered by `test_failed_dispatch_does_not_overwrite_dispatched_hash` and `test_successful_dispatch_persists_new_hash`.
- **Unchanged:** the `[MANUAL SIGN-OFF]` readability gate remains outstanding (human-only; see below).

### Minor observation (advisory, no change requested)
The Nit2 fix keys retry on `session_id is not None`. In the rare case where `valor-session create` succeeds but its stdout is unparseable, the session was created yet `dispatched_hash` is not recorded, so the next run re-dispatches (a duplicate triage session for the same set). This is the direct, intended consequence of the requested "only record on obtained session ID" fix, and `--json` success should always emit valid JSON, so the probability is negligible. Noting for the record; not a finding.

### Verification Results (Scope 1)
- `pytest tests/unit/test_nightly_regression_tests.py -q -n0` → **44 passed** (exit 0)
- `python -m ruff check` (changed files) → clean
- `python -m ruff format --check` (changed files) → 2 files already formatted
- `grep -c 'def test_'` on branch → 44 (PR-body claim accurate)
- flock / valor_session / --slug / --json / asyncio.run / run_typed present; claude_code_sdk absent (two-transport convention honored)

### Screenshots
- N/A — no UI files in diff.

### Outstanding Manual Gate (human-only, blocks merge on the readability criterion)
The plan's Success Criteria include a **[MANUAL SIGN-OFF]**: the "readable to a non-engineer" alert-prose criterion cannot be unit-tested and requires an explicit human 👍 in this thread before merge. The PR body pastes a sample `summarize_failures()` output and requests that sign-off. **As of this review it has not been recorded.** Not a code-review blocker — but a merge precondition the reviewer cannot clear.

_Standalone manual review (no supervised SDLC run / RUN_ID present), so no `sdlc-tool` verdict was recorded — this comment is the verdict. Posted as a comment because the PR is self-authored._

Issue #2179, opened 2026-07-20
Duplicate Telegram delivery: one reply sent twice (~2s apart) via response + pm_direct paths, both null message_id


valorengels:
> **Context:** The Telegram relay (`bridge/telegram_relay.py`) drains queued agent replies and ships them to Telegram. A single logical reply can reach the chat through more than one emit path: the executor's own `response` send (authored as "Valor") and the relay's `pm_direct` send (stored as sender `system`, `message_type="pm_direct"`). The relay already carries dedup machinery specifically to collapse these — see the `#1205-style duplicate` guard at `bridge/telegram_relay.py:855-866`, which records the PM self-send in `recent_sent_drafts` so `bridge.redundancy_filter` can suppress the executor's follow-up. This issue is a case where that dedup did **not** fire.

## Problem

**Current behavior:** On 07-18 in the "Eng: Valor" chat, a single Rich-Messages reply was delivered to the chat **twice, ~2 seconds apart**. One copy went out through the `response` path (executor, "Valor"); the other through the `pm_direct` path (relay, `system`). **Both deliveries carried a null `message_id`.**

**Desired outcome:** One logical reply is delivered exactly once. The `response` and `pm_direct` paths are reconciled by the existing dedup layer even when `message_id` is null.

## Why it matters

Two independent output paths fired for one logical reply and neither produced a real `message_id`. The null `message_id` on both copies is the likely reason dedup missed: the `#1205-style` guard and history-store both branch on `msg_id is not None` (`bridge/telegram_relay.py:868`), and the dedup keys on recorded draft text — so if the executor send never registered a draft (or registered under a different session_id), the relay's `pm_direct` copy ships anyway. User-visible double-posting.

## Definitions

| Term | Definition |
|------|------------|
| `response` path | The executor's own user-facing send, authored as "Valor", via the session's `send_cb`. |
| `pm_direct` path | The relay's direct send of a PM turn, stored with sender `system` and `message_type="pm_direct"` (`bridge/telegram_relay.py:880`). |
| `recent_sent_drafts` / `redundancy_filter` | The dedup mechanism (`bridge.redundancy_filter`) that records a PM self-send so a duplicate follow-up from the executor is suppressed (`telegram_relay.py:855-866`). |
| null `message_id` | A delivery that completed without a Telegram message id recorded — both 07-18 copies had this. |

## Investigate

[...]

Issue #1817, opened 2026-06-29
Resilience: correctness & delivery-integrity hazards (lost steers, double-exec, TUI scrape, silent drops)


valorengels:
> **Context.** One of four resilience workstreams (an internal resilience review). Where the other three target durability, the wedge, and event-loop coupling, this issue collects **correctness & delivery-integrity hazards** surfaced by a red-team sweep for *uncatalogued* failure modes — bugs that silently drop, duplicate, or corrupt work with **no crash required**. All file:line citations are HEAD-of-writing.

### A. Lost / dropped work (no crash needed)
- **A1 — Concurrent steering writes lose human course-corrections.** `AgentSession.queued_steering_messages` (`models/agent_session.py:225,2027-2066`) is a non-atomic read-modify-write; `save(update_fields=…)` HSETs the whole list wholesale (`popoto/models/base.py:1119-1140`). The worker pops using a session instance bound at start (`session_executor.py:995-1086,1528`), **not re-fetched**, so a steer landing after the bind is saved-over with `[]`. The resume path (`tools/valor_session.py:725-728`) uses a full `save()`, also clobbering `status`. **An atomic primitive already exists** (`agent/steering.py` uses `RPUSH`/`LPOP`) — the documented inbox just doesn't use it. → *Fix: route the inbox through the atomic primitive.*
- **A2 — OAuth/resolver failure silently discards customer email permanently.** `resolve_customer` (`bridge/routing.py:1404`, fail-closed) returns `None` on *any* resolver error; `_process_inbound_email` then drops the message and marks it `\Seen` (`email_bridge.py:1182-1189`). If the resolver's `gws`/OAuth token expires, **every inbound customer email is silently dropped** and is irrecoverable. → *Fix: distinguish "resolver unavailable" from "not a customer"; don't `\Seen` on resolver error.*
- **A3 — Permanent IMAP auth failure retries forever, no watchdog.** `email_bridge.py:1490-1496` treats a permanent `IMAP4.error` (revoked app password) like a transient one → 5-min backoff forever; no email watchdog exists. Only signal is a stale `email:last_poll_ts`. → *Fix: classify permanent auth failure; alert.*

### B. Duplicate execution / side-effects
[...]

PR #2298, opened 2026-07-23
Fix catchup/reconciler re-enqueuing already-handled messages (#2204)

### LINKED ISSUE #2204: Catchup/reconciler re-enqueues already-handled messages: dedup TTL (2h) < scan window (24h) + reply-only handled check
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/catchup-rehandles-handled-messages.md

> The Telegram bridge has a message-recovery layer that re-scans chat history for messages missed during downtime: a startup catchup scan (`bridge/catchup.py::scan_for_missed_messages`), a periodic reconciler (`bridge/reconciler.py`, every ~3 min), and an agent-judgment sweep (`valor-catchup`, `bridge/agent_catchup.py`, run as `/update` Step 9). All three re-enqueue any historical message they judge "not yet handled." That judgment is failing: already-handled messages are being re-enqueued, so the system responds again to messages it already answered.

## Problem

**Current behavior:** After a bridge restart (e.g. `/update`), the recovery scans re-enqueue Telegram messages that were already fully handled, producing duplicate sessions and duplicate replies to the human.

**Desired outcome:** Recovery scans never re-enqueue a message that was already dispatched, regardless of how it was answered (text reply, non-reply message, emoji reaction, or a deliberate no-reply judgment) or how long ago.

**Mitigation in place:** commit c366bdb84 added an operator kill switch — `touch data/catchup-disabled` pauses all three recovery scans (startup catchup, reconciler, valor-catchup) without affecting realtime handling. The flag is currently SET on the primary machine, meaning the entire recovery layer is disabled until this bug is fixed. Remove the flag (`rm data/catchup-disabled`) once fixed.

## Root-cause analysis (confirmed by code reading)

The "already handled" decision is a chain of guards, and each has a hole:

1. **Durable dedup TTL (2h) is shorter than the scan window (24h).** `bridge/dedup.py::is_duplicate_message` checks a Redis membership record written on dispatch, with a 2-hour TTL (module docstring, `Meta.ttl`). But the startup catchup uses a dynamic lookback from `last_connected` capped at **24 hours**, and the per-chat cursor (issue #1408) can extend lookback further. Any message handled more than 2 hours before a restart has aged out of dedup and falls through to guard 2.

2. **`_check_if_handled` only recognizes an explicit reply.** `bridge/catchup.py::_check_if_handled` fetches the 10 messages after the candidate and returns True only if one is outgoing AND has `reply_to_msg_id == message.id`. This misses:
   - **Emoji-reaction acks** — the repo's preferred "I heard you" signal (per operator feedback) leaves no reply message at all.
   - **Non-reply responses** — answers sent as plain channel messages without threading.
   - **Busy chats** — a reply that is more than 10 messages later is outside the `limit=10` fetch.
   - **Deliberate no-reply judgments** — messages the drafter/judge decided not to answer look permanently "unhandled."
   - On any exception it returns False ("better to double-process than miss") — a bias that compounds the above.

3. **The claim key is intentionally short-TTL** (`bridge:msgclaim:*`, seconds-scale per #1817), so it only prevents concurrent double-enqueue, not re-enqueue across restarts. Working as designed; listed for completeness.

The reconciler shares the same dedup module, and `valor-catchup`'s LLM judge has the same blind spot for reaction-only/no-reply handling.

## Definitions

| Term | Definition |
|------|------------|
| Startup catchup | `bridge/catchup.py::scan_for_missed_messages` — scans monitored groups on bridge startup for messages missed during downtime |
| Reconciler | `bridge/reconciler.py::reconciler_loop` — periodic in-process scan (~3 min) recovering messages dropped inside a live connection (issue #1408) |
| valor-catchup | `bridge/agent_catchup.py` CLI — LLM-judged sweep for sessioned-but-unanswered messages, invoked as `/update` Step 9 (issue #1709) |
| Durable dedup record | Redis membership entry written by `bridge/dedup.py::record_message_processed` on dispatch; 2h TTL |
| Message claim | Short-TTL atomic `SET NX` key (`bridge/dedup.py::claim_message`) preventing two producers racing on the same message (issue #1817 B1) |
| Kill switch | `data/catchup-disabled` flag file (commit c366bdb84) — pauses all three recovery scans while present |

## Open questions for /do-plan

The fix direction is genuinely open — do not treat any one of these as decided:

- Should the durable dedup TTL simply be ≥ the maximum possible lookback (24h + cursor margin), making guard 1 authoritative and guard 2 unnecessary? That is the smallest fix, but the TTL was presumably kept short for Redis hygiene — verify why before raising it.
- Should "handled" be recorded explicitly at dispatch time in a form the scans can trust (e.g. the per-chat last-processed cursor already written by `record_last_processed`), instead of inferring it from visible replies? A message at or below the per-chat cursor was by definition already dispatched.
- Is `_check_if_handled` worth keeping at all once a durable record is authoritative? (Repo has a no-scar-tissue policy; prefer deleting the heuristic over patching its holes.)
- Does `valor-catchup`'s judge need a "was dispatched, chose not to reply" signal, or does fixing the mechanical layer suffice?

## Acceptance criteria

- [ ] A message handled at any point within the scan window is never re-enqueued by startup catchup, reconciler, or valor-catchup after a restart — including messages answered by emoji reaction only, by non-reply message, or deliberately not answered.
- [ ] A genuinely missed message (no dispatch ever happened) is still recovered.
- [ ] Tests cover the >2h-old-handled-message restart case and the reaction-only-ack case.
- [ ] `data/catchup-disabled` is removed on the primary machine and recovery re-enabled.
- [ ] Docs updated (`docs/features/` recovery/catchup docs mention the kill switch and the fixed dedup contract).

## Downstream

This issue will be consumed by `/do-plan` to produce a plan document at `docs/plans/catchup-rehandles-handled-messages.md`, then executed by `/do-build`.

valorengels:
## Summary

Fixes a bug where the Telegram bridge's message-recovery layer (startup catchup + reconciler) re-enqueues messages that were already fully handled, producing duplicate sessions and duplicate replies to the human after a bridge restart.

**Root cause:** `DedupRecord`'s durable-dedup TTL (2h) was shorter than the startup-catchup scan window (up to ~30d via the cursor-extended lookback), and the fallback heuristic `_check_if_handled` only recognized explicit threaded replies — missing emoji-reaction acks, non-reply answers, and deliberate no-reply judgments.

**WS1 — mechanical scanners (dedup-authoritative):**
- `dedup_record_ttl_s` (settings-backed, defaults to `last_processed_ttl_s` ~30d, `TIMEOUTS__DEDUP_RECORD_TTL_S` env override) replaces the hardcoded 7200s TTL on `DedupRecord`
- Added a REQUIRED one-time per-chat dedup-seeding pass (`bridge/dedup_seed.py`) that runs at bridge startup before `scan_for_missed_messages` and before live dispatch begins, repopulating the already-handled window so the fix doesn't re-enqueue the entire historical backlog at rollout. Guarded by **per-chat** markers `data/dedup-seeded.{chat_id}` (never a single global marker) so a partial per-chat failure self-heals on the next restart
- Deleted `_check_if_handled` (the reply-only heuristic) — the dedup set is now authoritative over the full scan window
- Added structured seed/scan logging (per-chat seed summary + per-scan `re_enqueued`/`skipped_duplicate` counters) for post-rollout recurrence detection
- Added `_MAX_IDS >= max(scanner fetch limits)` invariant test instead of speculatively bumping `_MAX_IDS`

**WS2 — valor-catchup (LLM sweep) reaction awareness:**
- Live-Telethon spike found the self-reaction is reliably readable via `message.reactions.results[i].chosen_order` (NOT the bounded `recent_reactions`/`.my` field, which proved unreliable)
- `_valor_reacted()` treats a Valor emoji reaction on an inbound message as ANSWERED before invoking the LLM judge, closing the reaction-only-ack blind spot
- Scrubbed stale `_check_if_handled` docstring references

**Scope note:** the reconciler's fixed 30-min lookback never ages out of the dedup TTL, so it was never a bug site — it benefits from the fix only as a regression guard.

**Rollback:** `data/catchup-disabled` re-touch + bridge restart remains the documented one-command rollback lever if duplicate replies reappear post-rollout.

## Test plan
- [x] `pytest tests/unit/test_dedup.py -q` — 16 passed
- [x] `pytest tests/unit/test_reconciler.py tests/integration/test_reconciler.py -q` — 26 passed
- [x] `pytest tests/unit/test_catchup_claim.py tests/unit/test_catchup_seed.py tests/integration/test_per_chat_catchup_cutoff.py -q` — 17 passed
- [x] `pytest tests/unit/test_agent_catchup.py tests/integration/test_agent_catchup_recovery.py -q` — 34 passed
- [x] `grep -rn "_check_if_handled" bridge/catchup.py bridge/agent_catchup.py tests/` — 0 matches
- [x] `grep -n "7200" models/dedup.py` — 0 matches
- [x] `grep -rn "dedup-seeded" bridge/ | grep -v "dedup-seeded\."` — 0 matches (per-chat only)
- [x] `python -m ruff check .` / `ruff format --check .` — clean (one pre-existing, unrelated finding in `tests/unit/test_content_decode.py`, confirmed identical on `main`)
- [ ] Operator step post-merge: remove `data/catchup-disabled` on the primary machine to re-enable recovery (tracked in plan, not part of this PR)

Closes #2204


valorengels:

Solid, well-scoped bug fix for #2204. The dedup set is made authoritative over the full cursor-bounded scan window (settings-backed 30d TTL), the reply-only `_check_if_handled` heuristic is deleted, the one-time per-chat rollout seed closes the aged-out-keys gap with self-healing per-chat markers, and `valor-catchup` gains reaction-awareness via the spike-confirmed `results[i].chosen_order` field. Tests are thorough and exercise the real Redis dedup path. No blockers.

All three grep verification checks pass (`_check_if_handled`=0, `7200`=0 in models/dedup.py, no global `dedup-seeded` flag). Targeted tests all green on the PR branch: 57 unit + 21 integration passed. The one `ruff check .` finding is in `tests/unit/test_content_decode.py`, which is not in this diff and is identical on `main` — matches the PR-body disclosure.

Findings are minor comment-accuracy nits left behind by the TTL semantics change. `/do-patch` can clear them.

## Rubric

- [x] **1. Plan vs. implementation match** — pass — *all Success Criteria delivered; the `rm data/catchup-disabled` step is an [EXTERNAL] post-merge operator action, acknowledged.*
- [x] **2. New code quality** — pass — *`_valor_reacted`, `dedup_seed.py`, and the seed wiring are defensive by construction (fail-closed to "not reacted" / "not seeded", never raise into startup).*
- [x] **3. Test coverage** — pass — *new unit + real-Redis integration tests for reaction detection, per-chat seed isolation, >2h-old skip, fresh-chat recovery, and the per-chat-failure marker regression.*
- [x] **4. Regression risk to existing callers** — pass — *TTL change is backward-compatible (Popoto sets TTL on write); `_check_if_handled` removal is covered by updated scanner tests; reconciler behavior unchanged (regression guard only).*
- [x] **5. Data integrity** — pass — *no schema shape change; no migration required; seed is additive.*
- [x] **6. Security** — pass — *no secrets, no unsafe patterns.*
- [ ] **7. Documentation accuracy** — fail — *two inline comments in files touched by this change still say "2h" after the TTL became cursor-coupled (~30d) — see Nits.*
- [x] **8. PR body accuracy** — pass — *grep checks, test counts, and the disclosed pre-existing ruff finding all verified.*
- [x] **9. Disclosed deferrals** — pass — *[EXTERNAL] operator step and [SEPARATE-SLUG #2179] null-msgid dedup are sound out-of-scope exclusions.*
- [x] **10. Follow-up claims verified** — n/a — *no `filed as follow-up #N` claims requiring an open-issue check.*

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — PASS — *see Rubric item 1.*
- **2. No-Gos from plan — none violated** — PASS — *no cursor-based `id <= cursor` live skip added; seed uses `id <= cursor` only as the one-shot seed-time proxy, as the plan sanctions.*
- **3. New `except Exception` blocks — each has logger/raise/swallow-ok** — PASS — *`_valor_reacted` swallows-ok (documented); `seed_dedup_for_chat`/`seed_dedup_for_chats` log-and-continue; startup wiring logs a warning, never crashes.*
- **4. New integration tests — exercise serialization boundary (not in-memory only)** — PASS — *`test_catchup_seed.py` and `TestDurableDedupAuthoritativeOverFullScanWindow` hit real Redis dedup.*
- **5. Plan internal consistency — spike findings match task steps** — PASS — *the `chosen_order` spike finding is reflected in code, tests, and docs.*
- **7. New public APIs — docstrings present** — PASS — *all new `dedup_seed.py` functions documented.*
- **8. Breaking changes — migration path documented** — PASS — *the one-time seed + Popoto-TTL-on-write is the documented rollout path; no migration.*
- **10. Tests cover the failure path (not just happy path)** — PASS — *Telethon failure leaves marker absent; `get_dialogs` failure skips pass; shape-surprise reaction defaults False.*
- **12. Docs updated for user-facing changes** — PASS — *`agent-judgment-catchup.md`, `message-reconciler.md`, and README index all updated.*

### Blockers
- None

### Tech Debt
- None

### Nits
- [ ] **`models/dedup.py:59`** — `# C3 (#1817): DedupRecord's short 2h TTL makes it the most likely model to accumulate ghost index members` — The TTL is no longer 2h; it is cursor-coupled (~30d). Per this plan's Risk 1, the longer TTL actually *reduces* ghost churn, so this comment now argues the opposite of the new reality. The file's module/class docstrings were updated for the 2h→cursor-coupled change; this inline comment was missed.
- [ ] **`bridge/telegram_bridge.py:3097-3104`** — the seed-wiring comment states running the seed inline "also closes the seed-vs-live-dispatch lost-update race (Race 3)". The `NewMessage` handler is registered at line 1152 and the client connects with `catch_up=True` (line 1145) before the seed runs (line 3099), so the live handler is already dispatching during the seed — the plan's *primary* Race 3 mitigation ("run the seed before the live NewMessage handler starts dispatching") is not actually realized. Runtime is still safe via the plan's documented *secondary* containment (a clobbered id is a live `id > cursor`, claim-guarded, and re-added on the next dispatch write; window is microseconds and self-healing), so this is a comment-accuracy issue, not a correctness bug. Suggest softening the comment to "narrows" rather than "closes", crediting the secondary containment.

### Miscellaneous
- **Advisory (not a finding against this PR):** `bridge/telegram_bridge.py:1192-1193`'s live-handler comment ("the 2h DedupRecord above can't cover a catch_up replay of a message we handled *yesterday*") is now stale for the same 2h→30d reason, though this PR does not touch those lines. The reasoning still holds (the cursor guard is the real backstop); only the "2h" figure is outdated.

- **`rm data/catchup-disabled` on the primary machine** — [EXTERNAL] post-merge operator step; correctly excluded from the PR.
- **Null/absent message-id handling in the dedup path** — [SEPARATE-SLUG] tracked in `docs/plans/issue-2179-null-msgid-dedup.md`; not touched here.

### Verification Results
- `grep _check_if_handled bridge/catchup.py bridge/agent_catchup.py tests/` → 0 ✅
- `grep 7200 models/dedup.py` → 0 ✅
- `grep dedup-seeded bridge/ | grep -v 'dedup-seeded\.'` → 0 (per-chat only) ✅
- `pytest test_dedup.py test_catchup_seed.py test_agent_catchup.py test_catchup_claim.py` → 57 passed ✅
- `pytest test_per_chat_catchup_cutoff.py test_agent_catchup_recovery.py test_catchup_revival.py` → 21 passed ✅
- `ruff check .` → 1 finding in `tests/unit/test_content_decode.py` (pre-existing on `main`, not in this diff; matches PR disclosure) ⚠️ (not attributable)
- `ruff format --check` (changed files) → clean ✅

### Screenshots

PR #2414, opened 2026-07-27
fix(#2396): break bridge watchdog level-5 livelock

### LINKED ISSUE #2396: Bridge watchdog livelocks at recovery level 5 — crash-count escalation defeats the wedge detector's own restart
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/bridge_watchdog_level_5_livelock.md

> **Context:** The Telegram bridge (`bridge/telegram_bridge.py`) uses [Telethon](https://docs.telethon.dev/), an archived/unmaintained Telegram client library, and has a known issue (#1408, closed) where Telethon silently stops delivering `NewMessage` events while remaining connected — no error, no disconnect. [monitoring/bridge_watchdog.py](monitoring/bridge_watchdog.py) runs as a separate launchd-scheduled process every 60s and implements a **5-level recovery escalation ladder** (documented in [docs/features/bridge-self-healing.md](docs/features/bridge-self-healing.md)) to detect and self-heal bridge problems, from a simple restart (level 1) up to reverting the last commit (level 4) or alerting a human (level 5).

## Problem

The watchdog's escalation logic has two independent rules that can set `recovery_level`, and they interact to produce a livelock: once triggered, the bridge never actually recovers and no human is ever notified, despite the code being designed to always take *some* corrective action.

**Current behavior:**

1. `assess_update_flow()` in `monitoring/bridge_watchdog.py` detects the known Telethon wedge and is explicitly documented and commented to cap `recovery_level` at 2 (a plain, known-safe `launchctl kickstart` restart) — see the code comment "Level cap: the wedge detector contributes at most level 2 ... It must never push recovery_level to 4" and the doc line "The level cap of 2 is hard — the wedge detector never escalates to level 4 (auto-revert), regardless of how many consecutive wedge ticks occur."
2. Immediately after, a second, reason-blind rule in `check_bridge_health()` runs: `if len(get_recent_crashes(1800)) >= 5: recovery_level = max(recovery_level, 5)`. This ignores what caused the crashes.
3. `execute_recovery()`'s level-5 branch does **only** `logger.critical(...)` + `log_crash("... Recovery exhausted")` — it never calls `restart_bridge()`. It also does not send any Telegram alert, despite the escalation table in `docs/features/bridge-self-healing.md` documenting level 5 as "Alert human via Telegram" — no such call exists anywhere in the codebase (confirmed via `grep -rn "Recovery exhausted" --include="*.py" .`, the only hit is the `log_crash()` call site itself).
4. Because the wedge re-fires on essentially every 60-second watchdog tick once it starts (nothing at level 5 resets the Redis `bridge:last_update_received` key that `assess_update_flow()` reads), the 30-minute rolling crash count never drops back below 5. Recovery is then permanently pinned at level 5, which takes no action, so the wedge condition never clears, so the next tick re-detects it — a self-sustaining livelock with no exit except a manual `./scripts/valor-service.sh restart`.

**Observed evidence** (2026-07-26, live at time of filing):
- `python monitoring/bridge_watchdog.py --check-only` reports `Healthy: False`, `50 crashes in last 30 minutes`, `Recovery level: 5`.
- `grep -c "Executing recovery level" logs/watchdog.log` shows level 5 logged on every ~60s cycle continuously for 7+ hours since the last manual bridge restart (~06:46 UTC), with zero interleaved level 1-3 restart attempts in between.
- `monitoring/crash_tracker.get_recent_crashes(86400)` returned 2743 crash events in 24h, spanning multiple git commits (from `352e469b` through the current `HEAD`) — this rules out a single bad deploy; the defect is in the escalation logic itself, not the code it's escalating about.
- Practical impact: bridge (and correlated worker) processes were observed cycling/restarting unpredictably during this window, and human operators trying to run SDLC pipeline work over Telegram experienced repeated session interruptions before an agent could start executing — consistent with the bridge never reaching a stable, actually-recovered state.

**Desired outcome:** A wedge-caused crash streak — already judged safe to handle with a plain restart — actually gets that restart, on a bounded cadence, instead of being silently swallowed into a no-op "alert" state once it recurs often enough. If the ladder is ever going to sit at level 5 for real, level 5 must actually notify a human (matching its own documentation), not just write to a log file nobody is tailing.

## Definitions

| Telethon | The (now-archived) Python Telegram client library the bridge uses to receive live messages | [Telethon docs](https://docs.telethon.dev/) |
| Update-loop wedge | Telethon silently stopping delivery of `NewMessage` events while the process and TCP/API layer stay healthy | [docs/features/bridge-self-healing.md § 3a](docs/features/bridge-self-healing.md), issue #1408 |
| Recovery escalation ladder | The 5-level scale (`monitoring/bridge_watchdog.py::execute_recovery`) the bridge watchdog uses to decide how aggressively to intervene, from plain restart (1) to alert-only (5) | [docs/features/bridge-self-healing.md § 5-Level Recovery Escalation](docs/features/bridge-self-healing.md) |
| `crash_tracker` | `monitoring/crash_tracker.py` — records crash events (reason, commit sha, timestamp) to Redis and exposes `get_recent_crashes()` / `detect_crash_pattern()` used to decide escalation level | [monitoring/crash_tracker.py](monitoring/crash_tracker.py) |

## Prior Context

Related work:
- #1408 (closed): "Messages permanently lost in Telethon update gap" — the original diagnosis of the silent Telethon delivery gap; fixed the catchup/reconciler dead zone, not the watchdog's own escalation behavior.
- #1712 / PR #1723: added the update-loop wedge detector (`assess_update_flow()`) and its documented level-2 cap — this issue is about that cap being defeated by an unrelated, pre-existing rule in the same function.

## Solution Sketch

Open questions for `/do-plan` to resolve (root cause is understood; the right design tradeoff is not yet decided):

- Should the crash-count rule (`>= 5 crashes in 30 min → level 5`) take remedial action *in addition to* alerting, rather than *instead of* the lower-level action it's overriding? i.e. always execute at least the ladder's already-capped action for the underlying reason, and layer the level-5 alert on top when the count crosses the threshold.
- Should the crash-count threshold be reason-aware — only escalate to 5 when recent crashes are heterogeneous (multiple distinct reasons), rather than N repeats of a single reason that already has its own documented-safe cap?
- Should there be a hard ceiling on consecutive level-5 no-op ticks (e.g. after K consecutive level-5 assessments with no intervening successful lower-level action, force one attempt at the capped action) as a livelock breaker?
- Level 5 must be made to actually notify a human (e.g. via the existing Telegram output path) to match its documented behavior — today it only logs to a file.
- Preserve existing safety intent: level 4 (auto-revert) must remain gated on `AUTO_REVERT_ENABLED_FILE` exactly as today, and the wedge detector must continue to never authorize an auto-revert on its own.

## Recon Summary

- The level-5 no-op livelock is reproducible and currently live (not historical) — verified directly via `--check-only` and `logs/watchdog.log` grep.
- The wedge detector's documented "never escalates past 2" guarantee is violated in practice by the separate crash-count rule (it jumps straight past 4 to 5, worse than the auto-revert the comment worries about).
- Level 5's "Alert human via Telegram" documented behavior does not exist in code.
- 2743 crash events / 24h and level-5 logged on every single 60s tick for 7+ continuous hours confirm this is not a transient blip but a standing defect.

- None blocking; this is a self-contained fix within `monitoring/bridge_watchdog.py`'s escalation logic. (No pre-requisite work identified.)

- Migrating off Telethon entirely (the archived library is the ultimate source of the wedge) is out of scope — the wedge itself is an accepted, already-mitigated condition (issue #1408); this issue is scoped strictly to the escalation ladder failing to actually apply its own mitigation.

## Acceptance Criteria

- [ ] A wedge-caused crash streak results in the bridge actually being restarted (the ladder's own level-2 action executes), not just logged, even after the 30-minute crash count crosses 5.
- [ ] Level 5 sends an actual human-visible notification (e.g. Telegram message) when reached, matching `docs/features/bridge-self-healing.md`'s documented behavior — or the doc is corrected to match new, intentional behavior.
- [ ] `logs/watchdog.log` no longer shows an unbroken streak of "Executing recovery level 5" with no interleaved restart attempts under a live wedge condition.
- [ ] Existing safety guarantees are preserved: level 4 (auto-revert) still requires `AUTO_REVERT_ENABLED_FILE`; the wedge detector still cannot unilaterally trigger auto-revert.
- [ ] Regression test covering the specific interaction: wedge detector fires repeatedly within a 30-minute window and the system still performs the capped restart action rather than silently downgrading to a no-op alert.

## Downstream

This issue feeds into `/do-plan`, which will produce a plan document at `docs/plans/bridge-watchdog-level-5-livelock.md`. The plan is then executed by `/do-build`.

Constraint for the planner: this is production self-healing infrastructure for the sole Telegram bridge process — changes must not introduce a restart storm (e.g. uncapped rapid restarts) or weaken the existing auto-revert gating. Prefer the smallest change that breaks the livelock and restores the documented human-alert behavior.

valorengels:
## Summary

- The bridge watchdog's 5-level recovery escalation permanently pinned at level 5 ("alert human," which took no action and sent no alert) once a 30-minute crash-count threshold was crossed — this livelocked recovery for a recurring Telethon update-loop wedge, since the same threshold re-crossed every tick and silently overrode the level-2 restart the wedge detector had already judged safe.
- Splits the crash-count signal into two independent `HealthStatus` fields instead of overriding `recovery_level`: `human_alert_needed` (deduplicated Telegram alert, 30-min cooldown, also now fires from the previously-silent level-4 "recovery exhausted" fallback) and `restart_circuit_open` (reason-aware throttle — a non-wedge-dominated storm suppresses that tick's restart; a wedge-dominated storm always restarts, every tick, no cap).
- `recovery_level` now only ever reaches 1-4. Level 5 is deleted as a dispatch target; former call sites route through a new `_recovery_exhausted()` helper. Three safety constants (`CRASH_STORM_THRESHOLD`, `WEDGE_DOMINANCE_FRACTION`, `WATCHDOG_ALERT_COOLDOWN_SECONDS`) are env-overridable. `--check-only` output gains the three new signal lines.
- Updated `docs/features/bridge-self-healing.md` (escalation table + wedge-detector recovery paragraph) and the stale `1-5` field comment.

Plan: `docs/plans/bridge_watchdog_level_5_livelock.md`

## Test plan

- [x] `scripts/pytest-clean.sh tests/unit/test_bridge_watchdog.py tests/integration/test_update_loop_wedge_recovery.py -q` — 87 passed
- [x] `python -m ruff check` / `ruff format --check` on changed files — clean
- [x] Verification grep table from the plan (level-5 dispatch removed, `_recovery_exhausted` present, `restart_circuit_open` present, `WEDGE_DOMINANCE_FRACTION` default `0.9`, no backoff mechanism, env-overridable constants, stale comment fixed, no `O_EXCL`, docs updated) — all pass
- [ ] Live post-deploy check (C4, plan Success Criteria): after merge and `/update` on the bridge machine, run `python monitoring/bridge_watchdog.py --check-only` and confirm `Recovery level:` is 0-4 (never 5) plus the new alert/circuit lines; tail `logs/watchdog.log` and confirm no pinned "Executing recovery level 5" loop

Closes #2396

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -383,7 +383,7 @@ The bridge includes automatic crash recovery (see `docs/features/bridge-self-hea
 - **Session lock cleanup**: Kills stale processes holding session-related files on startup
 - **Bridge watchdog**: Separate launchd service (`com.valor.bridge-watchdog`) monitors health every 60s
 - **Crash tracker**: Logs start/crash events to Redis via `monitoring/crash_tracker.py` with git commit correlation
-- **5-level escalation**: restart → kill stale → clear locks → revert commit → alert human
+- **4-level escalation, plus a decoupled human-alert signal** (#2396): restart → kill stale → clear locks → revert commit; a crash-storm alert (`human_alert_needed`) and a reason-aware restart throttle (`restart_circuit_open`) fire independently of `recovery_level` so a recurring wedge's capped restart is never silently overridden
 - **Update-loop wedged detector** (#1712): detects when the bridge is process-alive but Telethon's `NewMessage` handler has silently stopped firing — auto-restarts with `catch_up=True` for lossless backfill
 
 **Check watchdog**: `python monitoring/bridge_watchdog.py --check-only`
--- a/docs/features/bridge-self-healing.md
+++ b/docs/features/bridge-self-healing.md
@@ -75,15 +75,19 @@ Claude Code CLI subprocesses can become orphaned when their parent session ends 
 
 The `--check-only` output includes zombie count, PIDs, memory usage, and active instance count.
 
-**5-Level Recovery Escalation**:
+**4-Level Recovery Escalation, plus a decoupled human-alert signal** (issue #2396):
 
 | Level | Condition | Action |
 |-------|-----------|--------|
 | 1 | Process not running | Log crash event via `crash_tracker.log_crash("bridge_dead_on_watchdog_check")` + simple restart (launchd) |
 | 2 | Process running but logs stale — or update loop wedged | Kill stale + kill zombies + restart (the bridge always catches up missed messages on startup — see below) |
 | 3 | Lock files present | Kill stale + kill zombies + clear locks + restart |
-| 4 | Crash pattern detected | Kill stale + kill zombies + revert HEAD + restart (if enabled) |
-| 5 | Recovery exhausted | Alert human via Telegram |
+| 4 | Crash pattern detected | Kill stale + kill zombies + revert HEAD + restart (if enabled); if auto-revert is disabled or the revert fails, falls through to "Recovery exhausted" below |
+
+`recovery_level` never reaches 5 — level 5 ("alert human") used to be a silent no-op action level that permanently overrode any lower level once a crash-count threshold was crossed, which livelocked the ladder (the wedge detector's own capped, known-safe restart never got to run). It has been replaced by two independent signals computed alongside `recovery_level`, both on `HealthStatus`:
+
+- **`human_alert_needed`** — set when `get_recent_crashes(1800)` (30 min window) returns `>= CRASH_STORM_THRESHOLD` (default 5, env-overridable) crashes. When true, `_alert_human_of_crash_storm()` enqueues a deduplicated Telegram alert (same `AgentSession`-enqueue pattern as `send_hibernation_notification()`), gated by a 30-minute file-sentinel cooldown (`WATCHDOG_ALERT_COOLDOWN_SECONDS`, env-overridable) so a sustained storm alerts once per window, not once per 60s tick. The same alert fires from the level-4 "recovery exhausted" fallback (`_recovery_exhausted()`), so that previously-silent path now also surfaces a human-visible alert.
+- **`restart_circuit_open`** — a reason-aware restart throttle for *non-wedge* storms. `CrashEvent.reason` classifies each crash; when the storm is not wedge-dominated (`wedge_count < len(recent_crashes) * WEDGE_DOMINANCE_FRACTION`, default fraction `0.9` — a bare `0.5` majority would let a 50/50 wedge+real-bug storm through), `restart_circuit_open` is set and `run_health_check()` skips `execute_recovery()` for that tick entirely (still alerting). A **wedge-dominated** storm always leaves this False, so the wedge detector's already-capped restart (level 2, `launchctl kickstart` with `catch_up=True`) keeps running every tick with no attempt ceiling — restarting is cheap and idempotent, and throttling it would recreate the exact livelock this fix removes.
 
 Zombie cleanup is integrated into recovery levels 2+ to free memory before restarting.
 
@@ -118,7 +122,7 @@ A SECONDARY accelerator fires at `UPDATE_STALENESS_WARN` (before the ceiling) to
 - **Startup grace window**: `bridge:last_update_received` is absent on cold start (bridge has not received any messages yet). The grace window prevents false wedge verdicts during startup before Telegram delivers the first event.
 - **`None` process-start = fail-safe**: if `get_bridge_process_start_ts()` returns `None` (process info unavailable), the detector treats the verdict as inconclusive and suppresses the restart. This avoids a restart based on incomplete information.
 
-**Recovery**: when `update_flow_live=False`, the watchdog sets `recovery_level = max(recovery_level, 2)` and calls the standard `restart_bridge()` (a `launchctl kickstart` — it takes no arguments). The level cap of 2 is hard — the wedge detector never escalates to level 4 (auto-revert), regardless of how many consecutive wedge ticks occur. Lossless backfill is inherent to bridge startup, not a flag the watchdog passes: the bridge unconditionally initializes Telethon with `catch_up=True` and runs a missed-message catchup scan on every connect, so any restart recovers the messages that arrived during the wedge window.
+**Recovery**: when `update_flow_live=False`, the watchdog sets `recovery_level = max(recovery_level, 2)` and calls the standard `restart_bridge()` (a `launchctl kickstart` — it takes no arguments). The level cap of 2 is hard — the wedge detector never escalates to level 4 (auto-revert), regardless of how many consecutive wedge ticks occur. Lossless backfill is inherent to bridge startup, not a flag the watchdog passes: the bridge unconditionally initializes Telethon with `catch_up=True` and runs a missed-message catchup scan on every connect, so any restart recovers the messages that arrived during the wedge window. This level-2 cap is now honored end to end: previously, a wedge that recurred faster than the crash-count window could re-cross while the crash-count check was still active, silently overriding this capped restart with the level-5 no-op (issue #2396). Because that override no longer exists, a recurring wedge always gets its capped restart on every tick — the crash-count storm this produces (all `bridge_update_loop_wedged`-reason crashes) is wedge-dominated, so `restart_circuit_open` stays False and the restart is never throttled.
 
 **Log signals**:
 ```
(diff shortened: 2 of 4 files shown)


valorengels:
## Review: APPROVED

**PR #2414** — `fix(#2396): break bridge watchdog level-5 livelock`. Reviewed at head SHA `61a824341d89db648a802c6fc32cac7d0b949fa8`.

**Verdict: APPROVED** — 0 blockers, 0 tech-debt items.

### Verification (all green)
- Preflight: OPEN / MERGEABLE / CLEAN.
- Level-5 dispatch branch removed; both former `execute_recovery(5, ...)` call sites reroute through `_recovery_exhausted()`.
- Crash-count signal split off `recovery_level` into `human_alert_needed` + `restart_circuit_open`; reason-aware suppression via `WEDGE_DOMINANCE_FRACTION=0.9` (default), all three new constants env-overridable.
- Wedge storms always restart every tick (no cap); non-wedge-dominated storms suppress restart.
- No backoff mechanism, no `O_EXCL`; level-4 auto-revert gate intact; stale `1-5` comment fixed; docs updated (`docs/features/bridge-self-healing.md`).
- 87 tests pass (73 unit + 14 integration) at PR SHA in an isolated worktree; ruff `check` and `format --check` clean.
- Integration points confirmed against real source: `CrashEvent.reason`, `send_hibernation_notification` reuse (no new send path), `_hostname()`.
- Subtle correctness verified: a non-wedge storm no longer bumps `recovery_level` but still appends an issue string, so `healthy` is False and `run_health_check` reaches the alert/circuit wiring rather than short-circuiting; the shared cooldown gate dedups the two in-tick alert call sites.

### Follow-up (non-blocking)
Critique concern C1 (deferred): `wedge_count` can be contaminated by the wedge restart's own `os.kill` producing interleaved `bridge_dead_on_watchdog_check` crashes in the same 30-min window, potentially eroding the 0.9 fraction and flipping `restart_circuit_open` (a secondary livelock). Judged out of the Small appetite and not a present-day correctness blocker — recommend a follow-up issue rather than blocking this PR.

valorengels:
## Independent Review — APPROVE (zero findings)

Reviewed against plan `docs/plans/bridge_watchdog_level_5_livelock.md` and issue #2396 at head `a02fa4dfa`.

**Correctness** — The level-5 override (`recovery_level = max(recovery_level, 5)`) is removed and replaced by two independent `HealthStatus` signals (`human_alert_needed`, `restart_circuit_open`). The livelock root cause (reason-blind crash-count override silently masking the wedge detector's capped, known-safe level-2 restart) is fixed: a wedge-dominated storm now leaves the circuit closed and restarts every tick with no attempt ceiling. Level 5 is fully deleted as a dispatch target; both former internal `execute_recovery(5, …)` call sites route through `_recovery_exhausted()`. The `_alert_human_of_crash_storm()` enqueue mirrors `send_hibernation_notification()` exactly (same `AgentSession(session_type="teammate", …).save()` pattern), and the shared cooldown gate dedups the alert across both the `human_alert_needed` path and the level-4 fallback.

**Tests** — 87 pass (73 unit + 14 integration). New coverage maps 1:1 to the plan's success criteria: livelock regression, non-wedge circuit open, 50/50 mixed storm under the 0.9 dominance bar, env-override, level-4 recovery-exhausted fallback (auto-revert disabled + revert-failure), cooldown open/expired/no-stamp pair, and dedup.

**Verification table** — all plan greps pass (no `elif level == 5`, no `execute_recovery(5`, no `O_EXCL`, no backoff mechanism, 3 env-overridable constants, `WEDGE_DOMINANCE_FRACTION` default `0.9`, AUTO_REVERT gate intact, stale `recovery_level` `1-5` field comment corrected to `1-4`).

**Lint/format** — `ruff check` + `ruff format --check` clean.

**Docs** — `docs/features/bridge-self-healing.md` (escalation table + wedge-detector recovery paragraph), `CLAUDE.md` self-healing bullet, and the inline `recovery_level` field comment all updated to the corrected two-signal design.

Note: the only pre-merge item deferred is the live post-deploy check (C4) — it can only run on the bridge machine after `/update`.

Issue #2098, opened 2026-07-15
Make session-liveness-check single-owner: out-of-process reflection false-kills live sessions (confirmed #2091 root cause)


valorengels:
> **Context.** This system runs AgentSessions (queued units of agent work) on a single **worker** process (`python -m worker`). A separate **reflection worker** subprocess (`com.valor.reflection-worker`, `python -m reflections`) runs periodic maintenance jobs declared in `config/reflections.yaml`. This issue is the confirmed root cause of the **#2091 double-owner incident**, where two processes resumed the *same* session concurrently. Full forensic evidence is in [`docs/plans/session-recovery-observation-audit.md`](../blob/main/docs/plans/session-recovery-observation-audit.md) — see §"Executive diagnosis" point 1, §"Incident reconstruction: issue #2091", and P0 fix items 1–2. This is part of the #1818 resilience-hardening umbrella.

## Problem

`_agent_session_health_check` (in `agent/agent_session_queue.py`) is the actuation function that scans running sessions and recovers ones it judges dead. It runs in **two processes at once**:

1. **In the worker**, in-process, via `_agent_session_health_loop` (`worker/__main__.py:890`) — this is the legitimate owner.
2. **In the reflection subprocess**, as the `session-liveness-check` reflection (`config/reflections.yaml:35-42`, `enabled: true`, `every: 300s`).

The decisive liveness signal is **process-local**: the worker-presence branch reads `_active_workers[worker_key]` and `_active_sessions[agent_session_id]` (`agent/session_health.py:3398-3434`) to decide `worker_alive`. Those dictionaries are populated only inside the real worker's process. **In the reflection subprocess they are empty**, so `worker_alive` is always `False` there.

Consequence: any running session older than the 300s worker-presence guard (`AGENT_SESSION_HEALTH_MIN_RUNNING`) is declared `worker_dead` by the reflection process even though the real worker is actively driving it. The reflection then transitions the row `running -> pending`, starts its **own** queue worker, and resumes the same Claude conversation UUID — while the original executor is still alive. Two processes now own one work item; the DB status describes neither reliably.

### Current behavior (confirmed in code AND incident logs)

- `11:00:34` — real worker's in-process health check finds nothing to recover; deadman beacon age 2.9s (`logs/worker.log:72781-72783`).
[...]

Issue #2418, opened 2026-07-27
Bridge watchdog: wedge-restart os.kill can dilute WEDGE_DOMINANCE_FRACTION into a secondary restart-suppression livelock


valorengels:
> **Context:** The Telegram bridge (`bridge/telegram_bridge.py`) uses [Telethon](https://docs.telethon.dev/), an archived Telegram client library with a known failure mode (issue #1408) where it silently stops delivering `NewMessage` events while the process and TCP layer stay healthy — the "update-loop wedge." [`monitoring/bridge_watchdog.py`](monitoring/bridge_watchdog.py) runs every 60s via launchd and implements a recovery escalation ladder to self-heal the bridge. Issue #2396 (PR #2414, commit `b299805`) just fixed a livelock in that ladder. **This issue is a theoretical, deferred follow-up to that fix** — a second, structurally similar livelock that the #2396 fix can reach by a different route. It has NOT been observed in production; it is a reasoned failure mode derived from reading the control flow.

## Background: what #2396 changed

Before #2396, a reason-blind rule (`if len(get_recent_crashes(1800)) >= 5: recovery_level = max(recovery_level, 5)`) overrode the wedge detector's deliberate level-2 cap, and level 5 only logged (no restart, no real alert) — a self-sustaining no-op livelock where the top "emergency" rung prevented every lower rung from firing.

The fix (`monitoring/bridge_watchdog.py`) split that reason-blind crash-count signal off `recovery_level` into two independent `HealthStatus` fields:

- **`human_alert_needed`** — a deduplicated human-visible Telegram alert (30-min cooldown via `WATCHDOG_ALERT_COOLDOWN_SECONDS`, file sentinel `data/bridge-watchdog-alert-cooldown`).
- **`restart_circuit_open`** — a **reason-aware** restart suppressor. A crash storm that is *wedge-dominated* still gets the capped restart every tick; a storm that is *not* wedge-dominated opens the circuit and suppresses that tick's restart to avoid thrash.

The dominance test (`check_bridge_health`, `monitoring/bridge_watchdog.py:552`):

```python
if len(recent_crashes) >= CRASH_STORM_THRESHOLD:      # default 5
    human_alert_needed = True
    if wedge_count < len(recent_crashes) * WEDGE_DOMINANCE_FRACTION:   # default 0.9
        restart_circuit_open = True                   # suppress this tick's restart
```

[...]

PR #2450, opened 2026-07-29
Fix notify-less session creation + Popoto boolean string mis-cast (#2439)

### LINKED ISSUE #2439: Root-cause and patch: notify-less AgentSession creation stranding pending sessions + Popoto boolean string mis-cast surfaced by stuck crash-storm alerts
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/notify-less-session-creation-and-popoto-bool-miscast.md

> **Context:** This repo runs a Telegram bridge + a standalone [worker](docs/features/bridge-worker-architecture.md) that executes [`AgentSession`](models/agent_session.py) records via [Popoto](https://github.com/tomcounsell/popoto) (a Redis-backed ORM). A separate `monitoring/bridge_watchdog.py` launchd service escalates through 4 recovery levels when the bridge crashes repeatedly, and as a last resort enqueues a Telegram alert session for a human. Investigating why 48 of those alert sessions sat unprocessed for up to ~48 hours surfaced two distinct, previously-undetected bug classes below. **This issue is scoping only** — no root-cause fix has been applied, and no assumption should be made about which gap actually caused the 48h stall without live-system verification.

## Problem

**Current behavior:**

1. Dozens of `AgentSession` records (created by `monitoring/bridge_watchdog.py`'s crash-storm alert, `_alert_human_of_crash_storm()` at [`monitoring/bridge_watchdog.py:763-800`](monitoring/bridge_watchdog.py)) sat in `status="pending"` for up to ~48 hours, never delivered to Telegram. Each one's entire job is "send this exact alert text to the Eng: Valor chat" — none ever ran.
2. Separately, the dashboard (`ui/app.py`, `ui/data/sdlc.py`) reported `requires_real_chrome: true` for every one of those 48 sessions. Direct inspection of the Redis-backed record (`AgentSession.get_by_id(...)`) showed the real stored value is the string `'False'` — the dashboard's display was simply wrong, and this misled the initial investigation toward the [BYOB real-Chrome scheduler gate](agent/session_pickup.py) as a suspected cause before it was ruled out.

**Desired outcome:** A human (or an automated process) can identify, live, why a given `pending` `AgentSession` isn't running, and the dashboard's status fields reflect ground truth. Concretely: (a) the root cause of the 48h stall is identified and patched, (b) the boolean-field display bug is fixed everywhere it appears, not just in the one site that was noticed.

## Definitions

| Popoto | The Redis-backed ORM this repo uses for models like `AgentSession`. Untyped `Field(default=False)` booleans round-trip through Redis as the *string* `'True'`/`'False'`, not a Python `bool` — so `bool(getattr(obj, "field", False))` is a bug: `bool('False')` evaluates `True` in Python because it's a non-empty string. | [`agent/session_pickup.py:76-91`](agent/session_pickup.py) defines the correct canonicalizing helper, `_truthy()` |
| `valor:sessions:new` notify | A Redis pubsub channel the normal session-enqueue path (`_push_agent_session()`/`enqueue_agent_session()` in `agent/agent_session_queue.py`) publishes to, so the worker wakes immediately instead of waiting for its next periodic scan. | [`agent/agent_session_queue.py`](agent/agent_session_queue.py) — `notify_channel_for()`, `_session_notify_listener()` |
| Reflection | A recurring background job declared in `~/Desktop/Valor/reflections.yaml` (synced to `config/reflections.yaml`) and run by an out-of-process scheduler subprocess (`python -m reflections`). | [`docs/features/README.md`](docs/features/README.md), `agent/reflection_scheduler.py` |
| BYOB / `requires_real_chrome` | A scheduler-layer gate (issue #1256) that serializes sessions needing the user's real Chrome browser so two don't race on the same tab. Ruled out as the cause of this stall — see Problem §2. | [`agent/session_pickup.py`](agent/session_pickup.py) |

## Prior Context

Related work:
- #1804: Standalone worker runs in bridge mode (`VALOR_WORKER_MODE` unset); notify-listener miss strands sessions — closed, same symptom class (missed notify → stranded session) as Bug B below, different root cause.
- #1811: Notify-listener NUMSUB self-check livelocks: bytes-vs-str key mismatch makes it always read 0 subscribers — closed, another prior notify-pipeline defect.
- #2147 / PR #2163: Test-suite notify isolation (db-scoped notify channel + live-worker guard) — established the `notify_channel_for()` db-scoping this issue's Bug B section references.
- PR #2170 (#2143, #2165): Worker hang forensics + resume-notify pickup — added `_publish_resume_notify()` to `tools/valor_session.py`'s `resume_session()`. A memory note claiming resume doesn't publish the notify appears to predate this fix and should be reconciled (see Bug B).
- #2418: Bridge watchdog wedge-restart livelock (open) — same file (`monitoring/bridge_watchdog.py`) as Bug B's confirmed stall source, different mechanism (restart-suppression vs. alert-delivery); worth a shared look given both live in the watchdog's escalation path.

## Solution Sketch

Root cause is **not yet known** — these are open questions for `/do-plan` to investigate, not a prescribed approach:

- Why did these 48 specific sessions go up to 48h without being drained by either of the two mechanisms that are supposed to catch missed-notify `pending` sessions (the worker's one-shot startup scan, and its in-process 300s health-check loop, both in `agent/session_health.py`)? Was the live worker's health loop actually ticking during the stall window, or did `_ensure_worker`/the per-`worker_key` queue-loop registry have a gap specific to this `project_key`/`session_type` combination?
- Is the fix to make `_alert_human_of_crash_storm()` (and the two other notify-less construct-and-save sites found in recon) go through the standard `_push_agent_session()` enqueue path — or is the real gap that the periodic backstop itself is broken/absent for out-of-process-created pending sessions?
- Should the 3 drifted copies of the `_truthy()` boolean-canonicalizer be consolidated into one shared import, and should every naive `bool()`/truthiness read of a Popoto string-boolean field (full list in Recon Summary) be migrated to it in one pass, or split per call site?
- Is the `session-liveness-check` reflection's out-of-process no-op (see Recon Summary) intentional dead weight that should be removed, or a regression that should be revived as a real independent safety net?

Constraint: whatever fix is chosen for the notify-publish gap must not change behavior for the existing "instant wake" path used by ordinary Telegram-originated sessions — this is a targeted fix for the alert/notification-only construct-and-save call sites, not a rewrite of `_push_agent_session()`.

## Recon Summary

Two parallel Explore-agent passes over the codebase confirmed the following (file:line citations verified against current `main`):

**Confirmed — Bug A (Popoto boolean string round-trip mis-cast), sites NOT using `_truthy()`:**
- `ui/data/sdlc.py:1164` — `requires_real_chrome=bool(getattr(session, "requires_real_chrome", False))`
- `ui/data/sdlc.py:1177` — `user_facing_routed=bool(getattr(session, "user_facing_routed", False))` (contrast: `ui/data/sdlc.py:1147` computes a sibling field, `is_ledger`, correctly via `agent.session_health._is_ledger()` → `_truthy()`, in the same function)
- `ui/app.py:756` — `"requires_real_chrome": getattr(s, "requires_real_chrome", False)` — raw string exposed into the dashboard JSON API, no cast at all
- `ui/app.py:776` — `"user_facing_routed": s.user_facing_routed` — same, raw attribute
- `tools/valor_session.py:1514-1516` — `retain = getattr(s, "retain_for_resume", False); if not retain: continue` — **logic-inversion bug, not cosmetic**: a real `retain_for_resume=False` record (stored as `"False"`) makes `not retain` evaluate `False`, so the fast-path `continue` never fires and every completed session falls through into PR-matching/release logic below
- `ui/data/reflections.py:134-139` — `bool(getattr(state, "auto_delete_after_run"/"dead_letter_escalated", False))`
- `models/reflection.py:287` — `if not bool(self.dead_letter_escalated):` — the model's own method mis-casting its own field
- `bridge/enrichment.py:69`, `agent/session_executor.py:1684`, `scripts/migrate_model_relationships.py:162,189-190` — same pattern against `has_media` (declared `type=bool` on `models/telegram.py:37` — whether typed fields actually escape the Redis string round-trip was NOT verified live and needs confirming)
- Downstream UI blast radius of the two `ui/data/sdlc.py` sites: `ui/templates/_partials/session_modal_content.html:31,163` render wrong badges for every session in the dashboard modal
- Three independently-drifted copies of the canonicalizer exist: `agent/session_pickup.py:76-90` (canonical), `models/crash_signature.py:34-46` (separate copy), `reflections/crash_recovery.py:171-178` (inline `try/except ImportError` fallback copy)
- Other untyped boolean fields on `models/agent_session.py` sharing this risk shape, no confirmed bad read site yet found: `is_ledger` (:385, generally read correctly via `_truthy`), `budget_tripped` (:520, no production read site exists yet — "no naive-read bug (only test code reads it), but... any future read is a landmine")

**Confirmed — Bug B (new `AgentSession`s created bypassing the notify-publish path):**
- `monitoring/bridge_watchdog.py:789` (`_alert_human_of_crash_storm`) — confirmed source of the 48 stuck sessions
- `reflections/agents/circuit_health_gate.py:103-108` (`send_hibernation_notification()`) — registered in `config/reflections.yaml` as the `circuit-health-gate` reflection (callable `agent.sustainability.circuit_health_gate`, a re-export per `agent/sustainability.py:10`); fires on circuit open→closed transition
- `agent/sustainability.py:114-119` — an independently-defined second copy of `send_hibernation_notification()`, triggered from `agent/agent_session_queue.py:2752-2754` on the worker's own in-process `CircuitOpenError` handling
- `agent/session_archive.py:443` (`_rehydrate_row()`) — cold-start SQLite backup restore; only runs once before the startup pending-scan, likely self-healing but unconfirmed
- `tools/valor_session.py`'s `resume_session()` DOES publish (`_publish_resume_notify()`, `tools/valor_session.py:762-812`, added by PR #2170) — a stale memory note claiming otherwise should be reconciled once this issue's investigation confirms current behavior
- **Critical finding**: `session-liveness-check` (config'd to run `agent.agent_session_queue._agent_session_health_check`) is a structural no-op when run out-of-process — `agent/session_health.py:3670-3677` early-returns because the reflections runner sets `VALOR_REFLECTION_WORKER=1` (`reflections/__main__.py:167`) but never calls `mark_owning_worker_process()` (only the worker's own in-process loop does, `session_health.py:4553`). Despite its "recover stuck ones" description and a healthy-looking `run_count` in the dashboard, it currently does nothing — the only real periodic backstop for stuck-`pending` sessions is the worker's in-process 300s health loop (`agent/session_health.py:4546-4574`), with no independent out-of-process coverage if that loop itself isn't alive/ticking.

**Pre-requisites:** None identified that block starting the investigation — this is greenfield scoping, not a fix in progress.

**Dropped:** Nothing scoped out; both bug classes and the dead-reflection finding are all in scope for the follow-up investigation.

## Acceptance Criteria

- [ ] Live-system root cause identified for why the 48 bridge-watchdog alert sessions sat `pending` for up to 48h without being drained by either the worker startup scan or the 300s in-process health loop
- [ ] `_alert_human_of_crash_storm()` and the other notify-less new-session construct-and-save sites listed above are fixed so newly created alert/notification sessions are reliably picked up (via notify-publish, a real periodic backstop, or both)
- [ ] The `session-liveness-check` reflection's out-of-process no-op is either fixed to provide real independent coverage or removed/re-described so its dashboard status doesn't imply false coverage
- [ ] Every naive `bool()`/truthiness read of a Popoto string-boolean field listed in this issue is replaced with the canonical `_truthy()` helper (with the 3 drifted copies consolidated to one shared import), including the logic-inversion bug in `tools/valor_session.py:1514-1516`
- [ ] The `circuit-health-gate` reflection (temporarily disabled in `~/Desktop/Valor/reflections.yaml` as a stopgap for this issue) is re-enabled once its notify-less-session-creation path is fixed
- [ ] The stale `tools/valor_session.py` resume-notify memory note is confirmed against current code and corrected/removed

## Mitigation Already Applied (stopgap, not a fix)

The `circuit-health-gate` reflection has been temporarily set `enabled: false` in `~/Desktop/Valor/reflections.yaml` (synced to `config/reflections.yaml` by `scripts/install_reflection_worker.sh`), since it shares the Bug B pattern. Tradeoff accepted: while disabled, this reflection's other job — atomically clearing `queue_paused`/`worker:hibernating` flags when the Anthropic circuit closes — also does not run, so the queue could get stuck paused if the circuit trips while it's off. Re-enable once patched.

## Downstream

This issue feeds into `/do-plan`, which will produce a plan document at `docs/plans/{slug}.md`, then executed by `/do-build`. Constraint for the planner: the notify-publish fix must not alter behavior for the existing "instant wake" path used by ordinary Telegram-originated sessions.

valorengels:
## Summary

Fixes two independent bug classes surfaced by 48 bridge-watchdog crash-storm alert sessions sitting `status="pending"` for up to ~48h without being delivered to Telegram.

**Bug B — notify-less session creation.** `monitoring/bridge_watchdog.py::_alert_human_of_crash_storm()`, `reflections/agents/circuit_health_gate.py::send_hibernation_notification()`, and `agent/sustainability.py` construct an `AgentSession` and call `.save()` directly, bypassing the enqueue path that publishes a `valor:sessions:new` pubsub notify. Promoted the existing `_publish_resume_notify` helper into a shared `publish_session_notify(session)` in `agent/agent_session_queue.py` and wired it into all three construct-and-save sites (after `.save()`, per the race-condition ordering requirement). `_push_agent_session()`'s existing instant-wake publish is untouched. Also deduplicated `send_hibernation_notification` (previously defined in two places) to one canonical definition.

**Bug A — Popoto string-boolean mis-cast.** Untyped Popoto `Field(default=False)` fields (`requires_real_chrome`, `user_facing_routed`, `retain_for_resume`) round-trip through Redis as the *string* `'False'`/`'True'`, so naive `bool(x)` reads are always `True` for non-empty strings. Routed these three field's reads through the canonical `_truthy()` helper (`agent/session_pickup.py`) at `ui/data/sdlc.py`, `ui/app.py`, and fixed a genuine logic-inversion bug at `tools/valor_session.py:1485` (`cmd_release`'s retain fast-path). Consolidated a duplicate `_truthy` definition in `models/crash_signature.py` to import the canonical helper. Confirmed via spike that *typed* `Field(type=bool)` fields (`has_media`, reflection `auto_delete_after_run`/`dead_letter_escalated`) round-trip as real Python bools already and were correctly left untouched.

**Reflection cleanup.** Removed the dead `session-liveness-check` reflection (out-of-process actuation is unsafe by design — its logic keys off process-local registries that are empty in any non-owning process; issues #2098/#2091). Re-enabled `circuit-health-gate` now that its notify path is fixed.

**Live root-cause finding (documented, not a code fix in this PR).** The worker's in-process ~300s health-scan backstop was working as designed — it was ticking continuously throughout the stall window. The actual root cause is a queue-priority/starvation problem: all sessions sharing `worker_key="valor"` are served by a single serialized per-key queue loop, and the health-scan's nudge (`asyncio.Event.set()`) can wake an *idle* loop but cannot preempt one that's mid-turn on long-running pipeline work. This PR's notify fix removes the dependency on the slow backstop for *freshness* when the worker is idle, but does not resolve genuine starvation when the worker is busy for a long stretch — flagged as a known residual limitation in `docs/features/bridge-self-healing.md`.

## Test plan
- [x] Targeted suite covering every touched module (19 files, 479 tests) passes serially: `pytest tests/unit/test_agent_session_queue.py tests/unit/test_agent_session_queue_async.py tests/unit/test_bridge_watchdog.py tests/unit/test_sustainability.py tests/unit/test_sustainability_namespace.py tests/unit/test_resume_notify.py tests/unit/test_truthy_helper.py tests/unit/test_dashboard_pillar_a_fields.py tests/unit/test_valor_session_resume_release.py tests/unit/test_crash_signature.py tests/unit/test_crash_signature_cli.py tests/unit/test_crash_signature_library.py tests/unit/test_session_pickup_phantom_filter.py tests/unit/test_crash_recovery_gates.py tests/unit/test_reflection_scheduler.py tests/unit/test_ui_reflections_data.py tests/unit/test_session_health_sibling_phantom_safety.py tests/integration/test_notify_isolation.py -q -n0` → 479 passed
- [x] New end-to-end pubsub test (`tests/integration/test_notify_isolation.py`) proves `publish_session_notify` publishes on the exact channel the worker's listener subscribes to
- [x] `_truthy()` table test + dashboard/release regression tests guard the string-bool fixes
- [x] Diff-scoped `ruff check` and `ruff format --check` clean
- [x] Live reload verified: `./scripts/install_reflection_worker.sh` + `python -m reflections --dry-run` confirm `session-liveness-check` absent, `circuit-health-gate` present
- [x] Documentation updated: `docs/features/bridge-self-healing.md`, `docs/features/reflections.md`, `docs/features/agent-session-health-monitor.md`, `docs/features/redis-models.md`

Closes #2439


valorengels:
## Review: Changes Requested

Thorough, well-scoped fix. Verified end-to-end on the PR branch:

- **Bug B (notify-less creation):** all three live construct-and-save sites are covered — `bridge_watchdog._alert_human_of_crash_storm`, `circuit_health_gate.send_hibernation_notification` (which absorbed the dedup'd `sustainability` copy), and the `_publish_resume_notify` call-through. `agent/session_archive._rehydrate_row` is the empty-Redis restore path and correctly out of scope. The AFTER-save ordering and fail-quiet contract are respected at every site.
- **Bug A (Popoto string-bool):** `_truthy` consolidation to `agent.session_pickup` verified; the claimed circular-import risk in `models/crash_signature.py` is real-tested clean (`crash_signature` in `agent_session.py` is a Field name, not an import). The `cmd_release` logic-inversion fix is correct.
- **Tests:** 74 load-bearing tests pass (`test_truthy_helper`, `test_notify_isolation`, `test_valor_session_resume_release`). Diff-scoped `ruff check` + `ruff format --check` clean. All 12 rows of the plan's Verification table pass. Import/re-export chain intact.

### Blockers
- None

### Tech Debt
- **`agent/session_health.py` carries stale references to the removed `session-liveness-check` reflection** (lines ~3561, ~3663, ~4550). Three inline comments still describe `_agent_session_health_check` as being "also registered as the out-of-process `session-liveness-check` reflection" and describe its behavior "when this runs out-of-process (the session-liveness-check reflection)". Now that the reflection is deleted, nothing invokes this function out-of-process, so these comments describe a non-existent mechanism and misstate the rationale for the #2098 owner-process guard. This file was not touched by the "cascade sweep for session-liveness-check removal" commit. Update the comments (and note whether the #2098 guard is now purely defensive, since its only out-of-process caller is gone) to match the post-removal reality.

### Nits
- None

Verdict: **CHANGES REQUESTED** — one tech-debt item to sweep. Everything else is solid; the notify fix and its documented residual-starvation limitation are honestly scoped.

valorengels:
## Review: Approved

Re-review on a new HEAD (`b4f8658c1`) after the prior `Changes Requested`. The single tech-debt item from that pass — stale `session-liveness-check` comments in `agent/session_health.py` — has been resolved by the `fix(#2439): update stale session-health comments after liveness-check removal` commit. Verified end-to-end on the PR branch; no new issues.

- **Bug B (notify-less creation):** all three live construct-and-save sites publish the create-path notify AFTER `.save()`, fail-quiet: `bridge_watchdog._alert_human_of_crash_storm`, `circuit_health_gate.send_hibernation_notification` (the canonical home, from which `agent/sustainability` now re-exports — `test_reexported_...` asserts `reexported is canonical`), and the `_publish_resume_notify` thin call-through. `_push_agent_session()`'s instant-wake publish is untouched (verified by grep).
- **Bug A (Popoto string-bool):** `_truthy` consolidated to `agent.session_pickup`; `models/crash_signature.py` now imports it (inline `def _truthy` count == 0) and `reflections/crash_recovery.py` dropped its ImportError-fallback copy. Untyped reads at `ui/data/sdlc.py`, `ui/app.py` route through `_truthy`; the `cmd_release` logic-inversion (`if not _truthy(retain)`) is fixed. No import cycle (`agent.session_pickup` pulls in no crash/reflection module).
- **Reflection cleanup:** `session-liveness-check` removed from `ui/data/reflections.py`, docs, and the scheduler tests; `circuit-health-gate` re-enabled. Remaining `session-liveness-check` mentions in `session_health.py` are now historical ("used to be registered") and correctly frame the #2098 guard as defense-in-depth.
- **Root cause:** the 48h stall is honestly attributed to per-`worker_key` queue starvation (a busy worker mid-turn can't be preempted), documented as a known residual limitation in `docs/features/bridge-self-healing.md` — the notify fix removes the idle-worker freshness dependency, not the busy-worker starvation, and the PR says so plainly.
- **Tests / lint:** 138 targeted tests pass serially (`test_truthy_helper`, `TestPublishSessionNotify`, `TestAlertHumanOfCrashStormNotify`, `TestSendHibernationNotificationPublishesNotify`, `test_valor_session_resume_release`, `test_reflection_scheduler`). Diff-scoped `ruff check` + `ruff format --check` clean.

## Rubric

- [x] **1. Plan vs. implementation match** — pass — *all 8 success criteria delivered (root cause documented, 3 notify sites wired, reflection removed, untyped-bool reads fixed + logic-inversion, circuit-health-gate re-enabled, memory note reconciled, single hibernation def, docs).* 
- [x] **2. New code quality** — pass — *shared `publish_session_notify` has a thorough docstring, fail-quiet, correct AFTER-save ordering; no copy-paste (drifted copies collapsed).* 
- [x] **3. Test coverage** — pass — *new unit + integration (channel-parity) tests cover payload, fail-quiet, missing-field, string-bool table, retain skip/release, re-export identity.* 
- [x] **4. Regression risk to existing callers** — pass — *`_publish_resume_notify` kept as call-through; `send_hibernation_notification` re-exported so `agent/agent_session_queue` and tests are unchanged; instant-wake path untouched.* 
- [x] **5. Data integrity** — pass — *read-path only; no schema/field change, no migration needed (No-Go respected).* 
- [x] **6. Security** — pass — *no secrets, eval/exec, or request-derived subprocess input.* 
- [x] **7. Documentation accuracy** — pass — *4 feature docs updated; typed-vs-untyped Popoto bool distinction recorded; stale liveness-check comments corrected.* 
- [x] **8. PR body accuracy** — pass — *claims (3 sites, re-export, typed-bool scope revision, residual limitation) all match the diff.* 
- [x] **9. Disclosed deferrals** — pass — *No-Gos (#2418 wedge-livelock, field-type migration, backlog drain) are explicit out-of-scope with rationale.* 
- [x] **10. Follow-up claims verified** — n/a — *no `filed as follow-up #N` claims; #2418 is a pre-existing separate issue, not a claimed new follow-up.* 

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — PASS — *all delivered; reflection yaml removal is an operational vault edit per Update System.*
- **2. No-Gos from plan — none violated** — PASS — *no field-type migration, no backlog drain, #2418 untouched.*
- **3. New `except Exception` blocks — each has logger/raise/swallow-ok** — PASS — *`publish_session_notify` and both call sites log warnings and swallow by design (save already durable).*
- **4. New integration tests — exercise serialization boundary** — PASS — *`test_notify_isolation` subscribes to the real channel and asserts a real pubsub message lands (no mock).* 
- **5. Plan internal consistency** — PASS — *spike findings (typed-bool safe, `_publish_resume_notify` reusable, actuation unsafe) match the implementation.*
- **7. New public APIs — docstrings present** — PASS — *`publish_session_notify` docstring is exhaustive.*
- **8. Breaking changes — migration path documented** — N/A — *no breaking change.*
- **10. Tests cover the failure path** — PASS — *fail-quiet-on-publish-error tests at every site.*
- **11. UI changes (if any) — screenshot captured** — N/A — *no HTML/CSS/JS/template files changed; `ui/*.py` edits are read-path bool coercion in the `/dashboard.json` data layer, no markup surface — the markup-file visual-proof gate does not trigger.*
- **12. Docs updated for user-facing changes** — PASS — *bridge-self-healing, reflections, redis-models, agent-session-health-monitor, subconscious-memory.*

### Blockers
- None

### Tech Debt
- None

### Nits
- None

### Miscellaneous
- None

- **Bridge-watchdog wedge-restart livelock** — explicit out-of-scope exclusion ([SEPARATE-SLUG #2418], distinct mechanism), no tracking commitment required here.
- **Untyped `Field(default=False)` → typed `Field(type=bool)` migration** — explicit [DESTRUCTIVE] out-of-scope exclusion with sound rationale (existing string-valued Redis records would need a re-cast migration).
- **Draining the ~48 historical stranded pending sessions** — explicit [DESTRUCTIVE] out-of-scope exclusion (fix prevents new strands; backlog drain is a separate review-gated op).

### Review Delta (vs prior review on HEAD before b4f8658c1)
- **Resolved**: `agent/session_health.py` stale `session-liveness-check` comments — the prior review's sole tech-debt item. Comments now read "used to be registered" / "now purely defensive (#2439)" and correctly frame the #2098 owner-guard as defense-in-depth with its out-of-process caller gone.
- **New**: none.
- **Unchanged**: none.

### Verification Results
All plan `## Verification` gates pass on the PR branch: tests green; diff-scoped ruff check + format clean; `publish_session_notify` present in `bridge_watchdog.py`, `sustainability.py`/`circuit_health_gate.py`; `def _truthy` count in `crash_signature.py` == 0; bare `bool(getattr(session,"requires_real_chrome"...))` gone from `ui/data/sdlc.py`; logic-inversion routed through `_truthy`; single `def send_hibernation_notification`; instant-wake marker still present. (Repo-wide `ruff check .` reports pre-existing UP038 debt unrelated to this PR — including two hits on unchanged lines of `session_health.py` that also fail on `main`; not introduced here.)

### Screenshots
- None (no markup files changed; data-layer bool coercion only).

PR #2453, opened 2026-07-29
Hook registration: per-event dispatcher + manifest-generated scopes

valorengels:
## Summary
- Replaces hand-maintained hook registration (23 project `settings.json` entries + hardcoded `_SDLC_HOOK_DEFS`) with a single declarative `.claude/hooks/manifest.toml`, generated into both the project and user `settings.json` scopes.
- Collapses the 7 PreToolUse/Bash validators into one in-process dispatcher (first-block-wins, per-validator fail-open except `validate_merge_guard` which stays fail-closed).
- Detaches the Stop-hook memory/post-merge extraction into a real subprocess so `stop.py` exits immediately instead of racing (and losing to) the 10s harness wall; the detached worker enforces its own `HOOK_DETACH_DEADLINE_SECONDS` deadline and a `HOOK_DETACH_MAX_INFLIGHT` concurrency cap, and logs drops instead of silently swallowing them.
- Fixes four pre-existing sync-discipline bugs: missing `|| true` on the user-scope Stop hook, `hooks_audit.py` auditing only the project scope, no `"hooks"` kind in `RENAMED_REMOVALS`, and a stray `uv run` hook registration inconsistent with its sibling validators.
- Ships a one-time idempotent migration that rewrites the three deployed legacy user-scope entries in place (attaching `manifest_id`, the Stop `|| true` guard, and restoring `sdlc_reminder.py`'s intended `Write|Edit` matcher).
- Phase C cleanup: deletes 7 confirmed-dead unregistered validator scripts, fixes stale claims in `reclassify.md`, and resolves a `validate_commit_message.py` name collision between the project validator and the SDLC fork.

## Changes
- `.claude/hooks/manifest.toml`, `scripts/update/hook_manifest.py` — manifest + loader
- `.claude/hooks/dispatch/pre_tool_use_bash.py` — per-event dispatcher
- `.claude/hooks/stop.py`, `.claude/hooks/hook_utils/stop_detach_worker.py`, `.claude/hooks/hook_utils/detach_lock.py` — detached Stop extraction with deadline + concurrency cap
- `scripts/update/hardlinks.py`, `scripts/update/migrations.py` — generators, removal pass, `RENAMED_REMOVALS` hooks kind, migration
- `reflections/audits/hooks_audit.py` — both-scope audit + agent-hooks scan
- `.claude/hooks/sdlc/validate_commit_message_sdlc.py` — renamed to resolve name collision
- 7 dead validators deleted; `uv run` shebangs normalized on survivors
[...]

Issue #2435, opened 2026-07-28
Hook registration: consolidate behind a per-event dispatcher and generate both scopes from a declared manifest


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/hook-registration-manifest-dispatcher.md

> **Context for readers new to this repo.** Claude Code (Anthropic's CLI coding agent) supports *hooks*: scripts the harness runs automatically at lifecycle events (`UserPromptSubmit`, `PreToolUse`, `PostToolUse`, `Stop`, …). Each hook is registered as a JSON entry in a `settings.json`, carries a `timeout`, and **blocks the agent loop while it runs**. Registration exists at two scopes: *project* (`.claude/settings.json`, checked in) and *user* (`~/.claude/settings.json`, per-machine, applies to every repo).
>
> This repo is the orchestrator for the whole machine: `/update` deliberately propagates skills, agents, and a subset of hooks from here out to `~/.claude/`, so they work in every other repo. That propagation is the design, not an accident — but today it is only partially implemented for hooks, and the hand-maintained half has drifted.

## Problem

Hooks are blocking the agent loop for **~15s per user prompt, ~7s per Bash call, and ~10s per turn end**, and the most expensive one is being killed before it finishes its work.

### Current behavior

Measured from a scan of the 50 most recently modified session transcripts (2026-07-23 → 2026-07-28, 12 project directories), aggregating `attachment` entries of type `hook_cancelled` with `timedOut: true`:

| timeouts | runs | median ms | max ms | script |
|---|---|---|---|---|
| **126** | 131 | 10,034 | 11,548 | `.claude/hooks/stop.py` |
| 17 | 18 | 5,143 | 14,728 | `.claude/hooks/pre_tool_use.py` |
| 14 | 15 | 15,048 | 23,460 | `.claude/hooks/user_prompt_submit.py` |
| 14 | 14 | 7,074 | 14,721 | `validators/validate_merge_guard.py` |
| 9 | 9 | 7,424 | 12,813 | `validators/validate_no_raw_redis_delete.py` |
| 5 | 5 | 10,951 | 12,162 | `validators/validate_no_uv_sync_in_worktree.py` |
| 4 | 4 | 11,807 | 12,022 | `validators/validate_commit_message.py` |
| 2 | 2 | 14,696 | 14,696 | `validators/validate_design_system_sync.py` |
[...]

PR #2454, opened 2026-07-29
Fix SDLC hooks failing to launch in every foreign repo

valorengels:
## Problem

Every Bash/Edit tool call in any repo other than this one emitted:

```
PreToolUse:Bash hook error
Failed with non-blocking status code: /bin/sh: python: command not found
```

The user-scope SDLC hooks in `~/.claude/settings.json` were registered with a bare `python`. Claude Code runs hook commands via `/bin/sh` — non-interactive, never sources `~/.zshenv` — so the `python → python3` alias doesn't apply. This machine has no `python` binary; `python3` only. It resolved inside `~/src/ai` purely because `.venv/bin` is on PATH.

Net effect: the hooks have been dead in every foreign repo, silently.

## Two independent bugs

**1. Bare interpreter** — `scripts/update/hardlinks.py` now emits `python3`. Because dedup is by exact command string, a deployed legacy `python ...` entry is matched and **rewritten in place**; without that, the next `/update` would append a second block and leave the broken original firing alongside it.

**2. PEP 604 annotations under system Python** — surfaced only after fixing #1. `python3` on a stripped PATH is `/usr/bin/python3` = **3.9** on macOS, which evaluates `str | None` at runtime and raises `TypeError` at import. Swapping one failure mode for another. All four scripts are 3.9-syntax-clean, so `from __future__ import annotations` fully resolves it.

## Verification

All hooks exit 0 under the worst case (`env -i`, `/bin/sh`, PATH forced to Python 3.9):

```
sdlc_reminder                exit=0
validate_commit_message      exit=0
validate_sdlc_on_stop        exit=0
```

And functionally, not just non-crashing — staged `.py` commit to `main`:

```json
{"decision": "block", "reason": "Cannot commit code files to main: mod.py. ..."}
```

Applied to this machine's live `~/.claude/settings.json` (backup at `~/.claude/settings.json.bak-hotfix-python3`); diff was exactly 3 command rewrites, no duplicates, no non-SDLC hooks touched.

## Tests

New `tests/unit/test_hardlinks_hook_merge.py` (5 tests) — `_merge_hook_settings` had **zero** coverage before. Covers python3 registration, in-place legacy rewrite without duplication, non-SDLC hook survival, idempotency, and the `__future__` guard.

`tests/unit/test_update_hardlinks.py::test_renamed_removals_entries_are_not_stale` fails on this branch — verified pre-existing on a clean tree (stale `do-skills-audit` entry), unrelated.

## Follow-ups (not in this hotfix)

- **`track_quality_runs.py` was vestigial — removed, not adopted.** It was registered in the deployed settings but absent from `_SDLC_HOOK_DEFS` and from `.claude/hooks/sdlc/` (no repo source, file link count 1). I initially assumed it fed the Stop gate. It does not: `validate_sdlc_on_stop.py` scans the session **transcript**, and the only references to its `/tmp/sdlc_quality_<session_id>` marker are inside the file itself. It wrote a file nobody read, on every Bash call — and was the PostToolUse half of the reported log spam. Deregistered and deleted on this machine; `/update` never managed it, so removal is durable. The `_SDLC_HOOK_DEFS` baseline in `docs/plans/hook-registration-manifest-dispatcher.md` lists three deployed entries and not this one, indicating it was local to this machine.
- **`validate_commit_message.py` gates on `repo_name != "popoto"`** (line 74) — its module docstring claims general "block code commits to main" behavior it does not implement.
- **`docs/plans/hook-registration-manifest-dispatcher.md`** rewrites these same command strings via a migration. That plan should absorb the `python3` literal, or the two will fight.
- `.opencode/SYNC_MANIFEST.json` hashes for the four scripts are now stale; the next sync advances them. Deliberately left out to keep this diff minimal — running the sync also pulled in unrelated pre-existing drift.

### DIFF
--- a/.claude/hooks/sdlc/sdlc_context.py
+++ b/.claude/hooks/sdlc/sdlc_context.py
@@ -10,6 +10,8 @@ system. The AgentSession import is optional — it falls back gracefully to
 branch-only detection when Redis or the AI repo isn't available.
 """
 
+from __future__ import annotations
+
 import json
 import os
 import subprocess
--- a/.claude/hooks/sdlc/sdlc_reminder.py
+++ b/.claude/hooks/sdlc/sdlc_reminder.py
@@ -18,6 +18,8 @@ Claude Code hook protocol:
   Advisory: print message to stdout, exit 0
 """
 
+from __future__ import annotations
+
 import hashlib
 import os
 import sys
--- a/.claude/hooks/sdlc/validate_commit_message.py
+++ b/.claude/hooks/sdlc/validate_commit_message.py
@@ -19,6 +19,8 @@ Claude Code hook protocol:
   To ALLOW: print nothing, exit 0
 """
 
+from __future__ import annotations
+
 import os
 import subprocess
 import sys
--- a/.claude/hooks/sdlc/validate_sdlc_on_stop.py
+++ b/.claude/hooks/sdlc/validate_sdlc_on_stop.py
@@ -22,6 +22,8 @@ Claude Code hook protocol:
   Stdin: JSON with session_id, transcript_path (JSONL file), cwd, etc.
 """
 
+from __future__ import annotations
+
 import os
 import re
 import subprocess
--- a/scripts/update/hardlinks.py
+++ b/scripts/update/hardlinks.py
@@ -686,6 +686,13 @@ def sync_user_hooks(project_dir: Path) -> HardlinkSyncResult:
     return result
 
 
+# Claude Code runs hook commands via /bin/sh, which is non-interactive and never
+# sources ~/.zshenv -- so the `python -> python3` shell alias does not apply and a
+# bare `python` resolves only when a virtualenv happens to be on PATH. These hooks
+# are registered at *user* scope and fire inside every repo on the machine, most of
+# which have no such venv. Must stay an interpreter name that resolves bare.
+_HOOK_INTERPRETER = "python3"
+
 # SDLC hook definitions for ~/.claude/settings.json.
 # Each tuple: (hook_event, matcher, script_name, timeout)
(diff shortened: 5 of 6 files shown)


valorengels:
Closing in favor of #2503.

The problem statement here is sound and the diagnosis was credited in the new issue, but this diff is no longer applicable. PR #2453 (issue #2435) landed on 2026-07-31 and moved hook registration to `.claude/hooks/manifest.toml`:

- `_SDLC_HOOK_DEFS` was deleted; this PR's `_HOOK_INTERPRETER` + `_SDLC_HOOK_DEFS` block lands in removed territory.
- `_merge_hook_settings` changed signature and switched from exact-command-string dedupe to `manifest_id` marker identity, which makes this PR's core mechanism (rewrite-not-duplicate the legacy bare-`python` string) obsolete by construction.
- `.claude/hooks/sdlc/validate_commit_message.py` was renamed to `validate_commit_message_sdlc.py`.
- `tests/unit/test_hardlinks_hook_merge.py` imports `_SDLC_HOOK_DEFS` at module scope and now ImportErrors on main.

Two corrections worth carrying forward rather than re-deriving:

1. **`uv run` is forbidden.** `scripts/update/hooks.py:99` audits for and fails on it, and #2435 deliberately removed the repo's only `uv run` hook registration. Bare `python` was the *outcome* of that fix, not an oversight.
2. **`python3` is not a safe swap either.** Under a stripped PATH it resolves to 3.12.6 on this machine but 3.9.6 on a clean mac — it would pass locally and break fleet-wide.

The underlying bug is still live on main. #2503 carries it forward with the wider scope recon uncovered (the PEP 604 crash fires at *module import*, before `main()`'s fail-open guard; and 5 files import `datetime.UTC`, 3.11+, which `from __future__ import annotations` cannot fix).

Issue #2140, opened 2026-07-17
valor-session create never sets classification_type — CLI-created SDLC sessions show no stage progress and may skip auto-continue


tomcounsell:
> **Context:** SDLC pipeline progress is tracked in two stores: an issue-keyed [`PipelineLedger`](agent/pipeline_ledger.py) written by `sdlc-tool stage-marker` during pipeline runs, and a session-keyed `AgentSession.stage_states` JSON field that the web dashboard (`ui/app.py:674`) reads to render `current_stage`/`stages`. Sessions can be created two ways: via the Telegram bridge (which classifies the message) or via the CLI — `python -m tools.valor_session create` ([docs/features/session-isolation.md](docs/features/session-isolation.md)). This bug was found live: three CLI-created SDLC sessions were observed running a full pipeline with `ISSUE/PLAN/CRITIQUE` completed in the ledger while the dashboard showed no stage progress at all.

## Problem

`tools/valor_session.py create` never sets `classification_type` on the sessions it enqueues (the file contains no reference to classification). Everything downstream that keys off `ClassificationType.SDLC` silently degrades.

**Current behavior:**
- Enqueue-time `stage_states` initialization is gated on `classification_type == ClassificationType.SDLC` (`agent/agent_session_queue.py:358-375`, comment: *"so the dashboard shows pipeline progress from the start"*). CLI-created sessions skip it, so `AgentSession.stage_states` stays empty and the dashboard renders `current_stage: None, stages: []` for the entire run — even while `sdlc-tool stage-query --issue-number N` shows `ISSUE: completed, PLAN: completed, CRITIQUE: completed` from the issue-keyed ledger. Observed live on session `0_1784286820434` (issue #2133 pipeline).
- The output router's auto-continue rule matches `session_type == "eng" and classification_type == "sdlc"` → `nudge_continue` (`agent/output_router.py:158`). With `classification_type=None`, a turn-end status update falls through to `deliver` instead — on a bridge machine that means the pipeline may pause as if awaiting a human instead of auto-continuing. (Behavioral impact needs verification; the headless PM runner's structured `route` field may mask it.)
- `issue_url` is also never populated by the CLI path (`None` on the observed sessions) even when the message contains the issue URL, so the dashboard cannot link the session to its issue or its ledger-side stage state.

**Desired outcome:**
[...]

Issue #2051, opened 2026-07-13
do-build/do-merge phantom-wait under forked /do-sdlc: stage subagents await a notification a fork can't receive, stranding the pipeline


tomcounsell:
> **Context.** A local `/do-sdlc` run executes stages by spawning subagents (via the Agent/Task tool). When `/do-sdlc` is invoked as a **forked** skill execution, the fork runs to a stopping point and returns to its parent; its subagents report back via task-notifications. A stage subagent that finishes its work but then **stops to "wait for a background notification"** strands the pipeline, because a fork has no resume turn in which to receive that notification.

## Problem

The `/do-build` and `/do-merge` stage subagents (under forked `/do-sdlc`) repeatedly **phantom-wait**: they background a piece of work (a builder child, or a full `pytest` run), then end their turn with a message like *"waiting for the build to complete"* / *"waiting for the test run to finish"* — a notification that never lands because the fork can't be resumed. The stage's actual work is often already done; the wait is spurious, and the pipeline stalls until a supervisor notices and resumes the subagent synchronously.

### Current behavior
- A `/do-build` or `/do-merge` subagent spawns a background child (`run_in_background`) or a `Monitor`, then returns "waiting for X."
- No resume turn fires for a fork → the stage sits indefinitely as a completed-looking-but-stalled notification.
- Recovery is manual: the supervisor reconciles git/PR ground truth and re-dispatches the subagent with a **synchronous** mandate ("no background children, no Monitor, no `run_in_background`; record the result in-turn"), which completes reliably.

### Desired outcome
Stage subagents run their work synchronously (or the skill/harness resumes a backgrounded stage automatically), so no stage strands on a wait a fork cannot satisfy. A "waiting for a notification" terminal message from a stage subagent should be impossible by construction.

## Recurrence evidence

Observed across every lane of a 2026-07-12/13 multi-wave push:
- **#1925**: `/do-build` and `/do-merge` each backgrounded work and returned "waiting for a notification"; recovered by synchronous resume.
- **#1927**: three stage subagents (build + two review passes) stalled on background children/monitors; each recovered only after a synchronous re-dispatch. The reliable fix each time was a subagent brief with hard constraints: no Agent tool, no Monitor, no `run_in_background`, record the result in-turn.
- **#1968**: one BUILD phantom-wait on a background `pytest`; the consolidation work was already done, the wait was spurious.

[...]

Issue #1915, opened 2026-07-06
do-sdlc/do-build fork: background exits strand pipelines; parallel builders share one slug worktree (6-issue batch findings)


tomcounsell:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/sdlc-fork-sync-workers-worktree-isolation.md

## Summary

A supervised 6-issue parallel SDLC batch (2026-07-05: #1898 #1899 #1901 #1902 #1904 #1905, all shipped) surfaced two systematic dispatch defects in the renovated fork path. Every pipeline eventually merged, but 4 of 6 required manual supervisor recovery.

## Defect 1: fork exits its turn "waiting" on background children (phantom-wait)

Observed in at least 4 pipelines, at different stages:
- #1904: fork armed a background PR-poll and returned "awaiting notifications", leaving the build uncommitted with no PR.
- #1901: fork returned at MERGE reporting "merge agent + ground-truth poll running in the background".
- #1902: do-build's completion summary claimed it "launched the full unit test suite in the background and set up a monitor".
- #1898: do-build returned "Builder B running in the background, I'll continue when it completes" — a forked skill cannot continue; it left an unpushed branch, no PR, and uncommitted edits.

A fork that ends its turn cannot be resumed by its children's completion. Stage skills that spawn workers must drive them synchronously (run_in_background: false) or the fork must not exit until terminal state.

## Defect 2: parallel builders spawned into ONE shared slug worktree; no seam for an assigned worktree

The fork derives worktree+branch from the plan slug (.worktrees/{slug}, session/{slug}) and ignores externally designated lanes (supervisors' .worktrees/sdlc-{N} instructions never propagate). When a plan has parallel build tasks, all builders share that single worktree and git index. Consequences observed:
- #1904: builder A's files swept into builder B's commit; B then soft-reset and recommitted only its own files, orphaning A's work while A exited reporting success. Two divergent versions of one file needed manual reconciliation.
- #1899: git state mutated between supervisor inspections as two builders committed concurrently; an unrelated issue's commit (docs/plans/update-verify-running-release.md) leaked onto the shared branch mid-flight.
- #1898: a concurrent session kept committing to .worktrees/sdlc-1898 while the owner worked; one commit snapshotted the other's mid-edit tree.

## Compounding effect: duplicate PRs

[...]

PR #2455, opened 2026-07-29
PM fire-and-forget: dev subagent promise/wake handling (#2420)

### LINKED ISSUE #2420: PM fire-and-forget: conversationally-classified eng session backgrounds a dev subagent, acks with a promise, and can never be woken
> **Context:** This repo runs bridge-originated work through an **Eng session** — a long-lived agent session (`session_type="eng"`) that acts as a project manager (PM), delegating implementation to a `dev` subagent. Each PM turn is one `claude -p` subprocess managed by the [headless session runner](docs/features/headless-session-runner.md). When a PM turn ends, the [output router](agent/output_router.py) decides whether to **deliver** the output to the human (terminating the session) or **nudge** it (re-enqueueing so the PM gets another turn). This issue is about a session that delivered when it should have nudged, stranding completed work.

## Problem

A PM session can spawn a background subagent, immediately reply "I'll follow up," and then end its turn permanently. Nothing in the system can give it the turn it promised. The work completes invisibly and the human is never told.

**Current behavior:**

Traced from session `tg_psyoptimal_-1003743854645_413` (2026-07-26, one turn, $0.38):

| Time (UTC) | Event |
|---|---|
| 23:44:46 | Session picked up, PM turn starts |
| 23:45:09 | PM calls the `Agent` tool, `subagent_type: dev`, in **background** mode |
| 23:45:16 | PM emits `StructuredOutput {route: "user", message: "On it — ... I'll follow up with the PR once it's ready."}` — **turn ends, 7 seconds in** |
| 23:48–23:54 | The `dev` subagent keeps working inside the still-alive process: opens a PR, force-pushes, runs tests (26 pass), runs the review gate (0 blockers), runs `/do-docs` (no-op) |
| 23:54:50 | `dev` invokes `/do-merge`, reads PR state: `MERGEABLE` / `CLEAN` |
| 23:55:16 | Process dies — roughly **one tool call before the merge** |
| 23:55:19 | The ack queued 10 minutes earlier is flushed to Telegram |
| 23:55:20 | Session marked `completed` |

The `[unmerged-branch-guard]` preserved the branch at 23:55:22, which is the only reason the work survived at all. The PR sat open and unmerged; the human's only signal was an acknowledgement promising a follow-up that could never arrive.

**Root cause:** [`agent/output_router.py:159`](agent/output_router.py#L159) — `determine_delivery_action` returns `"nudge_continue"` only when `session_type == "eng"` **and** `classification_type == "sdlc"`. This session was classified conversational (a chat message asking for a small content change), so it fell through to `"deliver"` → terminal status → no wake path.

**Desired outcome:** A PM session that has backgrounded real implementation work cannot reach a terminal state until that work is accounted for — either it is prevented from backgrounding such work, or it gets a wake path equivalent to `nudge_continue`.

## Definitions

| Eng session / PM session | Long-lived agent session (`session_type="eng"`) that orchestrates work and delegates implementation to a `dev` subagent | [docs/features/eng-session-architecture.md](docs/features/eng-session-architecture.md) |
| `classification_type` | Label assigned to an inbound message deciding whether it is SDLC (pipeline) work or conversational | [agent/output_router.py](agent/output_router.py) |
| Nudge / auto-continue | Re-enqueueing a session so the agent gets another turn instead of the output being delivered to the human as final | [docs/features/bridge-worker-architecture.md](docs/features/bridge-worker-architecture.md) |
| `StructuredOutput {route: "user"}` | The mechanism a PM uses to emit a human-facing message; emitting it ends the turn | [agent/session_runner/adapter.py](agent/session_runner/adapter.py) |

## Prior Context

This failure family is well-trodden; each prior fix addressed a different root cause and none covers this one.

- #2139: *Promise gate can detect 'I'll report back' but nothing can fulfill it — add a schedulable check-in primitive* (CLOSED) — added an opt-in wake primitive. Does not help a session that never calls it.
- #2140: *valor-session create never sets classification_type* (CLOSED) — **same code line**, failing because the CLI path left `classification_type` unset. Here the classification was arguably *correct*; the router simply has no rule for this case.
- #1915 / #2051: fork-based stage subagents phantom-wait for notifications a fork cannot receive (both CLOSED) — mitigated with prose rules ("don't use `run_in_background`") rather than an enforced invariant.
- #1005: *PM session completes before merge gate — PR left open and unmerged* (CLOSED) — same symptom via the now-obsolete `[PIPELINE_COMPLETE]` marker.
- #1156: *waiting_for_children → completed bypasses child-liveness gate* (CLOSED) — parent terminates while a child lives, but via a different channel; this session was never in `waiting_for_children`.

Filed alongside two sibling defects found in the same trace: the promise gate not firing on the ack, and a merge-guard cross-repo blind spot.

## Solution Sketch

The root cause is understood, but the correct remedy is a design decision — recording open questions rather than a concrete approach, so `/do-plan` genuinely evaluates them:

1. **Should a conversationally-classified eng session be allowed to background implementation work at all?** Prior fixes (#1915, #2051) chose prose prohibition. Prose has now failed at least three times, which argues for an enforced invariant.
2. **If backgrounding is allowed, what wakes the PM?** Options include treating "a background subagent is alive" as a nudge condition independent of `classification_type`, or reusing the `waiting_for_children` state (noting #1156's bypass).
3. **Should classification be re-evaluated mid-turn?** The message was conversational on arrival but became SDLC-shaped the moment the PM spawned `dev`. A terminal routing decision made from arrival-time classification may be the deeper flaw.
4. **What is the correct terminal behavior when a process dies with work in flight?** Here the session was marked `completed` despite an interrupted merge; `completed` was factually wrong.

Constraint: whatever wake path is chosen must not resurrect the phantom-wait failures of #1915/#2051, where sessions blocked forever awaiting a notification that could not arrive.

## Recon Summary

- Root cause line `agent/output_router.py:159` read directly; the `session_type`/`classification_type` conjunction is verbatim as described.
- Full timeline reconstructed from `logs/worker.log`, `logs/session_telemetry/`, and the raw PM transcript (`~/.claude/projects/.../08f205e4-*.jsonl`), including the background `Agent` tool call and the `StructuredOutput` payload.
- The `dev` subagent reached `/do-merge` and read a `MERGEABLE`/`CLEAN` PR state ~26 seconds before the process died — the work was one step from done.
- All 14 cross-referenced issue/PR numbers verified to exist with matching titles.

- Initially framed as "missing auto-continue nudge." Widened after recon: #2140 shows the same line already failed once for a different reason, so a point-fix to the conditional is likely insufficient. Reframed around the routing rule's missing case plus the classification-staleness question.

**Pre-requisites:** 0 items

- The cause of the process exit at 23:55:16 (624s after turn start). No timeout in `agent/session_runner/harness/claude.py` matches; most plausibly the CLI's own background-agent drain window in `-p` mode. **Unverified — deliberately excluded** rather than asserted. Worth a separate investigation issue if it recurs.

## Acceptance Criteria

- [ ] A PM session that spawns a background subagent and emits a user-facing message does not reach a terminal status while that subagent is still running.
- [ ] A regression test reproduces the trace: eng session + non-SDLC `classification_type` + backgrounded `dev` subagent → asserts the session is not terminal after the PM's first turn ends.
- [ ] When a session process dies with a subagent mid-task, the resulting status is not `completed`.
- [ ] The chosen remedy is documented, including why it does not reintroduce the #1915/#2051 phantom-wait failure.
- [ ] If backgrounding is prohibited rather than supported, the prohibition is enforced in code, not prose.

## Downstream

Constraint for the planner: the Solution Sketch is deliberately a set of open questions, not a chosen approach. Question 1 (prohibit vs. support backgrounding) should be settled before any implementation is designed.

### LINKED ISSUE #2527: Non-blocking hooks' deny path is swallowed by `|| true` in generated settings.json
## Problem

Every hook declared `blocking = false` in `.claude/hooks/manifest.toml` has its **deny path silently neutralized** in the generated `.claude/settings.json`.

`scripts/update/hardlinks.py:777-778`:

```python
if not decl.blocking:
    command += " || true"
```

Deployed result for `pre_tool_use.py` (PreToolUse):

```
"$CLAUDE_PROJECT_DIR"/.claude/hooks/hook_python "$CLAUDE_PROJECT_DIR"/.claude/hooks/pre_tool_use.py || true
```

Shell semantics, confirmed:

```
sh -c 'exit 2'           -> 2   (Claude Code DENY)
sh -c 'exit 2' || true   -> 0   (Claude Code ALLOW)
```

A hook that calls `sys.exit(2)` prints its block message to stderr and then reports success. The tool call proceeds.

## Impact

**`_enforce_tool_budget` (`pre_tool_use.py:292`) cannot block today.** Its `sys.exit(2)` is swallowed. The per-tool budget backstop appears wired but is inert at the choke point.

This was found while reviewing PR #2455, whose Layer 1 foreground-guard depends on the same exit-2 path and is therefore also inert as registered. That PR did not introduce the defect and should not have to carry the infrastructure fix alone.

## Why it went unnoticed

Hook tests construct `hook_input` and call the enforcement function directly, asserting on `SystemExit`. None exercises the **deployed command string**, so the `|| true` wrapper is invisible to CI.

## The tension to resolve

`|| true` exists for a reason: a crashing hook should not wedge every tool call. But "do not let an unexpected crash block" and "do not let a deliberate deny block" have been collapsed into one flag. They are different:

- exit 2 = deliberate, structured deny → must propagate
- exit 1 / traceback / timeout = hook bug → should not block

Suggested direction: stop using `|| true` as the fail-open mechanism and let hooks own their own fail-open internally (`pre_tool_use.py` already does this correctly — `_foreground_guard_fail_open` catches internal errors and returns, while the deny's `sys.exit(2)` deliberately lives outside the `try`). A wrapper that maps only *unexpected* non-zero exits to 0 while passing 2 through would preserve both properties.

## Acceptance

- [ ] A deliberate exit-2 deny from a non-blocking hook actually blocks the tool call
- [ ] An unexpected hook crash still fails open
- [ ] A test asserts on the **generated command string**, not just the Python function, so this cannot regress silently
- [ ] Audit which existing hooks have been inert and for how long

valorengels:
Addresses issue #2420 (PM fire-and-forget: an eng session backgrounds a dev subagent, acks with a promise, and can never be woken).

Closes #2420

## Status change

This PR was recovered from a local-only branch during a repo cleanup and opened as a draft that had never been rebased, tested, or reviewed. All three of those have now happened.

## Rebase: verified clean

`git rebase origin/main` applied all 10 commits with **zero conflicts**, and the post-rebase diff is byte-identical to the pre-rebase one. The "103 commits behind" figure was misleading: exactly one of those commits touched any file this branch modifies, and it was disjoint. `.claude/hooks/pre_tool_use.py` and `agent/session_runner/router.py` had seen zero lines of churn.

## Tests: run for the first time

| Suite | Result |
|---|---|
| The PR's own three test files | **61 passed** |
| `tests/unit/session_runner/` (full dir) | **333 passed** |
| `tests/unit/hooks/` | **63 passed** |
| `ruff check` / `ruff format --check` | clean |

This clears the one risk the assessment could not: main's new spawn-stamping did not require the test stubs to grow attributes.

## The problem is still real and unowned

Verified against current main, not against the plan:

- `agent/session_runner/router.py` — `PM_USER = ("pm_user", True, True, False)`, still `is_clean=True`.
- `git grep run_in_background` over `.claude/hooks/` and `agent/` returns **0 matches**. No enforcement exists at the tool boundary.
- No supersession anywhere: `PM_USER_SUBAGENT_LIVE`, `subagent_in_flight`, and "foreground" in session_runner are all absent from main.
- The only prohibitions that exist are prose in fork skills, which is the layer that already failed three times (#1915, #2051, #2140) and which does not bind a top-level PM turn spawning its inline `dev`.

## New in this revision: the do-test collision

Layer 1 denies `run_in_background: true` in eng sessions. `.claude/skills-global/do-test/parallel-dispatch.md` was still issuing exactly that at lines 35 and 52, and `/do-test` is invoked by the SDLC pipeline **from eng sessions** — so shipping Layer 1 alone would have broken it.

This was not drift. The file was present at the merge base; the plan simply missed it, and the review task that should have caught it never ran.

Fixed by migrating `do-test` to the same-message foreground pattern that `do-build`, `do-sdlc`, `do-plan-critique`, and `do-pr-review` already use. Parallelism is preserved (the harness runs same-message foreground Tasks concurrently); the fork-has-one-turn failure mode of #1915 is not reintroduced. Step 3's polling loop and 2-minute timeout are replaced by a per-group direct-execution fallback that must name which groups fell back, so a partial dispatch cannot read as a full parallel run. `do-test`'s two `run_in_background: true` sites were the last *explicit* background dispatchers in the fleet.

**Correction (review, 2026-08-04):** an earlier version of this line claimed `do-test` was the last background dispatcher outright. That was wrong. It was based on grepping `run_in_background: true`, which cannot see a `Task({...})` template that omits the key entirely — and the Layer 1 predicate denies omission, not just an explicit `true`. At least `do-test/special-targets.md`, `do-test/baseline-verification.md`, `docs/sdlc/do-plan-critique.md`, and the always-loaded eng surface (`config/personas/engineer.md`, `.claude/agents/dev.md`, `.claude/commands/roles/prime-pm-role.md`) all spawn flaglessly. See the review comment.

## Quality note on the original build

The plan's HARD GATE was genuinely executed rather than time-boxed away. The shipped `subagent_in_flight` helper handles `stop_reason is None` (streaming CLI flush), a trailing `tool_use` content block, and a `user` record carrying `tool_result` — cases you only find by reading real on-disk transcripts. It is not the always-`False` no-op the critique warned would silently negate Layer 2. Layer 2's wiring is also scoped on `summary.exit_reason.is_clean` rather than two literals, which was the fix from its earlier critique round.

## Review blockers: resolved

REVIEW returned CHANGES REQUESTED with three interacting blockers. All three are fixed in `215529bc`, together, as the review required.

Closes #2527

### 1. `subagent_in_flight` now detects the just-spawned window

The shipped predicate inferred liveness from a content-shape snapshot and read `message.content` as a block list. The task-prompt record — a `user` record whose content is a plain **string** in 1571 of 1573 real sidechains — yielded no blocks and fell through to `False`, so it missed 100% of just-spawned subagents.

The predicate is inverted: in flight when the transcript is **both** being written to (mtime within `SUBAGENT_LIVE_MTIME_WINDOW_S`) and **not** closed by a finished assistant answer. Both halves are load-bearing. Recency alone flags a subagent that finished a second ago; shape alone flags every stranded transcript from an earlier turn, since sidechain files accumulate under the claude session id rather than the turn. Every subagent is checked instead of `ids[-1]`, so a live dev-A behind a more-recently-finished dev-B is seen.

Measured by replaying 1573 real on-disk sidechains:

| Measurement | Shipped | Now |
|---|---|---|
| Truncated to first record (just spawned, definitively live) | 0% detected | **100.0%** (1573/1573) |
| Every proper prefix (uniformly sampled live moment) | 60.4% | **87.4%** (120223/137539) |
| Complete transcript, stale mtime (earlier turn) | — | 0.0% flagged |
| Closed on a finished answer, fresh mtime | — | **0 of 1455** flagged |

The 118 transcripts flagged at full length are exactly the 118 that end mid-exchange: killed, interrupted, or stranded. None closed on a final answer.

`tests/unit/session_runner/test_subagent_in_flight.py:138,149` enshrined two false-negative shapes as correct. Both assertions are corrected, and the just-spawned shape is now its own test class.

### 2. The deny survives registration

`blocking` collapsed two different intents into one boolean. Exit 2 is Claude Code's structured deny; every other non-zero exit is a hook crash; `blocking = false` mapped both to 0. `_enforce_tool_budget`'s deny had been inert for as long as the manifest existed.

The field is replaced by `exit_policy`: `"propagate"` (bare), `"deny-only"` (a crash maps to 0, exit 2 passes through), `"suppress"` (`|| true`). `pre_tool_use` declares `deny-only`. The two Stop validators stay `suppress`, now as a stated decision rather than an accident — a Stop-event exit 2 blocks turn-end, and the standing decision is that a missed quality gate warns.

`.claude/settings.json` was regenerated by `sync_project_hooks`, never hand-edited. Exactly one command string changed.

`tests/unit/test_update_hardlinks.py` now runs each generated command through a real shell and asserts the shell's exit code across policies and hook exit codes. That is the check whose absence let the wrapper stay invisible to CI.

### 3. Blast radius closed at the surface, and the test widened

Probed against a live `claude -p` run: a `Task` call issued without the flag reaches the PreToolUse hook with `run_in_background` **absent** from `tool_input`. The harness injects no default, and the tool's contract is that an absent flag means background. Treating omission as foreground would admit the exact #2420 dispatch shape, so the predicate is right and the dispatch surface is what needed fixing.

Our own sweep found three flagless templates, including `do-patch/SKILL.md`, which the review did not name. All three now carry the flag, and the rule is stated on the always-loaded eng surface (`config/personas/engineer.md`, `.claude/agents/dev.md`, `prime-pm-role.md`, `prime-dev-role.md`, `docs/sdlc/do-plan-critique.md`).

`tests/unit/test_sdlc_fork_no_background.py` globbed `**/SKILL.md` across 2 named files. It now discovers every `Task({` / `Agent({` template across both skill trees, `.claude/agents/`, `.claude/commands/`, `config/personas/`, and `docs/sdlc/`, and asserts per template so one correct template cannot vouch for a flagless sibling. Replayed against the pre-fix tree it fails at exactly the three offending sites.

### Nits

The do-test fallback triggered only on an error return, so a hanging subagent stalled to the 7200s turn cap. A `GROUP_TIMEOUT` hang bound is carried verbatim into each dispatch prompt: foreground dispatch means the parent blocks, so only the child can end its own hang. `docs/features/sdlc-fork-turn-boundary.md` now cites lines 26 and 135.

## Tests

Measured after merging current `origin/main` into the branch (the branch is `MERGEABLE` / `CLEAN` at `049eb7fa`).

| Suite | Review baseline | Now |
|---|---|---|
| `tests/unit/session_runner/` | 313 | **327** |
| `tests/unit/hooks/` | 63 | **63** |
| Combined | 376 | **390** |
| Hook-registration and dispatch-surface suites | — | **241** |
| `ruff check` / `ruff format` | clean | clean |

The +14 delta is all new coverage, and no test was removed. Nine come from this branch's `test_subagent_in_flight.py`: the just-spawned class, the mtime window, and the every-agent-checked cases. The other five arrived with `origin/main`.

The merge conflict was purely additive in `tests/unit/test_update_hardlinks.py` (both sides appended a block at end of file). Both blocks are kept, and main's new `_global_decl` helper from #2561 carries onto the `exit_policy` field.


tomcounsell:
## Review: CHANGES REQUESTED

Reviewed at head `2e43cb5c`. Suite baseline reproduced: `tests/unit/session_runner/` + `tests/unit/hooks/` = 376 passed (313 + 63), ruff clean. Three blockers; they interact and must land together.

### BLOCKER 1 — `subagent_in_flight` misses the exact window the bug occurs in

`agent/session_runner/adapter.py` `_record_shows_pending_tool_exchange` returns `True` on only three shapes: assistant with `stop_reason == "tool_use"`, assistant with a `tool_use` block, user with a `tool_result` block. Everything else returns `False`.

A subagent that is **thinking or streaming text is in flight**, and every one of those records falls through to `False`. Worse, `blocks = content if isinstance(content, list) else []` means a record whose `message.content` is a plain string yields no blocks at all.

Measured by replaying real on-disk transcripts through the shipped predicate, independently on two corpora:

| Measurement | Reviewer (sidechains) | Reproduced here |
|---|---|---|
| Truncated to first record (just spawned, definitively live) | 400/400 → `False` | **299/299 → `False` (100%)** |
| Every proper prefix (all live moments) | 39.6% → `False` | 49.8% → `False` |

The just-spawned case is the fire-and-forget window itself: PM spawns a background dev, the transcript holds only the task-prompt record, PM acks and exits clean, Layer 2 sees "not in flight" and does not downgrade. It misses **100%** of the time.

The plan (line 202-218) named this a HARD GATE and its tolerance was an explicitly *rare* miss (line 283). This is not that.

`tests/unit/session_runner/test_subagent_in_flight.py:138,149` enshrine two of the top false-negative shapes as correct ("ambiguous → fail-safe False"), so the suite cannot catch this.

**Root cause:** liveness is inferred from a content-shape snapshot of an append-only log. Transcript **mtime recency** is the unused robust signal — a live subagent's file is being appended to.

### BLOCKER 2 — Layer 1's deny is swallowed by `|| true`

The hook logic is correct; the registration neutralizes it. `.claude/hooks/manifest.toml` declares `blocking = false` for `pre_tool_use.py`, and `scripts/update/hardlinks.py:777-778` turns that into a shell suffix. The deployed command in `.claude/settings.json` is verbatim:

```
"$CLAUDE_PROJECT_DIR"/.claude/hooks/hook_python "$CLAUDE_PROJECT_DIR"/.claude/hooks/pre_tool_use.py || true
```

Confirmed: `sh -c 'exit 2'` → 2; `sh -c 'exit 2' || true` → 0. The `sys.exit(2)` fires, stderr prints, the shell returns 0, Claude Code allows. **AC1 and AC5 are not met as shipped.**

This is a pre-existing infrastructure defect this PR did not introduce — it equally neuters the existing `_enforce_tool_budget` deny. Filing separately. But Layer 1 depends on it entirely, and every hook test calls the function directly, so none exercises the deployed command string.

### BLOCKER 3 — Blast radius is wider than the one collision already fixed

The predicate allows only an **explicit** `run_in_background: false` (`pre_tool_use.py:372`); **omitting the key is denied**. So every flagless `Task({...})` template in the eng surface is reachable.

- `.claude/skills-global/do-test/special-targets.md:19` and `baseline-verification.md:79` — both flagless, both loaded by the same skill in the same eng session. **This falsifies my own claim in the PR body that do-test was the last background dispatcher in the fleet.** I had grepped `run_in_background: true`, which cannot see a template that omits the key.
- `docs/sdlc/do-plan-critique.md:126` instructs "(foreground, no `run_in_background`)" — omit the key, which is exactly what gets denied.
- `config/personas/engineer.md`, `.claude/agents/dev.md`, and `.claude/commands/roles/prime-pm-role.md` never name the flag (0 hits each) while mandating parallel fan-out. `prime-pm-role.md` is the primary production dispatch path.

`tests/unit/test_sdlc_fork_no_background.py:265-268` globs only `**/SKILL.md` across 2 named files, so every sub-file above is invisible to CI. Widening it is the durable fix.

**Findings 2 and 3 interact:** none of this fires today *because* of Finding 2. Fixing the registration without first fixing the surface would break a large fraction of eng dispatch on the next `/update`.

### What is correct and should survive

- **Chokepoint wiring (`runner.py:872-883`) is right.** After the wrap-up guard and `_requeue_pending_steers()`, immediately before `publish_exit_summary`, scoped on `summary.exit_reason.is_clean` rather than literals. Single site, cannot double-fire or miss a clean-exit path. `PM_USER_SUBAGENT_LIVE` correctly flows to `"failed"`. This properly fixed the prior critique's blocker.
- **The hook's fail-open/deny split is correct in both directions.** The module wrapper's `except Exception` cannot catch `SystemExit`, so the deny propagates while internal bugs fail open.
- **The responsiveness test is honest work.** It samples from outside the runner mid-spawn and asserts strict ordering; `exit_reason is PM_EMPTY_TURN` closes the vacuous-pass route. Two limits: it calls `adapter.on_user_payload` directly rather than driving a real `route:"user"` block, and its "foreground Task" is an `asyncio.Event.wait()` with no connection to Layer 1.

### Nits

- `adapter.py:292` — `latest = ids[-1]` checks only the newest subagent by mtime. If dev-A is live but dev-B finished more recently, only dev-B is inspected. Consider `any(...)`.
- do-test fallback loses one real property: the old 2-minute timeout bounded a **hang**; the new fallback triggers only on an error return. A hanging subagent now stalls to the 7200s turn cap.
- `docs/features/sdlc-fork-turn-boundary.md:19-20` cites `do-sdlc/SKILL.md:24,99`; actual lines are 26 and 135.

tomcounsell:
## Blockers resolved at `215529bc`

All three landed together. Evidence per blocker below; the PR body carries the same summary.

### BLOCKER 1 — measured on the same corpus, the same way

Replayed 1573 real on-disk sidechains through the rewritten predicate:

| Measurement | Shipped | Now |
|---|---|---|
| Truncated to first record (just spawned, definitively live) | 0% detected | **100.0%** (1573/1573) |
| Every proper prefix (uniformly sampled live moment) | 60.4% | **87.4%** (120223/137539) |
| Complete transcript, stale mtime (earlier turn) | — | 0.0% flagged |
| Closed on a finished answer, fresh mtime | — | **0 of 1455** flagged |

Your root cause was right, and the corpus made the shape explicit: **1571 of 1573 sidechains open with a `user` record whose `message.content` is a plain string** — the task prompt. `blocks = content if isinstance(content, list) else []` discarded exactly the record that proves a subagent is live.

The predicate is inverted rather than extended. In flight means the transcript is **both** being written to (mtime within `SUBAGENT_LIVE_MTIME_WINDOW_S`, 900s) and **not** closed by a finished assistant answer. Both halves are required: recency alone flags a subagent that finished a second ago, and shape alone flags every stranded transcript from an earlier turn, since sidechain files accumulate under the claude session id rather than the turn.

On the false-downgrade worry: it did not materialize. Of the 1455 transcripts that closed on a finished answer, **zero** are flagged even with a freshly stamped mtime. The 118 that are flagged at full length are precisely the 118 that end mid-exchange. Completion is keyed on an assistant record carrying a `text` block with no `tool_use` and `stop_reason != "tool_use"`; `stop_reason` is otherwise ignored, since 207 finished transcripts close on `None` and 23 on `stop_sequence`.

`ids[-1]` is now `any(...)` over every subagent, with a test for the dev-A-live-behind-dev-B-finished case you named.

The two assertions at `test_subagent_in_flight.py:138,149` are corrected, not worked around. A trailing thinking-only record and a trailing plain user text both mean the assistant has not answered; 77 real transcripts end on the latter and every one was interrupted.

### BLOCKER 2 — `exit_policy`, and a test that runs the generated string

Taking the direction from #2527: the two intents are separated rather than sharing one flag. `blocking` is replaced by `exit_policy` with `"propagate"` / `"deny-only"` / `"suppress"`, and `_EXIT_POLICY_SUFFIX` maps each to its shell suffix. Verified against a real shell:

```
deny-only, hook exits 0 -> 0    suppress, hook exits 2 -> 0
deny-only, hook exits 1 -> 0    propagate, hook exits 2 -> 2
deny-only, hook exits 2 -> 2    propagate, hook exits 1 -> 1
deny-only, hook exits 3 -> 0
```

Your audit question — which hooks have been inert — has an answer. Three declarations call `sys.exit(2)` under a `|| true` guard: `pre_tool_use` (10 sites, including `_enforce_tool_budget`), and the two `validate_sdlc_on_stop` twins. `pre_tool_use` becomes `deny-only`. The Stop pair stays `suppress`, because a Stop-event exit 2 blocks turn-end and the manifest already records the standing decision that a missed quality gate warns rather than wedges the turn fleet-wide. That is now a stated decision in the manifest rather than an invisible side effect.

`.claude/settings.json` was regenerated with `sync_project_hooks`, never hand-edited. Exactly one command string changed.

Your point that no test exercised the deployed command string is addressed directly: `tests/unit/test_update_hardlinks.py` builds the command from a `HookDeclaration`, runs it through `sh` with a real script exiting each code, and asserts the shell's exit status.

### BLOCKER 3 — the harness settled the design question

Rather than reason from the tool docs, we probed a live `claude -p` run with a hook that dumps `tool_input`. A `Task` call issued without the flag arrives with `run_in_background` **absent**:

```
{"keys": ["description", "prompt", "subagent_type"], "rib": "<ABSENT>"}
```

The harness injects no default, and the tool's contract is that an absent flag means background. So treating omission as foreground would admit the exact dispatch shape #2420 exists to prevent. The predicate stays as written; the dispatch surface is what needed fixing.

Our sweep found three flagless templates: the two you named plus **`do-patch/SKILL.md`**, which your list did not include. All three now carry `run_in_background: false`, and the rule is stated on the always-loaded eng surface (`config/personas/engineer.md`, `.claude/agents/dev.md`, `prime-pm-role.md`, `prime-dev-role.md`) and in `docs/sdlc/do-plan-critique.md`, whose "(foreground, no `run_in_background`)" phrasing described the shape that gets denied.

The durable fix you asked for: `test_sdlc_fork_no_background.py` now discovers every `Task({` / `Agent({` template across `.claude/skills-global/`, `.claude/skills/`, `.claude/agents/`, `.claude/commands/`, `config/personas/`, and `docs/sdlc/`, and asserts **per template** so one correct template cannot vouch for a flagless sibling in the same file. A companion test asserts discovery reaches known sub-files, so a discovery bug cannot make it pass vacuously — the failure mode of the glob it replaces. Replayed against the pre-fix tree it fails at exactly the three offending sites and passes the three that were already correct.

### Nits

Restored the hang bound: `GROUP_TIMEOUT` (10 minutes) is carried verbatim into each do-test dispatch prompt with instructions to kill and report `TIMEOUT`. Foreground dispatch means the parent blocks, so the bound has to live inside the child's own command — the only place that can end a hang. Step 3 treats a `TIMEOUT` report as a group that falls back to direct execution.

`sdlc-fork-turn-boundary.md` now cites lines 26 and 135.

### Tests

`tests/unit/session_runner/` + `tests/unit/hooks/` = **385 passed** against your 376 baseline (322 + 63). The +9 is entirely new coverage in `test_subagent_in_flight.py`: the just-spawned class, the mtime window, and the every-agent-checked cases. No test was removed. The hook-registration and dispatch-surface suites add **220 passed**. `ruff check` and `ruff format` clean.

tomcounsell:
Heads up from the #2567 lane: PR #2579 landed on `main` as `39ae8fc3` and puts this branch into conflict. Flagging with specifics rather than leaving it to be discovered on your next push.

Measured against your branch rather than guessed:

| Shared file | #2455 | #2579 | Result |
|---|---|---|---|
| `scripts/update/hardlinks.py` | 16+/2- | 349+/4- | auto-merges |
| `docs/features/hook-manifest.md` | 30+/5- | 51+/2- | auto-merges |
| `tests/unit/test_update_hardlinks.py` | 96+/6- | 609+/3- | **conflict** |

One file, and it is a test file, so this is a rebase rather than a redesign.

What changed in `hardlinks.py` that may matter to your tests: `sync_user_hooks` now migrates the legacy `~/.claude/hooks` directory symlink and registers only declarations whose script is on disk, and `_cleanup_renamed` skips the whole `"hooks"` kind while the root aliases a checkout. The `fake_project` fixture also gained a `.git` marker, since the new `user_hooks_root_is_repo_aliased()` helper only treats a resolved path as tracked source when it is a real checkout's `.claude/hooks`. If your new tests build their own project fixture, they will need the same marker.

Sorry for the churn. Ping me if the rebase turns up anything that looks like a real behavior clash rather than a textual one.

tomcounsell:
## Review: APPROVED

Reviewed at head `32cb2ba4` (post-rebase onto main after PR #2579 landed). Recording what I verified myself and what I took on the author's evidence, so the scope of this approval is legible.

### Verified independently

**The exit_policy migration is behavior-preserving except where intended.** I diffed every `[[hook]]` declaration on the branch against `origin/main` and mapped `blocking` to `exit_policy`. Twenty declarations, and exactly one changes behavior:

| | main `blocking` | branch `exit_policy` |
|---|---|---|
| `pre_tool_use.py` | `false` (suffixed `\|\| true`) | `deny-only` |
| every other declaration | `true` / `false` | `propagate` / `suppress` |

No declaration was removed, and nothing is silently promoted from suppressed to propagating. The one intended change is the point of the PR.

**The blast-radius sweep holds.** I swept independently, with a different method than the author's: every `Task({` / `Agent({` template across `.claude/skills-global/`, `.claude/skills/`, `.claude/skill-context/`, `.claude/agents/`, `.claude/commands/`, `config/personas/`, `docs/sdlc/`, `docs/features/`, `scripts/`, and `tools/`, plus the deployed global tree at `~/.claude/skills/`, plus dynamically constructed dispatch in Python. Seven templates exist; all seven carry an explicit `run_in_background: false` on the branch. The deployed global tree holds the same five files as inode-identical hardlinks, so it carries no independent surface. The only Python reference to a `Task` tool name is a read in `agent/health_check.py:421`, not a dispatch.

**Only two live deny sites, both gated fail-open.** `_enforce_tool_budget` and the foreground guard. Both fire only on a positively-resolved session, and both fail open on unresolved sessions, resolution errors, and internal bugs.

**Tests and lint, run rather than credited.** `tests/unit/session_runner/` + `tests/unit/hooks/` = **390 passed** (the body says 385; the rebase picked up five more from main). `test_update_hardlinks.py` + `test_sdlc_fork_no_background.py` = **86 passed**. `ruff check` and `ruff format --check` clean.

### Taken on the author's evidence

The blocker-1 corpus measurements (1573 sidechains, 100% just-spawned detection, 0 false positives on 1455 finished transcripts). The methodology is sound and the reported shape is consistent with what I read in the predicate, but I did not re-run the replay.

### Deployment note, not a blocker

`_enforce_tool_budget` has been inert in production for as long as the `|| true` suffix has been there, so its cap goes live for the first time on the next `/update`: 1000 tool calls and $50 per session, `TOOL_BUDGET_ENABLED` defaulting on. The caps are generous and fail safe on missing data, and `TOOL_BUDGET_ENABLED=false` is the instant kill switch. Worth watching the first long session after deploy.

Relatedly, the deny goes live in the same `/update` that re-hardlinks the fixed skill files, so the two land together. If a future `git pull` breaks those hardlinks (issue #2500) without a following `/update`, the deny would be live against stale flagless templates. Verify the hardlinks after deploying.

Issue #2521, opened 2026-08-04
Prune orphaned pre-manifest hook files under ~/.claude/hooks/


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/prune-orphaned-hook-files.md

Split out of #2503 during plan critique (BLOCKER 2).

## Context

`~/.claude/hooks/` still holds whole orphaned **pre-manifest** trees that `sync_user_hooks` neither created nor removes: `validators/` (20 files), `dispatch/`, `hook_utils/`, plus 8 top-level scripts and a stray `manifest.toml`. These predate the manifest-driven generator (#2435 / PR #2453), which is keyed on `# hook:<id>` markers and structurally cannot see them.

#2503 removes the one orphaned **registration entry** (the unmarked `validate_no_raw_redis_delete.py` command in `~/.claude/settings.json`). After that lands the orphaned files are inert — stale bytes on disk, not executing code. This issue covers deleting them.

## Why this was split out of #2503

The original plan gated deletion on a negative reference test: delete any file under `~/.claude/hooks/` that is neither a declared global script (nor an `sdlc/` sibling import) nor named by a surviving command string. Critique found that unsound and out of scope:

- **Unsound.** The gate only sees *command-string* references. A file reached solely via a Python `import` is invisible to it and gets deleted. "Nothing that anything still points at can be removed" is true only for the smaller command-string set.
- **Different risk class.** Deleting files from a users home directory, fleet-wide, is not the same operation as fixing an interpreter token. #2503 Risk 2 rated the failure mode "worse than the bug being fixed."

## Design constraint

The prune must use a **static allowlist** of exactly the trees to keep, so an unrecognized file is **kept by default** rather than deleted by default. Do **not** implement a transitive AST import walk to rescue the negative-reference approach — that is a third risk class on top of the second.

## Acceptance Criteria

- [ ] Orphaned pre-manifest files under `~/.claude/hooks/` are removed by an idempotent migration registered in `MIGRATIONS`
- [ ] The keep-set is a static allowlist; an unrecognized file under `~/.claude/hooks/` is NOT deleted, asserted by test
- [ ] A timestamped `.bak` (or equivalent recoverable step) is taken before any destructive operation
- [ ] `tests/unit/test_hook_migration.py:106`'s foreign `python /opt/my-own-tool/guard.py` fixture still survives
- [ ] Running `/update` twice is a no-op on the second run

Blocked by #2503 (its registration sweep must land first).

[...]

PR #2522, opened 2026-08-04
Resolve hook interpreter deterministically under stripped /bin/sh (#2503)

### LINKED ISSUE #2503: SDLC hooks are dead in every foreign repo: bare `python` never resolves under /bin/sh, and PEP 604 annotations crash the global-scope scripts at import
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/sdlc-hook-interpreter-resolution.md

> **Context:** [Claude Code hooks](https://docs.claude.com/en/docs/claude-code/hooks) are shell commands the CLI runs automatically around tool calls (before a `Bash` call, after a `Write`, when a session stops). This repo declares all of its hooks in [`.claude/hooks/manifest.toml`](.claude/hooks/manifest.toml) and generates two registration files from it: the repo-local `.claude/settings.json` (**project scope**, only active inside `~/src/ai`) and the machine-wide `~/.claude/settings.json` (**global scope**, active in *every* repo on the machine). The generator is [`scripts/update/hardlinks.py`](scripts/update/hardlinks.py), run by `/update`. This issue is about the interpreter those generated commands invoke.

## Problem

Every generated hook command starts with a bare `python`. Claude Code executes hook commands through `/bin/sh`, which is non-interactive and never sources `~/.zshenv`, so a user's `python` → `python3` shell alias does not apply. Verified on this machine:

```
$ env -i /bin/sh -c 'command -v python'
(exit 1 — no such binary)
```

Modern macOS ships no `/usr/bin/python` at all. The hooks only work inside `~/src/ai` because that repo's `.venv/bin` happens to be on `PATH` there. Everywhere else the **global-scope** hooks are dead, and every `Bash`/`Edit` tool call in a foreign repo emits:

```
PreToolUse:Bash hook error
Failed with non-blocking status code: /bin/sh: python: command not found
```

Naively switching the literal to `python3` swaps one failure for another, because a second latent bug is hiding behind the first. Two of the globally-deployed scripts use [PEP 604](https://peps.python.org/pep-0604/) union syntax (`str | None`) in a **return annotation**, which Python evaluates at *module import time*:

- [`.claude/hooks/sdlc/validate_commit_message_sdlc.py:32`](.claude/hooks/sdlc/validate_commit_message_sdlc.py#L32) — `def get_current_branch() -> str | None:`
- [`.claude/hooks/validators/validate_no_raw_redis_delete.py:107`](.claude/hooks/validators/validate_no_raw_redis_delete.py#L107) — `def find_violation(command: str) -> str | None:`

Neither file has `from __future__ import annotations`. Under Python 3.9 (`/usr/bin/python3` on macOS) both raise `TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'` **before `main()` ever runs**, so the `try/except Exception` fail-open guard inside `main()` cannot catch it. `validate_commit_message_sdlc.py` is registered `blocking = true` with no `|| true`, so its import crash surfaces as a hook error on every single `Bash` tool call.

**Current behavior:** Global-scope SDLC hooks silently do not run in any repo other than `~/src/ai`, and emit an error line on every tool call. The commit-message guard and the raw-Redis guard are unenforced everywhere except this one repo, while appearing to be installed.

**Desired outcome:** Generated hook commands invoke a **deterministic** interpreter that resolves identically on every machine under `/bin/sh` with no `PATH` inheritance, and the global-scope scripts import cleanly under the oldest interpreter they can land on. A regression test fails if either property breaks.

## Definitions

| Hook manifest | The single TOML file declaring every hook (event, matcher, script, scope, blocking). Both `settings.json` `hooks` blocks are generated from it; neither is hand-edited. | [`.claude/hooks/manifest.toml`](.claude/hooks/manifest.toml), [`docs/features/hook-manifest.md`](docs/features/hook-manifest.md) |
| Project scope vs global scope | A manifest `scope` field. `project` hooks are written to the repo's `.claude/settings.json` and run only in this repo; `global` hooks are written to `~/.claude/settings.json` and run in every repo on the machine. | [`scripts/update/hook_manifest.py`](scripts/update/hook_manifest.py) |
| `_build_hook_command` | The single function that renders one `HookDeclaration` into a shell command string. It hardcodes the interpreter token. | [`scripts/update/hardlinks.py`](scripts/update/hardlinks.py) (~line 707) |
| SDLC fork | The dependency-free copies of hook scripts under `.claude/hooks/sdlc/`, deliberately kept free of repo and third-party imports so they can run standalone in a foreign repo. | [`.claude/hooks/sdlc/`](.claude/hooks/sdlc/) |

## Prior Context

- **#2435** / **PR #2453** (merged): replaced the old hardcoded `_SDLC_HOOK_DEFS` list with manifest-driven generation for both scopes. This is the architecture any fix must target.
- **PR #2454** (open, branch `hotfix/sdlc-hook-python3`): an earlier attempt at exactly this fix, opened 2026-07-29, now `CONFLICTING`/`DIRTY`. Its diff patches `_SDLC_HOOK_DEFS` and the pre-manifest `_merge_hook_settings` shape that #2453 deleted, and edits `.claude/hooks/sdlc/validate_commit_message.py`, which has since been renamed to `validate_commit_message_sdlc.py`. **Recommend closing it in favor of this issue.** Its problem statement is sound; its diff is not salvageable.

## Solution Sketch

The root-cause fix shape is genuinely uncertain, and one obvious candidate is explicitly forbidden. Open questions for `/do-plan` to settle rather than a prescribed approach:

**Q1. What interpreter token should `_build_hook_command` emit?** `python3` is *not* automatically correct: it is machine-dependent. On this machine a stripped-`PATH` `python3` resolves to `/usr/local/bin/python3` (3.12.6), which shadows `/usr/bin/python3` (3.9.6). The worst failure mode is passing on the developer's box and breaking fleet-wide. An absolute interpreter path is deterministic but must differ per scope (see Q2). Should `HookDeclaration` gain an explicit `interpreter` field rather than the generator hardcoding one?

**Q2. Should project scope and global scope use different interpreters?** They have genuinely different requirements. Project-scope hooks import repo modules and use `datetime.UTC` (3.11+), so they need the repo venv (`/Users/valorengels/src/ai/.venv/bin/python`, currently 3.14.3). Global-scope hooks run in foreign repos where that venv is not on `PATH`, and are deliberately dependency-free.

**Q3. How is the interpreter path made portable across machines?** An absolute venv path is deterministic but machine-specific, and `~/.claude/settings.json` is generated per machine by `/update`, so this may be fine. Confirm.

**Constraint — `uv run` is off the table.** [`scripts/update/hooks.py:99`](scripts/update/hooks.py#L99) audits for and *fails* on `uv run` inside a hook ("can corrupt venv or trigger slow dependency resolution"). #2435's plan tracked this as ["Pre-requisite Bug 4"](docs/plans/completed/hook-registration-manifest-dispatcher.md) and deliberately **removed** the repo's only `uv run` hook registration, normalizing everything to bare `python`. The bare-`python` literal is a documented decision, not an oversight. Any plan that reintroduces `uv run` contradicts a shipped decision and fails the repo's own audit.

**Constraint — the current interpreter is baked into the test suite.** Several tests assert the exact generated string with a bare `python` prefix: `tests/unit/test_update_hardlinks.py` (lines 251, 738-739) and `tests/unit/test_hook_migration.py` (lines 58, 70, 82, 268). Changing the literal requires updating these in the same change, and the migration path must rewrite already-deployed legacy commands **in place** rather than appending a second block (dedup is by exact command string, so a stale entry would otherwise keep firing alongside the new one).

## Recon Summary

**Confirmed:** 3 items

- **Bare `python` does not resolve under a stripped `/bin/sh`.** Verified directly: `env -i /bin/sh -c 'command -v python'` exits 1. Single emission site: `_build_hook_command` in `scripts/update/hardlinks.py`. `HookDeclaration` has no interpreter field.
- **PEP 604 import-time crash is real and worse than described.** The failure is at module import, so the `try/except` fail-open guard in `main()` never runs. The two affected scripts' documented "never block the user" contract is silently void. Adding `from __future__ import annotations` was verified to fix both under 3.9.6.
- **PR #2454 is unsalvageable.** Targets deleted internals (`_SDLC_HOOK_DEFS`, old `_merge_hook_settings`) and a renamed file.

**Revised:** 3 items — scope adjusted

- **`python3` is not the fix.** Originally framed as a one-literal swap. Recon found resolution is machine-dependent (3.12.6 here, 3.9.6 on a clean mac). Reframed as Q1/Q3 above.
- **The fix is likely two-tier, not one literal.** Under 3.9, 27 of 35 hook scripts fail. Beyond PEP 604, five files import `datetime.UTC` (3.11+), which `from __future__ import annotations` **cannot** fix (it is a runtime symbol, not an annotation): `hook_utils/constants.py`, `hook_utils/detach_lock.py`, `post_tool_use.py`, `pre_tool_use.py`, `validators/validate_design_system_sync.py`. Because `hook_utils/constants.py` is a shared import, this poisons ~11 more files transitively. These are all project-scope and fine under the venv, which is exactly why scope may need to drive interpreter choice.
- **`__future__` alone is sufficient *for global scope only*.** All globally-deployed scripts are free of third-party and unguarded repo imports, so the standalone-fork discipline held.

**Pre-requisites:** 2 items — call out as blockers

- **`uv run` is forbidden by the repo's own audit** and was deliberately removed by #2435. Any plan proposing it is dead on arrival. Documented above as a hard constraint.
- **Manifest/reality drift:** `validators/validate_no_raw_redis_delete.py` is declared `scope = "project"` in the manifest but is actually registered in `~/.claude/settings.json` (global). It is one of the two PEP 604 crashers. Either the manifest or the deployed state is wrong; resolve before deciding which scripts must be 3.9-clean. Split to its own issue if it turns out to be a generator bug rather than stale deployed state.

**Dropped:** 2 items

- **Reusing PR #2454's diff or its `tests/unit/test_hardlinks_hook_merge.py`** — both test deleted internals.
- **"Consider `uv run` as the fix shape"** — explicitly forbidden, see above.

## Acceptance Criteria

- [ ] Every command generated by `_build_hook_command` invokes an interpreter that resolves successfully under `env -i /bin/sh` with no inherited `PATH`, on a machine with no `python` binary.
- [ ] Every script in `.claude/hooks/sdlc/` and every script registered in `~/.claude/settings.json` imports cleanly under the oldest interpreter it can be dispatched to, verified by actually executing it with a stripped environment.
- [ ] A regression test asserts no generated hook command begins with a bare `python` token.
- [ ] A regression test asserts every global-scope hook script parses and imports under the minimum supported interpreter (catching both PEP 604 and `datetime.UTC` classes of failure).
- [ ] Existing tests that hardcode the old interpreter string are updated: `tests/unit/test_update_hardlinks.py`, `tests/unit/test_hook_migration.py`.
- [ ] The migration rewrites already-deployed legacy hook commands **in place** in both `settings.json` files; running `/update` twice produces no duplicate hook entries.
- [ ] `scripts/update/hooks.py`'s `uv run` audit still passes against the regenerated settings.
- [ ] The `validate_no_raw_redis_delete.py` scope discrepancy is resolved or split into a tracked follow-up.
- [ ] Manually verified: a `Bash` tool call in a repo other than `~/src/ai` produces no hook error, and the commit-message guard actually blocks a code commit to `main` there.
- [ ] PR #2454 is closed with a pointer to this issue.

## Downstream

Constraints for the planner:

- **Do not use `uv run`** in any hook registration or shebang. It fails `scripts/update/hooks.py`'s audit and reverses a shipped decision from #2435.
- **Do not reintroduce `_SDLC_HOOK_DEFS`.** The manifest is the single source of truth; the generator must stay manifest-driven.
- **No new dependencies** in `.claude/hooks/sdlc/` scripts. Their whole value is running standalone in a foreign repo with no venv.
- **Verify empirically, not by inspection.** Both bugs here were latent precisely because the failure only reproduces under a stripped environment. Any fix must be tested with `env -i` against a real interpreter, not just reasoned about.

valorengels:
## Summary

Hook commands run through a non-interactive `/bin/sh` with no `PATH`, so every generated command's bare `python` token exits 127 — silently disabling the commit-message and raw-Redis guards in every repo except `~/src/ai`. This makes the interpreter a per-scope deployment contract instead of an implicit `PATH` dependency.

## Changes

- **`.claude/hooks/hook_python`** (new, mode 755, extensionless POSIX-sh shim): the project-scope interpreter. Resolves the **main checkout's** venv first via `git rev-parse --path-format=absolute --git-common-dir`, falling back to `$root/.venv/bin/python` only when git cannot answer. Main-checkout-first is deliberate: 9 of 57 worktrees carry an unmanaged, version-divergent `.venv`, so a local-first order would silently exec a stale interpreter. Fails open (stderr message, empty stdout, exit 0) via `exec` so `blocking=true` semantics pass through.
- **`scripts/update/hardlinks.py`**: `_build_hook_command` gains a required `interpreter` argument (no `HookDeclaration` field). Project scope emits the shim token; global scope emits a resolved absolute system `python3`, probed by **executing** `-V` (not `os.path.exists`, because the macOS CLT stub exists but exits non-zero). `MIN_GLOBAL_PYTHON = (3, 9)` declared once. Total probe failure records an error on `HardlinkSyncResult` rather than emitting a dead command. `.claude/settings.json` regenerated.
- **`.claude/hooks/sdlc/validate_commit_message_sdlc.py`**: `from __future__ import annotations` fixes the import-time PEP 604 crash under 3.9 (the only affected global script).
- **`scripts/update/migrations.py`**: new `sweep_legacy_unmarked_global_hooks` migration (registered in `MIGRATIONS`) removes the unmarked pre-manifest `validate_no_raw_redis_delete.py` registration from `~/.claude/settings.json`. Gated on a precise-match `_LEGACY_UNMARKED_SWEEP_TARGETS` tuple + unmarked + `~/.claude/hooks/` reference. `.bak` + JSON-validity guard. Deletes no files (orphaned-file prune deferred to #2521).
- **`reflections/audits/hooks_audit.py`**: dedicated finding on the shim's fail-open stderr message, so a machine failing open is distinguishable from generic hook-error noise.
- **`docs/features/hook-manifest.md`**: new Interpreter Contract section.

## Testing
- [x] `tests/unit/test_hook_interpreter.py` (new): bare-`python` bans, always-on AST pre-3.10 syntax walk, real `env -i` execution under the min interpreter, double-`/update` idempotence, and all three shim worktree-precedence cases + fail-open assertions
- [x] Targeted suite: 90 passed (test_hook_interpreter, test_update_hardlinks, test_hook_migration, test_hooks_audit)
- [x] Lint/format clean on the diff
- [x] All shim/interpreter behavior verified by execution under `env -i`

## Documentation
- [x] Interpreter Contract section added to `docs/features/hook-manifest.md`

## Definition of Done
- [x] Tested: Targeted tests passing (full-suite failures are pre-existing worktree env artifacts, unrelated to the diff)
- [x] Documented: Interpreter contract documented
- [x] Quality: Lint and format checks pass on the diff

Closes #2503


valorengels:
## Review:

**Verdict: CHANGES REQUESTED** — 0 blockers, 0 tech debt, 3 nits.

This is a strong, well-evidenced fix. Every empirical claim in the PR body was **reproduced independently in this review**, not taken on trust. The only findings are documentation staleness in the plan and one dead branch in a test helper.

### What I verified by execution (not by reading)

| Check | Result |
|---|---|
| Targeted suite (`test_hook_interpreter`, `test_update_hardlinks`, `test_hook_migration`, `test_hooks_audit`) | **90 passed** — matches the PR's claim exactly |
| Adjacent hook/settings consumers (`test_relink_global_skills`, `test_pipeline_integrity`, `test_reflections_package`, `test_sync_claude_to_opencode`, `test_pre_tool_use_dispatcher`, `test_validate_sdlc_on_stop`) | **130 passed** — no collateral breakage |
| `ruff check .` / `ruff format --check .` | clean (1254 files) |
| Shim from a real worktree with **no** local `.venv` | resolves `/Users/valorengels/src/ai/.venv` ✅ |
| Shim from a real worktree **with** its own `.venv` (`.worktrees/hybrid-retrieval-eval`) | resolves the **main** checkout's venv ✅ — the load-bearing precedence decision holds against production state, not just a fixture |
| Shim fail-open in a non-git dir | stderr message, **stdout empty**, exit 0 ✅ |
| All 3 declared global hooks + `sdlc/sdlc_context.py` under `env -i /usr/bin/python3` (3.9.6) with `{}` on stdin | all **exit 0** ✅ |
| `resolve_global_interpreter()` on this machine | `/usr/bin/python3` (the 3.9 floor — correct worst case, not the convenient one) |
| Anti-criteria (no bare `python` in settings, no `/Users/` committed, no `uv run`, no `_SDLC_HOOK_DEFS`, legacy match literal preserved, migration registered) | all pass |

**Shim overhead measured**, since it lands on the hot path of every tool call: ~5 ms per invocation over a direct venv `python` (2 extra process spawns). Negligible against hook timeouts of 5–20 s. Not a finding, but now measured rather than assumed.

### Rubric

| # | Item | Result |
|---|---|---|
| 1 | Correctness | pass — all three shim branches and both interpreter tiers exercised under `env -i` |
| 2 | Security | pass — sweep is a precise-match allowlist, `.bak` + JSON-validity guard, deletes no files |
| 3 | Error handling | pass — fail-open at run time (documented decision), fail-**closed** at generation time (probe failure records an error rather than emitting a dead command) |
| 4 | Tests | pass — 502 new lines that genuinely execute; the AST walk is always-on and cannot be skipped away |
| 5 | Code quality | pass |
| 6 | Documentation | pass — new Interpreter Contract section, plus the `hook-manifest.md:157` sharpening the plan called for |
| 7 | Plan / Success Criteria | pass — every non-`[EXTERNAL]` criterion verified above |
| 8 | No-Gos respected | pass — no `uv run`, no file deletion, no fail-closed behavior change, no `_SDLC_HOOK_DEFS` reintroduction |
| 9 | Verification table | pass — all rows run |
| 10 | Multi-machine / migration | pass — new migration registered in `MIGRATIONS`, no new env vars, idempotence pinned by test |

### Design calls worth naming as correct

- **Main-checkout-first precedence** is the right inversion. I confirmed 9 worktrees currently carry an unmanaged `.venv`; a local-first shim would silently exec a stale interpreter in every one of them. The `test_shim_from_worktree_with_own_venv_still_resolves_main` test pins this against regression.
- **Probing by execution rather than `os.path.exists`** correctly handles the macOS CLT stub. `TimeoutExpired` subclasses `SubprocessError`, so the GUI-dialog hang case is caught by the existing `except`.
- **Extensionless shim name** is load-bearing: `reflections/audits/hooks_audit.py` takes the first `.py`/`.sh` token as the script path, so a `.sh` shim would have made the audit inspect the wrapper instead of the hook. Verified against the audit source.
- The `assert interpreter is not None` in `_merge_hook_settings` is genuinely unreachable (the loop is empty whenever `global_decls` is), so stripping under `-O` is not a live risk.

### Blockers
- None

### Tech Debt
- None

### Nits

**1. `docs/plans/sdlc-hook-interpreter-resolution.md:224` describes the retired local-first precedence.**
The Solution > Key Elements bullet reads: *"Tries `$CLAUDE_PROJECT_DIR/.venv/bin/python`, then the main checkout via `git rev-parse ...`"*. That is the **inverse** of what shipped, and precedence was the subject of a BLOCKER-level critique correction. Spike-1 and the docs both say main-checkout-first; this one line still says local-first. A future reader (or re-build) taking Key Elements at face value would reintroduce the exact bug the critique caught.
*Fix:* reword to main-checkout-first to match the shim, `docs/features/hook-manifest.md`, and spike-1's correction.

**2. `docs/plans/sdlc-hook-interpreter-resolution.md:487` still carries the retired file-prune scope label.**
`Domain: redis-data (idempotent, reference-gated, .bak-guarded destructive sweep)` — this is the CONCERN at row 622 of the plan's own Critique Results table, still marked `pending`. The migration is registration-only and deletes no files (verified: `test_sweep_deletes_no_files` passes). The plan currently asserts two contradictory scopes for the same task.
*Fix:* apply the critique row's own suggested text, e.g. `Domain: settings-json — idempotent, .bak-guarded, registration-only (deletes no files)`, and flip both rows 621/622 from `pending` to `addressed` — both are in fact addressed in code.

**3. Dead branch in `tests/unit/test_hook_interpreter.py:87` (`_starts_with_bare_python`).**
`tok.startswith("python ")` can never be true: `tok` comes from `_first_token()`, which returns `command.split()[0]`, so it cannot contain a space. Harmless, but it reads as if it guards a case it does not.
*Fix:* drop the clause, or replace it with the check it looks like it intends.

### Deploy note

No `bridge/`, `agent/`, or `worker/` code changes, so **no service restart is required**. But every effect reaches a machine only through `/update` — each fleet machine must run it locally to regenerate `~/.claude/settings.json` and pick up the sweep migration. The plan correctly scopes that as `[EXTERNAL]`.

Also worth noting for the merge: the sweep migration is a **no-op on this machine** — the live `~/.claude/settings.json` no longer carries the unmarked `validate_no_raw_redis_delete.py` entry that spike-3 found. That is expected (the migration is idempotent and other fleet machines may still carry it), not a defect, but it means the sweep's real-world effect will not be observable here.

---

*Single-judge review. Findings were verified against code read on the PR branch at `cb1759a1`; numbers in the PR body were reproduced locally rather than credited as claimed.*

valorengels:

**0 blockers, 1 tech debt, 1 nit.** Down from 3 nits on the prior pass at `cb1759a1`; two of those three were plan-doc fixes landed on `main` and the third was fixed on the branch. One new finding, verified by execution.

Every empirical claim in the PR body was **reproduced in this review**, not credited as stated.

### What I verified by execution (not by reading)

- Targeted suite (`test_hook_interpreter`, `test_update_hardlinks`, `test_hook_migration`, `test_hooks_audit`) — **90 passed**, matching the PR body exactly.
- Adjacent consumers (`test_relink_global_skills`, `test_pipeline_integrity`, `test_pre_tool_use_dispatcher`, `test_validate_sdlc_on_stop`) — **69 passed**, no collateral breakage.
- `ruff check .` and `ruff format --check .` — clean (1254 files).
- Shim from a worktree with **no** local `.venv` → resolves `/Users/valorengels/src/ai/.venv/bin/python`.
- Shim from a worktree that **has** its own `.venv` (`.worktrees/hybrid-retrieval-eval`, live production state) → resolves `sys.prefix = /Users/valorengels/src/ai/.venv`. The load-bearing precedence inversion holds against reality, not just a fixture.
- Shim fail-open in a non-git dir → stderr message, **stdout empty**, exit 0.
- All four globally-deployed scripts (`sdlc_context.py`, `sdlc_reminder.py`, `validate_commit_message_sdlc.py`, `validate_sdlc_on_stop.py`) under `env -i /usr/bin/python3` (3.9.6) with `{}` on stdin → all exit 0.
- `resolve_global_interpreter()` on this machine → `/usr/bin/python3` — the 3.9 floor, i.e. the worst case, not the convenient one.

### Verification table (all rows run on the PR branch)

Every row in the plan's `## Verification` table passes: no bare `python` in `.claude/settings.json` (0), no `/Users/` committed (0), shim executable, `from __future__ import annotations` present (1), no `uv run` (0), no `_SDLC_HOOK_DEFS` (0), legacy match literal preserved (1), new migration registered (2 hook migrations), no `os.remove`/`unlink`/`rmtree` anywhere in `migrations.py`.

One cosmetic mismatch, not a finding: the "Shim fail-open is audited" row expects `grep -c` to output `1` but the literal now appears twice in `hooks_audit.py` (once in the explanatory comment, once in the code). The row's intent — the literal is present — holds.

## Rubric

- [x] **1. Plan vs. implementation match** — pass — *every non-`[EXTERNAL]` success criterion verified above by execution; the two `[EXTERNAL]` ones (fleet `/update`, live foreign-repo session) are correctly out of the agent's reach.*
- [x] **2. New code quality** — pass — *comments are unusually load-bearing and accurate throughout; the one exception is the Miscellaneous finding below, which is a localized latent defect rather than a systemic quality problem.*
- [x] **3. Test coverage** — pass — *the tests genuinely execute: a real `/bin/sh`-invoked shim against `git worktree add` fixtures, a real `env -i` interpreter run, and an always-on AST walk that cannot be skipped away. Each would fail on the pre-fix code.*
- [x] **4. Regression risk to existing callers** — pass — *`_build_hook_command` gained a required keyword-only `interpreter`; both call sites in `hardlinks.py` and the one in `migrations.py` were updated, and the adjacent 69-test consumer run is green.*
- [x] **5. Data integrity** — pass — *no Popoto/schema change. The `~/.claude/settings.json` edit is `.bak`-snapshotted and JSON-validity-asserted before touching disk.*
- [x] **6. Security** — pass — *the sweep is a precise-match allowlist (`_LEGACY_UNMARKED_SWEEP_TARGETS`), not a directory-prefix gate, so a hand-added user hook under `~/.claude/hooks/` survives. No secrets, no shell injection surface (the shim `exec`s a fixed path).*
- [ ] **7. Documentation accuracy** — fail — *`docs/features/README.md:89` still describes a single hook migration and omits the interpreter contract. See Nits.*
- [x] **8. PR body accuracy** — pass — *90 passed, 12 changed files, migration registered, `.bak` guard, deletes-no-files: all reproduced independently.*
- [x] **9. Disclosed deferrals** — pass — *the orphaned-file prune is deferred with sound rationale and tracked.*
- [x] **10. Follow-up claims verified** — pass — *#2521 confirmed OPEN.*

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — PASS — *all non-`[EXTERNAL]` criteria executed above.*
- **2. No-Gos from plan — none violated** — PASS — *no `uv run`, no file deletion, no fail-closed behavior change, no `_SDLC_HOOK_DEFS` reintroduction, file prune correctly deferred to #2521.*
- **3. New `except Exception` blocks — each has logger/raise/swallow-ok** — PASS — *`resolve_global_interpreter` catches `(OSError, SubprocessError)` and continues to the next candidate, which is the intended semantic; `TimeoutExpired` subclasses `SubprocessError`, so the CLT GUI-dialog hang is covered.*
- **4. New integration tests — exercise serialization boundary (not in-memory only)** — PASS — *the shim tests write real files, `git init` + `git worktree add`, and invoke `/bin/sh` under `env={}`.*
- **5. Plan internal consistency — spike findings match task steps** — PASS — *the two plan-doc inconsistencies flagged on the prior pass (line 224 precedence, line 487 scope label) are both corrected on `main`.*
- **7. New public APIs — docstrings present** — PASS — *`resolve_global_interpreter` and the new migration both carry docstrings that state the *why*, not just the what.*
- **8. Breaking changes — migration path documented** — PASS — *`## Update System` covers it; every effect reaches a machine only via `/update`.*
- **10. Tests cover the failure path (not just happy path)** — PASS — *fail-open branch, no-local-venv branch, has-local-venv branch, foreign-hook survival, unrecognized-file survival, double-`/update` idempotence.*
- **11. UI changes (if any) — screenshot captured** — N/A — *no HTML/CSS/JS/template files in the diff.*
- **12. Docs updated for user-facing changes** — FAIL — *`docs/features/hook-manifest.md` is thorough, but the `README.md` index line was not refreshed. Promoted to the Nit below.*

### Blockers
- None

### Tech Debt
- None

### Nits

**File:** `docs/features/README.md:89`
**Code:** `... a both-scope hooks_audit, and a migration rewriting the 3 deployed legacy user-scope entries in place`
**Issue:** The index line enumerates the feature's components but predates this PR: there are now **two** hook migrations (the manifest-id rewrite plus `sweep_legacy_unmarked_global_hooks`), and the interpreter contract — arguably the most operationally significant part of the feature — is absent. The plan's own Documentation section asks to "confirm the one-line description still reads true and update it if the interpreter contract changes its scope." Nothing in the line is false, but it is now an incomplete inventory of a feature whose index entry exists to be an inventory.
**Severity:** nit
**Fix:** Append the interpreter contract (project → `hook_python` shim, global → resolved absolute system `python3`, `MIN_GLOBAL_PYTHON` floor) and pluralize the migration clause.

### Miscellaneous

**File:** `scripts/update/hardlinks.py:1004-1009`
**Code:**
```python
# Add/update pass, in manifest declaration order. ``interpreter`` is
# non-None here: sync_user_hooks returns early on a total probe failure
# whenever global_decls is non-empty, so reaching this loop guarantees a
# resolved interpreter.
assert interpreter is not None
for decl in global_decls:
```
**Issue:** The comment's invariant does not hold, because the `assert` sits **before** the loop rather than inside it. `sync_user_hooks` returns early only when `interpreter is None AND global_decls` is truthy; when `global_decls` is empty and the probe fails, it still calls `_merge_hook_settings(..., interpreter=None, ...)` — precisely because the removal pass must run even with an empty global scope, as that function's own comment says. The `assert` then fires before the empty loop is ever reached, raising `AssertionError` out of `/update`'s hardlink step.

Verified by execution, not inference:

```
>>> _merge_hook_settings(p, d/"hooks", [], None, HardlinkSyncResult())
ASSERTIONERROR RAISED: AssertionError()
```

The trigger is doubly improbable in practice (every global hook removed from the manifest **and** no working `python3` among all three candidates), so this is not a blocker. But the durable defect is the comment: it asserts an invariant the code does not hold, on exactly the branch a future editor would consult before changing the early-return condition — and this PR's whole thesis is that reasoning by inspection instead of execution is what let the original bug survive two fixes.
**Severity:** tech_debt
**Fix:** Move `assert interpreter is not None` inside the `for` loop body (where the comment's reasoning is actually true), or make the guard real: `if interpreter is None: return` after the removal pass. Either way, correct the comment to say the *loop body* is unreachable with a `None` interpreter, not the statement.

- **`[DEFERRED]` Pruning the orphaned pre-manifest files under `~/.claude/hooks/`** — tracked by **#2521 (OPEN)**, "Prune orphaned pre-manifest hook files under ~/.claude/hooks/". Confirmed registration-only: `grep -E 'os\.remove|\.unlink\(|rmtree' scripts/update/migrations.py` returns nothing, and `test_sweep_deletes_no_files` pins it. Sound deferral — fleet-wide home-directory deletion is a different risk class from an interpreter-token fix, and the plan correctly mandates a static keep-by-default allowlist for the follow-up.
- **`[EXTERNAL]` Running `/update` on the other fleet machines** — explicit out-of-scope exclusion, no tracking required; the agent cannot reach those machines.
- **`[EXTERNAL]` The issue's live foreign-repo manual verification** — explicit out-of-scope exclusion; needs an interactive session a human must drive.
- **`[EXTERNAL]` Whether hooks should become fail-closed** — explicit out-of-scope exclusion with sound rationale (a wrong answer wedges every tool call fleet-wide; it is an owner policy call, not a cleanup).
- **`[EXTERNAL]` File locking around `~/.claude/settings.json` writes** — explicit out-of-scope exclusion; pre-existing and operator-bounded.

### Review Delta (vs prior review on HEAD `cb1759a1`)

- **Resolved — Nit 1** (`docs/plans/…:224` described the retired local-first precedence): the plan now reads "Precedence is **main-checkout-first**" with the 9-of-57-worktrees rationale inline. Fixed on `main`, per this repo's plans-on-main convention.
- **Resolved — Nit 2** (`docs/plans/…:487` carried the retired `redis-data … destructive sweep` scope label): now reads `Domain: settings-json — idempotent, .bak-guarded, registration-only (deletes no files)`, exactly the critique row's suggested text.
- **Resolved — Nit 3** (dead `tok.startswith("python ")` clause in `_starts_with_bare_python`): fixed on the branch in `ba2e5ac`. The replacement — `"/" not in tok and "$" not in tok and tok.startswith("python")` — is strictly better than dropping the clause: it now bans `python3.11` and any other bare variant, not just the two literals the old version enumerated.
- **New**: the `assert interpreter is not None` reachability defect (Miscellaneous, above). The prior review asserted this assert was "genuinely unreachable (the loop is empty whenever `global_decls` is)" — that was **wrong**, and I confirmed it wrong by executing the call rather than re-reading the same line. Recording the self-correction explicitly, since an un-flagged reversal between passes is the calibration drift this delta section exists to surface.
- **Unchanged**: none carried forward.

### Design calls worth naming as correct

- **Main-checkout-first precedence.** I confirmed against live state that a worktree carrying its own `.venv` still resolves the main interpreter. A local-first order would silently exec a stale interpreter in every such worktree; `test_shim_from_worktree_with_own_venv_still_resolves_main` pins the inversion against regression.
- **Probing by execution, not `os.path.exists`.** Correctly handles the macOS Command Line Tools stub — the exact class of "it exists so it must work" reasoning this issue exists to end.
- **Extensionless shim name.** Load-bearing: `hooks_audit.py` takes the first `.py`/`.sh` token as the script path, so a `.sh` shim would make the audit inspect the wrapper instead of the hook.
- **Precise-match sweep allowlist over a directory-prefix gate.** The CONCERN-level critique correction is fully implemented and fixture-covered.
- **Keeping the bare `python` in `_legacy_fork_command_prefix`** with a comment and a named test exemption. Matching bytes already on disk is genuinely different from generating output, and this is the one place the old string must survive.

### Deploy note

No `bridge/`, `agent/`, or `worker/` changes — **no service restart required**. But every effect of this PR reaches a machine only through `/update`; each fleet machine must run it locally to regenerate `~/.claude/settings.json` and apply the sweep. The plan correctly scopes that as `[EXTERNAL]`.

Also relevant to the merge: the sweep migration is a **no-op on this machine** — the live `~/.claude/settings.json` no longer carries the unmarked `validate_no_raw_redis_delete.py` entry that spike-3 found. Expected (the migration is idempotent and other fleet machines may still carry it), not a defect — but it means the sweep's real-world effect is not observable here.

---

*Single-judge review. Multi-judge consensus was dispatched (`code-quality`, `risk`); neither judge returned within a ~12-minute in-turn await across four nudges, so this verdict is recorded as single-judge rather than stalling the pipeline. The consensus rule is any-blocker-wins and this verdict is already `CHANGES REQUESTED`, so a missing judge could not have flipped a clean approval — the failure direction is conservative. The judge non-return itself is worth a separate issue against the fork-await path.*

*All findings were verified against code read and executed on the PR branch at `ba2e5ac`. Numbers in the PR body were reproduced locally rather than credited as claimed.*

valorengels:

**0 blockers, 1 tech debt, 0 nits.** Both findings from the prior pass at `ba2e5ac` are resolved on the branch. One new finding, verified by execution — and it is the same class of defect the prior passes kept surfacing: a claim about behavior that reads true and is not.

Every empirical claim in the PR body was reproduced in this review, not credited as stated.

### What I verified by execution (not by reading)

- Targeted suite (`test_hook_interpreter`, `test_update_hardlinks`, `test_hook_migration`, `test_hooks_audit`) — **91 passed**. The PR body says 90; the delta commit adds `test_merge_hook_settings_runs_removal_pass_with_no_interpreter`, so 91 is the correct current number and the body's figure is merely stale.
- `ruff check .` and `ruff format --check .` — clean, 1254 files.
- Shim from a live worktree with **no** local `.venv` (`.worktrees/_merge_baseline_main`) → `sys.prefix = /Users/valorengels/src/ai/.venv`.
- Shim from a live worktree that **has** its own `.venv` (`.worktrees/fix-test-isolation-cluster`, real production state) → `sys.prefix = /Users/valorengels/src/ai/.venv`. The load-bearing precedence inversion holds against reality, not just a fixture.
- Shim fail-open in a non-git dir → stderr line, **stdout empty**, exit 0.
- `resolve_global_interpreter()` on this machine → `/usr/bin/python3` (3.9.6): the floor, not the convenient candidate.
- All four scripts under `.claude/hooks/sdlc/` (`sdlc_context`, `sdlc_reminder`, `validate_commit_message_sdlc`, `validate_sdlc_on_stop`) under `env -i /usr/bin/python3` with `{}` on stdin → all exit 0, no stderr.
- Rendered all three global-scope commands through `_build_hook_command` against the live manifest → each leads with `/usr/bin/python3`, no bare token.

### Verification table

Every row on the plan's `## Verification` table passes on the PR branch: bare `python` in `.claude/settings.json` = 0; `/Users/` in `.claude/settings.json` = 0; shim executable; `uv run` = 0; `_SDLC_HOOK_DEFS` = 0; legacy match literal = 1; `from __future__ import annotations` = 1; `os.remove`/`unlink`/`rmtree` in `migrations.py` = 0; hook migrations registered = 2; both shim worktree-precedence rows resolve the main venv.

One cosmetic mismatch, not a finding: the "Shim fail-open is audited" row expects `grep -c` = 1, but the literal now appears twice in `hooks_audit.py` (comment + code).

## Rubric

- [ ] **1. Plan vs. implementation match** — fail — *thirteen of the fourteen non-`[EXTERNAL]` success criteria verified by execution. The fourteenth — "The shim's fail-open branch produces a dedicated finding in `reflections/audits/hooks_audit.py`" — has the code but not the effect. See Tech Debt.*
- [x] **2. New code quality** — pass — *comments are unusually load-bearing and accurate; the delta commit's rewritten `_merge_hook_settings` comment is now precise about which construct the invariant actually covers.*
- [x] **3. Test coverage** — pass — *the tests genuinely execute: real `/bin/sh` invocation of the shim against `git worktree add` fixtures, a real `env -i` interpreter run, an always-on AST walk. Each would fail on pre-fix code. The one exception is the new audit test, covered below.*
- [x] **4. Regression risk to existing callers** — pass — *`_build_hook_command` gained a required keyword-only `interpreter`; all three call sites updated. The delta commit's new test pins the `interpreter=None` removal-pass path the prior review found broken.*
- [x] **5. Data integrity** — pass — *no schema change. The `~/.claude/settings.json` edit serializes, asserts JSON validity, writes a timestamped `.bak` from the original bytes, and only then writes — in that order, verified by reading `scripts/update/migrations.py:620-640`.*
- [x] **6. Security** — pass — *precise-match allowlist (`_LEGACY_UNMARKED_SWEEP_TARGETS`), not a directory-prefix gate, so a hand-added hook under `~/.claude/hooks/` survives. The shim `exec`s a fixed path; no injection surface.*
- [ ] **7. Documentation accuracy** — fail — *`docs/features/hook-manifest.md`'s "Fail-open behavior and its audit signal" section asserts the same unreachable behavior as the Tech Debt finding. Otherwise the Interpreter Contract section is thorough and the `README.md` index line is now correct.*
- [x] **8. PR body accuracy** — pass — *all claims reproduced independently. The "90 passed" figure is stale by one after the delta commit; not a defect.*
- [x] **9. Disclosed deferrals** — pass.
- [x] **10. Follow-up claims verified** — pass — *#2521 confirmed OPEN.*

## Pre-Verdict Checklist

- **1. Plan criteria validated against diff** — PARTIAL — *13/14 pass; criterion 4 fails (Tech Debt).*
- **2. No-Gos — none violated** — PASS — *no `uv run`, no file deletion, no `_SDLC_HOOK_DEFS`, no fail-closed behavior change.*
- **3. New `except Exception` blocks** — PASS — *`resolve_global_interpreter` catches `(OSError, SubprocessError)` and falls through to the next candidate; `TimeoutExpired` subclasses `SubprocessError`, so the CLT GUI-dialog hang is covered.*
- **4. Integration tests exercise the serialization boundary** — PASS — *shim tests write real files, `git init` + `git worktree add`, and invoke `/bin/sh` under `env={}`.*
- **5. Plan internal consistency** — PASS.
- **7. New public APIs have docstrings** — PASS.
- **8. Breaking changes — migration path documented** — PASS.
- **10. Failure paths covered, not just happy path** — PASS.
- **11. UI changes — screenshot** — N/A — *no HTML/CSS/JS/JSX/Vue/template files in the diff.*
- **12. Docs updated for user-facing changes** — PARTIAL — *thorough, but see Tech Debt.*

### Blockers
- None

### Tech Debt

**File:** `reflections/audits/hooks_audit.py:187-201`
**Code:**
```python
# ... The literal matches the shim's stderr string
# byte-for-byte; the shim's fail-open test asserts the same literal,
# so the audit and the shim cannot drift apart silently.
shim_marker = "hook_python: no repo venv interpreter found"
shim_hits = sum(1 for e in recent if shim_marker in e.get("message", ""))
```
**Issue:** The dedicated fail-open detector is **unreachable in production**. It scans `<repo>/logs/hooks.log`, and that file has exactly one writer: `.claude/hooks/hook_utils/constants.py::log_hook_error()`, a Python function called from inside a hook script (`post_tool_use.py`, `stop.py`, `sdlc_reminder.py`, …). The shim's fail-open branch is precisely the branch on which **no Python runs at all** — it `echo`s to stderr and `exit 0`s, and that stderr goes to Claude Code's hook-error surface, never to `logs/hooks.log`. Nothing in the system can append the literal to the file the audit reads.

Verified, not inferred:

```
$ grep -rn "hooks.log" .claude/hooks/
.claude/hooks/hook_utils/constants.py:66:  log_path = get_project_dir() / "logs" / "hooks.log"
.claude/hooks/hook_utils/detach_lock.py:43:  ~/.claude/logs/hooks.log     # different path, not the one audited
$ tail -1 /Users/valorengels/src/ai/logs/hooks.log
2026-08-04 08:56:42 - dispatch.pre_tool_use_bash.validate_merge_guard - ERROR - RuntimeError: ...
```

Every real line in the live log carries the `TS - hook_name - ERROR - msg` shape `log_hook_error` emits. The new test at `tests/unit/test_hooks_audit.py:387` writes `f"{ts} ERROR - hook_python: no repo venv interpreter found under /Users/x/src/ai\n"` — a shape `log_hook_error` never produces, describing an event that never reaches this file. The test passes; it proves only that a substring check works on a hand-written line.

Two downstream claims are therefore false as written:

- The code comment: *"the shim's fail-open test asserts the same literal, so the audit and the shim cannot drift apart silently."* Nothing couples them. The test asserts a literal against a fixture, not against the shim's actual output path.
- `docs/features/hook-manifest.md`, *Fail-open behavior and its audit signal*: *"`reflections/audits/hooks_audit.py` scans `hooks.log` for that literal and raises a **dedicated finding** … so a machine whose shim has started failing open … is diagnosable rather than blending into ordinary hook-error noise."* It is not diagnosable. It does not even blend into the noise — it produces **no** signal in that file at all.

This is not a regression and it does not weaken the fix: the interpreter contract itself works, and I verified every part of it by execution. But the plan's Risk 5 mitigation and Success Criterion 4 are recorded as satisfied on the strength of code that cannot fire — and this PR's entire thesis is that reasoning by inspection instead of execution is what let the original bug survive two prior fixes. A dead detector for the successor failure mode is exactly the shape of thing this change exists to stop shipping.

**Severity:** tech_debt
**Fix:** Pick one, and make the docs match whichever:

1. Route the signal somewhere the audit can see. The cheapest honest version: have the shim's fail-open branch append the line to `"$root/logs/hooks.log"` in `log_hook_error`'s exact format before `exit 0` — e.g. `printf '%s - hook_python - ERROR - no repo venv interpreter found under %s\n' "$(date -u '+%Y-%m-%d %H:%M:%S')" "$root" >> "$root/logs/hooks.log" 2>/dev/null` — guarded so a failed append never changes the exit code or touches stdout. The audit check then becomes reachable, and the test can assert the shim's own output rather than a fixture.
2. Or scan the surface the message actually lands on, if one is durably readable.
3. Or drop the `hooks_audit.py` block, the docs paragraph, and Success Criterion 4, and record the fail-open branch as **unobserved** — honest, and still better than a claim of coverage that does not exist.

Option 1 keeps the criterion; option 3 is the smallest diff. Either is fine. Shipping the current state is not, because it writes a false mitigation into the feature doc.

### Nits
- None

- **`[DEFERRED]` Pruning orphaned pre-manifest files under `~/.claude/hooks/`** — tracked by **#2521 (OPEN)**. Registration-only confirmed: `grep -cE 'os\.remove|\.unlink\(|rmtree' scripts/update/migrations.py` = 0, and `test_sweep_deletes_no_files` pins it.
- **`[EXTERNAL]` `/update` on the other fleet machines** — out of scope; unreachable from here.
- **`[EXTERNAL]` Live foreign-repo manual verification** — out of scope; needs an interactive human-driven session.
- **`[EXTERNAL]` Whether hooks should become fail-closed** — out of scope; owner policy call.
- **`[EXTERNAL]` File locking around `~/.claude/settings.json` writes** — out of scope; pre-existing.

### Review Delta (vs prior review on HEAD `ba2e5ac`)

- **Resolved — Tech Debt** (`assert interpreter is not None` sat outside the add/update loop, so an empty `global_decls` plus a total probe failure raised `AssertionError` out of `/update`). `62467d4` moves the assert **inside** the loop and rewrites the comment to say the *loop body* is what is unreachable. Verified: `_merge_hook_settings(path, hooks_dir, [], None, result)` now completes with `errors == 0`, runs the removal pass, and leaves an unmarked foreign hook intact — pinned by the new `test_merge_hook_settings_runs_removal_pass_with_no_interpreter`.
- **Resolved — Nit** (`docs/features/README.md:89` was a stale inventory). Now names the scope-owned interpreter contract and pluralizes the migration clause.
- **New** — the dead audit detector above.
- Worth naming: across three passes, pass 1 asserted the `assert` was unreachable (wrong), pass 2 caught it by executing the call, and this pass finds the audit signal unreachable by tracing the writer. Every one of these survived a careful reading and died on an execution.

### Design calls worth naming as correct

- **Main-checkout-first precedence** — re-confirmed against live worktree state, not a fixture.
- **Probing by execution, not `os.path.exists`** — correctly handles the macOS CLT stub.
- **Extensionless shim name** — load-bearing, since `hooks_audit.py` takes the first `.py`/`.sh` token as the script path.
- **Precise-match sweep allowlist** over a directory-prefix gate.
- **`MIN_GLOBAL_PYTHON` enforced by an always-on AST test** rather than by the resolver. The resolver deliberately prefers `/usr/bin/python3`, the oldest and most-present candidate, so the deployed interpreter *is* the floor the AST test enforces. Consistent — and the new `manifest.toml` obligation comment closes the loop for whoever adds a fourth global hook.

### Deploy note

No `bridge/`, `agent/`, or `worker/` changes — **no service restart required**. But every effect reaches a machine only through `/update`; each fleet machine must run it locally to regenerate `~/.claude/settings.json` and apply the sweep. Correctly scoped `[EXTERNAL]`. The sweep is a **no-op on this machine** (the live `~/.claude/settings.json` no longer carries the unmarked entry), so its real-world effect is not observable here — expected for an idempotent migration, not a defect.

---

*Single-judge review. Multi-judge consensus was dispatched (`code-quality`, `risk`); neither judge returned within a ~7-minute in-turn await across two nudges each, so this is recorded as single-judge rather than stalling the pipeline. That is now the second consecutive pass with the same judge non-return and is worth its own issue against the fork-await path. The consensus rule is any-blocker-wins and this verdict is already `CHANGES REQUESTED`, so a missing judge could not have flipped a clean approval — the failure direction is conservative.*

*All findings were verified against code read and executed on the PR branch at `62467d446`.*

valorengels:
## Review: Blocked on Conflict

This PR cannot be reviewed until it merges cleanly against its base branch.

- **mergeable:** `CONFLICTING`
- **mergeStateStatus:** `DIRTY`

### Required action

Please rebase onto the current base (or merge base into your branch) and
resolve the conflicts, then push. Re-run `/do-pr-review` after the branch is
mergeable.

> No code review was performed. The mechanical mergeability gate runs before
> any diff reading — if the branch cannot merge, no amount of code review
> can make the PR mergeable.

valorengels:
## Review: Approved

**0 blockers, 0 tech debt, 0 nits.** The one Tech Debt finding from the prior pass at `62467d4` — the dead fail-open detector — is resolved, and I confirmed the fix is real by mutation, not by reading it.

Every empirical claim below was reproduced in this review at HEAD `4d41fcb`. Nothing in the PR body was credited as stated.

### The prior finding is genuinely closed

The previous pass found that `hooks_audit.py`'s dedicated shim finding was unreachable: it scans `logs/hooks.log`, whose only writer was a Python function, while the shim's fail-open branch is precisely the branch on which no Python runs. `67f0cf7d` takes Option 1 — the shim writes its own record. I verified the whole chain end to end rather than trusting the diff:

```
$ env -i /bin/sh .../hook_python -c "print('SHOULD NOT RUN')"   # non-git, no venv
EXIT=0   stdout=0 bytes
stderr:  hook_python: no repo venv interpreter found under <dir>
logs/hooks.log:
2026-08-04 10:39:12 - hook_python - ERROR - hook_python: no repo venv interpreter found under <dir>
```

Fed that real file to the audit's own parser:

```
extract_structured_errors(...) -> [{'timestamp': '2026-08-04 10:39:12', 'level': 'ERROR',
                                    'message': 'hook_python: no repo venv interpreter found under ...'}]
SHIM HITS: 1
```

The emitted format matches `log_hook_error`'s (`%Y-%m-%d %H:%M:%S - {hook} - ERROR - {msg}`, UTC) byte-for-byte, so the shim's record parses identically to every other hook error.

**Mutation test — the coupling is load-bearing, not decorative.** I stripped the log-append block from the shim and re-ran the audit test:

```
FAILED tests/unit/test_hooks_audit.py::TestShimFailOpenSignal::test_shim_fail_open_gets_dedicated_finding
  assert (tmp/repo/logs/hooks.log).exists() == False
```

It fails on pre-fix shim code and passes on post-fix code, driving the **real shim** rather than a hand-written fixture — which was the exact complaint against the previous implementation. Working tree restored clean afterward. Both the code comment and the `hook-manifest.md` paragraph that were false last pass are now accurate statements about verified behavior.

### The original bug is fixed — measured, not argued

The issue's thesis is that every generated command's bare `python` exits 127 under a stripped `/bin/sh`. I executed **all 17** generated project-scope commands from `.claude/settings.json` under `env -i CLAUDE_PROJECT_DIR=... /bin/sh -c` with `{}` on stdin:

```
17 commands, 17 exit 0, 0 occurrences of "command not found"
```

Other execution results:

- **Shim precedence against live worktrees** (not fixtures) — from a worktree with **no** `.venv`, from this worktree which **has its own**, and from the main checkout: all three resolve `sys.prefix = /Users/valorengels/src/ai/.venv`. The main-checkout-first inversion is load-bearing and I hit the divergence firsthand: this worktree's own `.venv` has neither `ruff` nor `pytest`, so a local-first shim would have exec'd a broken interpreter.
- **Global scope** — `resolve_global_interpreter()` returns `/usr/bin/python3` (3.9.6), the floor rather than the convenient candidate. All three global commands render as `/usr/bin/python3 /Users/.../<script>.py ... # hook:<id>`, no bare token.
- **All four `.claude/hooks/sdlc/` scripts** under `env -i /usr/bin/python3` with `{}` on stdin — exit 0, zero bytes of stderr each. The PEP 604 import-time crash is gone.
- **Targeted suite** — 92 passed (`test_hook_interpreter`, `test_update_hardlinks`, `test_hook_migration`, `test_hooks_audit`).
- **`ruff check .`** clean; **`ruff format --check .`** clean, 1258 files.

### Verification table

Every row on the plan's `## Verification` table passes at this HEAD: bare `python` in `.claude/settings.json` = 0; `/Users/` in `.claude/settings.json` = 0; shim executable; `uv run` = 0; `_SDLC_HOOK_DEFS` = 0; legacy match literal = 1; `from __future__ import annotations` = 1; `os.remove`/`unlink`/`rmtree` in `migrations.py` = 0; hook migrations registered = 2; audit marker occurrences = 2. Both shim worktree-precedence rows and the fail-open record row verified above. The cosmetic marker-count mismatch the prior pass noted is now corrected in the plan (expects 2, is 2).

## Rubric

- [x] **1. Plan vs. implementation match** — pass — *all 13 code-level success criteria verified by execution, including the fourteenth that failed last pass. The two remaining unticked items are stage-owned (`/do-test`, `/do-docs`).*
- [x] **2. New code quality** — pass — *comments are load-bearing and, unusually, now all true. `_GLOBAL_INTERPRETER_PROBE_TIMEOUT` carries the required grain-of-salt provisional note; `MIN_GLOBAL_PYTHON` is declared once.*
- [x] **3. Test coverage** — pass — *16 tests spanning every criterion. They execute rather than assert on strings: real `/bin/sh`, real `git worktree add`, real `env -i`, an always-on AST walk. Mutation-tested above.*
- [x] **4. Regression risk to existing callers** — pass — *`_build_hook_command` gained required keyword-only `interpreter`; all three call sites (`hardlinks.py:806`, `hardlinks.py:1013`, `migrations.py:807`) updated. The `interpreter=None` removal-pass path is pinned by its own test.*
- [x] **5. Data integrity** — pass — *no schema change. The `~/.claude/settings.json` write order is serialize → assert JSON-valid → write `.bak` from the **original** bytes → write. Empty-block pruning is computed but discarded when nothing was swept, so a no-op run writes nothing.*
- [x] **6. Security** — pass — *precise-match allowlist rather than a directory-prefix gate, so the hand-added `calendar_prompt_hook.sh` and any foreign hook survive (pinned by fixture). The shim `exec`s a fixed path; no injection surface. No secrets.*
- [x] **7. Documentation accuracy** — pass — *the `hook-manifest.md` fail-open paragraph asserted unreachable behavior last pass; I re-read it against the now-verified behavior and it is accurate. `README.md` index line updated.*
- [x] **8. PR body accuracy** — pass — *see the note on the test count below; it understates and cannot mislead.*
- [x] **9. Disclosed deferrals** — pass.
- [x] **10. Follow-up claims verified** — pass — *#2521 confirmed **OPEN**.*

## Pre-Verdict Checklist

- **1. Plan criteria validated against diff** — PASS — *13/13 code-level criteria verified by execution.*
- **2. No-Gos — none violated** — PASS — *no `uv run`, no file deletion, no `_SDLC_HOOK_DEFS` reintroduction, no fail-closed behavior change.*
- **3. New `except Exception` blocks** — PASS — *`resolve_global_interpreter` catches `(OSError, SubprocessError)` and falls through to the next candidate; `TimeoutExpired` subclasses `SubprocessError`, so the CLT GUI-dialog hang is covered.*
- **4. Integration tests exercise the serialization boundary** — PASS — *real files, real `git worktree add`, real `/bin/sh` under `env={}`.*
- **5. Plan internal consistency** — PASS.
- **6. No hardcoded secrets or debug artifacts** — PASS — *the only `print(` in the diff is inside the interpreter version probe.*
- **7. New public APIs have docstrings** — PASS.
- **8. Breaking changes — migration path documented** — PASS — *the signature change is internal; the deployed-state change is carried by a registered migration.*
- **10. Failure paths covered, not just happy path** — PASS — *fail-open, total probe failure, `interpreter=None` removal pass, and foreign-hook survival are all pinned.*
- **11. UI changes — screenshot** — N/A — *no HTML/CSS/JS/JSX/Vue/template files in the diff.*

### Blockers
- None

### Tech Debt
- None

### Nits
- None

### Miscellaneous
- None

- **`[DEFERRED]` Pruning orphaned pre-manifest files under `~/.claude/hooks/`** — tracked by **#2521 (OPEN)**. Registration-only confirmed: 0 file-deletion calls in `migrations.py`, pinned by `test_sweep_deletes_no_files`.
- **`[EXTERNAL]` `/update` on the other fleet machines** — out of scope; unreachable from here.
- **`[EXTERNAL]` Live foreign-repo manual verification** — out of scope; needs an interactive human session.
- **`[EXTERNAL]` Whether hooks should become fail-closed** — out of scope; owner policy call.
- **`[EXTERNAL]` File locking around `~/.claude/settings.json` writes** — out of scope; pre-existing.

### Claims I did not verify

Per this repo's rule that a number in a PR body is a claim rather than evidence, two items are recorded as **unverified** rather than credited:

- **"Targeted suite: 90 passed."** I measured **92** at this HEAD. The body is stale by two after later commits. It understates, so it cannot mislead — not a finding, but the review is written against my own measurement.
- **"Full-suite failures are pre-existing worktree env artifacts, unrelated to the diff."** I did not run the full suite and am not crediting this. The blast radius is contained (`scripts/update/`, `.claude/hooks/`, `reflections/audits/`, tests), all three `_build_hook_command` call sites are covered by passing tests, and lint/format are clean across all 1258 files — but the full-suite claim itself remains unconfirmed by this review.

### Review Delta (vs prior review on HEAD `62467d4`)

- **Resolved — Tech Debt** (dead fail-open detector). `67f0cf7d` has the shim write its own `hooks.log` record in `log_hook_error`'s format. Verified by execution end to end *and* by mutation: removing the append fails the audit test.
- **Resolved — Rubric 1 and 7** (Success Criterion 4 and the corresponding `hook-manifest.md` paragraph both asserted behavior that could not fire). Both now describe verified behavior.
- **Resolved — preflight.** The prior pass short-circuited on `CONFLICTING`/`DIRTY`; the branch is now `MERGEABLE`/`CLEAN`.
- **New** — none.
- Worth naming: across five passes, every finding that survived a careful reading died on an execution — the `assert` reachability, the audit signal, and now the mutation test. That is the PR's own thesis holding against the PR itself.

### Plan checkbox sync

Ticked 13 verified success criteria in `docs/plans/sdlc-hook-interpreter-resolution.md` and pushed as `4d41fcb` **before** posting this review, so it is strictly newer than every commit on the branch. `Tests pass (/do-test)` and `Documentation updated (/do-docs)` are left unticked for their own stages.

### Design calls worth naming as correct

- **Main-checkout-first precedence** — re-confirmed against live worktree state, and this review independently hit the divergent-venv case it exists to defend against.
- **Probing by execution rather than `os.path.exists`** — correctly rejects the macOS CLT stub.
- **`MIN_GLOBAL_PYTHON` as an authoring floor enforced by an always-on AST test, not as a resolver gate.** I checked whether the resolver ignoring version was a gap: it is not. The resolver prefers `/usr/bin/python3`, the oldest and most-present candidate, so the deployed interpreter *is* the floor the AST test enforces, and the docs say exactly that.
- **Extensionless shim name** — load-bearing, since `hooks_audit.py` takes the first `.py`/`.sh` token as the script path.
- **Precise-match sweep allowlist** over a directory-prefix gate.

### Deploy note

No `bridge/`, `agent/`, or `worker/` changes — **no service restart required**. Every effect reaches a machine only through `/update`, which each fleet machine must run locally to regenerate `~/.claude/settings.json` and apply the sweep. Correctly scoped `[EXTERNAL]`. The sweep is a **no-op on this machine** — the live `~/.claude/settings.json` carries only the three marked SDLC entries plus the foreign `calendar_prompt_hook.sh`, and the unmarked `validate_no_raw_redis_delete.py` entry is already gone. Expected for an idempotent migration, not a defect.

---

*Single-judge review. Two judge subagents for this PR were queried in-turn for a verdict before finalizing; no verdict was returned, matching the judge non-return seen on the two prior passes (worth its own issue against the fork-await path). Noting the direction honestly: unlike the prior passes, this verdict is APPROVED, so a missing judge is **not** conservative here — the aggregate rests on this reviewer's own execution-based verification rather than on consensus.*

*All findings were verified against code read, executed, and mutation-tested on the PR branch at `4d41fcb`.*

valorengels:
## Review: Approved

**0 blockers, 0 tech debt, 0 nits.** Sixth pass, at HEAD `18cb09b4`. The prior APPROVED verdict was recorded against `4d41fcb`; the only commit since is `18cb09b4`, the `/do-docs` cascade. This pass re-reviews that delta and re-runs the full verification table against the new HEAD, so the recorded verdict is no longer stale relative to the branch tip.

The delta is documentation-only (`.claude/skill-context/audit-hooks.md`, `docs/features/hook-manifest.md`, `docs/features/hooks-best-practices.md`, `docs/plans/...`, `tests/README.md` — +83/-17, no code). Every factual claim the new prose makes was checked against the code it describes rather than taken at face value, and all of them hold.

## Rubric

- [x] **1. Plan vs. implementation match** — pass — *All 13 substantive success criteria verified. Only `Tests pass (/do-test)` remains unticked; see "Plan checkbox sync" below for why I left it alone rather than ticking it.*
- [x] **2. New code quality** — pass — *No code changed in this delta. Re-verified the shim's shell quoting (`"$root"`, `"$common"`, `"$@"` all quoted; safe under paths with spaces) and that its fail-open append is guarded so it can neither touch stdout nor alter the exit code.*
- [x] **3. Test coverage** — pass — *Targeted suite green at this HEAD: 92 passed across `test_hook_interpreter`, `test_update_hardlinks`, `test_hook_migration`, `test_hooks_audit`. Tests assert real `env -i` execution, not mocks.*
- [x] **4. Regression risk to existing callers** — pass — *No signature changes in this delta. `_build_hook_command`'s required `interpreter` arg is unchanged since the prior pass and all call sites remain covered by passing tests.*
- [x] **5. Data integrity** — n/a — *No models, schema, or persisted-format changes.*
- [x] **6. Security** — pass — *No new `eval`/`exec`-on-input, no secrets. Re-read the sweep migration's gate: `_should_remove` requires all three of a `~/.claude/hooks/` reference, absence of a `# hook:` marker, and a precise-match against `_LEGACY_UNMARKED_SWEEP_TARGETS` — a full absolute path, not a directory prefix. A hand-added user hook cannot match.*
- [x] **7. Documentation accuracy** — pass — *This is the delta's whole surface; each claim verified individually below.*
- [x] **8. PR body accuracy** — pass — *One stale number, noted under "Claims I did not verify". It understates, so it cannot mislead.*
- [x] **9. Disclosed deferrals** — pass — *One deferral: the orphaned-file prune. Confirmed `#2521` is OPEN.*
- [x] **10. Follow-up claims verified** — pass — *`#2521` OPEN, title "Prune orphaned pre-manifest hook files under ~/.claude/hooks/" — it is the issue the disclosure claims it is, not just a live number.*

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — PASS — *13 of 14 verified by execution; the 14th is the TEST stage's own.*
- **2. No-Gos from plan — none violated** — PASS — *Anti-criteria re-run: no `uv run` (0), no `_SDLC_HOOK_DEFS` (0), no file-deleting call against `~/.claude/hooks/`, legacy match literal preserved (1).*
- **3. New `except Exception` blocks — each has logger/raise/swallow-ok** — N/A — *None added in this delta.*
- **4. New integration tests — exercise serialization boundary (not in-memory only)** — PASS — *Shim tests run the real script under `env -i` and read the real `logs/hooks.log` it writes.*
- **5. Plan internal consistency — spike findings match task steps** — PASS — *The 9-of-57 divergent-venv finding is what the main-checkout-first precedence implements, and I hit that exact case live (below).*
- **6. No hardcoded secrets or debug artifacts** — PASS — *`grep -c '/Users/' .claude/settings.json` is 0; no machine-specific path is committed.*
- **7. New public APIs — docstrings present** — N/A — *No new APIs in this delta.*
- **8. Breaking changes — migration path documented** — PASS — *The sweep migration is registered in `MIGRATIONS` and documented under its own heading in `hook-manifest.md`.*
- **9. Tests added for new behavior** — PASS — *Covered at the prior pass; re-run green here.*
- **10. Tests cover the failure path (not just happy path)** — PASS — *The shim's fail-open branch has dedicated tests and I reproduced it by hand.*
- **11. UI changes (if any) — screenshot captured** — N/A — *No HTML, CSS, JS/TS, JSX/TSX, Vue, or template files in the diff. The visual-proof gate is a no-op.*
- **12. Docs updated for user-facing changes** — PASS — *This delta is that update.*

### Blockers
- None

### Tech Debt
- None

### Nits
- None

### Miscellaneous
- None

- **Orphaned pre-manifest files under `~/.claude/hooks/` are not pruned; the migration is registration-only and deletes no files** — tracked by **#2521 (OPEN)**. Verified as a real, open, correctly-scoped issue, and verified as true of the code: no `os.remove` / `unlink` / `rmtree` targeting that directory appears anywhere in the diff. The deferral is also the safer half of the split — deregistering makes the files inert, so the residue is dead weight rather than live risk.

### Verification Results

Every row of the plan's `## Verification` table re-run against `18cb09b4`. All pass.

- Targeted tests pass — **92 passed** (4 files)
- Lint clean — `ruff check .` **All checks passed!**
- Format clean — `ruff format --check .` **1258 files already formatted**
- No bare `python` in generated project commands — **0**
- No machine-specific path committed — **0**
- Shim exists and is executable — **mode 755, confirmed**
- Shim resolves the venv from a worktree — **`Python 3.` under `env -i`, confirmed**
- Shim prefers the MAIN venv from a worktree that has its own — **confirmed, see below**
- Migration deletes no files — **no deleting call targets `~/.claude/hooks/`**
- Shim fail-open emits an auditable record — **confirmed, see below**
- Shim fail-open is audited — **2** (`shim_marker` literal + finding text)
- Global script runs under the minimum interpreter — **exit 0** under `env -i /usr/bin/python3`
- `__future__` import present in the fixed script — **1**
- No `uv run` reintroduced — **0**
- `_SDLC_HOOK_DEFS` not reintroduced — **0**
- Legacy match literal preserved — **1**
- New migration registered — **2** hook migrations in `MIGRATIONS`

Two rows I re-executed rather than reasoned about, because they are the PR's load-bearing claims:

**Main-checkout-first precedence, against a genuinely divergent case.** Run from this PR's own worktree, which carries its own `.venv`, with `CLAUDE_PROJECT_DIR` pointed at that worktree, the shim reported `sys.prefix` = `/Users/valorengels/src/ai/.venv` — the main checkout's, not the local one. This is the exact failure the precedence order exists to prevent, and the review hit it live rather than taking the plan's word for it.

**Fail-open emits its own auditable record.** With `CLAUDE_PROJECT_DIR` set to a non-repo temp directory under `env -i`: exit **0**, stdout **empty** (so a hook's JSON parser is never corrupted), stderr carrying the message, and `logs/hooks.log` created containing `<ts> - hook_python - ERROR - hook_python: no repo venv interpreter found under <dir>` — byte-compatible with `log_hook_error`'s format. The `shim_marker` literal in `reflections/audits/hooks_audit.py` matches that string exactly, so the dedicated audit finding is genuinely reachable. This branch runs no Python at all, which is precisely why the shim must write the record itself.

### Documentation delta — each claim checked against the code

- **`hooks-best-practices.md` repoints `BEST_PRACTICES.md` from `.claude/skills/audit-hooks/` to `.claude/skills-global/audit-hooks/`.** Verified: the `skills-global` path exists, and `.claude/skills/audit-hooks/` does not. The old link was broken; this fixes it rather than merely moving it.
- **`hook-manifest.md#interpreter-contract` anchor.** Resolves — `## Interpreter Contract` is a real heading (line 59), and the two inbound links from `audit-hooks.md` and `hooks-best-practices.md` land on it.
- **`audit-hooks.md` claims the shim's fail-open logs `hook_python: no repo venv interpreter found` and that the audit raises a dedicated finding on it.** Both true, and verified by execution rather than by reading — the marker string in the shim and the `shim_marker` literal in `hooks_audit.py` are the same bytes.
- **`audit-hooks.md`'s Rule 6 carve-out for the shim's `exec`.** Correct and worth having written down: `exec` is what preserves the wrapped hook's exit code, which is what preserves `blocking = true` semantics. Without the carve-out the audit would flag its own load-bearing mechanism.
- **`docs/features/README.md` index entry** rewritten to mention the interpreter contract and the *second* migration. Matches the diff — there are now two hook migrations registered, and I counted them.
- **`tests/README.md` row: `test_hook_interpreter.py` | 16.`** Correct. I initially flagged this as wrong, since pytest collects **22**. Checking the table's convention against its neighbours killed the finding: `test_remote_update.py` is listed as 29 with 31 collected, `test_config_bootstrap.py` as 13 with 13. The column counts `def test_` functions, not parametrized cases, and this file has exactly 16. Recording the near-miss because it is the kind of finding that reads as obviously true and is not.

### Claims I did not verify

Per this repo's rule that a number in a PR body is a claim rather than evidence:

- **"Targeted suite: 90 passed."** I measured **92** at this HEAD. Same discrepancy as the prior pass — the body is stale by two after later commits. It understates, so it cannot mislead, and I am not raising it as a finding; the review is written against my own measurement.
- **"Full-suite failures are pre-existing worktree env artifacts, unrelated to the diff."** I did not run the full suite and am **not** crediting this. The blast radius is contained (`scripts/update/`, `.claude/hooks/`, `reflections/audits/`, docs, tests), all `_build_hook_command` call sites are covered by passing tests, and lint/format are clean across all 1258 files — but the full-suite claim itself remains unconfirmed by this review.

### Review Delta (vs prior review on HEAD `4d41fcb`)

- **New** — none.
- **Resolved** — none outstanding; the prior pass closed its last Tech Debt finding (the dead fail-open detector) before approving.
- **Unchanged** — the stale "90 passed" count in the PR body, still recorded as an unverified claim rather than a finding.
- **Re-verified rather than assumed** — every verification-table row was re-run against `18cb09b4` rather than carried forward from `4d41fcb`. On a docs-only delta that is mostly ceremony, but it is what makes the recorded verdict fresh against the branch tip instead of stale by one commit, which is the entire reason this pass exists.

### Plan checkbox sync

**No plan-file write this pass.** Thirteen criteria were already ticked and verified; `Documentation updated (/do-docs)` was ticked by the docs commit itself. That leaves `Tests pass (/do-test)`.

I deliberately left it unticked. The TEST stage is recorded `completed` in the substrate and the targeted suite is green under my own hand, but "Tests pass" as written refers to the full-suite result — and I just declined to credit the PR body's full-suite claim. Ticking it on the strength of a stage marker would be crediting through the plan file exactly what I refused to credit in the review body. Treated as `n/a` (no write, existing state preserved) rather than `pass`.

### Deploy note

No `bridge/`, `agent/`, or `worker/` changes — **no service restart required**. Every effect reaches a machine only through `/update`, which each fleet machine runs locally to regenerate `~/.claude/settings.json` and apply the sweep. The sweep is a **no-op on this machine**: the unmarked `validate_no_raw_redis_delete.py` entry is already gone. Expected for an idempotent migration, not a defect.

---

*Judge dispatch: two judge subagents (code-quality, risk) were dispatched and awaited in-turn, with two follow-up prompts across roughly four minutes. Neither returned a verdict — the same non-return seen on all prior passes of this PR, which is worth its own issue against the fork-await path. This verdict therefore rests on a single reviewer, and the OUTCOME block omits the consensus fields rather than claiming a consensus that did not happen. Stating the direction plainly: the verdict is APPROVED, so a missing judge is **not** conservative here. The mitigation is that every load-bearing claim in this review was verified by executing the code, not by reading it.*

Issue #2544, opened 2026-08-06
Five rename-style migrations still call the unguarded raw rebuild_indexes()


valorengels:
Found while building #2524. Recording rather than expanding that PR's scope.

## Finding

#2524 removed the raw `AgentSession.rebuild_indexes()` call from the three *strip* migrations, replacing it with the production-safe `clean_indexes()` orphan sweep. A repo-wide grep shows five **rename-style** migrations still call it:

```
scripts/migrate_agent_session_keyfield_rename.py:180
scripts/migrate_parent_session_field.py:161
scripts/migrate_session_type_pm_to_eng.py:321
scripts/migrate_session_type_chat_to_pm.py:155
scripts/migrate_unify_parent_session_field.py:110
```

Two of these are live registry entries that still run on a machine that has not completed them: `agent_session_keyfield_rename` and `unify_parent_session_field`.

## Why it matters

The raw rebuild is the call that *opens* the #1720 class-set window: it deletes and re-adds `$Class:AgentSession` in batches, and any concurrent `query.all()` landing inside that window returns 0 rows with no exception. Migrations run at `/update` Step 3.6, **before** the service restart, so they are genuinely concurrent with a live worker. It also currently fails outright with `unpack(b) received extra data` on pre-existing phantom index metadata (#2536), and it bypasses the `$IndexF` cleanup and the #2101/#2207 phantom-re-inflation shim.

## Why it was not fixed in #2524

#2524 names only the two sibling *strip* migrations. These five are a different pattern with a different shape — they rename hash fields rather than delete them, and unlike the strip family they genuinely depend on index reconstruction rather than using it as a defensive sweep, so `clean_indexes()` may not be a drop-in substitute. That needs its own judgement per script, not a blanket find-and-replace.

All five are recorded complete on every current machine, so editing them in place is inert without the same rename-and-rerun decision #2524 made for the strip family.

## Work

- [ ] Per script, decide whether the rebuild is load-bearing (rename migrations may need it) or defensive (swap to `clean_indexes()`).
- [ ] For any that are load-bearing, document why the #1720 window is acceptable there, or gate the call.
- [ ] Decide reclaim-vs-accept for each, same as #2524's Decision 1.

## References

- #2524, #2518, #1720, #2536, #2101, #2207
- `scripts/_strip_migration.py` — the shared engine and the reasoning for `clean_indexes()`

valorengels:
[...]

Issue #2543, opened 2026-08-06
Zero-record migration guard fails closed forever on a genuinely empty keyspace


valorengels:
Split out from #2524 (its No-Gos section), which deliberately declined to relitigate this.

## Problem

`scripts/_strip_migration.py::run_strip_migration` fails closed when `AgentSession.query.all()` returns zero rows, because an empty scan is indistinguishable from one blinded by popoto's index-rebuild class-set window (#1720). The migration exits 2, `run_pending_migrations` never records it complete, and the next `/update` retries.

That is correct on a populated machine. On a **genuinely empty keyspace** — a fresh install — it is a permanent, unbounded failure: every `/update` runs the migration, sees zero records, and fails, forever.

#2518 saw this and accepted it, documenting the recurring `FAIL:` line as expected output. The accepted-consequence comment is still in the code:

> ACCEPTED CONSEQUENCE, by design: on a machine whose AgentSession keyspace is legitimately empty (a fresh install), every migration routed through this engine exits non-zero on EVERY `/update`, indefinitely [...] The recurring `FAIL:` lines are EXPECTED OUTPUT, not a live regression.

#2524 **multiplied the blast radius**: it registered `strip_pty_session_fields_v2` and `schema_diet_fields_v2` alongside the existing `strip_pid_fields_v2`, so a fresh machine now emits three permanent failures instead of one, all through the same shared engine.

Three recurring `FAIL:` lines that an operator is supposed to know are fine is a bad steady state. It trains people to ignore migration failures in `logs/update.log`, which is the one place a real one would show up.

## Proposed fix

Disambiguate the two cases the guard currently conflates. A detection-only `SCAN` for `AgentSession:*` hash **key names** (no value reads, so the #1038 binary-decode hazard does not apply — the same discipline the engine's `HKEYS` detection already uses) answers the question directly:

- SCAN finds zero keys -> keyspace is genuinely empty -> **succeed**, record complete.
- SCAN finds keys but `query.all()` returned zero -> the index really is blinding the query -> **fail closed**, exactly as today.

This makes the guard strictly more precise. It never weakens the #1720 protection; it only stops firing in the one case where the answer is knowable.

## Why it was not done in #2524

#2524 was a generalization of an existing pattern into a shared engine. Reversing a decision the parent issue's critique had explicitly taken belongs in its own change with its own review, not smuggled into a refactor.

## Acceptance criteria

[...]

PR #2538, opened 2026-08-05
Durability M1 fence: close nine unfenced consumers + permanent regression tests (#2518)

valorengels:
## Summary

Closes the durability M1 pid-fence gap left by #2516. That PR added `agent/pid_fence.py::fence_is_live` and applied it at nine decision sites, but at six others it rebound the pid source to `fence.get("pid")` and **discarded the `create_time` sitting in the same dict**. Every one kept compiling and kept passing tests, because the unguarded shape is still valid code. Review APPROVEd it twice.

This PR fences the remaining consumers, makes the legacy-row rule canonical, fixes the migration's observability hole, and adds an anti-criterion that can actually fail.

## The two HIGH sites fail in opposite directions

- **`_tier2_reprieve_signal` under-kills.** A recycled pid probes as "progressing" and reprieves every tick; `pid is not None` has been permanently true since #2494 stopped clearing the fence, so a dead session is held alive indefinitely. Fencing it is **kill-increasing**, so it ships **Phase A log-only** behind `# PHASE A — DELETE IN PHASE B`. Enforcement is deferred to Task 13, pending human review of the canary shadow log. No config flag: the switch is git, the evidence is the log.
- **`_has_progress` over-kills.** An unrelated process probing as "hung" bypasses the sticky-field honor and prematurely releases a progressing session to Tier-2 recovery. Fencing it is strictly **kill-reducing**, so it ships enforcing, with no shadow, and emits nothing in the canary by design.

## Also fenced

`_owned_task_hang_check`; the staged SIGKILL drain (`_pending_sigkill` now holds `(pid, create_time)` tuples re-verified at drain, because the 300s tick is the same order as the macOS recycle window); the ownership forward scan; `/update`'s stale-session cleanup; and the dashboard liveness probe.

## Legacy-row rule, made canonical

Written into `agent/pid_fence.py`: an unreadable or absent `create_time` on either side means **unknown**; unknown never authorizes a kill, but may authorize the gentler action already in place. Three divergent behaviors collapse to one. The pre-cancel snapshot stops failing open into a real SIGTERM, and the in-process orphan reap stops never-reaping legacy rows (SIGTERM on a liveness probe, but no SIGKILL escalation without a fence match).

[...]

Issue #2536, opened 2026-08-05
Investigate: AgentSession rebuild_indexes() fails with 'unpack(b) received extra data' on phantom index metadata


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/phantom-index-metadata-2536.md

## Recon Summary

Read-only investigation completed 2026-08-06 on host `Valor the Pirate`. Full evidence in [this comment](https://github.com/tomcounsell/ai/issues/2536#issuecomment-5200457495). **The issue's original premise (phantom index metadata, #2207 class) is disproven.**

### Confirmed

- `ExtraData: unpack(b) received extra data.` originates at `popoto/models/encoding.py:327-331` (**popoto 1.7.1** `decode_popoto_model_hashmap`), reached from `popoto/models/base.py:2805` inside `rebuild_indexes()` Step 2.
- The values it chokes on are popoto 1.8.0's `IndexedField` pointer fields `{field}\x00idxset`, stored inside each model hash. On `AgentSession` there are three: `status`, `task_type`, `claude_session_uuid`. They are raw Redis Set-key strings, deliberately not msgpack-encoded.
- **popoto 1.8.x skips them; 1.7.1 does not.** `popoto/models/encoding.py:324, 338, 422, 427` in the 1.8.1 dist carry `if b"\x00" not in key_b  # skip internal pointer fields (e.g. \x00idxset)`. 1.7.1 has no such guard.
- Read-only probe replaying `rebuild_indexes()` Step 2 (`scan_iter` → `hgetall` → `decode_popoto_model_hashmap`), zero writes:

  | Interpreter | popoto | records considered | decode failures |
  |---|---|---|---|
  | `.venv/bin/python` 3.14.3 (what launchd runs) | 1.8.0 | 4005 | **0** |
  | ambient `python3` 3.12.13 (homebrew + user site-packages) | 1.7.1 | 4005 | **4005** |

- `class_set_members=4005`, `scanned=4005`, `skipped_empty=0`, `non_hash_types={}` — every hash is in the class set and every hash decodes cleanly under the pinned popoto. **The keyspace is healthy.**
- `pyproject.toml:18` pins `popoto>=1.8.0`. `~/Library/Python/3.12/lib/python/site-packages/popoto` is **1.7.1**, below the floor. `uv sync` only writes `.venv`, so nothing on the machine corrects the ambient install.

### Revised (from the issue's original claims)

[...]

PR #2545, opened 2026-08-06
Generalize the migration zero-record guard and guarded index repair to sibling strip migrations

### LINKED ISSUE #2524: Generalize the migration zero-record guard and guarded index repair to sibling strip migrations
Split out from #2518 (see its Recon Summary, defects D1 and D6).

## Context

`scripts/migrate_strip_pid_fields.py` shipped in PR #2516 with two defects that it inherited by template from earlier strip migrations:

- **No `total_records == 0` guard.** A scan that sees nothing is indistinguishable from a clean machine: `errors=0` -> exit 0 -> recorded complete permanently. This fired on the canary machine, where `strip_pid_fields` is recorded complete while 20 terminal records still carry all four stale fields. Root cause is plausibly the `query.all()` / `rebuild_indexes()` class-set window documented at `agent/index_drift.py:1-12` (#1720), since migrations run at `/update` Step 3.6 *before* the service restart, i.e. against a live worker.
- **Unguarded `AgentSession.rebuild_indexes()`.** Bypasses the repo's `repair_indexes()` wrapper (`models/agent_session.py:2210`), which clears `$IndexF` keys popoto does not touch and installs the #2101/#2207 phantom-re-inflation shim.

#2518 fixes both in `migrate_strip_pid_fields.py` and re-registers it as `strip_pid_fields_v2` so every machine re-runs it once.

## Scope of this issue

The same two defects exist in the siblings, which #2518 deliberately leaves alone:

- `scripts/migrate_strip_pty_fields.py:161` — unguarded `rebuild_indexes()`, no zero-record guard
- `scripts/migrate_schema_diet_fields.py:230` — same

Both are already recorded complete on every machine, so editing them in place is inert — they need their own rename-and-rerun decision, which is a separate judgement about whether their stale fields are worth reclaiming at all.

Additionally: consider **generalizing the zero-record guard into `run_pending_migrations()`** in `scripts/update/migrations.py`, so every future migration inherits fail-closed-on-ambiguous-scan behavior rather than each script reimplementing it.

## Desired outcome

- A decision recorded for each sibling migration: reclaim (rename + rerun) or accept the orphaned fields and document why.
- The zero-record guard lives in one place rather than three.
- Neither sibling calls unguarded `rebuild_indexes()`.

## References

- Parent issue #2518, plan `docs/plans/durability-m1-fence-canary.md`
- PR #2516, #1720, #2101, #2207

valorengels:
Closes #2524

**Plan:** `docs/plans/generalize-migration-guards-2524.md`

## What was wrong

The three AgentSession "strip removed hash fields" migrations were near-verbatim clones. #2518 / PR #2538 hardened one of them (`migrate_strip_pid_fields`) with a zero-record guard and a production-safe `clean_indexes()` sweep. The other two drifted: `migrate_strip_pty_fields.py:161` and `migrate_schema_diet_fields.py:230` still reported success on an empty scan and still called popoto's raw `rebuild_indexes()` — the very call that *opens* the #1720 class-set window the guard insures against, and which currently fails outright on pre-existing phantom index metadata (#2536).

Fixing them by copy-paste would have reproduced the failure mode that created the issue.

## What changed

**One engine, three thin scripts.** `scripts/_strip_migration.py` (new) holds the single copy of the scan, the terminal-only atomic rewrite, the zero-record guard, the `clean_indexes()` sweep, and the exit-code contract (2 = blinded scan, 1 = per-record errors, 0 = clean). The three scripts keep only their module docstring and their `STALE_FIELDS` set. All three now log to stdout, which the update-log capture depends on.

**Both siblings re-registered as `_v2`** (`strip_pty_session_fields_v2`, `schema_diet_fields_v2`) so every machine re-runs them once. Migrations are skipped by name, so a rename is the auditable re-run mechanism — no hand-editing of `data/migrations_completed.json`. Each `_v2` has its own helper function purely so its captured output is attributable in `logs/update.log`; `strip_pid_fields_v2` is deliberately exempt because its label is #2518's canary gate artifact.

**Output capture generalized** via `_run_migration_script`, resolving the "tracked separately (#2524)" deferral in `_migrate_strip_pid_fields`'s docstring. No `MIGRATIONS` contract widening was needed — output goes to the module logger, so the `None | error-string` return is untouched. Six subprocess-shaped helpers now log both streams on the success path and return a both-tails error string.

**Three vacuous assertions fixed.** `tests/unit/test_migrate_strip_pid_fields.py` patched `AgentSession.repair_indexes`, a method hotfix `369d782c8` had already replaced with `clean_indexes()`. They passed against any implementation, including a completely broken one.

## The reclaim-vs-accept decision

The issue asked for a recorded decision per sibling: reclaim (rename + rerun) or accept the orphaned fields.

**Decided: reclaim both.** Dry-run evidence gathered on host *Valor the Pirate* (read-only):

```
migrate_strip_pty_fields:   {'total_records': 4006, 'clean': 4006, 'stripped': 0, 'deferred_non_terminal': 0, 'errors': 0}
migrate_schema_diet_fields: {'total_records': 4006, 'clean': 4006, 'stripped': 0, 'deferred_non_terminal': 0, 'errors': 0}
```

Both siblings are genuinely clean here — 4006 records scanned, so not a blinded zero-scan. That does **not** generalize to the fleet, and that asymmetry is the whole argument: the canary machine's `strip_pid_fields` was recorded complete while 20 terminal records still carried all four stale fields. "Recorded complete" has been demonstrated not to imply "actually did the work". The scripts are idempotent, so a clean machine pays one ~4k-record scan and gets a log line proving it is clean; a machine in the canary's state gets its fields reclaimed instead of keeping them forever with no signal.

Incidental observation from the same run: `migrate_strip_pid_fields` currently **defers 44 non-terminal records** on this host that still carry stale pid fields. That is the deferral path working as designed, not a defect.

## Test evidence

All runs from the lane worktree with `scripts/pytest-clean.sh`, `-n0`, focused paths.

```
./scripts/pytest-clean.sh tests/unit/test_strip_migration_shared.py \
  tests/unit/test_migrate_strip_pid_fields.py \
  tests/unit/test_migrations.py tests/unit/test_sdlc_stubs.py -n0 -q
  -> 97 passed in 27.54s
```

`python -m ruff check scripts/ tests/...` -> All checks passed. `ruff format --check` -> 4 files already formatted.

End-to-end dry-run of all three refactored scripts against live Redis: all exit 0 with the stats above.

Success-criterion gate: `grep -c rebuild_indexes` over `scripts/_strip_migration.py` and the three delegates -> **0, 0, 0, 0**.

No test failure was dismissed as pre-existing; there were none.

## Critique

Two passes. The first (plan-stage) raised one BLOCKER — an unsatisfiable Success Criterion — plus a CONCERN and a NIT. The second, adversarial pass ran against the built branch and verified the pid migration is behavior-preserving line by line, and found no path by which a non-terminal or in-flight row can be rewritten. Both passes' findings are recorded with resolutions in the plan's Critique Results section. Notable ones:

- `field_names` is a **required** keyword argument, not an optional one. A default was the real hazard: a caller omitting it would still run correctly while silently detaching the module-level `_raw_field_names`, turning every `patch.object(mod, "_raw_field_names", ...)` in the suite vacuous. Two tests pin it.
- Two "guard returns before the sweep" assertions were tautologies (with zero records `stripped` is 0, so the sweep gate already skips it). Replaced with an ordering assertion on the last emitted log record.
- `strip_migration_main` now reports a malformed stats dict with an attributed error and exit 1, rather than raising a bare `KeyError` traceback into `logs/update.log`.
- The `_run_migration_script` tests are functional (throwaway scripts, real subprocesses) rather than `inspect.getsource` token greps — the anti-pattern the pid test file's own docstring warns about.

## Deliberately out of scope

- **The guard's fail-closed behavior on a genuinely empty keyspace is unchanged.** #2518 deliberated and accepted it. This PR multiplies it from one recurring `/update` failure to three on a fresh install, which is worth fixing — but reversing a decision from the parent issue's critique belongs in its own change. Filed as **#2543**.
- **Five rename-style migrations still call the raw `rebuild_indexes()`** (`agent_session_keyfield_rename`, `parent_session_field`, `session_type_pm_to_eng`, `session_type_chat_to_pm`, `unify_parent_session_field`). Different pattern, different shape — they may genuinely depend on index reconstruction rather than using it as a defensive sweep, so `clean_indexes()` is not a blanket substitute. Filed as **#2544**.

## Operational note

No services were started or restarted — this is a build-only lane. Every migration run here was **dry-run**; the `--apply` path against production Redis is an operator action performed by `/update` after merge.

---

## Full-suite gate + failure classification

Final pre-merge full-suite run in this lane's worktree:

```
./scripts/pytest-clean.sh -q
  -> 45 failed, 12786 passed, 24 skipped, 782 warnings in 942.72s (15:42)
```

None of the 45 failures is in code this branch touches. Per the no-dismissal rule they were **classified against main rather than assumed pre-existing**. A 12-test representative sample was chosen to include every failure whose name touches migrations, update, popoto, index, or redis cleanup, and each was re-run isolated (`-n0`) on main in a clean checkout and on this branch:

| Node id | Fails on main | Fails on branch isolated | Classification |
|---|---|---|---|
| `test_remote_update.py::TestRemoteUpdateScript::test_already_up_to_date` | no | no | full-suite pollution |
| `test_remote_update.py::TestRestartFlag::test_check_restart_flag_defers_when_jobs_running` | yes | yes (identical) | pre-existing |
| `test_reflections_redis.py::TestRedisIndexCleanupReflection::test_reflection_entry_structure` | yes | yes (identical) | pre-existing |
| `test_agent_session_lifecycle.py::TestSummarizeWithSession::test_long_response_with_sdlc_session` | yes (ImportError StructuredDraft) | yes | pre-existing |
| `test_agent_session_scheduler.py::TestJobSchedulerCLI::test_push_and_pop` | no | no | full-suite pollution |
| `test_pipeline_integrity.py::...test_extract_agent_session_fields_includes_metadata` | yes | yes | pre-existing |
| `test_sdlc_dispatch.py::...test_missing_run_id_self_heals_on_free_lock[stage-marker]` | yes | yes | pre-existing |
| `test_stall_advisory_e2e.py::TestNeverStartedE2E::test_running_session_no_turn_start_classifies_stalled` | yes | yes | pre-existing |
| `test_sdlc_pipeline_lock.py::...test_write_meta_persists_plan_hash` | yes (`find_session` missing) | yes | pre-existing |
| `test_bot_loop_guard.py::test_non_bot_sender_unaffected` | no | no | full-suite pollution |
| `test_memory_mcp_server.py::test_fresh_shell_import_resolution` | yes (`ModuleNotFoundError: mcp`) | yes | pre-existing, env |
| `test_harness_resume.py::...test_stale_uuid_triggers_fallback` | yes | yes | pre-existing |

Nine of twelve produce **byte-identical tracebacks on main and on this branch**. The other three pass cleanly in isolation on both sides, so they are artifacts of the full-suite run sharing a machine with a parallel build lane's Redis, not failures at all.

The two tests closest to this branch's surface area, `test_remote_update.py` (exercises `scripts/update/*`) and `test_reflections_redis.py::TestRedisIndexCleanupReflection` (index/redis cleanup), fail identically on main. This branch neither introduced nor changed that failure mode.

Not checked: the remaining 33 failures were not individually run against main. The sample was selected to cover every plausibly-related area.

## Verification of the gates themselves

The two assertions this PR adds to protect against its own failure mode were **mutation-tested**, not just observed passing:

- Removing `args=("--apply",)` from `_migrate_schema_diet_fields_v2` -> `test_every_strip_registration_actually_passes_apply[schema_diet_fields_v2]` **FAILS**. Before this PR that same mutation left the suite fully green while the migration would run dry-run, exit 0, and be recorded complete having written nothing.
- One assertion is documented as *not* covering what an earlier draft claimed: `test_a_blinded_scan_does_no_index_work` pins the observable outcome, not the guard's early `return`, because deleting that return leaves it green (the sweep gate already skips on `stripped == 0`). The docstring says so rather than overclaiming.

---

## Patch cycle (post-review)

Review returned CHANGES REQUESTED (2 blockers), then APPROVE WITH COMMENTS twice as findings were cleared. Final head `f4ea3486f`. All blockers, both tech-debt items and all eleven review comments are addressed; nothing was deferred.

**Blockers.** `test_blank_lines_are_not_logged` was vacuous — records are formatted `[migration:probe] stdout: <line>`, so a logged blank line still yields a truthy `.strip()` and the assertion passed either way. Now asserts the record count, **mutation-verified**: deleting `if line.strip():` fails it with `assert 4 == 2`. And `docs/features/agent-session-fenced-execution-record.md` still described the pre-#2524 world and cited this PR's own issue as open future work; refreshed to the status quo.

**Tech debt, fixed rather than deferred.** `raw_field_names` swallowed `HKEYS` failures and returned an empty set, so a record whose detection failed was counted `clean`, `errors` stayed 0, the exit was 0, and `run_pending_migrations` recorded the migration **permanently complete**. A transient Redis blip would have manufactured the exact proof-of-cleanliness artifact these re-runs exist to produce. It now fails closed. The accepted consequence is documented symmetrically with the zero-record guard's, because it is not purely transient: `WRONGTYPE` on a phantom key (#2207, #2536) raises deterministically and pins the migration at exit 1 until an operator acts — the `logger.error` names the offending key.

Also: the safety narrative was duplicated between engine and pid delegate; it now lives in exactly one file, and the invariant is enforced across all three delegates rather than one.

## Final test evidence (supersedes the 97-passed figure above)

```
./scripts/pytest-clean.sh tests/unit/test_strip_migration_shared.py \
  tests/unit/test_migrate_strip_pid_fields.py \
  tests/unit/test_migrations.py tests/unit/test_sdlc_stubs.py -n0 -q
  -> 102 passed in 34.28s

./scripts/pytest-clean.sh tests/unit -q -x
  -> 2 failed, 6497 passed in 373.90s
```

Both unit failures (`test_pipeline_integrity.py::TestEnqueueContinuationFallback::test_extract_agent_session_fields_includes_metadata`, `test_sdlc_dispatch.py::TestRunIdRequiredFlag::test_missing_run_id_self_heals_on_free_lock[stage-marker]`) are in the set already baseline-verified against main earlier in this PR, with byte-identical tracebacks on both sides.

`ruff check` clean; `ruff format --check` clean. All three refactored scripts dry-run against live Redis at exit 0.

The single full-suite run was taken at `066553496`. Everything since is docstrings, comments, docs, and tests, plus the one runtime change (`raw_field_names` re-raising) covered by two new tests and the wide `tests/unit` run above.

## Cross-lane coordination (#2536)

The #2536 lane is adding a fail-closed popoto version-floor guard inside `AgentSession.repair_indexes()`, and its plan asserted `rebuild_indexes()` has exactly one in-repo caller. That is false: there are **nine** live call sites, verified on `origin/main`. Both plans' Freshness Checks wrongly claimed no overlapping active plan; both are corrected.

This PR removes exactly two of the nine (`migrate_strip_pty_fields.py:161`, `migrate_schema_diet_fields.py:230`), touches no other call site, does not modify `repair_indexes()`, and does not modify `models/agent_session.py` at all. **No file is edited by both lanes.** Details posted to [#2536](https://github.com/tomcounsell/ai/issues/2536#issuecomment-5200945295); #2544's census corrected to the full nine.

## Additional follow-up filed

**#2548** — `sdlc-tool verdict finalize` returns `ok: true` while its own readback reports `trailer_matches_head: false` and `marker_completed: false`. Found while recording this PR's REVIEW verdict. Fail-open shape in a gate tool; the verdict is recorded but the REVIEW stage marker did not land.

---

## Rebased onto latest main

Rebased onto `a85b7cc7c` (clean, no conflicts) so the final gate runs against the patched test system: pytest runs are now bounded by `--timeout=420 --timeout-method=thread` (`9232ef9b4`) and the machine-global suite lock is deleted (`3a91695b9`, #2535). Branch force-pushed; diff versus main is unchanged at 10 files.

Post-rebase focused gate: **102 passed**, ruff check and format clean, all three refactored scripts dry-run at exit 0.

**The full-suite run is pending a slot** allocated by the parent session, which is serializing suite runs across lanes now that the lock is gone.

**The rebase invalidated the recorded REVIEW verdict's head_sha trailer** (it pinned `f4ea3486f`). Combined with #2548 the REVIEW marker never persisted anyway, so the SDLC substrate needs reconciling before merge routing regardless.

## Live observation: a zero-record scan on a populated keyspace

While dry-running the three migrations back to back on this host, `migrate_schema_diet_fields` returned:

```
{'total_records': 0, 'clean': 0, 'stripped': 0, 'deferred_non_terminal': 0, 'errors': 0}
```

on a keyspace that reported 4006 records seconds earlier in the same loop, and 4006 again immediately after, and 4006 on eight further attempts. Never reproduced.

This is the exact ambiguity the zero-record guard exists to fail closed on. If it is what it appears to be, then on the **pre-#2524** `migrate_schema_diet_fields` that run would have exited 0 and been recorded permanently complete having stripped nothing — the #2518 canary failure mode, caught in the act. With this PR the same run exits 2 with an explicit refusal.

**The mechanism is corroborated.** The #2536 lane supplied a hypothesis and three falsification checks; all three came back pointing the same way.

`repair_indexes` runs on this host every ~31 minutes, each run logging `sessions_rebuilt=4006`. The last twelve today: 07:45:45, 08:16:43, 08:47:47, 09:18:44, 09:49:46, 10:20:43, 10:52:01, 11:23:03, 11:54:07, 12:28:18, 12:59:30, **13:31:36**. The zero-record scan ran after ~13:29:50; the same script returned 4006 at 13:31:54 and on eight further runs at 13:32:25-13:33:09. So it landed in the seconds before the 13:31:36 rebuild completion, and the next invocation eighteen seconds later was fine.

The checks: a rebuild was in flight; the migration reads `$Class:AgentSession` via `query.all()` (`_strip_migration.py:182`), the exact key popoto's rebuild deletes in Step 1 and re-adds in 1000-row batches in Step 2; and `popoto_index_cleanup.py:262` is ruled out because `_GUARDED_ELSEWHERE = frozenset({"AgentSession"})` excludes it deliberately as the #2207 fix.

I originally read the 8 clean retries as counter-evidence. That was wrong: they all ran inside a 45-second span between rebuilds, so they could not have hit a window that opens every ~31 minutes. The window is periodic, not persistent.

**The part that generalizes:** `repair_indexes` was measured at ~600ms for **150 sessions**. This host has **4006**. The window did not appear, it grew, and it keeps growing on machines that never prune AgentSession rows. So this guard is not insurance against a theoretical race. Without it, that 13:31 run would have exited 0 and been recorded permanently complete having stripped nothing.

The code comment carries this timeline, the header no longer claims the guard is "NOT a fix for a proven cause", and a test requires the #2549 evidence trail to stay reachable from the code that depends on it.

Precision caveat that remains: the zero-scan run's own timestamp and log line were not captured (its output went through a grep that kept only the stats line), so the overlap is a strong correlation across a bounded window, not a captured concurrent trace.

## Follow-ups filed by this lane

| Issue | Subject |
|---|---|
| #2543 | Zero-record guard fails closed forever on a genuinely empty keyspace |
| #2544 | Seven remaining raw `rebuild_indexes()` call sites (census corrected to nine total) |
| #2548 | `sdlc-tool verdict finalize` returns `ok: true` while its readback reports the trailer and marker did not persist |
| #2549 | Zero-record scan observed on a populated keyspace (candidate live #1720 window) |


valorengels:
## Review:

**Verdict: CHANGES REQUESTED** — 2 blockers, 2 tech debt, 4 nits.

The engineering here is strong. The consolidation is real (not relocated duplication), the pid migration is behavior-preserving line by line against `origin/main`, the reclaim-vs-accept decision is argued rather than asserted, and the two follow-ups (#2543, #2544) are actually filed and open. Both blockers are the same class of defect the PR itself set out to eliminate, applied to work the PR itself produced.

### Verified independently (not taken on the PR's word)

| Claim | Reproduced |
|---|---|
| `97 passed` on the focused paths | ✅ `97 passed in 120.31s` (`-n0`, this machine) |
| `ruff check` / `format --check` | ✅ All checks passed; 7 files already formatted |
| `grep -c rebuild_indexes` over engine + 3 delegates | ✅ `0, 0, 0, 0` |
| Guard text exists in exactly one file | ✅ only `scripts/_strip_migration.py` |
| Dry-run stats, both siblings | ✅ `{'total_records': 4006, 'clean': 4006, 'stripped': 0, 'deferred_non_terminal': 0, 'errors': 0}` — matches the PR body exactly |
| `strip_pid_fields` defers 44 non-terminal | ✅ `{'total_records': 4006, 'clean': 3962, 'stripped': 0, 'deferred_non_terminal': 44, 'errors': 0}` |
| `_v2` entries ordered before `purge_phantom_agent_sessions` | ✅ `migrations.py:1041`, `:1046` precede `:1051` |
| Plan has all four required sections | ✅ Documentation / Update System / Agent Integration / Test Impact |
| `docs/plans/durability-m1-fence-canary.md:431` row updated | ✅ already landed on `main` (plans commit to main), so it is correctly absent from this diff |

Behavior preservation of the pid migration was checked against `git show origin/main:scripts/migrate_strip_pid_fields.py`: scan order, `stats` key insertion order, the `delete`/`save`/`execute` pipeline sequence, the guard's early return, and the 2/1/0 exit ladder are identical. `field_names` being required (not defaulted) is the right call and is pinned by two tests.

### Blockers

- **`tests/unit/test_strip_migration_shared.py:482-486` — `test_blank_lines_are_not_logged` is vacuous, in the file whose own docstring warns against exactly this.** The runner emits `logger.info("[migration:%s] %s: %s", label, stream_name, line)` (`scripts/update/migrations.py:88-91`), so a suppressed-blank line, if it *were* logged, would produce the message `"[migration:probe] stdout: "` — whose `.strip()` is truthy. The assertion `[line for line in captured.splitlines() if not line.strip()] == []` therefore passes whether or not `if line.strip():` exists. I verified this empirically: with the suppression present the runner emits exactly 2 records for `print('a\n\n\nb')`; deleting the suppression yields 4 records and the assertion still returns `True`. This PR's stated deliverable includes "Three vacuous assertions fixed" and `TestSharedSubprocessRunner`'s docstring says these are "functional coverage — not source greps"; shipping a new vacuous assertion there contradicts the contract. Fix: assert on the record *count* (`len(recs) == 2`), not on the joined string.

- **`docs/features/agent-session-fenced-execution-record.md` still describes the pre-#2524 world and cites this PR's own issue as open future work.** Four stale statements, all verified against the branch:
  - `:172` — "This is scoped to this one migration ... Generalizing capture to every migration means widening that contract to carry `(error, output)` and updating every helper, tracked separately as [#2524]". This PR generalizes capture to six helpers and its own `_run_migration_script` docstring (`scripts/update/migrations.py:80-82`) says the opposite: "Widening the `MIGRATIONS` value contract turned out to be unnecessary."
  - `:172` — credits `_migrate_strip_pid_fields` with performing the capture; that helper is now a four-line delegate (`migrations.py:232-237`).
  - `:174` — "Bounding the retry would need new persisted state ... deliberately left unbounded." The engine replaced that rationale with a pointer to the deferred detection-only `SCAN` fix, now #2543. The doc keeps the retracted argument.
  - `:183` — "The Verification row for this greps `scripts/migrate_strip_pid_fields.py` ... so the script's own comment describes the excluded paths without naming them." That comment moved to `scripts/_strip_migration.py:230-242`, whose body was deliberately re-pointed to say "THIS FILE and the three delegate scripts". The doc's referent is stale in precisely the way the plan's own critique BLOCKER row flagged — fixed in the code comment, not in the doc describing it.

  Development Principle 1 is explicit: no historical artifacts in docs, describe only the new status quo. The plan's `## Documentation` section named `agent-session-model.md` (updated ✅) and the durability plan row (updated on main ✅) but missed this file.

### Tech Debt

- **`scripts/_strip_migration.py:106` — a detection failure is counted as `clean`, and the run is then recorded complete forever.** `raw_field_names` catches every exception, logs a warning, and returns an empty set; the caller at `:162-165` sees no stale fields and does `stats["clean"] += 1`. A record whose `HKEYS` read failed is indistinguishable from a genuinely clean one, `errors` stays 0, `strip_migration_main` returns 0 (`:296`), and `run_pending_migrations` (`migrations.py:1085-1088`) records the migration permanently complete. A transient Redis blip during the scan therefore manufactures the exact artifact Decision 1 says the re-run exists to produce ("a captured log line proving it is clean"). The zero-record guard fails closed; this path fails open. Pre-existing behavior carried forward from the pid script — but this PR is what extends it to two more migrations whose entire justification is audit proof, so it is worth counting the detection failures into the exit-1 ladder (or filing it alongside #2543).

- **`scripts/migrate_strip_pid_fields.py:19-54` duplicates `scripts/_strip_migration.py:28-61`.** The full "Safety properties" narrative (Idempotent / Atomicity-not-quiescence / Deferred-rows-do-not-age-out / TTL note) now exists in two near-verbatim copies. Both siblings correctly collapse it to a pointer (`migrate_strip_pty_fields.py:20-25`, `migrate_schema_diet_fields.py:76-81`), and the plan states the engine carries that commentary "in one place instead of three" — it is in two. The copies are free to drift, which is the failure mode this PR exists to eliminate, only now between engine and delegate. It is pinned there by `TestDocstringCorrections` (`tests/unit/test_migrate_strip_pid_fields.py:419-441`) asserting against `strip.__doc__`; repointing those four assertions at `shared.__doc__` would let the delegate keep only its field-set docstring like its siblings.

### Nits

- `tests/unit/test_strip_migration_shared.py:283-286` — `test_omitting_the_detection_function_is_an_error_not_a_fallback` sits inside `@pytest.mark.parametrize("module_name", STRIP_SCRIPTS)` and opens with `del module_name`, running the identical engine-level assertion three times. Its own docstring concedes the property "is the engine's". Move it to an unparametrized class.
- `tests/unit/test_strip_migration_shared.py:365` — the docstring rationale is wrong. "A v2 ahead of its v1 would run twice needlessly" cannot happen: `run_pending_migrations` skips by *name* (`migrations.py:1080-1082`) and the two names differ, so both run exactly once in either order. The assertion is harmless; the reason given for it is not the real one. The constraint that actually matters is the one at `:375`.
- `tests/unit/test_strip_migration_shared.py:343-350` — `test_each_v2_reruns_the_same_script_as_its_v1` scrapes the script name out of `v1_src` with a bare `next()` over lines starting `"migrate_` and ending `.py",`. If that formatting ever shifts the test dies with `StopIteration` rather than a readable assertion. A `re.search(r'"(migrate_\w+\.py)"', v1_src)` plus an explicit assert fails legibly.
- `scripts/update/migrations.py:932` — `_migrate_purge_phantom_agent_sessions` keeps `return f"exit code {result.returncode}: {result.stderr[-500:]}"`. Deliberately excluded (exit-3 contract, own time budget) and defensible, but it is now the last site with the stderr-only-empty-reason failure mode that `_run_migration_script`'s own docstring at `:80-83` argues against, and the module carries two error-string shapes. Worth a one-line comment naming it as a knowing exception, or routing it through the shared runner with an exit-code hook.

### Not blocking — checked and clean

- **Atomicity holds.** `POPOTO_REDIS_DB.pipeline()` defaults to `transaction=True`; `instance.delete(pipeline=pipe)` and `popoto.Model.save(instance, pipeline=pipe)` queue onto the same pipeline with one `EXEC` at `_strip_migration.py:196`. A crash or interleaved writer cannot lose a record — only a racing write to a terminal row is lost, which is documented at `:41-44`.
- **Terminal-only scope holds.** `TERMINAL_STATUSES` is checked at `:168` before any write, and the engine has no other write path.
- **Exit 2 is non-blocking with no escalating cost.** `run_pending_migrations` has no `break`; a failure appends to `failed`/`errors` and skips `_save_completed`, and `run.py` only logs `FAIL:`.
- **The error-string shape change is unobserved.** `grep -rn 'stderr\[-500'` returns only `migrations.py:106`, `:932`, and an unrelated file. Nothing parses `logs/update.log` for the old shape.
- **Real fleet cost is lower than the plan's "three passes per `/update`".** All three v1 entries plus `strip_pid_fields_v2` are already recorded complete, so the next `/update` runs two new passes once, then zero.
- No UI files in the diff — the visual-proof gate is a no-op. No new env vars. No `bridge/`, `agent/`, or `worker/` changes, so no service restart beyond the normal `/update`.

---

Both blockers are small, mechanical fixes. Fix the count assertion and refresh `agent-session-fenced-execution-record.md` to the post-#2524 status quo, and this is ready.

valorengels:
## Review:

**Verdict: CHANGES REQUESTED — Tech Debt** at head `65770a815`. 0 blockers, 2 tech debt, 1 nit.

**All four prior findings and all four prior nits are resolved.** Both blockers were re-verified with red-state proof, not taken on the PR's word. What remains is two factual inaccuracies in documentation this PR itself rewrote — the same class of defect the last round found, one layer out.

### Prior findings — re-verified at the current head

| Prior finding | Status | Evidence I produced |
|---|---|---|
| **BLOCKER** `test_blank_lines_are_not_logged` vacuous | ✅ Resolved | Now asserts `len(records) == 2` (`test_strip_migration_shared.py:589`). **Mutation-verified**: deleting `if line.strip():` at `migrations.py:101` makes it fail with `assert 4 == 2`, with all four records visible in the captured log. Reverted after. |
| **BLOCKER** stale `agent-session-fenced-execution-record.md` | ✅ Resolved | All four stale statements fixed: the `#2524`-as-open-future-work citation is replaced by a status-quo "Output capture" paragraph that names `_run_migration_script` and its one knowing exception; the `_migrate_strip_pid_fields` credit is gone; the retracted "bounding the retry needs new persisted state" rationale now points at #2543; the Verification-row sentence at `:178` is repointed at `_strip_migration.py` plus the three delegates and names the two enforcing tests. A "Shared strip engine" section and two Key-Files rows were added. |
| **TECH DEBT** `raw_field_names` fails open on `HKEYS` failure | ✅ Resolved | `scripts/_strip_migration.py:130` now `raise`s after the `logger.error`. The docstring documents the fail-closed choice **and** its accepted consequence symmetrically with the zero-record guard's — a `WRONGTYPE` phantom key (#2207/#2536) pins the migration at exit 1 indefinitely, named as expected output with the offending `redis_key` logged. That is the honest framing; the alternative was a silent "clean". |
| **TECH DEBT** safety narrative duplicated engine↔pid delegate | ✅ Resolved | `migrate_strip_pid_fields.py:15-20` is now a pointer, matching its two siblings. `TestDocstringCorrections` re-anchored on `shared.__doc__`, and `test_no_delegate_duplicates_the_narrative` enforces the one-copy invariant across **all three** delegates rather than only the pid one — wider than I asked for. |
| **NIT** parametrized engine-level assertion | ✅ Resolved | `test_omitting_the_detection_function_is_an_error_not_a_fallback` moved to module scope at `:338`, outside the `@parametrize` at `:289`. |
| **NIT** wrong docstring rationale at `:365` | ✅ Resolved | `test_the_v2_entries_run_after_their_v1` now says correctness does not depend on it and names the phantom-purge constraint as the one that bites. |
| **NIT** bare `next()` scrape → `StopIteration` | ✅ Resolved | `re.search(r'"(migrate_\w+\.py)"', v1_src)` plus an explicit `assert match` with a readable message (`:387-389`). |
| **NIT** `migrations.py:932` unnamed exception | ✅ Resolved | `_migrate_purge_phantom_agent_sessions` carries a `KNOWING EXCEPTION` paragraph naming the discarded success-path output, the stderr-only failure string, and the condition for adding a second helper in that shape. |

### Reproduced independently (Hard Rule 10 — not credited from the PR body)

| Claim | Reproduced |
|---|---|
| `102 passed` on the four focused files | ✅ `102 passed in 55.70s` (`-n0`, this machine) — see the Miscellaneous note below about a first attempt |
| Blank-line test is mutation-verified at `assert 4 == 2` | ✅ Exact string reproduced |
| `ruff check` / `ruff format --check` | ✅ All checks passed; 7 files already formatted |
| `grep -c rebuild_indexes` over engine + 3 delegates | ✅ `0, 0, 0, 0` (against `git show HEAD:`, not the working tree) |
| Guard text exists in exactly one file | ✅ `grep -rl "ZERO RECORDS SCANNED" scripts/` → `scripts/_strip_migration.py` only |
| `MIGRATIONS` contains both `_v2` names | ✅ `:1054`, `:1059` |
| Diff is 10 files vs `origin/main` | ✅ 10 files, +1348/-688 |
| Follow-ups #2543, #2544, #2548, #2549 | ✅ all four OPEN |
| Parent plan row updated on main | ✅ `durability-m1-fence-canary.md:437` carries the "Resolved by #2524" text |
| Rebase onto `a85b7cc7c` clean | ✅ `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN` |

### Blockers
- None

### Tech Debt

- **File:** `tests/README.md:344-345` (verified: read at `git show HEAD:tests/README.md`)
  **Code:** `` | unit | `test_migrate_strip_pid_fields.py` | 22 | ... `` and `` | unit | `test_strip_migration_shared.py` | 35 | ... ``
  **Issue:** Both test counts are wrong at this head. Actual collection: `test_migrate_strip_pid_fields.py` → **21**, `test_strip_migration_shared.py` → **56**. The shared-engine row is off by 21. This PR is what wrote both rows, and the index is the discovery surface for the suite — a count that understates a file by 60% is exactly the kind of stale artifact that makes the index untrustworthy.
  **Severity:** tech_debt
  **Fix:** `21` and `56`.

- **File:** `docs/features/agent-session-fenced-execution-record.md:169` (verified: read at `git show HEAD:`)
  **Code:** `Until that lands the retry is unbounded, and three migrations route through this guard, so a fresh install emits three recurring FAIL: lines.`
  **Issue:** Six registrations route through the guard, not three: `strip_pty_session_fields` (`migrations.py:993`), `schema_diet_fields` (`:1013`), `strip_pid_fields` (`:1017`), `strip_pid_fields_v2` (`:1027`), `strip_pty_session_fields_v2` (`:1054`), `schema_diet_fields_v2` (`:1059`). On a fresh install none is recorded complete, so all six run in the same `/update` and each emits its own `FAIL:` line. The sentence is this PR's own new text describing this PR's own new registrations, and it understates the operator-visible cost of the accepted #2543 consequence by half — which is the number an operator would use to decide whether #2543 is urgent.
  **Severity:** tech_debt
  **Fix:** Say six, and name that the v1 and v2 entries both run on a machine that has never run either.

### Nits
- None

### Miscellaneous

- **File:** `tests/unit/test_migrate_strip_pid_fields.py:278-293` and `:330-338` (verified: read, and observed failing)
  **Code:** `assert error is None, "this run must SUCCEED, or the test proves nothing"`
  **Issue:** `TestOutputIsCapturedNonEmpty` is timing-flaky under machine load. My **first** run of the exact command in the PR body gave `2 failed, 100 passed in 303.07s`; both failures were this class, both with the subprocess reporting `{'total_records': 0, ...}` and tripping the zero-record guard — on a test db where `_mk_row()` had just written a row. Re-running the class isolated gave `4 passed`, and re-running the full four-file command gave `102 passed in 55.70s`. The 303s vs 56s wall-clock gap says the first run was contended. Both failing tests are **unchanged from `origin/main`** by this diff, so this is pre-existing and not a regression this PR introduced — but it is worth naming twice over: it makes the PR's headline test evidence non-deterministic, and the observation itself is a second live sighting of the #2549 mechanism, this time inside the test keyspace rather than production. Worth attaching to #2549 as corroboration.
  **Severity:** nit
  **Fix:** Follow-up issue, not this PR. Either make the two tests retry the subprocess on a zero-record result, or have them assert against a directly-seeded scan rather than a live `query.all()`.

- **#2543 — zero-record guard fails closed forever on a genuinely empty keyspace** — tracked by #2543 (OPEN). Correctly scoped out: reversing a decision the parent issue's critique deliberated belongs in its own change.
- **#2544 — remaining raw `rebuild_indexes()` call sites** — tracked by #2544 (OPEN). Census corrected from five to nine in the PR body; different pattern, plausibly load-bearing rather than defensive, so `clean_indexes()` is not a blanket substitute.
- **#2548 — `sdlc-tool verdict finalize` fail-open readback** — tracked by #2548 (OPEN). Found while recording this PR's own verdict; genuinely out of scope here.
- **#2549 — zero-record scan on a populated keyspace** — tracked by #2549 (OPEN). The Miscellaneous item above corroborates it independently.
- **`_migrate_purge_phantom_agent_sessions` excluded from `_run_migration_script`** — explicit out-of-scope exclusion with sound rationale (exit-3 branch, own time budget), and since the last round it is named in code as a knowing exception with the condition for revisiting.

### Review Delta (vs prior review on HEAD `f4ea348`)
- **Resolved**: both blockers (vacuous blank-line assertion; stale fenced-execution-record doc), both tech-debt items (`HKEYS` fail-open; duplicated safety narrative), and all four nits. Each verified at the current head against code I read, with red-state proof on the blank-line assertion.
- **New**: `tests/README.md` test counts (22/35 vs actual 21/56); `agent-session-fenced-execution-record.md:169` says three guard-routed migrations where there are six. Both are in text this PR authored after the prior review, so neither existed to be found last round.
- **New (nit)**: `TestOutputIsCapturedNonEmpty` timing flakiness — pre-existing on `main`, surfaced because I ran the focused command under contention rather than accepting the PR's number.
- **Unchanged**: nothing. The prior finding set is fully cleared.

## Rubric

- [x] **1. Plan vs. implementation match** — pass — *All six Success Criteria verified at head: zero `rebuild_indexes` across engine + 3 delegates, all three fail closed via the shared engine, guard text in exactly one file, both `_v2` names registered, focused tests pass, pid migration behavior-preserving (verified line by line last round, untouched since).*
- [x] **2. New code quality** — pass — *The fail-closed change is three lines and documents its own accepted consequence. Delegates are uniformly thin. Type hints present, `field_names` correctly keyword-only and required.*
- [x] **3. Test coverage** — pass — *56 tests on the new engine. The two assertions guarding this PR's own failure mode are mutation-tested, and one is explicitly documented as NOT covering what an earlier draft claimed — that self-correction is worth more than the assertion.*
- [x] **4. Regression risk to existing callers** — pass — *Six subprocess-shaped helpers routed through `_run_migration_script`; the `MIGRATIONS` `str | None` contract is untouched. The one helper left out is named in code.*
- [x] **5. Data integrity** — pass — *Terminal-only scope checked before any write; delete+save on one transactional pipeline; `clean_indexes()` never a raw rebuild. Fail-closed detection strictly reduces the chance of a false "clean".*
- [x] **6. Security** — pass — *No secrets, no new eval/exec, no request-derived subprocess input. The `--apply` subprocess in tests is pinned to a non-db-0 URL with an explicit refusal assertion.*
- [ ] **7. Documentation accuracy** — fail — *Two inaccuracies in text this PR wrote: `tests/README.md` counts and the "three recurring FAIL: lines" claim. Both are the two Tech Debt findings.*
- [x] **8. PR body accuracy** — pass — *Every number I checked reproduced: 102 passed, `assert 4 == 2`, 10 files, `0,0,0,0`, four open follow-ups. The 45-failure full-suite classification table is from a pre-rebase run and is not re-verified here, but nothing in it is load-bearing for this verdict.*
- [x] **9. Disclosed deferrals** — pass — *All five disclosures are either tracked by an OPEN issue or are explicit exclusions with rationale that survives scrutiny.*
- [x] **10. Follow-up claims verified** — pass — *#2543, #2544, #2548, #2549 all OPEN.*

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — PASS — *All six re-run at this head, not carried over.*
- **2. No-Gos from plan — none violated** — PASS — *Scope stayed on the strip family; the five rename-style migrations were routed to #2544 rather than absorbed.*
- **3. New `except Exception` blocks — each has logger/raise/swallow-ok** — PASS — *`_strip_migration.py:128-130` logs then re-raises; that is the fix from last round.*
- **4. New integration tests — exercise serialization boundary (not in-memory only)** — PASS — *`TestSharedSubprocessRunner` spawns real subprocesses against throwaway scripts; the pid file's capture tests hit live Redis.*
- **5. Plan internal consistency — spike findings match task steps** — PASS — *Freshness Check self-corrected on the #2536 overlap and the correction is posted on both sides.*
- **7. New public APIs — docstrings present** — PASS — *`run_strip_migration`, `raw_field_names`, `strip_migration_main`, `_run_migration_script` all documented.*
- **8. Breaking changes — migration path documented** — PASS — *The rename-to-rerun mechanism is the migration path and is documented in code, docs, and tests.*
- **9. Tests added for new behavior** — PASS — *Fail-closed detection has dedicated coverage; the `--apply` pinning assertion is mutation-verified.*
- **10. Tests cover the failure path (not just happy path)** — PASS — *Zero-record guard, per-record errors, malformed stats dict, sweep failure, blinded scan.*
- **11. UI changes (if any) — screenshot captured** — N/A — *No UI files in the diff; the visual-proof gate is a no-op.*
- **12. Docs updated for user-facing changes** — FAIL — *Updated, but with two factual errors. Promoted to the two Tech Debt findings.*

### Verification Results

The plan has no machine-readable `## Verification` table — its gates live in `## Success Criteria`, and all six were run above.

### Screenshots

---

**Note on working-tree state.** While this review ran, the lane worktree acquired uncommitted edits to both files named in Tech Debt above, and those edits appear to fix both findings. They are **not pushed** — `65770a815` is the head under review and both errors are present in it. Push them and this is ready; the verdict flips on one commit.

valorengels:
## Review: PR Closed

This PR is no longer open (`state=MERGED`, merge commit `679b68f6e`, merged 2026-08-06T08:49:48Z); the formal review was skipped per the mergeability preflight. No `gh pr review` verdict was posted, and the recorded SDLC verdict was left untouched.

A delta re-review was dispatched at head `f9b42d7b3` to verify the two outstanding tech-debt findings from the `65770a815` pass. It arrived after the merge, so it is recorded here as forensics rather than as a gate.

### Delta verification (both findings resolved, verified against code — not the commit message)

**1. `tests/README.md` test counts — RESOLVED.** The doc previously claimed 22 / 35. Collected on `origin/main`:

- `tests/unit/test_migrate_strip_pid_fields.py` -> **21 collected**; doc now says 21.
- `tests/unit/test_strip_migration_shared.py` -> **56 collected**; doc now says 56.

Both match exactly.

**2. `docs/features/agent-session-fenced-execution-record.md` guard-registration count — RESOLVED.** The doc previously said "three migrations route through this guard, so a fresh install emits three recurring `FAIL:` lines". It now reads "Six registrations route through this guard (`strip_pty_session_fields`, `schema_diet_fields`, `strip_pid_fields`, and all three `_v2` names) ... so a fresh install emits six recurring `FAIL:` lines."

Confirmed against `scripts/update/migrations.py` on `origin/main` — six registrations at lines **993, 1013, 1017, 1027, 1054, 1059**, exactly the cited set.

### Notes

- `f9b42d7b3` is not an ancestor of `origin/main` (squash merge), but its content is present: both doc corrections are verified in the merged tree.
- All eight findings from the pass before `65770a815` remain confirmed resolved and were not re-litigated.
- Nit carried forward, non-blocking: the `TestOutputIsCapturedNonEmpty` flake is a second sighting of the [#2549](https://github.com/tomcounsell/ai/issues/2549) mechanism.
- Substrate note: the recorded `REVIEW` verdict for this issue is `CHANGES REQUESTED`, timestamped 08:50:38Z — **50 seconds after** the merge landed and after `MERGE` was already marked `completed`. The verdict record and the merge are out of order in the ledger. That is consistent with the [#2548](https://github.com/tomcounsell/ai/issues/2548) fail-open shape and is worth folding into that issue's evidence.

Net: nothing outstanding on the merged content. Both doc-accuracy errors are genuinely fixed in `main`.

valorengels:
## Review: PR Closed (merged mid-review)

This review was requested against head `65770a815` while the PR was `OPEN`/`MERGEABLE`/`CLEAN`. It merged during the pass (`679b68f6e`, 2026-08-06T08:49:48Z), and head advanced to `f9b42d7b3`. Posting as a comment rather than a formal review, since a merged PR cannot take one. Everything below is a **post-merge** observation against `main`.

**Verdict on the merged code: no blockers.** Two verified tech-debt items and three nits, all follow-up material. The one finding this pass raised against `65770a815` was already fixed by `f9b42d7b3` before merge.

### What was verified independently (not taken on the PR's word)

| Claim | Reproduced here |
|---|---|
| `102 passed` on the focused paths | `102 passed in 36.67s` (`-n0`) |
| `ruff check` / `ruff format --check` | All checks passed; 20 files already formatted |
| `grep -c rebuild_indexes` over engine + 3 delegates | `0, 0, 0, 0` |
| Full-suite gate, **which the PR body disclosed as "pending a slot" post-rebase** | Closed it: `tests/unit` on `65770a815` -> **6 failed, 11474 passed** in 28:50 |
| #2543 / #2544 / #2548 / #2549 / #2536 all OPEN | Confirmed, all five |

On the six unit failures: two (`test_pipeline_integrity.py::...test_extract_agent_session_fields_includes_metadata`, `test_sdlc_dispatch.py::...test_missing_run_id_self_heals_on_free_lock[stage-marker]`) are the ones already baseline-verified in this PR. The other four are all `test_persona_substitution.py`, which I re-ran isolated (`-n0`) on **both** `main` and the branch — **7 passed on each**. Full-suite pollution, not attributable to this branch. Nothing in the diff's surface area fails.

### Resolved before merge

The doc/code contradiction this pass found at `docs/features/agent-session-fenced-execution-record.md:169` — the paragraph still called the zero-record guard "not a fix for a proven cause; nothing establishes that window actually fired on any machine" while `scripts/_strip_migration.py:238-244` in the same PR documented the window observed firing on a 4006-row keyspace (#2549) — was fixed by `f9b42d7b3`. The doc now states the observation and the scaling argument. No action needed.

### Tech Debt (verified post-merge, on `main`)

- **`tests/unit/test_strip_migration_shared.py:335`** — `assert stats["total_records"] == 0`
  `test_the_guard_fires_for_it_too` is vacuous. Its docstring claims "Per-script proof, not just engine-level: the wiring must actually reach it", but with `AgentSession.query` patched to an empty iterator, `total_records == 0` holds for *any* implementation returning a stats dict — including a delegate that cloned the old guard-less loop. Deleting `logger.error(ZERO_RECORDS_MESSAGE)` and the early `return` leaves it green for all three parametrizations. Contrast the engine-level `test_the_guard_fires_and_says_why` (`:81`), which does assert the message.
  **Fix:** capture on `mod.logger.name` and require `shared.ZERO_RECORDS_MESSAGE` among the records — that proves the delegate's wiring reaches the guard *and* that the message is attributable to the right migration in `logs/update.log`.

- **`tests/unit/test_migrate_strip_pid_fields.py:417`** — `assert "The worker never writes terminal rows" not in (shared.__doc__ or "")`
  The retracted claim never existed in that capitalization. `git show 984d3bb7f:scripts/migrate_strip_pty_fields.py:24` reads `worker never writes terminal rows` — lowercase, mid-sentence — and `git log -S` confirms only that form ever existed. A verbatim reintroduction passes this assertion. This PR touched the line (retargeting `strip.__doc__` -> `shared.__doc__`) without noticing the string never matched.
  **Fix:** `assert "never writes terminal rows" not in (shared.__doc__ or "").lower()`. The companion positive assertions are fine as-is.

Both are vacuous-assertion defects in a PR whose stated deliverable included "Three vacuous assertions fixed" — the same class, one layer out. Worth a fast-follow.

### Nits

- **`tests/unit/test_migrate_strip_pid_fields.py:117`** — the docstring claims the guard returns *before* the sweep, a property the test cannot pin (with zero records `stripped` is 0, so the `if apply and stats["stripped"]:` gate at `_strip_migration.py:268` already skips it). The new file's sibling test documents exactly this caveat as "HONEST SCOPE ... verified by mutation" — the two files now say opposite things about the same mechanism, and the older one is wrong.
- **`scripts/update/migrations.py:109`** — `except subprocess.TimeoutExpired:` discards everything the script printed before it hung. `TimeoutExpired` carries partial `stdout`/`stderr`; neither is logged, and the returned reason carries no content. That is the same argument this function's own docstring makes for the success path. Bind the exception and run the same logging loop over its tails.
- **`scripts/_strip_migration.py:150`** — "wraps ``raw_field_names`` below" is directionally wrong; `raw_field_names` is defined at `:88`, above `run_strip_migration` at `:134`.
- **`scripts/update/migrations.py:1045`** — the "both must stay AHEAD of `purge_phantom_agent_sessions`" comment reads as a fleet-wide guarantee, but `purge_phantom_agent_sessions` is already recorded complete on existing machines, so it is skipped there and the ordering can only bind on a fresh install (where it is correct). Narrow the comment.

### Operational note worth its own follow-up

`scripts/update/run.py:1113` appends each failed migration to `result.errors`, which sets `result.success = False` (`:2155`), renders the run as `FAILED with N error(s)`, and sends a Telegram `update failed at {sha}`. Six registrations now route through the fail-closed zero-record guard, so a **fresh install reports `/update` as failed forever** and the `/update` success signal stops being usable as an alert source on that machine. #2543 is the right home for this; it is currently framed as a guard-precision issue rather than an alerting one.

### Risk posture — checked and clean

- `TERMINAL_STATUSES` (`models/session_lifecycle.py:65`) is `{"completed","failed","killed","abandoned","cancelled"}` — no in-flight status. Non-terminal rows take the `deferred_non_terminal` branch and are never written.
- Redis commands buffer client-side until `pipe.execute()`, so an exception between `delete()` and `save()` leaves Redis untouched. Record loss is not reachable.
- `STALE_FIELDS` is a **detection trigger only** — the rewrite serializes whatever the model currently declares. Checked anyway: zero intersection between all three sets (8/15/4 names) and the 89 live `*Field(` declarations on `AgentSession`.
- The fail-closed `raw_field_names` change has no escape path: `field_names(instance)` is called at `_strip_migration.py:185`, inside the `try` whose handler is at `:221`, so a re-raised `HKEYS` failure lands in `errors` -> exit 1 -> never recorded complete.
- Ordering holds: both `_v2` entries precede `purge_phantom_agent_sessions` in `MIGRATIONS`. `run_pending_migrations` has no `break`/`raise`, so one failing migration does not abort the rest.
- `_run_migration_script` is full parity against baseline `a85b7cc7c`: timeouts preserved exactly (120/120/120, 300/300/300), args preserved, `script.exists()` guard preserved, timeout string character-identical. The only exit-code branch that ever existed is `purge`'s exit-3, correctly left bespoke and documented.
- No UI files — visual-proof gate is a no-op. No new env vars. No `bridge/`/`agent/`/`worker/` changes.

### Method

Two-judge consensus (code-quality, risk) per this repo's REVIEW config, both awaited in-turn, aggregated under any-blocker-wins. Both returned APPROVED. No verdict was recorded to the SDLC substrate: `sdlc-tool stage-query --issue-number 2524` shows the pipeline already `completed` through MERGE, and writing a REVIEW verdict onto a closed pipeline would regress its state for no routing benefit.

PR #2565, opened 2026-08-06
Prune orphaned pre-manifest hook files under ~/.claude/hooks/ (#2521)

### LINKED ISSUE #2521: Prune orphaned pre-manifest hook files under ~/.claude/hooks/

valorengels:
Closes #2521

Plan: `docs/plans/prune-orphaned-hook-files.md`

## What

Adds `_migrate_prune_orphaned_premanifest_hook_files()` to `scripts/update/migrations.py`, registered in `MIGRATIONS` as `prune_orphaned_premanifest_hook_files`. `/update` Step 3.6 runs it automatically; no operator action.

The manifest generator hardlinks only the manifest's `scope = "global"` declarations, so it is structurally unable to remove anything the old hardcoded `_SDLC_HOOK_DEFS` path left in `~/.claude/hooks/`. #2503 (PR #2522) removed the last orphaned *registration*, making those files inert but deleting nothing. This removes the files.

**Three gates, all of which must permit a removal:**
1. named in `_PREMANIFEST_HOOK_ORPHANS` — a static, hand-maintained inventory. Unrecognized ⇒ kept, always.
2. not in `_USER_HOOK_KEEP_PATHS`.
3. not referenced by any command string in `~/.claude/settings.json` (permissive on absent/unreadable/malformed, one shared code path, WARNING logged, never aborts).

Every gate keys on `str(path.relative_to(hooks_root))`, never `path.name`: top-level `sdlc_reminder.py` is residue while `sdlc/sdlc_reminder.py` is a live global script, and the two share a basename. `sdlc/sdlc_context.py` is in the keep-set explicitly — it carries no manifest declaration and appears in no command string, but all three global fork hooks import it at runtime.

A full-tree `.bak` snapshot is taken only when the removal set is non-empty and before any unlink. Idempotency comes from disk state, not the ledger.

## Deviation from the plan — a blocking safety finding

**`~/.claude/hooks` on this machine is a SYMLINK to `<repo>/.claude/hooks`** (created Feb 3 2026, the legacy dir-symlink layout `hardlinks.py:219-227` migrates away from for `~/.claude/skills`). `/update` does not create it, but it exists.

On such a machine the "user tree" and this repo's git-tracked source tree are the *same directory*, so every inventory path resolves to a live repo file. An unguarded run during this build deleted 36 tracked repo files plus their bytecode (restored from git + the migration's own `.bak`).

This invalidates the plan's **spike-1**: the "shared inode 5917700" it measured between the user path and the repo path was this symlink, not a hardlink, so its "unlinking leaves the repo source intact" conclusion does not hold on this layout.

The migration now returns immediately (WARNING logged, no writes, no `.bak`) when `~/.claude/hooks` is a symlink or resolves to `<project_dir>/.claude/hooks`. Verified against the real machine: `result: None`, tree byte-identical, no `.bak`. Covered by `test_legacy_dir_symlink_into_repo_is_never_pruned`.

**Consequence for reviewers:** on a machine in this layout the prune is a permanent no-op, so the issue's stated outcome is not achieved there. Whether that layout should itself be migrated (as `~/.claude/skills` was) is a separate question worth filing.

## Verification

| Check | Result |
|---|---|
| `scripts/pytest-clean.sh tests/unit/test_hook_migration.py -n0 -q` | 36 passed |
| `tests/unit/test_migrations.py` | 10 passed |
| `ruff check` | All checks passed |
| `ruff format --check` | 2 files already formatted |
| Migration registered in `MIGRATIONS` | True |
| `sdlc/sdlc_context.py` in keep-set | 2 hits |
| Foreign `/opt/my-own-tool/guard.py` fixture intact | 4 hits, untouched |
| No AST import walk | pass |
| Not routed through `RENAMED_REMOVALS` | pass |
| No basename comparison in the prune | pass (2 hits are pre-existing `settings_path.name` backup-filename construction in older migrations, verified identical on `main`) |
| Live-tree cross-check | inventory covers the real tree exactly; 74 files removed, 6 `sdlc/` survivors, no live path caught |

## Docs

`docs/features/hook-manifest.md` gains a **User-tree Hygiene** section covering what the generator does and does not own, the three gates, the `sdlc/sdlc_context.py` import edge, and a **legacy dir-symlink** subsection. Docs gate: exit 2 (non-blocking stale-marker warning on the word "legacy" in the new prose).

### DIFF
--- a/docs/features/hook-manifest.md
+++ b/docs/features/hook-manifest.md
@@ -185,7 +185,31 @@ Exactly one such entry was found deployed: `validators/validate_no_raw_redis_del
 
 Condition 3 is an **explicit precise-match allowlist**, deliberately not a `validators/` or `~/.claude/hooks/` prefix test. A prefix gate would silently deregister a hand-added user hook that happens to live under `~/.claude/hooks/`; only the exact scripts named in the tuple are ever removed. Blocks and events emptied by the sweep are pruned.
 
-The migration is **registration-only: it deletes no files.** Once the registration is gone the orphaned pre-manifest files under `~/.claude/hooks/` are inert; pruning them is deferred to issue #2521, which must use a static keep-by-default allowlist rather than a negative reference test. Like its sibling it writes a timestamped `.bak` snapshot first, asserts JSON validity before touching disk, and is idempotent — an absent file, an empty `{}`, a missing `hooks` key, or nothing left to sweep are all no-ops that write nothing.
+The migration is **registration-only: it deletes no files.** Once the registration is gone the orphaned pre-manifest files under `~/.claude/hooks/` are inert; removing them from disk is the separate sibling migration described in [User-tree Hygiene](#user-tree-hygiene-claudehooks) below. Like its sibling it writes a timestamped `.bak` snapshot first, asserts JSON validity before touching disk, and is idempotent — an absent file, an empty `{}`, a missing `hooks` key, or nothing left to sweep are all no-ops that write nothing.
+
+## User-tree Hygiene (`~/.claude/hooks/`)
+
+The user generator owns a **narrow slice** of `~/.claude/hooks/`: it hardlinks exactly the manifest's `scope = "global"` declarations (today three scripts, all under `sdlc/`) and it registers exactly those in `~/.claude/settings.json`. It does **not** own, enumerate, or clean anything else in that directory. Everything the pre-manifest hardcoded `_SDLC_HOOK_DEFS` path put there — `validators/`, `dispatch/`, `hook_utils/`, `__pycache__/`, nine top-level scripts, and a stale copy of `manifest.toml` — was therefore permanent residue on every fleet machine, inert after the registration sweep above but never removed.
+
+`scripts/update/migrations.py::_migrate_prune_orphaned_premanifest_hook_files()` (issue #2521) is the one-time self-heal. A path is removed only when **all three** gates permit it:
+
+1. it is named in `_PREMANIFEST_HOOK_ORPHANS`, the static hand-maintained inventory of the pre-manifest layout;
+2. it is not in `_USER_HOOK_KEEP_PATHS`; and
+3. it is not referenced by any command string in `~/.claude/settings.json`.
+
+Gate 1 makes **unrecognized ⇒ kept** the default: a hand-added user hook, a foreign repo's artifact, or any path the inventory has never heard of is never touched. Naming a *tree* in the inventory does not license deleting its whole contents — each contained file is recognized individually (it must have a counterpart under this repo's `.claude/hooks/`, or be regenerable `.pyc` bytecode) and still passes gates 2 and 3, so a tree holding a hand-added file survives, holding just that file. Every gate compares the path **relative to `~/.claude/hooks/`**, never a bare basename: top-level `sdlc_reminder.py` is removable residue while `sdlc/sdlc_reminder.py` is a live manifest-declared global script, and the two share a basename.
+
+Gate 3 is deliberately **permissive** on an absent, unreadable, or malformed `settings.json` — one shared code path that logs a WARNING and proceeds with no registration guard. It is defense in depth, not the primary gate; a machine with a hand-broken settings.json must still get its residue pruned rather than being wedged forever on a file this migration does not own.
+
+`sdlc/sdlc_context.py` is in the keep-set and is the reason the keep-set exists as an explicit constant rather than being derived from command strings. **It carries no manifest declaration of its own** and appears in no hook command anywhere — it is a pure runtime import sibling, imported by all three global fork scripts (`sdlc/validate_commit_message_sdlc.py:35`, `sdlc/validate_sdlc_on_stop.py:34`, `sdlc/sdlc_reminder.py:29`). Any future keep-set, sweep, or generator change touching this directory must account for it explicitly; a command-string-derived keep-set deletes it and breaks all three global hooks in every foreign repo. (`sync_user_hooks()` likewise does not hardlink import siblings, so a genuinely fresh machine has a latent import failure — tracked separately as #2561.)
+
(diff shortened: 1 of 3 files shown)


valorengels:
Closing unmerged. The issue this implements (#2521) is closed as premise-disproved — there is no orphaned pre-manifest residue to prune, because `~/.claude/hooks` on this machine class is a symlink to the repo's own `.claude/hooks/`.

The migration here is correct code for a problem that does not exist. Shipping it would add a fleet-wide deletion path whose only safe behavior on this machine class is to do nothing.

The symlink guard is the salvageable piece and is carried into the follow-up issue verbatim, along with the 36-file near-miss as motivating evidence.

Issue #2563, opened 2026-08-06
_extract_agent_session_fields serves two incompatible contracts: recreate-same-session loses 50 fields, continuation would forge an execution fence


valorengels:
## Problem

`_extract_agent_session_fields` (`agent/agent_session_queue.py:214-221`) serves four production call sites that split into two semantically incompatible groups. There is no single field classification that is correct for both, so the helper is currently wrong for one group no matter what its field list contains.

**Group 1 — delete-and-recreate of the SAME session.** The session's identity is preserved; every field absent from `_AGENT_SESSION_FIELDS` is destroyed.

- `agent/session_health.py:4583-4587` — orphan repair, clearing the `parent_agent_session_id` KeyField.
- `tools/agent_session_scheduler.py:884-889` — priority bump.

Correct behavior here is to copy everything, **including** the fenced-execution-record fields, because the recreated row describes the same live process.

**Group 2 — creating a NEW session.** A distinct session with new identity.

- `agent/agent_session_queue.py:784` — `retry_agent_session` on a failed/cancelled session.
- `agent/session_executor.py:586` — auto-continue fallback recreate.

Correct behavior here is to copy history but **reset** the execution-fence fields. Copying `exec_pid`, `pid_create_time`, `exec_cwd`, `exec_harness`, `spawn_history`, `active_run_id`, `worker_pid` onto a new session forges a fence describing a process that never ran for it — the exact forged-liveness failure the #2518 durability fence work exists to prevent.

## Current impact (live bug, not hypothetical)

`_AGENT_SESSION_FIELDS` is a hand-maintained 38-entry list. Measured against the live model on `a85b7cc7c`, **50 model fields are missing from it**. For Group 1 that is silent data loss on every orphan repair and every priority bump. The lost fields include `issue_number`, `pr_number`, `model`, `task_type`, `project_config`, `retain_for_resume`, `budget_tripped` / `budget_tripped_reason`, `crash_signature`, the full `total_*_tokens` and `total_cost_usd` accounting, and the full `thread_*` counter set.

For Group 2 the same list is accidentally *protective* — it omits the fence fields, so continuations do not currently forge a fence. Naively "fixing the drift" by adding the missing 50 to the list would close the Group 1 data loss and simultaneously introduce forged fences in Group 2. That is why this cannot be fixed as a list edit.

## Evidence

- `agent/agent_session_queue.py:172-211` — the 38-entry hand-maintained list.
[...]

Issue #2518, opened 2026-08-04
Post-merge canary + permanent regression tests for the fenced execution record (Durability M1, PR #2516)


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/durability-m1-fence-canary.md

## Recon Summary

Ran 2026-08-04 against `3a5f1b50` (post-merge). Five parallel investigations plus direct inspection of the live canary machine (Tom's MacBook Air). **The canary has already run** — `/update` applied `strip_pid_fields` here at 11:48 UTC, ~90min after PR #2516 merged — so this issue's work is "fix what it found, re-verify, then roll" rather than first-contact validation.

> **Correction (2026-08-04).** An earlier version of this summary led with a HIGH-severity claim that the migration "self-certified" and had "stripped nothing". That claim was wrong and is retracted; D1 below carries the corrected, evidence-backed account and is downgraded to LOW. **The substance of this recon is the fence, not the migration** — four HIGH-severity defects (D2-D4, D8) where PR #2516 introduced `fence_is_live` and then left six consumers reading the fenced pid without the `create_time` compare. Those are unaffected by the correction.

**Confirmed:** 4 items — ready to include

- **Fence IS stamped on the real production spawn path.** `AgentSession.stamp_execution_spawn` (`models/agent_session.py:1173`) is called from `SessionRunner._on_turn_spawn` (`agent/session_runner/runner.py:633`, stamp at `:665-677`), wired via `_build_driver` `on_spawn=` (`:521`) → `role_driver._dispatch_turn_event` TURN_SPAWNED (`:407`) → `harness/claude.py:1042` firing immediately after `asyncio.create_subprocess_exec`. Not a side branch.
- **Staleness-by-compare confirmed.** The fence is re-stamped every turn and never nulled — the only writer of `exec_pid` in production is `stamp_execution_spawn:1202`; no `exec_pid = None` exists anywhere. Deliberate non-clear on finalize is documented at `models/session_lifecycle.py:484-491`.
- **Migration is terminal-only, ORM-safe, atomic.** `TERMINAL_STATUSES` gate at `scripts/migrate_strip_pid_fields.py:118`; delete+recreate is one MULTI/EXEC pipeline (`:140-148`) so there is **no non-existence window**; zero raw-Redis writes; `updated_at` correctly preserved. Registered at `scripts/update/migrations.py:983`, completion recorded per-migration (no batch-loss).
[...]

Issue #2473, opened 2026-07-30
Stale data/catchup-disabled kill switch silently disabled all message recovery for 7 days


valorengels:
## What happened

`data/catchup-disabled` (0-byte operator kill switch, mtime 2026-07-22 13:45) remained on disk for 7 days after the bug it was waiting on was fixed. It has now been removed, but the window it created needs review.

The flag gates **all three** recovery scans:
- `bridge/reconciler.py:102` -> `catchup_disabled()` -> logs warning, `return 0`
- `bridge/catchup.py:73` -> `scan_for_missed_messages` returns 0
- `agent_catchup.py:794` -> `valor-catchup`

It was set pending #2204. #2204 was CLOSED 2026-07-23T08:52:28Z by PR #2298. The issue body's own instruction was "Remove the flag (`rm data/catchup-disabled`) once fixed." Nobody did.

## Impact

Between 2026-07-25 and 2026-07-30 there was a sustained `bridge_update_loop_wedged` crash storm (36 events in 72h on this machine). Throughout that storm the app-level missed-message recovery **did not run**. The only backstop that executed was Telethon's native `catch_up()` (channel-update replay), which does not apply the dedup/handled-check recovery logic.

Consequence worth noting: the reconciler hotfix `5d9515671`, authored 2026-07-30, was **dead code on arrival** — `catchup_disabled()` short-circuits `reconcile_once()` before the new per-chat cutoff logic is reached.

## Status

Flag removed 2026-07-30. `catchup_disabled()` verified returning `False`. Check is `Path.exists()` at call time, so no restart was required.

## Work to do here

**Do not just close this as "flag removed."** The questions worth answering:

1. **Quantify the loss.** Diff Telegram history against dispatched sessions across 07-22 -> 07-30 to determine whether messages were dropped and never recovered. This was explicitly NOT verified during the audit. If people went unanswered, that matters more than the flag.
2. **Was re-enabling safe?** Verified before removal: #2204's root cause was dedup TTL (2h) < scan window (24h). Dedup record TTL is now 2592000s / 30 days (`bridge/dedup.py:122`), coupled to the cursor TTL, against a 24h scan cap (`bridge/catchup.py:87`). Ordering is correct with a wide margin. **Re-validate this independently rather than trusting the summary.**
[...]

PR #2568, opened 2026-08-06
Repair the test suite's measuring instrument: four false-signal defects (#2552, #2553, #2556, #2557, #2523)

### LINKED ISSUE #2523: skills-audit FAIL: do-skills-audit (rule 19)
The `skills-audit` reflection observed a FAIL finding on **2 consecutive runs** for skill `do-skills-audit` (rule 19) in `valor`.

**Message:** Husk directory: no SKILL.md (empty) — delete or restore

This issue was auto-filed by the `skills-audit` reflection. It will NOT be re-filed for 30 days even if the finding persists. Close this issue to silence; the streak counter will reset naturally when the underlying rule passes.

### LINKED ISSUE #2552: Catchup kill switch (data/catchup-disabled) leaks production control state into the test suite: 19 nodes red from main checkout, green from worktrees
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/wave2-test-instrument.md

## Root cause

`bridge/catchup.py:40` resolves the `data/catchup-disabled` kill-switch path relative to the source file's parent, not the working directory. The flag file has existed since Jul 22 (0 bytes, gitignored), so any pytest run started from the main checkout sees catchup as disabled and 19 catchup/reconciler/dedup nodes go red. The same nodes pass from a worktree, because the worktree has no `data/catchup-disabled`.

This is production control state leaking into the test suite through a gitignored file. Nothing in the repo records it, so the failure is invisible to anyone reading the diff.

## Evidence

- `bridge/catchup.py:40` — flag path anchored to the module's parent directory.
- Flag present on this host since 2026-07-22, 0 bytes.
- Main checkout: 19 catchup/reconciler/dedup nodes red. Worktree: same nodes green.

## Fix shape

1. Autouse fixture that points `CATCHUP_DISABLED_FLAG` at `tmp_path` so no test ever reads the real flag.
2. Decide the production flag's fate separately. It is operator-gated state, tracked in #2473.
3. Audit for other readers that treat a gitignored file as control state. Any of them has the same test-contamination shape.

## Verification

Must include a negative control: remove the fixture and confirm the 19 nodes go red again. A green suite alone does not prove the fixture is what is doing the work.

## Supersedes

- #2532 Cluster B
- #2429 Group A
- #2488 root-cause B

## Recon Summary

Re-verified by direct measurement on host `Valor the Pirate` at main `a85b7cc7c`, serial `-n0`, before any code change.

**Confirmed:**
- `bridge/catchup.py:40` still anchors `CATCHUP_DISABLED_FLAG` to `Path(__file__).resolve().parent.parent / "data" / "catchup-disabled"`. Working-directory independent, source-tree relative.
- `data/catchup-disabled` present, 0 bytes, mtime Jul 22 13:45.
- Two-arm measurement over 7 focused files (`tests/unit/test_agent_catchup.py`, `test_reconciler.py`, `test_catchup_claim.py`, `test_dedup.py`, `test_pipeline_integrity.py`, `test_reflections_yaml_migration.py`, `test_update_hardlinks.py`):
  - Arm 1, production flag present: **18 failed / 120 passed**.
  - Arm 2, identical selection with `CATCHUP_DISABLED_FLAG` monkeypatched to a temp path via a throwaway plugin (production flag untouched on disk): **1 failed / 137 passed**.
- The flag alone therefore causes exactly **17** red nodes, and is the sole cause of each: 2 in `test_agent_catchup`, 13 in `test_reconciler`, 2 in `test_catchup_claim`. The single arm-2 survivor is the unrelated #2553 stale assertion.

**Revised:**
- The issue says 19 nodes. The measured unit-level count is 17. The remaining 2 are plausibly in the integration catchup files (`tests/integration/test_reconciler.py`, `test_agent_catchup_recovery.py`, `test_catchup_revival.py`, `test_per_chat_catchup_cutoff.py`, `test_update_loop_wedge_recovery.py`), which were not in the two-arm selection. The fixture is autouse and conftest-scoped, so it covers those regardless of the exact count; the count is not load-bearing for the fix.

**Confirmed readers of the kill switch** (the audit item in the issue's fix shape) — three, all reached through `bridge.catchup`:
- `bridge/catchup.py:73`
- `bridge/reconciler.py:182-186` (imports from `bridge.catchup`)
- `bridge/agent_catchup.py:792-795` (imports from `bridge.catchup`)

All three import the symbol from `bridge.catchup` rather than re-deriving the path, so a single autouse fixture patching `bridge.catchup.CATCHUP_DISABLED_FLAG` covers every reader. No fourth gitignored-file-as-control-state reader was found in the non-worktree tree.

**Pre-requisites:** None.

**Dropped:** Nothing. The production flag stays on disk, operator-gated per #2473.

Recon by the Wave 2 build lane, host `Valor the Pirate`, slug `wave2-test-instrument`.

### LINKED ISSUE #2553: AgentSession field round-trip: test asserts a removed 'expectations' field, and _AGENT_SESSION_FIELDS silently drops 6 fields added by df6097fe6
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/wave2-test-instrument.md

## Root cause

Two related defects in the AgentSession field round-trip.

**Stale assertion.** `tests/unit/test_pipeline_integrity.py:176` asserts the presence of a field named `"expectations"`. That field has zero occurrences in `agent/agent_session_queue.py`. The test is asserting against a contract that no longer exists.

**No drift detection.** `_AGENT_SESSION_FIELDS` has no mechanism to detect divergence from the actual model. Six fields added by `df6097fe6` are absent from the list and are therefore silently dropped on session recreate. Nothing fails when the model grows a field, which is how the six went unnoticed.

The stale assertion is cosmetic. The silent field drop is the real bug: recreating a session loses state with no error.

## Evidence

- `tests/unit/test_pipeline_integrity.py:176` — asserts `"expectations"`.
- `agent/agent_session_queue.py` — zero occurrences of `expectations`.
- `df6097fe6` — added 6 fields not reflected in `_AGENT_SESSION_FIELDS`.

## Fix shape

Drop the stale assertion, add the 6 missing fields, and add a drift test that derives the expected field set from the model rather than restating it. A hand-maintained list will drift again.

## Supersedes

#2532 Cluster A.

## Recon Summary

Re-verified by direct measurement on host `Valor the Pirate` at main `a85b7cc7c`. **This recon materially corrects the issue's stated fix shape — do not implement the issue as written.**

**Confirmed:**
- `tests/unit/test_pipeline_integrity.py:176` asserts `"expectations"` is in `_AGENT_SESSION_FIELDS`. It fails today: `AssertionError: Critical field 'expectations' missing from _AGENT_SESSION_FIELDS` at line 184. Verified flag-independent (it is the sole failure surviving the #2552 kill-switch redirect).
- `df6097fe6` ("Durability M1: fenced execution record + pid-field deletion (#2494 Tasks 6-8)") deleted the `expectations` field, which is why the assertion is stale. That commit added exactly six model fields: `last_authored_at`, `exec_pid`, `pid_create_time`, `exec_cwd`, `exec_harness`, `spawn_history`.
- `_AGENT_SESSION_FIELDS` is a hand-maintained list at `agent/agent_session_queue.py:172-211` (38 entries) consumed by `_extract_agent_session_fields` at :214-221, which does a blind `getattr` over the list.

**Revised:** — two corrections, both load-bearing

1. **The drift is 51 fields, not 6.** Measured against the live model:
   `set(AgentSession._meta.fields) - {"id"} - set(_AGENT_SESSION_FIELDS)` yields 51 names. The 6 from `df6097fe6` are a small subset; the other 45 predate that commit. Any fix scoped to "add the 6" leaves 45 fields still silently dropped.

2. **Blindly adding the named 6 would introduce a real bug.** Five of them (`exec_pid`, `pid_create_time`, `exec_cwd`, `exec_harness`, `spawn_history`) are the fenced-execution-record fields introduced by the #2518 durability work. Copying them onto a recreated session would attach a fence describing a process that never ran for that session — precisely the forged-liveness failure mode the fence exists to prevent. The correct disposition for these is *explicit exclusion*, not inclusion.

**Corrected fix shape.** Keep the copy list explicit (a blind derive-everything would copy the fence fields and reintroduce the bug above). Add a named EXCLUDED set carrying a one-line rationale per entry. Make the drift guard derive the model's field set at runtime and fail when any field appears in neither COPY nor EXCLUDED — so the guard catches the next field addition regardless of which bucket it belongs in, and forces a deliberate classification rather than a silent list edit.

**Blast radius of `_extract_agent_session_fields`** — 7 non-test call sites and consumers to keep working: `agent/agent_session_queue.py:784`, `agent/session_executor.py:586`, `agent/session_health.py:4583-4586`, `tools/agent_session_scheduler.py:884-886`. Existing tests asserting on it: `tests/unit/test_agent_session_queue.py`, `test_session_completion_zombie.py`, `tests/integration/test_agent_session_queue_race.py`, `test_agent_session_scheduler.py`, `test_agent_session_health_monitor.py`.

**Pre-requisites:** None.

**Dropped:** The issue's literal instruction "add the 6 missing fields" — superseded by the corrected fix shape above, for the forged-fence reason.

Recon by the Wave 2 build lane, host `Valor the Pirate`, slug `wave2-test-instrument`.

### LINKED ISSUE #2556: Reflections cutover tests assert on YAML comments that reflection_machine_filter's safe_load/safe_dump round-trip destroys (latent red, not a stable green)
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/wave2-test-instrument.md

## Root cause

`TestSentryTriageCutover` and `TestPrReviewAuditCutover` assert on pointer comments inside `config/reflections.yaml`. YAML comments are not survivable across the derived-artifact rewrite:

- `tools/reflection_machine_filter.py:110` reads with `safe_load`, which discards comments.
- `tools/reflection_machine_filter.py:146` writes with `safe_dump`, which cannot restore them.

The tests pass on this host right now only because the current `config/reflections.yaml` (mtime Jul 29) still carries both pointer comments. The next time the machine filter runs, the comments are destroyed and both tests go red. This is a latent red, not a stable green.

## Evidence

- `tools/reflection_machine_filter.py:110` — `safe_load`.
- `tools/reflection_machine_filter.py:146` — `safe_dump`.
- `config/reflections.yaml` mtime Jul 29, both pointer comments present.

## Fix shape

Assert on a structural marker that survives the round-trip (a real YAML key), not on a comment. Alternatively use a comment-preserving round-trip, but the cheaper and more durable fix is to stop encoding contract state in comments.

## Supersedes

#2430 item 2.

## Recon Summary

Re-verified by direct measurement on host `Valor the Pirate` at main `a85b7cc7c`.

**Confirmed:**
- `tools/reflection_machine_filter.py:110` uses `yaml.safe_load`; `:146` uses `yaml.safe_dump`. Neither preserves comments — this is a property of PyYAML's plain loader/dumper, not a bug in the call sites.
- `tests/unit/test_reflections_yaml_migration.py::TestSentryTriageCutover::test_sentry_issue_triage_absent_from_repo_registry` asserts `"sentry-issue-triage migrated to a Claude Code Routine" in registry_path.read_text()` — a raw substring match against a YAML *comment*. `TestPrReviewAuditCutover` carries the structurally identical assertion.
- Both tests **pass today**: the whole file is green in the baseline run. That is the issue's point — this is a latent red, not a stable green. The comments survive only because `config/reflections.yaml` (gitignored, machine-local, mtime Jul 29) has not been rewritten by the filter since they were added.

**Confirmed:** the tests are already correctly gated Both classes sit behind `@pytest.mark.skipif(_..._cutover_pending())`, so on machines without the file or pre-cutover they skip cleanly. The defect is narrowly the comment-substring assertion inside the non-skipped path, not the gating.

**Design ruling — take the issue's recommended option, not its alternative.** The issue offers two paths: (a) assert on a structural marker that survives the round-trip, or (b) make the filter comment-preserving. Path (a) is the recommendation ("the cheaper and more durable fix is to stop encoding contract state in comments") and is what this plan implements. Path (b) would mean swapping PyYAML for `ruamel.yaml` across a derived-artifact writer to protect a test assertion — a new runtime dependency and a rewrite of the filter's I/O to serve a test's convenience, inverting the cost. Rejected.

The durable signal already exists and needs no new key: the real contract is *absence of the reflection entry from the registry*, which the preceding assertion (`assert "sentry-issue-triage" not in names`) already checks against parsed YAML structure. The comment assertion adds no contract coverage the structural one lacks — it only adds fragility.

**Pre-requisites:** None.

**Dropped:** Making `reflection_machine_filter` comment-preserving (issue's alternative path).

Recon by the Wave 2 build lane, host `Valor the Pirate`, slug `wave2-test-instrument`.

### LINKED ISSUE #2557: Hardlinks staleness probe fooled by __pycache__-only husks: .is_dir() reads a dead skill directory as present
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/wave2-test-instrument.md

## Root cause

`tests/unit/test_update_hardlinks.py:412-413` probes for a live skill directory with `.is_dir()`. A directory whose source files have all been deleted but which still contains a `__pycache__` subdirectory passes that probe. `.claude/skills-global/do-skills-audit` is exactly that: a `__pycache__`-only husk that the staleness probe reads as a present, healthy skill.

The probe therefore cannot detect the class of staleness it exists to detect.

## Evidence

- `tests/unit/test_update_hardlinks.py:412-413` — `.is_dir()` probe.
- `.claude/skills-global/do-skills-audit` — survives as a `__pycache__`-only husk.

## Fix shape

Probe for the artifact that actually defines a skill (`SKILL.md`), not for directory existence. Then sweep the existing husks so the corrected probe starts from a clean state.

## Supersedes

Parts of #2430. Related: #2523.

## Recon Summary

Re-verified by direct inspection on host `Valor the Pirate` at main `a85b7cc7c`. **Recon widens the fix: the issue cites the less important of three probes, and the husk is untracked, which changes what "sweep" has to mean.**

**Confirmed:**
- `tests/unit/test_update_hardlinks.py:413-414` probes `.is_dir()` for skill liveness in `test_renamed_removals_entries_are_not_stale` (the issue cites 412-413; the actual predicate lines are 413-414 — one-line drift, claim holds).
- `.claude/skills-global/do-skills-audit/` is a husk: contents are exactly `references/metadata.json` plus two `.pyc` files under `scripts/__pycache__/`. No `SKILL.md`. It is not hardlinked into `~/.claude/skills/` (`ls` returns no such file).
- History explains it: `44026cb96` renamed `do-skills-audit` → `audit-skills`, `ad4e95c8d` pruned stale skills-global skills. `("skills", "do-skills-audit")` is present in `RENAMED_REMOVALS` at `scripts/update/hardlinks.py:159`. The live successor `.claude/skills-global/audit-skills/SKILL.md` exists. So the rename completed everywhere except the source-tree directory itself.

**Revised:** — three findings that change the fix

1. **There is a third `.is_dir()` probe, at line 25, and it is the one that actually produces a false pass.** `_SKILL_ROOTS` liveness is `any((_REPO_ROOT / root / name).is_dir() for root in _SKILL_ROOTS)`, consumed by `test_renamed_removals_covers_deleted_skills`. That test excuses a deleted skill from needing a `RENAMED_REMOVALS` entry if it "currently exists on disk in any skill root". A husk makes that read `True`, so a genuinely deleted skill that left a husk behind silently escapes the completeness check. The two probes the issue cites (413-414) can only produce a false *failure*, not a false pass. Line 25 must be fixed too, and is the higher-value fix.

2. **The husk is entirely untracked** — `git ls-files .claude/skills-global/do-skills-audit/` returns nothing. `references/metadata.json` is untracked and the `.pyc` files are gitignored. Consequence: deleting it locally produces **no diff**, so a "sweep" committed in this PR would fix exactly one machine and be invisible to review. The durable artifact must therefore be a *guard test* that fails while any husk exists in either skill root, plus the local sweep. The guard is what carries the fix to other machines; the sweep is what makes it green here.

3. **`.claude/skills/_shared/` must be allowlisted, not swept.** It holds `test-quality.md` (git-tracked) and no `SKILL.md`. It is an intentional shared-resource directory whose `_`-prefix marks it as not-a-skill. A naive "every dir without SKILL.md is a husk" sweep would delete tracked content.

**Relationship to #2523.** #2523 is the `skills-audit` reflection reporting rule 19 FAIL on `do-skills-audit`: "Husk directory: no SKILL.md (empty) — delete or restore". Same directory, same root cause. Sweeping the husk resolves it, so this PR carries `Closes #2523`.

**Pre-requisites:** None.

**Dropped:** Nothing.

Recon by the Wave 2 build lane, host `Valor the Pirate`, slug `wave2-test-instrument`.

valorengels:
## Summary

Repairs four independent defects that break the relationship between "the suite is green" and "the code is correct". Two produced false reds, two produced false greens. Every guard added here ships with a negative control proving it can fail.

Plan: `docs/plans/wave2-test-instrument.md`

## What changed

**#2552 — catchup kill-switch isolation.** An autouse `isolate_catchup_kill_switch` fixture in `tests/conftest.py` repoints `bridge.catchup.CATCHUP_DISABLED_FLAG` at a per-test tmp path. All three production readers import that one symbol, so patching it covers every reader. The flag is *repointed, not stubbed* — `catchup_disabled()` still runs its real `Path.exists()`, so a test that wants disabled behavior can `touch` the redirected path. `CATCHUP_FLAG_ISOLATION=0` is a documented escape hatch that keeps the negative control permanently reproducible.

No file under `bridge/` was changed. The source-tree anchor is correct for production (the bridge runs under launchd with an unpredictable cwd); the defect is that tests read operator control state, and the fix belongs in the tests.

**#2553 — field-contract drift guard.** Deleted the stale `"expectations"` assertion (`df6097fe6` removed that field). Added `TestAgentSessionFieldContractDrift` with two guards: a phantom-name guard (every name in `_AGENT_SESSION_FIELDS` exists on the model) and a `KNOWN_GAP` set-*equality* guard recording the 50 model fields the list does not carry, with a pointer to #2563 as its termination condition. `_AGENT_SESSION_FIELDS` itself is untouched — the guard is added, the list is not edited.

The left-hand set excludes `AutoKeyField`s programmatically (`isinstance(f, AutoKeyField)`), never by subtracting a literal `{"id"}`, so a second auto-key field added later cannot silently reopen the same off-by-one. The 50 names were generated at build time from the exclusion-filtered diff, not transcribed.

**#2556 — durable cutover signal.** Both comment-substring assertions are gone; `tools/reflection_machine_filter.py`'s `safe_load`/`safe_dump` round-trip destroys YAML comments, so those tests were a latent red masquerading as a stable green. The structural assertions (against parsed YAML) remain and now name the contract they protect. A new `TestFilterRoundTripSignalDurability` regression test runs the filter over a synthetic `tmp_path` registry seeded with an other-machine `project_key` (required, or `filter_reflections_for_machine` no-ops and the control proves nothing) and asserts, against the same round-tripped artifact: the write happened, the comment is gone, the structural signal survived. The real `config/reflections.yaml` is never read or written.

**#2557 / #2523 — skill liveness probe.** All three `.is_dir()` liveness probes become `SKILL.md` presence. The uncited one at `:25` is the consequential one: it let `test_renamed_removals_covers_deleted_skills` excuse a deleted skill from needing a `RENAMED_REMOVALS` entry whenever a husk kept the directory readable — a false *pass*. A new `test_no_husk_directories_in_skill_roots` fails while any directory in either skill root lacks `SKILL.md`, gated by an explicit `HUSK_GUARD_ALLOWLIST = frozenset({"_shared"})` rather than an underscore-prefix rule.

## Acceptance evidence

Baseline (main checkout, serial, 7 focused files): **18 failed / 120 passed**. Post-fix: **141 passed, 2 skipped, 0 failed**.

| # | Criterion | Result |
|---|---|---|
| 1 | 17 flag-caused failures green | 141 passed, 2 skipped, 0 failed |
| 2 | **Negative control #2552** — fixture disabled, flag present | 17 failed, distribution 2/13/2 across `test_agent_catchup`/`test_reconciler`/`test_catchup_claim`, matching the baseline node list exactly. Reproduced independently by the orchestrator. |
| 3 | `grep -c '"expectations"' tests/unit/test_pipeline_integrity.py` | 0 |
| 4 | **Negative control** — bogus name appended in-memory | `AssertionError: _AGENT_SESSION_FIELDS names field(s) that no longer exist on AgentSession: ['bogus_phantom_field_xyz']` |
| 5 | **Negative control** — one name removed from `KNOWN_GAP` | `AssertionError: AgentSession field coverage drifted. Unclassified field(s) ...: ['worker_pid']` |
| 6, 7, 7a | **Negative + positive control #2556** — comment destroyed, structure survives, write confirmed | `TestFilterRoundTripSignalDurability` passes; all three assertions against the same round-tripped artifact |
| 8 | **Negative control** — temp husk dir | `AssertionError: husk directories found in the skill roots ...: ['.claude/skills/zz-orchestrator-husk-check']`. Reproduced independently by the orchestrator. |
| 9 | Husk swept | `.claude/skills-global/do-skills-audit` absent in both checkouts |
| 10 | `_shared` not swept | `.claude/skills/_shared/test-quality.md` present |
| 11 | All three former `.is_dir()` liveness sites route through `_skill_is_live` (a `SKILL.md` file check) | 3/3; no liveness decision reads directory existence |
| 12 | **Anti-criterion** — `git diff --name-only origin/main -- bridge/` | empty |
| 13 | **Anti-criterion** — `git diff origin/main -- agent/agent_session_queue.py` | empty |
| 14 | **Anti-criterion** — `data/catchup-disabled` | present, 0 bytes, mtime unchanged (Jul 22 13:45) |
| 15 | `ruff check` / `ruff format --check` | clean |

Measurement confirming the plan's arithmetic: `AgentSession._meta.fields` = 89 entries, 88 non-`AutoKeyField`, `_AGENT_SESSION_FIELDS` = 38, phantom names = 0, `KNOWN_GAP` = 50.

## Reviewer notes

- **The husk sweep produces no diff.** `.claude/skills-global/do-skills-audit/` is fully untracked, so its deletion is invisible here. It was a host-local operation on the main checkout (it never materialized in the worktree). The reviewable, durable artifact is the husk guard, which fails on any machine still carrying a husk. Verify with `test -e .claude/skills-global/do-skills-audit`.
- **Success Criterion 11 was corrected rather than satisfied.** The first build spelled `_find_husk_dirs` with `os.path.isdir` purely so a whole-file `grep -c 'is_dir()'` would keep returning zero. That is code bending to a bad measurement, in a PR about measurements that lie. Those calls filter directory entries and are not liveness probes at all, so the criterion was wrong, not the code. `e90a554d7` restored the natural `Path.is_dir` spelling and the criterion was amended on `main` (`6cd27be71`) to assert what matters: that all three former liveness sites route through a single `_skill_is_live` helper. The file now contains no `os.path` call at all.
- **Two skips in the focused run** are the existing cutover tests, which skip in a worktree because `config/reflections.yaml` is machine-local and gitignored. Their strengthened structural assertions are therefore not exercised from a worktree; the new `tmp_path` test is what carries real coverage, which is why it was added.
- Full-suite verification of the autouse fixture's blast radius is deliberately deferred — this host's full-suite slot is occupied by a diagnostic run. No test in the tree references `CATCHUP_DISABLED_FLAG` or `catchup_disabled` (verified by grep), so the fixture has no test that wants the flag set.

## Out of scope

- **#2563** — the `_extract_agent_session_fields` two-contract conflict and closing the 50-field gap. Production semantics change inside the #2518 fence blast radius; five of the six fields the original issue named are fenced-execution-record fields, and copying them into a newly created session would forge a fence describing a process that never ran.
- **#2473** — the fate of the production `data/catchup-disabled` flag. Operator-gated state.

Closes #2552
Closes #2553
Closes #2556
Closes #2557
Closes #2523

## Post-review patch (both review findings addressed)

Review verdict was Changes Requested — Tech Debt, no blockers. Both items are fixed.

**1. PR body accuracy.** Acceptance row 11 and the `os.path.isdir` reviewer note described code that `e90a554d7` had already removed. Both are corrected above.

**2. Quadratic cost in the autouse fixture** (`7f663d646`). The reviewer measured that `tmp_path_factory.mktemp` is numbered, so it scans the basetemp for the next free suffix and its cost grows with the directory count — 0.15s at 500 directories, 3.34s at 3000. Calling it once per test from a suite-wide autouse fixture made the aggregate quadratic and left one empty directory per test. The directory is now created once at session scope and only the *filename* varies per test, which costs a string format and no filesystem write.

Isolation is unchanged and re-verified after the refactor: each test still gets a path no other test can touch, `catchup_disabled()` still runs a real `Path.exists()`, the focused acceptance run is still **141 passed / 2 skipped / 0 failed**, and the `CATCHUP_FLAG_ISOLATION=0` negative control still reproduces exactly **17 failed / 40 passed** in the 2/13/2 distribution. Ruff clean. The production `data/catchup-disabled` remains 0 bytes at its original Jul 22 13:45 mtime.

### Exact commands behind every number in this PR

Listed so a reviewer reproduces the same thing rather than inferring it. All run from the worktree `.worktrees/wave2-test-instrument` at head `7f663d646`.

Focused acceptance (**141 passed, 2 skipped, 0 failed**):

```
./scripts/pytest-clean.sh -n0 -q --no-header -p no:randomly \
  tests/unit/test_agent_catchup.py tests/unit/test_reconciler.py \
  tests/unit/test_catchup_claim.py tests/unit/test_dedup.py \
  tests/unit/test_pipeline_integrity.py tests/unit/test_reflections_yaml_migration.py \
  tests/unit/test_update_hardlinks.py
```

#2552 negative control (**17 failed, 40 passed**, distribution 2 / 13 / 2). Requires the kill-switch flag to exist in the checkout the run starts from; `data/catchup-disabled` is gitignored, so create it in the worktree and remove it afterwards. Do **not** touch the main checkout's copy — it is operator state tracked in #2473.

```
touch data/catchup-disabled
CATCHUP_FLAG_ISOLATION=0 ./scripts/pytest-clean.sh -n0 -q --no-header -p no:randomly \
  tests/unit/test_agent_catchup.py tests/unit/test_reconciler.py tests/unit/test_catchup_claim.py
rm -f data/catchup-disabled
```

Field-drift arithmetic (89 total / 88 non-auto / 38 listed / `KNOWN_GAP` 50 / 0 phantoms):

```
.venv/bin/python -c "
from popoto import AutoKeyField
from models.agent_session import AgentSession
import agent.agent_session_queue as q
m = AgentSession._meta
nonauto = {n for n, f in m.fields.items() if not isinstance(f, AutoKeyField)}
print(len(m.fields), len(nonauto), len(q._AGENT_SESSION_FIELDS),
      len(nonauto - set(q._AGENT_SESSION_FIELDS)),
      sorted(set(q._AGENT_SESSION_FIELDS) - set(m.fields)))
"
```

Lint: `.venv/bin/python -m ruff check .` and `.venv/bin/python -m ruff format --check .`

### Full-suite status: attempted, wedged, NOT claimed as evidence

A full `tests/unit/` run at this head was ordered and attempted. It reached 99%, printed `[gw5] node down: Not properly terminated`, and stalled with the controller alive at 0% CPU and all xdist workers gone. It never reached the `-rf` summary, so there is no node list and no comparison against the canonical 35-failure baseline in `/tmp/unit-main-baseline.log`.

Recording it here as an open item rather than omitting it: **this PR does not currently carry full-suite evidence.** Memory was 91% free and no crash report was produced, so the worker did not die of resource exhaustion, but the diff ships a suite-wide autouse fixture and that is exactly the blast radius a full run exists to probe — so the wedge is not being attributed to anything until a second run says so.

### DIFF
--- a/docs/features/skills-global.md
+++ b/docs/features/skills-global.md
@@ -71,6 +71,32 @@ stat -f %i .claude/skills-global/<name>/SKILL.md ~/.claude/skills/<name>/SKILL.m
 
 Differing inodes mean the live copy is stale; `/update` re-establishes every link.
 
+## Skill Liveness and Husks
+
+A skill is *live* if and only if its directory holds a `SKILL.md` file. Directory
+existence alone is not liveness — a directory can survive on disk after its `SKILL.md`
+is deleted (a `__pycache__` dir, a stray reference file, an incomplete rename), and a
+bare `Path.is_dir()` probe reads that leftover as a live skill. Such a directory is a
+**husk**.
+
+`tests/unit/test_update_hardlinks.py::test_no_husk_directories_in_skill_roots` fails
+the build whenever a husk exists under `.claude/skills-global/` or `.claude/skills/`.
+All liveness checks in that file (`_skill_exists_in_any_root`, `test_renamed_removals_entries_are_not_stale`)
+route through the same `_skill_is_live` helper, so a directory-existence check cannot
+silently reappear at a second call site.
+
+The one directory allowed to lack a `SKILL.md` is `.claude/skills/_shared/` — an
+intentional shared-resource directory (it tracks `test-quality.md`, consumed by other
+skills rather than being one itself). It is permitted via an explicit
+`HUSK_GUARD_ALLOWLIST` frozenset in `test_update_hardlinks.py`, not an underscore-prefix
+convention: an allowlist of one names exactly what's exempt and fails loudly the moment
+a second non-skill directory shows up, instead of a naming convention silently absorbing
+anything that happens to start with `_`.
+
+When a skill is renamed or removed, delete its directory entirely (not just `SKILL.md`)
+and add a `RENAMED_REMOVALS` entry — see "Global vs. Project-Only Skills" in the root
+`CLAUDE.md`. Leaving the emptied directory behind is exactly what creates a husk.
+
 ## Invocation Types
 
 - **User + Model**: Both user and agent can trigger via `/skill-name`
--- a/docs/plans/wave2-test-instrument.md
+++ b/docs/plans/wave2-test-instrument.md
@@ -250,9 +250,9 @@ No agent integration required. Every change is in test code, test fixtures, or a
 
 ## Documentation
 
-- [ ] Update `tests/README.md` — add the catchup kill-switch isolation fixture to the fixture/blind-spot notes, recording that operator flag files are neutralized suite-wide and why (a test reading a gitignored production flag is a contamination class, not a one-off).
-- [ ] Update `docs/features/skill-context-convention.md` or the nearest skills-infrastructure doc — record that skill liveness is defined by `SKILL.md` presence, not directory existence, and that `_`-prefixed directories under the skill roots are shared resources rather than skills.
-- [ ] No new feature doc. This plan repairs existing behavior rather than adding capability; a `docs/features/` entry would describe a test fixture, which belongs in `tests/README.md`.
+- [x] Update `tests/README.md` — added the `isolate_catchup_kill_switch` fixture to the Fixtures table and a new "Operator control-state isolation (issue #2552)" subsection recording the neutralization, the `CATCHUP_FLAG_ISOLATION=0` escape hatch, and the contamination-class lesson.
+- [x] Update `docs/features/skills-global.md` (nearest skills-infrastructure doc, not `skill-context-convention.md` — that doc covers repo-specific skill behavior layering, not liveness) — added a "Skill Liveness and Husks" section recording that liveness is `SKILL.md` presence, that `.claude/skills/_shared/` is exempted via the explicit `HUSK_GUARD_ALLOWLIST` frozenset rather than an underscore-prefix convention, and the husk-guard test that enforces it.
+- [x] No new feature doc. This plan repairs existing behavior rather than adding capability; a `docs/features/` entry would describe a test fixture, which belongs in `tests/README.md`.
 
 ## Success Criteria
 
--- a/tests/README.md
+++ b/tests/README.md
@@ -73,6 +73,32 @@ Fix at source, not by suppression: the three #2118 leaks (`run_email_bridge`,
 via `bridge/promise_gate.py::_run_async_safely`, `_worker_loop` in
 `test_slow_redis_no_loop_freeze.py`) were each closed where the coroutine was dropped.
 
+### Operator control-state isolation (issue #2552)
+
+`bridge/catchup.py` anchors `CATCHUP_DISABLED_FLAG` to its own source tree (not to a
+fixture-controlled path), which is correct for production but means the suite reads
+live operator control state by default. On a host where an operator has paused message
+recovery via `data/catchup-disabled`, 17 unit tests with no relation to catchup went red
+purely because the file existed on that machine and not on another — a false red with no
+explanation anywhere in the diff.
+
+The autouse `isolate_catchup_kill_switch` fixture (`tests/conftest.py`) repoints
+`bridge.catchup.CATCHUP_DISABLED_FLAG` at a per-test tmp path for the duration of every
+test, covering all three production readers (`bridge/catchup.py`, `bridge/reconciler.py`,
+`bridge/agent_catchup.py`) since they all import the symbol from `bridge.catchup` rather
+than re-deriving the path. The flag is *repointed*, not stubbed — `catchup_disabled()`
+still runs a real `Path.exists()` against a real filesystem path, so a test that wants
+genuine disabled behavior can `touch` the redirected path.
+
+- **Escape hatch:** set `CATCHUP_FLAG_ISOLATION=0` to skip the redirect. This exists so
+  the negative control for #2552 stays permanently reproducible — running the suite with
+  it set must reproduce the same 17 flag-caused failures on a host where the real
+  `data/catchup-disabled` exists.
+- **The general lesson:** a test reading a gitignored production flag file is a
+  contamination *class*, not a one-off. Any future operator-toggled flag under `data/`
+  needs the same treatment before it can silently flip unrelated tests red depending on
+  which machine or checkout ran them.
+
 ### Known-failing clusters resolved on `main` (issue #1578)
 
 The previously known-bad clusters on `main` were driven to green in #1578. The fixes were **test-only** — assertions were re-pointed to current source/templates, never weakened, and no test was deleted:
@@ -389,6 +415,7 @@ here.
 |---------|-------|--------|---------|
 | `mock_claude_sdk_cleanup` | autouse | `conftest.py` | SDK mock cleanup between tests |
 | `redis_test_db` | autouse | `conftest.py` | Per-worker Redis db isolation |
+| `isolate_catchup_kill_switch` | autouse | `conftest.py` | Repoints the operator catchup kill-switch flag at a per-test tmp path (issue #2552) |
 | `sample_config` | function | `conftest.py` | 3-project sample configuration |
 | `valor_project` | function | `conftest.py` | Single project config |
 | `mock_telegram_client` | function | `tests/e2e/conftest.py` | AsyncMock Telethon client |
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -219,6 +219,53 @@ def mock_claude_sdk_cleanup():
             del sys.modules[mod_key]
 
 
+# ---------------------------------------------------------------------------
+# Operator kill-switch isolation (#2552)
+# ---------------------------------------------------------------------------
+# `bridge/catchup.py` anchors CATCHUP_DISABLED_FLAG to its own source tree
+# (`Path(__file__).parent.parent / "data" / "catchup-disabled"`), which is
+# correct for production — the bridge runs under launchd with an unpredictable
+# cwd — but means the test suite reads live operator control state. On a host
+# where an operator has paused message recovery, ~17 unit tests that have
+# nothing to do with the kill switch go red, and the same nodes pass from a
+# worktree. That is a false red with no explanation anywhere in the diff.
+#
+# All three production readers (`bridge/catchup.py`, `bridge/reconciler.py`,
+# `bridge/agent_catchup.py`) import the symbol FROM `bridge.catchup` rather than
+# re-deriving the path, so patching this one module attribute covers every
+# reader. The fixture imports `bridge.catchup` itself rather than relying on the
+# test having imported it.
+#
+# The flag is REPOINTED at a per-test temp path, not stubbed: `catchup_disabled()`
+# still runs its real `Path.exists()` against a real filesystem path. A test that
+# wants genuine disabled behavior can `touch` the redirected path.
+_CATCHUP_FLAG_ISOLATION_DISABLED = os.environ.get("CATCHUP_FLAG_ISOLATION", "1") == "0"
+
+
+@pytest.fixture(autouse=True)
+def isolate_catchup_kill_switch(tmp_path_factory):
+    """Repoint the catchup operator kill-switch flag at a per-test temp path.
+
+    Escape hatch: set ``CATCHUP_FLAG_ISOLATION=0`` to skip the redirect. That
+    exists so the negative control for #2552 stays permanently reproducible —
+    running the suite with it set must reproduce the flag-caused failures on a
+    host where the real ``data/catchup-disabled`` exists.
+    """
+    if _CATCHUP_FLAG_ISOLATION_DISABLED:
+        yield
+        return
+
+    import bridge.catchup
+
+    redirect_dir = tmp_path_factory.mktemp("catchup-flag")
+    original = bridge.catchup.CATCHUP_DISABLED_FLAG
+    bridge.catchup.CATCHUP_DISABLED_FLAG = redirect_dir / "catchup-disabled"
+    try:
+        yield
+    finally:
+        bridge.catchup.CATCHUP_DISABLED_FLAG = original
+
+
 @pytest.fixture(autouse=True)
 def agent_hooks_consistency_guard():
     """Detect and repair a corrupt `agent` package/submodule cache state.
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -12,6 +12,7 @@ import sys
 import time
 import warnings
 from unittest.mock import MagicMock
+from uuid import uuid4
 
 import pytest
 
@@ -242,9 +243,27 @@ def mock_claude_sdk_cleanup():
 _CATCHUP_FLAG_ISOLATION_DISABLED = os.environ.get("CATCHUP_FLAG_ISOLATION", "1") == "0"
 
 
+@pytest.fixture(scope="session")
+def _catchup_flag_redirect_dir(tmp_path_factory):
+    """One directory for the whole session to hold per-test kill-switch paths.
+
+    Session-scoped on purpose. ``tmp_path_factory.mktemp`` is numbered, so it
+    scans the basetemp for the next free suffix and its cost grows with the
+    number of directories already there. Calling it once per test from an
+    autouse fixture makes that aggregate quadratic and leaves one empty
+    directory per test; the per-test fixture below instead varies only the
+    filename inside this single directory, which costs a string format and no
+    filesystem write at all.
+    """
+    return tmp_path_factory.mktemp("catchup-flag")
+
+
 @pytest.fixture(autouse=True)
-def isolate_catchup_kill_switch(tmp_path_factory):
+def isolate_catchup_kill_switch(_catchup_flag_redirect_dir):
     """Repoint the catchup operator kill-switch flag at a per-test temp path.
+
+    Each test gets a unique path no other test can touch, so a test that wants
+    genuine disabled behavior can create it without leaking into its neighbors.
 
     Escape hatch: set ``CATCHUP_FLAG_ISOLATION=0`` to skip the redirect. That
     exists so the negative control for #2552 stays permanently reproducible —
@@ -257,9 +276,10 @@ def isolate_catchup_kill_switch(tmp_path_factory):
 
     import bridge.catchup
 
-    redirect_dir = tmp_path_factory.mktemp("catchup-flag")
     original = bridge.catchup.CATCHUP_DISABLED_FLAG
-    bridge.catchup.CATCHUP_DISABLED_FLAG = redirect_dir / "catchup-disabled"
+    bridge.catchup.CATCHUP_DISABLED_FLAG = (
+        _catchup_flag_redirect_dir / f"catchup-disabled-{uuid4().hex}"
+    )
     try:
         yield
     finally:
--- a/tests/unit/test_pipeline_integrity.py
+++ b/tests/unit/test_pipeline_integrity.py
@@ -173,7 +173,6 @@ class TestEnqueueContinuationFallback:
 
         critical_fields = [
             "context_summary",
-            "expectations",
             "issue_url",
             "pr_url",
             "session_events",
@@ -193,6 +192,140 @@ class TestEnqueueContinuationFallback:
         assert isinstance(result, dict)
         # Should have hash_exists, popoto_query_matches, or error (if Redis not available)
         assert "hash_exists" in result or "error" in result
+
+
+def _non_autokey_model_fields() -> set[str]:
+    """AgentSession field names eligible to appear in a `create(**fields)` payload.
+
+    AutoKeyFields are excluded because Popoto generates them; passing one into
+    a create payload is an error, so their absence from `_AGENT_SESSION_FIELDS`
+    is correct rather than drift. The exclusion is derived by ``isinstance``
+    against ``AutoKeyField`` and NOT by subtracting a literal ``{"id"}``, so a
+    second auto-key field added later cannot silently reopen the same
+    off-by-one.
+    """
+    from popoto import AutoKeyField
+
+    from models.agent_session import AgentSession
+
+    return {
+        name
+        for name, field in AgentSession._meta.fields.items()
+        if not isinstance(field, AutoKeyField)
+    }
+
+
+# AgentSession fields deliberately NOT copied by `_extract_agent_session_fields`
+# as of today. This is an honest record of an existing gap, not an approval of
+# it: the helper serves two incompatible contracts (delete-and-recreate of the
+# SAME session, where dropping a field is data loss, vs. creation of a NEW
+# session, where the #2518 execution fence must be reset), and resolving that
+# conflict is tracked in issue #2563.
+#
+# TERMINATION CONDITION: when #2563 ships, entries leave this set and it empties.
+# An empty KNOWN_GAP is a valid, meaningful state — the guard below asserts set
+# EQUALITY, so it becomes a plain completeness assertion at that point and a
+# newly added model field can never be silently absorbed into the excuse list.
+#
+# The left-hand side of the comparison excludes auto-key fields; see
+# `_non_autokey_model_fields` for why and how.
+KNOWN_GAP = frozenset(
+    {
+        "active_run_id",
+        "auto_resume_attempts",
+        "budget_tripped",
+        "budget_tripped_reason",
+        "chat_message_log",
+        "claude_version",
+        "continuation_depth",
+        "crash_outcome_attributed",
+        "crash_signature",
+        "current_tool_name",
+        "current_tool_timeout_s",
+        "dev_agent_id",
+        "exec_cwd",
+        "exec_harness",
+        "exec_pid",
+        "exit_reason",
+        "exit_returncode",
+        "is_ledger",
+        "issue_number",
+        "last_authored_at",
+        "last_compaction_ts",
+        "last_tool_use_at",
+        "last_turn_at",
+        "model",
+        "owned_run_ids",
+        "pid_create_time",
+        "pr_number",
+        "project_config",
+        "recent_thinking_excerpt",
+        "requires_real_chrome",
+        "response_delivered_at",
+        "retain_for_resume",
+        "rework_triggered",
+        "runner_cwd",
+        "spawn_history",
+        "task_type",
+        "thread_first_created_at",
+        "thread_run_count",
+        "thread_tool_call_count",
+        "thread_turn_count",
+        "tool_timeout_count_default",
+        "tool_timeout_count_internal",
+        "tool_timeout_count_mcp",
+        "total_cache_read_tokens",
+        "total_cost_usd",
+        "total_input_tokens",
+        "total_output_tokens",
+        "unhealthy_reason",
+        "user_facing_routed",
+        "worker_pid",
+    }
+)
+
+
+class TestAgentSessionFieldContractDrift:
+    """Guard `_AGENT_SESSION_FIELDS` against drift in both directions.
+
+    `df6097fe6` deleted the `expectations` field and nothing detected that the
+    list guarding it had drifted; these two tests are what would have caught it.
+    Five other modules assert *membership* in `_AGENT_SESSION_FIELDS`
+    (`test_health_check_recovery_finalization.py`, `test_agent_session_hierarchy.py`,
+    `test_agent_session.py`, `test_nudge_loop.py`, `test_session_completion_zombie.py`);
+    these are the completeness counterparts.
+    """
+
+    def test_no_phantom_names_in_field_list(self):
+        """Every name in `_AGENT_SESSION_FIELDS` still exists on the model.
+
+        A phantom name makes `_extract_agent_session_fields` raise
+        AttributeError at runtime on the delete-and-recreate path.
+        """
+        from agent.agent_session_queue import _AGENT_SESSION_FIELDS
+
+        phantoms = sorted(set(_AGENT_SESSION_FIELDS) - _non_autokey_model_fields())
+        assert not phantoms, (
+            f"_AGENT_SESSION_FIELDS names field(s) that no longer exist on "
+            f"AgentSession: {phantoms}. Remove them from the list in "
+            f"agent/agent_session_queue.py, or restore the model field."
+        )
+
+    def test_unlisted_model_fields_match_known_gap(self):
+        """Model fields absent from the list equal the frozen KNOWN_GAP exactly."""
+        from agent.agent_session_queue import _AGENT_SESSION_FIELDS
+
+        actual_gap = _non_autokey_model_fields() - set(_AGENT_SESSION_FIELDS)
+        unclassified = sorted(actual_gap - KNOWN_GAP)
+        newly_classified = sorted(KNOWN_GAP - actual_gap)
+        assert actual_gap == KNOWN_GAP, (
+            f"AgentSession field coverage drifted. Unclassified field(s) missing "
+            f"from _AGENT_SESSION_FIELDS and from KNOWN_GAP: {unclassified}. "
+            f"KNOWN_GAP entries no longer in the gap: {newly_classified}. "
+            f"Decide explicitly whether each field should be copied by "
+            f"_extract_agent_session_fields (see issue #2563) and update the "
+            f"list or KNOWN_GAP accordingly."
+        )
 
 
 class TestMergeStageTracking:
--- a/tests/unit/test_reflections_yaml_migration.py
+++ b/tests/unit/test_reflections_yaml_migration.py
@@ -220,12 +220,19 @@ class TestSentryTriageCutover:
         repo_root = Path(__file__).resolve().parent.parent.parent
         registry_path = repo_root / "config" / "reflections.yaml"
 
+        # The contract under guard is *absence of the reflection entry from the
+        # registry*, asserted against parsed YAML — the only signal that survives
+        # tools/reflection_machine_filter.py's safe_load/safe_dump round-trip.
+        # (A comment-substring assertion was deliberately removed here: comments
+        # are destroyed by that round-trip. See TestFilterRoundTripSignalDurability.)
         data = yaml.safe_load(registry_path.read_text())
         names = [r["name"] for r in data["reflections"]]
-        assert "sentry-issue-triage" not in names
-
-        # The pointer comment documenting the migration is still present.
-        assert "sentry-issue-triage migrated to a Claude Code Routine" in registry_path.read_text()
+        assert "sentry-issue-triage" not in names, (
+            "sentry-issue-triage was reintroduced into the local reflections "
+            f"registry at {registry_path}; it has migrated to a Claude Code "
+            "Routine (cloud) and a local entry would double-run the triage. "
+            f"Registry entries: {sorted(names)}"
+        )
 
 
 def _pr_review_audit_cutover_pending() -> bool:
@@ -286,12 +293,111 @@ class TestPrReviewAuditCutover:
         repo_root = Path(__file__).resolve().parent.parent.parent
         registry_path = repo_root / "config" / "reflections.yaml"
 
+        # The contract under guard is *absence of the reflection entry from the
+        # registry*, asserted against parsed YAML — the only signal that survives
+        # tools/reflection_machine_filter.py's safe_load/safe_dump round-trip.
+        # (A comment-substring assertion was deliberately removed here: comments
+        # are destroyed by that round-trip. See TestFilterRoundTripSignalDurability.)
         data = yaml.safe_load(registry_path.read_text())
         names = [r["name"] for r in data["reflections"]]
-        assert "pr-review-audit" not in names
+        assert "pr-review-audit" not in names, (
+            "pr-review-audit was reintroduced into the local reflections registry "
+            f"at {registry_path}; it has been cut over to a Claude Code Routine "
+            "(cloud) and a local entry would double-run the audit. "
+            f"Registry entries: {sorted(names)}"
+        )
 
-        # The pointer comment documenting the cutover is still present.
-        assert "pr-review-audit removed from local scheduling" in registry_path.read_text()
+
+class TestFilterRoundTripSignalDurability:
+    """Why the cutover guards above assert on parsed YAML and never on comments.
+
+    ``tools/reflection_machine_filter.py`` rewrites ``config/reflections.yaml``
+    in place with ``yaml.safe_load`` → ``yaml.safe_dump`` (``:110`` / ``:146``).
+    That round-trip discards comments by construction, so a comment-substring
+    assertion is a latent red: it passes only until the next ``install_worker.sh``
+    run and then fails for a reason unrelated to any live contract.
+
+    This test demonstrates the destruction rather than citing it, and proves the
+    replacement assertion style survives the same round-trip. It operates
+    entirely on ``tmp_path`` and never reads or writes the real machine-local
+    ``config/reflections.yaml``.
+    """
+
+    def test_filter_destroys_comments_but_preserves_structure(self, tmp_path: Path):
+        registry = tmp_path / "reflections.yaml"
+        projects = tmp_path / "projects.json"
+
+        pointer_comment = "# sentry-issue-triage migrated to a Claude Code Routine (cloud)"
+
+        # The synthetic registry carries three things:
+        #   (a) a pointer comment, the substrate the deleted assertions relied on;
+        #   (b) the structural state the cutover tests assert on — a live entry
+        #       list from which "sentry-issue-triage" is absent;
+        #   (c) a reflection whose project_key maps to a machine OTHER than the
+        #       one we pass in. (c) is mandatory: filter_reflections_for_machine
+        #       only writes the file back when disabled_names is non-empty
+        #       (tools/reflection_machine_filter.py:145-146), so without it the
+        #       round-trip silently no-ops and this control would prove nothing.
+        _write(
+            registry,
+            f"""
+            {pointer_comment}
+            reflections:
+              - name: cutover-marker
+                every: 60s
+                priority: low
+                execution_type: function
+                callable: x.y
+                enabled: true
+              - name: other-machine-audit
+                every: 3600s
+                priority: high
+                execution_type: function
+                callable: x.z
+                project_key: otherproj
+                enabled: true
+            """,
+        )
+        projects.write_text('{"projects": {"otherproj": {"machine": "Some Other Machine"}}}')
+
+        assert pointer_comment in _read(registry)  # substrate present before the filter
+
+        from tools.reflection_machine_filter import filter_reflections_for_machine
+
+        count, disabled_names = filter_reflections_for_machine(
+            registry, projects, machine_name="This Machine"
+        )
+
+        # Success Criterion 7a — the round-trip actually wrote. Guards against a
+        # silent no-op control, which would make the two assertions below vacuous.
+        assert disabled_names, (
+            "filter_reflections_for_machine disabled nothing, so it never wrote the "
+            "file and the round-trip below is vacuous; the synthetic registry must "
+            "contain a reflection whose project_key maps to another machine"
+        )
+        assert count == len(disabled_names) == 1
+        assert disabled_names == ["other-machine-audit"]
+
+        round_tripped = _read(registry)
+
+        # Success Criterion 6 — negative control for the DELETED assertion style.
+        # A comment-substring assertion placed here would fail.
+        assert pointer_comment not in round_tripped, (
+            "the safe_load/safe_dump round-trip was expected to destroy the pointer "
+            "comment; if it now survives, the comment-based assertions removed from "
+            "the cutover tests could be restored"
+        )
+        assert "#" not in round_tripped  # no comment of any kind survives
+
+        # Success Criterion 7 — positive control for the REPLACEMENT style, against
+        # the SAME round-tripped artifact: parsed structure survives intact.
+        data = yaml.safe_load(round_tripped)
+        names = [r["name"] for r in data["reflections"]]
+        assert names == ["cutover-marker", "other-machine-audit"]
+        assert "sentry-issue-triage" not in names
+        by_name = {r["name"]: r for r in data["reflections"]}
+        assert by_name["cutover-marker"]["enabled"] is True
+        assert by_name["other-machine-audit"]["enabled"] is False
 
 
 class TestNoTempFileLeak:
--- a/tests/unit/test_update_hardlinks.py
+++ b/tests/unit/test_update_hardlinks.py
@@ -20,9 +20,28 @@ _REPO_ROOT = Path(__file__).resolve().parents[2]
 _SKILL_ROOTS = (".claude/skills-global", ".claude/skills")
 
 
+# Directories under a skill root that are legitimately not skills. An explicit
+# named allowlist, not a ``_``-prefix rule: an allowlist of one is honest about
+# how thin the evidence for a convention is, and it fails loudly when a second
+# non-skill directory appears instead of silently absorbing it.
+HUSK_GUARD_ALLOWLIST = frozenset({"_shared"})
+
+
+def _skill_is_live(path: Path) -> bool:
+    """True if ``path`` is a live skill, i.e. it holds a ``SKILL.md``.
+
+    Skill liveness is defined by ``SKILL.md`` presence, never by directory
+    existence. A directory whose ``SKILL.md`` was deleted but which still holds
+    a ``__pycache__`` or a stray reference file is a *husk*: the skill is gone,
+    but a bare ``Path.is_dir`` probe still reads True. That blind spot is what let the
+    ``do-skills-audit`` husk survive its own rename (#2557, #2523).
+    """
+    return (path / "SKILL.md").is_file()
+
+
 def _skill_exists_in_any_root(name: str) -> bool:
-    """True if a skill dir ``name`` currently exists under either skill root."""
-    return any((_REPO_ROOT / root / name).is_dir() for root in _SKILL_ROOTS)
+    """True if a live skill ``name`` currently exists under either skill root."""
+    return any(_skill_is_live(_REPO_ROOT / root / name) for root in _SKILL_ROOTS)
 
 
 @pytest.fixture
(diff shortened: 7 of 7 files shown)


valorengels:

Single-judge review (standalone invocation; `SDLC_REVIEW_JUDGES` unset in this environment, so the multi-judge path did not run). No blockers. The instrument repair is sound and, more importantly, **provably capable of failing** — I independently reproduced every negative control rather than crediting the acceptance table. Two tech-debt items hold the verdict: the PR body now describes code that the last commit removed, and the autouse fixture has a measurable quadratic cost that lands squarely in the blast radius the plan deferred.

### What I reproduced myself (not read from the PR body)

Per "a number the PR claims is a claim, not evidence", every figure below was measured in this worktree on `e90a554d7`.

| Check | Method | Result |
|---|---|---|
| Criterion 1 — focused acceptance | `scripts/pytest-clean.sh -n0` over the 7 baseline files | **141 passed, 2 skipped, 0 failed** — matches the PR body exactly |
| **Criterion 2 — negative control #2552** | `touch data/catchup-disabled` in the worktree (gitignored; removed after), then `CATCHUP_FLAG_ISOLATION=0 scripts/pytest-clean.sh -n0` over the 3 catchup files | **17 failed, 40 passed**, distribution **2 / 13 / 2** across `test_agent_catchup` / `test_reconciler` / `test_catchup_claim` — the plan's arm-1 distribution, node-for-node |
| Control arm — flag present, fixture ON | same 3 files, flag present, no escape hatch | **57 passed** — the fixture is what does the work, not the absence of the flag |
| Criterion 4 — phantom guard | appended `bogus_phantom_field_xyz` to `_AGENT_SESSION_FIELDS` in-memory | fired: `_AGENT_SESSION_FIELDS names field(s) that no longer exist on AgentSession: ['bogus_phantom_field_xyz']` |
| Criterion 5 — drift guard | removed `worker_pid` from `KNOWN_GAP` in-memory | fired: `Unclassified field(s) ...: ['worker_pid']` |
| Criteria 6/7/7a | ran `TestFilterRoundTripSignalDurability` as a named node | passed; asserts write-happened, comment-destroyed, structure-survived against the same artifact |
| Criterion 8 — husk guard | created `.claude/skills/zz-review-husk-check/leftover.txt` (removed after) | fired, naming the directory |
| Criterion 15 — lint | `ruff check .` / `ruff format --check .` | clean, 1273 files formatted |

**The `CATCHUP_FLAG_ISOLATION=0` escape hatch is live configuration, not dead configuration.** That was the single highest-value thing to verify and it holds: with the hatch set and the flag present, the exact 17-node failure set returns. The hatch is read once at conftest import into a module constant, which is correct for its stated use (set it before launching pytest); it is not per-test toggleable, and it does not need to be.

**The fixture repoints rather than stubs, and the mechanism is genuinely sound.** All three production readers call `catchup_disabled()` (`bridge/catchup.py:73`, `bridge/reconciler.py:184`, `bridge/agent_catchup.py:794`), which reads the module global **at call time** — `bridge/reconciler.py:182` and `bridge/agent_catchup.py:792` do function-local `from bridge.catchup import ...` only to get the name for the log line. So patching the one module attribute really does cover every reader, and the arm-B/arm-C split proves it empirically rather than by argument. `catchup_disabled()` still runs a real `Path.exists()` against a real path, so the disabled path remains exercisable by touching the redirect.

### Scrutiny items, resolved

- **`KNOWN_GAP` is set-equality, not subset or length.** `assert actual_gap == KNOWN_GAP` at `tests/unit/test_pipeline_integrity.py:322`. Measured live: `AgentSession._meta.fields` = 89, non-`AutoKeyField` = 88, `_AGENT_SESSION_FIELDS` = 38, actual gap = **50**, `len(KNOWN_GAP)` = **50**, sets equal, phantoms = 0. The plan's arithmetic reconciles exactly.
- **The `AutoKeyField` exclusion is `isinstance`-derived, not `- {"id"}`.** `_non_autokey_model_fields` (`:197`) filters `if not isinstance(field, AutoKeyField)`. A second auto-key field cannot reopen the off-by-one. This was a critique BLOCKER and it is genuinely resolved, not resolved-on-paper.
- **`TestFilterRoundTripSignalDurability` does seed an other-machine `project_key`.** `other-machine-audit` with `project_key: otherproj` mapped to `"Some Other Machine"` against `machine_name="This Machine"`, and the test asserts `disabled_names == ["other-machine-audit"]` **before** the comment/structure assertions. Without that seed `filter_reflections_for_machine` returns early at `tools/reflection_machine_filter.py:145-146` and the control is vacuous; the 7a assertion is what makes the vacuity impossible rather than merely unlikely. It operates entirely on `tmp_path` and `projects.json` written into `tmp_path` — the real `config/reflections.yaml` is never opened.
- **`HUSK_GUARD_ALLOWLIST` is `frozenset({"_shared"})`, checked by membership.** `tests/unit/test_update_hardlinks.py:27` and `:451` (`entry.name not in HUSK_GUARD_ALLOWLIST`). No `startswith("_")` anywhere in the file.
- **All three former liveness sites route through `_skill_is_live`.** `:44` (`_skill_exists_in_any_root`), `:432`, `:433` (`test_renamed_removals_entries_are_not_stale`), plus `:451` inside `_find_husk_dirs`. The surviving `is_dir()` calls at `:446` and `:451` filter directory entries and are not liveness probes — consistent with the amended Criterion 11 on `main` and with `e90a554d7`'s message. Agreed with the amendment: bending the code's spelling to keep a whole-file grep at zero was the worse artifact.
- **Anti-criteria (independently confirmed).** `git diff --name-only origin/main...HEAD -- bridge/` → 0 lines. `git diff origin/main...HEAD -- agent/agent_session_queue.py` → 0 lines. `data/catchup-disabled` in the main checkout: **0 bytes, mtime Jul 22 13:45:47 2026**, unchanged. The worktree flag I created for the negative control was removed and never existed in the main checkout.
- **Husk sweep (Criteria 9, 10).** `.claude/skills-global/do-skills-audit` absent in the worktree, in the main checkout, and in `~/.claude/skills/`. `.claude/skills/_shared/test-quality.md` present, and `_shared` is the only `SKILL.md`-less directory in either root on both checkouts — the allowlist of one is honest today.
- **Merge blockers.** No `Co-Authored-By` and no "Generated with Claude Code" trailer on any of the six commits.
- **Blast-radius grep is sufficient for the deferred item.** `grep -rn "CATCHUP_DISABLED_FLAG\|catchup_disabled" tests/` returns only `tests/README.md` prose and the fixture itself. No test wants the flag set, so the correctness half of the deferred full-suite verification is answered by construction. The *performance* half is not — see Miscellaneous below.

## Rubric

- [x] **1. Plan vs. implementation match** — pass — *all 15 Success Criteria verified independently; 6 negative controls all fire.*
- [x] **2. New code quality** — pass — *type hints present, guards name their termination condition, failure messages name the offending field/directory. The efficiency issue is bucketed under Miscellaneous rather than here — it is not a correctness or readability defect.*
- [x] **3. Test coverage** — pass — *this PR is test coverage; `test_find_husk_dirs_edge_cases` covers missing root, empty root, empty `SKILL.md`, allowlisted dir, loose file, real husk.*
- [x] **4. Regression risk to existing callers** — pass — *no production file changed; the autouse fixture's only reachable effect is relocating one module attribute, and no test reads it.*
- [x] **5. Data integrity** — n/a — *no schema, migration, or persisted-model change.*
- [x] **6. Security** — pass — *no secrets, no `eval`/`exec`/`subprocess` on external input, no debug artifacts.*
- [x] **7. Documentation accuracy** — pass — *`tests/README.md` records the fixture, the escape hatch and the contamination class; `docs/features/skills-global.md` records the `SKILL.md` liveness definition and the allowlist rationale. Both match the code.*
- [ ] **8. PR body accuracy** — fail — *acceptance row 11 and a Reviewer note describe code deleted by the final commit. See Tech Debt.*
- [x] **9. Disclosed deferrals** — pass — *#2563 and #2473 both OPEN; the ORDERED full-suite deferral is an explicit plan No-Go with sound rationale.*
- [x] **10. Follow-up claims verified** — pass — *#2552, #2553, #2556, #2557, #2523 all OPEN and correctly targeted by `Closes`; #2563, #2473 OPEN.*

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — PASS — *15/15 verified by independent execution.*
- **2. No-Gos from plan — none violated** — PASS — *all three anti-criteria confirmed by direct measurement; the four SEPARATE-SLUG No-Gos are untouched.*
- **3. New `except Exception` blocks — each has logger/raise/swallow-ok** — N/A — *none introduced.*
- **4. New integration tests — exercise serialization boundary (not in-memory only)** — PASS — *`TestFilterRoundTripSignalDurability` runs the real filter against a real file and re-reads the written artifact; the husk guard walks the real filesystem.*
- **5. Plan internal consistency — spike findings match task steps** — PASS — *the four critique findings are each implemented as specified; the Technical Approach / Risk 4 contradiction is resolved in favor of the allowlist.*
- **7. New public APIs — docstrings present** — PASS — *`_skill_is_live`, `_find_husk_dirs`, `_non_autokey_model_fields`, and the fixture all carry docstrings stating the why.*
- **8. Breaking changes — migration path documented** — N/A.
- **10. Tests cover the failure path (not just happy path)** — PASS — *this is the plan's whole thesis; six negative controls, all reproduced here.*
- **11. UI changes (if any) — screenshot captured** — N/A — *no HTML/CSS/JS/TSX/template files in the diff.*

### Verified
- [x] Code correctness
- [x] Plan requirements met

### Blockers
- None

### Tech Debt

**File:** PR description, "Acceptance evidence" table row 11 and the second "Reviewer notes" bullet
**Code:** `| 11 | \`grep -c 'is_dir()' tests/unit/test_update_hardlinks.py\` | 0 |` and "**`_find_husk_dirs` uses `os.path.isdir`, not `Path.is_dir`.** Deliberate: Success Criterion 11 is a whole-file `grep -c 'is_dir()'`..."
**Issue:** Both statements were made obsolete by the final commit `e90a554d7`, which restored `Path.is_dir` for entry-type filtering. Measured on this head: `grep -c 'is_dir()' tests/unit/test_update_hardlinks.py` returns **2**, not 0, and `grep -n "os.path" tests/unit/test_update_hardlinks.py` returns **nothing** — the file contains no `os.path` call at all. Success Criterion 11 was correspondingly amended on `main` to "all three former liveness sites route through a single `_skill_is_live(path)` helper", which the code satisfies. The code is right and the amendment is right; the PR body is the stale artifact. It is also the merge record, and it currently tells a future reader that a whole-file grep gates this change and that a deliberate `os.path.isdir` workaround exists — a reader who checks either claim finds it false and has no way to tell whether the code or the description drifted.
**Severity:** tech_debt
**Fix:** Update row 11 to the amended criterion ("3/3 liveness sites route through `_skill_is_live`; no liveness decision reads directory existence") with the observed result, and replace the `os.path.isdir` reviewer note with the reasoning from `e90a554d7`'s commit message — that the code stopped bending to a bad measurement and the measurement was corrected instead. No code change required.

### Nits
- None

### Miscellaneous

**File:** `tests/conftest.py:260`
**Code:** `redirect_dir = tmp_path_factory.mktemp("catchup-flag")`
**Issue:** This is an autouse, unconditional, suite-wide fixture, so `mktemp` runs once per test. `tmp_path_factory.mktemp` with the default `numbered=True` goes through `_pytest.pathlib.make_numbered_dir`, which scans the basetemp's existing entries to find the next suffix — cost grows with the number of directories already created, making the aggregate quadratic. Measured directly on this host: 500 dirs = 0.15s, 1500 = 0.93s, 3000 = 3.34s. This suite collects **12,851 tests**, so a serial (`-n0`) run — the mode this repo's rails mandate — extrapolates to roughly a minute of pure directory bookkeeping and 12,851 empty directories per basetemp, retained across the last three run roots. Under `-n auto` the cost divides across per-worker basetemps and is minor; serial is where it bites. I am labelling the 12,851-test figure an extrapolation from the micro-benchmark, not a measured full-suite delta — the full-suite slot is occupied and running it is an ORDERED No-Go in the plan. This is not a correctness problem and the guard works exactly as designed; it is the one part of the deferred blast-radius question that the `CATCHUP_DISABLED_FLAG` grep does not answer, since the grep addresses semantic interference and this is resource cost.
**Severity:** tech_debt
**Fix:** Create the directory once at session scope and give each test a unique *filename* inside it rather than a fresh numbered directory — e.g. a `session`-scoped fixture returning `tmp_path_factory.mktemp("catchup-flag")` once, with the per-test fixture setting `CATCHUP_DISABLED_FLAG = base / f"catchup-disabled-{uuid4().hex}"`. Per-test isolation is preserved (each test still gets a path no other test can touch, and a test that wants disabled behavior can still create it), the real `Path.exists()` still runs, and the per-test cost drops to a string format with no filesystem write at all.

- **#2563 — the `_extract_agent_session_fields` two-contract conflict and the 50-field gap** — tracked by #2563 (OPEN). Correctly deferred: five of the six fields the original issue named are #2518 fenced-execution-record fields, and copying them into a newly created session would forge a fence describing a process that never ran. `KNOWN_GAP` records the gap as set-equality with #2563 named as its termination condition, so it cannot silently absorb a new field.
- **#2473 — the fate of the production `data/catchup-disabled` flag** — tracked by #2473 (OPEN). Operator-gated state; correctly untouched, and confirmed intact at 0 bytes / mtime Jul 22 13:45.
- **Full-suite verification of the autouse fixture's blast radius** — explicit `[ORDERED]` No-Go in the plan, not a tracked issue. Sound: the host's full-suite slot is occupied. The grep evidence (no test references `CATCHUP_DISABLED_FLAG` or `catchup_disabled`) is sufficient for the semantic half of the question and I confirmed it independently. It does not cover the resource half — see Miscellaneous.
- **The husk sweep produces no reviewable diff** — explicit out-of-scope disclosure with sound rationale; `do-skills-audit` is fully untracked. Verified absent in both checkouts and in `~/.claude/skills/`, and the durable artifact (the guard) is reviewable and was proven to fire.

### Verification Results

Plan `## Success Criteria` table, all 15 rows run on this branch:

- **1** 141 passed / 2 skipped / 0 failed (7 focused files, serial) — **pass**
- **2** 17 failed, 2/13/2 distribution, with `CATCHUP_FLAG_ISOLATION=0` and the flag present — **pass**
- **3** `grep -c '"expectations"' tests/unit/test_pipeline_integrity.py` = 0 — **pass**
- **4** phantom guard fired naming the bogus field — **pass**
- **5** `KNOWN_GAP` guard fired naming `worker_pid` — **pass**
- **6, 7, 7a** `TestFilterRoundTripSignalDurability` passes; `disabled_names == ["other-machine-audit"]`, comment gone, structure intact — **pass**
- **8** husk guard fired naming `.claude/skills/zz-review-husk-check` — **pass**
- **9** `do-skills-audit` absent (worktree, main checkout, `~/.claude/skills/`) — **pass**
- **10** `.claude/skills/_shared/test-quality.md` present — **pass**
- **11** (amended form on `main`) 3/3 liveness sites route through `_skill_is_live`; no liveness decision reads directory existence — **pass**
- **12** `git diff --name-only origin/main...HEAD -- bridge/` empty — **pass**
- **13** `git diff origin/main...HEAD -- agent/agent_session_queue.py` empty — **pass**
- **14** `data/catchup-disabled` present, 0 bytes, mtime Jul 22 13:45:47 2026 — **pass**
- **15** `ruff check .` and `ruff format --check .` clean — **pass**

Note for the merger, not a finding: this branch's copy of `docs/plans/wave2-test-instrument.md:274` still carries the pre-amendment Criterion 11 wording, because the branch predates the correction on `main` and touched only the `## Documentation` section of that file. A merge keeps `main`'s corrected line and adds the branch's documentation ticks — `mergeStateStatus` is `CLEAN` and no conflict exists. Editing that line on the branch would be the riskier move; leave it.

### Screenshots
None — no UI files in the diff.

valorengels:
Both review findings addressed; requesting re-review.

**Tech debt 1 — PR body accuracy.** Row 11 and the `os.path.isdir` reviewer note are rewritten to match the code at head. The row now states the amended criterion (3/3 liveness sites route through `_skill_is_live`), and the note explains that the criterion was corrected rather than the code contorted.

**Tech debt 2 — quadratic `mktemp`** (`7f663d646`). Good catch, and the fix is the one you specified: a session-scoped fixture creates the redirect directory once, and the per-test fixture varies only the filename via `uuid4().hex`. Per-test cost is now a string format with no filesystem write, and the suite no longer leaves one empty directory per test.

Re-verified after the refactor, not assumed: focused acceptance run is **141 passed / 2 skipped / 0 failed**, and the `CATCHUP_FLAG_ISOLATION=0` negative control still yields exactly **17 failed / 40 passed** with the 2/13/2 distribution across `test_agent_catchup` / `test_reconciler` / `test_catchup_claim`. Ruff check and format clean. Production `data/catchup-disabled` still 0 bytes, mtime Jul 22 13:45, untouched.

Head is now `7f663d646`.

valorengels:
## Delta re-review — APPROVED

Head resolved git-first via `git ls-remote refs/pull/2568/head` = `7f663d64685b980e4a5fd3f7586f93736255117f`. Both findings from the prior review (0 blockers, 2 tech-debt) are fixed, and nothing else rode in.

**Diff since the reviewed head.** `git diff e90a554d7..7f663d646 --stat` is a single file, `tests/conftest.py` (+23/-3), in one commit (`7f663d646`). No other code, test, or doc file changed. The PR-body correction is a body edit and produces no diff, as expected.

**TD2 — quadratic autouse cost. Fixed.** `tmp_path_factory.mktemp("catchup-flag")` moved into a new session-scoped `_catchup_flag_redirect_dir` fixture, so the numbered-directory scan runs once per session instead of once per test. The autouse fixture now varies only the filename, `f"catchup-disabled-{uuid4().hex}"`, which is a string format against a `Path` and performs no filesystem write. Isolation is preserved and arguably stronger: every test still gets a path no other test can reach, `catchup_disabled()` still runs a real `Path.exists()` against a real filesystem path, and a test wanting genuine disabled behavior can still `touch` its own redirect target. The `CATCHUP_FLAG_ISOLATION=0` escape hatch still short-circuits before the redirect.

**TD1 — stale PR body. Fixed.** Acceptance row 11 now reads "All three former `.is_dir()` liveness sites route through `_skill_is_live` (a `SKILL.md` file check)", matching the criterion as amended on main in `6cd27be71`. The reviewer note no longer claims `os.path.isdir` is in the code; it explains that `e90a554d7` restored the natural `Path.is_dir` spelling and that the file contains no `os.path` call. A "Post-review patch" section records both fixes.

**Evidence re-run independently at this head**, in the lane worktree, serial `-n0` via `scripts/pytest-clean.sh`:

- Focused run over the seven files (`test_agent_catchup`, `test_reconciler`, `test_catchup_claim`, `test_dedup`, `test_pipeline_integrity`, `test_reflections_yaml_migration`, `test_update_hardlinks`): **141 passed, 2 skipped, 0 failed** in 30.83s. Matches the claim.
- Negative control, `CATCHUP_FLAG_ISOLATION=0` over the three catchup files: **17 failed, 40 passed**, distribution **2 / 13 / 2** across `test_agent_catchup` / `test_reconciler` / `test_catchup_claim`. Matches the claim exactly, including the node list.

One wrinkle worth recording for whoever reproduces this next: the flag path is source-tree anchored, so the negative control is inert in a worktree, which carries no `data/catchup-disabled`. Run bare in the worktree it yields 57 passed, not 17 failed. I reproduced the real control by creating a 0-byte `data/catchup-disabled` inside the *worktree* (gitignored), running, then removing it. The production flag in the main checkout was never touched: still 0 bytes at its original Jul 22 13:45 mtime, and anti-criterion 14 holds. The worktree is clean afterward (`git status --porcelain` empty).

`ruff check` and `ruff format --check` are clean on `tests/conftest.py`.

Non-blocking observation, not counted as debt: because `_catchup_flag_redirect_dir` is a parameter of the autouse fixture, its one `mktemp` still runs even when `CATCHUP_FLAG_ISOLATION=0` makes the redirect a no-op. That is a single empty directory per session and not worth code.

**Verdict: APPROVED. 0 blockers, 0 tech debt.**

valorengels:
## Review: Approved

Single-judge **delta re-review** at head `7f663d646` (`SDLC_REVIEW_JUDGES` unset in this environment, so the multi-judge path did not run). The prior review on `e90a554d7` returned Changes Requested — Tech Debt, 0 blockers, 2 items. **Both are resolved, and I verified the TD2 fix behaviourally rather than by reading it.**

Concentrated on what changed since that verdict (`e90a554d7`, `7f663d646`, and the PR body edits) rather than re-litigating the 15 Success Criteria the prior reviewer independently passed — but I re-measured every number this review leans on, using the commands the PR body now publishes.

### What I re-measured (all from the worktree at `7f663d646`)

| Measurement | Claimed | Observed |
|---|---|---|
| Focused acceptance (7 files, `-n0`) | 141 passed / 2 skipped / 0 failed | **141 passed, 2 skipped in 45.32s** ✅ |
| #2552 negative control (`CATCHUP_FLAG_ISOLATION=0`, flag present) | 17 failed / 40 passed, 2/13/2 | **17 failed, 40 passed** — 2 `test_agent_catchup` / 13 `test_reconciler` / 2 `test_catchup_claim` ✅ |
| Field-drift arithmetic | 89 / 88 / 38 / KNOWN_GAP 50 / 0 phantoms | **`89 88 38 50 []`** ✅ |
| Anti-criterion 12 (`bridge/` untouched) | empty | `git diff --name-only origin/main -- bridge/` → **0 lines** ✅ |
| Anti-criterion 13 (`agent_session_queue.py` untouched) | empty | **0 lines** ✅ |
| Anti-criterion 14 (operator flag intact) | present, 0 bytes, Jul 22 13:45 | main checkout: **0 bytes, Jul 22 13:45** — untouched by my runs ✅ |
| Criterion 3 (`"expectations"` gone) | 0 | **0** ✅ |
| Criterion 9 / 10 (husk swept, `_shared` kept) | absent / present | **absent in both checkouts / present** ✅ |
| Criterion 15 (lint) | clean | `ruff check` **All checks passed**; `format --check` **1273 files already formatted** ✅ |

The negative control was run with `data/catchup-disabled` created in the **worktree** and removed afterwards; the operator's copy in the main checkout was never touched (mtime confirmed unchanged above).

### TD2 — quadratic autouse fixture (`7f663d646`): fixed, and isolation independently proven

The refactor is correct in shape (`_catchup_flag_redirect_dir` session-scoped, one `mktemp`; the per-test fixture varies only the filename via `uuid4().hex`). Shape is not proof, so I wrote a throwaway probe test file, ran it, and deleted it. All four invariants hold:

- **Unique per test** — `test_b` and `test_c` each observed a `CATCHUP_DISABLED_FLAG` different from every prior test's.
- **One shared directory** — all three tests resolved to the same parent (`.../pytest-238/catchup-flag0`), confirming the per-test `mktemp` is genuinely gone.
- **Real `Path.exists()` still runs** — `test_a` touched its redirected path and `catchup_disabled()` immediately returned `True`; the flag is repointed, not stubbed.
- **No leak** — `test_b`, running straight after `test_a`'s touch, observed `catchup_disabled() == False`.

Second, stronger check the PR does not claim: I ran the three catchup files with `data/catchup-disabled` **present** and isolation **on** → **57 passed**. Same three files with isolation **off** → **17 failed**. The fixture is doing exactly the work it exists to do, after the refactor.

The escape hatch still reproduces the failures exactly, including the 2/13/2 distribution.

### TD1 — PR body accuracy: fixed

Acceptance row 11 now reads "All three former `.is_dir()` liveness sites route through `_skill_is_live`", matching the amendment on `main` (`6cd27be71`), and the stale `os.path.isdir` reviewer note is replaced with `e90a554d7`'s reasoning. I checked the body's one falsifiable claim: `grep -c "os\.path" tests/unit/test_update_hardlinks.py` → **0**. Accurate.

I also verified the amended criterion against the code rather than against the body: liveness is decided at exactly three sites — `_skill_exists_in_any_root` (:44) and both probes in `test_renamed_removals_entries_are_not_stale` (:432, :433) — and all three call `_skill_is_live`, whose body is `(path / "SKILL.md").is_file()`. The two surviving `is_dir()` calls in `_find_husk_dirs` (:446, :451) filter directory entries and make no liveness decision. **3/3.**

Two smaller things I checked because the delta could have broken them and nobody would notice:

- `os` is still imported and still used (`os.stat`, `os.link`) after `e90a554d7` removed the `os.path` calls — no dead import, consistent with ruff passing.
- `tests/README.md`'s fixture description ("repoints … at a per-test tmp **path**") survives the refactor unchanged: it is still a per-test path, and it was never described as a per-test *directory*. No doc staleness introduced by `7f663d646`.

### Plan-vs-branch note (not a finding)

The plan copy on this branch still carries the pre-amendment row 11, because the branch predates `6cd27be71`. That is the expected consequence of the plans-on-`main` convention. I confirmed via `git merge-tree` that the merge result keeps `main`'s amended row 11, so nothing regresses on merge — consistent with `mergeStateStatus=CLEAN`.

### Open item for the merge decision: no full-suite evidence

The PR body discloses this honestly and does not claim it — a `tests/unit/` run at this head wedged at 99% with `[gw5] node down`. I did not attempt one (build-only rails), so I cannot close it, and I am not treating it as a blocker. My assessment of whether the remaining evidence is a sufficient merge basis:

- **The diff is test-and-docs-only.** Changed files: `tests/conftest.py`, three `tests/unit/` files, `tests/README.md`, `docs/features/skills-global.md`, `docs/plans/…`. Anti-criteria 12 and 13 re-verified empty. No production behaviour can regress from this diff.
- **The only suite-wide surface is the autouse fixture**, and its blast radius is now characterized rather than assumed: it rebinds one module attribute inside `try/finally`, restores it unconditionally, and **no test in the tree references `CATCHUP_DISABLED_FLAG` or `catchup_disabled`** (I re-ran that grep — only `tests/README.md` prose matches). There is no test that wants the flag set for the fixture to fight with.
- **`7f663d646` strictly reduces that surface** versus the version the prior reviewer already assessed: one `mktemp` per session instead of one per test, and no per-test directory left behind. The delta lowers full-suite risk; it cannot have introduced a full-suite hazard the prior assessment missed.
- The plan's Success Criteria (rows 1–15) contain no full-suite row; row 1 is measured by the focused run, which I reproduced exactly.

On that basis, the focused evidence plus the prior reviewer's independent verification of all 15 criteria is a sufficient merge basis, with the wedged run recorded as a disclosed open item for the merge gate to weigh rather than as an unmet criterion.

One caveat I will not paper over: the "Baseline (main checkout): 18 failed / 120 passed" figure is the only number in the body I did **not** independently re-measure this pass (it requires a run on `main`, and it is settled ground the prior reviewer covered). I did check it for internal consistency — 138 baseline nodes + the 5 tests this PR adds = 143 = the 141 passed + 2 skipped I observed, and 17 flag-caused + 1 stale-`expectations` failure = 18. It holds arithmetically, but treat it as carried-forward, not re-verified here.

## Rubric

- [x] **1. Plan vs. implementation match** — pass — *All 15 criteria satisfied against the amended plan on `main`; rows 1, 2, 3, 5, 9, 10, 11, 12, 13, 14, 15 re-verified directly this pass.*
- [x] **2. New code quality** — pass — *The session-scoped fixture is well-named, documents why it is session-scoped, and the escape-hatch branch still short-circuits cleanly. `finally` restores unconditionally.*
- [x] **3. Test coverage** — pass — *5 new tests, each shipping a reproduced negative control. Focused run green at 141/2/0.*
- [x] **4. Regression risk to existing callers** — pass — *Autouse fixture touches one module attribute with `try/finally` restore; no test in the tree reads the flag (grep re-run). Positive control with the flag present and isolation on → 57 passed.*
- [x] **5. Data integrity** — n/a — *No schema, model, or migration changes; `_AGENT_SESSION_FIELDS` is guarded, not edited (anti-criterion 13 empty).*
- [x] **6. Security** — pass — *No secrets, no `eval`/`exec`/`subprocess` on derived input. `uuid4().hex` is a filename discriminator, not a security primitive.*
- [x] **7. Documentation accuracy** — pass — *`tests/README.md` and `docs/features/skills-global.md` both still describe the code at head after the refactor; fixture table row added.*
- [x] **8. PR body accuracy** — pass — *Was the prior TD1. Row 11 and the reviewer note now match the code; the `os.path` claim verified at 0; every re-measured number matched.*
- [x] **9. Disclosed deferrals** — pass — *#2563 and #2473 both open; the full-suite gap is disclosed rather than omitted.*
- [x] **10. Follow-up claims verified** — pass — *See Acknowledged Deferrals.*

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — PASS — *11 of 15 re-measured this pass; remainder carried from the prior reviewer's independent verification.*
- **2. No-Gos from plan — none violated** — PASS — *Anti-criteria 12/13/14 all re-verified; operator flag mtime unchanged.*
- **4. New integration tests — exercise serialization boundary** — PASS — *`TestFilterRoundTripSignalDurability` asserts against the round-tripped artifact, not an in-memory object.*
- **5. Plan internal consistency — spike findings match task steps** — PASS — *Row 11 amendment on `main` reconciles the criterion with the code; merge result preserves it.*
- **6. No hardcoded secrets or debug artifacts** — PASS — *None; my probe test file was removed (`git status` clean).*
- **7. New public APIs — docstrings present** — PASS — *Both new fixtures carry docstrings explaining the why, not the what.*
- **8. Breaking changes — migration path documented** — N/A — *Test-only diff.*
- **9. Tests added for new behavior** — PASS — *5 new tests.*
- **10. Tests cover the failure path (not just happy path)** — PASS — *Every guard ships a negative control; I reproduced the #2552 one and the isolation invariants directly.*
- **11. UI changes (if any) — screenshot captured** — N/A — *No UI files in the diff.*
- **12. Docs updated for user-facing changes** — PASS — *`tests/README.md`, `docs/features/skills-global.md`.*

### Verified
- [x] Code correctness
- [x] Test coverage
- [x] Plan requirements met

### Blockers
- None

### Tech Debt
- None

### Nits
- None

### Miscellaneous
- None

- **#2563 — `_extract_agent_session_fields` two-contract conflict / the 50-field gap** — tracked by #2563 (OPEN); `KNOWN_GAP` names it as its own termination condition.
- **#2473 — fate of the production `data/catchup-disabled` flag** — tracked by #2473 (OPEN); operator-gated state, correctly untouched.
- **Full-suite verification of the autouse fixture's blast radius** — explicitly disclosed in the PR body as attempted-and-wedged, not claimed as evidence. No tracking issue; see "Open item for the merge decision" above for why the remaining evidence is judged a sufficient merge basis and what residual risk is being accepted.

### Review Delta (vs prior review on HEAD `e90a554d`)
- **Resolved — TD1 (PR body accuracy).** Row 11 and the `os.path.isdir` reviewer note rewritten to match the code and the `main` amendment. `grep -c "os\.path"` → 0, confirming the body's replacement claim.
- **Resolved — TD2 / Miscellaneous (`tests/conftest.py:260`, per-test `mktemp`).** `7f663d646` moves the `mktemp` to a session-scoped fixture and varies only the filename. Isolation re-proven behaviourally (unique paths, shared dir, live `Path.exists()`, no cross-test leak) plus a flag-present/isolation-on positive control at 57 passed. Negative control unchanged at 17/40 in 2/13/2.
- **New** — none.
- **Unchanged** — the absent full-suite evidence, which was outside the prior verdict's findings and remains an open item rather than a finding.

### Screenshots

valorengels:
## Merging — basis stated explicitly

Merge gate passed via the shared deterministic predicate (`tools.merge_predicate`, the same helper the merge-guard hook enforces): `allowed: true`, no failed checks, "REVIEW verdict fresh: head_sha trailer matches PR head commit", "single-owner MERGE: merge actor holds the issue lease". Head resolved git-first via `resolve_pr_head_sha` to `7f663d64685b980e4a5fd3f7586f93736255117f`. Ruff check and format clean in the worktree (1273 files).

**What this merge rests on:**

- Focused acceptance: **141 passed / 2 skipped / 0 failed** on the 7 baseline files, serial `-n0`. Baseline before this PR was 18 failed / 120 passed.
- **Six negative controls, all independently reproduced by a reviewer** rather than credited from this body — including the #2552 control at exactly **17 failed / 40 passed** in a 2/13/2 distribution, node-for-node against the Step 0 baseline.
- **Two independent APPROVED reviews.** The delta re-review is recorded at exact head `7f663d646` with **0 blockers / 0 tech debt**.
- **Failure-set diff against a same-environment control** (`origin/main` in a worktree): zero regressions, and the #2553 stale-assertion fix visible as the single removed failure.
- Three anti-criteria confirmed: `bridge/` untouched, `agent/agent_session_queue.py` untouched, `data/catchup-disabled` intact at 0 bytes / mtime Jul 22 13:45.

**What this merge does NOT rest on — stated plainly rather than omitted:**

**Full-suite verification is not part of this basis.** Three full-suite attempts from the worktree wedged at 99% (xdist worker death, controller hung at 0% CPU, no summary emitted). That wedge is **pre-existing** — reproduced across unrelated branches (#2536 recorded 4/4) and observed before this diff existed — and one of the three wedged runs had this PR's autouse fixture reduced to a bare `yield` via its own `CATCHUP_FLAG_ISOLATION=0` escape hatch, so the fixture's per-test behavior is not the mechanism.

Its attribution is also confounded: worktrees on this host run **Python 3.13.2** while the main checkout runs **3.14.3**, so every wedged run and every clean control differed by interpreter as well as by code. That confound was found while deconfounding this PR and is why the earlier "control implicates the diff" reading was withdrawn.

Three issues filed from this lane's findings, all tracking work deliberately kept out of this PR:

- **#2572** — session worktrees run Python 3.13.2 vs main's 3.14.3; confounds every worktree-vs-main comparison.
- **#2573** — checkout portability: ≥18 test outcomes flip on machine-local gitignored files, so worktree and main-checkout failure counts are incommensurable; the 35-failure baseline is main-checkout-only and must be labeled so.
- **#2574** — the tail wedge itself, including the xdist attribution trap (a dead worker is reported against its *first collected item*, not the test it was running — this misled two investigations today).

Also deferred and tracked: **#2563** (the `_extract_agent_session_fields` two-contract conflict and its 50-field gap) and **#2473** (the production kill-switch flag's fate).

Issue #2567, opened 2026-08-06
~/.claude/hooks is still a legacy directory symlink into the repo: undetected, unmigrated, and able to delete tracked source


valorengels:
`~/.claude/skills` was deliberately migrated off the legacy directory-symlink layout. `~/.claude/hooks` was not, and nothing detects that it is still in that layout. The result is that any code writing to or deleting from `~/.claude/hooks/` can silently operate on this repo's tracked source tree.

## Incident evidence

This is not theoretical. During the #2521 build, a migration that enumerated `~/.claude/hooks/` and removed what it believed were orphaned user-level files **deleted 36 tracked repo files plus their bytecode**. They were recovered from git and the migration's own `.bak`, and the tree was verified byte-identical to `HEAD`. The migration was correct about its own logic; it was wrong about what directory it was standing in.

```
$ ls -ld ~/.claude/hooks
lrwxr-xr-x  1 valorengels  staff  39 Feb  3  2026
  /Users/valorengels/.claude/hooks -> /Users/valorengels/src/ai/.claude/hooks
```

That symlink also invalidated #2521 entirely (closed as premise-disproved) and caused a documented reasoning error in its plan: a `stat` showing the user path and the repo path sharing inode `5917700` was read as evidence of a *hardlink*, supporting a "deleting the user copy leaves the repo source intact" conclusion. Same inode, but the cause was the symlink — the two paths were the same file, and the conclusion was false. Any future audit of `~/.claude/hooks/` is exposed to exactly this inference.

## Why the asymmetry exists

`sync_claude_dirs` explicitly handles the legacy layout for skills (`scripts/update/hardlinks.py:219-227`):

```python
user_skills = user_claude / "skills"
if user_skills.is_symlink():
    user_skills.unlink()
    ...  # migrated from dir-symlink to hardlinks
```

There is no equivalent for `hooks`. `sync_user_hooks` (`hardlinks.py:873-943`) writes into `hooks_root = user_claude / "hooks"` via `_ensure_hardlink` without ever checking whether that path is a symlink, so on this layout it hardlinks the repo's files on top of themselves — a silent no-op that looks like success.

## Consequences beyond the deletion hazard

[...]

PR #195, opened 2026-02-26
SDLC user-level hooks + PR #185 tech debt cleanup

valorengels:
## Summary
- Extract shared `sdlc_context.py` module — single source of truth for SDLC context detection, imported by all 3 hooks instead of 3x duplication
- Add `sync_user_hooks()` to `scripts/update/hardlinks.py` — copies hooks to `~/.claude/hooks/sdlc/` and merges entries into `~/.claude/settings.json` with deduplication
- Fix PR #185 tech debt: replace hardcoded `MAX_AUTO_CONTINUES` with `effective_max` in log guards, refactor classifier path to call `_enqueue_continuation()` instead of duplicating coaching+enqueue logic
- Fix test class typo: `TestMaxAutoContiuesConstants` → `TestMaxAutoContinuesConstants`

## Changes
- `.claude/hooks/sdlc/sdlc_context.py` — new shared module
- `.claude/hooks/sdlc/validate_commit_message.py` — refactored to import from sdlc_context
- `.claude/hooks/sdlc/sdlc_reminder.py` — refactored to import from sdlc_context
- `.claude/hooks/sdlc/validate_sdlc_on_stop.py` — refactored to import from sdlc_context
- `scripts/update/hardlinks.py` — new `sync_user_hooks()`, `_merge_hook_settings()`, integrated into `sync_claude_dirs()`
- `agent/job_queue.py` — `effective_max` in log guards, classifier path uses `_enqueue_continuation()`
- `tests/test_stage_aware_auto_continue.py` — typo fix
- `docs/features/sdlc-enforcement.md` — user-level deployment section
- `docs/features/README.md` — updated index entry

## Testing
- [x] 32/32 stage-aware auto-continue tests passing
- [x] 16/16 auto-continue tests passing
- [x] ruff check passing (pre-existing E501 on unrelated line)
- [x] black format check passing

## Documentation
- [x] sdlc-enforcement.md updated with user-level deployment section
- [x] README.md index updated

Closes #183

PR #2569, opened 2026-08-06
Restore directory-granular deployment for global-scope hook scripts (#2561)

### LINKED ISSUE #2561: sync_user_hooks never syncs a global hook script's runtime import siblings, so sdlc_context.py is missing on a fresh machine
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/sync-global-hook-import-siblings.md

Split out of #2521 during planning (spike-3).

## Context

`sync_user_hooks` (`scripts/update/hardlinks.py:873-943`) is the only thing that populates `~/.claude/hooks/` under the manifest-driven regime. It filters the manifest to `scope == "global"` declarations and hardlinks exactly one file per declaration:

```python
for decl in global_decls:
    src_file = project_dir / ".claude" / "hooks" / decl.script
    dst_file = hooks_root / decl.script
    ...
    _ensure_hardlink(src_file, dst_file, dst_file.parent, result)
```

There are three global declarations today (`manifest.toml:274-306`), all under `sdlc/`.

## Problem

All three of those scripts import a shared helper module that is **not itself a declaration**:

- `sdlc/validate_commit_message_sdlc.py:35` — `from sdlc_context import allow, block, read_stdin`
- `sdlc/validate_sdlc_on_stop.py:34` — `from sdlc_context import is_sdlc_context, read_stdin`
- `sdlc/sdlc_reminder.py:29` — `from sdlc_context import is_sdlc_context, read_stdin`

`sdlc/sdlc_context.py` has no `[[hook]]` entry of its own (correctly — it registers no hook), so `sync_user_hooks` never hardlinks it. On a machine where it is absent from `~/.claude/hooks/sdlc/`, all three global hooks die at import with `ModuleNotFoundError` on every Bash call, every Stop, and every Write/Edit in **every foreign repo** — the exact blast radius #2503 was filed about.

This is latent rather than active today because the pre-manifest sync path already deposited the file on existing fleet machines (verified on this host: `~/.claude/hooks/sdlc/sdlc_context.py` and the repo source share inode `1526758`). The exposure is a genuinely fresh machine, or any machine where the user tree is ever rebuilt from scratch — including after #2521's prune, if its keep-set were ever narrowed.

### Current behavior

A global-scope hook script's runtime import siblings are not synced to `~/.claude/hooks/`. The generator's implicit contract is "a declaration is a single self-contained file", and that contract is already violated by the only three declarations that exist.

### Desired outcome

`~/.claude/hooks/` contains everything a declared global script needs in order to import cleanly, on a machine that has never run the pre-manifest sync path.

## Recon Summary

**Confirmed**

- `sync_user_hooks` hardlinks `decl.script` and nothing else (`hardlinks.py:922-936`). There is no sibling-file, package, or directory handling anywhere in the global sync path.
- All three `scope = "global"` declarations import `sdlc_context`, and `sdlc_context.py` carries no declaration. Verified by reading all four files.
- The file is present on this host only as a pre-manifest artifact. Nothing in the current code would restore it.

**Revised**

- This is not a bug in `sdlc_context.py`'s design. Keeping shared helpers out of the manifest is right — they register no hook. The gap is that the *generator* assumes declarations are self-contained.

**Dropped**

- A transitive AST import walk. #2521 rules that class of solution out for its own prune, and the same reasoning applies here: the real edge set is tiny and enumerable.

## Design notes for `/do-plan`

Two shapes are worth weighing, and the plan should pick one rather than hedge:

1. **An explicit `also_sync` / companion-files list** — either a new optional manifest key on the declaration, or a module-level constant beside `RENAMED_REMOVALS`. Cheap, explicit, matches the repo's established "static allowlist over inference" posture (#2503 critique, #2521 design constraint). Cost: a second place to remember when adding a helper.

2. **Sync the whole declaring directory** — if any global declaration lives under `sdlc/`, hardlink every `.py` under `.claude/hooks/sdlc/`. Self-maintaining, and `sdlc/` exists precisely to hold the fork scripts. Cost: it syncs by proximity rather than by need, and would silently start shipping any file someone drops in that directory.

Also worth deciding: should a global script that fails to import be *detected* rather than only prevented? A cheap import smoke-check of each synced global script at the end of `sync_user_hooks` would have caught this class directly.

## Acceptance criteria

- [ ] A fresh `~/.claude/hooks/` built solely by `sync_user_hooks` contains everything the three global scripts need to import — asserted by test, not by inspection
- [ ] Each of the three global scripts imports cleanly under the resolved global interpreter after a sync into an empty temp home
- [ ] The mechanism is explicit (static declaration or an enumerated directory), not a transitive import walk
- [ ] Adding a fourth global declaration with a new helper does not silently reintroduce the gap — the failure is caught by a test, not discovered in a foreign repo

## Related

- #2521 — prunes the pre-manifest residue from `~/.claude/hooks/`; its keep-set names `sdlc/sdlc_context.py` explicitly precisely because of this gap. That plan's spike-3 surfaced this issue.
- #2503 / PR #2522 — the "global hooks are dead in every foreign repo" failure mode this would reproduce on a fresh machine.
- #2435 / PR #2453 — introduced the manifest generator whose self-contained-declaration assumption is the root cause.

valorengels:
## Summary

All three `scope = "global"` hook scripts import a shared sibling helper at module scope (`from sdlc_context import ...`). `sdlc/sdlc_context.py` registers no hook, so it has no `[[hook]]` declaration — and `sync_user_hooks` hardlinked exactly one file per declaration. The helper was therefore never deployed to `~/.claude/hooks/sdlc/`, and on any machine whose `~/.claude/hooks/` is a real directory, all three global fork hooks died at import with `ModuleNotFoundError` on every Bash call, every Stop, and every Write/Edit — **in every foreign repo**.

This is a regression, not an original omission. PR #195 deployed the `sdlc/` directory by `glob("*.py")` and extracted the shared helper in the same change; PR #2453 correctly made the manifest authoritative for *registration* but silently narrowed *deployment* from the directory to the declared file. This PR restores directory-granular deployment.

The break was invisible on the development machine only because its `~/.claude/hooks` is a legacy symlink into the repo (#2567), so the helper resolved to the repo's own copy.

## Changes

- **`scripts/update/hardlinks.py::sync_user_hooks`** — the deployment directory set is now *derived* from the global declarations (`{PurePosixPath(d.script).parent for d in global_decls}`) and every `*.py` in each such directory is hardlinked. Never hardcodes a directory name, so a future global hook in a new directory — or a helper added beside an existing one — is covered with no manifest change. A declaration at the hooks root deploys only its own file (globbing the root would sweep every project-scope script into the user tree). Registration is unchanged and stays declaration-granular: helpers are deployed but never registered. Deployment is additive — nothing is deleted, since that needs the symlink guard #2567 owns.
- **`scripts/update/run.py`** — prints a greppable fleet signal after `sync_claude_dirs`: `hooks: sdlc_context deployed` or `hooks: MISSING sdlc_context (see #2561)`. This dev machine cannot verify the fix for real (its `~/.claude/hooks` is the #2567 symlink, so a local `/update` shows success either way), so this line is how the first real fleet machine reports whether the fix actually took.
- **`tests/unit/test_update_hardlinks.py`** — 9 new tests, including an end-to-end import smoke-check that executes each declared global script under the resolved interpreter and asserts on stderr *content*, not exit status (a hook that exits 0 while printing an import error is exactly the silent failure being fixed). The child env is built explicitly with all `CLAUDE_*` keys stripped so `is_sdlc_context()` cannot reach production Redis from a unit test.
- **`docs/features/hook-manifest.md`** — documents the registration-vs-deployment split and records the #2567 ordering constraint.

The smoke-check is the part that generalizes: it fails for *any* future global script with an unsatisfiable import, regardless of how deployment is implemented.

## Ordering constraint

**This must land before #2567.** #2567 migrates `~/.claude/hooks` from a symlink to a real directory. If it landed first, that real directory would be built by the pre-fix `sync_user_hooks` — three scripts, no helper — unmasking this bug on the development machine too. A working machine would become a broken one.

## Verification

`scripts/validate_build.py`: **11 PASS, 0 FAIL, 0 SKIP**.

| Check | Result |
|---|---|
| Focused sync tests (`test_update_hardlinks`, `test_hook_interpreter`) | pass |
| Existing hook migration/audit tests | pass |
| `ruff check .` / `ruff format --check .` (repo-wide) | pass |
| Helper deployed, not just declared (`-k sdlc_context`) | pass |
| Global scripts import cleanly after sync (`-k import_smoke`) | pass |
| Anti-criterion: no hardcoded directory name in the sync body | pass |
| Anti-criterion: no deletion added under user hooks | pass |
| Anti-criterion: no AST import walk | pass |

101 tests pass across the four touched test modules. **Red-state proof:** stashing only `hardlinks.py` and re-running the suite fails exactly the 5 new tests that assert the fix — they are not vacuous.

One plan correction landed on `main` during the build (f156a9ce2): three Verification rows escaped a pipe as `re.S\|re.M` for the Markdown table, and the verification runner splits rows on `|` without unescaping, truncating the command mid-cell and reporting a bogus `exit=2`. Rewritten as `re.DOTALL+re.MULTILINE` so the rows actually execute. This was the fourth grep/table trap recorded in that plan section.

Closes #2561


valorengels:
## Review: Approved

Restores directory-granular deployment for global-scope hook scripts. The fix is small, the mechanism is derived rather than hardcoded, and the regression guard generalizes beyond the specific helper it was written for.

**Multi-judge consensus (K=2, `any-blocker-wins`):** code-quality judge APPROVED (0 blockers, 0 tech debt); risk judge APPROVED (0 blockers). No disagreement.

### Independently reproduced (not taken from the PR body)

Per the "a number the PR claims is a claim, not evidence" rule, every number below was re-measured in a fresh worktree at head `8e45200`:

- **101 passed** across `test_update_hardlinks.py`, `test_hook_interpreter.py`, `test_hook_migration.py`, `test_hooks_audit.py` (`-n0`). Matches the PR body.
- **Red-state proof reproduced.** Reverting only `scripts/update/hardlinks.py` to `origin/main` fails **exactly 5** tests — `deploys_sdlc_context_helper`, `global_hooks_import_smoke`, `deployment_dir_is_derived_not_hardcoded`, `sdlc_context_is_deployed_but_unregistered`, `missing_declared_source_does_not_abort_siblings`. The new tests are not vacuous.
- **`validate_build.py`: 11 PASS, 0 FAIL, 0 SKIP.** All four anti-criterion rows re-run by hand and confirmed independently.
- `ruff check` / `ruff format --check` clean on all three changed Python files.

**Reproduction note for whoever re-runs this:** running `validate_build.py` against the plan copy *inside the PR worktree* reports 3 bogus FAILs (`exit=2` on the three Python one-liner rows). That is the branch carrying a pre-`f156a9ce2` copy of the plan — the pipe-escaping fix landed on `main` only, per the plans-on-main convention. Run it against the `main` copy of the plan. Not a defect in this PR, but it will waste someone's afternoon otherwise.

### Verified by reading, not by assertion

- `helper_dirs` is derived (`{PurePosixPath(d.script).parent}`), never hardcoded — confirmed by the anti-criterion row *and* by reading the loop body.
- The `seen_scripts` dedup key `str(rel_dir / src_file.name)` matches the `decl.script` string format, so declared files are not re-linked.
- The hooks-root guard (`str(parent) != "."`) is real and covered.
- `sync_claude_dirs` genuinely calls `sync_user_hooks` at `hardlinks.py:272`, so the `run.py` fleet signal is placed after the code it reports on. `log()` accepts `always=` (`run.py:172`) and `UpdateResult.warnings` exists (`run.py:164`).
- `.claude/hooks/sdlc/` contains exactly 4 `.py` files and the manifest declares **zero** project-scope scripts under `sdlc/` — so the directory glob ships nothing unintended today.
- `sdlc_context.py`'s docstring claim ("STANDALONE module deployed to `~/.claude/hooks/sdlc/` by the update system") is true again after this change. The plan asked for this to be verified rather than assumed; it was.
- `#2567` confirmed **OPEN**; `f156a9ce2` confirmed present on `main`.

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — PASS — *all 5 Success Criteria satisfied; plan checkboxes ticked on `main` (`cda9f657b`)*
- **2. No-Gos from plan — none violated** — PASS — *both No-Gos are `[SEPARATE-SLUG #2567]` (symlink migration, stale-file deletion); the diff adds no `unlink`/`rmtree` under the user tree, asserted by an anti-criterion row and by `test_sync_user_hooks_deletes_nothing_under_user_hooks`*
- **3. New `except Exception` blocks — each has logger/raise/swallow-ok** — N/A — *none added; the existing `OSError` posture in `_ensure_hardlink` is unchanged*
- **4. New integration tests — exercise serialization boundary (not in-memory only)** — PASS — *the smoke test runs each script as a real subprocess under the resolved interpreter against a real temp-`HOME` filesystem tree*
- **5. Plan internal consistency — spike findings match task steps** — PASS — *spike-2's mandatory `CLAUDE_*`-stripping mitigation is implemented verbatim*
- **7. New public APIs — docstrings present** — N/A — *no new public API; one internal loop*
- **8. Breaking changes — migration path documented** — N/A — *additive deployment; self-heals on next `/update`*
- **9. Tests added for new behavior** — PASS — *9 new tests*
- **10. Tests cover the failure path (not just happy path)** — PASS — *missing declared source, empty global scope, hooks-root case, re-run no-op, nothing-deleted*
- **11. UI changes (if any) — screenshot captured** — N/A — *no UI files in the diff*
- **12. Docs updated for user-facing changes** — PASS — *`hook-manifest.md` gains the registration-vs-deployment section and the #2567 ordering constraint; inline loop comment present*

## Rubric

- [x] **1. Plan vs. implementation match** — pass — *every Task 1 bullet delivered, including the derived-directory requirement and the root-declaration carve-out*
- [x] **2. New code quality** — pass — *typed (`set[PurePosixPath]`), derived not hardcoded, and the comment explains why the directory is the deployment unit so the next refactor does not re-narrow it*
- [x] **3. Test coverage** — pass — *red-state proven at exactly 5 failures*
- [x] **4. Regression risk to existing callers** — pass — *only caller is `sync_claude_dirs`; the change is purely additive to the deployed set, registration path untouched, `test_sync_user_hooks_empty_global_scope_deploys_nothing_and_still_removes` guards the removal pass*
- [x] **5. Data integrity** — n/a — *no schema, no migration, no Popoto model*
- [x] **6. Security** — pass — *no `eval`/`exec`/secrets; the one new `subprocess` call is in a test with an explicitly-constructed env*
- [x] **7. Documentation accuracy** — pass
- [x] **8. PR body accuracy** — pass — *101 tests, 11 PASS, and the 5-test red state all reproduced independently*
- [x] **9. Disclosed deferrals** — pass
- [x] **10. Follow-up claims verified** — pass — *#2567 is OPEN*

### Blockers
- None

### Tech Debt
- None

### Nits
- None

### Miscellaneous
- None

- **Deletion of stale files under `~/.claude/hooks/` is out of scope; it needs the symlink guard #2567 owns** — tracked by #2567 (OPEN). Correctly deferred: deleting "user-tree files with no source" on a symlinked machine would operate on tracked repo source.
- **`~/.claude/hooks` symlink-to-real-directory migration, and the ordering constraint that this PR must land first** — tracked by #2567 (OPEN). The ordering claim is sound and I verified the mechanism: `_ensure_hardlink` takes the inode-match skip branch on a symlinked machine, so landing #2567 first would have built a real directory from the pre-fix per-declaration deployment and broken the one machine that currently works.
- **This development machine cannot verify the fix for real** — explicit out-of-scope exclusion with a shipped mitigation rather than a bare admission: the greppable `hooks: sdlc_context deployed` / `hooks: MISSING sdlc_context (see #2561)` line in `run.py` is how the first real fleet machine reports the truth. No tracking issue required.

### Legacy Cruft
- None. The change restores PR #195's behavior rather than adding a parallel path, and leaves no dual implementation or compatibility shim behind.

### Deploy note
`scripts/update/` only — no bridge, worker, or agent code touched, so no service restart is required. Delivery is the next `/update` on each machine.

Issue #1997, opened 2026-07-10
Centralize duplicated machine-name resolution into config/machine.py


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/centralize_machine_name_resolution.md

## Recon Summary

Verified against main (2026-07-10). The five copies are **not byte-identical** — they diverge on four axes, so a naive merge would change behavior at call sites.

**Confirmed (all five exist):**
- `ui/data/machine.py::get_machine_name` — scutil, no timeout, `""` on exception, no fallback.
- `bridge/update.py::_get_machine_name` (L88) — scutil `timeout=5`, falls back to `platform.node().split(".")[0]` on failure. Used at L169/345/459 for Telegram/log display only (cosmetic).
- `tools/google_workspace/auth.py::_get_machine_name` (L98) — scutil `timeout=5`, **filesystem-safe transform** (`.lower().replace(" ", "-")`), fallback `platform.node().split(".")[0].lower()`. Feeds per-machine token filename — MUST stay non-empty.
- `scripts/update/readme_check.py::_get_machine_name` (L56) — scutil, **no timeout, no try/except (raises on failure)**, no fallback. Caller guards `if not machine_name`.
- `monitoring/sentry_config.py::_get_machine_name` (L99) — scutil, no timeout, `""` on exception; self-contained per #1834 to avoid monitoring->ui inversion.

**Divergence axes:** (1) timeout=5 vs none; (2) platform.node() fallback present in bridge+google_workspace only; (3) filesystem-safe slug transform in google_workspace only; (4) readme_check RAISES on failure vs others return `""`.

**Ownership-resolution copies:** `ui/data/machine.py::get_machine_project_keys` (list, `[]` on failure) and `monitoring/sentry_config.py::_owned_project_key` (first-or-None, with an explicit **empty-machine fail-to-development guard** from #1834, test-pinned).

**Two projects.json sources (important):** ui + monitoring read the vault `~/Desktop/Valor/projects.json`; `readme_check.py` reads the repo-local `config/projects.json` with a different output shape (working_directory). So readme_check centralizes only `get_machine_name()`, not the project-key logic.

[...]

Issue #1834, opened 2026-07-01
Sentry: gate init so dev/test runs stop reporting as production


tomcounsell:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/sentry-gate-dev-test-init.md

## Problem

Local `pytest` runs on dev machines are reporting synthetic errors into the **production** Sentry project (`yudame/valor`). 13 of 20 current unresolved `valor` issues originate from `Mac.local` (a dev machine), tagged `environment: production`, most in a single `10:02–10:03` test batch. Examples of test fixtures leaking as prod errors:

- `VALOR-BW` — `RuntimeError: crash` emitted directly from `tests/unit/test_worker_supervisor.py:_crashing`
- `VALOR-A3` — `[granite-exit-anomaly] session=granite-anomaly-001` (test session id)
- `VALOR-B4` — `Dead letter replay failed for chat 12345` (test fixture chat)
- `VALOR-AW/AV/AT` — `[executor-guard] Refusing to start session with None ...` (guards firing under test, working as designed)
- `VALOR-BX` — `KeyError: 'subscription_type'` in `ui/app.py:dashboard_json` (surfaced by a test)

## Root cause

Sentry is initialized only in `bridge/telegram_bridge.py:74-84`:

```python
_sentry_dsn = os.getenv("SENTRY_DSN")
if _sentry_dsn:
    sentry_sdk.init(
        dsn=_sentry_dsn,
        release=...,
        traces_sample_rate=0.1,
        environment=os.getenv("SENTRY_ENVIRONMENT", "production"),
        before_send=_sentry_before_send,
    )
```

`environment` defaults to `"production"` with no test-mode or machine guard. Any process that imports the bridge with `SENTRY_DSN` set — including a local `pytest` run — captures events into the production project.

## Acceptance criteria

- Sentry init is suppressed (or routed to a non-production environment) when running under pytest — detect via `PYTEST_CURRENT_TEST` env var or equivalent.
- Optionally: gate real reporting to designated bridge machines only (this machine `Tom's MacBook Pro` / dev machines should not report as `production`). Cross-reference the single-machine-ownership model in `docs/features/single-machine-ownership.md`.
- A test asserts that with `PYTEST_CURRENT_TEST` set, `sentry_sdk.init` is not called with `environment="production"` (or not called at all).
[...]

PR #2578, opened 2026-08-06
Validate that a project's owning machine exists (#2541)

### LINKED ISSUE #2541: projects.json ownership validation passes when the owning machine does not exist
> **Context.** `projects.json` (private, at `~/Desktop/Valor/projects.json`) is the multi-project routing config. Per [CLAUDE.md's "Single-Machine Ownership (Strict)"](../blob/main/CLAUDE.md), `projects.<key>.machine` is the source of truth for which host answers a project's Telegram DMs, groups, and email; every other identifier inherits ownership from its project. The rule is enforced by `bridge/config_validation.py::validate_projects_config` and gated by `scripts/update/run.py` Step 4.6, which blocks the bridge restart on a malformed config.

## Problem

`validate_projects_config` checks that each identifier has **exactly one** owner. It never checks that the owner **exists**. A project can name a machine that was decommissioned, renamed, or never existed, and validation passes — so Step 4.6 greenlights the bridge restart on a config that routes real traffic to nobody.

This is fail-open in a subsystem documented as strict. The failure is silent: no host claims the project, so its messages are simply never picked up. There is no error, no warning, and no log line naming the unowned project.

### Current behavior

Observed on Tom's MacBook Air, 2026-08-06, against the live config (last modified 2026-07-15):

- 20 projects total.
- **19** assign `machine` to `Tom's MacBook Pro` — a laptop that has since been **wiped and sold**. 18 use a straight apostrophe (`U+0027`), 1 (`cuttlefish`) uses a curly one (`U+2019`).
- 1 (`valor`) assigns `Tom's MacBook Air` (curly), a host that runs no bridge.
- **Zero** entries reference `Valor the Cowboy`, `Valor the Captain`, or `Valor the Bald`, the hosts that actually constitute the fleet.

`validate_projects_config(config)` returns without raising against exactly this data.

The apostrophe split is a second, independent trap: `scutil --get ComputerName` returns the curly `U+2019` form on macOS (confirmed by `od -c` on this host), so a straight-apostrophe owner string cannot match its own machine even when that machine is alive.

### Desired outcome

A config whose owners do not resolve to real machines should not pass validation silently. At minimum the validator should surface unresolvable owners; ideally Step 4.6 treats an unowned project as the routing defect it is.

## Definitions

| Term | Definition |
|------|------------|
| **Ownership** | `projects.<key>.machine` names the single host that answers that project's traffic. See [`docs/features/single-machine-ownership.md`](../blob/main/docs/features/single-machine-ownership.md). |
| **Step 4.6** | The `scripts/update/run.py` gate that validates `projects.json` before restarting the bridge; on failure the running bridge keeps serving the last-validated config. |
| **Unowned project** | A project whose `machine` value matches no live host, so no bridge ever claims its identifiers. |

## Solution sketch

The root question is what "exists" should mean, and that is worth deciding before implementing:

- **Closed set.** Validate `machine` against an explicit roster of known fleet hostnames. Strongest check, but adds a list that must be maintained and will itself drift.
- **Self-check only.** Each host verifies that the projects claiming *it* resolve, and warns about owners it does not recognize. Cannot see the whole fleet from one machine, but needs no new state.
- **Liveness-derived.** Treat a machine as existing if it has checked in recently. Catches decommissioned hosts automatically, but couples config validation to runtime state and would flag a host that is merely powered off.

Independent of that choice, two narrower fixes stand alone:

1. **Normalize apostrophes** (`U+2019` ↔ `U+0027`) on both sides of the comparison, so an encoding mismatch cannot silently unown a project.
2. **Report ownership on every `/update`** — print which projects this host claims and which owners it does not recognize. The current silence is what let a three-week-stale config go unnoticed.

Note the observed config drift is a *data* problem for the operator to fix by reassigning those 19 projects to live hosts. This issue is about the *validator* that failed to notice.

## Acceptance criteria

- [ ] A project naming a nonexistent machine no longer passes validation silently.
- [ ] Apostrophe encoding cannot cause a live host to fail to match its own name.
- [ ] `/update` reports this host's claimed projects and any unrecognized owners.
- [ ] Test coverage for: nonexistent owner, straight-vs-curly mismatch, and a valid config (negative control, so the guard is proven able to fail).

## Downstream context

This issue will be consumed by `/do-plan` to produce a plan document at `docs/plans/{slug}.md`. Found while verifying a fleet rollout for [#2518](../issues/2518), where the deploy could reach no remote machine and the local ownership gate passed against three-week-stale data.

tomcounsell:
Closes #2541

## Problem

`validate_projects_config` checks that each bridge identifier has **exactly one** owner. It never checks that the owner **exists**. A project can name a machine that was decommissioned, renamed, or never existed, and validation passes, so `scripts/update/run.py` Step 4.6 greenlights the bridge restart on a config that routes real traffic to nobody.

Reproduced against `2e517d3b`:

```python
cfg = {"projects": {"ghost": {"machine": "A Machine That Never Existed",
                              "telegram": {"groups": {"Eng: Ghost": {}}},
                              "email": {"contacts": ["a@b.com"]}}},
       "dms": {"whitelist": [{"id": 1, "name": "x", "project": "ghost"}]}}
validate_projects_config(cfg)   # returns cleanly
```

And the apostrophe split, against the live config on this host:

```python
get_machine_project_keys(machine="Tom's MacBook Air")   # []      (U+0027)
get_machine_project_keys(machine="Tom’s MacBook Air")   # 20 keys (U+2019)
```

`scutil --get ComputerName` returns the U+2019 form, so a straight-apostrophe owner string cannot match its own machine even when that machine is alive.

## The design decision

The issue named three candidate meanings for "exists" and asked for a decision before implementing. This PR takes the **closed set**, declared in the config:

```json
"machines": {
  "Valor the Cowboy": "Primary bridge host",
  "Valor the Captain": "Second bridge host"
}
```

- **Self-check only** was rejected because one host cannot see the fleet, and deriving a roster from the config's own owner set would only ever confirm itself.
- **Liveness-derived** was rejected because it couples config validation to runtime state and would flag a host that is merely powered off.
- The roster's maintenance cost is one line per machine, in the file the operator already edits when adding a machine, and `/update` reports it every run so it cannot drift unnoticed.

**When no roster is declared the check is skipped.** That keeps this from hard-failing every machine in the fleet the moment it lands; `/update` instead prints, on every run, that owner names are unverifiable and how to fix it. Declaring the roster is the operator's opt-in to strictness.

## Changes

| File | Change |
|---|---|
| `config/machine.py` | `normalize_machine_name()`: strip, casefold, fold `’ ‘ ʼ ´ \`` to `'`. `get_machine_project_keys` routes through it. The #1834 empty-machine guard is unchanged. |
| `bridge/config_validation.py` | `validate_machine_roster`, `declared_machines`, `project_owners`, `summarize_ownership`. Roster check joins the aggregated suite. |
| `ui/data/machine.py`, `ui/data/memories.py` | The two drifted raw-`.lower()` ownership comparisons #1997 missed; `memories.py` now just calls `get_machine_project_keys()`. |
| `scripts/update/verify.py` | `ToolCheck.detail`; `check_projects_json` returns the ownership summary on pass **and** fail. |
| `scripts/update/run.py` | Step 4.6 logs the summary with `always=True`. |
| `config/projects.example.json` | Roster block plus `_doc` entry. |
| `docs/features/single-machine-ownership.md` | Roster, reporting, normalization, two new failure-mode rows, adding-a-machine step. |

## Acceptance criteria

- [x] A project naming a nonexistent machine no longer passes validation silently — hard error with a roster, reported line without one.
- [x] Apostrophe encoding cannot cause a live host to fail to match its own name.
- [x] `/update` reports this host's claimed projects and the config's other owners.
- [x] Test coverage for nonexistent owner, straight-vs-curly mismatch, and a valid config (negative control).

## Verification

```
tests/unit/test_dm_whitelist_validation.py  49 passed
tests/unit/test_config_machine.py           18 passed
tests/unit/test_ui_data_memories.py + test_reflection_machine_filter.py + above: 117 passed
```

End-to-end against the live config on this host:

```
available: True
version: valid (0 DM contacts, 0 groups, 0 email patterns)
ownership: this host (Tom’s MacBook Air) claims 20 project(s): ai-skills, cuttlefish, ...
ownership: no 'machines' roster declared — owner names are unverifiable; ...
```

## Note on the reported data drift

The issue observed 19 projects owned by `Tom's MacBook Pro`. The live config has since been reconciled by the operator: all 20 projects now name `Tom’s MacBook Air`, and `_doc.machine_this` reads "sole remaining Mac; MacBook Pro was wiped and sold 2026-08". The data symptom is gone; the validator defect this PR fixes is not, and reproduces on current main.

## DOCS

`docs/features/single-machine-ownership.md` updated: new "The fleet roster" and "Ownership reporting" sections, `normalize_machine_name` added to the machine-identity function list, roster check added to "What the validator catches", two rows added to failure modes, roster step added to "Adding a new machine", test-coverage paragraph rewritten.

### DIFF
--- a/bridge/config_validation.py
+++ b/bridge/config_validation.py
@@ -13,6 +13,8 @@ by the update script as a green-light gate before service restart.
 
 from __future__ import annotations
 
+from config.machine import normalize_machine_name
+
 
 class ConfigValidationError(RuntimeError):
     """Raised when projects.json fails validation. Update script blocks restart."""
@@ -390,6 +392,138 @@ async def validate_bot_live_flags(config: dict, resolver) -> tuple[set[int], str
     return (quarantine_ids, detail)
 
 
+def declared_machines(config: dict) -> dict[str, str]:
+    """Return the declared fleet roster as ``{normalized_name: as_written}``.
+
+    The roster is the top-level ``machines`` block of ``projects.json``. It
+    accepts either a list of ComputerNames or a dict keyed by ComputerName
+    (values are free-form operator notes), so an operator can annotate a host
+    without changing the shape the validator reads.
+
+    An absent or malformed block yields ``{}``, which
+    :func:`validate_machine_roster` reads as "no roster declared".
+    """
+    raw = config.get("machines")
+    if isinstance(raw, dict):
+        names = list(raw.keys())
+    elif isinstance(raw, list):
+        names = [n for n in raw if isinstance(n, str)]
+    else:
+        return {}
+    return {normalize_machine_name(n): n for n in names if normalize_machine_name(n)}
+
+
+def project_owners(config: dict) -> dict[str, list[str]]:
+    """Return ``{normalized_owner_name: [project_key, ...]}`` for every owned project.
+
+    Projects with a missing or blank ``machine`` are omitted; the per-shape
+    validators already fail those when they declare any bridge identifier.
+    """
+    owners: dict[str, list[str]] = {}
+    for proj_key, proj_cfg in config.get("projects", {}).items():
+        if not isinstance(proj_cfg, dict):
+            continue
+        normalized = normalize_machine_name(proj_cfg.get("machine"))
+        if normalized:
+            owners.setdefault(normalized, []).append(proj_key)
+    return owners
+
+
+def validate_machine_roster(config: dict) -> None:
+    """Enforce that every project owner names a machine the fleet actually has.
+
+    The per-shape validators prove each identifier has *exactly one* owner. They
+    never prove that owner **exists**, so a project can name a host that was
+    decommissioned, renamed, or typo'd and validation passes — leaving the
+    project unowned, its traffic picked up by nobody, and no error anywhere
+    (issue #2541).
+
+    "Exists" is answered by the config itself: the top-level ``machines`` roster
+    lists the fleet's ComputerNames. Ownership is checked against that roster
+    under :func:`config.machine.normalize_machine_name`, so a curly-vs-straight
+    apostrophe cannot unown a live host.
+
+    When no roster is declared the check is skipped rather than guessed at: a
+    single host cannot see the fleet, and inventing a roster from the config's
+    own owner set would only ever confirm itself. The operator learns the roster
+    is missing from the ownership summary that ``/update`` prints on every run
+    (:func:`summarize_ownership`).
+
+    Raises:
+        ConfigValidationError: if any project names a machine outside the roster.
+    """
+    roster = declared_machines(config)
+    if not roster:
+        return
+
+    errors: list[str] = []
+    for normalized, proj_keys in sorted(project_owners(config).items()):
+        if normalized in roster:
+            continue
+        as_written = next(
+            (
+                config["projects"][k].get("machine")
+                for k in proj_keys
+                if isinstance(config["projects"].get(k), dict)
+            ),
+            normalized,
+        )
+        errors.append(
+            f"projects {sorted(proj_keys)} are owned by '{as_written}', which is not in "
+            f"the machines roster {sorted(roster.values())} — those projects are unowned "
+            f"and no bridge will answer their traffic"
+        )
+
+    if errors:
+        raise ConfigValidationError(
+            "projects.json machine ownership failed validation:\n  - " + "\n  - ".join(errors)
+        )
+
+
+def summarize_ownership(config: dict, this_machine: str) -> str:
+    """Render the human-facing ownership report ``/update`` prints on every run.
+
+    Names this host, the projects it claims, the other owners in the config, and
+    whether a roster is declared. Silence is what let a three-week-stale config
+    route twenty projects at a sold laptop; this is the line that makes the
+    current ownership state visible without anyone going looking for it.
+    """
+    roster = declared_machines(config)
+    owners = project_owners(config)
+    me = normalize_machine_name(this_machine)
+
+    lines: list[str] = []
+    mine = sorted(owners.get(me, []))
+    label = this_machine or "unresolved ComputerName"
+    if mine:
+        lines.append(f"this host ({label}) claims {len(mine)} project(s): {', '.join(mine)}")
+    else:
+        lines.append(
+            f"this host ({label}) claims NO projects — every incoming message for "
+            f"this machine's contacts will be handled by some other host, or by nobody"
+        )
+
+    others = {k: v for k, v in owners.items() if k != me}
+    if others:
+        rendered = ", ".join(
+            f"{name} ({len(keys)})" for name, keys in sorted((k, v) for k, v in others.items())
+        )
+        lines.append(f"other owners in config: {rendered}")
+
+    if roster:
+        idle = sorted(roster[n] for n in roster if n not in owners)
+        lines.append(f"machines roster: {', '.join(sorted(roster.values()))}")
+        if idle:
+            lines.append(f"roster hosts owning nothing: {', '.join(idle)}")
+    else:
+        lines.append(
+            "no 'machines' roster declared — owner names are unverifiable; add a "
+            'top-level "machines" list to projects.json to make a decommissioned '
+            "or typo'd owner a hard validation failure (#2541)"
+        )
+    return "\n".join(lines)
+
+
 # NOTE (plan #1924): the per-project ``transport`` block is no longer
 # validated — there is one execution transport (headless ``claude -p``) and
 # no seam. Stale ``transport`` keys still present in a not-yet-migrated
@@ -401,9 +535,9 @@ def validate_projects_config(config: dict) -> None:
     """Run the full bridge-contact ownership validation suite.
 
     Aggregates errors from every shape (DM whitelist, Telegram groups,
-    email routing, bot registry) into a single ConfigValidationError so the
-    operator sees every problem at once instead of fixing them one round-trip
-    at a time.
+    email routing, bot registry, machine roster) into a single
+    ConfigValidationError so the operator sees every problem at once instead of
+    fixing them one round-trip at a time.
     """
     errors: list[str] = []
     for fn in (
@@ -411,6 +545,7 @@ def validate_projects_config(config: dict) -> None:
         validate_telegram_groups,
         validate_email_routing,
         validate_telegram_bots,
+        validate_machine_roster,
     ):
         try:
             fn(config)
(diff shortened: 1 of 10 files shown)


tomcounsell:
## Review: APPROVED

Verified against the merged result of `session/ws-fleet` onto `origin/main` (46d5aee3), not the PR body alone.

**The defect is real and the fix lands where it is claimed.** `validate_projects_config` aggregated four per-shape validators, each proving "exactly one owner", none proving "that owner exists". `validate_machine_roster` joins that tuple in `bridge/config_validation.py:545`, so the roster check reaches Step 4.6 through the same path the operator already trusts.

**The skip-when-no-roster decision is the right call and is honestly implemented.** `validate_machine_roster` returns early on an empty roster rather than deriving one from the config's own owner set, which would only ever confirm itself. That could have been a silent hole; it is not, because `summarize_ownership` emits the "no 'machines' roster declared" line and `scripts/update/run.py:1404` logs the detail with `always=True` on both the pass and fail paths. The `ToolCheck.detail` field is populated on both `return` branches of `check_projects_json`, which is what makes "reported on every run" true rather than aspirational.

**The apostrophe fold is a genuinely separate defect, correctly scoped.** `normalize_machine_name` is stdlib-only in `config/machine.py`, which has no imports that could cycle back through `bridge/`, so the new module-level import in `config_validation.py` is safe. All four ownership comparison sites now route through it, including the two drifted `.lower()` copies in `ui/data/machine.py` and `ui/data/memories.py` that #1997 missed. `memories.py` collapsing to a `get_machine_project_keys()` call removes a copy rather than adding a caller.

**The #1834 empty-machine guard survives.** `normalize_machine_name("")` returns `""`, and `get_machine_project_keys` still short-circuits on an unresolved machine before any file read, so an unknown host cannot match a `"machine": ""` entry. `test_blank_machine_is_not_a_roster_violation` pins that the roster validator stays out of the per-shape validators' territory.

**Test coverage discriminates.** `test_owner_not_in_roster_fails` asserts the failing project is named and the passing one is not, so it would fail if the validator flagged everything. `test_owner_in_roster_passes` is a real negative control. `test_summary_handles_unresolved_computer_name` covers the `get_machine_name() == ""` path that the live code actually hits.

### Verification run

Merged onto current `origin/main` in an isolated worktree with `PYTHONPATH` pinned to that checkout:

```
tests/unit/test_dm_whitelist_validation.py + test_config_machine.py   67 passed
tests/unit/test_ui_data_memories.py + test_reflection_machine_filter.py   50 passed
python -m ruff format --check . 1275 files already formatted
```

### Non-blocking observation

`summarize_ownership` renders the "other owners in config" line from normalized (casefolded) keys, so a second host prints as `valor the captain (8)` rather than its as-written ComputerName, while the roster line just below prints as-written. Cosmetic inconsistency in an operator-facing report; it does not affect any routing decision and the docs example already shows this shape. Not worth holding the merge.

No blockers.

tomcounsell:
## Review: Approved

Verified the behavior claims by executing them, not by reading the diff alone.

### The design decision is the right one, and it lands safely

The issue asked for a decision between closed-set, self-check, and liveness-derived. The PR picks the closed set and defends it in the body. I agree, and the **skip-when-absent** default is what makes it safe to land: with no `machines` block declared, `validate_machine_roster` returns immediately.

I checked the live config on this host before anything else, since this change sits in front of `/update` Step 4.6 and a hard failure there blocks the bridge restart fleet-wide:

```
top-level keys: ['$schema', '_doc', 'defaults', 'dms', 'personas', 'projects']
has machines roster: False
  owner 'Tom’s MacBook Air': 20 projects
```

No roster declared, so this merges as a validation no-op here. Strictness is opt-in, exactly as the body claims.

Worth noting the live data has moved since #2541 was filed: all 20 projects now name `Tom’s MacBook Air` (U+2019), not the 19-on-a-sold-MacBook-Pro split the issue describes. The data problem was fixed separately. This PR is the validator, which is the right scope.

### Behavior verified by execution

| Scenario | Result |
|---|---|
| Ghost owner, no roster declared | returns cleanly (documented skip) |
| Ghost owner, roster declared | **raises** `projects ['ghost'] are owned by 'A Machine That Never Existed', which is not in the machines roster [...]` |
| Straight-apostrophe owner vs curly-apostrophe roster entry | accepted |
| `normalize_machine_name` across case, whitespace, U+2019/U+0027 | folds correctly |
| `get_machine_project_keys("")` | `[]`, #1834 guard intact |

The pre-fix reproduction from the PR body does reproduce, and the post-fix behavior is what the body claims.

### Acceptance criteria

All four met. Test names map one-to-one, including `test_validate_projects_config_runs_the_roster_check`, which pins the new validator into the aggregated suite. That wiring test is the one that would catch a future refactor silently dropping the check, and I am glad it is there rather than only the unit-level tests.

### Tests and gates

- 160 passed across `test_config_machine.py`, `test_dm_whitelist_validation.py`, `test_ui_data_memories.py`, `test_ui_app.py`, `test_reflection_machine_filter.py`
- `ruff check` clean, `ruff format --check` 1273 files, `uv lock --locked` PASS, fence census clean
- No dependency changes; branch is 1 ahead of `main`, merges clean

### One defect found and fixed (`358c2445`)

`config/projects.example.json` declared its roster as:

```json
"machines": {
  "Machine Name": "Primary bridge host",
  "Valor the Cowboy": "Second bridge host"
}
```

The file's own convention is the placeholder `"Machine Name"`, which is what its single example project names as owner (line 40). Pairing that placeholder with a real fleet host meant an operator copying the block got a roster that was half placeholder and half someone else's machine. It also did not match the roster the PR body documents. Both entries are placeholders now, and the example validates against itself: its only project owner resolves to a roster entry.

### Two nits, not blocking

1. `get_machine_project_keys` guards `if not machine` before the file read, so `""` is handled. A whitespace-only string (`"   "`) passes that guard and then normalizes to `""`, which would match a `"machine": "   "` entry. No live config can hit this and no caller passes it, so it is theoretical, but the guard would be exact as `if not normalize_machine_name(machine)`.
2. `summarize_ownership` renders other owners from the normalized (casefolded) key while the roster line renders as-written, so the report reads `other owners in config: tom's macbook pro (19)` next to `machines roster: Valor the Cowboy`. Legible either way, just inconsistent casing in operator output.

### On the docs

The `docs/features/single-machine-ownership.md` update is genuinely complete rather than perfunctory: the roster and reporting sections, the new failure mode added to the troubleshooting table, the extra step in "Adding a new machine", and a corrected test-coverage paragraph. The DOCS work is done in this PR.

Issue #2462, opened 2026-07-29
Nightly triage (15-node batch): 10 net-new test-isolation flakes (persona mock-patch pollution + shared-Redis order dependence); 5 nodes defer to #2429/#2430


valorengels:
> **Context.** The nightly regression detector (`scripts/nightly_regression_tests.py`, streamed to `logs/nightly_tests.log`) re-runs the unit suite and flags node IDs that newly transitioned from passing to failing. It confirmed a **15-node** batch. This is an **investigation-only** issue: it documents the failures, tracebacks, and root-cause grouping so a planner can scope the fix. **No fix is proposed beyond suggested next steps; no source or test code was modified.**
>
> **Relationship to #2429 and #2430.** Two sibling investigation issues from the same detector are already OPEN. **5 of the 15 nodes in this batch are already tracked** and are deferred to those issues (see the table below). The **10 remaining nodes are net-new** — not covered by either #2429 or #2430 — and are the substantive contribution here. They are all **test-isolation / order-dependent flakes**: every one passes deterministically in isolation and only fails when co-scheduled with a polluting sibling file under the parallel runner (`-n auto --dist=loadfile`, per `pyproject.toml`).

## Problem

15 unit-test node IDs across six test files were newly confirmed failing by the nightly detector. Investigation shows they split into **two already-tracked deterministic regressions** and a **net-new cluster of order-dependent test-isolation defects**. The isolation defects are the dangerous kind: they are green in CI-in-isolation and red only in the full parallel suite, so they erode trust in the nightly signal without pointing at a product bug.

**Current behavior:** nightly suite red across bridge-watchdog / persona-substitution / sdlc-stage-marker / teammate-cold-start / reflections-cutover / inbound-storage / update-hardlinks.
**Desired outcome:** the 10 net-new nodes are made isolation-robust (or the shared-state leak that pollutes them is fixed at the source) so they pass under the parallel runner; the 5 overlapping nodes are resolved via #2429/#2430.

## Full list of newly-confirmed failing node IDs

```
tests/unit/test_bridge_watchdog.py::TestHibernationSuppression::test_not_hibernating_proceeds_to_recovery
tests/unit/test_persona_substitution.py::test_load_persona_prompt_customer_service_substitution
tests/unit/test_persona_substitution.py::test_load_persona_prompt_no_substitutions_backward_compat
tests/unit/test_persona_substitution.py::test_load_persona_prompt_substitutions_none_is_safe
[...]

Issue #2430, opened 2026-07-28
Nightly regression (6-node batch): reflections cutover comment-loss + hardlinks __pycache__ cruft (net-new) plus 3 nodes overlapping #2429


valorengels:
> **Context.** The nightly regression detector (`scripts/nightly_regression_tests.py`, streamed to `logs/nightly_tests.log`) confirmed a **6-node** batch of newly-failing unit tests. This is an **investigation-only** issue: it documents the failures, exact tracebacks, and root-cause grouping. **No fix is proposed beyond suggested next steps; no code was modified.**
>
> **Relationship to #2429.** A sibling investigation issue (#2429, "Nightly regression cluster: 18 unit-test failures across four independent root causes") was filed the same day from the same detector. Three of the six nodes in *this* batch (compose baseline, bridge_watchdog AgentSession, sdlc_stage_marker self-heal) are already root-caused there — this issue reconfirms them and defers to #2429's Groups B/C/D as authoritative. The **three remaining nodes** (the reflections cutover pair + the hardlinks RENAMED_REMOVALS staleness test) are **NOT covered by #2429** and are the net-new contribution here.

## Investigation Finding

**Component:** test suite / reflections registry / update hardlinks / persona baseline / bridge watchdog / sdlc stage-marker
**Detected:** 2026-07-28 (nightly regression detector, tomcounsell/ai)
**Status:** Persistent (all six reproduced deterministically or explained by environment coupling)

## Symptoms

Six newly-confirmed failing unit-test node IDs:

```
tests/unit/test_compose_system_prompt.py::test_eng_cell_byte_stable_against_local_fixture
tests/unit/test_recovery_respawn_safety.py::TestBridgeWatchdogSafe::test_bridge_watchdog_has_no_agent_session_import
tests/unit/test_reflections_yaml_migration.py::TestPrReviewAuditCutover::test_pr_review_audit_absent_from_repo_registry
tests/unit/test_reflections_yaml_migration.py::TestSentryTriageCutover::test_sentry_issue_triage_absent_from_repo_registry
tests/unit/test_sdlc_dispatch.py::TestRunIdRequiredFlag::test_missing_run_id_self_heals_on_free_lock[stage-marker]
tests/unit/test_update_hardlinks.py::test_renamed_removals_entries_are_not_stale
```

[...]

Issue #2429, opened 2026-07-27
Nightly regression cluster: 18 unit-test failures across four independent root causes (catchup Redis isolation, stale persona baseline, watchdog AgentSession invariant, sdlc_stage_marker self-heal)


valorengels:
> **Context.** The nightly regression detector (`scripts/nightly_regression_tests.py`, streamed to `logs/nightly_tests.log`) re-runs the unit suite and flags nodes that newly transitioned from passing to failing. It confirmed a batch of 18 failing unit-test node IDs. This is an **investigation-only** issue: it documents the failures, tracebacks, and root-cause grouping so a planner can scope the fix. **No fix is proposed here beyond suggested next steps.**

## Problem

18 unit-test node IDs across six unrelated test files were newly confirmed failing. A cluster this broad often points to one shared break, but investigation shows **four independent root causes**, three of them deterministic regressions traceable to specific recent commits, and one order-dependent test-isolation defect. Each group is documented below with symptom, evidence (`file:line` + commit), and a concrete next step.

Current behavior: nightly suite red across catchup / reconciler / compose-prompt / duplicate-delivery / bridge-watchdog / sdlc-stage-marker.
Desired outcome: each group triaged and fixed (or the guarding test updated) so the nightly suite is green, with test isolation hardened so Group A cannot recur order-dependently.

## Full list of newly-confirmed failing node IDs

```
tests/unit/test_agent_catchup.py::test_main_exits_zero_on_partial_failure_and_prints_errored_chat
tests/unit/test_agent_catchup.py::test_main_exits_zero_when_run_async_raises
tests/unit/test_catchup_claim.py::TestCatchupClaimGate::test_enqueue_exception_releases_claim_no_orphan
tests/unit/test_catchup_claim.py::TestCatchupClaimGate::test_won_claim_enqueues_and_records_dedup
tests/unit/test_compose_system_prompt.py::test_eng_cell_byte_stable_against_local_fixture
tests/unit/test_duplicate_delivery.py::TestCatchupRedisDedup::test_catchup_processes_non_duplicate_message
tests/unit/test_duplicate_delivery.py::TestTelethonDuplicateDialogDedup::test_catchup_skips_duplicate_dialog_id
tests/unit/test_reconciler.py::TestReconcileOnce::test_enqueue_exception_releases_claim_no_orphan
tests/unit/test_reconciler.py::TestReconcileOnce::test_missed_message_is_enqueued_and_recorded
[...]

Issue #2327, opened 2026-07-24
nightly-tests LaunchAgent silently fails to load .env every night (bash lacks TCC Desktop-folder access)


valorengels:
> **Context:** [`.env`](../.env) in this repo is a symlink to `~/Desktop/Valor/.env`, the canonical secrets vault (see [CLAUDE.md → Secrets](../CLAUDE.md#secrets)). macOS's TCC (Transparency, Consent, and Control) privacy system requires each individual binary — not each user — to be separately granted access to protected folders like `~/Desktop`; a binary with no grant gets `EPERM` ("Operation not permitted") on any read, silently, with no error dialog for a background process.

## Problem

`com.valor.nightly-tests` (the LaunchAgent that runs [`scripts/nightly_regression_tests.py`](../scripts/nightly_regression_tests.py) every night at 3 AM) has been failing to load its environment on every single run.

**Current behavior:**

The LaunchAgent's `ProgramArguments` is:
```
/bin/bash -c "set -a; source /Users/valorengels/src/ai/.env; set +a; exec /Users/valorengels/src/ai/.venv/bin/python scripts/nightly_regression_tests.py"
```
`/bin/bash` has no TCC grant for the `~/Desktop` folder on this machine (confirmed via a direct query of `~/Library/Application Support/com.apple.TCC/TCC.db`: only specific interpreter binaries — e.g. `/opt/homebrew/.../python3.12`, `/opt/homebrew/.../python3.14` — hold `kTCCServiceSystemPolicyDesktopFolder` with `auth_value=2`; `/bin/bash` holds none). Every night, `logs/nightly_tests_error.log` records repeated:
```
/bin/bash: /Users/valorengels/src/ai/.env: Operation not permitted
```
Bash exits without exporting any vars, then `exec`s the venv python anyway — so the nightly regression run proceeds with **no environment variables from `.env`** (API keys, feature flags, DB settings, etc.), every night, indefinitely. There is no alerting on this — it fails silently inside a background LaunchAgent with no GUI to surface an error.

**Desired outcome:**

The nightly regression run gets the same environment as every other worker process (`com.valor.worker` gets 115 vars injected into its plist directly, sidestepping this problem entirely — see `[update] Injected 115 env vars into worker plist` in `logs/update.log`). No more `Operation not permitted` lines in `logs/nightly_tests_error.log`.

## Definitions

| Term | Definition | Reference |
|------|-----------|-----------|
[...]

PR #2581, opened 2026-08-06
Dedup nightly triage dispatch per node (#2559)

### LINKED ISSUE #2559: Nightly detector re-dispatches already-filed confirmed_failing nodes under a 'Newly-confirmed' header, causing duplicate-issue churn
## Root cause

The nightly regression detector dispatches nodes that are already in `confirmed_failing` state under a header reading "Newly-confirmed", and re-triages nodes that have already been filed. The result is issue churn: #2429, #2430 and #2462 each independently re-listed the same dead watchdog node as a fresh finding.

The detector has no memory of what it has already reported, so every run re-surfaces the standing set as if it were new. Three bundle issues were opened over the same underlying facts.

## Evidence

Verified by detector dry-run, not by pytest. The re-listed watchdog node appears in all three bundles.

Note: that watchdog node was not only fixed but RENAMED by `df6097fe6` to `test_bridge_watchdog_has_no_module_level_agent_session_import` (now AST-based). The old node ID can never match again, so any detector output still citing it is reporting on a node that does not exist.

## Fix shape

Suppress already-filed nodes from the dispatch set, and correct the header so `confirmed_failing` carryover is not presented as newly confirmed. A node that has an open issue against it should not generate a second one.

## Supersedes

#2532 Cluster D, part of #2488.

tomcounsell:
Closes #2559

## Root cause

`main()` computes `new_failures` correctly for the Telegram alert, then hands the **full** confirmed set to the dispatcher:

```python
summary_text = summarize_failures(new_failures, serial_report)                 # correct
triage_session_id, dispatch_hash = maybe_dispatch_triage_session(confirmed_failing, prev)   # wrong
```

The dispatch prompt then lists that whole set under a header reading `Newly-confirmed failing node IDs:`. So every run with any new failure re-triaged the standing set and filed it again. #2429, #2430 and #2462 each opened an issue over the same dead watchdog node.

The set-wide sha256 dedup could not catch it: adding one new failure changes the hash, which re-dispatches everything.

Reproduced against `2e517d3b`:

```
run 1: new=['tests/unit/x.py::test_a']
         DISPATCHED SET = ['tests/unit/test_watchdog.py::test_dead_node', 'tests/unit/x.py::test_a']
run 2: new=['tests/unit/x.py::test_b']
         DISPATCHED SET = ['tests/unit/test_watchdog.py::test_dead_node', 'tests/unit/x.py::test_b']
```

**Why the tests missed it:** they call `maybe_dispatch_triage_session(new_failures, ...)` while production calls it with `confirmed_failing`. The test asserted a contract main() did not use.

## Fix

The alert and the dispatch answer different questions. The alert asks "is this a regression since last night"; the dispatch asks "does this node already have an issue". They now have separate inputs.

| Function | Question |
|---|---|
| `compute_new_failures(prev, confirmed)` | new since last night → drives the Telegram alert |
| `compute_dispatch_set(prev, confirmed)` | never handed to triage → drives the dispatch |
| `carry_dispatched_nodes(prev, confirmed, just_dispatched)` | what to persist |

- **`dispatched_nodes`** replaces `dispatched_hash`. Per-node suppression strictly supersedes the set-wide hash, and the hash could only ever block a legitimate retry, so it is removed rather than kept alongside.
- **Retries survive.** A node whose dispatch subprocess failed is no longer "new" but is still unfiled, so it stays in the dispatch set. This is exactly why the dispatch set is not `compute_new_failures`.
- **Fixed nodes drop out**, so a genuine re-regression is dispatchable again later. This also retires the stale ID the issue calls out: `df6097fe6` renamed the watchdog node, so it stops appearing in the confirmed set and falls out of state with no special case.
- **Baseline runs seed instead of dispatching.** Without the seed the *next* run would file the entire standing set as fresh discoveries.
- **Header corrected** to `Not-yet-triaged failing node IDs:`, plus an instruction to search open issues and comment rather than open a duplicate.
- `prior_dispatched` falls back to the prior run's confirmed set when `dispatched_nodes` is absent, so the first run after this lands does not replay the churn one last time.

## Verification

```
tests/unit/test_nightly_regression_tests.py   62 passed
```

Same scenario against the fix:

```
run 1: DISPATCHED = ['tests/unit/x.py::test_a']
run 2: DISPATCHED = ['tests/unit/x.py::test_b']

standing node fixed, then regresses again:
  after clean run, dispatched_nodes = []
  regression re-dispatches: ['tests/unit/test_watchdog.py::test_dead_node']
```

New coverage: `TestComputeDispatchSet` (5), `TestCarryDispatchedNodes` (4), and `TestMainDispatchPersistence` (7) driving the whole of `main()` including the end-to-end standing-plus-new case, the failed-dispatch retry, and the baseline seed.

## Incidental fix

Three `main()`-driving tests were already red on `main` when run from a git worktree: `.env` is untracked, so it is absent there and the #2327 env guard exits before the assertions run. `load_env_or_die` is now patched in those tests, which are about dispatch persistence rather than env loading. Baselined before and after:

```
before (origin/main, in a worktree):  3 failed, 46 passed
after:                                0 failed, 62 passed
```

## DOCS

- `docs/features/nightly-alert-triage.md` — "Triage Session Dispatch" rewritten: new signature, per-node dedup semantics, alert-vs-dispatch distinction, retry/re-regression/rename/baseline behavior. Summarizer signature and files table updated.
- `docs/features/nightly-regression-tests.md` — state-file field list.
- `scripts/nightly_regression_tests.py` module docstring — new "Triage dispatch is deduped per node" section.

### DIFF
--- a/docs/features/nightly-alert-triage.md
+++ b/docs/features/nightly-alert-triage.md
@@ -26,9 +26,9 @@ three behaviors layered around that base run:
    suite and both sending Telegram alerts for the same window.
 2. **Failure summarizer** — turns a list of raw pytest node IDs into a short,
    human-readable sentence or two for the Telegram alert.
-3. **Triage dispatch** — spins up an Eng session to investigate newly-confirmed
-   failures and file a GitHub issue, deduped so the same failing set doesn't
-   re-dispatch every night.
+3. **Triage dispatch** — spins up an Eng session to investigate failures that have not
+   been triaged before and file a GitHub issue, deduped per node so a failure that
+   already has an issue open against it never generates a second one.
 
 ## Run Lock (Race 1)
 
@@ -51,9 +51,9 @@ loading prior state or running any tests.
 
 ## Best-Effort Failure Summarizer
 
-`summarize_failures(confirmed_failing, report)` turns the newly-confirmed failing
-node IDs into a 1-3 sentence plain-English summary for the Telegram alert, instead of
-a raw list of pytest node IDs.
+`summarize_failures(node_ids, report)` turns the newly-confirmed failing node IDs into
+a 1-3 sentence plain-English summary for the Telegram alert, instead of a raw list of
+pytest node IDs.
 
 - **What it does**: groups the failing node IDs by file, pulls the last line of each
   test's traceback (`call.longrepr` or `crash.message`) from the pytest
@@ -61,8 +61,8 @@ a raw list of pytest node IDs.
   (`agent/llm/wrapper.run_typed`, `config.models.MODEL_FAST`, the project's
   standard non-harness PydanticAI transport) for a short summary and likely root
   cause area.
-- **Fallback discipline**: this is best-effort only. An empty `confirmed_failing`
-  list short-circuits before any LLM call. Any other failure — network error, empty
+- **Fallback discipline**: this is best-effort only. An empty `node_ids` list
+  short-circuits before any LLM call. Any other failure — network error, empty
   or malformed LLM response, `LLMCallError`, missing `ANTHROPIC_API_KEY` — is caught
   broadly, logged as a warning, and the function falls back to the raw node-ID
   preview format (`_raw_failure_preview`: first 5 node IDs + "+N more") that
@@ -75,9 +75,9 @@ a raw list of pytest node IDs.
 
 ## Triage Session Dispatch
 
-`maybe_dispatch_triage_session(confirmed_failing, prev)` fires off an Eng-role
-`AgentSession` to investigate newly-confirmed failures, whenever the confirmed-failing
-set has changed since the last dispatch.
+`maybe_dispatch_triage_session(dispatch_nodes)` fires off an Eng-role `AgentSession` to
+investigate failing tests that have never been triaged before. The set comes from
+`compute_dispatch_set(prev, confirmed_failing)`.
 
 - **Invocation contract**: shells out to
   `python -m tools.valor_session create --role eng --slug nightly-triage-<hash8>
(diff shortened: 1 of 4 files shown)


tomcounsell:
## Review: APPROVED

Rebased onto current `origin/main` (7123245c) and reviewed at the rebased head `10382488`.

**The root cause is correctly identified and the fix addresses it rather than its symptom.** `main()` computed `new_failures` for the alert and then handed `confirmed_failing` to the dispatcher, so every run with any new failure re-triaged the whole standing set under a header reading "Newly-confirmed". The set-wide sha256 could not suppress it, because adding one node changes the hash and re-dispatches everything. Splitting the two questions into `compute_new_failures` (is this a regression since last night) and `compute_dispatch_set` (has this node already been filed) is the right decomposition, because they genuinely are different questions with different answers.

**Replacing `dispatched_hash` outright rather than keeping both is correct.** Per-node suppression strictly supersedes the set-wide hash, and the hash could only ever block a legitimate retry. Keeping it alongside would have left a second, weaker gate that can only cause false negatives. I confirmed the removal is complete: no `dispatched_hash` reference survives in `scripts/nightly_regression_tests.py`, and the only mention left in the tests is `test_dispatched_hash_is_gone_from_persisted_state`, which asserts its absence.

**The three behavioral edges are each covered by a test that would fail without the fix.** A failed dispatch leaves the node unfiled so it retries (`test_failed_dispatch_leaves_nodes_unfiled_for_retry`); a fixed node drops out so a genuine re-regression is dispatchable later (`test_node_that_stopped_failing_drops_out`); a baseline run seeds instead of filing the whole standing suite as fresh discoveries (`test_baseline_run_seeds_rather_than_dispatches`). The `prior_dispatched` fallback to the prior confirmed set is what stops the first run after this lands from replaying the churn one final time, which is an easy thing to miss.

**Worth naming: the reason this defect survived is a test that asserted a contract production did not use.** The old tests called `maybe_dispatch_triage_session(new_failures, ...)` while `main()` called it with `confirmed_failing`. The test was green and the code was wrong. `TestMainDispatchPersistence` now drives `main()` end to end rather than the helper in isolation, which is what closes that gap. This is the same false-signal family as #2568.

### Rebase and conflict resolution

The branch was DIRTY against main. Three conflicts, all in `tests/unit/test_nightly_regression_tests.py`, all the same underlying story: this PR and #2573 (landed on main as `cbb07617`) independently made the same incidental fix, patching `load_env_or_die` so the #2327 env guard does not exit before the assertions in a git worktree where `.env` is absent.

I resolved for intent rather than by taking a side:

1. **Sentinel value.** Main used `return_value=42`; this PR used `return_value=0`. I took main's `42` at all three sites. It is the discriminating choice: these tests assert `nrt.main() == 0`, so a `0` sentinel would pass even if `load_env_or_die`'s return leaked into `main()`'s return value, while `42` fails loudly. Taking the PR's `0` would have reintroduced exactly the vacuous-assertion shape #2568 was cleaning up. I added a comment at the helper explaining why the sentinel is deliberately non-zero, so a future reader does not "simplify" it back to `0`.
2. **Helper structure.** I kept this PR's parameterized `_run_main` helper over main's inline test bodies, since the helper is the new structure the added coverage depends on, and folded main's sentinel into it.
3. **`dispatched_hash` assertion.** Main's side still asserted `saved["dispatched_hash"] == "new-hash-for-current-failures"`. I took this PR's side, since that field is deliberately deleted here. Main's version also used a `mock_send_telegram` binding inconsistent with the shared tail of the test, which the PR's side gets right.

### Verification (post-rebase, against the merge onto `origin/main` 7123245c)

```
tests/unit/test_nightly_regression_tests.py    62 passed
```

62 passing matches the PR's pre-rebase claim, so the conflict resolution neither dropped nor duplicated coverage.

### Non-blocking observation

The incidental env-guard fix is now redundant with #2573 at the two sites that overlap, but it is not dead: this PR's new `_run_main` helper is a fourth site that #2573 never touched and genuinely needs the patch. No cleanup owed.

No blockers.

Issue #2494, opened 2026-07-31
Durability refactor: Room / Job / AgentSession — recovery keys on work owed, not process status


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/durability-room-job-agentrun.md

> **Context:** This system runs autonomous agents that answer messages arriving over Telegram and email. All operational state lives in Redis via [Popoto](https://github.com/tomcounsell/popoto), a Redis ORM. Today a single model, [`AgentSession`](models/agent_session.py), represents four different things at once: a conversation with a human, a queued work item, an agent's context, and a running OS process. This issue proposes splitting it, because that conflation is the root cause of a class of failures in which **work is silently lost and nothing detects it**.

## Problem

Session durability in this repo is not designed — it is an emergent property of roughly fourteen independent recovery mechanisms ("healers") that accumulated one per incident. Every one of them is keyed on `AgentSession.status`.

That means every healer answers the same question: **"did a process die badly?"** No mechanism anywhere answers **"was the work that was promised actually delivered?"**

Consequently, every real loss happens outside the guarded perimeter. A five-agent code audit on 2026-07-31 confirmed the following by reading code (not by inference):

**Current behavior:**

| Failure | Evidence | Detected today? |
|---|---|---|
| Direct messages are permanently lost if the bridge crashes before enqueue | All three recovery scanners start with `if not chat_title: continue` ([reconciler.py:212](bridge/reconciler.py#L212), [catchup.py:109](bridge/catchup.py#L109), [agent_catchup.py:243](bridge/agent_catchup.py#L243)). A Telegram private chat is a `User` entity with no `title`. | ❌ never |
| Inbound email lost in batches | IMAP `\Seen` is set for the whole batch *before* fetch ([email_bridge.py:1596](bridge/email_bridge.py#L1596)); the search is `UNSEEN`, so a crash mid-batch loses every message in it | ❌ never |
| A "stop" message destroys the instructions queued with it | [health_check.py:511](agent/health_check.py#L511) drains the whole steering list, finds an abort, and `return`s before the re-push. Unconditional — no crash required. | ❌ never |
| Steering messages orphan forever and are later mis-consumed | `steering:{session_id}` has no TTL; `clear_steering_queue` has **zero** production callers. A stranded message becomes the first turn of the next session reusing that id. | ❌ never |
[...]

Issue #2479, opened 2026-07-30
Make machine-ownership a structural precondition for recovery-path enqueues, not a single call site


valorengels:
## Context

Hotfix `5332768ff` fixed a real bug: on a machine that does not own `valor`, every crash-storm alert enqueued a permanently-pending AgentSession that no local worker could ever claim. 75 stranded sessions accumulated as dashboard noise on the sibling machine.

The fix is sound as far as it goes. `_machine_owns_alert_project()` (`monitoring/bridge_watchdog.py:763`) resolves through `config.machine.get_machine_project_keys`, the canonical ownership seam, and **fails closed on enqueue while falling back to filing a GitHub issue** (`:898`) — so a suppressed alert still reaches a human. That fail-direction is correct and worth preserving in any refactor.

## The structural gap

Single-machine ownership is a **system-wide invariant** (see `docs/features/single-machine-ownership.md`, enforced for config by `bridge/config_validation.py`). But the runtime gate is applied at exactly **one call site** (`:898`). It is a hand-applied convention, not a precondition. The next author who adds a session enqueue anywhere in the recovery path must remember to call it — and nothing catches them if they don't. That is how this bug happened the first time.

## Secondary: the hardcoded default

`monitoring/bridge_watchdog.py:896`
```python
project_key = os.environ.get("VALOR_PROJECT_KEY", "valor").strip() or "valor"
```

The ownership *gate* was added, but the *key* still defaults to `"valor"` on every machine. On any non-`valor` machine the gate therefore always fails and the GitHub-issue fallback becomes the only working path. The alert works, but never via its intended channel — a permanently degraded path that reads as "working" because the fallback fires.

## Work to do here

1. Make ownership a **structural precondition** for any recovery-path session enqueue rather than a call-site convention. Options to weigh: push the check into the enqueue seam itself so it cannot be bypassed; or add a guard/test that fails when an enqueue path lacks it. Prefer the seam — a test only catches what someone remembered to enumerate.
2. Resolve the `project_key` default from machine ownership rather than hardcoding `"valor"`, so the intended Telegram path works on every machine instead of silently degrading to the fallback.
3. Preserve the current fail-direction: unresolvable ownership must keep suppressing the enqueue **and** keep the GitHub fallback. Do not "fix" the fail-closed behavior into fail-open.

## Before patching

[...]

PR #2582, opened 2026-08-06
Gate direct-to-main commits on an explicit issue disposition (#2540)

### LINKED ISSUE #2540: Hotfixes landed directly on main never close the issues they resolve; the PR path enforces Closes #N and the hotfix path has no equivalent gate
> **Context:** This repo ships work two ways. The **SDLC pipeline** routes an issue through plan, build, test, review and merge, landing via a pull request. The **hotfix path** lets small or urgent fixes land directly on `main` with no PR, sanctioned by `CLAUDE.md`'s quick-task threshold. GitHub only auto-closes an issue when a *closing keyword* (`Closes`/`Fixes`/`Resolves` followed by `#N`) appears in a merged PR body or in a commit message on the default branch. A bare `#N` mention closes nothing.

## Problem

The SDLC merge path **enforces** issue linkage. `do-merge` requires the PR body to carry a closing keyword ([`.claude/skills-global/do-merge/SKILL.md:147`](.claude/skills-global/do-merge/SKILL.md#L147)), and `tools/sdlc_stage_query.py` deliberately refuses to treat a bare `#N` as a link ([`:278`](tools/sdlc_stage_query.py#L278): *"A bare `#N` mention with no closing keyword is intentionally NOT a match"*).

The hotfix path has **no equivalent gate**. Nothing requires a hotfix commit to name the issues it resolves, and nothing sweeps open issues afterward. So a hotfix can silently resolve or fully obsolete an open issue, and that issue stays open indefinitely, indistinguishable from live work.

This is not hypothetical. It was found by manually auditing all 29 open `bug`-labeled issues on 2026-08-06.

**Current behavior:**

| Commit | Issue linkage | Result |
|---|---|---|
| [`9fe58f45d`](https://github.com/tomcounsell/ai/commit/9fe58f45d) "Hotfix: remove the watchdog crash-storm alert delivery path" (2026-08-04) | **none** | Deleted `_machine_owns_alert_project()`, the gated enqueue, and the hardcoded project-key default. This fully obsoleted **#2479** (whose three named targets were exactly those symbols) and resolved **#2429**'s Group C. Both stayed open for two days until a manual audit caught it. #2479 was closed by that audit; #2429 remains open on its other groups. |
| [`f695d2bed`](https://github.com/tomcounsell/ai/commit/f695d2bed) "Hotfix SDLC skill-file defects: #2493, #2492, #2465, #2419" | **bare `#N`, no keyword** | #2493 is closed; #2492, #2465 and #2419 are still open. Because a bare mention carries no resolved/partial semantics, a reader cannot tell whether those three were left open deliberately or by omission. The commit message is not evidence either way. |
| [`ac3a87d51`](https://github.com/tomcounsell/ai/commit/ac3a87d51) "Fix #2489: clear `deferred_self_draft_pending` on successful resend" | **closing keyword** | #2489 auto-closed correctly. This is the working counter-example: the mechanism exists and works on the hotfix path when used. |

The audit's broader finding sharpens the point. Of 29 open bug issues, **exactly one** was invalidated, and it was invalidated by a hotfix, not by any merged PR. The single largest source of stale open issues in that batch was the one path with no linkage gate.

**Desired outcome:**

- A hotfix that resolves an issue closes it, or explicitly records that it only partially addresses it.
- "This issue was silently resolved by a commit nobody linked" is not a state a human has to discover by reading source.
- The fix does not make the hotfix path heavy enough that people route around it. The hotfix path exists because it is fast; a gate that costs a PR's worth of ceremony defeats its purpose.

## Definitions

| Hotfix path | Landing a small or urgent change directly on `main` with no PR and no SDLC stages. Sanctioned for quick/small tasks. | `CLAUDE.md`, Development Workflow |
| SDLC merge path | The full pipeline (plan → build → test → review → docs → merge), landing via PR. | [`.claude/skills/sdlc/SKILL.md`](.claude/skills/sdlc/SKILL.md) |
| Closing keyword | `close`/`closes`/`closed`/`fix`/`fixes`/`fixed`/`resolve`/`resolves`/`resolved` followed by `#N`. GitHub auto-closes on merge to the default branch. | [GitHub docs](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue) |
| Obsoleted by deletion | An issue whose defect ceased to exist because the code was removed, not repaired. #2479's shape. Produces no test signal and no diff a reviewer would connect to the issue. | this issue |

## Open questions

The root cause is a missing gate, but the right *shape* of gate is genuinely unsettled. Deciding this is the plan's job, not this issue's:

1. **Enforce at commit time, or sweep after?** A `pre-push`/`prepare-commit-msg` hook on direct-to-`main` commits could demand an issue reference or an explicit `no-issue` opt-out. Alternatively a periodic sweep could re-verify open issues against current source. These have very different failure modes: a hook is fail-closed and annoying; a sweep is fail-open and lossy.
2. **Can obsoleted-by-deletion even be caught mechanically?** #2479 is the hard case. The hotfix author was not thinking about #2479 and had no reason to. A gate that only asks "which issue does this fix?" would not have caught it. Detecting it requires checking issues against code they cite, which is closer to the manual audit that actually found it.
3. **Is the real unit of work a recurring audit?** The 2026-08-06 audit read 29 issues against current source and found one stale. That is a low hit rate but a real one, and it also surfaced #2497, whose *headline diagnosis was false when filed*. A scheduled re-validation pass may be worth more than a commit-time gate, or may be worth having in addition. See the `reflections` system for existing scheduled-maintenance precedent.
4. **Does this generalize to bare-`#N` hotfixes?** `f695d2bed` shows a second, quieter failure: mentions that record neither resolution nor partiality. Whether the fix should also demand explicit per-issue disposition is scope the plan should rule on.

## Acceptance criteria

- [ ] A hotfix landing on `main` that fully resolves an open issue results in that issue being closed, without the author having to remember unprompted.
- [ ] The chosen mechanism is documented in `CLAUDE.md` or a `docs/features/` page, whichever the plan rules correct.
- [ ] The hotfix path remains fast enough that it is still preferable to a PR for genuinely small changes. The plan must state how it verified this rather than asserting it.
- [ ] There is a stated, tested answer for the obsoleted-by-deletion case, even if that answer is an explicit "not mechanically detectable, handled by the audit instead."
- [ ] Whatever is built fails loudly. A silently-dead sweep reproduces the exact class of defect that #2494 catalogues: *"a dead detector emits silence, and silence is indistinguishable from health."*

## Prior context

- **#2494** — the durability refactor. Its central thesis, that silent loss with no owner is the real defect class, is the same shape as this issue at the project-management layer rather than the runtime layer.
- **#2479** — closed 2026-08-06 by the audit. The concrete instance that motivated this issue.
- **#2429** — still open; its Group C was resolved by `9fe58f45d` and never marked.
- **#2497** — separate defect surfaced by the same audit: an issue whose headline claim was already false when filed. Argues for question 3 above.

## Downstream context

This issue will be consumed by `/do-plan` to produce a plan document at `docs/plans/{slug}.md`, then executed by `/do-build`. The plan should resolve the four open questions above before proposing an implementation. **Do not treat the open questions as a solution sketch to execute.**

tomcounsell:
Closes #2540

## Problem

The PR path enforces issue linkage — `do-merge` requires a closing keyword in the PR body, and `tools/sdlc_stage_query.py` deliberately refuses to treat a bare `#N` as a link. The hotfix path had no equivalent gate, so a hotfix could resolve an open issue and leave it open indefinitely, indistinguishable from live work.

The 2026-08-06 audit of 29 open `bug` issues found exactly one stale issue, and it was stale because of a hotfix rather than any merged PR. The single largest source of stale open issues was the one path with no linkage gate.

## The rule

A commit landing code on `main` must carry one of:

| Form | Meaning |
|---|---|
| `Closes #123` (also `Fixes`, `Resolves`) | Resolves #123; GitHub auto-closes on push |
| `Refs #123` | Touches #123 but does **not** resolve it |
| `No-issue: <reason>` | Nothing to link, and why. A reason is required |

A bare `#123` satisfies nothing — that ambiguity is what the gate removes.

## Resolving the issue's four open questions

1. **Commit time, not a sweep.** A sweep is fail-open and lossy; #2494's thesis is that silent loss with no owner is the defect class. The gate decision is fail-closed; infrastructure failures fail open so a broken environment never bricks every commit.
2. **Obsoleted-by-deletion is not mechanically detectable.** Stated, not papered over. #2479's author was not thinking about #2479 and had no reason to be; a gate asking "which issue does this close?" cannot surface an issue nobody considered. That class stays with the periodic audit.
3. **The recurring audit is complementary** and not built here.
4. **Yes, this covers bare `#N`.** `f695d2bed` mentioned four issues with no keyword and left three open; it is now blocked until each gets a disposition.

## Two legs, one predicate

There are two ways to land on `main`, so there are two hooks sharing `find_violation()`:

- **`.githooks/commit-msg`** — commits authored *on* `main`. The commit-msg stage rather than a PreToolUse Bash hook because it sees the final message however it arrived; `git commit -F -` carries the message on stdin where no inspector of the command string can read it.
- **`.githooks/pre-push`** — commits authored on a side branch and pushed *to* `main`. This is the shape a worktree hotfix takes and the commit-msg leg cannot see it. This hook also wires up `tools/push_ancestry_guard.py`, whose docstring claimed pre-push installation that did not actually exist.

## Keeping the hotfix path fast

Only `docs/plans/` is exempt, covering the `Migrate completed plan:` / `Plan (slug):` traffic that is the bulk of legitimate direct-to-`main` commits and never resolves an issue by itself. Cost on a real hotfix is one trailer line. Prose is deliberately **not** exempt — `f695d2bed` changed only `.md` skill files. Feature branches are never gated.

## Failing loudly

These hooks run only when `core.hooksPath` is `.githooks`, which `/update` sets and `/setup` does not. A machine can silently skip them, and a skipped gate looks exactly like a passing one. `python -m tools.doctor` now reports `git_hooks_installed` with a one-line fix.

## Verification

`tests/unit/test_commit_issue_disposition.py` (37 tests): the predicate against the issue's three real commits, plus both hooks subprocessed against an ephemeral repo with a bare remote so the wiring is proven rather than assumed.

| Case | Result |
|---|---|
| `9fe58f45d` hotfix, no linkage | BLOCKED |
| `f695d2bed` bare `#N` on skill files | BLOCKED |
| `ac3a87d51` closing keyword | ALLOWED |
| `-F -` heredoc, no disposition | BLOCKED |
| plan-only commit | ALLOWED |
| normal hotfix-to-main push | ALLOWED |
| side branch pushed to main, undisposed | BLOCKED |
| side branch pushed to its own ref | ALLOWED |
| normal PR-merge flow | ALLOWED |

## Rollout note

This activates on merge for every machine whose `core.hooksPath` is set. Agents landing hotfixes will need one trailer line. `--no-verify` is the break-glass.


tomcounsell:
Self-review finding, folded into the commit before review: `.githooks/pre-push` was first committed `100644`. git silently **ignores** a non-executable hook (it prints only a `hint:` line), so the push leg — the one that covers the worktree hotfix shape — was dead on arrival.

The repo-fixture tests did not catch it because the fixture `chmod`s its own copies. Added `test_hook_is_committed_executable`, which asserts the mode in the git index directly for both hooks. Same failure class the PR is about: a gate that does not run looks exactly like a gate that passes.

tomcounsell:
## Review: APPROVED

Verified against the merge onto current `origin/main`. Reviewed last in tonight's queue deliberately, because this one changes how every agent's commits are gated and the others were still landing.

**The gap is real and correctly located.** The PR path already enforces linkage — `do-merge` requires a closing keyword, and `tools/sdlc_stage_query.py` refuses to read a bare `#N` as a link. The hotfix path had no equivalent, so the one path with no gate was the single largest source of stale open issues. Putting the same judgement on both sides is the right shape.

**The fail direction is split correctly, and that split is the whole safety story.** The gate *decision* is fail-closed: no disposition means refused. Infrastructure failures fail open — no git, unreadable message file, unknown remote SHA on a first push, unresolvable range on a shallow clone. That is the right asymmetry. A gate that bricks every commit on a machine with a broken environment would be worse than the gap it closes, and each fail-open branch carries a comment saying which infrastructure condition it covers.

**Two legs are genuinely necessary, not belt-and-braces.** `commit-msg` catches commits authored on `main` and sees the final message however it arrived, including `git commit -F -` where the message is on stdin and no inspector of the command string could read it. `pre-push` catches the worktree hotfix shape — commit on `session/{slug}`, then `git push origin HEAD:main` — which the commit-msg leg structurally cannot see, because at commit time the branch was not `main`. Both share `find_violation()`, so they cannot drift.

### The blast-radius question I most wanted answered

The rollout note says this activates on merge for every machine with `core.hooksPath` set, with other agents actively landing commits. So I measured it rather than reasoning about it. Applying `find_violation` to the last 25 commits on `main`:

```
9 allowed, 16 blocked
```

16 looks alarming. It is not, and the reason is the part worth recording:

```
blocked & created server-side by GitHub squash (hooks never see these): 14
blocked & authored locally direct-to-main (genuinely gated):             2
   -> Retire the dangling _AGENT_SESSION_FIELDS references left by #2563
   -> Stop equating the 651ms p99 with the measured-rebuild table's first row
```

A `gh pr merge --squash` commit is created on GitHub's side. There is no local `git commit` and no local `git push`, so neither hook ever evaluates it. And when a machine later pushes something else to `main`, `check_pushed_range` uses `rev-list remote_sha..local_sha`, which is exactly the set of commits the push *adds* — anything already on the remote is never re-judged. That exclusion is explicitly documented in the docstring and it is what keeps the gate from retroactively blocking history on the next push.

So the real blast radius is 2 commits in 25, and both are precisely the target shape: locally-authored hotfixes landing code on `main` while citing issues by bare `#N` with no statement of whether they resolve them. Those are the commits #2540 is about. The gate hits what it aims at and essentially nothing else.

### The self-review catch was the best thing in this PR

`.githooks/pre-push` was first committed `100644`. Git silently ignores a non-executable hook — it prints only a `hint:` line — so the push leg, the one covering the worktree hotfix shape, was dead on arrival. The repo-fixture tests missed it because the fixture `chmod`s its own copies, which is the classic way a fixture launders the very property under test.

That is the same failure class this PR exists to fix: a gate that does not run looks exactly like a gate that passes. `test_hook_is_committed_executable` now asserts the mode in the git index directly. I verified it at source rather than trusting it:

```
100755  .githooks/commit-msg
100755  .githooks/pre-push
100755  scripts/check_issue_disposition.py
```

**The `git_hooks_installed` doctor check belongs here for the same reason.** These hooks run only when `core.hooksPath` is `.githooks`, which `/update` sets and `/setup` does not. A machine can silently skip the whole gate, and a skipped gate is indistinguishable from a passing one. Shipping the enforcement without the "is it even installed" report would have reproduced the defect one level up.

### Scope decisions, all defensible

- **Only `docs/plans/` is exempt.** That covers the `Migrate completed plan:` / `Plan (slug):` traffic which is the bulk of legitimate direct-to-`main` commits and never resolves an issue by itself. Cost on a real hotfix is one trailer line.
- **Prose is deliberately not exempt**, and the PR is right about why: `f695d2bed` changed only `.md` skill files, mentioned four issues with no keyword, and left three open. Exempting docs would have exempted the motivating commit.
- **`is_generated_message` exempts merge/revert/fixup/squash/amend**, so the gate fights authors rather than git's own machinery. It correctly skips comment lines and blanks before testing the first real line.
- **The closing-keyword regex matches GitHub's own set**, so the gate passes precisely when GitHub will actually auto-close. Any narrower or wider would let the two disagree.

**Question 2 is answered honestly rather than papered over.** An issue obsoleted by deletion (#2479's shape) is not mechanically detectable, because that hotfix's author was not thinking about #2479 and had no reason to be. A gate asking "which issue does this close?" cannot surface an issue nobody considered. Stating that as a limitation and leaving it to the periodic audit is the correct call, and it is in the module docstring where the next reader hits it.

### Verification

```
tests/unit/test_commit_issue_disposition.py + test_doctor.py   80 passed
python -m ruff format --check . 1281 files already formatted
git index modes                 100755 on both hooks and the script
```

### Operational note for the fleet

Once this lands, any agent landing a direct-to-`main` hotfix needs one of `Closes #N`, `Refs #N`, or `No-issue: <reason>`. `--no-verify` is the documented break-glass. Both of the two genuinely-gated commits I found above were authored tonight, so this will be felt immediately by agents still working. That is the intended effect and the block message names the three accepted forms inline, so the fix is self-evident at the point of failure.

No blockers.

tomcounsell:
## Review: Approved (posted post-merge; landed as 88825e31 at 19:28:43)

Verified the predicate against **real commits from this repo**, including ones made today, rather than only the fixtures.

### The exemptions are correctly scoped

| Commit | Verdict | Why it matters |
|---|---|---|
| `922fb915` "Migrate completed plan: ..." | **ALLOWED** | plans-only, the exempt shape. This was my main worry: `/do-merge` commits these to `main` after every merge, and gating them would have broken the merge workflow. It does not. |
| `0f19460e` docs one-liner, direct to main today | **BLOCKED** | prose deliberately not exempt |
| `a18770fc` doc-retirement follow-up, direct to main today | **BLOCKED** | same |
| same commit on `session/foo` | **ALLOWED** | feature branches never gated |

Trailer handling is exactly as documented — `Closes`/`Fixes`/`Resolves`/`Refs`/`No-issue:` all pass, bare `#1` and an empty body both block. Getting bare `#N` to fail is the whole point and it does.

### The error message is the best part

```
COMMIT BLOCKED (#2540): this commit lands code on  with no issue disposition.
...
  Closes #123          this commit resolves #123 (also Fixes / Resolves)
  Refs #123            this commit touches #123 but does NOT resolve it
  No-issue: <reason>   nothing to link, and here is why

Staged files outside docs/plans/ (1):
  docs/features/popoto-index-hygiene.md

Bypass with On branch main
Your branch is up to date with 'origin/main'.

nothing to commit, working tree clean if this gate is wrong for your case.
```

It names the rule, the three forms with their distinct meanings, the specific offending files, the bypass, and where the rationale lives. A gate whose message does not tell you how to satisfy it just becomes a thing people `--no-verify` past reflexively; this one does not have that problem.

Splitting the two legs is right too. A PreToolUse Bash inspector genuinely cannot see `git commit -F -`, and the pre-push leg catches the worktree-hotfix shape the commit-msg leg structurally cannot.

### Verification

`tests/unit/test_commit_issue_disposition.py` — **39 passed**. Fail-open paths confirmed at `check_issue_disposition.py` lines 131, 235, 283, 292, so a broken environment degrades to permissive rather than bricking every commit. That asymmetry is the correct one for a commit-time gate.

### Rollout note for whoever is watching the fleet

`core.hooksPath` is **already** `.githooks` on this machine, so this went live the moment it merged, for every agent working in this repo right now. Two commits made today would have been blocked under it, both legitimate small docs fixes. That is the intended behavior and the cost is one trailer line, but agents mid-task will hit it without warning and the block will look like a bug to them.

The `python -m tools.doctor` `git_hooks_installed` check is the right mitigation for the opposite failure (a machine silently skipping the gate), and I am glad it shipped in the same PR rather than as a follow-up. A skipped gate looking identical to a passing one is exactly the class of thing that goes unnoticed for weeks.

### One observation

A GitHub squash-merge never touches the local pre-push path, so PR merges are not gated by this and do not need to be — the PR body requirement in `do-merge` already covers them. Noting it only because I initially tested a squash commit against the predicate and it reported BLOCKED, which is an artifact of testing the predicate directly rather than a real gap.

PR #2583, opened 2026-08-06
Bound class-set retry by a tunable budget; make exhaustion observable (#2550)

### LINKED ISSUE #2550: Class-set retry cap is sized for 150 sessions but the fleet runs 4000+, and it degrades silently
Split out from #2549, where the underlying window was corroborated. This is a **live-path** defect, not a migration one, which is why it is filed separately.

## The sizing

`tools/valor_session.py:58-70` documents the retry design:

> Measured (spike-1, 150 sessions): `rebuild_indexes()` takes ~600ms; p99 class-set-empty window = 651ms. Retry cap: 5 attempts x 200ms = 1000ms total, which covers the measured p99 with ~35% margin.

Two readers depend on that cap:

- `tools/valor_session.py::_find_session`
- `tools/sdlc_stage_query.py::_find_session_by_id` (`:74-81`)

## The problem

**The measurement was taken at 150 sessions. Host `Valor the Pirate` currently holds 4006** — about 27x. `repair_indexes` there logs `sessions_rebuilt=4006` on a ~31-minute cadence, and `models/agent_session.py` re-adds class-set members in `batch_size=1000` pipeline batches, so the delete-to-re-add window grows with row count.

If the window scales even sub-linearly, the real p99 is now well beyond the 1000ms cap. The "~35% margin" is margin against a measurement that no longer describes any production machine.

## Why it is worse than a stale constant

**The failure is silent and it looks like something else.** When the cap is exhausted, `_find_session` returns `None` — indistinguishable from a session that genuinely does not exist. The caller reports "session not found" for a **live** session. Nothing logs "I retried 5 times and gave up during an index rebuild".

So the symptom is a rare, unreproducible "session not found" that resolves on the next attempt, on paths the SDLC pipeline depends on for session lookup. That is a hard bug to attribute, and it gets steadily more likely as the keyspace grows, with no signal that anything changed.

#2549 documents the same window swallowing a whole migration scan on this host, which is the same root cause landing on a different reader.

## What to do

- [ ] Re-measure the class-set-empty p99 at realistic keyspace sizes (4000+, not 150). Record the row count alongside the number this time.
- [ ] Re-size `_CLASS_SET_RETRY_ATTEMPTS` / the backoff against that measurement, and name the constant's provenance so the next reader can tell it is stale.
- [ ] **Make cap exhaustion observable.** Log at WARNING when the retry loop exhausts, distinguishing "index was mid-rebuild" from "genuine absence". Without this the defect stays invisible no matter how the cap is sized.
- [ ] Consider whether a fixed cap is the right shape at all, given the window scales with row count. Options: derive the cap from the observed row count, or detect the rebuild directly rather than guessing a duration.
- [ ] Audit for other `query.filter` / `query.all` readers with no retry at all. #2549 found one (the strip migrations, now fail-closed via PR #2545); there may be more.

## References

- #2549 (the corroborated window and its timeline), #1720 (original class-set window), #2536 (adjacent index work)
- `tools/valor_session.py:58-70`, `tools/sdlc_stage_query.py:74-81`
- `models/agent_session.py` `repair_indexes()` and popoto's `rebuild_indexes()` Step 1 delete / Step 2 batched re-add

tomcounsell:
Closes #2550

## What I could and could not verify

**Verified on current main:** the constant's provenance is exactly as reported (`tools/valor_session.py:66-68` documents "spike-1, 150 sessions"), both readers share it, and exhaustion is silent — a budget-exhausted read returns the same `None` as a genuine miss with only a `logger.debug` per attempt.

**Could not verify locally:** this host holds **28** `AgentSession` rows, not 4000+. I did not reproduce the p99 window at fleet scale, and I deliberately did not create 4000 synthetic sessions in shared Redis with five other agents running against the same DBs. So the re-measurement checkbox is **not** ticked by this PR, and the fix is designed to not need it.

## Why not just re-measure and re-size

`rebuild_indexes()` re-adds members in `batch_size=1000` pipeline batches, so the window scales with row count. A constant measured at 4000 rows goes stale at 40000 exactly as the 150-row constant went stale at 4000. The issue's own fourth checkbox asks whether a fixed cap is the right shape at all; this PR answers no.

The ceiling is now a **wall-clock budget**, `TIMEOUTS__CLASS_SET_RETRY_BUDGET_S` (default 3s, was effectively 1s). A large-keyspace host raises its own ceiling with no code change, and the field description in `config/settings.py` carries the 150-session provenance so the next reader can see where the old number came from.

## Exhaustion is no longer silent

This is the item the issue calls essential ("without this the defect stays invisible no matter how the cap is sized"). `log_class_set_exhaustion` logs at WARNING with the attempt count, the budget, and the env key to raise.

The two readers can say different things, and they now do:

- **`_find_session`** falls through to `get_by_id`, a direct key lookup that never touches the class set. A hit after exhaustion **proves** the class set was mid-rebuild, and the message says so.
- **`_find_session_by_id`** takes a `session_id`, not the primary key, so it has no such fallback. Its message states plainly that stale-index and genuine-absence are indistinguishable rather than asserting either.

Live output:

```
WARNING _find_session: class-set read for 'definitely-not-a-real-session-id' came back
empty on all 14 attempt(s) over a 3.0s budget. This is reported to the caller as
'not found', which is indistinguishable from genuine absence — if the row does exist,
the class set was still rebuilding and TIMEOUTS__CLASS_SET_RETRY_BUDGET_S is too small
for this host's keyspace.
```

## Bug found while wiring this up

`_find_session` accepts either a `session_id` or a 32-char hex `agent_session_id`. For the UUID form, an empty `query.filter(session_id=<uuid>)` is the **documented expected outcome** before `get_by_id` fires. The retry loop did not know that, so every UUID lookup burned the whole cap first — and under the larger budget it would have burned 3s and then logged a rebuild that never happened. UUID-form arguments now route straight to `get_by_id`.

```
by session_id -> True 0.009s
by UUID       -> True 0.061s   (was ~1s, would have been ~3s)
```

## Audit of unretried readers (issue's last checkbox)

~75 call sites do `AgentSession.query.filter(session_id=...)` with no retry, across `bridge/`, `agent/`, `worker` paths, `.claude/hooks/`, and other `tools/` modules. **The bulk of these are the deliberate hot-path exclusion** from #1720: worker recovery, steering delivery, and hook paths handle `None` gracefully and latency matters there.

Fixing them is not a list edit — each needs a judgment call on whether a false "not found" is harmless or load-bearing at that site. I have left that out of this PR rather than doing it shallowly, and it is worth its own issue. The shared `tools/class_set_retry.py` this PR introduces is what makes that follow-up cheap.

## Changes

| File | Change |
|---|---|
| `tools/class_set_retry.py` | New. `class_set_retry_attempts` (budget-driven generator) and `log_class_set_exhaustion`. |
| `config/settings.py` | `TimeoutSettings.class_set_retry_budget_s`, `.class_set_retry_backoff_s`, with provenance in the description. |
| `tools/valor_session.py` | Uses the shared helper; UUID-form fast path; WARNING on exhaustion with the `get_by_id` proof. Duplicated constants removed. |
| `tools/sdlc_stage_query.py` | Uses the shared helper; WARNING on exhaustion. Duplicated constants removed. |

## Verification

```
tests/unit/test_class_set_retry.py                                        10 passed
+ test_sdlc_stage_query.py, test_valor_session_cli.py,
  test_valor_session_enumeration_seam.py                                 105 passed
```

New tests cover the first-attempt-always-yields contract, budget adherence, a larger budget buying more attempts, zero cost on a first-read hit, WARNING level, and both exhaustion messages including the proven-vs-ambiguous distinction.

**Known tradeoff:** a genuinely absent session now costs up to 3s instead of 1s on the miss path. At operator/dispatch invocation rates this is the intended trade — the whole premise is that a "miss" may be a live session.

## DOCS

- `docs/features/popoto-index-hygiene.md` — retry paragraph rewritten: budget not attempt count, the row-count scaling argument, WARNING exhaustion with the proven-vs-ambiguous split, UUID fast path.
- `docs/features/session-lifecycle.md` — "Read-path retry defense" section: stale `5 × 200ms = 1000ms` parameters replaced, 651ms/150-session measurement kept but labelled as the historical spike it is.


tomcounsell:
## Review: APPROVED

Verified against the merge of `session/ws-fleet-2550` onto `origin/main` (1d9a26c1).

**The reframing from attempt-count to wall-clock budget is the right answer to the issue, and the PR argues for it correctly.** `rebuild_indexes()` re-adds members in `batch_size=1000` pipeline batches, so the empty-class-set window scales with row count. Any fresh constant would decay exactly as the 150-session one did. Moving the ceiling into `TimeoutSettings` lets a large-keyspace host raise it via `TIMEOUTS__CLASS_SET_RETRY_BUDGET_S` with no code change, and keeping the spike-1 provenance in the field description means the next reader can see where 651ms came from instead of finding another undocumented number.

**I want to credit the honesty in the PR body specifically.** It states plainly that the fleet-scale re-measurement checkbox is *not* ticked, that this host holds 28 `AgentSession` rows rather than 4000+, and that it deliberately declined to create 4000 synthetic sessions in shared Redis while five other agents were running against the same DBs. That was the correct call on both counts, and declaring an unticked checkbox rather than quietly implying coverage is what makes the rest of the report trustworthy. The fix is designed to not need that measurement, which is why the gap is acceptable.

**Exhaustion observability is the load-bearing half and it is done properly.** `log_class_set_exhaustion` logs at WARNING with attempt count, budget, and the env key to raise. The `resolved_by_fallback` discriminator is the part worth calling out: `_find_session` can fall through to `get_by_id`, a direct key lookup that never reads the class set, so a hit after budget exhaustion *proves* the class set was mid-rebuild. `_find_session_by_id` takes a `session_id` rather than the primary key and has no such fallback, so its message states that stale-index and genuine-absence are indistinguishable instead of asserting a cause it cannot know. Two readers, two different epistemic positions, two different messages. That is the distinction the issue asked for.

**The UUID fast path is a real bug found in passing, not scope creep.** `_find_session` accepts either form, and for the 32-hex `agent_session_id` an empty `query.filter(session_id=...)` is the *documented expected* outcome before `get_by_id` fires. The old loop burned the full cap on every UUID lookup. Under the new 3s budget that regression would have been three times worse and would have logged a rebuild that never happened, turning the new WARNING into a false alarm on a routine code path. Fixing it here is necessary for the budget change to be safe, not incidental.

### Generator semantics check

I read `class_set_retry_attempts` for the two ways this shape usually goes wrong:

- The first attempt always yields before any deadline test, so a zero budget still performs one read. Retry is an addition to the read, never a precondition for it.
- The sleep is after the `yield`, so a caller that `return`s on a first-read hit closes the generator without ever sleeping. Zero added cost on the hit path, as claimed.
- The loop variable is read after the loop (`attempts`), which is only safe because the generator is guaranteed to yield at least once. It is. The `attempts = 0` pre-initialization plus `noqa: B007` makes the intent explicit rather than accidental.
- `time.monotonic()` rather than `time.time()`, so a clock adjustment cannot extend or collapse the budget.

### Verification

```
tests/unit/test_class_set_retry.py + test_sdlc_stage_query.py
  + test_valor_session_cli.py + test_valor_session_enumeration_seam.py   105 passed
```

Confirmed `time` is still imported and used elsewhere in `tools/valor_session.py` after the `time.sleep` removal, so the import is not orphaned.

### Non-blocking observations

1. The UUID fast path rests on the assumption that no `session_id` scheme produces 32 lowercase hex characters. The code comment names this assumption and it inherits from `_find_session`'s pre-existing docstring rather than introducing it, and `.fullmatch()` is correctly used so a longer string cannot partially match. Real but pre-existing, and correctly disclosed.
2. The stated tradeoff is genuine: a truly absent session now costs up to 3s on the miss path instead of 1s. At operator/dispatch invocation rates that is the intended trade, and it is confined to the non-hot-path readers, since worker recovery and steering delivery remain excluded per #1720.
3. The ~75 unretried `query.filter(session_id=...)` call sites are correctly left out of scope. Each needs a per-site judgment on whether a false "not found" is load-bearing, and the shared `tools/class_set_retry.py` makes that follow-up cheap. Worth its own issue.

No blockers.

tomcounsell:
## Post-merge review (#2583 merged before I finished the gate)

Assigned this as merge-queue slot 3; it landed at 18:56:37 while I was still measuring. Recording what I found, since two items are actionable.

### The design call is right

Replacing a re-measured constant with a wall-clock budget is the correct answer to the issue's fourth checkbox. A constant sized at 4000 rows goes stale at 40000 exactly as the 150-row one did. The PR body's honesty about **not** reproducing at fleet scale, and declining to create 4000 synthetic sessions in shared Redis with other agents live, is the right call and well documented.

The UUID fast path is a genuine bug fix. Measured on the branch:

| Call | Before | After |
|---|---|---|
| `_find_session`, UUID form | ~1s (burned the whole cap) | **0.02s** |

### Regression: two integration tests broke on main

The PR removed `_CLASS_SET_RETRY_ATTEMPTS` / `_CLASS_SET_RETRY_BACKOFF_S` but left `tests/integration/test_class_set_rebuild_retry.py` monkeypatching them. Bisected to this merge: **12 passed** at `1d9a26c1`, **10 passed / 2 failed** at `7123245c`.

Filed as #2591, fix in #2592.

### Genuine misses now cost 3s, and log WARNING every time

Measured on the branch, this host:

```
_find_session       (genuine miss, session_id form) -> None in 3.03s
_find_session_by_id (genuine miss)                  -> None in 2.95s
```

Both emit the exhaustion WARNING. That is a 3x latency increase on the genuine-absence path (was ~1s), which is defensible at operator CLI rates. The logging is the part worth revisiting.

`_find_session` already computes the discriminator: `resolved_by_fallback` is `True` only when `get_by_id` finds the row, which **proves** a mid-rebuild class set. When the fallback also misses, that is strong evidence of genuine absence, and it still logs WARNING. So the normal "you asked for a session that does not exist" case produces a warning that says a rebuild may have happened.

This matters because the PR's stated goal is making exhaustion visible. A WARNING that fires on every ordinary miss is one people learn to skip, which is the same failure mode the silent `None` had, just louder. Suggestion: keep WARNING for `resolved_by_fallback=True` (proven stale index) and drop the no-fallback path to INFO or DEBUG. `_find_session_by_id` has no fallback so it has no proof either way; the message correctly says the two causes are indistinguishable, but it will still fire on every miss.

Not filing this one as an issue since it is a judgment call the author is better placed to make than I am.

### Smaller notes

- `log_class_set_exhaustion` re-reads `settings.timeouts.class_set_retry_budget_s` rather than the budget actually used, so a caller passing an explicit `budget_s` to `class_set_retry_attempts()` gets a message quoting the default. No caller does this today, but it makes an observability message capable of stating a number that was not used.
- `_AGENT_SESSION_ID_RE` is lowercase-only, so an uppercase-hex argument falls to the retry path and spends the full budget. `AutoKeyField` emits lowercase, and the comment states the assumption, so this is fine as-is.
- `gt=0.0` on `class_set_retry_backoff_s` correctly rules out a zero-backoff busy-loop against Redis. Good constraint to have picked.

### Gates

`tests/unit/test_class_set_retry.py` 10 passed. Budget-zero and budget-exhaustion edges behave as documented: one attempt at budget 0, ~14 attempts at the 3s default.

PR #2584, opened 2026-08-06
Fix the Verification-table pipe split; reject unreadable rows instead of truncating (#2570)

### LINKED ISSUE #2570: Verification-table parser splits rows on | with no escaping: any pipe-bearing command is silently truncated and misattributed
`/do-build` executes the `## Verification` table from a plan document via `agent/verification_parser.py`. That parser splits each table row on `|` with no escape handling, so **any command containing a pipe character is silently truncated** — and the truncation also destroys the row's `Expected` cell.

The result is a check that runs a different command than its author wrote, fails for an unrelated reason, and reports that failure against the code under test.

## Mechanism

`agent/verification_parser.py:69`:

```python
cells = [c.strip() for c in row.split("|")]
```

A three-column row whose command contains a pipe yields four cells. Cell 1 (command) is truncated at the pipe; cell 2 (expected) receives the fragment *after* the pipe; the real expectation lands in cell 3 and is discarded.

Worked example, run against the real parser:

```
row:  | Anti-criterion: sdlc not hardcoded | `! grep -nE '"sdlc/"\|/ "sdlc"' scripts/update/hardlinks.py` | exit code 0 |

cmd  parsed as:  '! grep -nE \'"sdlc/"\\'
exp  parsed as:  '/ "sdlc"\' scripts/update/hardlinks.py`'
exit: 2   (/bin/sh: unexpected EOF while looking for matching `'')
verdict: FAIL
```

The garbage expectation matches none of `_evaluate`'s patterns and hits its terminal `return False` (`verification_parser.py:165`).

## Escaping does not help — there is no working form

Tested all three candidate forms end to end through the parser:

| Row form | Command as parsed | Verdict |
|---|---|---|
| `'foo\|bar'` — backslash-escaped for Markdown | truncated at the pipe | FAIL |
| `'foo|bar'` — unescaped | truncated at the pipe | FAIL |
| no pipe at all | intact | PASS |

**Both escaped and unescaped pipes truncate.** There is currently no way to express a pipe-bearing command in a Verification row. That silently rules out:

- regex alternation (`grep -E 'a|b'`)
- shell pipelines (`cmd | wc -l`) — note `docs/features/machine-readable-dod.md` documents `grep -r ... | wc -l` as a *canonical idiom* for the `match count == 0` expectation, so the documentation recommends a shape the parser cannot execute
- Python bitwise-or (`re.S|re.M`)

## Why this went unnoticed: the failure is misattributed, not loud

The check fails, so this is not a fail-open hole — `_evaluate`'s terminal `return False` is good design and is what contains the blast radius. The harm is that the red is attributed to the **code under test** rather than to the malformed row. An author sees a named check failing with a shell syntax error and reasonably concludes the implementation is wrong. The plausible outcomes are someone "fixing" working code, or deleting a legitimate check because it appears broken.

## Evidence from a real plan

Four separate traps were hit while authoring one Verification table for #2561, three of them pipe-related. All four are recorded in `docs/plans/sync-global-hook-import-siblings.md` under "Why these rows are shaped this way":

1. `grep -E 'a\|b'` matches a *literal* `a|b` — a backslash-escaped pipe is not alternation under `-E`.
2. `glob` is a substring of `global`, so a proxy grep counted 50 hits on unmodified `main` and proved nothing.
3. `sdlc` appears elsewhere in the file (`RENAMED_REMOVALS`, `hardlinks.py:174`), so whole-file greps false-positive; rows had to be scoped to a function body.
4. **This issue** — found only when the row was executed by the runner rather than by hand.

Trap 4 is the one an author cannot discover by testing their command in a shell, because the command works fine there. It only manifests through the parser. Parse-time rejection would have caught it at authoring time.

Notably, the plan critique's prescribed remedy for trap 1 ("drop the backslash for true alternation") would **also** have failed, for trap 4's reason. Two review passes and a build pass went by before the row was executed through the real parser.

## Proposal

**1. Fail closed on a malformed row, loudly.** At parse time, reject any row whose command cell contains a pipe that cannot be unambiguously attributed to the command, and surface it as a *plan authoring error* distinct from a check failure. A row that cannot be parsed as intended must never be silently executed as something else.

**2. Provide a real escape mechanism**, so the ruled-out command classes become writable again. Options worth weighing in the plan:
- honor a backslash escape (`\|`) by unescaping it after the split, which makes the Markdown-safe form also the correct form
- split on a column count from the header rather than greedily on every `|`
- allow a fenced/quoted command cell whose interior is not split

**3. Correct `docs/features/machine-readable-dod.md`**, which currently recommends `grep -r ... | wc -l` as a canonical idiom the parser cannot run.

## Acceptance criteria

- [ ] A Verification row whose command contains a pipe is either executed correctly or rejected as an authoring error — never silently truncated
- [ ] A rejected row is reported distinguishably from a failed check, naming the row and the reason
- [ ] At least one of regex alternation, a shell pipeline, and `re.S|re.M` is expressible in a Verification row — asserted by test
- [ ] The three forms in the table above are covered by regression tests, including the currently-passing pipe-free control
- [ ] `docs/features/machine-readable-dod.md`'s `| wc -l` idiom either works or is replaced

## Related

- #2561 / PR #2569 — where all four traps were found; its plan carries the trap subsection and the red-state proof table
- #2521 — closed premise-disproved; its plan carried an earlier `\|` row with the same defect (as basic-regex `grep`, where `\|` *is* valid alternation, so that one worked by luck)
- `agent/verification_parser.py:69` — the split
- `agent/verification_parser.py:165` — the terminal `return False` that keeps this fail-closed

tomcounsell:
Closes #2570

## Problem

`agent/verification_parser.py:69` split every table row on every `|`. A pipe-bearing command was truncated at the pipe, the `Expected` cell received the fragment *after* it, and the real expectation was discarded. The check then ran a command its author never wrote, failed for an unrelated reason (usually a shell syntax error), and reported that failure **against the code under test**.

Escaping did not help. `\|` and a bare `|` truncated identically, so there was no working form at all — regex alternation, shell pipelines, and `re.S|re.M` were all unwritable. The failure was misattributed rather than fail-open, which is why it survived: a persistently red check that looks broken gets ignored or deleted.

## Fix

Rows split on **unescaped** `|` only, and `\|` is unescaped to a literal pipe afterwards. That is standard GitHub-flavored Markdown, so the Markdown-safe way to write a pipe is also the correct way.

The escape composes, which matters because most anti-criteria use basic-regex `grep -c` alternation:

| In the table cell | Reaches the shell as | `grep -E` | `grep -c` (BRE) |
|---|---|---|---|
| `a\|b` | `a\|b` | alternation | literal |
| `a\\\|b` | `a\\|b` | literal | alternation |
| `a\|b` unescaped | rejected as malformed | — | — |

A row that still does not yield the header's column count is reported as a `MalformedRow` and **never executed**. `parse_verification_table` returns `ParsedTable(checks, malformed)`; `format_results` reports authoring errors in their own section headed "fix the plan, not the code" and fails the run, because a row nobody can execute is not a passing check. The column count comes from the header rather than a hardcoded 3, so the one 4-column table in `docs/plans/` parses instead of having every row rejected.

## Two corrections to the issue as filed

1. **The same defect was in a second parser.** `scripts/validate_build.py` has its own `parse_verification_table`, and `do-build` Step 14 runs it as a *blocking* gate (exit 1 routes to `/do-patch`). It now shares `split_row_cells` with the runner, so the validator and the runner cannot disagree about what a row says.
2. **The `| wc -l` idiom lives in the parser docstring**, not `docs/features/machine-readable-dod.md` (the issue comment already caught this). Both are corrected: the docstring shows the escaped form, and that feature doc's two canonical anti-criterion rows are corrected to `\\|` — verified to round-trip through the parser and pass.

## Measurement

| Measure | Count |
|---|---|
| Rows with the truncation signature under the old split | 544 (reproduces the issue's 545) |
| Recovered by escape-aware splitting | 508 |
| Still malformed (genuinely bare `\|`), now rejected loudly | 36 across 28 plans |

Of the 31 recovered rows in **active** plans, 10 fail — all in plans whose work is not yet merged, where red is the expected state. None revealed a hidden violation in merged code.

## Tests

`tests/unit/test_verification_parser.py` and `tests/unit/test_validate_build.py`, 88 passing including `test_architectural_constraints`. Covers all three forms from the issue's table (backslash-escaped, unescaped, pipe-free control), the issue's verbatim worked example, each previously-ruled-out command class (alternation, pipeline, `re.S|re.M`), and header-driven column counts.

`test_malformed_verification_table` previously asserted exit 0 — "nothing parseable, so nothing to say". That silence was the defect; it now asserts the loud failure.

### DIFF
--- a/agent/verification_parser.py
+++ b/agent/verification_parser.py
@@ -11,6 +11,39 @@ Plan documents contain a ``## Verification`` section with a markdown table:
 
 This module extracts those rows into ``VerificationCheck`` objects and provides
 ``evaluate_expectation`` to decide pass/fail based on command results.
+
+Pipes in commands (#2570)
+-------------------------
+Rows are split on **unescaped** ``|`` only, and ``\\|`` is unescaped to a
+literal ``|`` afterwards. That is standard GitHub-flavored Markdown, and it
+makes the Markdown-safe way to write a pipe also the correct way:
+
+    | Anti-criterion | `grep -rc 'a\\|b' src/ \\| wc -l` | match count == 0 |
+
+runs ``grep -rc 'a|b' src/ | wc -l``. Before this, every row was split on every
+``|``, so a pipe-bearing command was truncated at the pipe, the ``Expected``
+cell received the fragment *after* it, and the real expectation was discarded.
+The check then ran a command its author never wrote, failed for an unrelated
+reason (usually a shell syntax error), and reported that failure against the
+code under test. Escaping did not help: both ``\\|`` and a bare ``|``
+truncated, so there was no working form and regex alternation, shell
+pipelines, and ``re.S|re.M`` were all unwritable.
+
+A row that still does not yield the header's column count -- an author wrote a
+bare ``|`` and meant it as part of the command -- is now reported as a
+:class:`MalformedRow` rather than executed as something else. Silently running
+a different command than the one on the page is the failure this module exists
+not to have.
+
+The escape composes, which matters for basic-regex ``grep``: in a BRE,
+alternation is spelled ``\\|``, and to get that through the table you double
+the backslash. What lands in the shell is one level of unescaping::
+
+    table cell        reaches the shell as     grep -E / grep -c meaning
+    ---------------   ----------------------   -------------------------
+    a\\|b              a|b                      -E: alternation. -c: literal "a|b"
+    a\\\\|b             a\\|b                     -c (BRE): alternation
+    a|b               (rejected: malformed)    --
 """
 
 from __future__ import annotations
@@ -18,6 +51,25 @@ from __future__ import annotations
 import re
 import subprocess
 from dataclasses import dataclass
+
+# Split on a `|` that is not backslash-escaped. A row's cells are the pieces
+# between these; `\|` inside a cell survives the split and is unescaped after.
+_UNESCAPED_PIPE_RE = re.compile(r"(?<!\\)\|")
+
+
+def split_row_cells(row: str) -> list[str]:
+    """Split one markdown table row into its cells, honoring ``\\|`` escapes.
+
+    Leading/trailing empties from the row's border pipes are dropped; interior
+    empties are preserved, because an empty cell is a real (if unusable) cell
+    and collapsing it would shift every column after it.
+    """
+    parts = _UNESCAPED_PIPE_RE.split(row)
+    if parts and not parts[0].strip():
+        parts = parts[1:]
+    if parts and not parts[-1].strip():
+        parts = parts[:-1]
+    return [p.replace("\\|", "|").strip() for p in parts]
 
 
 @dataclass(frozen=True)
@@ -27,6 +79,26 @@ class VerificationCheck:
     name: str
     command: str
     expected: str
+
+
+@dataclass(frozen=True)
+class MalformedRow:
+    """A table row that could not be read as the author wrote it.
+
+    Reported separately from a failed check: this is a plan-authoring error,
+    not evidence about the code under test.
+    """
+
+    line: str
+    reason: str
+
+
+@dataclass(frozen=True)
+class ParsedTable:
+    """Everything a ``## Verification`` table yielded, including its rejects."""
+
+    checks: list[VerificationCheck]
+    malformed: list[MalformedRow]
 
 
 @dataclass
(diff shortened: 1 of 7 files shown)


tomcounsell:
## Review: APPROVED

Verified against the merge of `session/ws-2570` onto `origin/main` (3216c2d9).

**The fix is at the right layer.** `split_row_cells` splits on `(?<!\\)\|` and unescapes `\|` afterwards, so the Markdown-safe spelling and the correct spelling are the same string. That is the property that makes this stick: the previous state had no working form at all, which is why the broken rows propagated instead of being worked around.

**Malformed rows are now evidence about the plan, not the code.** This is the part that mattered most in the issue. `MalformedRow` carries the raw line and a reason, `format_results` prints it under "Fix the plan, not the code" above the checks, and `all_passed` is ANDed with `not malformed`, so an unreadable row fails the run instead of vanishing. The old `continue`-on-short-row was a silent drop; the new behavior is loud in the right direction.

**Both parsers were fixed, and they now share one splitter.** `scripts/validate_build.py` had an independent copy of this defect and runs as a *blocking* `do-build` Step 14 gate. It now imports `split_row_cells` from `agent.verification_parser`, so the validator and the runner cannot disagree about what a row says. Fixing only the one named in the issue would have left the blocking gate broken.

**Caller audit: no breakage.** `parse_verification_table` changed return type from `list[VerificationCheck]` to `ParsedTable`, which is a breaking signature change, so I checked every call site. The only non-test consumers are the inline `python -c` snippets in `docs/sdlc/do-build.md` and `docs/sdlc/do-pr-review.md`, and both are updated in this diff to unpack `t.checks` / `t.malformed` and to exit non-zero on malformed. In `scripts/validate_build.py`, `check_verification_table` is the single consumer of the parsed list and tests `check.get("malformed")` before touching `command`, so a malformed entry with an empty command cannot reach execution. `format_results` gained an optional second parameter, so it stays backward-compatible.

**Header-driven column count is a real fix, not incidental.** `expected_columns = max(len(split_row_cells(rows[0])), 3)` is what stops the one 4-column table in `docs/plans/` from having every row rejected. The `max(..., 3)` floor keeps a malformed 2-column header from silently lowering the bar.

### Independent verification

I did not take the measurement section on faith. Running the new parser across `docs/plans/**/*.md` myself:

```
parsed checks total:                                 3894
checks containing a literal pipe (recovered)          501
malformed rows now rejected:            36 across 28 plans
```

The malformed count and plan count match the PR's claim exactly. My recovered count of 501 differs from the stated 508 only because I counted checks whose command *or expected* cell contains a pipe post-unescape rather than rows carrying the old truncation signature. The claim holds.

```
tests/unit/test_verification_parser.py + test_validate_build.py   82 passed
python -m ruff format --check . 1275 files already formatted
```

`test_malformed_verification_table` flipping from asserting exit 0 to asserting a loud failure is the correct test change: the old assertion pinned the defect.

### Non-blocking observations

1. A cell whose content genuinely ends in a trailing backslash immediately before a separator pipe (`... \|`) will read that separator as escaped and the row will be rejected as malformed. This is the standard GFM escaping edge and it now fails loudly rather than silently, which is the right side to fail on. Not worth extra machinery.
2. A row with *more* cells than the header is now malformed where it was previously accepted by taking the first three. Stricter and deliberate, consistent with the "never execute a row we cannot read" principle, but it is a behavior change beyond the issue's literal scope. The 36-row rejection set already accounts for it.
3. `scripts/validate_build.py` gains a `sys.path.insert` to import from `agent/`. Necessary for the shared splitter and it keeps the script runnable standalone.

No blockers.

PR #2585, opened 2026-08-06
Split the delete-and-recreate helper along its two contracts (#2563)

### LINKED ISSUE #2563: _extract_agent_session_fields serves two incompatible contracts: recreate-same-session loses 50 fields, continuation would forge an execution fence

tomcounsell:
Closes #2563

## Verified before fixing

The 50-field drift is exact. Against `2e517d3b`:

```
total model fields: 89   (auto: ['id'])
listed in _AGENT_SESSION_FIELDS: 38
MISSING: 50 — issue_number, pr_number, model, task_type, project_config,
  retain_for_resume, budget_tripped, budget_tripped_reason, crash_signature,
  total_cost_usd, total_input_tokens, total_output_tokens, total_cache_read_tokens,
  thread_turn_count, thread_run_count, thread_tool_call_count, thread_first_created_at,
  exec_pid, pid_create_time, exec_cwd, exec_harness, spawn_history, active_run_id,
  owned_run_ids, worker_pid, ...
```

All four call sites confirmed, and they split exactly as the issue describes.

## Why this cannot be a list edit

Group 1 (`agent/session_health.py` orphan repair, `tools/agent_session_scheduler.py` priority bump) recreates the **same** session, so every field absent from the list is destroyed. Group 2 (`retry_agent_session`, `agent/session_executor.py` auto-continue fallback) creates a **new execution** of the same `session_id`, where the list's omission of the fence fields was accidentally protective.

Adding the missing 50 would close the Group 1 loss and simultaneously forge fences in Group 2. A continuation row is `pending` and therefore non-terminal, so it is visible to `find_live_session_by_pid`'s ownership scan; a copied `exec_pid` makes it claim a process that never ran for it.

## The split

| Helper | Contract | Call sites |
|---|---|---|
| `clone_agent_session_fields` | Same session, same live process. Copy everything. | orphan repair, priority bump |
| `continuation_agent_session_fields` | New execution of the same `session_id`. Copy history, reset the fence. | `retry_agent_session`, auto-continue fallback |

Both derive from `AgentSession._meta` at runtime via `_copyable_agent_session_fields()`, which retires the drift class rather than re-freezing it.

`_EXECUTION_FENCE_RESET_FIELDS` carries a one-line rationale per entry:

```python
"exec_pid",         # OS pid of a harness subprocess this row never spawned.
"pid_create_time",  # The PID-reuse fence for that pid; meaningless without it.
"exec_cwd",         # Working dir that spawn ran in; resume is cwd-scoped.
"exec_harness",     # Which harness ran it; the new run picks its own.
"spawn_history",    # Append-only spawn timeline of the previous run.
"active_run_id",    # Names the run that just ended, not the one being queued.
"owned_run_ids",    # Runs the previous execution owned.
"worker_pid",       # Worker process that owned the previous execution.
```

**Scope call worth flagging:** this set is deliberately the fence and run identity, **not** a general freshness reset. `last_heartbeat_at`, `last_sdk_heartbeat_at` and `last_stdout_at` are in today's 38-entry list and therefore already carry onto continuations; they keep doing so. Clearing them is arguably also right, but it changes watchdog inputs and belongs to a separate decision rather than riding in here. `test_reset_set_is_the_fence_and_run_identity` pins the set explicitly so widening it is a decision, not a refactor.

Fields reset to each field's **declared default** rather than a hardcoded `None`, so this stays correct if a field's unset state ever stops being null.

## Bug found while wiring this up

Several tests patch `agent.agent_session_queue.AgentSession` with a mock to intercept `create`. Deriving the schema off that name would read the mock's `_meta` and silently produce an **empty** payload. `_agent_session_model_fields()` imports the model directly, so schema derivation is independent of module-level patching. This surfaced as two real red tests, not a hypothetical.

## Relationship to #2553

`KNOWN_GAP` is **removed, not emptied**. The issue anticipated it dropping to empty, but that assumed the hand-maintained list survived. It does not, so the drift class the constant recorded cannot recur and a frozen-empty set would guard nothing. Its guard is replaced by assertions on the two new contracts.

## Verification

Round trip against real Redis (`dbg-2563` project key, ORM-scoped, cleaned up):

```
derived copy-set size: 88            (was 38)
CLONE keeps history: 2563 99 claude-opus-5 1.25 7
CLONE keeps fence:   424242 [{'pid': 424242}]
CONT keeps history:  2563 1.25 7
CONT clears fence:   all 8 -> None
CLONE create OK -> issue_number=2563 exec_pid=424242
CONT  create OK -> issue_number=2563 exec_pid=None
remaining rows: 0
```

```
8 affected unit test files    357 passed
4 affected integration files   89 passed
```

Baselined against `origin/main` on the same 8-file set (350 passed) before and after.

New coverage includes the assertion pair the issue asks for: `test_clone_preserves_exec_pid` and `test_continuation_clears_exec_pid`, plus `test_previously_lost_fields_are_now_copied` naming the fields the 38-entry list destroyed.

## DOCS

- `docs/features/agent-session-queue.md` — new "Two contracts, two field payloads (#2563)" section with the call-site table, the reset set and its rationale; code sample and orphan-children bullet updated.
- `docs/features/agent-session-model.md` — continuation now notes the fence clear.
- `docs/features/bridge-self-healing.md` — the "all fields are in `_AGENT_SESSION_FIELDS`" claim replaced with the derivation, cross-linked.

### DIFF
--- a/agent/agent_session_queue.py
+++ b/agent/agent_session_queue.py
@@ -163,62 +163,113 @@ _SESSION_TYPE_UNSET = object()
 PRIORITY_RANK = {"urgent": 0, "high": 1, "normal": 2, "low": 3}
 
 
-# Fields to extract from AgentSession for delete-and-recreate pattern.
-# Used by callers that legitimately need a fresh AutoKeyField-generated ID:
-# retry, orphan fix, and the continuation fallback in _enqueue_nudge. The pop
-# path (_pop_agent_session) does NOT use this — it mutates in place via
-# transition_status() because status is an IndexedField, not a KeyField, and
-# the secondary index is updated correctly on save().
-_AGENT_SESSION_FIELDS = [
-    "project_key",
-    "status",
-    "priority",
-    "scheduled_at",
-    "created_at",
-    "session_id",
-    "working_dir",
-    "initial_telegram_message",
-    "chat_id",
-    "extra_context",
-    "task_list_id",
-    "auto_continue_count",
-    "started_at",
-    "telegram_message_key",
-    "updated_at",
-    "completed_at",
-    "turn_count",
-    "tool_call_count",
-    "log_path",
-    "branch_name",
-    "tags",
-    "session_events",
-    "issue_url",
-    "plan_url",
-    "pr_url",
-    "context_summary",
-    "correlation_id",
-    "claude_session_uuid",
-    "parent_agent_session_id",
-    "session_type",
-    "slug",
-    "pm_sent_message_ids",
-    "recent_sent_drafts",
-    "last_heartbeat_at",
-    "last_sdk_heartbeat_at",
-    "last_stdout_at",
-    "recovery_attempts",
-    "reprieve_count",
-]
+# ---------------------------------------------------------------------------
+# Delete-and-recreate field sets (issue #2563)
+# ---------------------------------------------------------------------------
+# Four call sites delete an AgentSession row and create a replacement, because
+# KeyFields cannot be mutated in place without corrupting their index. They
+# split into two contracts that no single field list can serve at once:
+#
+#   CLONE — the recreated row IS the same session. Orphan repair
+#   (agent/session_health.py) clears a KeyField; the scheduler's priority bump
+#   (tools/agent_session_scheduler.py) changes one value. Both describe the same
+#   live process, so anything not copied is destroyed.
+#
+#   CONTINUATION — the recreated row is a NEW execution of the same session_id.
+#   retry_agent_session below and the auto-continue fallback in
+#   agent/session_executor.py both produce a `pending` row that no process has
+#   ever run. History carries; the execution fence must not.
+#
+# A hand-maintained list served both and was wrong for both: it had drifted 50
+# fields behind the model, silently destroying issue_number, pr_number, model,
+# the token/cost accounting and the thread counters on every clone, while its
+# omission of the fence fields was load-bearing-by-accident for continuation.
+# Deriving the copy set from the model removes that drift class permanently.
+#
+# The pop path (_pop_agent_session) uses neither — it mutates in place via
+# transition_status(), since status is an IndexedField and save() maintains the
+# secondary index correctly.
+
+# Fields cleared on the CONTINUATION path. Each describes ONE execution, so
+# copying it onto a row that has never run forges a record of a process that
+# never ran for it — the exact forged-liveness failure the #2518 durability
+# fence exists to prevent, since a `pending` row is non-terminal and therefore
+# visible to find_live_session_by_pid's ownership scan.
+#
+# This set is deliberately the execution fence and run identity, NOT a general
+# freshness reset. The heartbeat and liveness timestamps carry, as they always
+# have; changing those has watchdog blast radius and is a separate decision.
+_EXECUTION_FENCE_RESET_FIELDS = frozenset(
+    {
+        # The fenced execution record (docs/features/agent-session-fenced-execution-record.md).
+        "exec_pid",  # OS pid of a harness subprocess this row never spawned.
+        "pid_create_time",  # The PID-reuse fence for that pid; meaningless without it.
+        "exec_cwd",  # Working dir that spawn ran in; resume is cwd-scoped.
+        "exec_harness",  # Which harness ran it; the new run picks its own.
+        "spawn_history",  # Append-only spawn timeline of the previous run.
+        # Run identity.
+        "active_run_id",  # Names the run that just ended, not the one being queued.
+        "owned_run_ids",  # Runs the previous execution owned.
+        "worker_pid",  # Worker process that owned the previous execution.
+    }
+)
 
 
-def _extract_agent_session_fields(redis_session: AgentSession) -> dict:
-    """Extract all non-auto fields from an AgentSession instance.
+def _agent_session_model_fields() -> dict:
+    """The real ``AgentSession._meta.fields``, independent of module-level patching.
 
-    Returns a dict suitable for AgentSession.create(**fields) or
-    AgentSession.async_create(**fields). Excludes agent_session_id since that is
-    an AutoKeyField and will be auto-generated on create.
+    Schema derivation is a different concern from row construction. Several tests
+    patch this module's ``AgentSession`` name with a mock to intercept ``create``;
+    reading the schema off that mock would yield an empty field set and silently
+    produce empty payloads. Importing the model here keeps the derivation reading
+    the actual schema in every context.
     """
-    return {field: getattr(redis_session, field) for field in _AGENT_SESSION_FIELDS}
+    from models.agent_session import AgentSession as _AgentSessionModel
+
+    return _AgentSessionModel._meta.fields
+
+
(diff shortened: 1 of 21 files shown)


tomcounsell:
## Review: Approved

Verified the two contracts by executing them against a constructed session rather than reading the diff alone.

### The split is correct, and the reasoning for it is the load-bearing part

The insight worth calling out: the old list was wrong in *both* directions at once, and a list edit could only ever fix one. Adding the missing 50 would have closed the clone-path loss and simultaneously forged fences on the continuation path. That is why this had to become two helpers, and the PR body argues it exactly right.

Exercised both, with a source session carrying a real fence and real history:

```
CLONE keeps the fence (same live process):
  exec_pid: 44321 | active_run_id: run-OLD | worker_pid: 999

CONTINUATION clears the fence:
  non-null fence fields: NONE — fence fully reset

CONTINUATION preserves history/identity:
  session_id 'sess-abc' | issue_number 2563 | pr_number 2585 | model 'opus'
  total_cost_usd 1.23 | turn_count 7 | thread_turn_count 9 | branch_name 'session/x'
```

### Drift claim is exact

```
total model fields: 89   AutoKeyFields: ['id']
copyable: 88            old hand-maintained list: 38
50 fields now preserved that were previously destroyed on every clone
  e.g. response_delivered_at, last_authored_at, thread_first_created_at,
       thread_turn_count, thread_run_count, task_type, rework_triggered
```

### The key-exclusion question

The one thing that could have gone quietly wrong is the derived set including the key and defeating the fresh-AutoKey purpose. It does not: `agent_session_id` is not a model field at all, `id` is the sole `AutoKeyField` and is excluded, and the five non-auto `KeyField`s (`session_type`, `project_key`, `chat_id`, `slug`, `parent_agent_session_id`) are values that *should* carry. Confirmed `'id' in copyable == False`.

### Call-site routing

All four route as the table claims, and the retired helper and list are gone from source entirely:

| Contract | Sites |
|---|---|
| `clone_agent_session_fields` | `agent/session_health.py:4588`, `tools/agent_session_scheduler.py:887` |
| `continuation_agent_session_fields` | `agent/agent_session_queue.py:837` (retry), `agent/session_executor.py:593` |

### Gates

| Check | Result |
|---|---|
| 8 unit suites | **357 passed** |
| 6 integration suites | **98 passed** |
| `ruff check` / `format --check` | clean / 1273 files |
| `uv lock --locked` | PASS |
| `tools/check_fence_census.py` | clean, 11 fenced-pid consumers all guarded |

The fence census passing matters more than usual here, since the fence is precisely what this PR redistributes.

### Two observations, neither blocking

1. All eight `_EXECUTION_FENCE_RESET_FIELDS` currently declare `default=None`, so reading the declared default is presently identical to hardcoding `None`. The comment already frames it as future-proofing and I agree with the choice; noting it so nobody reads the mechanism as load-bearing today.
2. `tests/unit/test_pipeline_integrity.py` lines 200 and 220 still name `_AGENT_SESSION_FIELDS` in prose. Given the repo's no-historical-artifacts rule that could read as a leftover, but here it is the rationale for why the invariant is pinned, which is worth keeping. Mentioning it only so a future cleanup pass does not strip the explanation by reflex.

The scope call on `last_heartbeat_at` / `last_sdk_heartbeat_at` / `last_stdout_at` is right. Those change watchdog inputs, and `test_reset_set_is_the_fence_and_run_identity` pinning the set means widening it later is a decision rather than a drive-by.

### Note on process

Reviewed from a detached worktree at `687a18e9` rather than checking out `session/ws-fleet-2563`, since `.worktrees/ws-fleet` holds that branch. Nothing of mine touched it.

Issue #2574, opened 2026-08-06
Full tests/unit/ run wedges at 99%: xdist worker dies, controller hangs at 0% CPU with no summary (and misattributes the crash to the wrong test)


valorengels:
## Problem

A full `tests/unit/` run dies near the end of the suite: an xdist worker goes down with `node down: Not properly terminated`, the run stops producing output, and the controller never reaches the summary. The controller process stays alive at 0% CPU indefinitely and must be killed by pid.

Reproduced **3 times** on 2026-08-06, and it **predates any single branch's diff**.

## Signature

- Worker death at **99%** every time, but a **different worker** each run: `gw5`, then `gw8`, then `gw4`.
- On one run the controller crashed in `pytest_sessionfinish` with `OSError: cannot send (already closed?)`.
- The run never emits the `-rf` summary, so a wedged run yields **no failure list at all** unless `-v` was passed.

## The attribution trap — read this before investigating

When an xdist worker dies, pytest reports **the first item collected by that worker**, not the item it was actually executing. This is actively misleading and it misled two separate investigations today.

Both wedged runs reported `test_session_executor_runner_dispatch.py::TestExecutorGuardEmptyTurnInput::test_empty_turn_input_triggers_guard[none]` as `FAILED` at 99%. Reading the full `-v` log rather than the tail shows that **the same test ran at 26% and PASSED on the very worker that later died** (`/tmp/wave2-suite-v.log:6085` PASSED at 26%, then `:22962` FAILED at 99%, same `gw4`). The `Terminated: 15` on that line is an operator SIGTERM during cleanup, not the test's own signal.

**The crashing test is therefore unidentified.** Any investigation that starts from the reported node name will chase the wrong test.

A hypothesis worth checking, but not yet supported: `worker/__main__.py:101` `_self_kill` does `os.kill(os.getpid(), ...)`. A dispatch test reaching a self-kill path would kill its own xdist worker and match the signature. This was proposed and is *not* confirmed — the moving worker/node identity argues against a single poisonous test.

## Why it is not attributable to any current branch

- Reproduced 3/3 on the PR #2568 worktree, including once with that PR's autouse fixture reduced to a bare `yield` via its `CATCHUP_FLAG_ISOLATION=0` escape hatch — so the fixture's per-test behavior is not the mechanism.
- #2536 recorded **4/4 branch-worktree hangs at 77-99%** on unrelated branches while the main checkout completed.
[...]

Issue #2551, opened 2026-08-06
test_reap_killlist.py fails only when run alongside test_session_lifecycle.py / test_recovery_ownership.py — pre-existing cross-file isolation defect verified on main (from PR #2538)


valorengels:
> **Context for readers unfamiliar with this area.** `agent/reap_killlist.py` is the durable cross-boot "kill list" for wedge-reap survivors (issue #2146): when a session's teardown cannot confirm that a tool subtree died, the surviving PIDs are persisted to a Redis hash so the worker's orphan reaper can `SIGKILL` them later, guarded by a `create_time` match so a recycled PID is never collaterally killed. Its unit tests at `tests/unit/test_reap_killlist.py` never touch real Redis or real processes: they patch the module's `_redis` seam to a hash-backed fake. This issue is about those tests failing for reasons unrelated to the code under test.

## Problem

Two tests in `tests/unit/test_reap_killlist.py` fail when the file is run in the same pytest process as `tests/unit/test_session_lifecycle.py` and/or `tests/unit/test_recovery_ownership.py`. The same file passes when run alone. The failure is **not** a regression from any recent change — it reproduces identically on `origin/main`.

**Current behavior**

- `pytest tests/unit/test_reap_killlist.py` alone: passes.
- `pytest tests/unit/test_reap_killlist.py tests/unit/test_session_lifecycle.py tests/unit/test_recovery_ownership.py`: two failures in `test_reap_killlist.py`.
- The same combination on `origin/main`: the same two failures.

**Desired outcome**

`tests/unit/test_reap_killlist.py` passes deterministically regardless of which other test modules share its pytest process, so its result is a signal about `agent/reap_killlist.py` and nothing else.

## Evidence

This was verified during the Durability M1 fence work and recorded only in the body of PR #2538 ("Durability M1 fence: close nine unfenced consumers + permanent regression tests"), which stated verbatim:

> Two `test_reap_killlist.py` failures appear only when run alongside `test_session_lifecycle.py`/`test_recovery_ownership.py`. The same combination fails identically on `origin/main` and the file passes alone on both — **pre-existing cross-file isolation defect, verified on main, not assumed.**

That PR correctly scoped the defect out of its own work and merged. The finding has lived only in a PR body since; this issue gives it a durable home.

PR #2538 did not record the two failing test names or their assertion output. Reproducing the combination and capturing that output is the first task of any plan for this issue.

## Suspected shared state

[...]

PR #2586, opened 2026-08-06
Repair the agent module tree by rebinding, not by evicting (#2558)

### LINKED ISSUE #2558: Named: agent_hooks_consistency_guard sys.modules eviction makes mock.patch target a fresh module while the test's collection-time binding holds the old one
## Root cause

This names the module-identity divergence that #2462 Group B could only hypothesize.

The autouse fixture `agent_hooks_consistency_guard` (`tests/conftest.py:222-271`) deletes every `agent.*` key from `sys.modules` at SETUP whenever it finds the parent-attribute link between `agent` and `agent.hooks` severed.

`tests/unit/test_persona_substitution.py:17` binds `load_persona_prompt` at collection time, before any eviction. After the fixture evicts, `patch("agent.sdk_client._assemble_segments")` imports and patches a FRESH `agent.sdk_client` module object, while the already-bound `load_persona_prompt` still closes over the OLD module's `__globals__`. The patch applies to a module nothing under test is looking at.

Sibling eviction at `tests/conftest.py:216-219` was ruled out as the cause: `_SDK_IMPORTABLE` is True on this host.

## Still unknown

What severs the `agent` / `agent.hooks` parent-attribute link in the first place. That needs a bisect. The eviction fixture is the mechanism, not the origin.

## Evidence

- `tests/conftest.py:222-271` — autouse guard, deletes all `agent.*` from `sys.modules` at setup.
- `tests/conftest.py:216-219` — sibling eviction, ruled out (`_SDK_IMPORTABLE` True here).
- `tests/unit/test_persona_substitution.py:17` — collection-time binding of `load_persona_prompt`.

## Constraint on any fix

The eviction fixtures guard other corruptions. Removing or narrowing them without full-suite verification trades one isolation defect for another. Any change here requires a FULL-SUITE run, not a targeted one, because `tests/conftest.py` touches everything.

## Supersedes

#2462 Group B.

tomcounsell:
Closes #2558

## Problem

`tests/conftest.py`'s autouse `agent_hooks_consistency_guard` fires when the parent-attribute link between `agent` and `agent.hooks` is severed, and repaired it by deleting every `agent.*` key from `sys.modules`.

That self-heals the attribute walk and replaces one isolation defect with a quieter one. Test modules bind names at collection time (`from agent.sdk_client import load_persona_prompt`, the ordinary idiom, in **46 unit files** here), before any eviction. Afterwards the bound function still closes over the OLD module's `__globals__` while `patch("agent.sdk_client.X")` imports and patches a FRESH module object. The patch lands on a module nothing under test is looking at.

## Reproduction

A scratch test that severs the link, run ahead of `tests/unit/test_persona_substitution.py` on unmodified main:

```
FAILED tests/unit/test_persona_substitution.py::test_load_persona_prompt_customer_service_substitution
FAILED tests/unit/test_persona_substitution.py::test_load_persona_prompt_no_substitutions_backward_compat
FAILED tests/unit/test_persona_substitution.py::test_load_persona_prompt_substitutions_none_is_safe
FAILED tests/unit/test_persona_substitution.py::test_load_persona_prompt_unreferenced_braces_preserved
4 failed, 4 passed
```

Every patch in those `with` blocks was a no-op, so the calls read the real persona tree. Here the divergence surfaces as failures. A test asserting a mock was CALLED would instead have read **green while measuring nothing**, which is the reason this is worth fixing rather than routing around.

## Fix

Rebind rather than evict. `repair_agent_module_tree()` walks the cached `agent.*` keys and rebinds each onto its parent, restoring the whole tree in one pass. That repairs the actual corruption (a severed attribute link) and leaves `sys.modules` identity alone, so every already-bound reference stays valid and a patch reaches the object the test holds.

`mock_claude_sdk_cleanup`'s eviction is deliberately left alone: it evicts after an SDK entry swap precisely because it WANTS `agent.*` re-imported against the restored SDK, which rebinding would not accomplish.

## Red / green

`tests/unit/test_agent_module_tree_repair.py` drives the real fixture body (predicate included, via `__wrapped__`) and pins both halves: module identity preserved, and a collection-time-bound function still sees its patch. Both tests sever the link themselves and restore it in `finally`, so they are order-independent and leave the interpreter as they found it.

Restoring the eviction behavior inside `repair_agent_module_tree`:

```
FAILED tests/unit/test_agent_module_tree_repair.py::test_the_guard_repairs_the_link_without_replacing_the_modules
FAILED tests/unit/test_agent_module_tree_repair.py::test_a_collection_time_binding_still_sees_a_patch_after_a_repair
2 failed
```

With the fix in place: 2 passed.

## Verification scope

The issue asked for a full-suite run because conftest touches everything. A full `tests/unit/` run currently wedges at ~99% (#2574, reproduced during this work), so I ran the population that is affected **by construction** instead: every unit test file that both patches `agent.*` and binds a name from `agent.*` at collection time, 46 files, plus the new test and `test_persona_substitution.py`.

| Batch | Result |
|-------|--------|
| chunk 1 (16 files) | 597 passed |
| chunk 2 (16 files) | 329 passed |
| chunk 3, individually + grouped | 364 passed |

One file in chunk 3, `tests/unit/test_session_executor_runner_dispatch.py`, wedges the controller at 0% CPU with no summary. **It wedges identically on baseline with this change stashed**, so it is pre-existing (#2574) and is now a single-file reproducer for that issue.

`python -m ruff check` and `ruff format` clean.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -286,6 +286,53 @@ def isolate_catchup_kill_switch(_catchup_flag_redirect_dir):
         bridge.catchup.CATCHUP_DISABLED_FLAG = original
 
 
+def repair_agent_module_tree() -> list[str]:
+    """Rebind every cached ``agent.*`` submodule onto its parent package.
+
+    The corruption this repairs: ``sys.modules`` is a flat name->module cache
+    and does not, by itself, keep the attribute tree in sync. CPython binds a
+    submodule as an attribute on its parent package only at the moment that
+    submodule is freshly imported. If something replaces or partially rebuilds
+    ``sys.modules["agent"]`` while ``sys.modules["agent.hooks"]`` survives from
+    an earlier import, the new ``agent`` object never gets ``hooks`` bound onto
+    it, and every dotted-string ``monkeypatch.setattr``/``mock.patch`` that
+    walks through ``agent.hooks`` raises ``AttributeError`` at setup.
+
+    Repairs by REBINDING, not by evicting (#2558). Evicting every ``agent.*``
+    key also self-heals, because the next import rebuilds the whole chain --
+    but it rebuilds it around NEW module objects, and that is a second bug
+    wearing the first one's clothes. A test module that bound a name at
+    collection time (``from agent.sdk_client import load_persona_prompt``, the
+    ordinary idiom, present in ~50 files here) still closes over the OLD
+    module's ``__globals__``. Its ``patch("agent.sdk_client._assemble_segments")``
+    then imports and patches a FRESH module object that nothing under test is
+    looking at. The patch silently applies to nobody: in
+    ``tests/unit/test_persona_substitution.py`` that surfaced as four
+    unexplainable failures, but the same divergence in a test asserting a mock
+    was CALLED would have read green while measuring nothing.
+
+    Rebinding fixes the actual defect -- the severed attribute link -- and
+    leaves module identity alone, so every already-bound reference stays valid
+    and a patch reaches the object the test holds.
+
+    Returns the dotted names rebound, so callers and tests can assert on the
+    repair instead of inferring it.
+    """
+    repaired: list[str] = []
+    for name in sorted(k for k in sys.modules if k == "agent" or k.startswith("agent.")):
+        parent_name, _, attr = name.rpartition(".")
+        if not parent_name:
+            continue
+        parent = sys.modules.get(parent_name)
+        module = sys.modules.get(name)
+        if parent is None or module is None:
+            continue
+        if getattr(parent, attr, None) is not module:
+            setattr(parent, attr, module)
+            repaired.append(name)
+    return repaired
+
+
 @pytest.fixture(autouse=True)
 def agent_hooks_consistency_guard():
     """Detect and repair a corrupt `agent` package/submodule cache state.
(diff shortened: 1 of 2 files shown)


tomcounsell:
Superseded by da044ecf on main (#2551), which ws-harness and I arrived at independently and near-simultaneously. Same diagnosis, same fix shape, same decision to leave `mock_claude_sdk_cleanup`'s eviction alone.

I verified the collision rather than taking it on trust, and da044ecf does fix #2558's actual symptom. Against pure main at c7705909, my original repro (sever the `agent`/`agent.hooks` link in a test ordered ahead of `tests/unit/test_persona_substitution.py`) now passes: **11 passed**. Reverting only da044ecf's rebind hunk back to eviction on top of that same main brings the failures straight back:

```
FAILED tests/unit/test_persona_substitution.py::test_load_persona_prompt_customer_service_substitution
FAILED tests/unit/test_persona_substitution.py::test_load_persona_prompt_no_substitutions_backward_compat
FAILED tests/unit/test_persona_substitution.py::test_load_persona_prompt_substitutions_none_is_safe
FAILED tests/unit/test_persona_substitution.py::test_load_persona_prompt_unreferenced_braces_preserved
FAILED tests/unit/test_persona_substitution.py::test_load_persona_prompt_substitutes_over_real_segments
5 failed, 6 passed
```

So the fix on main is load-bearing and this PR's conftest hunk is redundant.

Dropping my `repair_agent_module_tree()` extraction too. Its only caller was my own test's `finally` block, which is self-justifying rather than a real need, and refactoring a shared file on that basis is churn.

Dropping `tests/unit/test_agent_module_tree_repair.py` as well. `TestAgentHooksGuardRepair` on main already pins both halves, and does the identity half better than mine did (a substituted `types.ModuleType` rather than my `delattr`). The one thing mine exercised that main's does not is a collection-time-bound *callable* being called through a `mock.patch` on its dotted path, but identity preservation already implies that, so it is a restatement rather than coverage.

PR #2587, opened 2026-08-06
Make the persona byte-stability baseline run on every host (#2555)

### LINKED ISSUE #2555: Persona byte-stability baseline is inert on most hosts: hostname-derived fixture gate SKIPS silently, and the baseline has no freshness enforcement
## Root cause

The persona byte-stability baseline has two independent holes.

**The hostname gate skips silently.** `tests/unit/test_compose_system_prompt.py:30-36` derives a fixture slug from `socket.gethostname()`. `tests/fixtures/` contains only `Mac-local`. On a host named `MacBookPro.local` the test does not fail, it SKIPS, with no signal that the identity guard is not running. Renaming a machine on the network, or joining a different one, flips the persona guard between skip and fail with nothing to distinguish the two states from a passing run.

**No freshness enforcement.** `tests/unit/test_compose_system_prompt.py:65-70` compares against a checked-in baseline with no mechanism to detect that the baseline is stale relative to the persona segments it is supposed to be guarding.

Together: the test that is supposed to catch unintended persona drift can be inert on a given machine and nobody would know.

## Evidence

- `tests/unit/test_compose_system_prompt.py:30-36` — slug from `socket.gethostname()`.
- `tests/unit/test_compose_system_prompt.py:65-70` — baseline comparison, no freshness check.
- `tests/fixtures/` — only `Mac-local` present.

## Fix shape

Make the missing-fixture case fail rather than skip, or generate the baseline host-independently. Add freshness enforcement tying the baseline to the segments it guards.

## Verification

Negative control: touch a persona segment and confirm the guard fires. A skip is not a pass.

## Supersedes

- #2429 Group B
- #2430 item 1a

tomcounsell:
Closes #2555

## Problem

The persona byte-stability guard had two independent holes, and between them it measured nothing.

**The hostname gate skipped silently.** `tests/unit/test_compose_system_prompt.py` derived its fixture directory from `socket.gethostname()`, so the test ran only on a machine whose name matched a checked-in directory. Only `Mac-local` was ever committed, and that matches no host in the current fleet. Confirmed on this host:

```
SKIPPED [1] No baseline for hostname 'Toms-MacBook-Air-local'; run
scripts/capture_persona_baseline.py to record one.
14 passed, 1 skipped
```

A passing run looked identical to a running one, and renaming a machine moved the guard between the two states silently.

**No freshness enforcement.** Nothing tied the baseline to the segments it guarded, so even where it did run it could be arbitrarily stale.

## Fix

Both holes close with the same move: compose from repo-tracked inputs only, and record one checked-in baseline that runs everywhere.

Three inputs vary per machine and live outside the repo. The baseline pins them out:

| Input | Pinned to |
|-------|-----------|
| `~/Desktop/Valor/identity.json` | a path that does not exist, so `config/identity.json` stands alone |
| `~/Desktop/Valor/personas/` | a path that does not exist, so `config/personas/` overlays are used |
| `config/PRINCIPAL.md` | `tests/fixtures/persona/principal.md` |

None of the three is in the repo, so no pull request can drift them and no test could ever have gated them. Pinning them costs no coverage. What remains is exactly the surface a diff can change: `WORKER_RULES`, the segments and `manifest.json`, the in-repo overlays, the composition order, and `CLAUDE.md`'s completion criteria.

**Freshness then needs no separate mechanism.** The baseline is a pure function of repo-tracked inputs, so a stale baseline and a drifted persona are the same red. A change to a segment must carry a re-recorded baseline in the same commit.

The test imports `compose_repo_only_eng_worker_prompt` from the capture script rather than restating the recipe. Two copies would be free to drift, and the test would then compare two different compositions and report the difference as persona drift.

## Red / green

Three negative controls, each verified:

1. **Persona drift** — appended a line to `config/personas/segments/work-patterns.md`:
   `AssertionError: ENG cell drifted from the baseline: composed 62073 chars, baseline 62015 chars` → 1 failed.
2. **Missing baseline fails, never skips** — moved the baseline aside:
   `AssertionError: ...eng_worker_repo_baseline.txt is missing. Run scripts/capture_persona_baseline.py` → 2 failed, **0 skipped**.
3. **Broken pin** — made all three `patch.object` calls no-ops, so the baseline would silently go back to measuring the local machine:
   `AssertionError: the pinned composition equals the unpinned one on a host that HAS a private persona layer` → 2 failed.

Green with everything restored: **17 passed, 0 skipped** in that file; 113 passed across the persona/compose/principal neighbourhood.

The negative control in (3) ships as a permanent test, `test_the_baseline_pins_out_the_private_layer_it_claims_to`. Without it a broken pin would go unnoticed on any machine whose private layer happens to be absent.

`python -m ruff check` and `ruff format` clean.

## Deletions

`tests/fixtures/Mac-local/` is removed. Its `pm_system_prompt_baseline.txt` was a snapshot of a `config/personas/project-manager.md` that no longer exists, and no test read it or `eng_system_prompt_baseline.txt`.

Note for whoever owns `docs/plans/pipeline-graph-single-source-of-truth.md`: lines 194-195 list regenerating and deleting those `Mac-local` fixtures as a prerequisite. This PR satisfies it. I left the plan doc alone rather than edit another lane's active plan.


tomcounsell:
Rebased onto `deb10d0b` (picks up ws-harness's three `tests/conftest.py` changes, including the #2574 calendar-heartbeat fixture). Now MERGEABLE / CLEAN.

Re-verified on the new base:

- `scripts/capture_persona_baseline.py --check` → `Baseline is current (62015 chars)`, so the recorded bytes still match what this checkout composes.
- 109 passed across `test_compose_system_prompt.py`, `test_conftest_isolation_guards.py`, `test_migrate_strip_pid_fields.py` and `test_strip_migration_shared.py`, with **0 skipped**.

Still needs the DOCS marker and a REVIEW verdict. I authored it, so the verdict needs someone else.

tomcounsell:
## Review: APPROVED

Verified against the merge of `ws-2555` onto current `origin/main` (3264ce7f). Picking up the author's note: the verdict needed someone other than the author, and this is it.

**I independently confirmed the guard was inert, rather than taking the report's word for it.** On this host `socket.gethostname()` is `Toms-MacBook-Air.local`, slugging to `Toms-MacBook-Air-local`, and `git ls-tree origin/main tests/fixtures/` shows the only committed persona fixture directory was `Mac-local`. So the byte-stability assertion could never execute here, and a passing suite was indistinguishable from a running one. The defect is exactly as described.

**The fix's central argument is sound: pinning the machine-local inputs costs no coverage.** `~/Desktop/Valor/identity.json`, `~/Desktop/Valor/personas/`, and `config/PRINCIPAL.md` are all absent from the repo, so no pull request could ever drift them and no test could ever have gated them. Removing them from the composition subtracts nothing that was being protected while making the remaining surface identical on every host. What is left is precisely what a diff can change: `WORKER_RULES`, the segments and `manifest.json`, the in-repo overlays, the composition order, and `CLAUDE.md`'s completion criteria.

**Freshness falling out for free is the elegant part.** Because the baseline is a pure function of repo-tracked inputs, a stale baseline and a drifted persona produce the same red. There is no second mechanism to build, and no way for one to hide behind the other. That is a better answer than the separate freshness check the issue implied.

**Importing `compose_repo_only_eng_worker_prompt` from the capture script rather than restating the recipe is the right call.** Two copies of the composition would be free to drift, and the failure mode would be especially nasty: the test would compare two different compositions and report the difference as persona drift, sending someone hunting a persona change that never happened.

**The negative control is what makes this durable, and it actually ran.** `test_the_baseline_pins_out_the_private_layer_it_claims_to` asserts the pinned composition differs from the unpinned one on a host that has a private layer. Without it, a pin that silently stopped taking effect would look green on any machine lacking a private layer, and the baseline would quietly revert to being host-dependent. On this host the control did **not** skip, meaning a private layer is present and the pin is demonstrably doing its job. That is the strongest available result.

**The `Mac-local` deletion is correct.** `pm_system_prompt_baseline.txt` snapshotted a `config/personas/project-manager.md` that no longer exists, and no test read it or `eng_system_prompt_baseline.txt`. Deleting rather than regenerating matches the repo's no-legacy stance. Leaving `docs/plans/pipeline-graph-single-source-of-truth.md` untouched was also right; editing another lane's active plan to claim a prerequisite is not this PR's call to make, and the note in the description is the correct way to hand that off.

### Verification (merged onto `origin/main` 3264ce7f)

```
python scripts/capture_persona_baseline.py --check
  Baseline is current (62015 chars)

tests/unit/test_compose_system_prompt.py    17 passed, 0 skipped
```

The `--check` pass on the merged tree matters independently of the test run: it proves none of tonight's four already-merged PRs perturbed the composed prompt, so the recorded bytes are still accurate against current main rather than against the base this branch was cut from.

### Non-blocking observation

`test_the_baseline_pins_out_the_private_layer_it_claims_to` still skips on a host with no private layer. That is correct scoping rather than a repeat of the defect being fixed: the control genuinely requires a private layer to have anything to detect, and the primary byte-stability guard next to it runs unconditionally. The skip reason says so explicitly, so it cannot be mistaken for the old silent-skip behavior.

No blockers.

PR #2579, opened 2026-08-06
Make ~/.claude/hooks a real user tree and guard the legacy directory-symlink alias (#2567)

tomcounsell:
Closes #2567

## Problem

Under the legacy layout `~/.claude/hooks` is a directory symlink into a checkout's `.claude/hooks/`. There is then no separate user tree: every path beneath it is tracked source, so a pass that enumerates the directory and removes what it reads as an orphaned user file removes the checkout. That deleted 36 tracked files during the #2521 build.

Two mechanics make the layout easy to misread, and both showed up in that incident:

- The same inode appears under both paths, which looks like a hardlink pair and invites the conclusion that deleting the user copy leaves the repo source intact. It is one file under two names.
- `_ensure_hardlink()` sees matching inodes and reports `exists`, so a sync through the alias registers as success while having written nothing.

## Change

**One shared answer to "is this a real user tree?"** `user_hooks_root_is_repo_aliased(project_dir, user_claude=None)` walks the ancestry from `hooks` down to `.claude` for a symlink, then asks whether the resolved path is some checkout's `.claude/hooks` (`.git` probed with `exists()`, so worktrees count). Every writer, deleter, and the migration itself consult it, so detection and destruction share one definition. Two distinctions the callers depend on: the answer is **not** scoped to `project_dir` (an agent in a worktree while `~/.claude` points at the main checkout is the same hazard), and a symlink to a plain user directory such as `~/dotfiles/claude-hooks` is **not** an alias (a real user tree, just not where it appears, and migrating it away would destroy a deliberate setup).

**The migration lives next to its rebuild.** The distance between the unlink and the rebuild is the hazard, not the unlink. Anything that fails in between leaves `~/.claude/hooks` absent while `~/.claude/settings.json` still registers blocking global hooks against it, and a python interpreter on a missing script exits 2, the PreToolUse deny code. That denies every Bash call in every repo, including the `/update` that would repair it. So the migration sits inside `sync_user_hooks()`, after both fallible steps (manifest load, interpreter probe) have returned and immediately before the deployment loop.

**Nothing under `~/.claude/hooks/` stays registered against a missing script.** `_register_deployed_only()` filters undeployed declarations before writing, then sweeps `~/.claude/settings.json` for any entry, marked or not, naming a missing file under the hooks root. Both call sites go through it. The sweep is narrower than a prefix rule (the test is the referenced file's absence) and expands `~/` and `$HOME/` first, since `shlex` leaves them literal while the `/bin/sh` that runs the hook does not.

[...]

PR #2594, opened 2026-08-06
Pin the two unpinned #2567 guards and fold the _tilde copies (#2593)

### LINKED ISSUE #2593: Two #2567 guards are unpinned, and the merged commit message overclaims that every guard was mutation-checked
Follow-up to #2567 / PR #2579 (merged as `39ae8fc3`).

## The claim that is wrong

The squash commit message on `main` says:

> Every guard was mutation-checked in isolation, one at a time

That is false for two of them, and two rows of its mutation table understate. The commit message is immutable, so the fix is to make the claim true rather than to correct the text.

This matters more than a normal docs nit. The whole 2026-08-06 batch was about tests that report success while measuring nothing (#2552, #2553, #2556, #2557, #2523). Shipping an unbacked "everything was mutation-checked" claim is that same defect one level up, in the record a future reader trusts.

## Measured, one mutation at a time, against `main` at `88825e31`

| Guard | Mutation | Tests killed |
|---|---|---|
| `hardlinks.py` OSError fail-closed (`return hooks_root`) | → `return None` | **0** |
| `_deregister_missing_hook_scripts` `isinstance(command, str)` | → `entry.get("command") or ""` | **0** |
| `sync_user_hooks` `hooks_root.mkdir` after unlink | removed | 2 |
| Deregistration sweep | call removed | 4 (commit says 1) |
| `_cleanup_renamed` hooks guard | `continue` removed | 2 (commit says 1) |
| Migration unlink | moved back to `sync_claude_dirs` | 1 (commit says 1, correct) |
| Migration unlink | removed entirely | 3 |

The two zero rows are real gaps. The `mkdir` guard was reported as unpinned in review triage; it is not, and this issue exists partly to correct that too.

The sweep and `_cleanup_renamed` rows drifted because they were measured in review round 3 and not re-measured after round 4 added tests. Re-measuring after every subsequent round is the process lesson.

## Scope

1. Add tests that kill the two zero-row mutations:
   - `user_hooks_root_is_repo_aliased` must fail **closed** on `OSError`, returning a truthy value so callers decline to write or delete. A `None` there reads as "real user tree, safe to delete", which is the one answer an unresolvable path cannot support.
   - The sweep must skip a non-string `command` rather than coerce it. Coercing hides a malformed entry instead of leaving it alone, and this pass only ever removes.
2. Fold the 11 inline `str(...).replace(str(Path.home()), "~")` copies into the existing `_tilde()` helper (Development Principle 1), and use it at `hardlinks.py:1087` where a bare `str(hooks_root)` still produces a second path form in one action list.
3. `run.py`'s removed-count block reports three unrelated events (the hooks migration, a deregistered dead hook, a renamed-hardlink cleanup) under one "stale hardlink(s)" heading.
4. Docs: state why the settings.json sweep carries no `.bak`, and note the parent-alias / `expandvars` interaction.

## Not in scope

Nothing about the #2567 behavior itself. The guards are correct and the fix is landed and verified; this issue is about the evidence for them.

tomcounsell:
Closes #2593

Follow-up to #2567 / PR #2579, merged as `39ae8fc3`. Rebased onto `addff9eb` (#2455).

## The problem

The squash commit on `main` says "Every guard was mutation-checked in isolation, one at a time." That was false for two guards, and two rows of its table understated. The commit message is immutable, so this makes the claim true rather than correcting the text.

That matters more than a normal docs nit here. The whole 2026-08-06 batch was about tests that report success while measuring nothing (#2552, #2553, #2556, #2557, #2523). An unbacked "everything was mutation-checked" line is that same defect one level up, in the record a future reader trusts.

## Two guards now pinned

Both under the natural "remove the guard" mutation, not a contrived one.

- **`user_hooks_root_is_repo_aliased` must fail closed on `OSError`.** `None` reads to every caller as "real user tree, safe to write and delete", and `_cleanup_renamed` acts on that answer by unlinking. Failing open is a deletion decision taken on a reading the code just admitted it could not make.
- **The sweep must skip a non-string `command` rather than coerce it.** Dropping the check raises `AttributeError` out of `shlex.split`.

Honest note on the second: a `str(... or "")` coercion mutation *survives* both tests, because Python's `repr` escaping is robust enough that a coerced list or dict never produces a token matching the hooks root. The guard is pinned against removal, not against that particular rewrite.

## A live crash the second test found

A non-string, non-falsy `command` (`42`, a list) got past the `or ""` guard added in #2579 and raised `TypeError` out of `_extract_manifest_id` into an unguarded `/update`. `or ""` only ever handled falsy values, so `null` was covered and `42` was not.

`_extract_manifest_id` now accepts any type and returns `None` for anything that is not a string. `~/.claude/settings.json` is hand-editable, so `null`, numbers, and lists all really arrive, and anything that is not a string carries no marker by definition. The now-misleading `or ""` at the call site is gone.

## Interaction with #2455's `exit_policy` (verified, not assumed)

#2455 replaced `blocking` with `exit_policy` and `_EXIT_POLICY_SUFFIX`. I re-ran the wedge argument against the real suffixes rather than reasoning about them:

```
propagate  -> exit 2      (no suffix)
deny-only  -> exit 2      (re-raises exactly 2)
suppress   -> exit 0      (|| true)
```

**`exit_policy` raises the stakes rather than lowering them.** Under the old `blocking = false` default every hook carried a blanket `|| true`, so a missing script was inert. `deny-only` deliberately stops swallowing exit 2, which is the point of that change, and a missing script is then indistinguishable from a deliberate deny. Two of three policies now wedge on a missing script where one of two did before.

The deployment filter, not the policy, is what keeps `deny-only` safe — and #2455's own `test_deny_only_declaration_that_fails_to_deploy_is_never_registered` independently pins that filter, which is a nice confirmation the two changes were designed to meet. Docs and test docstrings that described the old `blocking` semantics are corrected here.

## Review nits from #2579

- Ten inline copies of `_tilde`'s body folded into the helper (Development Principle 1), plus the one bare `str(hooks_root)` that put a second path form in a single action list.
- `run.py`'s removed-count block called three unrelated events "stale hardlink(s)": a `RENAMED_REMOVALS` sweep, the dir-symlink migration, and a deregistered hook. The count is generic now and each line carries its own detail.
- Docs: why the settings.json sweep carries no `.bak`, and the parent-alias interaction with `$HOME` expansion.

On the `.bak`: the sweep only ever deletes entries it has proven point at a missing file, and `if not removed: return` means a healthy machine never opens the file for writing. A `.bak` would add a write to the common path to protect against a removal that is already evidence-gated, on a file that gates every Bash call. Recovery if it ever removes wrongly is the next `/update`, which re-registers every declared hook from the manifest.

## Measurements, re-run against this head

| Guard | Mutation | Killed | #2579 said |
|---|---|---|---|
| OSError fail-closed | → `return None` | 1 | not measured |
| Sweep `isinstance` | guard removed | 1 | not measured |
| `_extract_manifest_id` type guard | removed | 2 | did not exist |
| Registration filter | removed | 2 | 0 (1 pre-rebase) |
| Deregistration sweep | call removed | 4 | 1 |
| `_cleanup_renamed` hooks guard | `continue` removed | 2 | 1 |
| Migration unlink | moved back | 1 | 1 (correct) |
| Migration unlink | removed entirely | 3 | not measured |
| `mkdir` after unlink | removed | 2 | not measured |

The `mkdir` guard was reported in triage as unpinned. It is not; it kills 2.

The sweep and `_cleanup_renamed` rows drifted because they were measured in review round 3 and never re-measured after round 4 added tests. Re-measuring after every round is the process lesson, and it is the one I would most want carried forward.

## Verification

`PYTHONPATH=<worktree> ./scripts/pytest-clean.sh tests/unit/test_update_hardlinks.py tests/unit/test_hook_migration.py tests/unit/test_hook_manifest.py tests/unit/test_hook_interpreter.py tests/unit/test_hooks_audit.py -q` — **154 passed** (141 before the rebase; #2455 contributes the rest). `ruff check` and `ruff format` clean.

The rebase conflict with #2455 was one hunk in `tests/unit/test_update_hardlinks.py`, append-vs-append at end of file. Both sides' tests kept; `hardlinks.py` and `hook-manifest.md` auto-merged. Every guard mutation re-run after the rebase, in case the merge masked one — none did.

No behavior change to the #2567 fix itself beyond the `_extract_manifest_id` crash guard. `~/.claude/hooks` on this machine is unaffected: still a real directory, helper still returns `None` from both the main checkout and a worktree.

## DOCS

`docs/features/hook-manifest.md` gains the non-string-command rationale, the no-`.bak` reasoning, the parent-alias / `$HOME`-expansion interaction, and corrected `exit_policy` semantics replacing the stale `blocking` prose.


tomcounsell:
## Review: APPROVED

Reviewed at head of `session/ws-hooks-2593`. Small, self-contained, and it does what #2593 asks.

**Both previously-unpinned guards are now pinned by tests that name the behavior they protect:**

- `test_an_unresolvable_hooks_root_fails_closed` covers the `OSError` fail-closed path.
- `test_the_sweep_skips_a_non_string_command_rather_than_coercing_it` covers the `isinstance` guard, and asserts the malformed entries survive untouched (`None in commands and 42 in commands`) rather than merely asserting no crash.

**The `_tilde` fold is the Development Principle 1 cleanup it claims.** Ten inline copies of the helper's own body are replaced by calls to the helper.

**Verified rather than credited:** `tests/unit/test_update_hardlinks.py` + `tests/unit/hooks/` = **140 passed**. `ruff check` and `ruff format --check` clean. Merges clean against `main` at `addff9ebb`, which already carries PR #2455's `exit_policy` rework of the same file.

This closes out the nit set left open when #2579 merged, including the commit-message overclaim that every guard had been mutation-checked.

tomcounsell:
## Review: PR Closed

**This PR merged while the review was in flight.** The review below ran to completion against `5177be48`, the exact SHA that landed, so every finding here now describes code on `main`. Posting it rather than discarding it: the verdict cannot gate anything, but the findings still hold and want a follow-up issue.

Had it still been open the verdict would have been **Changes Requested — Tech Debt**: no blockers, but four accuracy findings, one of which is the same shape as the defect this PR exists to remove.

Two judges (`code-quality`, `risk`) both returned APPROVED with zero blockers and no disagreement (`any-blocker-wins`, k=2, n=2, mean confidence 0.85, tied=false); their findings are folded in below.

### The `exit_policy` claim you asked about — verified, and correct

I did not take the PR body's word for it. `_EXIT_POLICY_SUFFIX` (`scripts/update/hardlinks.py:903-907`) is exactly:

```python
_EXIT_POLICY_SUFFIX = {
    "propagate": "",
    "deny-only": '; __hook_rc=$?; [ "$__hook_rc" = 2 ] && exit 2 || exit 0',
    "suppress": " || true",
}
```

I then ran a missing script through each suffix in a real `sh`:

```
propagate  -> exit 2
deny-only  -> exit 2
suppress   -> exit 0
```

Two of three wedge. The corrected prose in `hook-manifest.md` and in the `test_hooks_symlink_survives_a_manifest_that_fails_to_load` docstring matches what the code does, and the "raises the stakes rather than lowering them" framing is right: under `blocking = false` a missing script was inert, and `deny-only` deliberately stops swallowing 2. It is also live rather than hypothetical — `manifest.toml`'s only global `PreToolUse`/`Bash` declaration, `validate_commit_message_fork`, carries `exit_policy = "propagate"`.

### Measurements reproduced in my own environment

Per the "a number the PR claims is a claim, not evidence" rule, I re-ran the load-bearing numbers in a detached worktree at `5177be48`:

| Claim | PR says | I measured |
|---|---|---|
| Named-files suite | 154 passed | **154 passed** |
| OSError fail-closed → `return None` | 1 | **1** (`test_an_unresolvable_hooks_root_fails_closed`) |
| Sweep `isinstance` guard removed | 1 | **1** (`test_the_sweep_skips_a_non_string_command…`) |
| `_extract_manifest_id` type guard removed | 2 | **2** (also `test_an_unparseable_hook_command_is_reported_not_raised`) |
| `str(... or "")` coercion mutant | survives (disclosed) | **survives** (77 passed) |
| `ruff check` / `ruff format --check` | clean | **clean** |

Every number I checked reproduced exactly. I did not re-run the other six mutation rows; those remain the author's measurement.

### The finding that matters

The PR body says "Docs and test docstrings that described the old `blocking` semantics are corrected here." Both judges independently found that sentence is not yet true, and it is not true *in the file this PR edits*. `hardlinks.py:1208` still reads "`_build_hook_command` appends `|| true` only for non-blocking declarations" — a statement about a field `HookDeclaration` no longer has, in the docstring of the very function whose wedge-safety argument rests on it. Its twin at `:1089` is the source copy of the doc paragraph this PR *did* correct.

That is an overclaim in the record a future reader trusts, in a PR whose stated purpose is removing an overclaim from the record a future reader trusts. Two lines make it true.

## Rubric

- [x] **1. Plan vs. implementation match** — pass — *no plan doc (chore-shaped follow-up); all four scope items from #2593 delivered and verified: two zero-row mutations now killed, `_tilde` fold complete, `run.py` count reworded, all four DOCS additions present.*
- [x] **2. New code quality** — pass — *`_extract_manifest_id(command: object)` is the honest signature for hand-editable JSON; the `or ""` removal at the call site is correct and its now-wrong comment went with it.*
- [x] **3. Test coverage** — pass — *both new guards measured at 1 kill each, reproduced independently.*
- [x] **4. Regression risk to existing callers** — pass — *`_tilde`'s body is byte-identical to the ten copies it replaces; the one changed action string (`str(hooks_root)` → `~/.claude/hooks`) is asserted by no test; nothing asserts the old `"stale hardlink(s)"` wording; six `_extract_manifest_id` call sites all safe under the relaxation.*
- [x] **5. Data integrity** — n/a — *no schema, model, or migration-registry change.*
- [x] **6. Security** — pass — *no new subprocess/eval/secret surface.*
- [ ] **7. Documentation accuracy** — fail — *`hook-manifest.md:141`'s "the other writers of this file snapshot it first" overstates; `_merge_hook_settings`, called on the immediately preceding line, has no `.bak` either.*
- [ ] **8. PR body accuracy** — fail — *"Docs and test docstrings that described the old `blocking` semantics are corrected here" is not yet true of `hardlinks.py:1089`, `:1208`, or `migrations.py:733`. Every other body claim I checked reproduced exactly.*
- [x] **9. Disclosed deferrals** — pass — *the coercion-mutant-survives note is honest and I confirmed it (mutant survives, 77 passed); explicit out-of-scope exclusion with sound rationale, no tracking required.*
- [x] **10. Follow-up claims verified** — n/a — *the body claims no filed follow-ups.*

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — PASS — *validated against issue #2593's four scope items; no plan file exists.*
- **2. No-Gos from plan — none violated** — PASS — *#2593's "Not in scope: nothing about the #2567 behavior itself" is respected; the `_extract_manifest_id` change is a crash guard, not a behavior change to the fix.*
- **4. New integration tests — exercise serialization boundary** — PASS — *the sweep test round-trips real JSON through `~/.claude/settings.json` on disk rather than asserting in memory.*
- **5. Plan internal consistency — spike findings match task steps** — N/A — *no plan doc.*
- **7. New public APIs — docstrings present** — PASS — *`_extract_manifest_id`'s widened docstring states why the type is `object`.*
- **10. Tests cover the failure path (not just happy path)** — PASS — *both new tests are failure-path tests, and both are mutation-verified.*
- **11. UI changes (if any) — screenshot captured** — N/A — *no HTML/CSS/JS/JSX/Vue/template files in the diff; visual-proof gate is a no-op.*
- **12. Docs updated for user-facing changes** — PASS — *`hook-manifest.md` carries all four additions; accuracy nit noted under rubric 7.*

### Blockers

- None

### Tech Debt

- [ ] **`scripts/update/hardlinks.py:1208`** — `` `_build_hook_command` appends `|| true` only for non-blocking declarations, and a python interpreter invoked on a missing file exits 2 `` — `HookDeclaration` has no `blocking` field after #2455/#2527; line 932 is `command += _EXIT_POLICY_SUFFIX[decl.exit_policy]`, and `|| true` comes only from `suppress`. This is the exact sentence the PR rewrote in `hook-manifest.md` and in the test docstring at `test_update_hardlinks.py:1247`, so the body's claim that the stale prose is corrected is not yet true. Twin at `:1089` ("still registers blocking global hooks against it") and one more at `migrations.py:733` ("the `|| true` guard the manifest's `blocking` flag calls for"). **Fix:** rewrite all three to the `exit_policy` wording already written for the doc.
- [ ] **`scripts/update/migrations.py:661`** — `if f"{hooks_root}/" not in command:` — the same non-string-`command` crash this PR just fixed in `hardlinks.py` is still live in the two migrations that walk the same hand-editable file on the same `/update` run. `_should_remove` receives `e.get("command", "")`, whose `""` default never fires for a present `"command": null`, so `"…" not in None` raises `TypeError`; `:829`'s `stripped.endswith("|| true")` raises `AttributeError` on `None` and on `42`. `run_pending_migrations` calls `fn(project_dir)` with no `try`/`except`. Pre-existing and reachable only where these one-shot migrations are not yet recorded complete — i.e. the `/setup` path — but the PR now *documents* the premise ("`null`, a number, or a list all really arrive"), which makes leaving the adjacent sites unguarded a gap rather than a non-decision. **Fix:** add the same `isinstance(command, str)` skip at both sites, or say in the PR body that the hardening covers `hardlinks.py` and file a follow-up for `migrations.py`.

### Nits

- [ ] **`scripts/update/run.py:769`** — `f"Removed {result.hardlink_result.removed} item(s) from ~/.claude/"` — the new comment claims "the count is generic and each line carries its own detail", but count and lines are not one-to-one: `_merge_hook_settings` does `result.removed += removed` (`hardlinks.py:1449`) for its marker-keyed removal pass (`:1384-1386`) while emitting only a `"created"` action. Delete a global declaration from `manifest.toml`, or let one fail to deploy, and `/update` prints `Removed 1 item(s) from ~/.claude/` followed by zero detail lines. That deregistration is the fleet event the docs call the safe direction. **Fix:** append a `LinkAction("", rel_path, "removed", f"Deregistered undeclared hook: {mid}")` in that pass so every counted unit has a line.
- [ ] **`docs/features/hook-manifest.md:141`** — "The other writers of this file snapshot it first; this one does not" — true of `migrations.py:698`/`:892` and `cal_integration.py:97`, but not of `_merge_hook_settings`, the sibling writer of the same file invoked on the immediately preceding line (`hardlinks.py:1253-1254`), which writes at `:1435` with a plain `write_text` and no snapshot either. **Fix:** narrow to "the migration writers", or state the real distinction — the sweep is remove-only and evidence-gated, the migrations rewrite command strings.
- [ ] **`tests/unit/test_update_hardlinks.py:1900`** — `def test_the_sweep_skips_a_non_string_command_rather_than_coercing_it` — I confirmed the disclosed gap: under `command = str(entry.get("command") or "")` the suite is 77 passed, green. The PR body is honest about this; the test's name and docstring are not, and the test outlives the PR body. Given #2593 is specifically about a name claiming more than it measures, that belongs in the docstring. **Fix:** rename toward what it pins (removing either guard raises), and move the honest note into the docstring.
- [ ] **`tests/unit/test_update_hardlinks.py:1873`** — `assert hardlinks.user_hooks_root_is_repo_aliased(...) is not None` — the docstring says "must return a truthy value", and callers (`_cleanup_renamed`, `sync_user_hooks`) use the returned `Path`. **Fix:** `== hooks_root`. Kills the same mutant, and additionally rules out a non-None sentinel callers would misuse.

### Miscellaneous

- None

- **"a `str(... or "")` coercion mutation *survives* both tests… The guard is pinned against removal, not against that particular rewrite."** — explicit out-of-scope exclusion with sound rationale, independently reproduced (mutant survives, 77 passed). No tracking required. The nit above is about where the disclosure lives, not whether it is true.
- **"No behavior change to the #2567 fix itself beyond the `_extract_manifest_id` crash guard."** — verified against the diff; matches #2593's stated No-Go.

### Verification Results

The plan has no machine-readable `## Verification` table (no plan doc). Ran the PR body's stated command in a detached worktree at `5177be48`:

```
tests/unit/test_update_hardlinks.py tests/unit/test_hook_migration.py
tests/unit/test_hook_manifest.py tests/unit/test_hook_interpreter.py
tests/unit/test_hooks_audit.py    ->  154 passed

ruff check      -> All checks passed!
ruff format --check -> 3 files already formatted
```

### Screenshots

### Consensus

`any-blocker-wins` over 2 judges: verdict APPROVED on blockers (0), tech_debt max 3, mean confidence 0.85, tied=false. The two judges converged independently on the stale-`blocking` finding and on the `migrations.py` crash gap, which is the strongest signal in this review. Under the zero-findings-only approval rule that would have been Changes Requested; the PR merged first, so the recorded verdict is `PR_CLOSED` and the two tech-debt items belong in a follow-up issue against `main`.

PR #2601, opened 2026-08-06
Finish the stale exit_policy prose and guard migrations.py the same way (#2600)

### LINKED ISSUE #2600: Post-merge findings on #2594: the stale "blocking" prose I claimed to have fixed, and the same non-string-command crash live in migrations.py
The #2594 review completed against `5177be48`, the exact SHA that landed, so every finding describes code on `main`. Zero blockers; two tech-debt items and four nits, all reproduced independently before filing.

Two of these are worth naming plainly because of what #2594 was for.

## 1. The stale prose I claimed to have corrected is still in the source

#2594's body says the stale `blocking` prose is corrected. It corrected `docs/features/hook-manifest.md` and the test docstring at `test_update_hardlinks.py:1247`, and missed three source sites:

- `hardlinks.py:1208` — "`_build_hook_command` appends `|| true` only for non-blocking declarations"
- `hardlinks.py:1089` — "still registers blocking global hooks against it"
- `migrations.py:733` — "applies the `|| true` guard the manifest's `blocking` flag calls for"

`HookDeclaration` has had no `blocking` field since #2455/#2527. Line 932 is `command += _EXIT_POLICY_SUFFIX[decl.exit_policy]`, and `|| true` comes only from `suppress`.

This is the same defect #2594 exists to remove, in the PR that removes it. A claim in a merged body that a class of staleness was fixed, when a third of the instances were not touched.

## 2. The non-string-`command` crash is still live in `migrations.py`

#2594 fixed this in `hardlinks.py` and documented the premise — "`null`, a number, or a list all really arrive" in a hand-editable `~/.claude/settings.json`. Two migrations walk the same file on the same `/update` run and are unguarded:

```
migrations.py:661  f"{hooks_root}/" not in command   -> TypeError: argument of type 'NoneType' is not iterable
migrations.py:829  stripped.endswith("|| true")      -> AttributeError on None and on 42
```

Both reproduced. `_should_remove` receives `e.get("command", "")`, whose `""` default never fires for a present `"command": null`. `run_pending_migrations` calls `fn(project_dir)` with no `try`/`except`.

Reachable only where these one-shot migrations are not yet recorded complete, i.e. the `/setup` path on a new machine. Pre-existing, but documenting the premise while leaving the adjacent sites unguarded turns a non-decision into a gap.

## 3. `/update` can print a removal count with no lines

`_merge_hook_settings`'s marker-keyed removal pass (`hardlinks.py:1384-1386`) does `result.removed += removed` but emits only a `"created"` action. `run.py:769` prints `Removed N item(s) from ~/.claude/` and then iterates `action == "removed"`, finding none.

Delete a global declaration from `manifest.toml`, or let one fail to deploy, and the operator sees a count with zero detail. That deregistration is the fleet event the docs call the safe direction. #2594's own comment claims "the count is generic and each line carries its own detail", which is not true of this path.

## 4. Three smaller accuracy items

- `hook-manifest.md:141` "The other writers of this file snapshot it first" — false for `_merge_hook_settings`, the sibling writer invoked on the immediately preceding line, which also uses a plain `write_text`. The real distinction is that the sweep is remove-only and evidence-gated while the migrations rewrite command strings.
- `test_the_sweep_skips_a_non_string_command_rather_than_coercing_it` — the `str(... or "")` coercion mutant survives. #2594's body disclosed this honestly; the test's name and docstring do not, and the test outlives the body. Given #2593 was specifically about a name claiming more than it measures, the disclosure belongs in the docstring.
- `test_an_unresolvable_hooks_root_fails_closed` asserts `is not None` where callers use the returned `Path`. `== hooks_root` kills the same mutant and additionally rules out a sentinel callers would misuse.

## Note on process

The #2594 review's two judges converged independently on items 1 and 2, which the reviewer called the strongest signal in that review. Under the zero-findings approval rule the verdict would have been Changes Requested; the PR merged while the review was in flight, so it recorded `PR_CLOSED` instead. Nothing was bypassed — the merge simply beat the review.

tomcounsell:
Closes #2600

Post-merge findings from the #2594 review, which ran to completion against `5177be48` — the exact SHA that landed — and so describes code on `main`. Zero blockers; two tech-debt items and four nits, all reproduced before fixing.

## 1. The stale prose #2594 claimed to have fixed

#2594's body said the stale `blocking` prose was corrected. It corrected `docs/features/hook-manifest.md` and one test docstring, and left three source sites untouched:

- `hardlinks.py:1208` — "`_build_hook_command` appends `|| true` only for non-blocking declarations"
- `hardlinks.py:1089` — "still registers blocking global hooks against it"
- `migrations.py:733` — "applies the `|| true` guard the manifest's `blocking` flag calls for"

`HookDeclaration` has had no `blocking` field since #2455/#2527; `|| true` now comes only from `suppress`. So a PR whose entire purpose was removing a false claim shipped one. All three fixed, and `grep` confirms none remain outside `docs/plans/`.

## 2. The same crash, still live in `migrations.py`

#2594 fixed the non-string-`command` crash in `hardlinks.py` and documented the premise — `null`, numbers, and lists really do arrive in a hand-editable `~/.claude/settings.json` — while leaving the two migrations that walk that same file on the same `/update` run unguarded. Both reproduced:

```
migrations.py:661  f"{hooks_root}/" not in command  -> TypeError: argument of type 'NoneType' is not iterable
migrations.py:829  stripped.endswith("|| true")     -> AttributeError on None and on 42
```

`run_pending_migrations` calls each migration with no `try`/`except`, so either takes down `/update`. Reachable on the `/setup` path, where these one-shot migrations are not yet recorded complete. Both now skip a non-string command, which can match no target and no legacy fork command anyway.

**Writing that test reproduced this lineage's defect one more time, and I want it on the record.** My first version passed `tmp_path` as `project_dir`. The rewrite migration loads `.claude/hooks/manifest.toml` from `project_dir` and returns before the loop when it is absent, so the test never reached the code it was written for and survived its own mutation while reporting green. It passes `_REPO_ROOT` now. Both guards fail under removal:

| Mutation | Result |
|---|---|
| Sweep `isinstance` guard removed | `test_migration_survives_a_non_string_command[legacy_unmarked_sweep]` fails |
| Rewrite `isinstance` guard removed | `test_migration_survives_a_non_string_command[manifest_id_rewrite]` fails |

## 3. `/update` could print a removal count with no lines

`_merge_hook_settings`'s marker-keyed removal pass incremented `result.removed` while emitting only a `"created"` action, so `run.py` printed `Removed N item(s) from ~/.claude/` followed by zero detail lines — for the deregistration the docs call the safe direction. It emits a matching `"removed"` action now, and the comment states the invariant: every pass that increments the count must emit an action.

## 4. Three accuracy items

- `hook-manifest.md` said "the other writers of this file snapshot it first." False for `_merge_hook_settings`, the sibling writer invoked on the immediately preceding line, which also uses a plain `write_text`. Narrowed to the migration writers, with the real distinction stated: migrations rewrite command strings where a bad rewrite is unrecoverable, while this pass and the merge only add or remove entries the manifest can reconstruct.
- `test_the_sweep_skips_a_non_string_command_rather_than_coercing_it` claimed more than it measured. Renamed to `test_a_non_string_command_neither_crashes_the_sweep_nor_is_removed`, with the surviving-coercion-mutant disclosure moved out of the PR body and into the docstring, where it outlives the PR. Given #2593 was specifically about names that overclaim, that belonged in the code.
- `test_an_unresolvable_hooks_root_fails_closed` asserted `is not None` where callers use the returned `Path`. Now `== hooks_root`, which kills the same mutant and additionally rules out a sentinel callers would misuse.

## Verification

`PYTHONPATH=<worktree> ./scripts/pytest-clean.sh tests/unit/test_update_hardlinks.py tests/unit/test_hook_migration.py tests/unit/test_hook_manifest.py tests/unit/test_hook_interpreter.py tests/unit/test_hooks_audit.py -q` — **156 passed**. `ruff check` and `ruff format --check` clean.

No behavior change on a healthy machine. The two migration guards only skip entries that could never have matched; the new `LinkAction` is log output; everything else is prose.

## DOCS

`docs/features/hook-manifest.md`: corrected `.bak` claim.


tomcounsell:
## Review: APPROVED

Verified against the PR head rather than the body.

**Both #2600 findings are genuinely closed.**

All three stale claims are gone from the source: `_build_hook_command` "appends `|| true` only for non-blocking declarations", "still registers blocking global hooks against it", and "applies the `|| true` guard the manifest's `blocking` flag calls for". The `blocking` mentions that remain (`hardlinks.py:901`, `migrations.py:543`, `:854`) describe the concept of a blocking hook rather than the removed field, and read correctly.

The non-string `command` guard is now present in `migrations.py` at the sites that walk the same hand-editable `settings.json` (`:667`, alongside the `:642` / `:646` / `:681` shape guards), matching what #2594 fixed in `hardlinks.py`.

**Tests and lint, run rather than credited:** `test_update_hardlinks.py` + `test_hook_migration.py` + `tests/unit/hooks/` = **166 passed**. `ruff check` clean. Merges clean against `main`.

This closes the loop opened when I merged #2594 while its review was still in flight. Filing #2600 from that review rather than letting the findings evaporate was the right call.

Issue #2060, opened 2026-07-13
test-isolation flake: test_cli_hook_denies_over_budget_exit_2 fails intermittently even in single-test isolation


valorengels:
## Context

Follow-up instance under umbrella #1897 (xdist test-isolation flakes). During validation of PR for #1897's fix, `tests/integration/test_tool_budget_enforcement.py::test_cli_hook_denies_over_budget_exit_2` was observed failing intermittently — exit code 0 instead of the expected 2 — **even when run completely standalone** (single test, `pytest path::test -q`, no cross-file ordering, no `--dist=loadfile` composition).

Reproduced on both `main` (pre-#1897-fix) and the #1897 fix branch, confirming this is a **separate, unproven root cause** from the two #1897 targeted (popoto db-cache split-brain; agent-hooks hooks-less-parent corruption). #1897's own Success Criteria explicitly downgrades this specific test's poisoning-ordering re-run to "best-effort, may pass vacuously" for exactly this reason — the authoritative acceptance for #1897 is the deterministic `tests/unit/test_conftest_isolation_guards.py::Test B` binding gate, which is unaffected.

## Repro

```
cd <repo root>
for i in 1 2 3 4; do
  pytest "tests/integration/test_tool_budget_enforcement.py::test_cli_hook_denies_over_budget_exit_2" -q
done
```

Observed: passes ~2-3 times out of 4, fails once with:
```
E   assert 0 == 2
E    +  where 0 = CompletedProcess(args=[... .claude/hooks/pre_tool_use.py], returncode=0, stdout='', stderr='').returncode
```
Empty stdout/stderr on the failing run — worth investigating whether the CLI-hook subprocess is silently short-circuiting (fail-open on missing session) or something in the write path (`make_session` → real Redis) hasn't landed before the subprocess reads it, independent of the popoto-db-cache mechanism #1897 already fixed.

## Suggested next step

Root-cause with a code-read/prototype spike (same method #1897 used for its two instances) before planning a fix. Log under the #1897 umbrella per its explicit "No-Gos" framing: future instances get filed here rather than expanding #1897's scope.

Related: #1897 (parent umbrella), #2037 (resolved by #1897's Fix 1).

PR #2606, opened 2026-08-07
Repair two shared-state leaks that make the suite's failure set unreproducible (#2603, #2605)

### LINKED ISSUE #2603: test_teammate_cold_start_finalize.py leaks across files under randomized ordering: 4 tests fail in a full run, all 9 pass alone
Four tests in `tests/unit/test_teammate_cold_start_finalize.py` fail during a full `tests/unit/` run under `pytest-randomly`'s default randomized ordering, and all nine pass when the file runs alone.

Failing under randomized order:

- `TestDefectBCompletionExitGuard::test_guard_own_status_conflict_error_treated_as_success`
- `TestDefectAPopLoopEscalation::test_primary_escalation_deletes_terminal_duplicate_pending_preserved_error_once`
- `TestDefectAPopLoopEscalation::test_last_resort_cancels_pending_with_reason_when_terminal_duplicate_has_children`
- `TestDefectAPopLoopEscalation::test_status_conflict_error_carries_real_session_id_on_pop_path`

## Evidence that ordering is the cause

Three runs of the same tree at `5011d8a8e`:

| Run | Ordering | Result |
|---|---|---|
| Full `tests/unit/` | randomized (default) | **5 failed**, 11808 passed |
| The file alone | `-p no:randomly` | **9 passed** |
| Full `tests/unit/` | `-p no:randomly` | **1 failed**, 11812 passed |

The one failure in the deterministic full run is `test_no_zombies_still_populates` (#2596), which is environment-dependent and unrelated. So under a fixed order the file is green in-suite; only randomized interleaving breaks it.

## Why this matters

This is the same class as #2551 (`test_reap_killlist.py` failing only alongside `test_session_lifecycle.py` / `test_recovery_ownership.py`) and #2469 (`TestFinalize` leaking lease state into `test_sdlc_stage_marker`), both addressed for specific file pairs. The general defect is still present: session and lifecycle state written by one test file is visible to another, and randomized ordering decides whether that is fatal on any given run.

The practical cost is that a full-suite failure count is not reproducible across runs, so "did my change break something" cannot be answered by comparing counts. That undermines the merge-gate baseline the same way #2573 did for machine-local state.

## Fix direction

Find the leaking writer rather than pinning the order. The four failures all sit in the pop-loop escalation and completion-exit-guard paths, which touch `AgentSession` status transitions and the pop lock, so the likely culprit is a sibling file that leaves sessions or lock keys behind. Per the repo's Redis hygiene rule, cleanup has to go through the ORM scoped by a recognizable `project_key`, never a bulk unscoped delete.

Found while verifying `main` after the 2026-08-06 bug batch landed. Not a regression from that batch: the failures do not reproduce under deterministic ordering, and the file passes alone.

### LINKED ISSUE #2605: TestKillCommandIntegration is order-dependent: its subprocess reads the main checkout and shared Redis, not the checkout under test
`tests/integration/test_agent_session_scheduler.py::TestKillCommandIntegration` fails non-deterministically when run alongside the other five session integration files, with a **different test failing each run**. All 29 pass in isolation.

Observed by the merge-queue agent while verifying tonight's PRs, and baselined against clean `main` specifically so it could not be blamed on a PR:

| Run | Failing tests |
|---|---|
| 1 (PR merged) | `test_kill_by_agent_session_id_pending`, `test_status_shows_killed_count` |
| 2 (PR merged) | `test_kill_by_session_id` |
| 3 (PR merged) | `test_kill_all_with_pending_jobs` |
| 4 (**clean main**) | `test_kill_all_with_pending_jobs`, `test_status_shows_killed_count` |
| isolation | 29 passed, on both merged and clean main |

Pre-existing, not a regression from this batch.

## Mechanism

These tests shell out to `/Users/tomcounsell/src/ai/.venv/bin/python3 -m tools.agent_session_scheduler`. Two consequences:

1. The subprocess reads the **main checkout's** code regardless of which checkout the test runs from, so a worktree cannot test its own scheduler.
2. It shares the Redis DB with every other test and every concurrent agent, so a row the test just created can be invisible to it, or a sibling file's fixture can remove it.

The observed failure is `"Session push-XXXX not found."` — the subprocess cannot see a row the test created moments earlier.

Concurrent agents on the machine make it strictly worse, which is how it surfaced tonight.

## Why it matters

Same false-signal family as the #2568 batch and #2603. A suite whose failure set changes run to run cannot answer "did my change break something", so the merge-gate baseline degrades into noise. This one is worse than most because the subprocess silently ignores the checkout under test: a real regression in a worktree's `agent_session_scheduler` would pass its own integration suite.

Related: `AI_REPO_ROOT` in `scripts/sdlc-tool` has the same main-checkout-pinning property, noted separately tonight when fixes on `origin/main` were found to be invisible to agents until the main checkout pulled.

## Fix direction

Have the subprocess inherit the checkout under test rather than an absolute path to the main venv, and give these tests an isolated Redis DB or a `project_key`-scoped keyspace so a sibling file cannot remove their rows. Per the repo's Redis hygiene rule, any cleanup goes through the ORM scoped by a recognizable key, never a bulk unscoped delete.

Reproduction is to run `tests/integration/` for the six session files together, repeatedly; the failing member rotates.

tomcounsell:
Closes #2603
Closes #2605

Two independent root causes. One PR because four of the ten touched files overlap (`tests/conftest.py`, `tests/unit/test_conftest_isolation_guards.py`, `tests/README.md`, `docs/features/test-isolation-hardening.md`), so split PRs would conflict with each other.

Both were reproduced before the fix and mutation-checked after it. Neither is fixed by pinning order or adding skips.

---

## #2603 — a reload splits a shared exception class in two

**The leaking writer:** `importlib.reload(models.session_lifecycle)`.

A reload keeps the module *object* but rebinds every class defined in it to a brand-new class. Every module that had already done `from models.session_lifecycle import StatusConflictError` keeps the OLD class — every test module, bound at collection, plus the function-local imports inside `agent/agent_session_queue.py::_worker_loop` and `agent/session_executor.py`'s completion-exit guard. After the reload, `except StatusConflictError` and `pytest.raises(StatusConflictError)` compare an exception against a class it is no longer an instance of, and silently miss. The damage lands in a different file from the one that reloaded, which is what made it look like an ordering flake.

**The specific state left behind**, by two writers, each sufficient alone:

| Node | What it left behind |
|---|---|
| `tests/unit/test_session_lifecycle.py::TestIssueLockTtlDefault::test_default_is_1800_seconds` | `models.session_lifecycle.StatusConflictError` rebound to a class no other module holds |
| `tests/unit/test_session_lifecycle_consolidation.py::TestImportSafety::test_lazy_imports_not_triggered_at_module_level` | same |

Running **either single node** before `tests/unit/test_teammate_cold_start_finalize.py`, serially, reproduces exactly the four failures named in the issue:

```
$ pytest-clean.sh -p no:randomly -n0 \
    "tests/unit/test_session_lifecycle.py::TestIssueLockTtlDefault::test_default_is_1800_seconds" \
    tests/unit/test_teammate_cold_start_finalize.py
FAILED ...::TestDefectBCompletionExitGuard::test_guard_own_status_conflict_error_treated_as_success
FAILED ...::TestDefectAPopLoopEscalation::test_primary_escalation_deletes_terminal_duplicate_pending_preserved_error_once
FAILED ...::TestDefectAPopLoopEscalation::test_last_resort_cancels_pending_with_reason_when_terminal_duplicate_has_children
FAILED ...::TestDefectAPopLoopEscalation::test_status_conflict_error_carries_real_session_id_on_pop_path
4 failed, 6 passed
```

**Both reloads were removable, and neither was earning anything.** The first recomputed a value the test asserts one line earlier is not env-overridden, so it could only ever produce the number the module already held. The second reloaded and then **asserted nothing at all** — an in-process reload cannot observe a *first* import, everything already being cached, so it could not have detected the module-level import it claimed to guard. It now runs its probe in a fresh interpreter and fails when a heavy dependency moves to module scope.

**The general defense**, so the class cannot return silently: `shared_exception_identity_guard`, an autouse fixture that snapshots the identity of every exception class in `_SHARED_EXCEPTION_MODULES` and rebinds any that changed at each test's teardown, then warns naming the test that reloaded. Same doctrine as #2551 — repair in place, one object per name, never pin the order. The warning is a teardown failure under `-W error::RuntimeWarning`.

---

## #2605 — a subprocess pointed at a db its parent does not own

**Not the checkout.** `python -m` puts the subprocess's cwd first on `sys.path`, and these tests already pass `cwd=<checkout under test>`, so a worktree's subprocess *was* resolving the worktree. Verified directly by printing `tools.agent_session_scheduler.__file__` from the subprocess: `/Users/tomcounsell/src/ai/.worktrees/ws-isolation/tools/agent_session_scheduler.py`. Correcting the issue's mechanism #1.

**The live defect is the Redis db.** `_subprocess_env` derived it from `PYTEST_XDIST_WORKER` (`gw{N} → db{N+1}`), the pre-#2060 rule. That agrees with the parent only when its *claim* happened to land on `worker_id + 1`. Any other live pytest process on the machine holding a lower slot moves the claim, and then the parent reads its claimed db while its own children read and write a db a different process is `flushdb`-ing at every test setup and teardown. Measured live, with another agent's run holding db1:

```json
{"parent_claimed_db": 2, "subprocess_db": 1, "match": false}
```

Three sibling call sites carried the same derivation: `test_bot_await_reply.py` and two in `test_email_bridge.py`. All four now route through one helper, `tests/db_claim.py::subprocess_env`.

**Why the claim moved out of `conftest.py`.** It is memoized in a module global backed by held file locks. Importing it as `tests.conftest` risks a second copy of that state, since pytest loads conftest through its own machinery — and a process holding two claims would hand its children a db it is not using, which is this bug again by another route. One plain module, imported normally by both conftest and the tests, cannot fork.

The checkout pinning stays anyway, as defense: the shared venv carries a `.pth` naming the main checkout, so the moment cwd stops winning a worktree would exercise `main`'s scheduler and could not fail for the right reason. `TestSubprocessIsolationContract` asserts all three properties — db match, a CLI-written row visible through the parent's ORM, and the checkout.

---

## Mutation-check evidence

Every fix was broken deliberately, shown red, restored, shown green.

| # | Mutation | Result |
|---|---|---|
| 1 | Re-add `importlib.reload(sl)` to `test_default_is_1800_seconds`, guard **active** | **10 passed** + `RuntimeWarning: ...test_default_is_1800_seconds reloaded models.session_lifecycle, rebinding StatusConflictError...; identity restored` |
| 2 | Same reload, guard **disabled** (`_SHARED_EXCEPTION_MODULES = ()`) — negative control | **4 failed, 6 passed**, the same four nodes |
| 3 | `ISSUE_LOCK_TTL_SECONDS` default `1800 → 1801` in source | RED: `assert 1801 == 1800` |
| 4 | Add module-level `import tools.session_tags` to `models/session_lifecycle.py` | RED: `heavy dependencies imported at module level: ['tools.session_tags']` — the previous version of this test could not have caught this |
| 5 | Delete the rebind from `_repair_shared_exception_identity` | RED: `assert <StatusConflictError> is <StatusConflictError>` |
| 6 | Revert `_subprocess_env` to the `PYTEST_XDIST_WORKER` derivation, db1 held by another process | RED ×2: `subprocess would use db=1 while this process uses db=2`, and `the CLI wrote to a keyspace this process cannot read` |
| 6b | Same mutation, db1 **free** | GREEN — the defect needs the concurrency, which is why it only ever showed up alongside other agents |
| 7 | Fix restored, db1 still held | **32 passed** |

## Suite results

- `tests/unit/{test_session_lifecycle.py::TestIssueLockTtlDefault, test_session_lifecycle_consolidation.py::TestImportSafety}` + `test_teammate_cold_start_finalize.py` + `test_conftest_isolation_guards.py` — **28 passed**
- `tests/integration/test_agent_session_scheduler.py` alone — **32 passed** (29 + 3 new), including under a held db1
- All seven touched integration files together — **151 passed, 3 failed**

The 3 are pre-existing on clean `main` and unrelated: `test_agent_session_lifecycle.py::TestSummarizeWithSession` imports `StructuredDraft` from `bridge.message_drafter`, where the symbol no longer exists. Present in the pre-change baseline; not touched here.


tomcounsell:
## Review: APPROVED

Verified independently rather than credited, including the parts that correct my own issue text.

**#2603 reproduction confirmed exactly.** On `main`, one node before the file produces precisely the four failures named in the issue:

```
pytest-clean.sh -p no:randomly -n0 \
  "tests/unit/test_session_lifecycle.py::TestIssueLockTtlDefault::test_default_is_1800_seconds" \
  tests/unit/test_teammate_cold_start_finalize.py
-> 4 failed, 6 passed
```

The same command on this branch clears all four. The `importlib.reload` diagnosis holds: a reload keeps the module object but rebinds its classes, so every module that already imported `StatusConflictError` keeps the old class and `except` / `pytest.raises` silently stop matching. The damage landing in a different file from the one that reloaded is exactly what made it read as an ordering flake.

Removing both reloads is right, and the reasoning for why neither was earning anything is correct: one recomputed a value the test asserts one line earlier is not env-overridden, the other reloaded and asserted nothing, and an in-process reload cannot observe a first import anyway.

**On #2605, this PR corrects my issue and is right to.** I wrote that the subprocess reads the main checkout. It does not: `python -m` puts cwd first on `sys.path` and the tests already pass `cwd=<checkout under test>`. The PR verified this directly by printing `tools.agent_session_scheduler.__file__` from the subprocess. The live defect is the Redis db derivation from `PYTEST_XDIST_WORKER` (the pre-#2060 rule), which agrees with the parent's claim only by luck. Measured mismatch (`parent_claimed_db: 2, subprocess_db: 1`) is the real mechanism. Keeping the checkout pinning as defense is a reasonable call given the shared venv `.pth`, and the stated reason for moving the claim into `tests/db_claim.py` rather than importing `tests.conftest` (avoiding a second memoized copy of lock-backed state) is sound.

**Mutation evidence is real**, including a negative control that disables the guard and reproduces the original four. The `_SHARED_EXCEPTION_MODULES` guard follows the #2551 doctrine: repair identity in place, warn naming the offending test, never pin the order.

**One note for the record, not a blocker.** While verifying I saw a *different* member of `test_teammate_cold_start_finalize.py` fail on both branch and main, rotating between runs. That is ambient Redis contention from two concurrent full-suite runs on this machine, not this PR: under identical load `main` showed 3 failures in that file and this branch showed 1. It is the pre-existing shared-db pressure this PR's `db_claim` work reduces rather than something introduced here.

Suite state before this merge: `tests/unit/` was fully green at `805b7f1ae` (11817 passed) under deterministic ordering.

PR #2035, opened 2026-07-11
fix(merge_predicate): resolve SDLC-tracked issue on multi-issue-closure PRs (#2034)

tomcounsell:
Closes #2034

## Problem
`tools/merge_predicate.py::_extract_issue_number()` used `.search()` (first match) against `Closes/Fixes/Resolves #N` in the PR body. On a multi-issue-closure PR under an umbrella tracking issue, the merge predicate keyed its DOCS (group b) and REVIEW-verdict (group c) checks on the FIRST closed sub-issue instead of the umbrella issue where the recorded SDLC state lives, producing false merge-gate failures (`no recorded REVIEW verdict`, `DOCS marker not authoritative`) that required a break-glass override on every such PR (observed on PR #2033 / issue #2029).

## Fix
Adds a tri-state tracked-issue resolver `_resolve_tracked_issue(head_ref, repo_root)`:
- Derives the branch slug via the existing `_derive_slug`; empty/`main`/`master`/`HEAD`/`session/` → no signal.
- Two broad `except Exception` guards (lazy imports; then the `AgentSession.query.filter(slug=...)` call) so import/query failure degrades to the body issue and never crashes the in-process merge-guard hook.
- Project-scopes candidates via `resolve_project_key(cwd=str(repo_root), env={})` (the `env={}` forces cwd-only resolution, neutralizing an ambient `VALOR_PROJECT_KEY`); a cross-project slug collision is discarded.
- Filters to live, non-transitional sessions (`NON_TERMINAL_STATUSES - {superseded, paused_budget}`) so a stale session cannot inflate the distinct-issue set.
- Distinct `issue_number` set: one → tracked; zero → no signal (body fallback); more than one → **ambiguous → fail closed** with an explicit `tracked-issue lookup ambiguous` entry rather than guessing.

Wired into `evaluate_merge_predicate`: groups (b)/(c) key on the tracked issue when resolved (with a substitution note), else the body issue (with a distinguishing note); group (a) body-link presence check is unchanged. Single-issue PRs are invariant (tracked == body).

## Scope
Diff touches only `tools/merge_predicate.py` and the new `tests/unit/test_merge_predicate.py` (scope fence per plan). No raw Redis; read-only `AgentSession` lookup.

## Tests
[...]

Issue #2034, opened 2026-07-11
merge_predicate._extract_issue_number resolves wrong issue on multi-issue-closure PRs


tomcounsell:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/merge-predicate-tracked-issue-resolution.md

## Problem

`tools/merge_predicate.py::_extract_issue_number()` uses `.search()` (first match only) against `Closes/Fixes/Resolves #N` in the PR body. When a PR closes multiple sub-issues under an umbrella tracking issue, the predicate resolves to the FIRST closed issue instead of the tracking issue where the SDLC substrate (REVIEW verdict, DOCS marker, stage state) actually lives.

## Repro (observed on PR #2033 / issue #2029)

PR #2033 had a body with:
```
Closes #1871
Closes #1267
Closes #1760
```
under umbrella tracking issue #2029. `_extract_issue_number` resolved to #1871 (which has zero SDLC substrate), producing false merge-gate failures:
- `"no recorded REVIEW verdict"`
- `"DOCS marker not authoritative"`

Ground truth on the correct issue #2029 was REVIEW=APPROVED (0 blockers), DOCS=completed, PR MERGEABLE/CLEAN, CI green. The merge had to proceed via the documented break-glass override (`data/merge_authorized_2033`).

## Impact

Hits **every** multi-issue-closure PR that follows this repo's umbrella-tracking-issue pattern (a PR that closes several sub-issues while its SDLC state is tracked on a separate umbrella issue). Each such PR will false-fail the merge predicate and require a manual override.

## Suggested fix

`_extract_issue_number` should not blindly take the first `Closes #N`. Options: prefer the issue that actually has SDLC substrate (query stage state for each candidate and pick the one with recorded verdicts/markers), or accept an explicit tracking-issue hint, or resolve against all matches and disambiguate by which carries the pipeline state.

Found during the #2029 SDLC run (PR #2033).
## Recon Summary

Verified against `tools/merge_predicate.py` @ `30fbebb6` (last touched by #2003 / #2010, `2f324bff`).

**Confirmed:**
- `_extract_issue_number()` (line 228-248) calls `_ISSUE_REF_CAPTURE_RE.search(body)` (line 237) — `.search()` returns the FIRST `Closes/Fixes/Resolves #N` match only. Multi-issue-closure bodies resolve to the first sub-issue, not the umbrella tracking issue.
[...]

Issue #2144, opened 2026-07-17
Resumed pipeline turns lose SDLC run identity — stage markers/verdicts silently fail after mid-turn interruption, ledger freezes while work proceeds


tomcounsell:
> **Context:** SDLC pipeline state (stage markers, verdicts) is written via `sdlc-tool` state-mutating subcommands that require a **run identity**: `sdlc-tool session-ensure` mints a `run_id`, persists it to `.sdlc-run` in the worktree, and pins it on a ≤300s-TTL Redis issue lock (`tools/sdlc_session_ensure.py`, issue #2003). Calls without a valid run-id fail with `RUN_ID_REQUIRED` (`tools/sdlc_dispatch.py:340-344`) — and the skill convention wraps marker writes in `2>/dev/null || true`, so those failures are **silent**. Observed live on 2026-07-17 during a worker restart mid-pipeline (session `0_1784286820434`, issue #2133).

## Problem

When a PM turn is killed mid-pipeline (worker restart — see #2141/#2143) and the session resumes via transcript, the resumed turn continues pipeline work **without re-establishing run identity**, so every subsequent stage-marker/verdict write silently fails and the ledger freezes while real work proceeds.

**Current behavior (observed):**
- Pre-restart (turn 1, run-id `1e3ab313…`): markers flowed correctly — `ISSUE/PLAN/CRITIQUE completed`, CRITIQUE verdict recorded 11:29:30 UTC.
- 11:33: worker restarted mid-turn; session resumed from transcript.
- Post-restart (turn 2): the resumed turn revised the plan, committed the fix, and **opened PR #2142 (11:43, CI green)** — but the ledger never moved: `BUILD` stayed `ready`, `TEST` stayed `pending`, no verdicts recorded. The old run-id lived only in the killed turn's context; the issue lock expired (≤300s TTL); no `session-ensure` re-ran; every marker write since presumably exited `RUN_ID_REQUIRED` into `|| true`.
- Turn 4 (13:35, after a second restart) *did* re-run `session-ensure` — `.sdlc-run` now holds a fresh run-id `6df037dd…`. Whether markers resume flowing in turn 4 is the live discriminator: if yes, the gap is "resumed turns skip session-ensure until something re-runs it"; if no, the refusal is deeper (foreign-run lease semantics).
[...]

PR #2614, opened 2026-08-07
Give the merge gate a satisfiable path for PRs that did not originate in the pipeline (#2577)

### LINKED ISSUE #2577: Merge gate has no satisfiable path for PRs that did not originate in the SDLC pipeline (dependabot and hand-authored)
Closing out PR #2526 and PR #2525 today required a break-glass override on **both**, for the same root cause. Neither PR had anything wrong with it. The gate has no representable path for a PR that did not originate inside the SDLC pipeline.

## What happens

`tools/merge_predicate.py` group (b) passes only when `stages.DOCS == completed`, degrading to a `docs/features/{slug}.md` existence check when the marker is `pending`/empty.

Reaching `DOCS == completed` calls `PipelineStateMachine.start_stage(..., backfill_predecessors=True)`, and `_backfill_predecessors` (`agent/pipeline_state.py:576`) refuses:

```
Cannot backfill predecessors of DOCS: CRITIQUE would be force-completed for
issue #N but carries no finalized verdict (verdict invariant unsatisfied)
```

That refusal is **correct**. CRITIQUE genuinely never ran, because there is no plan document to critique. The invariant is doing its job: it is telling us the stage did not happen.

The degraded fallback does not help either. It derives the slug from the PR head ref, so it only ever resolves for a `session/{slug}` branch. `fix/dangling-command-table-refs` and `dependabot/uv/uv-374350d79f` both produce a path that cannot exist.

The result is a gate that cannot be satisfied by any honest action. The only two ways through are to write a synthetic CRITIQUE verdict for a stage that never ran, which puts false stage completions into the durable ledger, or to break-glass.

## Skill and hook disagree about dependabot

`.claude/skills-global/do-merge/SKILL.md` has a **Dependabot Exemption** section instructing the skill to skip Steps 0, 2, 3, and 5 for a dependabot PR: no stage marker, no REVIEW verdict, no `Closes #N`.

`tools/merge_predicate.py` has no dependabot branch at all. It enforces the issue link, the DOCS gate, and verdict freshness on every PR. So the documented fast path is unreachable in practice: follow the skill exactly on a dependabot PR and the merge-guard hook blocks you on all three of the legs the skill just told you to skip.

This is the same skill-versus-hook drift class that `tools/merge_predicate.py` was introduced to eliminate (module docstring, "#1944 class"). The predicate unified the checks but the dependabot carve-out stayed behind in the skill body.

## Evidence

Both overrides today reduced to a single failing leg after all the substantive work was done:

- **#2526** (hand-authored docs fix): `DOCS marker not authoritative (status=<empty>) and docs/features/fix/dangling-command-table-refs.md absent`. Groups (a), (c), (d) all passed, with a real REVIEW verdict and a posted `## Review:` artifact.
- **#2525** (dependabot uv bump): identical leg, with `docs/features/dependabot/uv/uv-374350d79f.md`. Groups (a), (c), (d) all passed the same way.

Both override files were written with an `override:` reason, used once, and deleted immediately. Two `merge_guard.override_used` metrics should be visible on the dashboard.

## Worth fixing

Every break-glass that becomes routine stops being a signal. Two in one afternoon on two unrelated, healthy PRs means the next genuinely dangerous merge will pass the same way without anyone blinking.

Directions worth considering, not a decision:

1. Teach `evaluate_merge_predicate` the dependabot exemption the skill already documents, keyed on PR author and the `dependencies` label, so skill and hook agree again.
2. Give group (b) a defensible answer for a PR whose head ref is not `session/{slug}`: for example, pass when the PR diff touches no non-docs source, or when the tracked issue has no plan at all.
3. Let the DOCS gate accept a recorded "not applicable" disposition written through `sdlc-tool`, so the honest statement has somewhere to live other than an override file.

## Also observed

`gh pr edit --body` on a dependabot PR is **not durable**. Dependabot force-pushed a rebase mid-review and its bot-authored body replaced the edit, silently removing the `Closes #2576` line the gate requires. Anything that depends on editing a dependabot PR body has to re-check after every force-push.

tomcounsell:
Closes #2577

## The failure

`tools/merge_predicate.py` group (b) needs `stages.DOCS == completed`, and group (c) needs a recorded REVIEW verdict. Both writes route through `PipelineStateMachine._backfill_predecessors`, which walks the ISSUE-rooted spine and refuses:

```
Cannot backfill predecessors of REVIEW: CRITIQUE would be force-completed for issue #N
but carries no finalized verdict (verdict invariant unsatisfied)
```

The refusal is correct. Its remediation ("re-run CRITIQUE") is unsatisfiable for the dominant real shape: a dependabot bump, a hand-authored fix, or a follow-up filed from another PR's review findings has no plan document, so there is nothing to critique and no honest verdict to record. The degraded `docs/features/{slug}.md` fallback never resolves for those branches either.

That left forging a CRITIQUE verdict — which `tools/sdlc_stage_marker.py` says in as many words would turn any `--stage DOCS --status completed` call into a way to manufacture an approval — or breaking glass. Eleven merges broke glass in one night.

## The fix

The pipeline now distinguishes **"this stage has not run yet"** from **"this stage was never dispatched and does not apply"**. `skipped` joins the stage statuses. `completed` and `skipped` form `SETTLED_STATUSES`, which every ordering check accepts (predecessor satisfaction, backfill scan, `has_remaining_stages`, `next_stage`, the router's row-10 readiness test) while the two stay distinct on read.

The skip is **verified, not asserted**, and the verification lives in `tools/sdlc_stage_marker.py::_skip_precondition_error`:

- `SKIPPABLE_STAGES` is exactly `{PLAN, CRITIQUE}`. REVIEW, DOCS and MERGE are the legs the merge predicate reads; a skippable one would be a way to merge without the guarantee that stage exists to provide. Refused as `STAGE_NOT_SKIPPABLE` in the tool and again in `skip_stage`.
- A resolvable plan document for the issue refuses the skip (`PLAN_EXISTS_NOT_SKIPPABLE`). You cannot skip the CRITIQUE of an issue that has a plan.
- A recorded verdict, a recorded `_sdlc_dispatches` entry, or any status past `pending`/`ready` refuses it (`STAGE_RAN_NOT_SKIPPABLE`). A CRITIQUE that actually ran keeps its verdict requirement.
- The backfill treats an existing `skipped` as settled but **never writes one**. Every skip is an explicit, lease-gated call, so no `--stage DOCS --status completed` can produce one as a side effect.
- Every probe fails closed: an unreadable ledger or an errored plan lookup refuses the skip.

The disposition persists with its reason and timestamp under a `_stage_skips` ledger key.

## Also in scope

- **The Dependabot Exemption is gone** from `.claude/skills-global/do-merge/SKILL.md`. It instructed the skill to skip the stage marker, the REVIEW verdict, and the `Closes #N` link — none of which `tools/merge_predicate.py` has ever allowed, so the documented fast path was unreachable and following it blocked on all three legs. Dependabot PRs now run the same gate as everything else. The two genuinely dependabot-specific behaviors are kept: the `mergeable == "UNKNOWN"` retry, and a warning that a rebase silently rewrites the PR body and drops a hand-added issue link.
- **Spent break-glass overrides authorize nothing.** An override whose PR is already MERGED or CLOSED is classified `spent` and ignored (issue comment, three such files found still on disk). An unresolvable PR state still honors the override — break-glass has to work in a degraded environment.
- `docs/sdlc/do-merge.md`'s tracked-issue-resolution paragraph described the branch-slug `AgentSession` mechanism that PR #2035 shipped and #2034 replaced as empirically inert; it now describes the `PipelineLedger.pr_number` lookup that actually runs.

## Verification

`tests/integration/test_off_pipeline_merge_path.py` drives a real issue lease, a real `PipelineLedger`, and the real `write_marker` / `finalize` / `evaluate_merge_predicate` paths; only the live GitHub boundaries are stubbed. It encodes the pre-fix repro and all four required properties, including the forge attempts (skip REVIEW/DOCS/MERGE; complete REVIEW with the spine cleared; put REVIEW `in_progress` first to bypass the backfill branch; reach past REVIEW via the MERGE marker; skip a CRITIQUE that has a verdict; skip with a plan present; skip without the lease).

- `tests/unit/test_pipeline_state_machine.py` — 20 new cases for `skip_stage` and settled-status semantics
- `tests/unit/test_sdlc_stage_marker.py` — 18 new cases for the skip preconditions and write path
- `tests/unit/test_validate_merge_guard.py` — 6 new cases for spent-override classification
- `python -m ruff check` / `ruff format --check` clean

## Docs

New: `docs/features/off-pipeline-merge-path.md`. Updated: `docs/features/pipeline-state-machine.md`, `docs/sdlc/do-merge.md`, `docs/sdlc/do-pr-review.md`, `docs/tools-reference.md`, `.claude/skills/sdlc/SKILL.md`, `.claude/skills-global/do-merge/SKILL.md`, `docs/features/README.md`.


tomcounsell:
## Review: CHANGES REQUESTED — 1 blocker

The design is right and I verified the security property myself. Blocking only on six regressions.

### Verified good: the forge defenses hold

Attacked `skip_stage` directly, bypassing the CLI preconditions:

| Attempt | Result |
|---|---|
| `skip_stage("REVIEW" / "DOCS" / "MERGE")` | all refused; `SKIPPABLE_STAGES` is exactly `{PLAN, CRITIQUE}` |
| `skip_stage("CRITIQUE")` at `in_progress` / `completed` / `failed` | all refused |
| empty and whitespace-only reason | refused |
| never-run CRITIQUE | accepted, recorded with reason under `_stage_skips` |

Distinguishing "has not run yet" from "was never dispatched and does not apply" is the right shape, and keeping REVIEW/DOCS/MERGE permanently outside the skippable set is what makes it safe. Removing the unreachable Dependabot Exemption and treating spent overrides as authorizing nothing are good calls beyond the ask.

### BLOCKER: six tests regress against a green main baseline

Same five files, same machine, same load: **main 266 passed / 0 failed**, **this branch 6 failed / 319 passed**.

**The #2144 run-id self-heal is broken across all four tools.** Each refuses `RUN_ID_REQUIRED` where it must heal on a free lock:

```
tools.sdlc_dispatch must self-heal a missing --run-id on a free-lock issue (issue #2144),
not refuse RUN_ID_REQUIRED; got: '{"error": "RUN_ID_REQUIRED"}'
```

Identical for `sdlc_verdict`, `sdlc_stage_marker`, `sdlc_meta_set`. The probable root cause shows in the lock result:

```
IssueLockResult(acquired=True, owner_session_id=None, owner_run_id=None, ...)
assert None == '53ebedacdf0546558ac5120216d2f0d3'
```

The lock acquires without recording its owner, so the heal has nothing to attach to. One sibling failure surfaces as `LEASE_ABSENT`, consistent with the same cause.

This matters beyond the tests: `session-ensure` then `verdict finalize --run-id` is the exact path this PR exists to make usable. Trading the CRITIQUE deadlock for a broken run-id heal is not a net win.

**One unexplained error-code change**, possibly intentional but not mentioned in the body:

```
assert 'REVIEW_VERDICT_MISSING' == 'STATE_MACHINE_REJECTED'
```

### Also

The PR body states the suites are clean. They are not, on the branch as pushed. Please say what was run that reported green. Two other PRs tonight were caught on exactly this kind of verification claim, and the inaccurate claim is the more important defect of the two.

tomcounsell:
## Status: blocked, and pickup-ready

Re-measured at `7dbeb859d`, both trees on this machine within minutes of each other:

| Tree | Result |
|---|---|
| `main` @ `ea796f302` | **275 passed, 0 failed** |
| `session/ws-gate` @ `7dbeb859d` | **10 failed, 329 passed** |

The count moved **6 → 10** after the second push, so that commit went the wrong way.

### Do this first: rebase

The branch is one commit behind main, and that commit is **#2615, "Key issue-lock same-machine liveness on stable machine_id, not hostname"** — 73 changed lines in `models/session_lifecycle.py` plus a new `config/machine.py`. The failures live in exactly that area:

```
AssertionError: IssueLockResult(acquired=True, owner_session_id=None, owner_run_id=None, ...)
assert None == '53ebedacdf0546558ac5120216d2f0d3'
```

Some of the run-id-heal cluster may dissolve on the rebase. Nobody should rewrite lock-ownership logic to chase a failure a merge from main fixes. Rebase, re-run the five files, compare against **275**.

### Then: the design question `7dbeb859d` opens

The first design had a property I verified and want preserved in spirit: the backfill treats an existing `skipped` as settled but **never writes one**, so every skip is explicit, lease-gated, and verified. `7dbeb859d` moves the skip into the backfill, which converts a *structural* guarantee into a *predicate-correctness* guarantee. Before, a bug in the precondition could not mint a skip because no code path existed. Now one does.

The motivation is legitimate and I do not want it dropped: requiring a second command to have been run earlier is still an unsatisfiable gate for the agent standing in front of it. But it needs, with evidence:

1. Does the backfill's skip route run the **identical** predicate as the explicit call, or a re-implementation? If the latter, unify them.
2. Can it mint a skip for a CRITIQUE that has a plan document, a recorded verdict, or a recorded dispatch? Each must be shown refused **on the implicit path**, not only the explicit one.
3. Is it reachable for PLAN/CRITIQUE only, never REVIEW/DOCS/MERGE, implicitly as well as explicitly?
4. The explicit path fails closed when its probes raise. Confirm the implicit one does too.

If the implicit skip cannot be made safe, the fallback keeps the backfill read-only and instead has `verdict finalize` emit a precise refusal naming the exact `stage-marker --status skipped` command. A gate that tells you the one command to run is satisfiable, which was the original complaint, and it needs no mutation in the backfill.

### What is already verified good, and should not be redone

I attacked `skip_stage` directly, bypassing the CLI, against the first push:

- `skip_stage("REVIEW" / "DOCS" / "MERGE")` — all refused; `SKIPPABLE_STAGES` is exactly `{PLAN, CRITIQUE}`
- `skip_stage("CRITIQUE")` at `in_progress` / `completed` / `failed` — all refused
- empty and whitespace-only reason — refused
- never-run CRITIQUE — accepted and recorded with reason under `_stage_skips`

Removing the unreachable Dependabot Exemption and treating spent overrides as authorizing nothing are both good and beyond the original ask.

### Why this still matters

#2577 was hit **four** times in the last day, and twelve merges have gone through break-glass. This is the right fix to the right problem; it just is not green yet.

tomcounsell:
## Update: the six failures are branch-attributable, with a deterministic repro

My earlier comparison was single-sample on a suite I had myself filed #2628 about for having a non-reproducible failure set, which made it invalid. Redone properly, the answer did not change, and now it does not depend on sampling at all.

**Controlled repro.** One test node, `-n0` so no xdist, `-p no:randomly` so no ordering, both trees on the same machine seconds apart:

```
MAIN     test_sdlc_dispatch.py::TestRunIdRequiredFlag::test_missing_run_id_self_heals_on_free_lock
         -q -p no:randomly -n0   ->  4 passed
BRANCH   same node, same flags   ->  4 failed
```

No concurrency, no worker index, no shared-db race. The contention mechanism described earlier is real but is not what this is.

**Counts, five runs each, inline paths:**

| | `main` | `session/ws-gate` @ `205718f8d` |
|---|---|---|
| run 1 | 275 passed | 6 failed, 321 passed |
| run 2 | 1 failed, 274 passed | 6 failed, 321 passed |
| run 3 | 275 passed | 6 failed, 321 passed |

Also ruled out: the branch is **not** behind main (`git merge origin/main` reports "Already up to date"), and dropping the new `test_off_pipeline_merge_path.py` from the set changes nothing, so it is not a collision with the new integration file.

**Hypothesis for where to look.** All four self-heal params fail, including `dispatch-record` and `meta-set` whose tools this diff does not touch. What they share is the `heal_missing_run_id` chain. This diff adds a new edge into that neighbourhood: `PipelineStateMachine._skip_precondition` lazily imports `tools.sdlc_stage_marker._skip_precondition_error`, creating an `agent/pipeline_state.py` → `tools/sdlc_stage_marker.py` dependency. An import cycle through that edge causing the heal to return `None` would produce exactly the observed `RUN_ID_REQUIRED` and `owner_run_id=None` signature, and would explain untouched tools failing.

That is a hypothesis. The repro is the finding, and it bisects cheaply now that it reproduces on a single node at `-n0`.

## Unchanged from my last review

The design decision is settled: **keep the implicit auto-skip.** The guarantee the merge gate rests on — the backfill can never mint a REVIEW/DOCS/MERGE disposition — remains structural via short-circuited frozenset membership. What moves to predicate-correctness is an unwarranted PLAN/CRITIQUE skip, whose worst case is a dishonest disposition on two stages `tools/merge_predicate.py` never reads.

The `verdict_invariant_satisfied` fix remains the most valuable thing here: `finalize` wrote verdict and trailer before the artifact check, and the invariant checked only verdict and trailer, so a correctly-refused `REVIEW_ARTIFACT_MISSING` run left behind exactly what a later `--stage DOCS --status completed` needed to force-complete a REVIEW nobody posted. Pre-existing, unreachable only because CRITIQUE blocked first.

#2577 has now been hit **five** times; the most recent was PR #2624 an hour ago, which broke glass for the same reason. This fix is wanted.

tomcounsell:
## Re-verified at `6f472dce9`: still blocked, now with a one-command repro

**Correction to my previous comment first.** The branch head I compared against (`205718f8d`) did not contain PR #2624, and my `git merge origin/main` check reported "Already up to date" against a stale local ref. That comparison was invalid. The conclusion survives on new evidence, but the earlier reasoning did not.

**The repro needs no pytest, no shared Redis db, no ordering, and no worker index.** Run from each checkout with `PYTHONPATH` set to it, against issue numbers that have never existed:

```
python -m tools.sdlc_dispatch record --skill /do-build --issue-number <virgin-number>
```

```
issue 987001  BRANCH: {"error": "RUN_ID_REQUIRED"}
issue 987001  MAIN:   {"ok": true, "history_length": 1}
issue 987002  BRANCH: {"error": "RUN_ID_REQUIRED"}
issue 987002  MAIN:   {"ok": true, "history_length": 1}
```

Order-independent. Main-first on a virgin number succeeds; the branch on the next virgin number still fails:

```
issue 987003  MAIN:   {"ok": true, "history_length": 1}
issue 987004  BRANCH: {"error": "RUN_ID_REQUIRED"}
```

**The diagnostic tell.** Once main has created state for an issue, the branch succeeds on that same issue:

```
issue 999999  BRANCH: RUN_ID_REQUIRED       <- cold
issue 999999  MAIN:   ok, history_length 1  <- creates the session
issue 999999  BRANCH: ok, history_length 2  <- now warm, branch works
```

So the regression is not in finding an identity. It is that **the branch cannot create the session on a cold issue** where main can. That explains why all four self-heal params fail, including `dispatch-record` and `meta-set` whose tools this diff does not touch: they do not need the changed code, they need a session to exist, and creation is what regressed.

Also at `-n0` with `-p no:randomly`, three alternating runs: main **4 passed** each time, branch **4 failed** each time.

My earlier import-cycle hypothesis is wrong and should be discarded — `tools._sdlc_run_identity.heal_missing_run_id` imports cleanly on the branch.

Where I would look: whether constructing or backfilling `PipelineStateMachine` for a brand-new issue now routes through `_skip_precondition` / `_qualifies_as_never_dispatched`, and whether that refuses when the ledger is completely empty. A cold issue is the one case where every probe has no data, and those probes are documented as failing closed. Failing closed is correct for a skip decision and incorrect for session creation.

## Unchanged

The in-pipeline documentation added in `6f472dce9` is what was asked for. The design decision stands: keep the implicit auto-skip. The `verdict_invariant_satisfied` artifact fix remains the most valuable change here, and #2577 has now been hit five times.

tomcounsell:
## Retraction: the six failures are environmental. This PR causes no regression.

I blocked this PR three times on a regression that does not exist. Here is the control I should have run first.

**The confound: I was comparing `main` in the main checkout against the branch in a fresh worktree.** The variable was the directory, not the code.

```
988001  MAIN code,   fresh worktree      -> {"error": "RUN_ID_REQUIRED"}
988002  BRANCH code, fresh worktree      -> {"error": "RUN_ID_REQUIRED"}
988003  MAIN code,   MAIN checkout       -> {"ok": true, "history_length": 1}
```

Main fails exactly the same way when run from a worktree. The SDLC CLIs need a checkout that `projects.json` maps; from an unmapped worktree a cold issue cannot mint a session, and `heal_missing_run_id` returns `None`, which surfaces as `RUN_ID_REQUIRED`. That mimics a code regression precisely.

**Apples-to-apples, both trees from equivalent fresh worktrees:**

| run | `main` worktree | `session/ws-gate` worktree |
|---|---|---|
| 1 | 6 failed, 269 passed | 6 failed, **321** passed |
| 2 | 6 failed, 269 passed | 6 failed, **321** passed |

Identical failure counts. The branch adds 52 passing tests and **zero** new failures. And on the single node, both trees fail 4/4 from a worktree while main passes 4/4 from the main checkout.

**Every claim I made against this PR was an artifact of an uncontrolled comparison**, four times running: a single sample on a suite I had myself filed #2628 about for non-reproducible failures; a stale `origin/main` ref that made a `git merge` check report "Already up to date"; and finally this directory asymmetry. The author's structural argument — that two of the four failing tools are untouched by the diff and never enter the changed code — was correct from the start and I should have treated it as a reason to hunt for my own confound rather than as something to out-argue with counts.

## Verdict: APPROVED

Everything I actually verified holds up:

- **The forge defenses are real.** Attacked `skip_stage` directly, bypassing the CLI: REVIEW/DOCS/MERGE all refused (`SKIPPABLE_STAGES` is exactly `{PLAN, CRITIQUE}`), a CRITIQUE at `in_progress`/`completed`/`failed` refused, empty and whitespace reasons refused, a never-run CRITIQUE accepted and recorded with its reason.
- **The guarantee the merge gate rests on stays structural.** Frozenset membership is short-circuited before the predicate is consulted, so a predicate bug cannot reach a non-skippable stage. What moves to predicate-correctness is an unwarranted PLAN/CRITIQUE skip, whose worst case is a dishonest disposition on two stages `tools/merge_predicate.py` never reads.
- **The `verdict_invariant_satisfied` fix is the most valuable change here.** `finalize` wrote the verdict and trailer before checking for a posted artifact, and the invariant checked only verdict and trailer, so a correctly-refused `REVIEW_ARTIFACT_MISSING` run left behind exactly what a later `--stage DOCS --status completed` needed to force-complete a REVIEW nobody posted. Pre-existing, unreachable only because CRITIQUE blocked first, and surfaced by an acceptance test I expected to pass trivially.
- **The in-pipeline behaviour change is documented** in `off-pipeline-merge-path.md`, framed against what the old code did in the same situation.

#2577 has been hit five times in two days and every one of those merges broke glass. Merging.

Issue #2305, opened 2026-07-23
SDLC multi-pipeline coordination hardening: hollow-session liveness mirage, xdist gate deadlock, cleanup-vs-gate race, verdict-less REVIEW markers


valorengels:
> **Context:** On 2026-07-23 a batch of seven `bug`-labeled issues (#2203, #2204, #2205, #2207, #2208, #2211, #2212) was shipped end-to-end by seven concurrent local `/do-sdlc` supervision runs on one machine. All seven merged, but the batch surfaced four recurring coordination defects in the SDLC infrastructure. `/do-sdlc` is the local pipeline supervisor skill (`.claude/skills-global/do-sdlc/SKILL.md`); `sdlc-local-{N}` sessions are AgentSession tracking records minted by `sdlc-tool session-ensure`; the merge lease / issue lock are the single-owner guards in `agent/sdlc_router.py` + `tools/merge_predicate`.

## Problem

Under multi-pipeline load, four distinct failure modes each independently stalled pipelines. Every stall required manual orchestrator intervention; an unattended batch would have wedged indefinitely.

**Current behavior** (each observed at least once, most repeatedly):

### 1. Liveness mirage: hollow tracking sessions look alive forever

`sdlc-local-{N}` sessions created by `session-ensure` are pure tracking records — `claude_pid=None`, `harness_pid=None`, `turn_count=0`, no heartbeats ever. But `updated_at` is refreshed by monitoring probes and sibling `session-ensure` calls, so `_run_id_has_live_session()` reports them live indefinitely. Consequences observed:

- Five successive builder/supervisor contexts stood down on #2211, each deferring to a "live eng session `sdlc-local-2211`" that had zero processes — the worktree had been frozen for 3.7 hours while every agent inferred liveness from `updated_at`.
- Merge leases for dead runs stayed `orphaned_lock: false`, refusing merges of fully-approved PRs #2298 and #2297 for 30+ minutes each. Unblocking required manually killing the hollow session (`valor-session kill --id sdlc-local-{N}`).

This directly contradicts the repo's single-authoritative-liveness principle (session_runner owns liveness; no multi-field inference).

### 2. xdist finalization deadlock on merge full-suite gates (recurred twice)

[...]

Issue #2012, opened 2026-07-10
SDLC pipeline deadlocks at MERGE when a takeover session finishes an issue whose driver session went terminal (stage state lost)


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/sdlc-issue-keyed-stage-ledger.md

> **Context for the reader.** This repo drives feature work through an **SDLC pipeline** (Issue -> Plan -> Critique -> Build -> Test -> Review -> Patch -> Docs -> Merge). Pipeline progress is not inferred from artifacts; it is tracked explicitly in Redis as **stage state** keyed to the **PM/driver session** that owns a GitHub issue (see [`docs/features/sdlc-stage-tracking.md`](../blob/main/docs/features/sdlc-stage-tracking.md)). The `MERGE` stage will only fire when a **passing REVIEW verdict** is recorded for the issue. This issue reports a state-loss defect that leaves a fully-mergeable PR stuck because that recorded state is missing.

## Problem

**Current behavior.** PR #2008 (issue #1997, "Centralize machine-name resolution into `config/machine.py`") is functionally complete and mergeable, but the SDLC pipeline cannot advance it to MERGE.

Ground truth I verified:

- `gh pr view 2008` -> `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`, `isDraft: false`, no failing CI (`gh pr checks 2008` -> "no checks reported"; this repo runs no required GitHub checks).
- The code is clean: `tools/machine_identity.py` is correctly deleted, no dangling imports of it remain (remaining `machine_identity` hits are all the unrelated `check_machine_identity` function in `scripts/update/verify.py`), and the module is internally consistent (`get_machine_display_name` is spelled the same in the module, callers, and tests — the PR *description* prose misspells it as `get_display_machine_name`, but the code is fine).
- I ran the six affected unit test files serially in a fresh worktree on the branch tip (`e77168e0`): **89 passed** (`test_config_machine`, `test_bridge_update`, `test_crash_recovery_gates`, `test_reflection_arm`, `test_reflection_register`, `test_worker_sentry_init`). No regression from this PR.

So nothing is *failing*. What blocks merge is missing pipeline state:

```
$ sdlc-tool stage-query --issue-number 1997
{"stages": {}, "_meta": {"latest_critique_verdict": null,
 "latest_review_verdict": null, "pr_number": null, ...}}
$ sdlc-tool verdict get --stage REVIEW --issue-number 1997
{}
[...]

Issue #2446, opened 2026-07-29
Local /do-sdlc cannot finish a pipeline: issue lease expires mid-stage, supervisor mistakes its own orphaned session for a rival and loses stage markers


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/sdlc-run-self-recognition.md

## Summary

A local `/do-sdlc` supervision run cannot complete any pipeline whose stages take longer than the issue-lease TTL (~300s). The lease is minted by `sdlc-tool session-ensure` and is **not renewed while a stage subagent runs**. When it lapses mid-stage, the supervisor's own orphaned `sdlc-local-{N}` session record starts reading as a *live foreign owner*, and the supervisor — correctly following the documented "treat a foreign-owned lock like a router block: stop and report" rule — stands down in favour of a rival that does not exist.

Observed on three independent runs in one session (#2420, #2421, #2422). All three stalled at post-CRITIQUE with zero PRs. Every stage subagent's work was correct; only the state substrate failed.

## Failure chain

1. `session-ensure` mints `run_id` A and takes the issue lock (TTL ≤300s).
2. Supervisor dispatches a stage subagent (PLAN/CRITIQUE routinely run several minutes). Nothing renews the lease during this window.
3. Lease lapses. The `sdlc-local-{N}` session record persists.
4. A state-mutating call (`stage-marker`, `verdict record`) is refused: `ISSUE_LOCKED run_id=<B>` or `{"blocked": true, "reason": "SUPERVISED_RUN_ACTIVE", "owner_session_id": "sdlc-local-{N}"}`.
5. Supervisor reads the refusal as proof of a second live supervisor and stops.
6. Because the marker writes were refused, **completed stage state is silently lost**. `sdlc-tool stage-query` returns `{"stages": {}}` and `next-skill` regresses to `/do-plan` — for an issue whose plan is already finished and committed.

## Evidence

- #2422: supervisor reported the incumbent as `sdlc-2422` — **its own agent name**. `stage-query --issue-number 2422` returned completely empty stages despite a committed plan (`ba9f7f9c4`) and surviving critique forensics in `.critique-runs/2422-1785311700931707000`.
- #2421: `session-ensure` returned `owner_session_id: "sdlc-local-2421"` — **its own tracking session**. Left the issue hard-blocked on guard G7 (`plan_revising` set, no recent `/do-plan` dispatch).
[...]

PR #2615, opened 2026-08-07
Key issue-lock same-machine liveness on stable machine_id, not hostname

### LINKED ISSUE #2537: SDLC issue lock: renaming the host makes _lock_owner_is_live fail open forever, stranding resumed runs for the full TTL
> **Context for readers new to this repo.** The SDLC pipeline lets several agent runs work different GitHub issues at once. To stop two runs from grabbing the same issue, each issue gets a Redis **issue lock** (`session:issuelock:{N}`) whose payload records who owns it. Issue #2305 replaced an unreliable ownership test (the "liveness mirage" — any non-terminal session row made a dead run look alive) with a single authoritative check, [`_lock_owner_is_live`](../blob/main/models/session_lifecycle.py#L897), which trusts only what was stamped into the lock payload: `pid` + `hostname` + `create_time`. This issue reports that the new check has a hole that swallows it whole.

## Problem

`_lock_owner_is_live` decides "is the run holding this lock still alive?" by first asking "is this lock even from my machine?" — and it answers that question with `socket.gethostname()`:

```python
# models/session_lifecycle.py:933-934
if hostname != socket.gethostname():
    return True
```

A foreign host's pid genuinely cannot be checked locally, so failing toward "assume live" is defensible on its own terms. The bug is that **`hostname` is a mutable label, not a machine identity**. Rename the Mac and every lock payload stamped before the rename starts reporting itself as belonging to a different machine — while sitting on the same machine, with the pid right there, locally checkable.

That is not hypothetical. It stranded a resumed run on #2518 today:

| | value |
|---|---|
| hostname in the stranding lock payload | `Mac.local` |
| `socket.gethostname()` on this host now | `Valor-the-Cowboy.local` |
| `scutil --get ComputerName` | `Valor the Cowboy` |

Same machine, renamed. So `_lock_owner_is_live` took the cross-host branch and returned `True` — permanently. Both protections #2305 added (dead-pid detection and the pid-recycling `create_time` guard at `models/session_lifecycle.py:939-946`) were skipped entirely, and never ran again for the life of that lock.

**Two things make it permanent rather than transient.**

1. **The renewal path never refreshes the stale name.** The self-healing renewal branch (`models/session_lifecycle.py:1145-1170`) deliberately spreads the existing payload and overrides *only* `target_repo`, with a comment explaining it must "never reconstruct a subset — that would silently drop `pid`/`hostname`". Correct for its own purpose (issue #2012), but the consequence is that `hostname` is written once at acquire time (`models/session_lifecycle.py:1104`) and is never corrected. A lock stamped `Mac.local` renews forever as `Mac.local`.

2. **The heartbeat ignores the orphan signal it is already handed.** `touch_issue_lock`'s peek computes `orphaned_lock=not _lock_owner_is_live(payload)` (`models/session_lifecycle.py:1078`) and returns it on the result object. But `run_heartbeat` reads only `owner_run_id` from that same peek (`tools/sdlc_lease_heartbeat.py:110-111`) and renews on a run_id match alone (`:123-128`). Nothing in the renew loop consults `orphaned_lock`.

The net effect is a **fail-open on a gate whose entire purpose is to fail closed** — the same class as the stale-head-SHA verdict gate documented in [`docs/features/gh-stale-state-verdict-gate.md`](../blob/main/docs/features/gh-stale-state-verdict-gate.md). The only remaining backstop is the lock TTL, `ISSUE_LOCK_TTL_SECONDS`, default **1800s / 30 minutes** (`models/session_lifecycle.py:847`). Every resumed run on a renamed machine waits out the full TTL behind a lease nobody holds.

### Current behavior

A lock acquired before a host rename is treated as live forever, regardless of whether its owning process is dead. A resumed SDLC run blocks until the 30-minute TTL lapses.

### Desired outcome

Ownership liveness is keyed on something that survives an operator renaming their Mac, so a dead same-machine owner is detected as dead on the first check.

## Corrections to the original report

Filed from an operator summary; two details in it did not survive contact with the source, and are corrected above rather than repeated:

- *"`sdlc_lease_heartbeat` held a lease for a dead supervisor."* Not the mechanism. The heartbeat is the supervisor's own in-process 60s tick (`tools/sdlc_lease_heartbeat.py:4`), so it dies when the supervisor dies; the lease then simply persists to TTL. The heartbeat's real defect is narrower — it ignores `orphaned_lock`.
- *"There is no owner-liveness check on the held lease."* There is one, and it is well-built: `_lock_owner_is_live`. It is defeated by the stale-hostname comparison, not absent.

One point from the report is confirmed as written: when a supervisor is minted outside `/do-sdlc` there is no heartbeat process at all, so the lease only ever lapses on TTL.

## Definitions

| Term | Meaning |
|---|---|
| **Issue lock** | Redis key `session:issuelock:{N}` granting one run exclusive claim on GitHub issue N. Payload: `run_id`, `session_id`, `pid`, `hostname`, `create_time`, `target_repo` (`models/session_lifecycle.py:1104`). |
| **Lease heartbeat** | `tools/sdlc_lease_heartbeat.py` — peek-then-renew loop that extends the lock TTL under the owner's own identity. Never re-acquires; re-acquiring would be lease-theft (its Risk 2, `:112-113`). |
| **`orphaned_lock`** | Field on the peek result meaning "the payload's owner is verifiably dead" — `not _lock_owner_is_live(payload)` (`models/session_lifecycle.py:1078`). |
| **Fail-open** | A check that returns "permit" when it cannot reach a verdict. Correct for unverifiable cases; a defect when the case *was* verifiable. |
| **pid recycling** | The OS reissuing a dead process's pid to an unrelated process. Guarded by comparing `create_time` (`models/session_lifecycle.py:939-946`) — a guard the cross-host branch skips. |
| **Liveness mirage** | The pre-#2305 bug where any non-terminal `AgentSession` row made a dead run read as live. |

## Open questions for `/do-plan`

The root cause is structural — identity keyed on a mutable label — so these are questions, not a prescribed fix:

1. **What is the stable machine identity?** This repo already has a machine-identity concept: `projects.<key>.machine` in `projects.json`, the source of truth for [single-machine ownership](../blob/main/docs/features/single-machine-ownership.md). Should the lock payload adopt that key, a hardware UUID, or record both and compare on the stable one while keeping `hostname` as a human-readable breadcrumb?
2. **Should cross-host stay fail-open at all?** #2305's intent was to fail *closed* on verifiable death. With TTL as the backstop, is "assume live" still the right answer for a genuinely foreign host, or should a foreign lock be reclaimable at a shorter horizon?
3. **Immutable identity vs. self-healing renewal.** Should the renewal branch refresh identity fields (contradicting its deliberate `:1154` comment), or should identity be immutable and any mismatch simply barred from renewal so the lock lapses at TTL? These pull in opposite directions and the plan should pick one.
4. **Migration for locks already in the wild.** Payloads stamped with a now-stale hostname exist right now. Do they self-heal, or are they left to lapse?

## Acceptance criteria

1. `_lock_owner_is_live` correctly reports a dead owner as dead when the lock was acquired on the same physical machine under a previous hostname.
2. Lock ownership is compared on an identifier that does not change when the operator renames the machine.
3. `run_heartbeat` stops renewing when its peek reports `orphaned_lock=True`, rather than renewing on `owner_run_id` match alone.
4. The no-lease-theft invariant is preserved: the heartbeat still never re-acquires a lock it does not own (`tools/sdlc_lease_heartbeat.py:112-113`).
5. Genuine cross-host locks, and payloads that are truly indeterminate (missing `pid` or `create_time`), still fail toward live — this must not become a hard-kill on unverifiable evidence.
6. Regression test: lock acquired under hostname A, current hostname B, same machine, owner pid dead → owner reported not-live. This is the exact #2518 shape and must fail before the fix.
7. Regression test: heartbeat exits without renewing when the peek reports `orphaned_lock=True`.
8. Existing `_lock_owner_is_live` coverage in `tests/unit/test_session_lifecycle.py:537-590` still passes or is updated with a stated reason.

## Downstream context

This issue will be consumed by `/do-plan` to produce a plan document at `docs/plans/{slug}.md`, then executed by `/do-build`. Related: #2305 (introduced `_lock_owner_is_live`, CLOSED), #2012 (`target_repo` pinning and the self-healing renewal branch), [`docs/features/sdlc-run-identity-self-heal.md`](../blob/main/docs/features/sdlc-run-identity-self-heal.md), [`docs/features/sdlc-stage-tracking.md`](../blob/main/docs/features/sdlc-stage-tracking.md).

Found while resuming the SDLC run on #2518; not caused by that work and independently fixable.

valorengels:
Closes #2537

## Problem

`_lock_owner_is_live` decided "is this lock from my machine?" by comparing the payload's `hostname` against `socket.gethostname()`. Hostname is a mutable label: after a machine rename, every pre-rename lock payload took the cross-host branch and returned live **forever**, skipping the dead-pid and pid-recycling guards #2305 added. A resumed run stranded behind a lease nobody held for the full 30-min TTL (the #2518 incident). Compounding it, `run_heartbeat` renewed on `owner_run_id` match alone and ignored the `orphaned_lock` signal its own peek already computed.

## Design (per the issue's open questions)

1. **Stable identity** = hardware machine id: `config.machine.get_machine_id()` — macOS `IOPlatformUUID` via `ioreg`, `/etc/machine-id` elsewhere; cached; fail-soft to `""`. Survives renames, unlike ComputerName / hostname / `projects.json` machine (all operator-editable labels). `hostname` stays in the payload as a human-readable breadcrumb.
2. **Cross-machine stays fail-open** toward live (AC5); the lock TTL remains the backstop for a genuinely foreign dead owner.
3. **Renewal self-heals** a missing `machine_id` (same pattern as the #2012 `target_repo` heal — safe because renewal requires a run_id match and run_ids are minted on exactly one machine). Identity is immutable once stamped: an existing `machine_id` is never overwritten; an unresolvable local id (`""`) is never written.
4. **Migration**: legacy payloads without `machine_id` keep today's hostname comparison until their next same-owner renewal heals them, or they lapse at TTL.
5. **Heartbeat** (AC3/AC7): exits without renewing when the peek reports `orphaned_lock=True`. Still peek-first, renew-only — never re-acquires (no lease-theft, AC4 / Risk 2 preserved).

## Acceptance criteria → tests

| AC | Test |
|---|---|
| 1, 6 (#2518 shape: hostname A → B, same machine, dead pid → not live; failed before fix) | `TestLockOwnerIsLiveMachineId::test_renamed_host_same_machine_dead_pid_is_not_live` |
| 2 (rename-stable identifier) | `test_acquire_stamps_machine_id`, `test_renamed_host_same_machine_live_pid_is_live`, `config.machine` id tests |
| 3, 7 (heartbeat stops on orphaned_lock) | `test_exits_without_renew_when_peek_reports_orphaned_lock` |
| 4 (no lease-theft) | same test asserts peek-only, no mutating call; existing peek-first tests unchanged |
| 5 (foreign / indeterminate still fail toward live) | `test_foreign_machine_id_fails_toward_true`, `test_unresolvable_local_machine_id_falls_back_to_hostname`, existing indeterminate tests unchanged |
| 8 (existing coverage) | all pre-existing `TestLockOwnerIsLive` / `TestTouchIssueLock` tests pass unchanged; one test-fixture update: `_peek_result` now sets `orphaned_lock=False` explicitly (a bare MagicMock attribute is truthy and would falsely trip the new guard) |

## Verification (RED → GREEN)

- RED: 12 new tests failed before the implementation (including the exact #2518 regression shape).
- GREEN: `scripts/pytest-clean.sh tests/unit/test_session_lifecycle.py tests/unit/test_sdlc_lease_heartbeat.py tests/unit/test_config_machine.py` → **104 passed**.
- Adjacent consumers: `tests/unit/test_sdlc_session_ensure.py tests/unit/test_sdlc_next_skill.py` → **129 passed**.
- `ruff check` + `ruff format --check` clean on all touched files.

### DIFF
--- a/config/machine.py
+++ b/config/machine.py
@@ -154,7 +154,12 @@ def get_machine_id() -> str:
     ``/etc/machine-id`` (or the dbus fallback). Fail-soft to ``""`` -- callers
     treat an unresolvable id as indeterminate and fall back to hostname
     comparison, never raise. Cached for the process lifetime (the value is
-    immutable per machine and the lookup shells out).
+    immutable per machine and the lookup shells out). Note the cache also
+    memoizes a FAILED lookup: a process that once resolved ``""`` keeps
+    answering ``""`` until it restarts. Acceptable because every consumer
+    treats ``""`` as indeterminate-with-fallback, and the processes involved
+    (CLIs, heartbeat, worker turns) are short-lived relative to a transient
+    ``ioreg`` failure clearing.
     """
     try:
         if sys.platform == "darwin":
--- a/config/machine.py
+++ b/config/machine.py
@@ -36,10 +36,14 @@ matching. All former consumers now share this stricter contract.
 
 from __future__ import annotations
 
+import functools
 import json
 import logging
+import re
 import socket
 import subprocess
+import sys
+from pathlib import Path
 
 from config.paths import VALOR_DIR
 
@@ -134,6 +138,48 @@ def get_machine_slug() -> str:
         return name.lower().replace(" ", "-")
     slug = get_machine_display_name().split(".")[0].lower().replace(" ", "-")
     return slug or "unknown"
+
+
+@functools.cache
+def get_machine_id() -> str:
+    """Return this machine's stable hardware identity; ``""`` on failure.
+
+    Unlike hostname / ComputerName / ``projects.json``'s ``machine`` field --
+    all mutable labels an operator can rename -- this identifier survives a
+    machine rename, so it is what same-machine comparisons key on (issue
+    #2537: the SDLC issue-lock liveness check failed open forever after a
+    rename because it compared hostnames).
+
+    Resolution: macOS ``IOPlatformUUID`` via ``ioreg``; elsewhere
+    ``/etc/machine-id`` (or the dbus fallback). Fail-soft to ``""`` -- callers
+    treat an unresolvable id as indeterminate and fall back to hostname
+    comparison, never raise. Cached for the process lifetime (the value is
+    immutable per machine and the lookup shells out).
+    """
+    try:
+        if sys.platform == "darwin":
+            result = subprocess.run(
+                ["ioreg", "-rd1", "-c", "IOPlatformExpertDevice"],
+                capture_output=True,
+                text=True,
+                timeout=_SCUTIL_TIMEOUT_SECONDS,
+            )
+            if result.returncode == 0:
+                match = re.search(r'"IOPlatformUUID"\s*=\s*"([^"]+)"', result.stdout)
+                if match:
+                    return match.group(1)
+            logger.debug("ioreg IOPlatformUUID lookup failed (rc=%s)", result.returncode)
+        else:
+            for path in ("/etc/machine-id", "/var/lib/dbus/machine-id"):
+                try:
+                    value = Path(path).read_text().strip()
+                except OSError:
+                    continue
+                if value:
+                    return value
+    except Exception as exc:
+        logger.debug("machine-id lookup failed: %r", exc)
+    return ""
 
 
 def get_machine_project_keys(machine: str | None = None) -> list[str]:
--- a/docs/features/sdlc-issue-ownership-lock.md
+++ b/docs/features/sdlc-issue-ownership-lock.md
@@ -72,10 +72,10 @@ Gated this way: `sdlc-tool dispatch record`, `sdlc-tool stage-marker`, and `sdlc
 **Stored payload** (JSON):
 
 ```json
-{"run_id": "a1b2c3...", "session_id": "sdlc-local-1954", "pid": 42317, "hostname": "worker-1", "target_repo": "owner/repo"}
+{"run_id": "a1b2c3...", "session_id": "sdlc-local-1954", "pid": 42317, "machine_id": "A1B2C3D4-...", "hostname": "worker-1", "create_time": 1754500000.0, "target_repo": "owner/repo"}
 ```
 
-Ownership is decided **solely** by comparing the caller's `run_id` against the payload's `run_id` -- a fresh live check on every mutation. `session_id`/`pid`/`hostname` ride along purely for human-readable display (`owner_session_id` in `IssueLockResult` and in the blocked-dispatch JSON shape) -- they are never compared for ownership, because two independent live processes can resolve the identical deterministic `session_id` (`sdlc-local-{issue_number}`) for the same issue. `target_repo` (issue #2012) rides along for a different purpose: it is the single authoritative source the issue-keyed `PipelineLedger`'s writers and readers use to assemble their `(target_repo, issue_number)` key, so no writer or reader needs to shell out to `gh repo view` per call.
+Ownership is decided **solely** by comparing the caller's `run_id` against the payload's `run_id` -- a fresh live check on every mutation. `pid`/`machine_id`/`create_time` feed the owner-liveness predicate `_lock_owner_is_live` (below); `session_id` and `hostname` ride along purely for human-readable display (`owner_session_id` in `IssueLockResult` and in the blocked-dispatch JSON shape) -- they are never compared for ownership, because two independent live processes can resolve the identical deterministic `session_id` (`sdlc-local-{issue_number}`) for the same issue. `target_repo` (issue #2012) rides along for a different purpose: it is the single authoritative source the issue-keyed `PipelineLedger`'s writers and readers use to assemble their `(target_repo, issue_number)` key, so no writer or reader needs to shell out to `gh repo view` per call.
 
 ### The `target_repo` Field (issue #2012)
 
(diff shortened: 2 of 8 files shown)


valorengels:
Review addressed in 319b8d225dd6ff7b3a5c8db48ecfa7837216a903.

**Heartbeat regression fixed as directed.** The peek now carries the guarded run_id (`touch_issue_lock(issue, run_id, peek=True)` — a peek never mutates regardless of run_id), so every exit is an **ownership** check: lease gone, or held by a different run_id. No pid-liveness inference anywhere in the loop. This is a deliberate deviation from the issue's AC3/AC7 as literally written: the payload pid is stamped by the ephemeral `sdlc-tool session-ensure` CLI (`os.getpid()` at fresh acquire) and is dead before the detached heartbeat's first tick, so `orphaned_lock` reads every locally-minted lease as orphaned — gating the renew on it lapses the lease mid-stage, the exact #2446/#2451 failure. Run-liveness is the heartbeat's own existence; its bounded max-lifetime remains the death backstop. AC4 (no lease-theft) unchanged: still peek-first, renew-only, never re-acquires.

**Regression test on the real path**: `test_renews_when_self_owned_payload_pid_is_dead` stubs only Redis and psutil and drives the real `touch_issue_lock` peek — same-machine payload, dead pid, matching run_id → the mutating renew fires. Verified it FAILS against the pre-fix heartbeat (orphaned_lock guard) and passes now. The hand-fed `orphaned_lock` test is deleted.

**Docs de-staled**: `docs/features/sdlc-issue-ownership-lock.md` now documents machine-id-first same-machine comparison with hostname as breadcrumb/fallback plus `config/machine.get_machine_id()` as the stable-identity primitive; `docs/features/sdlc-run-self-recognition.md` documents the ownership-not-pid-liveness exit conditions; `config/machine.py` notes the `functools.cache` memoizing a failed `""` for the process lifetime.

Targeted runs via `scripts/pytest-clean.sh`: lifecycle + heartbeat + machine + session-ensure suites → 193 passed, 0 failed (pre-import-fix run); heartbeat suite re-run post-fix → 6 passed. `ruff check` / `ruff format --check` clean.

Issue #2636, opened 2026-08-07
Bound Job.recent_for_room via a popoto sorted-range pushdown (blocks #2494 phase-2 cutover)


tomcounsell:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/job-recent-for-room-bounded-read.md

Split out of #2634 item 3, because it needs an upstream popoto release and should not hold the other two scaling notes hostage.

**This is a cutover prerequisite.** The #2494 phase-2 authoritative cutover is gated on it, per the M3 block in `docs/plans/durability-room-job-agentrun.md`.

## The defect

`models/job.py:239` `Job.recent_for_room` hydrates every Job in the Room and only then sorts and slices:

```python
jobs = list(cls.query.filter(room_id=room_id, last_active_at__gte=_EPOCH))
jobs.sort(key=lambda j: j.last_active_at or _EPOCH, reverse=True)
return jobs[:limit]
```

## Why the existing popoto `.limit()` does not fix it

Verified against the local `~/src/popoto` checkout at 1.8.1. Two independent reasons:

1. **The SortedField range query is unbounded at the source.** `src/popoto/fields/sorted_field_mixin.py:717` issues `zrangebyscore(key, min, max)` with no `start`/`num`, returning every key in the partition. A commented-out `zrange` carrying `byscore=True, offset=None, num=None` sits directly below at `:720-724` — the bounded form was contemplated and left dormant.
2. **The early-limit optimization is gated to KeyField ordering.** `src/popoto/models/query.py:2687` fires the pre-hydration slice only when `order_by_attr_name in model._meta.key_field_names`. `last_active_at` is a SortedField, so it misses and execution falls through to hydrate-everything-then-slice at `:2560-2564`. The method docstring says so: "limit: Maximum objects to load. Applied early if sorting by KeyField."

So calling `.limit(5)` here would produce the identical behavior and **measure unchanged while looking fixed**.

## Fix: upstream, then consume

`tomcounsell/popoto`, roughly 50-80 lines plus tests:

- `sorted_field_mixin.py` — accept an optional bound and issue `zrange(..., byscore=True, desc=..., offset=0, num=limit)`.
- `query.py` — extend the early-limit gate beyond KeyFields so a SortedField-ordered, SortedField-filtered query threads `limit` down to the range read.

[...]

PR #2646, opened 2026-08-07
Bound the at-rest promise backstop to an index intersection (#2634)

### LINKED ISSUE #2634: Job model scaling follow-ups: pipeline the  scan, bound the at-rest promise scan, bound recent_for_room hydration
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/job-model-scaling-followups.md

Three non-blocking scaling notes from PR #2631's substrate review, all mechanical, none urgent while Job routing runs in shadow:

1. **Pipeline the stale-member scan in `Job.repair_indexes` leg 2** — currently a bare per-member `exists()` round-trip. AgentSession batches the identical scan in 5000s because a bloated index otherwise turns a daily maintenance path into a multi-hour hang. Mechanical port of the existing precedent.
2. **Bound `at_rest_with_open_promises()`** — rest-by-age (72h) bounds the ACTIVE set, but the at-rest population grows forever and this scan JSON-parses every at-rest Job's goal on every health cycle. Add an `at_rest_since`-style bound or cap the scan.
3. **Bound `Job.recent_for_room`** — the top-5 candidate cap is applied after hydrating every Job in the Room (popoto's sorted lookup has no LIMIT). Needs a direct bounded ZREVRANGE against the partition key **before the phase-2 authoritative cutover** — this one is a cutover prerequisite, not just hygiene.

Context: docs/features/durability-model.md; plan docs/plans/durability-room-job-agentrun.md M3 block. Refs #2494.

---

## Recon Summary

Explore fan-out over `models/job.py`, `agent/session_health.py`, `bridge/job_router.py`, installed popoto 1.8.0, and `docs/plans/durability-room-job-agentrun.md`.

**Confirmed:** 2 items — ready to include

- **Item 3 (bound `Job.recent_for_room`)**: `models/job.py:239-251` does `filter(room_id=..., last_active_at__gte=_EPOCH)` — popoto resolves that to `ZRANGEBYSCORE -inf +inf` and then HGETALLs *every* Job in the Room (`popoto/models/query.py:2376,2680`) before the Python-side `[:limit]` at `models/job.py:250-251`. `.limit()` exists in popoto but truncates *after* hydration (`query.py:2516`), so it is not a fix. Direct ZREVRANGE is feasible and has a working in-repo precedent: `agent/memory_retrieval.py:110-122` derives the key via `DecayingSortedField.get_sortedset_db_key(...)`, calls `zrevrange(key, 0, limit-1)`, and decodes bytes. Hydration back through the ORM uses `Model.query.get_many(keys, skip_none=True)` (`query.py:1655-1717`), which pipelines HGETALLs and **preserves input order**.
- **Cutover linkage**: `docs/plans/durability-room-job-agentrun.md:708` names #2634 explicitly as the bounded-ZREVRANGE prerequisite for the phase-2 authoritative cutover (checklist at :643-648, gate at :655). At phase 2 the router moves from backgrounded shadow (`bridge/room_inbox.py:76-121`) into the synchronous dispatch ordering, so its latency becomes intake latency. Jobs are immortal by design (`models/job.py:25-26`, no `Meta.ttl`), so the per-Room candidate set only grows.

- **Item 2 (bound `at_rest_with_open_promises`)**: the pathology is confirmed and worse than stated. `agent/session_health.py:917` (via `_agent_session_health_check` :4505 → `_agent_session_health_loop` :5143) runs on `AGENT_SESSION_HEALTH_CHECK_INTERVAL = 300` (`session_health.py:442`), **per worker process**, and does *two* full Job scans per tick (`sweep_to_rest` over `status=active`, then `at_rest_with_open_promises` over `status=at-rest` with a `json.loads` of every goal). But the suggested `at_rest_since`-style bound is not the drop-in the issue implies:
  - popoto SortedFields **cannot be null** (`popoto/fields/sorted_field_mixin.py:201-206`) and their value lives in a Redis ZSET populated only by `on_save` — so every pre-existing Job would be **absent** from the new ZSET and a `__gte` window would silently return nothing for exactly the backlog it exists to find. That needs a **backfill-write** migration, not the read-probe pattern used for all prior field additions (`scripts/update/migrations.py:288-357`, whose stated rule is "additive nullable fields need no backfill").
  - `last_active_at`'s `partition_by="room_id"` (`models/job.py:86`) makes it unusable for a *global* window: popoto raises `QueryException` if a partition value is missing (`sorted_field_mixin.py:697-715`).
  - An unpartitioned `at_rest_since` reintroduces a global ordered index, against which the M3 schema gate ratified (`durability-room-job-agentrun.md:528,552`). An alternative — a derived low-cardinality `has_open_promises` IndexedField intersected with `status` — would trip the same gate's "status is the ONLY IndexedField" anti-criterion (:736) and risks drift from the JSON truth.
  - Scope is therefore revised to: **choose between a partitioned/unpartitioned `at_rest_since` SortedField (with backfill migration) and a derived open-promise index, resolve the choice by spike, and take it back to the schema gate.** Not a mechanical change.

**Pre-requisites:** 1 item — must fix first (folded into this issue)

- **Naive-datetime score skew must be fixed before `recent_for_room` trusts ZSET scores.** popoto encodes datetimes with `strftime` and decodes with a bare `strptime` — **no tzinfo** (`popoto/models/encoding.py:87-96`). A Job loaded from Redis therefore carries a *naive* `last_active_at`; any subsequent `save()` recomputes the ZSET score as `naive.timestamp()`, which Python reads as **local** time, shifting the stored score by the machine's UTC offset (7h on the dev box). Affected write paths: `mark_at_rest()` (`models/job.py:226`) and `_write_goal_data()` → `append_goal_version` / `add_promise` / `remove_promise` (`models/job.py:152`). Today this is invisible because `recent_for_room` re-sorts in Python on the uniformly-naive decoded value (`models/job.py:250`). A ZREVRANGE trusts the score, so a Job whose last write was a promise-append would sort as if it were hours older than one whose last write was `revive()`. Existing test `tests/unit/test_job_model.py:144` would not catch it (7 consecutive `mint()`s all store aware-derived scores).

- **Item 1 (pipeline the stale-member scan in `Job.repair_indexes` leg 2)**: **already done.** Landed on `main` in commit `0a53ad22f` ("Pipeline Job.repair_indexes' stale-member scan instead of one round trip each") ~46 minutes after this issue was filed. `models/job.py:393-403` now batches `EXISTS` at `batch_size = 5000` through a non-transactional pipeline, byte-for-byte the `models/agent_session.py:2423-2428` precedent the issue asked for. No work remains.

**Incidental finding (not in scope, worth a follow-up):** `.claude/hooks/validators/validate_no_raw_redis_delete.py:83` guards on `\$SortedF:`, a prefix popoto never emits. The metaclass at `popoto/fields/field.py:124` does `f"${name.strip('Field')}F"`, and `"SortedField".strip('Field')` strips the *character set* `{F,i,e,l,d}` → `"Sort"`, so the real prefix is **`$SortF`** (and `$DecayingSortF`). `Job` is also absent from that validator's model list (`Room` is present at :100).

tomcounsell:
Closes #2634.

Scope note: #2634 is the at-rest promise scan only. Item 1 (pipelining the `$IndexF` repair scan) shipped as `0a53ad22f`. Item 3 (bounding `Job.recent_for_room`) moved to #2636, which is blocked on an upstream popoto release.

## The problem

`Job.at_rest_with_open_promises` iterated `filter(status="at-rest")` and called `open_promises()` on every row, which `json.loads`-es that Job's `goal`. It runs from `agent/session_health.py::_check_jobs_at_rest_with_open_promises` every 300 seconds, per worker.

Rest-by-age bounds the *active* set at 72h. Nothing retires an at-rest Job, so the at-rest population grows without limit and the backstop's cost grows with it.

## The change

`Job` gains a second low-cardinality IndexedField, `has_open_promises`: a two-valued bool derived from the `goal` JSON. The backstop intersects it with `status` and hydrates the flagged set rather than the at-rest set.

Measured on a 3,000-Job at-rest population carrying five open promises:

| | `json.loads` | wall clock |
|---|---|---|
| before | 3,000 | 65.0ms |
| after | 5 | 2.5ms |

Both forms return an identical Job set, asserted in the benchmark rather than eyeballed.

## Three things that make it safe

**`type=bool` is load-bearing.** Without the declared type the value round-trips as the string `"False"`, which is truthy, so every Job would look promise-owing. This was flagged as a blocking caveat by the plan's spike-4 and is covered by `test_flag_is_a_real_bool_not_a_truthy_string`.

**The flag is derived at a chokepoint, not by convention.** Every promise mutation already funnels through `_write_goal_data` (`mint`, `add_promise`, `remove_promise`, `append_goal_version`), and external callers in `tools/job_tool.py` and `bridge/job_router.py` all route through those methods. Deriving it there makes it un-bypassable rather than something a future caller has to remember.

**The flag is a projection, never an authority.** `goal` JSON remains the single source of truth, and each candidate is re-verified against `open_promises()`. A drifted flag can therefore cost a hydration but never produce a wrong answer. `test_stale_flag_never_produces_a_wrong_answer` sets the flag `True` on a Job with no open promise and asserts it does not surface.

## Backfill

Rows written before the field existed hold no value for it, and an absent value indexes as nothing rather than as `False`, so they would be silently invisible to the backstop. `rebuild_indexes()` cannot help, because it reconstructs index sets from what the hashes already hold.

`backfill_open_promises_index()` stamps them. It runs on the daily maintenance path via `repair_indexes()`, is idempotent, and writes only where the derived value disagrees with what is stored, so steady-state passes cost reads alone.

## Fail-open contract

Unchanged. This is a backstop query called inside a health cycle and it still returns `[]` rather than raising. `test_backstop_never_raises_into_the_health_cycle` covers it directly.

## Schema gate

Schema Gate Amendment 1 (`bc3f682a5`) authorized the second IndexedField before this was written. The gate's enforced rule is a cardinality rule, never index a pid, uuid, or timestamp, and a two-valued derived bool honors it. Stale assertions that `status` is the only IndexedField are corrected in `models/job.py` and the test module docstring.

Also repoints `docs/plans/durability-room-job-agentrun.md` at #2636 for the bounded ZREVRANGE, since that sentence still described the pre-split issue and named the already-shipped repair-scan item as outstanding.

## Testing

9 new tests in `tests/unit/test_job_model.py::TestOpenPromiseIndex`.

Five mutations, each caught by the intended test:

| mutation | caught by |
|---|---|
| bound removed, scan the whole at-rest set | `test_backstop_hydrates_only_the_flagged_set` + 1 |
| `type=bool` dropped | `test_flag_is_a_real_bool_not_a_truthy_string` + 3 |
| derivation removed from the chokepoint | `test_flag_tracks_promise_lifecycle` + 3 |
| re-verification against `open_promises()` removed | `test_stale_flag_never_produces_a_wrong_answer` |
| backfill neutered | `test_backfill_stamps_a_row_the_flag_missed` |

Runs, from the worktree with `PYTHONPATH` set to it, via `scripts/pytest-clean.sh`:

- `tests/unit/test_job_model.py` — 27 passed
- `tests/unit -k "job or promise or session_health"` — 495 passed, 1 skipped

`python -m ruff check` clean, `python -m ruff format` applied.

Benchmark rows used a `dbg-` prefixed `room_id` and were removed via the ORM scoped to that key.


tomcounsell:
## Review: APPROVED

Verified independently rather than credited.

- `tests/unit/test_job_model.py`: **27 passed**, run from the branch worktree with `PYTHONPATH` set.
- **Mutation-checked the property the whole design rests on.** Replacing the `if job.open_promises()` re-verification with `if True` fails exactly `TestOpenPromiseIndex::test_stale_flag_never_produces_a_wrong_answer` and nothing else. So "a stale flag can only cost a hydration, never a wrong answer" is a tested claim, not a comment.
- Fail-open contract intact: the `except Exception` backstop wrapper is unchanged, which matters because this is called from `agent/session_health.py::_check_jobs_at_rest_with_open_promises` on the health cadence and must never raise into it.

The three properties are the right ones and they are load-bearing in the right order: derived at the `_write_goal_data` chokepoint so it cannot be bypassed, `goal` left authoritative with the flag as a pure projection, and `backfill_open_promises_index()` for legacy rows — that last one is easy to skip and would have made pre-existing Jobs silently invisible, since an absent value indexes as nothing rather than as `False`.

Declaring `IndexedField(type=bool, default=False)` with the explicit `type=bool` is the detail worth calling out. Without it the default hydrates as the truthy string `"False"`, which is exactly the class of defect that ships green.

**On the mutation run that reported five survivors:** catching that as a stripped `PATH` rather than a real result, because five-for-five surviving is not believable, is the right instinct. A mutation harness that cannot fail is worth more scrutiny than one that reports a survivor.

Measured 3,000 at-rest Jobs with five open promises: 3,000 `json.loads` → 5, 65.0ms → 2.5ms, same Job set asserted. That is the bar this issue was set at.

tomcounsell:
Remediation note, with a measurement that has a method attached.

The planning trail carried a claim that the live at-rest Job population is zero, sourced to Schema Gate Amendment 1 (`bc3f682a5`). That commit says "measured" but records no method and no Redis DB, and scopes the claim to an instant. `@do-sdlc` declined to be cited for it because its pipeline read it there rather than measuring it. **That claim was never used in this PR's body**, and it is not the basis for anything here.

Re-measured directly instead, read-only through the ORM:

```
redis: host=localhost port=6379 db=0
Job total:                    0
  active:                     0
  at-rest:                    0
  at-rest with open promise:  0
  has_open_promises unset:    0
at_rest_with_open_promises(): 0
```

Method: `config.redis_bootstrap.configure_resilient_redis()`, then `Job.query.filter()` iterated through the ORM, statuses counted in Python. Taken 2026-08-07 after this PR merged.

**Two things this does not establish.**

It is **this host only**. Job records live in each machine's own Redis, and this is one worker host. It says nothing about the rest of the fleet, and scoping a per-host count as though it were a global property is the same mistake that made the original claim uncitable.

The full-scan-versus-index agreement it printed is **vacuous at a population of zero**. Both sides return an empty list, so it confirms nothing about the derived flag. The real evidence for that property is `test_stale_flag_never_produces_a_wrong_answer` and its mutant, not this census.

**What it does support:** on this host the backfill is currently a no-op, so `backfill_open_promises_index()` running on the daily `repair_indexes` path costs a scan of an empty keyspace. The reasoning that a write-backfill is cheapest while the population is small holds regardless of the exact count, and does not depend on it being exactly zero.

PR #2653, opened 2026-08-07
Scope the Job open-promise backfill write to prevent clobbering a concurrent promise

### LINKED ISSUE #2647: Job open-promise backfill can clobber a concurrent promise write; durability-model doc still says status is the only IndexedField
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/job-open-promise-backfill-clobber.md

Two findings from the `/do-plan-critique` pass on `docs/plans/job-model-scaling-followups.md` survived the merge of PR #2646. The critique verdict was recorded at 06:52:27Z; the PR merged at 06:48:11Z, four minutes earlier, so nothing routed the findings back into the shipped code. #2634 is closed and stays closed — these carry forward here.

## Problem

**1. `Job.backfill_open_promises_index()` can silently lose a promise write.**

The shipped backfill (`models/job.py:449-487`) runs from `Job.repair_indexes()` on the **daily maintenance path** — deliberately while the bridge and workers are live — and re-runs every day rather than once. Its loop is read-derive-write with no re-read before the write:

```python
for job in cls.query.filter():
    derived = any(entry.get("removed_ts") is None for entry in job._goal_data().get("promises", []))
    if job.has_open_promises is not derived:
        job.has_open_promises = derived
        job.save()
```

popoto's `save()` writes the whole hash. An `add_promise` / `append_goal_version` landing between this row's hydration and its `save()` is overwritten by the stale in-memory `goal` — promise data loss, not a lost timestamp.

Two facts bound the exposure without closing it: the write fires only where the stored flag disagrees with the derived value (so a steady-state pass writes nothing), and the loop assigns no `_now()` (so the write is otherwise convergent, including across machines running the daily tick concurrently against shared Redis).

**2. `docs/features/durability-model.md:119` contradicts the shipped schema.**

It still reads "`Job.status` is the only IndexedField (low-cardinality active/at-rest)". `models/job.py:103` declares a second one, `has_open_promises = IndexedField(type=bool, default=False)`. `docs/plans/durability-room-job-agentrun.md:543` (Schema Gate Amendment 1) was amended by the PR; the feature doc was not, and neither it nor the `docs/features/README.md:67` Durability Model row mentions `has_open_promises` at all.

## Recon Summary

**Confirmed:** (verified against `main` @ `b44c4ba7a`)

- `models/job.py:449-487` — `backfill_open_promises_index()` body is as quoted; no re-read, no CAS, no lock beyond `repair_indexes`' own `_repair_lock` (which is per-process, so it does not serialize across machines).
- `models/job.py:442` — `repair_indexes()` calls `cls.backfill_open_promises_index()` after `rebuild_indexes()`, on the daily maintenance path.
- `git grep has_open_promises -- scripts/update/migrations.py` → **no match**. The plan specified a one-shot `MIGRATIONS` entry; the shipped design is the daily maintenance call instead. The plan's Race 1 mitigation ("`/update` sequences migrations before the bridge restart") described neither the plan's design nor the shipped one — `scripts/update/run.py` runs migrations at `:1155` and Step 5 service management at `:1653`, with nothing stopping the bridge in between.
- `docs/features/durability-model.md:119` — stale claim present.
- `git grep -l has_open_promises -- docs/` → only `docs/plans/durability-room-job-agentrun.md` and `docs/plans/job-model-scaling-followups.md`. No feature-doc coverage.
- `models/job.py:103`, `:171`, `:334` — the field, its derivation in `_write_goal_data`, and the narrowed `filter(status="at-rest", has_open_promises=True)` query all shipped and are correct.

**Dropped:** (checked because each would have been a defect; neither is)

- The backfill does **not** round-trip `goal` through the lossy `_goal_data()` parser — it assigns the flag and calls `save()`, so a malformed blob survives verbatim. This was critique BLOCKER 1 and the shipped design avoids it.
- Test coverage is real: `tests/unit/test_job_model.py:152-265` covers flag lifecycle, `type=bool` hydration, the hydration row-count bound, stale-flag correctness, backfill stamping, backfill idempotence, and the fail-open contract.

**Revised:** desired outcome

1. The daily backfill cannot clobber a concurrent promise write. The critique's prescription: re-read each row immediately before writing and skip rows whose `goal` changed since the hydration, turning a lost update into a skipped row that the next daily pass self-heals. Keep the no-`_now()` rule — it is what makes the write convergent across concurrent machines.
2. `docs/features/durability-model.md:119` names both IndexedFields and states the invariant the schema gate actually enforces (no IndexedField holds an unbounded-cardinality value — never index a pid, uuid, or timestamp), per Schema Gate Amendment 1. The `docs/features/README.md:67` row picks up the indexed at-rest promise backstop.

## Refs

- Plan (Complete, with full critique disposition): `docs/plans/job-model-scaling-followups.md`
- Shipped: PR #2646 / `84f07272c`, closes #2634
- Schema gate: `docs/plans/durability-room-job-agentrun.md:543` (Amendment 1, `bc3f682a5`)
- Parent epic: #2494

valorengels:
## Summary

`Job.backfill_open_promises_index()` ran the daily maintenance sweep as a
read-derive-write loop that ended in a bare `job.save()` — a full-hash write
derived from a snapshot hydrated at the top of the loop, on a path that runs
concurrently with live promise mutation (`add_promise` / `remove_promise` /
`append_goal_version`). If a PM's promise write landed on a Job between this
loop's hydration and its `save()`, the backfill's stale `goal` overwrote it,
silently discarding a promise the system had made to a human.

Per spike-1's direct read of the pinned popoto 1.8.0 source: `save()` with no
`update_fields` re-encodes and HSETs every non-indexed field, but
`save(update_fields=[...])` is index-maintaining, and when every listed field
is an `IndexedField` (as `has_open_promises` is), popoto excludes it from the
HSET mapping entirely and takes an EVAL-only path — zero `goal` bytes are ever
transmitted. This converts the fix from a narrowed timing window into a
structural one: the clobber becomes unreachable, not merely unlikely.

## Changes

- `models/job.py::backfill_open_promises_index` — re-fetches each row by both
  KeyFields (`id`, `room_id`) immediately before deriving the flag (so the
  derivation reads fresh `goal`, not a possibly-stale enumeration snapshot),
  and writes only via `save(update_fields=["has_open_promises"])`. The
  re-fetch sits inside the per-row `try` so a raising `query.get` costs one
  row, never the whole daily sweep. Signature is unchanged
  (`(cls) -> int`) — no `room_id` parameter was added. Docstring corrected:
  removed the disproven claim that `rebuild_indexes()` "cannot help" with
  legacy rows (it does, via `default=False` fill-in) and replaced it with the
  real rationale (daily re-derivation/self-healing of any drift), and states
  the write-scope and no-`_now()` convergence invariants.
- `tests/unit/test_job_model.py` — added the monkeypatched-enumeration
  regression test (`test_backfill_does_not_clobber_a_concurrent_promise`,
  the red-state proof below), a write-scope pin via a `save()` spy
  (`test_backfill_write_is_scoped_to_the_flag`), and two fail-open tests
  (per-row re-fetch raise with sibling still stamped; whole-loop enumeration
  raise returns `0`). Updated `test_backfill_stamps_a_row_the_flag_missed` to
  assert `goal` is byte-identical after the backfill.
- `docs/features/README.md` / `docs/features/durability-model.md` — record
  that the at-rest promise backstop is index-served and its daily backfill
  is write-scoped to the derived field.

## Red-state proof

Measured against unmodified `main`'s `backfill_open_promises_index` body,
`test_backfill_does_not_clobber_a_concurrent_promise` fails on exactly the
promise-survival assertion:

```
tests/unit/test_job_model.py:282: in test_backfill_does_not_clobber_a_concurrent_promise
    assert promises_survived == 1, "the concurrent promise must survive the backfill"
E   AssertionError: the concurrent promise must survive the backfill
E   assert 0 == 1
```

Against the fixed loop, the same test passes (`promises_survived == 1`).

## Testing

- [x] `scripts/pytest-clean.sh tests/unit/test_job_model.py -q` — 31 passed
- [x] `python -m ruff check` / `ruff format --check` clean
- [x] All Verification-table checks manually re-run and confirmed (write is
      field-scoped, no bare `save()` in the method, no `_now()`/
      `last_active_at` in executable code, re-fetch inside per-row `try`,
      signature unchanged, no `room_id=` test call sites, docstring
      corrected, no `scripts/update/migrations.py` entry)

## Documentation
- [x] `docs/features/README.md` Durability Model row updated
- [x] `docs/features/durability-model.md` Index safety paragraph extended

## Definition of Done

Closes #2647


valorengels:
## Review: Approved

Reviewed against `docs/plans/job-open-promise-backfill-clobber.md` (issue #2647) on a clean checkout of the PR head. Every Verification-table row was re-run in my own environment, and every number the PR body claims was reproduced rather than taken on trust.

**What the change is.** `Job.backfill_open_promises_index()` stops writing the whole hash. It re-fetches each row by both KeyFields inside the per-row `try`, derives `has_open_promises` from the fresh `goal`, and writes via `save(update_fields=["has_open_promises"])`. Because the sole listed field is an `IndexedField`, popoto filters it out of the HSET mapping entirely and takes the EVAL-only Lua path — no `goal` bytes are transmitted. That is the difference between "narrowed the window" and "removed the outcome", and the plan's spike-1 source read is what earns the stronger claim.

**Independent verification (my environment, not the author's):**

- `scripts/pytest-clean.sh tests/unit/test_job_model.py -q` — **31 passed** (matches the PR body).
- `ruff check` / `ruff format --check` on the two changed source files, and repo-wide — clean.
- Red state reproduced. With `models/job.py` reverted to `main` and the branch's tests in place, `test_backfill_does_not_clobber_a_concurrent_promise` fails on exactly the promise-survival assertion: `AssertionError: the concurrent promise must survive the backfill / assert 0 == 1`. That is the correct failure reason, not a `TypeError` or a signature mismatch, so Success Criterion 4's paper trail holds. The other two new tests also fail on `main` (`stamped == 1` / `assert 0 == 1`), which is expected and is not the red-state artifact.
- The write-scope pin's save-spy (`assert calls == [["has_open_promises"]]`) is the round-4 critique's NIT 1 taken up: it distinguishes a scoped write from a bare one, which no side-effect assertion in the original recipe could. The plan's central invariant now has CI protection rather than a one-shot grep.

**Blast radius.** `backfill_open_promises_index` has exactly one production caller (`Job.repair_indexes()` at `models/job.py:443`, argument-free) and one test module. The signature is unchanged, so no caller can silently change behavior. `Job` declares no `auto_now` field and no `save()` override, so the partial-save path skips nothing this model needs — matching spike-2's reasoning rather than merely repeating it.

**Note for deploy:** this touches `models/`, which the worker and bridge import. It needs a service restart to take effect (`./scripts/valor-service.sh restart` via the normal `/update` path), though the maintenance path only runs on the daily tick so there is no urgency.

## Rubric

- [x] **1. Plan vs. implementation match** — pass — *All 13 Success Criteria verified individually; all 18 Verification rows run and passing, including the mandatory positive control (prints `2`).*
- [x] **2. New code quality** — pass — *Loop body matches the plan's mandated shape verbatim in structure. Both `except Exception` layers, their `logger.warning` bodies, the `stamped` semantics and the summary `logger.info` are untouched. The `except` logs `getattr(job, ...)` not `fresh`, which is correct since `fresh` may be unbound.*
- [x] **3. Test coverage** — pass — *Four new tests (regression, write-scope spy, per-row fail-open, whole-loop fail-open) plus a write-scope assertion added to an existing test. 31 pass on the branch.*
- [x] **4. Regression risk to existing callers** — pass — *`git grep backfill_open_promises_index` → one production call site (unchanged, argument-free) and one test module. `inspect.signature` prints `() -> 'int'`.*
- [x] **5. Data integrity** — pass — *No schema delta, so no `scripts/update/migrations.py` entry is correct (`git diff main --name-only -- scripts/update/migrations.py` is empty). `Job` has no `auto_now` field, so no `modified_at`-style companion belongs in `update_fields`. `last_active_at` is deliberately excluded and its exclusion is asserted by a test. Index maintenance in both directions is exercised by the existing at-rest-membership assertion, which passes.*
- [x] **6. Security** — pass — *No secrets, no `eval`/`exec`/`subprocess`, no raw Redis. The change narrows what a maintenance pass may assert about a row.*
- [x] **7. Documentation accuracy** — pass — *`docs/features/README.md` Durability Model row and `docs/features/durability-model.md`'s Index safety paragraph both updated; the IndexedField enumeration at `:119` correctly left alone (already fixed by `0cee84389`). The docstring's disproven `rebuild_indexes()` claim is replaced, not merely extended, and the docstring names both `_now()` and `last_active_at` as the Verification row requires.*
- [x] **8. PR body accuracy** — pass — *"31 passed", "ruff clean", the pasted red-state failure, "signature unchanged", and "no migrations entry" all reproduced independently.*
- [x] **9. Disclosed deferrals** — n/a — *The PR body discloses no deferrals. The plan's two No-Gos ([SEPARATE-SLUG #2494] cadence change; [SEPARATE-SLUG #2494] repo-wide `save()` audit) are respected — neither is touched.*
- [x] **10. Follow-up claims verified** — n/a — *No follow-up claims made in the PR body.*

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — PASS — *13/13, each re-run.*
- **2. No-Gos from plan — none violated** — PASS — *Cadence untouched; no repo-wide `save()` sweep.*
- **3. New `except Exception` blocks — each has logger/raise/swallow-ok** — N/A — *No new blocks; both pre-existing handlers retained verbatim with their `noqa: BLE001` rationale comments, and both are now covered by tests.*
- **4. New integration tests — exercise serialization boundary (not in-memory only)** — PASS — *Every assertion re-reads from Redis via `Job.query.get(...)`; no assertion trusts an in-memory instance, which is the only shape that can prove this bug.*
- **5. Plan internal consistency — spike findings match task steps** — PASS — *spike-1's EVAL-only claim is what the implementation relies on and what the tests confirm empirically.*
- **7. New public APIs — docstrings present** — N/A — *No new public API; the existing docstring was corrected.*
- **8. Breaking changes — migration path documented** — N/A — *Signature and behavior contract unchanged; strictly narrower write scope.*
- **10. Tests cover the failure path (not just happy path)** — PASS — *Both fail-open layers now have dedicated tests; the per-row test specifically pins the re-fetch's placement inside the inner `try`, which is the placement the plan argued was load-bearing.*
- **11. UI changes (if any) — screenshot captured** — N/A — *No HTML/CSS/JS/TS/template files in the diff.*
- **12. Docs updated for user-facing changes** — PASS — *Both mandated doc edits present.*

### Verified
- [x] Code correctness
- [x] Test coverage
- [x] Plan requirements met

### Blockers
- None

### Tech Debt
- None

### Nits
- None

### Miscellaneous
- None

- None

### Verification Results

All rows pass on the PR branch:

- Job model tests pass — exit 0, 31 passed
- Lint clean / Format clean — exit 0
- Positive control — prints `2`
- Write is field-scoped — `1`
- No bare `save()` in the backfill — `0`
- Convergence: no timestamp in the backfill's code — `0`
- Row re-fetched by both KeyFields — `1`
- Re-fetch is inside the per-row `try` — `True`
- Signature unchanged — `() -> 'int'`
- Production sweep call unchanged — `models/job.py:443` (no argument)
- Test call sites unchanged — `0`
- Regression test forces a stale enumeration — `5` (>= 2)
- Docstring rationale corrected — `False`
- Docstring states the convergence rule — `True`
- No migrations entry added — empty
- README row updated — `1`
- durability-model records write scope — `1`

### Screenshots

### Plan Checkbox Sync
All 13 Success Criteria ticked and pushed as `44318d7` before this review was posted.

PR #2665, opened 2026-08-07
Doctor: flag console scripts that resolve outside the repo venv

### LINKED ISSUE #2566: critique-resume-probe / critique-roster-check console shims crash with ModuleNotFoundError: No module named 'tools' — same wrong-interpreter class as #2536
Reported during the #2536 lane (final report, 2026-08-06). The stale console-script shims for critique-resume-probe and critique-roster-check crash at import with ModuleNotFoundError: No module named 'tools'. Same defect class as #2536's root cause: a console entry point resolving against an interpreter/environment that never had the repo installed, or a stale shim left behind by a removed/renamed [project.scripts] entry. Needs: identify whether the entry points still exist in pyproject.toml, regenerate or delete the shims (pip install -e . refresh), and consider a doctor check for console-script shims whose imports fail (the popoto_floor doctor check from #2546 is the precedent). Related: #2536, #2562 (host tooling hygiene).

tomcounsell:
Closes #2566

## Problem

`[project.scripts]` entry points are invoked by bare name across skills, hooks, and SDLC addenda. Two of them are fail-closed critique gates (`critique-roster-check`, `critique-resume-probe`) and one gates pushes (`sdlc-push-guard`). What those names resolve to is host state that correct packaging cannot guarantee.

On `valorengels` and `MacBookPro.local`, `.venv/bin` was absent from PATH while a stale `~/Library/Python/3.12/bin` sat on it, producing two distinct surfaces at once:

- names with a stale shim resolved, ran under a system interpreter with no editable install of this repo, and died with `ModuleNotFoundError: No module named 'tools'`
- `sdlc-push-guard`, which had no stale shim, failed as `command not found`

Deleting the stale shims only converts the first shape into the second, so the durable fix is detection, not shim hygiene.

## Change

`tools/doctor.py::_check_console_scripts_resolve` reads every `[project.scripts]` name from `pyproject.toml` and `shutil.which`es it against the repo's venv bin directories. Three states, each with its own remedy:

- the name resolves into a repo venv bin directory: healthy, no action needed
- the name resolves somewhere else, or nowhere at all, and is present in the venv: something else is winning the PATH race. The fix is putting the venv first on PATH
- the name is declared in `pyproject.toml` but was never built into the venv: the puller has not re-synced since a `[project.scripts]` entry was added. The fix is `uv sync`, not a PATH reorder

A host can hit both problems at once, a genuinely shadowed name and a separately absent one, and the failure message and fix line report both rather than picking one. The failure message names the offending scripts, where each resolved (or that it did not), and whether `.venv/bin` is off PATH entirely or merely shadowed. When the offending-script list or the missing-entry-point list runs long, both the message and the fix line truncate to the first five and note how many more were omitted.

Two deliberate non-flags, so the check does not manufacture false positives on healthy machines:

- a hardlinked copy outside the venv is accepted, because `/update` installs some entry points that way
- the main checkout's `.venv/bin` is accepted, because doctor frequently runs from a worktree whose lane uses the main venv

The check is registered before `_check_system_tools`, whose `scripts.update.verify` import prepends `~/Library/Python/3.12/bin` to `os.environ["PATH"]`. Measuring first keeps the reported resolution the one a shell would actually get.

## Verification

Live, against the real `pyproject.toml` on this machine (where the issue does **not** reproduce):

```
$ python -c 'from tools.doctor import _check_console_scripts_resolve as c; ...'
True
25 console scripts resolve into /Users/tomcounsell/src/ai/.venv/bin
```

Live, under a PATH that reproduces the broken machines (no `.venv/bin`, a stale shim dir holding two of the three names):

```
passed: False
25/25 console scripts do not resolve into the repo venv — .../.venv/bin is not on PATH;
bare-name callers run the wrong copy or nothing at all (#2566):
critique-resume-probe -> .../fakelib/bin/critique-resume-probe;
critique-roster-check -> .../fakelib/bin/critique-roster-check;
sdlc-push-guard -> not found; ... (+20 more)
```

Both #2566 surfaces are named in one message.

## Tests

`tests/unit/test_doctor_console_scripts.py` — 18 tests, real files and a real PATH, no mocking of `shutil.which`. Covers the `valorengels` shape (shim ahead of an on-PATH venv), the `MacBookPro.local` shape (venv off PATH, mixed shim/not-found), the declared-but-not-installed state and its `uv sync` remedy, a name that is both shimmed and never installed, the hardlink and worktree non-flags, truncation of both the message and the fix line, and registry ordering.

Mutation-tested across three review rounds, each mutation applied and reverted separately, baseline and post-restore both green:

| Mutation | Result |
|---|---|
| `found is None` no longer flagged | caught |
| any resolution accepted (shadowing undetected) | caught |
| hardlink allowance removed (false positive) | caught |
| check unregistered from `get_checks` | caught |
| main-checkout-venv allowance removed | caught |
| `path_note` ternary collapsed to always-"not on PATH" | caught |
| truncation limit 5 -> 50 | caught |
| empty-`[project.scripts]` branch flipped to pass | caught |
| unreadable-pyproject branch flipped to pass | caught |
| not-installed tagging stripped from the resolved-elsewhere branch | caught |
| not-installed tagging stripped from the `found is None` branch | caught |
| aggregate "; N also not installed" suffix removed | caught |
| `_same_file` hardlink clause removed | caught |
| `os.path.realpath(...)` parent-match clause removed | survived — redundant, `_same_file` already covers this case |
| `misresolved[:5]` truncation removed | caught |
| `not_installed` fix-line truncation suffix removed | caught |

`tests/unit/test_doctor_console_scripts.py` + `tests/unit/test_doctor.py`: 65 passed. `ruff format` and `ruff check` clean.

## Out of scope

No stale shim was deleted on any machine — that is host state, and the check now makes it visible with a remediation line. The issue's secondary question (whether callers distinguish exit 127 from a real non-zero verdict) is a separate gate-attribution concern and is not addressed here. Tracked as #2749, so it keeps a home once this PR merges and closes #2566.

### DIFF
--- a/tests/unit/test_doctor_console_scripts.py
+++ b/tests/unit/test_doctor_console_scripts.py
@@ -198,6 +198,64 @@ class TestDeclaredButNotInstalled:
         assert result.fix and "uv sync" not in result.fix
 
 
+class TestShimmedAndNeverInstalled:
+    """The sub-state where both problems land on the same name.
+
+    Every other fixture shims a name that `_fake_checkout` also installs, so
+    "resolves to a stale shim AND was never built into the venv" was never
+    constructed — and the not-installed tagging on that branch could be
+    stripped with the suite still green.
+
+    It is not exotic: pull a branch that adds a `[project.scripts]` entry, skip
+    the sync, and if a stale user-site directory happens to carry that name you
+    are in exactly this state. Telling the operator only to reorder PATH would
+    leave them with a name that still does not exist.
+    """
+
+    def _checkout_with_a_declared_but_unbuilt_name(self, tmp_path):
+        root = _fake_checkout(tmp_path)
+        (root / "pyproject.toml").write_text(
+            (root / "pyproject.toml").read_text() + 'never-installed = "tools.never:main"\n'
+        )
+        return root
+
+    def test_shimmed_absent_name_is_tagged_not_installed(self, tmp_path, monkeypatch):
+        root = self._checkout_with_a_declared_but_unbuilt_name(tmp_path)
+        # The shim carries the one name the venv does NOT have.
+        shims = _stale_shim_dir(tmp_path, ("never-installed",))
+        result = _run(root, [shims, root / ".venv" / "bin"], monkeypatch)
+
+        assert result.passed is False
+        assert "1/4" in result.message
+        assert "never-installed" in result.message
+        assert "not installed in the repo venv" in result.message, (
+            "resolving to a shim does not make the name present in the venv"
+        )
+        assert str(shims) in result.message, "the operator still needs to know what won"
+
+    def test_fix_names_uv_sync_for_a_shimmed_absent_name(self, tmp_path, monkeypatch):
+        root = self._checkout_with_a_declared_but_unbuilt_name(tmp_path)
+        shims = _stale_shim_dir(tmp_path, ("never-installed",))
+        result = _run(root, [shims, root / ".venv" / "bin"], monkeypatch)
+
+        assert result.fix
+        assert "uv sync" in result.fix, "PATH advice alone leaves the name non-existent"
+        assert "never-installed" in result.fix
+
+    def test_mixed_host_aggregate_suffix_counts_the_absent_names(self, tmp_path, monkeypatch):
+        """One genuinely shadowed name plus one shimmed-and-absent name."""
+        root = self._checkout_with_a_declared_but_unbuilt_name(tmp_path)
+        shims = _stale_shim_dir(tmp_path, ("critique-roster-check", "never-installed"))
+        result = _run(root, [shims, root / ".venv" / "bin"], monkeypatch)
+
+        assert "2/4" in result.message
+        assert "shadowed" in result.message
+        assert "1 also not installed in the venv" in result.message, (
+            "the aggregate suffix must count the absent names, not just mention them"
+        )
+        assert result.fix and "export PATH" in result.fix and "uv sync" in result.fix
+
+
 class TestRegisteredInDoctor:
     def test_check_is_registered_before_system_tools(self):
         """`_check_system_tools` imports verify.py, which prepends the stale dir."""
--- a/tests/unit/test_doctor_console_scripts.py
+++ b/tests/unit/test_doctor_console_scripts.py
@@ -255,6 +255,34 @@ class TestShimmedAndNeverInstalled:
         )
         assert result.fix and "export PATH" in result.fix and "uv sync" in result.fix
 
+    def test_fix_signals_omission_when_more_than_five_names_are_not_installed(
+        self, tmp_path, monkeypatch
+    ):
+        """The `fix` string truncates the not-installed list at 5; it must say so.
+
+        `message` already appends `(+N more)` on truncation (see
+        `test_truncates_a_long_failure_list`); the `fix` string's own
+        `uv sync` remediation list truncated silently, so an operator who
+        installed exactly the five named entries and re-ran would still have
+        unnamed broken scripts.
+        """
+        names = tuple(f"valor-tool-{i}" for i in range(9))
+        root = _fake_checkout(tmp_path, names=())
+        (root / "pyproject.toml").write_text(
+            (root / "pyproject.toml").read_text()
+            + "\n".join(f'{n} = "tools.{n.replace("-", "_")}:main"' for n in names)
+            + "\n"
+        )
+        result = _run(root, [root / ".venv" / "bin", Path("/usr/bin")], monkeypatch)
+
+        assert result.passed is False
+        assert result.fix
+        assert "uv sync" in result.fix
+        assert "(+4 more)" in result.fix, (
+            "9 not-installed names, 5 shown -- the remaining 4 must be signaled, "
+            "not silently dropped"
+        )
+
 
 class TestRegisteredInDoctor:
     def test_check_is_registered_before_system_tools(self):
--- a/tests/unit/test_doctor_console_scripts.py
+++ b/tests/unit/test_doctor_console_scripts.py
@@ -137,6 +137,67 @@ class TestConsoleScriptsResolve:
         assert "no [project.scripts]" in result.message
 
 
+class TestDeclaredButNotInstalled:
+    """The third state: venv on PATH, nothing shadowing, name simply absent.
+
+    This is the routine one — a teammate pulls a new `[project.scripts]` entry
+    and has not re-synced — and it used to be reported as "shadowed" with a
+    remediation (reorder PATH) that was already satisfied.
+    """
+
+    def test_absent_name_is_not_called_shadowed(self, tmp_path, monkeypatch):
+        root = _fake_checkout(tmp_path)
+        # Declared in pyproject but never built into .venv/bin.
+        (root / "pyproject.toml").write_text(
+            (root / "pyproject.toml").read_text() + 'never-installed = "tools.never:main"\n'
+        )
+        result = _run(root, [root / ".venv" / "bin", Path("/usr/bin")], monkeypatch)
+
+        assert result.passed is False
+        assert "1/4" in result.message
+        assert "not installed" in result.message
+        assert "shadowed" not in result.message, "PATH is already correct; nothing shadows it"
+        assert "never-installed" in result.message
+
+    def test_fix_points_at_uv_sync_not_at_path(self, tmp_path, monkeypatch):
+        root = _fake_checkout(tmp_path)
+        (root / "pyproject.toml").write_text(
+            (root / "pyproject.toml").read_text() + 'never-installed = "tools.never:main"\n'
+        )
+        result = _run(root, [root / ".venv" / "bin", Path("/usr/bin")], monkeypatch)
+
+        assert result.fix
+        assert "uv sync" in result.fix
+        assert "export PATH" not in result.fix, "advising a PATH reorder that is already done"
+        assert "never-installed" in result.fix
+
+    def test_mixed_shadowed_and_absent_reports_both_remedies(self, tmp_path, monkeypatch):
+        """A host can have both problems; the fix must not name only one."""
+        root = _fake_checkout(tmp_path)
+        (root / "pyproject.toml").write_text(
+            (root / "pyproject.toml").read_text() + 'never-installed = "tools.never:main"\n'
+        )
+        shims = _stale_shim_dir(tmp_path, ("critique-roster-check",))
+        result = _run(root, [shims, root / ".venv" / "bin"], monkeypatch)
+
+        assert result.passed is False
+        assert "2/4" in result.message
+        assert "shadowed" in result.message
+        assert "not installed" in result.message
+        assert result.fix
+        assert "export PATH" in result.fix and "uv sync" in result.fix
+
+    def test_a_shadowed_name_still_reads_as_shadowed(self, tmp_path, monkeypatch):
+        """Control: the not-installed branch must not swallow the real thing."""
+        root = _fake_checkout(tmp_path)
+        shims = _stale_shim_dir(tmp_path, ("critique-roster-check",))
+        result = _run(root, [shims, root / ".venv" / "bin"], monkeypatch)
+
+        assert "shadowed" in result.message
+        assert "not installed" not in result.message
+        assert result.fix and "uv sync" not in result.fix
+
+
 class TestRegisteredInDoctor:
     def test_check_is_registered_before_system_tools(self):
         """`_check_system_tools` imports verify.py, which prepends the stale dir."""
--- a/tests/unit/test_doctor_console_scripts.py
+++ b/tests/unit/test_doctor_console_scripts.py
@@ -1,0 +1,148 @@
+"""Tests for doctor's `[project.scripts]` PATH-resolution check (#2566).
+
+The defect these guard is pure host state: `.venv/bin` missing from PATH while
+a stale `~/Library/Python/3.12/bin` sits on it holding shims for a system
+interpreter with no editable install of this repo. Two of the three SDLC entry
+points resolved to those shims and died with `ModuleNotFoundError`; the third
+had no shim and failed as `command not found`. Both shapes are simulated here
+with real files and a real PATH, because a check that only ever sees a healthy
+machine proves nothing.
+"""
+
+from __future__ import annotations
+
+import os
+import stat
+from pathlib import Path
+from unittest.mock import patch
+
+from tools.doctor import _check_console_scripts_resolve
+
+SCRIPTS = ("critique-roster-check", "critique-resume-probe", "sdlc-push-guard")
+
+
+def _executable(path: Path, body: str = "#!/bin/sh\nexit 0\n") -> Path:
+    path.parent.mkdir(parents=True, exist_ok=True)
+    path.write_text(body)
+    path.chmod(path.stat().st_mode | stat.S_IXUSR | stat.S_IXGRP | stat.S_IXOTH)
+    return path
+
+
+def _fake_checkout(tmp_path: Path, names: tuple[str, ...] = SCRIPTS) -> Path:
+    """A checkout with a `[project.scripts]` table and a populated `.venv/bin`."""
+    root = tmp_path / "repo"
+    (root / ".git").mkdir(parents=True)
+    table = "\n".join(f'{n} = "tools.{n.replace("-", "_")}:main"' for n in names)
+    (root / "pyproject.toml").write_text(f'[project]\nname = "x"\n\n[project.scripts]\n{table}\n')
+    for name in names:
+        _executable(root / ".venv" / "bin" / name)
+    return root
+
+
+def _stale_shim_dir(tmp_path: Path, names: tuple[str, ...]) -> Path:
+    """Stand-in for `~/Library/Python/3.12/bin` — shims for a foreign interpreter."""
+    shim_dir = tmp_path / "Library" / "Python" / "3.12" / "bin"
+    for name in names:
+        _executable(shim_dir / name, "#!/usr/bin/python3\nraise SystemExit(1)\n")
+    return shim_dir
+
+
+def _run(root: Path, path_entries: list[Path], monkeypatch):
+    monkeypatch.setenv("PATH", os.pathsep.join(str(p) for p in path_entries))
+    with patch("tools.doctor.PROJECT_DIR", root):
+        return _check_console_scripts_resolve()
+
+
+class TestConsoleScriptsResolve:
+    def test_passes_when_venv_bin_leads_path(self, tmp_path, monkeypatch):
+        root = _fake_checkout(tmp_path)
+        result = _run(root, [root / ".venv" / "bin", Path("/usr/bin")], monkeypatch)
+        assert result.passed is True, result.message
+        assert "3 console scripts" in result.message
+        assert str(root / ".venv" / "bin") in result.message
+
+    def test_stale_shim_ahead_of_venv_is_caught(self, tmp_path, monkeypatch):
+        """The `valorengels` shape: venv on PATH, but shadowed for two names."""
+        root = _fake_checkout(tmp_path)
+        shims = _stale_shim_dir(tmp_path, ("critique-roster-check", "critique-resume-probe"))
+        result = _run(root, [shims, root / ".venv" / "bin"], monkeypatch)
+        assert result.passed is False
+        assert "2/3" in result.message
+        assert "critique-roster-check" in result.message
+        assert str(shims) in result.message
+        assert "shadowed" in result.message
+        assert result.fix and str(root / ".venv" / "bin") in result.fix
+
+    def test_venv_absent_from_path_reports_both_failure_shapes(self, tmp_path, monkeypatch):
+        """The `MacBookPro.local` shape: shims for two names, nothing for the third."""
+        root = _fake_checkout(tmp_path)
+        shims = _stale_shim_dir(tmp_path, ("critique-roster-check", "critique-resume-probe"))
+        result = _run(root, [shims, tmp_path / "empty"], monkeypatch)
+        assert result.passed is False
+        assert "3/3" in result.message
+        assert "not on PATH" in result.message
+        # ModuleNotFoundError shape and command-not-found shape, both named.
+        assert f"critique-roster-check -> {shims / 'critique-roster-check'}" in result.message
+        assert "sdlc-push-guard -> not found" in result.message
+
+    def test_nothing_on_path_at_all_fails(self, tmp_path, monkeypatch):
+        root = _fake_checkout(tmp_path)
+        result = _run(root, [tmp_path / "empty"], monkeypatch)
+        assert result.passed is False
+        assert "not found" in result.message
+
+    def test_hardlinked_copy_outside_the_venv_is_accepted(self, tmp_path, monkeypatch):
+        """`/update` hardlinks entry points into ~/.local/bin — not a wrong copy."""
+        root = _fake_checkout(tmp_path, names=("sdlc-push-guard",))
+        local_bin = tmp_path / "local" / "bin"
+        local_bin.mkdir(parents=True)
+        os.link(root / ".venv" / "bin" / "sdlc-push-guard", local_bin / "sdlc-push-guard")
+        result = _run(root, [local_bin, root / ".venv" / "bin"], monkeypatch)
+        assert result.passed is True, result.message
+
+    def test_main_checkout_venv_accepted_from_a_worktree(self, tmp_path, monkeypatch):
+        """Doctor often runs from a worktree whose lane uses the main venv."""
+        root = _fake_checkout(tmp_path)
+        worktree = root / ".worktrees" / "lane-a"
+        worktree.mkdir(parents=True)
+        (worktree / ".git").write_text(f"gitdir: {root / '.git' / 'worktrees' / 'lane-a'}\n")
+        (worktree / "pyproject.toml").write_text((root / "pyproject.toml").read_text())
+        result = _run(worktree, [root / ".venv" / "bin"], monkeypatch)
+        assert result.passed is True, result.message
+        # The pass message names where they actually resolved (the main venv),
+        # not the worktree's own bin that nothing on PATH points at.
+        assert str(root / ".venv" / "bin") in result.message
+        assert str(worktree / ".venv" / "bin") not in result.message
+
+    def test_truncates_a_long_failure_list(self, tmp_path, monkeypatch):
+        names = tuple(f"valor-tool-{i}" for i in range(9))
+        root = _fake_checkout(tmp_path, names=names)
+        result = _run(root, [tmp_path / "empty"], monkeypatch)
+        assert result.passed is False
+        assert "+4 more" in result.message
+
+    def test_unreadable_pyproject_fails(self, tmp_path, monkeypatch):
+        root = tmp_path / "repo"
+        root.mkdir()
+        result = _run(root, [tmp_path / "empty"], monkeypatch)
+        assert result.passed is False
+        assert "could not read" in result.message
+
+    def test_empty_scripts_table_fails(self, tmp_path, monkeypatch):
+        root = tmp_path / "repo"
+        root.mkdir()
+        (root / "pyproject.toml").write_text('[project]\nname = "x"\n')
+        result = _run(root, [tmp_path / "empty"], monkeypatch)
+        assert result.passed is False
+        assert "no [project.scripts]" in result.message
+
+
+class TestRegisteredInDoctor:
+    def test_check_is_registered_before_system_tools(self):
+        """`_check_system_tools` imports verify.py, which prepends the stale dir."""
+        from tools.doctor import _check_console_scripts_resolve as check
+        from tools.doctor import _check_system_tools, get_checks
+
+        checks = get_checks()
+        assert check in checks
+        assert checks.index(check) < checks.index(_check_system_tools)
--- a/tools/doctor.py
+++ b/tools/doctor.py
@@ -156,19 +156,25 @@ def _check_console_scripts_resolve() -> CheckResult:
     name resolves to is therefore load-bearing, and it is pure host state that
     no amount of correct packaging can guarantee.
 
-    Two live failure modes, both observed on other machines in #2566:
+    Three failure modes, with three different remedies:
 
     * `.venv/bin` is absent from PATH while a stale `~/Library/Python/3.12/bin`
       is present. The stale directory holds shims for a *system* interpreter
       with no editable install of this repo, so the name resolves, runs, and
-      dies with `ModuleNotFoundError: No module named 'tools'`.
+      dies with `ModuleNotFoundError: No module named 'tools'`. Observed on
+      another machine in #2566.
     * The same PATH with no stale shim for a given name, which surfaces as
-      `command not found` (exit 127) instead.
+      `command not found` (exit 127) instead. Also #2566.
+    * The venv leads PATH and nothing shadows it, but the name was never
+      installed into it — someone added a `[project.scripts]` entry and the
+      puller has not re-synced. The remedy is `uv sync`, and reporting this as
+      "shadowed" sends an operator to reorder a PATH that is already correct.
 
     Deleting stale shims only converts the first shape into the second, so the
     check measures resolution rather than shim hygiene: a name is healthy when
     it resolves into a repo venv bin directory, and unhealthy when it resolves
-    anywhere else or nowhere at all.
+    anywhere else or nowhere at all. Distinguishing the third shape from the
+    first two is what `_installed_in_venv` is for.
 
     Ordering note: this runs before `_check_system_tools`, whose
     `scripts.update.verify` import prepends `~/Library/Python/3.12/bin` to
@@ -207,12 +213,26 @@ def _check_console_scripts_resolve() -> CheckResult:
     }
     on_path = [d for d in venv_bins if os.path.realpath(d) in path_entries]
 
+    def _installed_in_venv(script_name: str) -> bool:
+        """Does a repo venv bin dir actually carry this entry point?
+
+        This is what separates "something else is winning the PATH race" from
+        "nothing was ever built to win it". Both produce an unresolvable bare
+        name, and they have opposite remedies.
+        """
+        return any((bin_dir / script_name).exists() for bin_dir in venv_bins)
+
     misresolved: list[str] = []
+    not_installed: list[str] = []
     resolved_into: list[str] = []
     for name in sorted(scripts):
         found = shutil.which(name)
         if found is None:
-            misresolved.append(f"{name} -> not found")
+            if _installed_in_venv(name):
+                misresolved.append(f"{name} -> not found")
+            else:
+                not_installed.append(name)
+                misresolved.append(f"{name} -> not installed in the repo venv")
             continue
         found_path = Path(found)
         match = next(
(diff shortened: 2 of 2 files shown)


tomcounsell:
## Review: CHANGES REQUESTED

A well-built check. It measures the right thing (resolution, not shim hygiene), the reasoning for that choice is correct, and the tests are unusually strong — real files, a real PATH, no `shutil.which` mocking. One tech-debt item stands between this and approval, and it is in the part of the check this PR advertises as a deliverable: the diagnosis attached to the failure.

### Blockers

- None

### Tech Debt

- [ ] **`tools/doctor.py:141-145` and `155-159`** — the `path_note` ternary and the `fix` string — **"declared but not installed" is misdiagnosed as "shadowed", with a remediation that is already satisfied.**

  `path_note` has only two states: if no repo venv bin is on PATH it says "not on PATH", otherwise it says "is on PATH but is shadowed". There is a third real state — the venv bin is on PATH, nothing shadows it, and the name is simply not installed in it — and it lands in the "shadowed" branch. The `fix` line is unconditional, so it tells the operator to put the venv first on PATH when it already is first.

  Reproduced with a real checkout and a real PATH, venv bin first, one declared name deliberately not installed:

  ```
  message: 1/3 console scripts do not resolve into the repo venv —
           .../fakeproj/.venv/bin is on PATH but is shadowed; ...: never-installed -> not found
  fix    : Put the repo venv first on PATH: export PATH=".../fakeproj/.venv/bin:$PATH" ...
  ```

  Nothing is shadowing anything, and the PATH advice is a no-op. The correct remediation is `uv sync` (or `uv pip install -e .`) to install the missing entry point.

  This matters more than it looks, because it is the case most likely to fire from here on. The two shapes in #2566 are host rot on two specific machines; "someone added a `[project.scripts]` entry, I pulled, I have not re-synced" is the routine one, and it is the one the check will misexplain. It is also the one case with no test: the suite covers venv-leads-PATH, shim-ahead-of-venv, venv-off-PATH, nothing-on-PATH, hardlink, worktree, truncation, unreadable pyproject, and empty table — but not on-PATH-and-absent.

  **Fix:** branch on whether `(bin_dir / name).exists()` for the primary venv. When the name is absent from a venv that is on PATH, say so and point at `uv sync`; reserve "shadowed" for the case where the name exists in the venv and `which` found an earlier copy.

  Scope note on reachability: `_check_console_scripts_resolve` is in the `--quick` set (position 2 of 24), which doctor's `--install-hook` pre-push hook runs under `set -e` and blocks the push on. That path is inert on this machine — `core.hooksPath` points at `.githooks`, whose `pre-push` does not invoke doctor — so today this is a wrong message on a human-invoked command, not a blocked push. Worth knowing before the hook path is ever revived.

### Nits

- None

### Verified correct

- **Both #2566 surfaces are detected, in one message.** Reproduced the `valorengels` shape (stale shim dir ahead of an on-PATH venv) and the `MacBookPro.local` shape (venv off PATH, some names shimmed and some absent) against real directories: the first reports `shadowed` with the shim path named, the second reports `not on PATH` with the `-> not found` and `-> <shim path>` entries side by side.
- **No false positive in either real-world configuration.** Pointing `PROJECT_DIR` at the live main checkout: `passed=True`, "25 console scripts resolve into /Users/tomcounsell/src/ai/.venv/bin", which reproduces the PR's claimed measurement exactly. Pointing it at a live worktree (`.worktrees/sdlc-2140`): `passed=True`, resolving through the main checkout's venv, which is the stated non-flag working as designed. `_repo_venv_bin_dirs()` correctly returned both the lane's own bin and the main checkout's.
- **The ordering rationale is real, not assumed.** `scripts/update/verify.py:28-31` prepends `~/Library/Python/3.12/bin` (among others) to `os.environ["PATH"]` at import time, so registering this check ahead of `_check_system_tools` genuinely preserves the resolution a shell would get. `get_checks()` places it at index 2, before `_check_system_tools`, in both full and `--quick` modes.
- **Mutation-checked, wider than the PR claims: 9 valid mutations, 9 caught.** Beyond the four in the description I also mutated the main-checkout-venv allowance (caught, 1 failed), the `path_note` ternary collapsed to always-"not on PATH" (caught, 1 failed), the truncation limit 5 → 50 (caught, 1 failed), the empty-`[project.scripts]` branch flipped to pass (caught, 1 failed), and the unreadable-pyproject branch flipped to pass (caught, 1 failed). Baseline 10 passed, restored clean afterward. One further mutation I attempted was a no-op on my part and is not counted.
- **Tests and lint reproduce.** `tests/unit/test_doctor_console_scripts.py` + `tests/unit/test_doctor.py`: 57 passed, matching the claimed count exactly. `ruff check` and `ruff format --check` clean on both changed files.
- **Function-local `import shutil` / `import tomllib` is house style here**, not an inconsistency — `tools/doctor.py` already has 9 function-local imports including `platform`, `subprocess`, and `socket`. Checked before raising it, and it is not a finding.
- **The hardlink and worktree allowances do not open a hole.** `_same_file` compares by inode via `samefile`, so a *copy* outside the venv is still flagged; only a genuine hardlink to the venv's own entry point is accepted. `test_hardlinked_copy_outside_the_venv_is_accepted` exercises that with a real `os.link` rather than a mock.

### Screenshots

Not applicable. No UI files in the diff.

tomcounsell:
Patched at `915ed93876dc1d5b49462db4f9b4eb56ad2263a5`. The tech debt finding is fixed; thanks for the wider mutation set and for reproducing the third state live — it was a real hole and the one most likely to fire.

## What changed

`_installed_in_venv()` asks whether a repo venv bin dir actually carries the entry point. That is what separates "something else is winning the PATH race" from "nothing was ever built to win it" — the two produce the same unresolvable bare name and have opposite remedies.

The message and the fix now branch on it:

| State | Message | Fix |
|---|---|---|
| Venv not on PATH | `<venv> is not on PATH` | `export PATH=...` |
| On PATH, name exists, earlier copy wins | `<venv> is on PATH but is shadowed` | `export PATH=...` |
| On PATH, nothing shadowing, name absent | `N declared but not installed in <venv>` | `uv sync` |
| Both problems on one host | shadowed `; N also not installed in the venv` | both, in one string |

Per-name detail also reads `name -> not installed in the repo venv` rather than the bare `-> not found`, which was accurate but unhelpful.

## Your repro, before and after

Same shape you used — real checkout, venv first on PATH, one declared name deliberately absent:

```
before  message: 1/3 ... — .../fakeproj/.venv/bin is on PATH but is shadowed; ...
        fix    : Put the repo venv first on PATH: export PATH=".../fakeproj/.venv/bin:$PATH" ...

after   message: 1/3 ... — 1 declared but not installed in .../fakeproj/.venv/bin; ...
        fix    : Install the missing entry point(s) with `uv sync` in .../fakeproj
                 (or `uv pip install -e .`): never-installed. Then re-run.
```

No PATH advice at all in the third state now. I also ran the other two states against the same fixture to confirm they still report as before, and the mixed host reports both remedies.

## Tests

4 new, in `TestDeclaredButNotInstalled`: the absent-name state is not called shadowed, the fix points at `uv sync` and not at PATH, a mixed host names both remedies, and — the control — a genuinely shadowed name still reads as shadowed and does not get swallowed by the new branch. 61 pass (was 57).

**Mutation-checked, 7 mutations, 7 caught:** installed-probe forced True (the old behaviour, 3 failed) and forced False (5), the branch selector forced True (2) and False (4), each remedy arm dropped (2 each), and the not-found path made to ignore install state (3). Baseline restored green.

Doctor still passes in situ on this machine: `25 console scripts resolve into /Users/tomcounsell/src/ai/.venv/bin`.

## On your scope note

Agreed and worth keeping visible: this check sits at position 2 of the `--quick` set, which `--install-hook`'s pre-push runs under `set -e`. That path is inert here because `core.hooksPath` points at `.githooks`, whose `pre-push` does not call doctor. So today a wrong message costs a human a confused minute; if that hook path is ever revived it would have blocked a push with advice that could not help. That is most of why this was worth fixing rather than deferring.

tomcounsell:
## Review: CHANGES REQUESTED

Delta review at `915ed9387` by the reviewer who filed the tech-debt item at `be9ce553`. **The finding is fully closed** — the three-way split is correct in every state I could construct, and the control genuinely controls. Two mutations survived that the PR's own count did not cover; one is cosmetic and one leaves the fixed behavior unpinned. Read the disposition note at the bottom before sequencing the merge.

### Blockers

- None.

### Prior finding: closed

`_installed_in_venv()` is the right discriminator, and the remedies branch correctly. Verified against real directories and a real PATH in five states, not from the diff:

| state | `path_note` | remedy |
|---|---|---|
| venv leads PATH, name absent (my original scenario) | `2 declared but not installed in <venv>/bin` | `uv sync` only, no PATH advice |
| pure shadowing, every name installed (**control**) | `<venv>/bin is on PATH but is shadowed` | PATH advice only, no `uv sync` |
| mixed host | `…is shadowed; 2 also not installed in the venv` | both, and both named in `fix` |
| venv off PATH *and* some names absent | `…is not on PATH; 2 also not installed` | both |
| venv empty *and* off PATH (never-synced clone) | `2 declared but not installed` | `uv sync` only |

The last row is the one I went looking for as a possible hole: with nothing installed, the PATH problem is real but unmentioned. It resolves correctly in sequence rather than misleading — the message ends "Then re-run.", and re-running after a notional `uv sync` flips it to `<venv>/bin is not on PATH` with the PATH remedy. I confirmed that two-step transition directly. Not a finding.

The control test at line 190 is real and non-vacuous: `_fake_checkout` installs every declared name into `.venv/bin`, so the shadowed names genuinely are installed, and the test asserts `"not installed" not in result.message` and `"uv sync" not in result.fix`. Mutating the probe to always return False — the exact "new branch swallows the real thing" failure — is caught with 5 failures.

### Tech Debt

- [ ] **`tools/doctor.py`, the resolved-elsewhere arm of the not-installed tagging** — the sub-state "name resolves to a stale shim **and** was never installed in the venv" has no test. Removing both the `not_installed.append(name)` and the `(not installed in the repo venv)` annotation from that branch leaves all 14 tests green.

  It is uncovered because of how the fixtures compose: `_fake_checkout` installs every declared name into `.venv/bin`, and all four `_stale_shim_dir` call sites shim names drawn from that same set, so a shimmed-but-absent name is never constructed. The mixed-host test's `never-installed` is absent from the venv but is not shimmed, so it exercises the `found is None` arm instead.

  The shipped behavior is correct — I confirmed a shimmed-and-absent name is tagged not-installed and drives the `uv sync` remedy. But a regression here would silently reclassify it as a pure PATH problem and hand back only the PATH advice, which is the exact wrong-remedy behavior this patch exists to remove, just narrowed to the shimmed subcase.

  **Fix:** one test with a name declared in `pyproject.toml`, present in the shim dir, and absent from `.venv/bin`; assert the message carries `(not installed in the repo venv)` and the fix carries `uv sync`.

### Nits

- [ ] **`tools/doctor.py`, the mixed-host `path_note` suffix** — dropping `; {n} also not installed in the venv` leaves all 14 tests green. `test_mixed_shadowed_and_absent_reports_both_remedies` asserts `"not installed" in result.message`, but that substring is already satisfied by the per-name entry in the truncated list, so the aggregate clause is unpinned. Cosmetic: both remedies still reach `fix`, which is separately pinned.

### Verified correct

- **Mutation-checked, wider than the PR's claimed seven: 10 mutations, 8 caught.** Caught: probe forced True (3 failed), probe forced False (5 failed), selector forced to path-problem (2), selector forced to not-installed (4), PATH-advice arm removed (2), `uv sync` arm removed (2), and the not-found branch reverted to the old undifferentiated message (3). The two survivors are the items above. Baseline 14 green, restored green, working tree clean afterward.
- **61 tests pass across the two doctor files, exactly the claimed count** (was 57, +4 new). `ruff check` and `ruff format --check` clean on both changed files.
- **The four new tests are the right four** — the previously uncovered state, the fix pointing at `uv sync` rather than PATH, the mixed host, and the control.
- **Nothing from the first review regressed.** The check still sits at index 2 in both full and `--quick` mode, ahead of `_check_system_tools`; the hardlink and worktree allowances are untouched; the `fix` string can never be empty, since `a_path_problem` false implies `not_installed` is non-empty inside the `if misresolved:` block.

### Disposition

The finding I filed is closed and I verified it five ways, including the control the coordinator asked me to check specifically. The verdict is CHANGES REQUESTED only because approval is reserved for zero-finding reviews, and the outstanding work is a single test on behavior that is already correct. I would not argue against merging this now and adding that test as a follow-up — the substance is sound, and this is the same standard I applied to #2681's re-review rather than a new bar raised for this PR.

### Screenshots

Not applicable. No UI files in the diff.

tomcounsell:
Patched at `ca066dbbba98cb3c58ceec1a8e4c175d0432ccff`. The sub-state finding is closed, and the cosmetic nit came along with it as a one-liner.

## The gap was real

Confirmed your diagnosis before writing anything: every `_stale_shim_dir` call shims a name that `_fake_checkout` also installs, so "resolves to a stale shim **and** was never built into the venv" was never constructed. The not-installed tagging on the resolved-elsewhere branch could be stripped with the suite green.

Worth saying why it matters beyond coverage arithmetic: it is not an exotic host. Pull a branch that adds a `[project.scripts]` entry, skip the sync, and if a stale user-site directory happens to carry that name you are in exactly this state. PATH advice alone would leave the operator with a name that still does not exist after they follow it — the same failure the original finding was about, one layer in.

## Three tests

- `test_shimmed_absent_name_is_tagged_not_installed` — the message tags it not-installed **and** still names the shim that won, since the operator needs both facts.
- `test_fix_names_uv_sync_for_a_shimmed_absent_name` — the remedy is `uv sync`, not a PATH reorder.
- `test_mixed_host_aggregate_suffix_counts_the_absent_names` — one genuinely shadowed name plus one shimmed-and-absent one. Both remedies present, and the aggregate suffix reads `1 also not installed in the venv`. That pins your cosmetic nit too.

## Mutation results

Your exact mutation, plus the nit and a control:

| Mutation | Before | Now |
|---|---|---|
| shim branch never tags not-installed | 14 passed | 3 failed |
| aggregate suffix dropped | (nit, unpinned) | 1 failed |
| not-found branch tagging removed (control) | — | 3 failed |

17 tests in the file, 64 with `test_doctor.py`. `ruff` clean.

Thanks for checking that my swallow-control was non-vacuous rather than taking the assertion at face value — that is the check I most want someone else running on my tests, since a control that cannot fail is worse than no control.

tomcounsell:
## Review: CHANGES REQUESTED — Tech Debt

Third pass, at `ca066dbb`. **Both prior findings are closed and I confirmed each with the reviewer's own mutation**, so the code is where it should be. What is left is not in `tools/doctor.py` at all: the PR description has drifted two commits behind the behavior it describes, and the tracking issue it claims to close was closed as `NOT_PLANNED` on 2026-08-11 while this branch sat open. Neither is a code defect; both need a hand before merge.

Everything below was measured on this machine at `ca066dbb`, not read off the diff.

### Rubric

- [x] **1. Plan vs. implementation match** — pass — *No plan doc exists (`docs/plans/sdlc-2566.md` absent — hand-authored fix straight off the issue), so I judged against #2566 itself. The issue asked for "a doctor check for console-script shims whose imports fail"; the PR measures resolution instead and argues in the docstring why that is the durable measure. Strictly wider than the ask, and correct: deleting a shim converts a `ModuleNotFoundError` into a `command not found`, and both shapes now land in one message.*
- [x] **2. New code quality** — pass — *Three new functions, all with docstrings that carry the reasoning, not just the signature. The single `except Exception` in `_repo_venv_bin_dirs` is a deliberate swallow with a stated rationale and a `noqa` tag.*
- [x] **3. Test coverage** — pass — *17 tests in the new file, mutation-verified below.*
- [x] **4. Regression risk to existing callers** — pass — *The change is additive: one new check appended to `get_checks`, no existing check touched. Registration order ahead of `_check_system_tools` is itself pinned by a test, which matters because that import mutates `os.environ["PATH"]`.*
- [x] **5. Data integrity** — n/a — *No models, no migrations, no persisted state.*
- [x] **6. Security** — pass — *No `eval`/`exec`/`subprocess`, no secrets. `shutil.which` over `os.environ["PATH"]` and `Path.exists()` reads only.*
- [x] **7. Documentation accuracy** — pass — *`docs/features/local-doctor.md` enumerates check **categories**, not individual checks, and this one falls under the existing `Environment` row. Same treatment `_check_popoto_floor` (#2546), `_check_git_hooks_installed` (#2540), and `_check_worktree_interpreters` already get. No stale reference introduced.*
- [ ] **8. PR body accuracy** — fail — *See Tech Debt below. Test counts and the change description both describe `be9ce553`, two commits back.*
- [x] **9. Disclosed deferrals** — pass — *Both are explicit out-of-scope exclusions with sound rationale; see the Acknowledged section.*
- [x] **10. Follow-up claims verified** — n/a — *The body claims no `filed as follow-up #N` anywhere.*

### Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — N/A — *No plan document for this slug.*
- **3. New `except Exception` blocks — each has logger/raise/swallow-ok** — PASS — *`_repo_venv_bin_dirs`, `# noqa: S110 -- the current checkout's venv is enough on its own`. The swallow is correct: losing the main-checkout candidate degrades to checking only the current checkout.*
- **4. New integration tests — exercise serialization boundary (not in-memory only)** — N/A — *No serialization boundary. The tests do use real files and a real `PATH` rather than mocking `shutil.which`, which is the analogous discipline here.*
- **6. No hardcoded secrets or debug artifacts** — PASS — *No `print`, no `TODO`, no commented-out code in the diff.*
- **7. New public APIs — docstrings present** — PASS — *All three new module-level functions plus the nested `_installed_in_venv`.*
- **8. Breaking changes — migration path documented** — N/A — *Additive check only.*
- **9. Tests added for new behavior** — PASS — *17, covering all three states plus the two non-flags.*
- **10. Tests cover the failure path (not just happy path)** — PASS — *15 of the 17 are failure-path or non-flag tests; only two assert a clean pass, and one of those is the worktree allowance.*
- **11. UI changes (if any) — screenshot captured** — N/A — *`gh pr diff --name-only` returns two Python files. No HTML, CSS, JS/TS, JSX/TSX, Vue, or template files.*
- **12. Docs updated for user-facing changes** — PASS — *See Rubric item 7.*

### Blockers

- None

### Tech Debt

- **File:** PR description, `## Tests` and `## Change` sections
  **Code:** `` `tests/unit/test_doctor_console_scripts.py` — 10 tests `` / `` 57 passed `` / *"A name is healthy when it lands in a repo venv bin directory; unhealthy when it resolves anywhere else or nowhere at all."*
  **Issue:** The body still describes `be9ce553`, the first of three commits. Measured at `ca066dbb`: the file has **17** tests, and `test_doctor_console_scripts.py` + `test_doctor.py` is **64 passed**, not 57. The mutation table lists 4 rows; the branch has since been mutation-checked far wider (the `ca066dbb` commit message alone claims three more). More substantively, `## Change` describes a **two-state** check, while `915ed93` shipped a **three-state** split — "declared but not installed" is now distinguished from "shadowed" and carries a separate `uv sync` remedy instead of PATH advice. That third state is the one a teammate actually hits after pulling a new `[project.scripts]` entry without re-syncing, and it is invisible in the description. This repo squash-merges, so the stale body becomes the permanent commit message for a check whose shipped behavior it misstates.
  **Severity:** tech_debt
  **Fix:** Refresh `## Change` to describe the three states and the two remedies, and update `## Tests` to 17 tests / 64 passed with the current mutation table. No code change, no re-test.

  *(Verified separately: the pasted red-state block under `## Verification` is **not** stale. I traced that scenario through the current code — every name installed in the venv, `.venv/bin` off PATH — and it still produces `25/25 … is not on PATH` with no `also not installed` suffix, exactly as pasted.)*

### Nits

- **File:** `tools/doctor.py:289`
  **Code:** ``f"(or `uv pip install -e .`): {', '.join(sorted(not_installed)[:5])}."``
  **Issue:** The `fix` string truncates the missing-name list at 5 **silently**, while the `message` truncation two blocks up appends `(+N more)`. Constructed live with 9 declared-and-unbuilt names: the message ends `valor-tool-4 … (+4 more)` and the fix ends `valor-tool-4.` — an operator who installs exactly the five named entries and re-runs gets the same failure back with no hint that four more were elided. Never-synced clones and large `[project.scripts]` tables (this repo declares 25) are precisely where this lands.
  **Severity:** nit
  **Fix:** Mirror the message's suffix — append `(+{len(not_installed) - 5} more)` when the list is truncated, or drop the `[:5]` in the fix string, since `uv sync` repairs all of them in one go regardless of how many are named.

### Miscellaneous

- **File:** PR description, line 1 — `Closes #2566`
  **Code:** `Closes #2566`
  **Issue:** #2566 is **CLOSED** with `stateReason: NOT_PLANNED`, closed 2026-08-11 by `valorengels` with "Cannot reproduce — closing as obsolete." That closure was verified only on this machine (`.venv/bin/critique-roster-check --help` plus seven worktree venvs), which is the same single-machine check that produced the **first** close on 2026-08-06 — and that one was reopened a day later when `valorengels` posted a real traceback and `MacBookPro.local` re-confirmed it with `.venv/bin` absent from PATH entirely. The issue's own history refutes the method the second closure used. Merging as-is is not harmful (`Closes` on an already-closed issue is a no-op), but it leaves a merged fix pointing at a not-planned issue, and it quietly drops the issue's secondary question about exit-127 attribution, which this PR explicitly deferred to #2566's thread.
  **Severity:** tech_debt
  **Fix:** Reopen #2566 before merge so `Closes` resolves honestly and the deferred sub-question stays tracked — the check's whole premise is that host state differs per machine, so a clean probe on one host is not evidence of absence. If the human decides the issue really is obsolete, change the trailer to `Refs #2566` and say in the body that the check ships as durable detection rather than a fix for a live defect.

- **"No stale shim was deleted on any machine — that is host state, and the check now makes it visible with a remediation line."** — explicit out-of-scope exclusion, no tracking required. Sound: the PR's own argument is that shim deletion converts one failure shape into another, so detection is the durable fix.
- **"The issue's secondary question (whether callers distinguish exit 127 from a real non-zero verdict) is a separate gate-attribution concern and is not addressed here."** — explicit out-of-scope exclusion, `tracked=false`. The rationale is sound, but see Miscellaneous: with #2566 closed as `NOT_PLANNED` there is now nowhere this sub-question lives.

### Review Delta (vs prior review on HEAD `915ed93`)

- **Resolved — the shimmed-and-never-installed tech debt.** `TestShimmedAndNeverInstalled` (3 tests) constructs exactly the state the prior reviewer said no fixture reached: `never-installed` declared in `pyproject.toml`, present in the shim dir, absent from `.venv/bin`. I re-ran the prior reviewer's exact mutation — stripping `not_installed.append(name)` and the `(not installed in the repo venv)` annotation from the resolved-elsewhere branch — and it now **fails 3** where it previously survived clean.
- **Resolved — the aggregate `path_note` suffix nit.** Dropping `; {n} also not installed in the venv` now **fails 1** (`test_mixed_host_aggregate_suffix_counts_the_absent_names`), and that assertion is on the counted string `1 also not installed in the venv`, so the per-name entry can no longer satisfy it by accident.
- **New** — PR body drift (Tech Debt above).
- **New** — silent truncation in the `fix` string (Nit above).
- **New** — #2566 closed `NOT_PLANNED` (Miscellaneous above). Not present at the prior review; the closure happened 2026-08-11, a day after it.
- **Unchanged** — nothing carried forward. Both prior findings are closed.

### Verification Results

No `## Verification` table to run — this branch has no plan document, so the checks below are the ones I chose.

- **Suite** — `PYTHONPATH=$PWD ./scripts/pytest-clean.sh tests/unit/test_doctor_console_scripts.py tests/unit/test_doctor.py` → **64 passed** in 20.87s. This matches the `ca066dbb` commit message; it is *not* the 57 in the PR body.
- **Lint** — `ruff check` and `ruff format --check` on both changed files: clean, exit 0.
- **Live, on this machine** — `_check_console_scripts_resolve()` from `.worktrees/sdlc-2566`: `passed=True`, `25 console scripts resolve into /Users/tomcounsell/src/ai/.venv/bin`. That is the worktree allowance doing its job — the lane has no populated venv of its own and the main checkout's is what PATH serves.
- **Mutation battery — 7 mutations, 6 caught, each applied and reverted separately, baseline and post-restore both green, working tree clean afterward:**

  - *not-installed tagging stripped from the resolved-elsewhere branch* → 3 failed. **Caught.**
  - *not-installed tagging stripped from the `found is None` branch* → 3 failed. **Caught.**
  - *aggregate `; N also not installed` suffix removed* → 1 failed. **Caught.**
  - *`_same_file` hardlink clause removed from the venv match* → 1 failed. **Caught** (`test_hardlinked_copy_outside_the_venv_is_accepted`).
  - *main-checkout venv dropped from `_repo_venv_bin_dirs`* → 1 failed. **Caught** (`test_main_checkout_venv_accepted_from_a_worktree`).
  - *`misresolved[:5]` truncation removed* → 1 failed. **Caught.**
  - *`os.path.realpath(found_path.parent) == os.path.realpath(bin_dir)` clause removed from the match* → **17 passed. Survived.** Deliberately not filed as a finding: it is redundant rather than unpinned. Any PATH entry that is a symlink to a venv bin dir also makes `bin_dir / name` the same inode, so `_same_file` already accepts it — and removing `_same_file` instead does fail a test. The clause costs nothing and defends the same case twice; a test for behavior that cannot differ would be theatre.

### Screenshots

Not applicable. No UI files in the diff — `gh pr diff 2665 --name-only` returns `tests/unit/test_doctor_console_scripts.py` and `tools/doctor.py`.

### Disposition

The code is done. Two prior findings closed, verified by mutation rather than by reading, and the one mutation that survived is genuinely redundant defense. The verdict is CHANGES REQUESTED because approval is reserved for zero-finding reviews and there are three items — but two of them are edits to the PR description and the third is a one-line string change. The thing worth a human's attention before `/do-patch` runs is the issue state: #2566 is closed `NOT_PLANNED` on evidence from one machine, and this check exists precisely because one machine is not the fleet.

tomcounsell:
## Review: Approved

Fourth pass, at `63433c7e`. **All three findings from round 3 are closed, and I confirmed each one myself rather than reading the commit messages.** The nit is fixed and pinned by a new test, the PR body now describes the behavior that actually shipped, and #2566 is `OPEN` again so the `Closes` trailer resolves honestly.

Everything below was measured on this machine at `63433c7e`, in a detached review worktree, not read off the diff or inherited from the prior rounds.

### Rubric

- [x] **1. Plan vs. implementation match** — pass — *No plan doc (`docs/plans/` has no `sdlc-2566` or console-script entry), so I judged against #2566 itself. The issue asks for detection of console-script shims whose imports fail; the check measures resolution instead and argues in its docstring why that is the durable measure. Wider than the ask and correct: deleting a shim converts `ModuleNotFoundError` into `command not found`, and both shapes now land in one message.*
- [x] **2. New code quality** — pass — *Three new module-level functions plus a nested helper, each with a docstring carrying the reasoning rather than restating the signature. The lone `except Exception` in `_repo_venv_bin_dirs` is a deliberate swallow with a stated rationale and a `noqa` tag.*
- [x] **3. Test coverage** — pass — *18 tests, mutation-verified below.*
- [x] **4. Regression risk to existing callers** — pass — *Purely additive: one entry appended to the `get_checks` list, no existing check touched. I checked every other `get_checks` caller (`tests/unit/test_doctor.py` only) and the full 65-test doctor suite passes. Registration ahead of `_check_system_tools` is itself pinned by a test, which matters because that import mutates `os.environ["PATH"]`.*
- [x] **5. Data integrity** — n/a — *No models, no migrations, no persisted state.*
- [x] **6. Security** — pass — *No `eval`/`exec`/`subprocess`, no secrets. `shutil.which` over `os.environ["PATH"]` plus `Path.exists()` reads only.*
- [x] **7. Documentation accuracy** — pass — *I read `docs/features/local-doctor.md` directly. Its "Check Categories" table enumerates categories, not individual checks, and this one falls under the existing `Environment` row. `_check_popoto_floor`, `_check_git_hooks_installed`, and `_check_worktree_interpreters` all get the same treatment. `tests/README.md` lists neither `test_doctor.py` nor any sibling, so the new test file breaks no index convention.*
- [x] **8. PR body accuracy** — pass — *The round-3 drift is gone. Every number re-measured below: 18 tests, 65 passed, both pasted live blocks reproduce.*
- [x] **9. Disclosed deferrals** — pass — *Both are explicit out-of-scope exclusions with sound rationale; see the Acknowledged section.*
- [x] **10. Follow-up claims verified** — n/a — *The body makes no `filed as follow-up #N` claim.*

### Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — N/A — *No plan document for this slug.*
- **3. New `except Exception` blocks — each has logger/raise/swallow-ok** — PASS — *One, in `_repo_venv_bin_dirs`, tagged `# noqa: S110 -- the current checkout's venv is enough on its own`. Losing the main-checkout candidate degrades to checking only the current checkout, which is a safe floor.*
- **4. New integration tests — exercise serialization boundary (not in-memory only)** — N/A — *No serialization boundary. The analogous discipline here is real files and a real `PATH` instead of a mocked `shutil.which`, and the tests do that. The only patch in the file is `tools.doctor.PROJECT_DIR`.*
- **6. No hardcoded secrets or debug artifacts** — PASS — *No `print`, no `TODO`, no commented-out code in the diff.*
- **7. New public APIs — docstrings present** — PASS — *`_repo_venv_bin_dirs`, `_same_file`, `_check_console_scripts_resolve`, and the nested `_installed_in_venv`.*
- **8. Breaking changes — migration path documented** — N/A — *Additive check only.*
- **9. Tests added for new behavior** — PASS — *18, covering all three states plus the two deliberate non-flags.*
- **10. Tests cover the failure path (not just happy path)** — PASS — *16 of 18 are failure-path or non-flag tests; only two assert a clean pass.*
- **11. UI changes (if any) — screenshot captured** — N/A — *`gh pr diff 2665 --name-only` returns two Python files. No HTML, CSS, JS/TS, JSX/TSX, Vue, or template files, so the visual-proof gate is a no-op.*
- **12. Docs updated for user-facing changes** — PASS — *See Rubric item 7.*

### Verified

- [x] Code correctness
- [x] Test coverage
- [x] Plan requirements met (judged against #2566; no plan doc exists)

### Blockers

- None

### Tech Debt

- None

### Nits

- None

### Miscellaneous

- None

- **"No stale shim was deleted on any machine — that is host state, and the check now makes it visible with a remediation line."** — explicit out-of-scope exclusion, no tracking required. Sound, and it is the PR's central argument: shim deletion converts one failure shape into another, so detection is the durable fix.
- **"The issue's secondary question (whether callers distinguish exit 127 from a real non-zero verdict) is a separate gate-attribution concern and is not addressed here."** — explicit out-of-scope exclusion. Round 3 flagged that this sub-question had nowhere to live because #2566 was closed `NOT_PLANNED`. #2566 is now `OPEN` (`stateReason: REOPENED`), so the deferral is tracked again and the concern is resolved.

### Review Delta (vs prior review on HEAD `ca066db`)

- **Resolved — PR body drift (was Tech Debt).** The body now reads 18 tests and 65 passed, both of which I re-measured and both of which match. `## Change` describes the three states explicitly ("Three states, each with its own remedy") including the `uv sync` remedy for declared-but-not-installed, which is the state round 3 said was invisible. The mutation table has grown from 4 rows to 16.
- **Resolved — silent truncation in the `fix` string (was a Nit).** `63433c7e` mirrors the message's suffix. I applied the reviewer's own mutation, stripping `not_installed_suffix` back out of the `fix` f-string, and `test_fix_signals_omission_when_more_than_five_names_are_not_installed` **fails 1**, so the fixed behavior is genuinely pinned rather than merely present.
- **Resolved — #2566 closed `NOT_PLANNED` (was Miscellaneous).** The issue is now `OPEN` with `stateReason: REOPENED`, so `Closes #2566` resolves honestly and the deferred exit-127 sub-question keeps a home. This was the item round 3 said was worth a human's attention, and the human made the call.
- **New** — nothing.
- **Unchanged** — nothing carried forward.

### Verification Results

No `## Verification` table to run (no plan document), so these are checks I chose.

- **Suite** — `PYTHONPATH=$PWD ./scripts/pytest-clean.sh tests/unit/test_doctor_console_scripts.py tests/unit/test_doctor.py` → **65 passed** in 121.90s. Matches the PR body.
- **Test count** — 18 `def test_` in `test_doctor_console_scripts.py`. Matches the PR body.
- **Lint** — `ruff check` and `ruff format --check` on both changed files: clean, exit 0 both.
- **Live, healthy shape** — `_check_console_scripts_resolve()` with the real `pyproject.toml`: `passed=True`, `25 console scripts resolve into /Users/tomcounsell/src/ai/.venv/bin`. Reproduces the body's first pasted block exactly. Run from a worktree with no venv of its own, so this is also the worktree allowance doing its job.
- **Live, broken shape** — rebuilt the body's second pasted block from scratch under `PATH` with no `.venv/bin` and a stale shim dir holding two of the three SDLC names. Got `passed=False` and `25/25 console scripts do not resolve into the repo venv — /Users/tomcounsell/src/ai/.venv/bin is not on PATH; ... critique-resume-probe -> .../fakelib/bin/critique-resume-probe; critique-roster-check -> .../fakelib/bin/critique-roster-check; sdlc-push-guard -> not found; ... (+20 more)`, with no `also not installed` suffix. Matches the paste, including the absence of the suffix, which is correct here because all 25 names are installed in the real venv.
- **Mutation battery — 2 mutations, both caught, each applied and reverted separately, tree clean afterward.** I targeted the two guards I had not personally exercised rather than re-running the author's table:
  - *`not_installed_suffix` stripped from the `fix` f-string* (this round's entire behavioral change) → **1 failed. Caught.**
  - *`a_path_problem = len(not_installed) < len(misresolved)` collapsed to `True`* (the heart of the three-state split) → **2 failed. Caught** (`test_absent_name_is_not_called_shadowed`, `test_fix_points_at_uv_sync_not_at_path`).

**Two risks I probed that no prior round raised, both cleared:**

- **Push gating.** The check runs under `--quick`, and `--quick` backs the `--install-hook` pre-push hook, which exits 1 on doctor failure. There is direct precedent for worrying about this: `_check_catchup_kill_switch` is deliberately excluded from `--quick` with a comment saying a failure there "would block every push". I checked whether this check inherits that hazard. It does not on this fleet: `core.hooksPath` is `.githooks`, so the generated `.git/hooks/pre-push` is inert, and the active `.githooks/pre-push` never invokes doctor. Where the opt-in hook *is* installed, a failure means the fail-closed gates really are broken, and the hook already calls `sdlc-push-guard` by bare name one line earlier, so the check adds no new failure mode.
- **Background alert noise.** I grepped every `tools.doctor` reference in the repo. Doctor is only ever invoked interactively or by that opt-in hook. Nothing runs it under launchd, cron, or a scheduled reflection, so a PATH-shaped failure cannot become recurring background noise.

**One observation, deliberately not filed as a finding.** The `_same_file` hardlink allowance is currently latent: `USER_BIN_SCRIPTS` contains exactly one entry, `("scripts/sdlc-tool", "sdlc-tool")`, and `sdlc-tool` is not among the 25 `[project.scripts]` names this check iterates, so no name it examines is hardlinked outside the venv today. The allowance is cheap, tested, correct, and becomes load-bearing the moment a `[project.scripts]` name is added to `USER_BIN_SCRIPTS`. Flagging forward-defensive code that a test already pins would be churn, not a finding.

I also re-derived the one mutation the PR honestly reports as surviving (the `os.path.realpath(found_path.parent) == os.path.realpath(bin_dir)` clause). It is genuinely redundant rather than unpinned: `shutil.which(name)` always returns a path ending in `name`, so if the parent realpaths to a venv bin dir then `bin_dir / name` is the same inode and `_same_file` already accepts it. Disclosing it beat quietly deleting it.

### Screenshots

Not applicable. No UI files in the diff.

### Disposition

The code was already done at round 3, and this round closed the last three items without touching the check's logic beyond a four-line truncation suffix. Every number the body claims reproduces on this machine, both pasted live blocks reproduce, the two guards I mutated are pinned, and the two fresh risks I went looking for are both clear. Approving.

_Single-judge review: `SDLC_REVIEW_JUDGES` is unset in this environment and all three prior rounds on this PR were single-judge._

tomcounsell:
Follow-up filed: #2749, for the exit-127 vs real-gate-verdict attribution question this PR deferred.

Timing note for the record: the follow-up was filed and the `## Out of scope` section updated to reference it a few minutes after the squash landed, so `2560a191c` on main does not carry the reference. The linkage lives in this PR body, in #2749 (which refs both #2566 and this PR), and in a comment on #2566.

PR #2651, opened 2026-08-07
react() transport derivation: stop chatless sessions enqueueing reactions to telegram:outbox:0

valorengels:
## Summary
- `react()` now derives transport via the same `_resolve_transport(session, chat_id)` path `send()` already uses (from PR #2627), short-circuiting on `"system"` before any Redis write.
- The `session_executor` reaction guard now checks `session.telegram_message_id` falsy (not `is not None`) and passes `agent_session or session` so the reflection scheduler's `telegram_message_id=0` / `chat_id="0"` sessions never reach the callback with a dead transport.
- Extracted `utils/peer.py` (stdlib-only) so `tools/react_with_emoji.py` can check peer deliverability without paying the `popoto` import cost on every invocation. This module lives under `utils/`, not `models/`: a first pass placed it at `models/peer.py`, but importing any submodule of `models` runs `models/__init__.py` first regardless of the submodule's own imports, which still pulled in `popoto`/`redis` (measured ~110x regression, 0.02s → 2.2–3.5s, on the CLI happy path — review on this PR). Relocating to `utils/peer.py` (whose `__init__.py` is empty) fixes it; re-measured `tools/react_with_emoji.py`'s `_resolve_transport()` on the Telegram happy path post-fix at ~0.003–0.005s, and confirmed via a subprocess check that `popoto`/`redis` are absent from `sys.modules` after `from utils.peer import ...` (`tests/unit/test_room_resolution.py::TestPeerParseSingleHome::test_utils_peer_import_stays_cheap`, which replaced the AST-only predecessor test that couldn't see this cost).
- `bridge/telegram_bridge.py`'s `_react` closure and `bridge/email_bridge.py`'s `EmailOutputHandler.react` both accept the widened 4-arg `ReactionCallback` signature.
- `tools/react_with_emoji.py` no-ops on system transport for both `react()` and `--standalone`.

## Test plan
- [x] `./scripts/pytest-clean.sh tests/unit/test_output_handler.py tests/unit/test_react_with_emoji.py tests/unit/test_email_bridge.py tests/unit/test_room_resolution.py -q` — 230 passed
- [x] `./scripts/pytest-clean.sh tests/unit/test_bridge_react_transport.py tests/unit/test_agent_session_queue.py tests/unit/test_worker_down_reactions.py tests/unit/test_reaction_never_hostile.py tests/unit/test_bridge_worker_liveness_reaction.py tests/unit/test_session_executor_runner_dispatch.py -q` — 174 passed
[...]

Issue #2629, opened 2026-08-07
react() cannot derive a transport: reflection emoji reactions still RPUSH to telegram:outbox:0


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/react-transport-derivation.md

Residual half of #2497, surfaced during PR #2627's review. `agent/output_handler.py:1410-1442` `react()` takes no session parameter, so it cannot derive the system-Room transport; a chatless session calling `tools/react_with_emoji.py` still RPUSHes to `telegram:outbox:0` and produces the misleading "invalid Telegram peer" WARNING that #2497 complained about. Fix requires a signature change (pass the session or enough context to derive the transport), then route system-derived reactions to the same null sink #2627 added (or drop with a debug log — a reaction has no meaning in the system Room). Out of scope of #2627 by deliberate choice.

## Recon Summary

Performed 2026-08-07 against `main` @ `e6d0e2bc7` (after #2627 `68480ac08` and #2631 `f08ab6d3d` landed).

**Confirmed:**
- `agent/output_handler.py:1443-1476` `TelegramRelayOutputHandler.react(chat_id, msg_id, emoji)` still hardcodes `session_id = chat_id` and RPUSHes to `telegram:outbox:{chat_id}`. For a reflection session (`chat_id="0"`) the queue key is literally `telegram:outbox:0`. No session parameter exists, so `_resolve_transport(session, chat_id)` — added by #2627 at `agent/output_handler.py:466` — cannot be consulted.
- The live production caller is `agent/session_executor.py:2508`: `await react_cb(session.chat_id, session.telegram_message_id, emoji)`, fired at the end of *every* session. `react_cb` resolves via `_resolve_callbacks` (`agent/agent_session_queue.py:1511`) to `TelegramRelayOutputHandler.react` — registered in `worker/__main__.py:727`. Reflection sessions are `session_type="eng"`, `chat_id="0"` (`agent/reflection_scheduler.py:626-628`), so this path fires for them on every run.
- The resulting payload has `reply_to=None` (reflection sessions have no `telegram_message_id`), so the relay drops it at `bridge/telegram_relay.py:141` with `"Relay: skipping malformed reaction payload"`. The wasted RPUSH, the shared-bucket `telegram:outbox:0` key, and the misleading WARNING are all real.
[...]

PR #2666, opened 2026-08-07
Relay: one zero-peer guard for all three send paths (#2644)

### LINKED ISSUE #2644: Relay reaction path has no zero-chat_id guard (message path does)
## Problem

\`bridge/telegram_relay.py\` guards the **message** path against \`chat_id == 0\` at line 452-457, dropping with \`\"Relay: dropping message with zero chat_id (invalid Telegram peer)\"\`. The **reaction** path, \`_send_queued_reaction\` (line 117), has no equivalent guard: it only checks \`int(chat_id)\` parses (line 145-149), so \`chat_id=\"0\"\` passes straight through to \`set_reaction(client, 0, reply_to, emoji)\` and raises \`PeerIdInvalidError\`.

Today this is masked because chatless sessions also have a null \`reply_to\`, which trips the malformed-payload check at line 141 first. A chatless session that *does* carry a \`telegram_message_id\` would hit the raise.

## Context

Split out of #2629 as a deliberate No-Go. #2497 ruled the relay guard out of scope as defence-in-depth and #2629 fixes the source side (\`TelegramRelayOutputHandler.react\` derives the system transport and never enqueues). This issue is the remaining defence-in-depth leg.

## Desired outcome

\`_send_queued_reaction\` rejects \`chat_id == 0\` with the same warning shape the message path uses, so the two paths agree.

## Recon: Skipped

Trivial — recon unnecessary because the finding is a three-line asymmetry between two adjacent functions in one file, verified by direct read during #2629's recon: message-path guard at \`bridge/telegram_relay.py:452-457\`, reaction path at \`:117-150\` with no such guard. No unknowns to surface.

tomcounsell:
Closes #2644

## What

`bridge/telegram_relay.py` guarded the **message** path against `chat_id == 0` and dropped it with a warning. The **reaction** path did not, and neither did the **custom emoji message** path (which the issue does not name — same class, found while making the probe as wide as the claim).

`chat_id="0"` is the chatless-session placeholder. It survives every existing check on those two paths:
- `if not chat_id` — `"0"` is a truthy string, passes.
- `int(chat_id)` parse check — `int("0")` succeeds, passes.

It then reaches `set_reaction(client, 0, ...)` / `send_message(0, ...)` and raises `PeerIdInvalidError`.

The three inline checks are replaced by one `_deliverable_peer(chat_id, kind, message)` helper. Log shapes are unchanged: a local session id (`local-<uuid>`) stays at `debug` because it is routine — `FileOutputHandler` already wrote that output — and a zero peer stays at `warning` because it is a bug worth seeing.

## Verification

`tests/unit/test_bridge_relay.py::TestZeroPeerGuard`, six tests: each of the three send paths drops `chat_id="0"` without calling Telethon, negative supergroup ids and ordinary ids still send, and a local session id still drops quietly.

**Mutation-checked.** Deleting the zero branch from `_deliverable_peer` turns exactly the three drop tests red (`assert result is False` → `AsyncMock`, `assert result is None` → `AsyncMock`) while the three positive/quiet-drop tests stay green. The guards are reached.

78 tests green across `test_bridge_relay.py`, `test_relay_length_guard.py`, `test_telegram_relay_chat_log.py`, `test_telegram_relay_voice_note.py`. `ruff check` and `ruff format` clean.

## The issue's second leak site: mostly refuted, one real gap

The 2026-08-07 comment on #2644 says `tools/react_with_emoji.py --standalone` still enqueues reactions with `chat_id="0"`.

**On the default path it does not.** `standalone()` calls the same `_resolve_transport()` as `react()`, which routes a non-deliverable peer to `"system"` via `utils.peer.deliverable_telegram_peer` and returns before building any payload. PR #2651 (`baea50bf8`) fixed both entry points at once by putting the derivation in the shared helper.

**But there was one reachable configuration where it did reproduce**, found in review: `_resolve_transport()` returned the `VALOR_TRANSPORT` override *before* the peer check was consulted, so `VALOR_TRANSPORT=telegram` with `TELEGRAM_CHAT_ID=0` skipped it entirely and built a `chat_id="0"` payload. Measured directly:

```
TELEGRAM_CHAT_ID=0, no override              -> 'system'    (early return, no payload)
TELEGRAM_CHAT_ID=0, VALOR_TRANSPORT=telegram -> 'telegram'  (proceeds, builds chat_id='0')
```

`VALOR_TRANSPORT` is a documented operator override with no production setter, so the likelihood was low — and a residual hole on the source side is exactly the argument for the relay-side leg this PR adds. It is now fixed here too: the override forces the *transport*, not the peer's reachability, so a resolved `telegram` still consults the peer check. An unset `TELEGRAM_CHAT_ID` deliberately still resolves to `telegram`, so callers keep emitting their "must be set" error rather than silently no-opping.

## Not merged

Held for the coordinator's serial merge queue.


tomcounsell:

No blockers. The fix is correct, the sweep is genuinely as wide as the headline claims, and every load-bearing assertion in the PR body reproduced in my environment. Two tech-debt items and three nits keep this off approval.

I reproduced rather than credited the body's numbers (78 green, ruff clean, the mutation result) and ran an independent mutation battery per call site. I also independently re-derived the `VALOR_TRANSPORT` finding that `review-2672` raised in its parallel pass; both of us landed on it from different starting points, and it is recorded below with my own evidence.

**What I verified in my own environment** (private `git archive` of `bfb9bcc77` at `scratchpad/mut2666`, `PYTHONPATH` pinned to that tree, all runs serial):

- 78 passed across `test_bridge_relay.py`, `test_relay_length_guard.py`, `test_telegram_relay_chat_log.py`, `test_telegram_relay_voice_note.py`. Matches the body exactly.
- `ruff check` and `ruff format --check` clean on both changed files.
- **Mutation A** — neutralize the `chat_id_int == 0` branch in `_deliverable_peer`: `3 failed, 3 passed`. Exactly the three drop tests red, the three positive/quiet-drop tests green. Matches the body's claim precisely.
- **Mutations B/C/D** (mine, not in the body) — remove the `_deliverable_peer` call from each of the three sites individually. Each turns exactly one test red (`1 failed, 5 passed` in every case). Every call site is independently load-bearing; none is coasting on another's coverage. This is the strongest evidence that the third site (`_send_custom_emoji_message`) was genuinely unguarded rather than incidentally covered.
- **Completeness of "all three send paths"** — `KNOWN_MESSAGE_TYPES` has exactly three members and the `process_outbox` dispatcher has exactly three branches, all guarded. The fourth Telethon-touching function, `_maybe_send_oversized_text_as_file`, carries no guard of its own but is called only from `_send_queued_message` at `:589` and `:619`, both downstream of the `:472` guard. Transitively protected, not a gap.
- **Retry semantics unchanged** — a dropped zero-peer payload returns falsy, so `process_outbox` treats it as a failure and re-queues up to `MAX_RELAY_RETRIES`, then `_dead_letter_message` discards it (reaction/custom-emoji are discarded outright; the message path has its own `chat_id_int == 0` discard at `:839`). No infinite loop is introduced, and this is identical to the pre-existing message-path behavior.
- **Log-string changes are safe** — the two `debug` strings were reworded. `grep` across the repo finds no test, doc, or script that matches on them. The body's "log shapes are unchanged" refers to levels (debug for local-session, warning for zero peer) and that reading is accurate.
- **Referenced artifacts resolve** — #2629 CLOSED, PR #2651 MERGED, commit `baea50bf8` exists.

## Rubric

- [x] **1. Plan vs. implementation match** — n/a — *No plan document (`plan_exists: false`; PLAN and CRITIQUE both `skipped` for #2644, which recorded "Recon: Skipped — trivial"). Judged against the issue's Desired outcome instead: "`_send_queued_reaction` rejects `chat_id == 0` with the same warning shape the message path uses." Delivered, and correctly widened to the third path.*
- [x] **2. New code quality** — pass — *One helper replaces three inline checks. Docstring explains the debug-vs-warning split and names the issue. The bare `int()` parse is the pre-existing message-path logic moved verbatim, not new code — the divergence it carries is recorded under Miscellaneous rather than charged here.*
- [x] **3. Test coverage** — pass — *All three call sites mutation-verified independently (Mutations B/C/D), plus the shared zero branch (Mutation A). Positive coverage for both an ordinary peer and a negative supergroup id. The one soft spot is a non-mutation-sensitive quiet-drop test, recorded as a nit.*
- [x] **4. Regression risk to existing callers** — pass — *`_send_custom_emoji_message` and `_send_queued_reaction` are each called from exactly one place (`process_outbox`). The new early return produces the same falsy result the pre-existing failure paths already produced, so retry and dead-letter behavior is unchanged.*
- [x] **5. Data integrity** — n/a — *No schema, model, or migration changes.*
- [x] **6. Security** — pass — *No new `eval`/`exec`/`subprocess`/`raw`, no secrets. The warning logs the full payload dict, which for the message path can include user text — pre-existing behavior, unchanged in level or scope by this PR.*
- [x] **7. Documentation accuracy** — pass — *No stale references introduced and no user-facing surface changed. `docs/features/relay-retry-guard.md` is now an incomplete home for the zero-peer story; raised as a nit for the DOCS stage rather than charged as an inaccuracy.*
- [ ] **8. PR body accuracy** — fail — *The refutation section asserts an absolute that does not hold in one reachable configuration. See Tech Debt 1. Every other claim in the body reproduced exactly.*
- [x] **9. Disclosed deferrals** — pass — *The "Not merged / held for the coordinator's serial merge queue" disclosure is an explicit, sound scope statement. The "second leak site does not reproduce" section is a refutation rather than a deferral, and is judged under item 8.*
- [x] **10. Follow-up claims verified** — pass — *#2629 CLOSED, PR #2651 MERGED 2026-08-07, commit `baea50bf8` resolves. All three cited correctly.*

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — N/A — *No plan; validated against issue #2644's Desired outcome, which is met and exceeded.*
- **2. No-Gos from plan — none violated** — N/A — *No plan. The issue's own scope note (defence-in-depth leg only, source side already fixed by #2651) is respected.*
- **3. New `except Exception` blocks — each has logger/raise/swallow-ok** — PASS — *No new `except Exception`. The new `except (ValueError, TypeError)` in `_deliverable_peer` is narrow and logs at debug.*
- **4. New integration tests — exercise serialization boundary (not in-memory only)** — N/A — *Unit-level guard tests; the serialization boundary is unchanged by this PR.*
- **5. Plan internal consistency — spike findings match task steps** — N/A — *No plan.*
- **7. New public APIs — docstrings present** — PASS — *`_deliverable_peer` is module-private and carries a docstring that explains the two drop levels and cites #2644.*
- **8. Breaking changes — migration path documented** — N/A — *No breaking changes; two debug log strings reworded with no dependents.*
- **9. Tests added for new behavior** — PASS — *Six tests in `TestZeroPeerGuard`, all reached (mutation-verified per site).*
- **10. Tests cover the failure path (not just happy path)** — PASS — *Three drop tests and three send/quiet-drop tests. The quiet-drop one is over-determined; see Nits 1.*
- **11. UI changes (if any) — screenshot captured** — N/A — *Diff is `bridge/telegram_relay.py` and `tests/unit/test_bridge_relay.py`. No HTML, CSS, JS/TS, JSX/TSX, Vue, or template files. Visual proof gate is a no-op.*
- **12. Docs updated for user-facing changes** — N/A — *Internal defence-in-depth guard with no user-facing surface. A doc extension is suggested under Nits for the DOCS stage.*

### Blockers

- None

### Tech Debt

- [ ] **1. `tools/react_with_emoji.py:34-36`** — `override = os.environ.get("VALOR_TRANSPORT"); if override: return override.strip().lower()`

  **Issue:** The PR body's refutation says of the issue's second leak site: *"It does not, on current main."* That absolute does not hold. `_resolve_transport()` returns the `VALOR_TRANSPORT` override **before** `deliverable_telegram_peer` is ever consulted, so the peer check is skipped entirely in that configuration. I ran the function directly rather than reading it:

  ```
  TELEGRAM_CHAT_ID=0, no override                 -> _resolve_transport() == 'system'    (early return, no payload)
  TELEGRAM_CHAT_ID=0, VALOR_TRANSPORT=telegram    -> _resolve_transport() == 'telegram'  (proceeds, builds chat_id='0' payload)
  ```

  `react()` has the identical structure at `:57-66`. `VALOR_TRANSPORT` is a documented operator override (`tools/send_message.py:27,67`) with no production setter anywhere in the repo, so real-world likelihood is low.

  **This strengthens the PR rather than undermining it.** A residual hole on the source side is exactly why the relay needs a defence-in-depth leg, so the body's conclusion is *more* right than its evidence claims. The ask is only that the wording be narrowed — an absolute that is false in one reachable config is the kind of sentence a future reader will rely on when deciding whether the source side is safe to lean on.

  **Fix:** Narrow "It does not, on current main" to the default path, e.g. "It does not on the default path; the `VALOR_TRANSPORT` override at `:34-36` short-circuits the peer check, which is a separate, low-likelihood gap and is precisely what this defence-in-depth leg covers." No code change needed in this PR.

- [ ] **2. `bridge/telegram_relay.py:131`** — `chat_id_int = int(chat_id)`

  **Issue:** `utils/peer.py` already owns this predicate and documents itself as such — `agent/output_handler.py:456` calls it "the single home", and `tools/react_with_emoji.py` and `models/room.py` both route through it. `_deliverable_peer` re-derives it with a bare `int()` instead, and the two implementations disagree on real inputs. I compared them directly:

  ```
  chat_id          utils.peer   relay._deliverable_peer
  "+5"             False        True                     <-- divergent
  5.9              False        True                     <-- divergent
  True             False        True                     <-- divergent
  "0"              False        False
  "-1003900483201" True         True
  "--5"            False        False
  ```

  `utils.peer._numeric_peer` rejects a leading `+`, non-integral floats, and bools; bare `int()` accepts all three. A payload carrying any of those is treated as deliverable by the relay and non-deliverable by the source side — the exact drift class this PR set out to close, reintroduced one layer down. Reachability is low (chat_id normally arrives as a decimal string in JSON), which is why this is tech debt and not a blocker. It is also inherited rather than introduced: the bare `int()` came over verbatim from the old message-path check.

  **Fix:** Have `_deliverable_peer` derive from `utils.peer` so there is one parse. The two-level logging still works because the reason is recoverable from the parse result:

  ```python
  from utils.peer import _numeric_peer  # promote to a public name as part of this change

  peer = _numeric_peer(chat_id)
  if peer is None:
      logger.debug(...)   # local session — routine
      return False
  if peer == 0:
      logger.warning(...) # chatless placeholder leaked — a bug worth seeing
      return False
  return True
  ```

### Nits

- [ ] **1. `tests/unit/test_bridge_relay.py:1360`** — `test_reaction_path_drops_local_session_id_quietly`

  **Issue:** This test is not mutation-sensitive to the guard it names. In Mutation C I removed the `_deliverable_peer` call from `_send_queued_reaction` entirely and this test stayed green (`1 failed, 5 passed` — only the zero-peer test went red). The mechanism: with the guard gone, `int("local-abc123")` at `:188/:190` raises `ValueError` *inside* the `try`, which `except Exception` at `:201` catches, logs, and returns `False` from — so `set_reaction` is still never called and both assertions still hold. The test therefore passes for a reason unrelated to the code it is named after.

  Worse for its stated intent: on that path the failure is logged at **warning** (`"Relay: reaction send failed: ..."`), which is the precise outcome the docstring says must not happen ("no send, **no warning**"). The test asserts nothing about log level, so it cannot tell the two apart.

  **Fix:** Add a `caplog` assertion so the test discriminates — assert the drop is recorded at `DEBUG` and that no `WARNING` record was emitted. That makes it fail under Mutation C and actually pin the debug-vs-warning split the helper's docstring is built around.

- [ ] **2. `docs/features/relay-retry-guard.md:28`** — *"Both the persist side (`_dead_letter_message` ...) and the replay side (`replay_dead_letters` ...) guard against invalid peers by checking `chat_id_int == 0` ... The two guards must stay in lockstep"*

  **Issue:** That paragraph is the repo's written home for the zero-peer story, and it now describes only two of the guard sites. After this PR the send path has a named helper covering three more. A reader following the doc would conclude the send path is unguarded.

  **Fix:** Extend the paragraph to name `_deliverable_peer` and the three send paths it covers. Natural work for the DOCS stage rather than this PR, but flagging it here so it is not lost.

- [ ] **3. `bridge/telegram_relay.py:830-844`** — `_dead_letter_message`'s inline `int(chat_id)` + `chat_id_int == 0` pair

  **Issue:** The PR's stated goal is that the deliverability checks "cannot drift apart again", and it achieves that for the three send paths. `_dead_letter_message` still carries a fourth hand-rolled copy of the same two-part check. The `kind` parameter exists precisely to parameterize the log line, so routing this through the helper is nearly free.

  **Fix:** Either call `_deliverable_peer(chat_id, "dead letter", message)` there, or add a one-line comment explaining why the dead-letter path deliberately keeps its own copy (its log text is about *discarding a stored record*, not *dropping a send*, which is a defensible reason to keep it separate). Deliberate is fine; silent is what invites the next drift.

### Miscellaneous

- None

- **"Held for the coordinator's serial merge queue"** — not a code deferral; an explicit and sound merge-sequencing statement. No tracking issue required.

### Verification Results

No `## Verification` table to run — this PR has no plan document (`plan_exists: false`, PLAN and CRITIQUE both `skipped` for #2644). The PR body's own verification section was reproduced instead, and is recorded above.

### Screenshots

None — no UI files in the diff. The visual proof gate is a no-op for this PR.

tomcounsell:
Patch round applied — head SHA **`daf40e21f0231944f4cdefdabb62f6f52d3d78b4`**. All five items addressed; both tech-debt items turned out to warrant code, not just wording.

## TD1 — `VALOR_TRANSPORT` override

Reproduced your measurement exactly before touching anything:

```
VALOR_TRANSPORT=telegram + chat_id=0     -> telegram     (pre-patch: hole)
no override + chat_id=0                  -> system
```

Body section rewritten from "does not reproduce" to **"mostly refuted, one real gap"**, with your two-line evidence table.

I also **fixed it** rather than only narrowing the claim. You wrote "no code change needed in this PR," but issue #2644's own comment scopes this pass as "one guard + one call-site fix here closes the class," and leaving a known reachable hole documented-but-open seemed worse than a four-line change. The override forces the *transport*, not the peer's reachability, so a resolved `telegram` now consults the peer check:

```
VALOR_TRANSPORT=telegram + chat_id=0     -> system      (post-patch)
VALOR_TRANSPORT=telegram + real peer     -> telegram    (override honored)
VALOR_TRANSPORT=email                    -> email       (untouched)
nothing set                              -> telegram    (unchanged)
```

That last row is deliberate: an unset `TELEGRAM_CHAT_ID` still resolves to `telegram` so callers keep emitting their "must be set" error instead of silently no-opping. Demoting it to `system` would have been the tidier-looking change and a behavior regression.

## TD2 — one predicate

Confirmed the drift table in my own environment, all three rows. `_deliverable_peer` now derives from `utils.peer`.

`_numeric_peer` is promoted to **`numeric_peer`** as you suggested. Worth noting the underscore was already fiction: `models/room.py:39` imports it across a module boundary and `test_room_resolution.py:284` pins that import *by identity*. Renaming touched `utils/peer.py`, `models/room.py`, `agent/output_handler.py` (a docstring reference), and that test.

**Nit 3 folded in here rather than deferred.** `_dead_letter_message` was the fourth hand-rolled copy, and the parse is the part that actually drifts. It now shares `numeric_peer` but keeps its own log wording, with a comment saying why: it discards a stored record rather than dropping a send, and deliberately logs only the chat_id instead of the whole payload — which also avoids widening the payload-in-logs surface your rubric item 6 flagged. Deliberate and stated, not silent.

## Nit 1 — the over-determined test

This was the best catch in the review. Your Mutation C diagnosis is exact. The test now asserts the DEBUG record and the **absence of any WARNING**, which is the property the docstring was always claiming and never checking.

Re-running your Mutation C against the patched tree — it now takes the test with it:

| | before patch | after patch |
|---|---|---|
| Mutation C (guard removed from `_send_queued_reaction`) | `1 failed, 5 passed` — quiet test green | **`3 failed, 4 passed`** — quiet test red |

Two further mutations on the new guards:

| Mutation | Result |
|---|---|
| `_deliverable_peer` back to a bare `int()` (the drifted copy) | `test_guard_agrees_with_utils_peer_on_odd_inputs` red, `1 failed, 6 passed` |
| `_resolve_transport` returns the override bare (the pre-patch hole) | 3 red in `TestTransportOverrideDeliverability`, `3 failed, 18 passed` |

All restored by file copy, never `git checkout` — a checkout reverts the feature along with the mutation.

## Nit 2 — docs

Done in this PR rather than deferred to DOCS, since the paragraph is actively misleading right now. `docs/features/relay-retry-guard.md` now carries a five-row table (three send paths, persist, replay), names `_deliverable_peer` and the DEBUG-vs-WARNING split, and states that all five derive from `utils.peer.numeric_peer` and why that parse is deliberately stricter than `int()`.

## Verification

243 tests green: `test_bridge_relay.py`, `test_react_with_emoji.py`, `test_room_resolution.py`, `test_relay_length_guard.py`, `test_telegram_relay_chat_log.py`, `test_telegram_relay_voice_note.py`, `test_bridge_react_transport.py`, `test_output_handler.py`, `test_dead_letters.py`. `ruff check` and `ruff format` clean.

## Note for the merge queue

This round touches **`models/room.py`** (two lines: the import and one call site, from the rename). That is Workstream C's collision surface, so it is worth sequencing against anything of theirs that touches `models/`. Nothing else in the diff goes near `models/`.

tomcounsell:

Delta re-review of `bfb9bcc77..daf40e21f`. **No blockers.** The code is right everywhere I probed, all four requested changes landed, and every mutation reproduces. One finding, and it is a single false sentence in a doc this PR adds — but it is false about precisely the property the PR exists to establish, so it needs a line either way before merge.

Under Hard Rule 4 (approval is zero-finding only) that makes this CHANGES REQUESTED rather than APPROVED. I want to be plain that this is a one-line fix and I would approve on it alone; nothing here needs rework.

### The finding

`docs/features/relay-retry-guard.md` adds a five-row table, and the five rows are individually accurate — I checked each against the code, not the prose. The sentence underneath them is not:

> All five derive the peer from `utils.peer.numeric_peer`

Four do. The fifth does not. `bridge/dead_letters.py:55-58` still parses locally:

```python
try:
    chat_id_int = int(chat_id)
except (ValueError, TypeError):
    chat_id_int = 0
if chat_id_int == 0:
```

`git grep -n numeric_peer -- bridge/dead_letters.py` returns nothing. So the replay row — the one row outside `telegram_relay.py` — is the fifth hand-rolled copy, and this PR's own comment in `_dead_letter_message` already names that file and line as a lockstep partner ("narrowing only this guard without `dead_letters.py:57` is a no-op").

The behavioral divergence is narrow but real and is the same class the PR closes. `int("+5")` is 5, so a stored record with `chat_id="+5"` gets **replayed** by `replay_dead_letters` and would be **dropped** by every send path, since `numeric_peer("+5")` is `None`. Reachability is low, and this PR is the reason: `_dead_letter_message` now passes `persist_failed_delivery` a `numeric_peer`-validated int, so newly written records are clean and only legacy rows could carry an odd form.

**Fix — either is fine.** Convert `dead_letters.py:55` to `numeric_peer` (its `except → 0` collapses to `is None` plus the existing zero branch, and the outcome for every currently-reachable input is unchanged), or narrow the sentence to "Four of the five... the replay side still parses locally" with a follow-up issue. I mildly prefer the code fix, because a doc that says "four of five share the predicate" is an invitation to the drift the helper exists to prevent. **Severity:** tech_debt.

### Everything else verified

**The override fix (TD1).** I ran the state table rather than reading it — ten cases through `_resolve_transport()`, all correct. The hole is closed (`VALOR_TRANSPORT=telegram` + `chat_id=0` → `system`), a real peer under override still resolves `telegram`, `email` is untouched in both its forms, and an unsupported override still passes through. Most importantly the non-regression holds: **nothing set → `telegram`**, and `VALOR_TRANSPORT=telegram` with `TELEGRAM_CHAT_ID` unset → `telegram`, so callers keep producing their "must be set" error instead of silently no-opping. Splitting `_telegram_or_system()` out is what makes the override path and the default path provably share one answer rather than two that look alike.

Taking this in-scope was the right call. #2644's own comment scopes it as "one guard + one call-site fix here closes the class", and leaving the source-side hole open while shipping the relay-side guard would have left the class half-closed.

**The single predicate (TD2) and nit 3.** `numeric_peer` is the right promotion — the underscore was already fiction given the cross-module import, and `test_models_room_reexports_peer_helpers` pins it by `is`-identity rather than equality, which is the assertion that actually catches a re-implementation. `_dead_letter_message` sharing the parse while keeping its own wording is the correct split: it discards a stored record rather than dropping a send, and logging only the chat_id there avoids putting a payload in the logs. The rename is complete in code; the only surviving `_numeric_peer` strings are in a completed plan document, correctly untouched.

**Nit 1, the mutation-insensitive test.** Fixed and confirmed by the mutation that previously missed it. Removing the reaction-path guard now turns `test_reaction_path_drops_local_session_id_quietly` **red** — it asserts a DEBUG record and the absence of any WARNING-or-above, so it can finally distinguish the two paths its docstring always claimed to. Under that mutation: `3 failed, 57 passed` file-wide.

**The two new guard mutations**, both reproduce: reverting `_deliverable_peer` to a bare `int()` turns `test_guard_agrees_with_utils_peer_on_odd_inputs` red (`1 failed, 59 passed`), and returning the override bare turns exactly the three `TestTransportOverrideDeliverability` cases red (`3 failed, 18 passed`).

**Collision surface.** `models/room.py` carries exactly the rename — one import line and one call site — and `agent/output_handler.py` is a docstring reference. Nothing else in the diff goes near `models/`.

**Counts.** 243 passed across the nine named suites, reproduced exactly. `ruff check` and `ruff format --check` clean on all seven changed source and test files.

One housekeeping note: the PR body's Verification section still reads "78 tests green" from the previous round. The delta comment has the current 243. Worth reconciling so the body is not stale at merge.

tomcounsell:
Final round applied — head SHA **`09f7ae397197474689f82c5a91635989159d0e62`**.

## The finding was mine to catch and I missed it

My own `_dead_letter_message` comment names `dead_letters.py:57` as the lockstep partner of that guard, and I still wrote "All five derive the peer from `utils.peer.numeric_peer`" without checking the fifth. The sentence was false, and it was false in the one direction that matters: it told a reader the convergence was complete.

Fixed in code rather than by narrowing the sentence, per your preference — the sentence is now true instead of hedged.

```python
chat_id_int = numeric_peer(chat_id)
if chat_id_int is None or chat_id_int == 0:
```

Unparseable and zero collapse into one branch because `numeric_peer` returns `None` where the old `except -> 0` was already folding unparseable into zero. **Outcome unchanged for every currently-reachable input**, which the six existing dead-letter tests confirm. The send call below now reuses `chat_id_int` instead of re-parsing an already-validated value — re-parsing is how the sixth copy would have started.

## Mutation

`+5` is the natural probe and, as it turns out, the *only* one: `int("5.9")` and `int("--5")` raise, so the old `except -> 0` discarded them anyway. Only a value that `int()` accepts and `utils.peer` rejects can distinguish the two parses.

Restoring the local `int()` there:

```
FAILED test_replay_discards_peers_the_send_paths_would_drop[+5]
E   AssertionError: Expected 'send_message' to not have been called. Called 1 times.
E   Calls: [call(5, 'legacy record with an odd peer', reply_to=None)]
1 failed, 10 passed
```

That is the divergence itself in the failure output — replayed to peer 5. Restored by file copy, not `git checkout`.

New tests are parametrized both ways: five undeliverable forms discarded without a send, two ordinary peers (a supergroup and a plain id) still replayed, so the stricter parse cannot start eating good records. Each undeliverable case asserts its own premise against `deliverable_telegram_peer` first, so the test cannot quietly stop testing what it claims.

## Sixth-copy sweep

Three probes, wider than the claim: local `int()` parses of a chat id, zero-peer comparisons, and the `except (ValueError, TypeError)` shape near anything chat- or peer-named.

**Zero-peer comparisons repo-wide — exactly three, all now on `numeric_peer`:**

| Site | |
|---|---|
| `telegram_relay.py:144` | `_deliverable_peer` (the three send paths) |
| `telegram_relay.py:849` | `_dead_letter_message` (persist) |
| `dead_letters.py:64` | `replay_dead_letters` (replay) |

No sixth copy exists. Two other sites reach `send_message` with an `int(chat_id)` and neither is a copy of this guard:

- `telegram_bridge.py:3027` — gated by `_resolve_transport`, which returns before the `int()` for a `system` transport. The comment at `:3026` says so explicitly: *"BEFORE int(chat_id) -- int(\"0\") succeeds"*.
- `hibernation.py:215` — buffered-log replay. **No guard at all**, but also no parse/compare pair, so it is not a copy. A `chat_id="0"` entry raises `PeerIdInvalidError` into the per-entry `except`, which logs a warning and continues to the next entry — loud and self-limiting, unlike the relay's retry loop.

I am **not** fixing `hibernation.py` in this PR. I have expanded scope twice already at reviewer prompting (the custom-emoji path, then the `VALOR_TRANSPORT` override), both times correctly, but a third unrequested expansion is how a "final round" stops being final. It is also a recovery path I cannot exercise on this machine, and guarding it would convert a logged failure into a silent skip — a judgement call that deserves its own issue rather than a drive-by. Flagged here so it is on the record.

## Verification

250 tests green across the nine relay/peer/dead-letter files. `ruff check` and `ruff format --check` clean.

Docs updated so the paragraph states the divergence as fact rather than theory, and names both sites that keep their own log wording while sharing the parse.

## Merge queue

Unchanged from the last round: this branch touches **`models/room.py`** (two lines, from the `numeric_peer` rename), so it wants sequencing against any Workstream C work in `models/`.

tomcounsell:
## Review: Approved

Delta re-review of `daf40e21f..09f7ae397`. My finding is closed the way I hoped it would be — the code fix rather than the wording fix — and the doc sentence is now true because the code makes it true. **250 passed** across the nine named suites, `ruff check` and `ruff format --check` clean on all seven changed files. No findings.

### The convergence

`bridge/dead_letters.py` now parses with `numeric_peer` and collapses unparseable-or-zero into one branch, which is right: the old `except -> 0` was already folding `None` into the zero case, so the discard decision is unchanged for every input either version could reach. The detail I like most is `send_message(chat_id_int, ...)` reusing the validated int instead of re-parsing `int(chat_id)` at the call site. "Re-parsing is how the sixth copy would start" is exactly the right instinct, and it is the difference between converging the predicate and merely calling it once.

The mutation is my divergence made executable. Restoring the local parse fails `test_replay_discards_peers_the_send_paths_would_drop[+5]` with `AssertionError: Expected 'send_message' to not have been called. Called 1 times.` — the replay to peer 5 visible in the failure output rather than inferred. `1 failed, 10 passed`. Each undeliverable case asserting its own premise against `deliverable_telegram_peer` before exercising the path is a nice touch: it means the test cannot quietly stop testing what it claims to if the predicate ever moves.

### Sixth-copy sweep — reproduced

I ran my own three probes rather than trusting the count. Zero-peer comparisons repo-wide: exactly three, `dead_letters.py:64`, `telegram_relay.py:144`, `telegram_relay.py:849`, each paired with a `numeric_peer` call on the line above. Hand-rolled `try/except int(chat_id)` parses: one left, `tools/valor_telegram.py:115`, which is a bot-registration lookup that returns `None` for unregistered ids and never compares against zero — not a deliverability decision, correctly untouched.

Both near-misses classified correctly:

- **`telegram_bridge.py:3027`** is genuinely gated. `_resolve_transport(session, chat_id)` runs first and `if transport == "system": return` sits immediately above the `int(chat_id)`, with an inline comment saying why the order matters. Cleared.
- **`hibernation.py:215`** is unguarded, and leaving it out was the right call. There is no parse/compare pair — `send_message(int(chat_id), ...)` sits inside a `try/except Exception` that logs a WARNING and continues the loop, so a zero peer produces one logged failure per entry rather than a silent drop. Guarding it would convert a visible failure into a silent skip on a hibernation-replay recovery path, which is the wrong trade on a path whose whole job is not losing buffered output. Filing it separately is scope discipline, not a miss.

### One correction to the reasoning, not to the PR

The note that `"+5"` is the *only* distinguishing probe does not hold. I enumerated the divergence between the old local parse and `numeric_peer` across 28 inputs, and among strings — which is what `DeadLetter.chat_id` (`KeyField()`) stores — three diverge, not one:

| input | old local parse | `numeric_peer` |
|---|---|---|
| `"+5"` | sends to peer 5 | discards |
| `"5_0"` | sends to peer **50** | discards |
| `"1_000"` | sends to peer **1000** | discards |

`int()` accepts PEP 515 underscore separators in strings, so `int("1_000")` is `1000`. The rest of the reasoning is right: `"5.9"`, `"--5"`, `"0x10"`, and the unicode-digit forms all agree between the two, and `5.9`/`True` diverge only as non-string types that a `KeyField` round-trip cannot produce.

**I am deliberately not recording this as a finding**, and I want to be explicit about why rather than leave it to inference. Nothing shipped is wrong: the claim lives in PR discussion, not in the code, the test docstring, or the docs. The test docstring says only that `int("+5")` is 5, which is true. The parametrized list is a reasonable spread and fully serves the test's purpose, which is catching a reversion to a local parse — `"+5"` catches that on its own. Adding `"5_0"` would widen the probe to the whole divergence class rather than one instance, and I would take it whenever that file is next open, but it is optional strengthening rather than outstanding work, so it does not belong in the way of a merge under Hard Rule 4.

### Verdict

Approved. Four rounds in, #2644's class is genuinely closed: five sites, one predicate, each independently mutation-checked, with the two deliberate exclusions reasoned and tracked rather than overlooked.

Sequencing note for the merge queue: the only cross-workstream surface is the two-line `models/room.py` rename, which I confirmed is purely the import and one call site.

PR #2670, opened 2026-08-07
Wedge detector: require evidence something was missed, and reset on restart (#2475)

### LINKED ISSUE #2475: Root-cause the recurring Telethon update-loop wedge — detector+restart is not curing it (36 crashes/72h)
## The actual defect

Two hotfixes landed 2026-07-30 (`5332768ff`, `5d9515671`). Both are downstream mitigations of the same underlying failure: the Telethon **update-loop wedge** — the bridge is process-alive but `NewMessage` has silently stopped firing (issues #1408 / #1712).

The wedge is what drops live messages (so the reconciler must reach further back) **and** what drives the crash storm (so the watchdog fires alerts that strand sessions). Neither patch touches the wedge.

## Evidence the existing mitigation is not curing it

- 36 `bridge_update_loop_wedged` crash-tracker events in 72h on this machine (first ~07-25, last 07-30 ~05:00 UTC). **All 36 crashes were that single reason** — a sustained storm, not scattered noise.
- `logs/bridge_watchdog.log`: continuous `loop_wedged detected (beacon missing…)` every 5 minutes from 04:25 to 06:09 UTC on 07-30.
- The sibling machine ("Valor the Bald") saw 164 crashes in 24h from the same reason, per `5332768ff`'s own commit message.

CLAUDE.md documents the #1712 detector as auto-restarting with `catch_up=True` for "lossless backfill." A storm at this rate means detect-and-restart is **not durably curing the wedge** — it is a recovery loop that keeps re-firing.

## Why this is the priority

We have now hardened the blast radius twice without touching the cause. Each mitigation adds surface area (see the sibling issues on `limit=30` and the catchup/reconciler asymmetry) while the wedge keeps recurring.

## Work to do here

1. **Root-cause the wedge.** Why does Telethon's update loop stop firing while the process stays alive? Candidates to investigate rather than assume: connection/session state after network transitions, an exception swallowed inside the update handler, `catch_up` interaction with the update loop, asyncio task death, server-side session invalidation.
2. **Determine whether restart-with-catch_up is actually lossless.** If it were, the reconciler would not have needed a deeper lookback. Either the guarantee is overstated in the docs or the reconciler covers a path the detector does not. Resolve which — and fix the docs if the guarantee is not real.
3. **Consider detection quality.** The beacon-missing detector fires every 5 min during a storm; check whether restart is being attempted and failing, or succeeding and re-wedging.

## Before patching

Validate every claim above against the source and the live logs rather than trusting this issue. Search for side effects before changing the update loop or restart policy — this code path is the ingest hot path, and a wrong change here drops messages silently rather than loudly. Do not add another mitigation layer without first establishing why the existing one does not hold.

tomcounsell:
Closes #2475

Full root-cause narrative with file:line citations is in [this comment on #2475](https://github.com/tomcounsell/ai/issues/2475#issuecomment-5219890671). Summary below.

## The defect

`assess_update_flow()` declared a wedge when `last_probe_ok` was fresh and `bridge:last_update_received` was stale. Both halves fail.

`last_probe_ok` is refreshed every 180s by the reconciler (`bridge/reconciler.py:36,147-149`), so it is **always** fresh on a connected bridge. The rule therefore reduced to *"no inbound message in 4 hours"* — a quiet night, indistinguishable from a wedged update loop. The beacon has exactly one writer, `bridge/telegram_bridge.py:1196`, inside the `NewMessage` handler; it is a traffic signal being read as a liveness signal.

Nothing seeds that beacon on restart. `restart_bridge()` is a bare `launchctl kickstart` and the bridge writes an unrelated boot-SHA beacon at startup. So after a wedge restart the key still held its pre-restart timestamp, and the only thing between the new process and an identical verdict was the 5-minute startup grace against a 60-second tick:

| t | event |
|---|---|
| 0:00 | verdict → `log_crash("bridge_update_loop_wedged")` → SIGKILL + kickstart |
| 0:00–5:00 | new pid inside grace → suppressed |
| ~6:00 | first tick past grace; beacon now 4h06m old, never seeded → verdict again |

That predicts a restart every ~6 minutes, and the issue's own evidence records exactly that: *"continuous `loop_wedged detected` every 5 minutes from 04:25 to 06:09 UTC."* The cadence is the grace window, not a Telethon fault. It is also never throttled — `bridge_watchdog.py:554-563` exempts wedge-dominated storms from the restart circuit by design.

## The fix

**Silence is measured from `max(last_update_received, process_start_ts)`.** A restart clears the accusation; a real recurrence still re-fires after another full ceiling of silence. `get_process_start_ts` was already being fetched for the grace window, so this costs nothing.

**A verdict requires `bridge:last_missed_recovery`** — new key, stamped by the reconciler when a scan recovers ≥1 message the live path never delivered. The corroborating evidence has to come from outside the handler under suspicion: `last_update_received` and the per-chat `bridge:last_event:*` keys are both written by the `NewMessage` handler, so neither can testify that the handler has stopped — their silence is equally consistent with an idle account. The reconciler reaches Telegram over an independent API path, already computes `recovered`, and gives a quiet account nothing to stamp.

The secondary accelerator's per-chat `scan_iter` corroboration was circular for that same reason and is removed; the accelerator now applies the primary shape at the warn threshold with equally recent evidence.

**No bridge pid is now the same fail-safe as an unreadable start time.** Process age is the floor for the silence measurement, not just the grace window, so a verdict without it cannot be bounded to the current process.

## Not changed, deliberately

`catch_up` losslessness is fine and is not the defect. `catch_up=True` is unconditional (`telegram_bridge.py:1158`) and `bridge/catchup.py:218-244` takes `min(global_cutoff, cursor - epsilon)`, never `max`, so the short lookback during a restart storm does not truncate recovery. The reconciler needed its deeper lookback (`5d9515671`) for its own fixed 30-minute window. The docs are accurate on this point and are left alone.

## Verification

162 tests green: `test_bridge_liveness.py`, `test_bridge_watchdog.py`, `test_reconciler.py` (unit + integration), `test_update_loop_wedge_recovery.py`, `test_watchdog_to_bridge.py`, `test_dm_recovery.py`. `ruff check` and `ruff format` clean.

**Mutation-checked, three guards, each independently:**

| Mutation | Expected red | Result |
|---|---|---|
| Measure silence from the beacon only (pre-fix) | restart-reset test | `test_restart_resets_the_silence_clock` failed, 20 passed |
| Drop the recovery-evidence requirement from the primary rule (pre-fix) | the three false-positive tests | `test_quiet_account_past_ceiling_is_not_wedged`, `test_absent_beacon_on_fresh_redis_is_not_wedged`, `test_stale_recovery_evidence_does_not_corroborate` failed, 18 passed |
| Stop stamping the key in the reconciler | the stamp test | `test_stamps_when_messages_were_missed` failed, 1 passed |

## Tests whose contract this changes

`tests/integration/test_update_loop_wedge_recovery.py` cases (a) and (e) asserted that silence past the ceiling with a fresh probe **is** a wedge. Those pinned the false positive as the specification, so they are replaced rather than kept. Case (e)'s underlying concern — that reconciler *probe* activity must not re-green the update signal — still holds and is still tested; what changed is that reconciler *recovery* activity is now a required input rather than an irrelevance.

## Note for review

This machine is worker-only; the bridge is deliberately stopped here, so there is no live reproduction. Everything above is from code, git history, the installed Telethon 1.42.0, and the issue's own crash-tracker evidence. Local `logs/watchdog.log` is unit-test output (`monitoring/bridge_watchdog.py:87-93` attaches a `RotatingFileHandler` to the production log path at module import) and was not used as evidence.

## Not merged

Held for the coordinator's serial merge queue.


tomcounsell:

The diagnosis holds up under independent verification, and the fix is the right shape: the corroborating signal now comes from the reconciler, which reaches Telegram over a path the `NewMessage` handler cannot influence, so it can testify to that handler's silence in a way `bridge:last_update_received` and `bridge:last_event:*` never could. Measuring silence from `max(last_update, start_ts)` is what stops the verdict from surviving its own restart. Both guards were mutation-checked independently in this review and both are genuinely reached.

No blockers. One documentation-accuracy finding: two files outside the diff still describe the detector by the rule this PR removes, and one is a sentence a future reader would reasonably re-implement from.

**Independently reproduced in this review** (worktree at `da359a49`, PR head confirmed via `git ls-remote refs/pull/2670/head`):

- 162 tests green — `test_bridge_liveness.py` + `test_update_loop_wedge_recovery.py` + `test_bridge_watchdog.py` = 108 passed; `test_reconciler.py` (unit + integration) + `test_watchdog_to_bridge.py` + `test_dm_recovery.py` = 54 passed. Matches the PR body exactly.
- `ruff check .` → All checks passed. `ruff format --check .` → 1308 files already formatted.
- Five mutations, each applied and reverted in isolation:

| Mutation | Result |
|---|---|
| `silence_since = last_update` (drop the `max` with `start_ts`) | `test_restart_resets_the_silence_clock` failed, 20 passed |
| Drop `missed_age` requirement from the PRIMARY rule | `test_quiet_account_past_ceiling_is_not_wedged`, `test_absent_beacon_on_fresh_redis_is_not_wedged`, `test_stale_recovery_evidence_does_not_corroborate` failed, 18 passed |
| Drop `missed_age` requirement from the SECONDARY accelerator | `test_secondary_accelerator_needs_recent_evidence`, `test_secondary_accelerator_does_not_fire_without_evidence` failed |
| Remove the `record_missed_recovery()` call from the reconciler | `test_stamps_when_messages_were_missed` failed, 1 passed |
| `bridge_pid is None` no longer fail-safes | `test_missing_pid_suppresses_verdict` failed |

- Line citations in the PR body spot-checked against the branch: `reconciler.py:36` (`RECONCILE_INTERVAL_SECONDS = 180`), `reconciler.py:147-149` (`record_probe_ok()`), `telegram_bridge.py:1196` as the sole writer of `bridge:last_update_received` (confirmed by repo-wide grep — `record_update_received` appears in exactly one non-test call site), `telegram_bridge.py:1158` (`catch_up=True` unconditional), `catchup.py:~240` (`per_chat_cutoff = min(cutoff, candidate)`, never `max`), `bridge_watchdog.py:571-577` (wedge-dominated storms leave `restart_circuit_open` False). All accurate.
- `bridge/reconciler.py` still contains no call to `record_update_received`, so the new evidence key stays independent of the handler under suspicion. Enforced by the existing regression guard `test_reconciler_does_not_call_record_update_received`.

## Rubric

- [x] **1. Plan vs. implementation match** — n/a — No plan document exists for this issue (hand-authored root-cause fix; PLAN and CRITIQUE never ran). Reviewed against issue #2475 instead: work item 1 (root-cause) is answered with evidence that the storm was the detector's own re-arming loop, item 2 (`catch_up` losslessness) is resolved with a citation rather than a change, item 3 (detection quality) is the body of the PR.
- [x] **2. New code quality** — pass — `record_missed_recovery`/`get_last_missed_recovery` mirror the existing probe/update pair exactly, including the best-effort `except Exception` → WARNING contract. The `_wedge_issue` closure is only reachable on branches where `missed_age is not None`, so the `missed_age / 60` format is safe on both call sites.
- [x] **3. Test coverage** — pass — 8 new unit tests on the key pair, 2 new reconciler stamp/no-stamp tests, and a restructured wedge suite covering both false-positive shapes, the restart reset, the re-fire after a full new ceiling, and both fail-safes. Five independent mutations confirmed each guard is reached.
- [x] **4. Regression risk to existing callers** — pass — `assess_update_flow` has a single caller (`check_bridge_health`, `bridge_watchdog.py:616`); the signature and return type are unchanged. The `bridge_pid is None` path moves from "proceed without a grace window" to "suppress the verdict", which is strictly the safer direction and is pinned by `test_missing_pid_suppresses_verdict`.
- [x] **5. Data integrity** — pass — one new freeform Redis key with the same 7-day TTL as its siblings; no Popoto-managed keys touched, no migration surface, no schema change.
- [x] **6. Security** — n/a — no new subprocess, eval, deserialization, or credential surface.
- [ ] **7. Documentation accuracy** — fail — two files outside the diff still describe the pre-#2475 rule. See the Tech Debt finding.
- [x] **8. PR body accuracy** — pass — every number and line citation in the body was reproduced or verified here; nothing was overstated.
- [x] **9. Disclosed deferrals** — pass — both disclosures are explicit out-of-scope exclusions with sound, citation-backed rationale. See Acknowledged Deferrals.
- [x] **10. Follow-up claims verified** — n/a — the PR body claims no follow-up issues.

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — N/A — *no plan document; validated against issue #2475's three work items instead*
- **2. No-Gos from plan — none violated** — N/A — *no plan document*
- **3. New `except Exception` blocks — each has logger/raise/swallow-ok** — PASS — *both new blocks in `bridge/liveness.py` log a WARNING and return, matching the documented best-effort contract shared with `bridge.dedup.record_last_event`*
- **4. New integration tests — exercise serialization boundary (not in-memory only)** — PASS — *`test_reconciler.py::TestMissedRecoveryStamp` drives real `reconcile_once` including the live `claim_message` Redis key; the watchdog suite mocks only Redis reads and runs the real `assess_update_flow`*
- **5. Plan internal consistency — spike findings match task steps** — N/A — *no plan document*
- **6. No hardcoded secrets or debug artifacts** — PASS — *none*
- **7. New public APIs — docstrings present** — PASS — *`record_missed_recovery` and `get_last_missed_recovery` both carry docstrings that state the contract, including why an empty scan must stay silent*
- **8. Breaking changes — migration path documented** — N/A — *no breaking change; the new key is absent-tolerant and its absence means "not wedged"*
- **9. Tests added for new behavior** — PASS — *see rubric item 3*
- **10. Tests cover the failure path (not just happy path)** — PASS — *quiet account, absent beacon on fresh Redis, stale evidence, stale probe, missing pid, unreadable process start, and Redis exception are all pinned*
- **11. UI changes (if any) — screenshot captured** — N/A — *no UI files in the diff; the visual proof gate is a no-op*
- **12. Docs updated for user-facing changes** — FAIL — *`CLAUDE.md` and `docs/features/bridge-self-healing.md` are updated and accurate, but two other files still state the removed rule — promoted to the Tech Debt finding below*

### Verified
- [x] Code correctness
- [x] Test coverage
- [x] Issue requirements met

### Blockers
- None

### Tech Debt

- [ ] **`docs/features/bridge-worker-architecture.md:158`** — `` `bridge:last_update_received` — written by the `NewMessage` handler before dedup; staleness while `bridge:last_probe_ok` is fresh indicates the Telethon update loop has wedged `` — This sentence asserts precisely the inference this PR removes as a false positive, and it is the shortest statement of the rule anywhere in the docs, so it is the one a future reader or agent would re-implement from. The adjacent list (lines 157-159) also enumerates only two liveness signals. Both need `bridge:last_missed_recovery` and the corrected inference. Per the repo's "describe only the new status quo" principle, leaving the old rule standing in a second doc is the residue the principle exists to prevent.

- [ ] **`docs/features/README.md:30`** — `` wedged-update-loop detector (`bridge:last_update_received` + `bridge:last_probe_ok` signals, issue #1712) `` — The feature index still describes the detector as a two-signal mechanism. It should name the third signal and reference #2475 alongside #1712.

### Nits

- [ ] **`monitoring/bridge_watchdog.py:317`** — `missed_age = None if last_missed is None else now - last_missed` — The evidence window equals the silence window, which leaves one narrow path where a restart-boundary recovery can still be read as wedge evidence. Sequence: the bridge restarts for any routine reason at `T`; a message that arrived across the restart gap is missed by the startup catchup scan and picked up by the reconciler at `T+3m`, stamping `bridge:last_missed_recovery`; the account is then quiet for four hours. At `T+4h`, `update_age` is `4h` measured from `start_ts` and `missed_age` is `3h57m`, so the PRIMARY rule fires on evidence that predates the silence it is being used to explain. This is self-limiting — the next restart resets `start_ts` and the evidence ages out of the window, so it costs at most one spurious restart rather than the old storm — and the startup catchup's per-chat cursor lookback covers the gap in the common case, which is why I am filing it as a nit rather than a blocker. The symmetric hardening is one line: require the evidence to postdate the current process's grace window (`last_missed >= start_ts + STARTUP_GRACE_SECONDS`) rather than merely to be younger than the threshold, which is the same "bound the verdict to this process" reasoning already applied to the silence clock.

### Miscellaneous
- None

- **"`catch_up` losslessness is fine and is not the defect" — issue #2475's work item 2 resolved by citation rather than by a change (`telegram_bridge.py:1158`, `catchup.py:218-244`), docs left alone** — tracked=false. Verified independently: `catch_up=True` is unconditional at `telegram_bridge.py:1158`, and the per-chat cutoff at `catchup.py:~240` is `min(cutoff, candidate)` with an explicit "Never max()" comment. The exclusion is sound and the docs are correct as they stand.
- **"This machine is worker-only; the bridge is deliberately stopped here, so there is no live reproduction"** — tracked=false. Sound: the analysis is grounded in source, git history, and the issue's own crash-tracker evidence, and the author correctly declines to cite the local `logs/watchdog.log` as evidence given the import-time `RotatingFileHandler` at `monitoring/bridge_watchdog.py:87-93` (verified present on this branch). The five mutation checks reproduced above are the substitute for a live repro and they carry the weight.
- **"`tests/integration/test_update_loop_wedge_recovery.py` cases (a) and (e) are replaced rather than kept"** — tracked=false. Correct call: those cases pinned the false positive as the specification, so preserving them would have made the fix untestable. Case (e)'s real concern (reconciler *probe* activity must not re-green the update signal) is still covered — `test_stale_probe_is_a_disconnect_not_a_wedge` and the probe-freshness gate at `bridge_watchdog.py:304-306` both survive.

### Deploy note

This PR touches `bridge/liveness.py`, `bridge/reconciler.py`, and `monitoring/bridge_watchdog.py`. Both the bridge and the watchdog run as separate launchd services, so the change is inert on any host until `./scripts/valor-service.sh restart` cycles them. Worth calling out at merge time given that the machines actually suffering the storm are the ones that need the new code running.

### Screenshots

tomcounsell:
## Review: Changes Requested (supplement — fail-safe direction)

Supplementing the review above, which I stand behind: its two documentation findings are real (I confirmed both files and that neither is in the diff), its nit about restart-boundary evidence is a correct catch, and I independently reproduced its verification. Adding one finding it did not cover, because it is the specific question this review was asked to answer: **is the new positive-evidence requirement itself fail-safe?**

It is not, in one reachable case, and the cost is not written down anywhere.

### The false negative this introduces

A wedge verdict now requires `bridge:last_missed_recovery`. Only the reconciler stamps it, and only when `recovered > 0`. But `record_probe_ok()` is called at `bridge/reconciler.py:149`, immediately after `get_dialogs()` at `:142` — **before** the per-chat scan loop. The loop's terminal handler at `:402` is `except Exception ... continue`, which swallows any per-chat failure.

So this state is reachable:

- `get_dialogs()` succeeds → `bridge:last_probe_ok` fresh → `probe_is_fresh` True
- every per-chat `iter_messages` scan throws → `recovered` stays 0 → `bridge:last_missed_recovery` never stamped
- Telethon's update loop is genuinely wedged → `update_age` grows past the ceiling
- `missed_age is None` → **no wedge verdict, no restart, indefinitely**

That is not hypothetical partitioning of failure modes. A half-wedged Telethon client where the dialog list still resolves but per-chat history iteration faults is exactly the correlated failure this detector exists to catch, and it is the shape where recovery and detection fail together.

Before this PR that same state restarted the bridge after four hours. Now it never does, and every message in the wedge window is lost with a `logger.error` as the only trace.

**The test suite already contains this case and asserts the new behavior is correct.** `test_quiet_account_past_ceiling_is_not_wedged` passes precisely because no evidence exists — and the watchdog has no way to tell that state apart from a wedged bridge whose reconciler scan is broken. The test is right about the quiet account and silent about the other reading.

To be fair to the design: I do not think this is fixable by making the detector cleverer, and I am not asking for that. The PR's central insight — that corroboration must come from outside the handler under suspicion — is correct, and some false negative is the unavoidable price of refusing to treat silence as evidence. The alternative is the storm. What I am asking for is that the price be named.

### Why it matters more than it reads

The reconciler's **recovery** path is now load-bearing for wedge detection. It was not before: previously only its **probe** gated the verdict. Nothing monitors it. `grep` for `silent_stream`/`silent_gap` in `monitoring/bridge_watchdog.py` returns nothing — the silent-gap check in `bridge/silent_stream.py` is observability-only, and the watchdog never reads it. A reconciler that probes cleanly while every scan throws is invisible to every automated signal on the machine.

Related and also undocumented: the operator kill switch returns at `bridge/reconciler.py:124`, before `get_dialogs()` and `record_probe_ok()`. With `CATCHUP_DISABLED_FLAG` set the probe goes stale and the wedge detector is suppressed entirely. That is the safe direction and arguably correct — with catchup off a restart could not backfill anyway — but an operator flipping that switch should know it silently disables wedge recovery too.

### Requested changes (consolidating both reviews)

1. Fix `docs/features/bridge-worker-architecture.md:158` and `docs/features/README.md:30` — per the review above. The first states the removed inference verbatim and is the shortest statement of the rule in the docs.
2. In `docs/features/bridge-self-healing.md`, name the residual false negative. The diff **deletes** the line "*This avoids false negatives on quiet-but-monitored bridges*" and replaces it with nothing that says what replaced that property. One sentence under Key design decisions is enough: a wedged bridge whose reconciler probes but recovers nothing will not be restarted, and that is accepted as the cost of refusing silence as evidence.
3. File a follow-up for reconciler-scan health monitoring. Recovery is now an input to a safety-critical verdict and has no health signal of its own. Not in this PR.

### What I reproduced independently

Worktree at `da359a49`, PR head confirmed via `resolve_pr_head_sha` and `git ls-remote refs/pull/2670/head`.

- Control: 50 passed on the three changed test files; 162 passed across all six files named in the body, matching the claim exactly. `ruff check` and `ruff format --check` clean on all six changed source and test files.
- All three claimed mutations reproduce exactly: beacon-only silence → `test_restart_resets_the_silence_clock` (1 failed, 20 passed); drop the PRIMARY recovery requirement → the three false-positive tests (3 failed, 18 passed); stop stamping in the reconciler → `test_stamps_when_messages_were_missed` (1 failed, 8 passed).
- On clock handling for `max(last_update, start_ts)`: `get_process_start_ts` parses `ps -o lstart=` as naive local time and converts with `time.mktime`, which interprets it as local — correct, and I verified empirically that a known pid yields an age of 0.7s rather than a timezone-offset value. A backward clock jump makes `update_age` negative and suppresses the verdict, which is the safe direction. The residual is DST ambiguity in `mktime` on the repeated hour, bounded at one hour twice a year against a four-hour ceiling, and it cannot produce a false positive on its own because the evidence requirement still has to be met. Adequate.

A process note, since it bears on how much weight to give any single number here: my first mutation run reported 6 failures where three consecutive re-runs of the identical mutation reported 1. The cause was mine — I used bare `pytest` instead of `scripts/pytest-clean.sh`, so my run collided with another agent's on this machine. Every figure above was re-measured under the sanctioned runner.

tomcounsell:
Patch round applied — head SHA **`e6b3e63740d1ad3d2854994ffb99ad49b3b31733`**. All three items addressed.

## 1 + 2. Docs stating the old rule

`docs/features/bridge-worker-architecture.md:158` said staleness while `last_probe_ok` is fresh *indicates* the update loop has wedged — the exact circular inference this PR removes. Rewritten, and while in there I fixed **line 159 as well** (it framed `last_probe_ok` only as a disconnect discriminator) and added the third key with the reason it has to be the reconciler that writes it:

> - `bridge:last_update_received` — … This is a *traffic* signal: its silence is equally consistent with a wedged update loop and with nobody having sent anything, so on its own it is never a verdict
> - `bridge:last_missed_recovery` — … This is the positive evidence a wedge verdict requires, and it has to come from here: the `NewMessage` handler writes both keys above, so neither can testify that the handler has stopped.

`docs/features/README.md:30` no longer describes the detector as two-signal.

**Sweep result (wider than the two named lines):** grepped `last_probe_ok`, `last_update_received`, `update loop wedg`, `update-loop wedg`, `wedge detector`, `wedged detector`, `two positive`, `two-signal`, `two signal` across `docs/` and `CLAUDE.md`. **No further occurrences of the old rule.** The remaining hits are unrelated uses of the phrase "two signal" in `subconscious-memory.md`, `skill-context-convention.md`, `sdlc-pipeline.md`, and `headroom-evaluation.md`. `bridge-self-healing.md` and `CLAUDE.md` were already rewritten in the first commit.

## 3. The nit, taken

You were right, and I'd underweighted it because I checked only that the evidence postdates `start_ts` — which it does automatically, since `now - CEILING >= start_ts` whenever the silence clears the ceiling. The failure is one window narrower than that: the stamp can land in the **grace window** specifically.

That is not a corner case, it is the common path. Every restart *creates* a gap, and catchup plus the reconciler recover it on the way back up — so a routine restart reliably stamps `bridge:last_missed_recovery` within its own first five minutes. Since the evidence window and the silence window are both `UPDATE_STALENESS_CEILING`, that stamp stayed admissible for a full four hours, and four quiet hours after any restart produced one spurious verdict. Every deploy would have had a chance of manufacturing exactly the crash this PR exists to stop.

```python
evidence_floor = start_ts + STARTUP_GRACE_SECONDS
missed_age = (
    None if last_missed is None or last_missed <= evidence_floor else now - last_missed
)
```

Same principle as the silence clock: bound the verdict to the process it accuses.

### Tests

`test_recovery_stamped_during_startup_grace_does_not_corroborate` isolates the case precisely — the stamp is at `start_ts + 290s`, inside the 300s grace, yet `now - last_missed` is ~3.9h, inside the 4h evidence window, so the window check alone would admit it. Three inline assertions guard the construction so the test cannot silently drift off the case it exists to cover and pass for the wrong reason.

`test_recovery_stamped_just_after_grace_does_corroborate` is its complement (`start_ts + 310s` → still wedged), proving the floor is a floor and not a blanket suppression of post-restart evidence.

**Mutation-checked.** Dropping the floor (`None if last_missed is None else now - last_missed`) turns the grace test RED and leaves the other 22 green, including the just-after-grace complement:

```
FAILED test_recovery_stamped_during_startup_grace_does_not_corroborate
1 failed, 22 passed
```

Restored by file copy, not `git checkout` — a checkout would have reverted the feature along with the mutation.

### Verification

164 tests green: `test_bridge_liveness.py`, `test_bridge_watchdog.py`, `test_reconciler.py` (unit + integration), `test_update_loop_wedge_recovery.py`, `test_watchdog_to_bridge.py`, `test_dm_recovery.py`. `ruff check` and `ruff format` clean.

Deploy note acknowledged — `bridge/` and `monitoring/` changes stay inert until a `valor-service` restart, which is in the coordinator's post-fan-out queue. No services touched from here.

tomcounsell:
Addendum applied — head SHA **`8c409247c03f7294c4aaacd96531b6e57b7b3ea0`**.

## The finding reproduces exactly

Verified against the code rather than taken on trust:

- `record_probe_ok()` — `bridge/reconciler.py:149`, immediately after `get_dialogs()` at `:142`
- `for dialog in dialogs:` — `:151`, i.e. **after** the probe stamp
- per-chat terminal handler — `except Exception as e: logger.error(...)` / `continue` at `:403-405`

So a half-wedged client keeps `last_probe_ok` fresh off the dialog call alone, every per-chat `fetch_messages_back_to` throws into that handler, `recovered` stays 0, `record_missed_recovery()` never fires. The detector sees fresh probe + silence + no evidence — byte-identical to a quiet account. `test_quiet_account_past_ceiling_is_not_wedged` passes on that state for precisely that reason. **Pre-#2475 that state restarted at 4h; now it never does.** That is a real regression in that one mode and it should be written down, not glossed.

You were right that my diff deleted "This avoids false negatives on quiet-but-monitored bridges" and put nothing in its place. That is the worst of both worlds — the old claim was false, but silence about it leaves the next reader to rediscover this the hard way.

## What I wrote

A **"What this detector can no longer see, and why we took that trade"** block in `bridge-self-healing.md`, covering four things: the mechanism with file:line, why no cleverer detector fixes it (every signal the wedged component emits is circular; the reconciler is the only independent observer, and when *it* is the failing part there is nothing left to ask), what bounds the exposure today, and the pointer to **#2691** for the monitoring work — stating current behavior plainly and pointing at the issue for the bounded-exposure work rather than promising a fix, including that the open question of whether persistent all-chat scan failure should re-enter the restart rule as evidence or should only page belongs there.

Compact versions also went into `bridge-worker-architecture.md` (one trailing clause on the evidence-key bullet) and into the `assess_update_flow` docstring — that one prefixed **"read this before concluding the detector is broken because a wedged bridge never restarted"**, since the code is where someone debugging that will actually land.

One thing I added that the review did not ask for, because it changes how the regression reads: **the pre-#2475 behavior in this state was not a working recovery either.** The restart re-armed the identical verdict six minutes later, so unless it happened to cure the fault, that state produced the storm rather than a fix. What this PR gives up there is closer to "a restart that probably was not helping" than "a restart that was."

## The floor's cost on the false-negative side

Stated rather than left implicit, and it is small:

- **Half-wedge case: no effect at all.** No stamp is ever written, so there is nothing for the floor to exclude. The detector was already blind there.
- **Real wedge with ongoing traffic:** every 180-second scan re-stamps, so the floor discards only the first stamp and evidence returns within a scan or two of the grace window closing.
- The only case it genuinely delays is a wedge whose sole recovery ever landed inside the grace window with no traffic afterwards — where there is no ongoing loss to detect.

## Verification

110 tests green (`test_update_loop_wedge_recovery.py`, `test_bridge_liveness.py`, `test_bridge_watchdog.py`); the 164-test set from the previous round is unchanged by this commit, which is docs plus one docstring. `ruff check` and `ruff format` clean.

Ready for delta re-review.

tomcounsell:
## Review: Approved

Delta re-review of the patch round, `da359a49..21adbd999` (three commits). Every requested change landed, both claims I was asked to hold to evidence survive scrutiny, and the mutation reproduces. **164 passed** across the seven bridge/watchdog suites — the 162 from before plus the two new precision cases. `ruff check` and `ruff format --check` clean.

### My blind-spot finding is properly named

This is what I asked for, and the final wording is better than what I asked for. The "What this detector can no longer see, and why we took that trade" block gets the mechanism right down to the line numbers — `record_probe_ok()` at `reconciler.py:149` firing before the per-chat loop at `:151`, the `except Exception ... continue` at `:403-405`, and the half-wedge state being byte-identical to a quiet account. It says outright that the old text claimed the opposite, which is the honest way to retire a deleted claim.

The last commit is the one that matters most. Rewriting "what bounds the exposure today" from a list of mitigations into **"nothing does, automatically... until some unrelated cause happens to restart the bridge, which is luck rather than a mechanism. Treat that as the honest status, not as a safety net"** is exactly right. A reader who skims that cannot come away thinking the gap is covered. Naming the log line that nobody reads (`[reconciler] Error scanning` at ERROR, once per chat per 180s) makes the gap concrete rather than abstract.

Coverage confirmed: `git grep -n 2691` over `docs/` and `monitoring/` returns exactly the three claimed hits, and #2691 exists and is scoped correctly ("Reconciler per-chat scan is now load-bearing for the wedge verdict but has no health monitoring"). Folding the open design question — whether persistent all-chat scan failure should re-enter the restart rule or only page — into that issue rather than deciding it here is the right call.

### The evidence floor

Clean, and the reviewer who raised it was right to upgrade it from a nit: a routine restart stamping its own backfill inside the grace window is the common path, not a corner. `evidence_floor = start_ts + STARTUP_GRACE_SECONDS` applies the same "bound the verdict to the process it accuses" rule already used for the silence clock, which makes it one idea rather than two.

The precision pair is exact — stamp at `start_ts + 290s` excluded, `start_ts + 310s` corroborates — and the excluded case carries three construction assertions that fail loudly if the fixture ever drifts off the state it exists to cover. That is the same "guard the guard" discipline as the inventory test, and it is what stops a precision test from quietly becoming a tautology. Mutation reproduces exactly as claimed: dropping the floor turns `test_recovery_stamped_during_startup_grace_does_not_corroborate` red, **1 failed, 22 passed**.

### Doc sweep — negative claim spot-checked and holds

I ran my own sweep rather than trusting the nine-phrasing one: every file in the repo mentioning `last_update_received` or `last_probe_ok`, filtered to those that never mention `last_missed_recovery`. Two survive, and both are correct to leave alone — `docs/plans/completed/bridge_stale_update_stream_detector.md` (the original #1712 plan, describing the state *before* the detector existed) and `docs/plans/completed/bridge_watchdog_level_5_livelock.md` (a dated incident record). Completed plan documents are historical by definition; rewriting them would falsify the record rather than clean it. No live documentation still asserts the old inference.

Incidentally, that second file corroborates the reframe below: it records the bridge sitting "wedged and undelivered" through 7+ continuous hours of recovery attempts.

### The two claims held to evidence

**"The pre-#2475 restart was not a working recovery."** Defensible, and strictly it is understated. I checked the old rule at the merge base (`probe_is_fresh and update_is_stale`, no process-start floor) and confirmed by grep that `record_update_received` has exactly one writer, `bridge/telegram_bridge.py:1196` in the `NewMessage` handler — nothing seeds it on restart. So after the SIGKILL the beacon still reads its pre-restart value, and the first tick past grace re-arms the identical verdict.

The caution that a restart *can* cure a wedged client session is fair, and the sentence already hedges for it. But curing the fault is necessary and **not sufficient** to stop the storm: the beacon only clears when a message is actually delivered by the live path. On a quiet account a fully cured bridge keeps storming until real traffic arrives. So "unless the restart happened to cure the fault it produced the storm" is if anything generous to the old behavior. I would not change the wording — it is defensible as written — but the mechanism is worth knowing if it ever gets re-litigated.

**The floor's ongoing-traffic cost.** Verified, and it is the load-bearing one. `STARTUP_GRACE_SECONDS = 300`, `RECONCILE_INTERVAL_SECONDS = 180`, so scans land at +180s, +360s, +540s and the first stamp clearing the floor is the one at +360s — a single scan after the grace window closes. The premise holds too: `record_missed_recovery()` fires whenever `recovered > 0`, and a wedged bridge with continuing traffic produces new undelivered messages on every scan, so it re-stamps every 180s. Since the verdict also needs four hours of silence, evidence has been re-stamped roughly eighty times before the ceiling is even reachable. The floor discards one stamp and costs nothing. The other two cases are stated correctly: zero effect in the half-wedge (no stamp is ever written), and delayed-but-not-lost when the sole recovery lands in grace.

### Verdict

Approved. The one blind spot this design cannot close is now documented in three places with its mechanism, its bound (none), and an issue to close it. That is the right disposition for a trade that is real and unavoidable.

Deploy note stands from the earlier review: this touches `bridge/liveness.py`, `bridge/reconciler.py`, and `monitoring/bridge_watchdog.py`, all of which run as launchd services, so the change is inert on every host until those services are cycled. The machines that suffered the storm are the ones that need the new code running.

Issue #2649, opened 2026-08-07
Reliability risk: do-plan-critique — forked skill contexts may lack the Agent tool, degrading roster lenses to serial execution


valorengels:
## Investigation Finding

**Component:** SDLC pipeline — `/do-plan-critique` roster dispatch
**Detected:** 2026-08-07 (reported by the `do-sdlc` fork driving the #2628 pipeline, session `sdlc-local-2628`)
**Status:** Unverified

## Symptoms

The fork running the #2628 SDLC pipeline reported that forked skill execution contexts have no Agent-spawn tool available, so `/do-plan-critique` roster lenses run **serially** rather than in parallel.

This has not been independently confirmed. It is a second-hand report from an agent that was mid-pipeline, and no traceback or tool-listing was captured at the time. It is filed because the roster barrier's design assumes subagent dispatch, so if the report is accurate the assumption is being silently violated on every forked critique run.

## Diagnostic Output

No raw output was captured — the report was narrative, and the reporting agent did not dump its available tool list. The closest available evidence is the skill's own stated dependency on the Agent tool.

`.claude/skills-global/do-plan-critique/SKILL.md:14`:

```
When the file is absent (the common case in a foreign repo), this skill runs
entirely on `git`, `gh`, and the Agent tool — it dispatches critics in the
foreground, waits for each to return its findings, aggregates, and prints a verdict
```

`.claude/skills-global/do-plan-critique/SKILL.md:226` draws the distinction that matters here:

```
The barrier is the robust form when the agent driver may return early from a
background dispatch; foreground-and-wait suffices when the driver reliably blocks.
```

So the roster barrier (`_roster.json` + per-critic `{name}.result.md` with a terminal completion fence) exists precisely to tolerate a driver that returns early from a *background* dispatch. If the Agent tool is absent entirely, there is no dispatch of either kind, and the driver must inline each lens in its own context instead.

Observed circumstantially during the same session: the #2628 critique ran three full rounds (PLAN, CRITIQUE, revision, re-CRITIQUE, second revision) over roughly an hour, which is consistent with serial lens execution but is not evidence on its own.

## Steps Attempted

None. The finding was relayed by another agent while it held the SDLC run lock for #2628, and verifying it would have required interrupting that pipeline to inspect the fork's tool availability. Filed for a clean-room check instead.

## Impact

If confirmed:

[...]

PR #2669, opened 2026-08-07
Stop resolving /update's merge target through the shared FETCH_HEAD (#2650)

tomcounsell:
Closes #2697. Refs #2650 — closes shape 2 outright, hardens shape 1 inside `/update`, and states the rules behind shapes 3 and 4. Does **not** close the issue; the plan-doc single-writer lease is the remaining acceptance criterion and needs a plan.

## Shape 2: the FETCH_HEAD race

`.git/FETCH_HEAD` is per-*repository*, not per-worktree. With concurrent SDLC lanes across many worktrees of one repo, bare `git pull` races: its merge step resolves the target through FETCH_HEAD, a peer lane's fetch lands between our fetch and our merge, and the merge reads their refs.

```
fatal: Cannot fast-forward to multiple branches
```

Three consecutive `/update` attempts died this way on 2026-08-07. The workaround was to pull only when the machine was quiet.

**The sweep found more than the issue reported.** The issue named `scripts/update/git.py`. The primary offender was `scripts/remote-update.sh` — it pulls *first*, before any Python runs. Widening further found four more, including `scripts/migrate_completed_plan.py`, which rebases on the shared main checkout where plan-doc migrations run.

| Site | Now |
|---|---|
| `scripts/remote-update.sh` | `git fetch origin main` + `git merge --ff-only origin/main` |
| `scripts/update/git.py::pull_ff_only` | fetch + merge/rebase the resolved `@{upstream}` |
| `scripts/update/run.py` | `git fetch origin main` + `git rebase origin/main` |
| `scripts/migrate_completed_plan.py` | same |
| `.claude/commands/update.md` (body + `allowed-tools`) | fetch + ff-merge |
| `.claude/skills/{update,setup,do-deploy}/SKILL.md` | fetch + ff-merge |
| `.claude/skills-global/weekly-review/SKILL.md` | `git fetch && git merge --ff-only @{upstream}` — the branch-portable form, since this skill is global |
| `.claude/skills-global/do-deploy-example/SKILL.md` | `git -C "$REPO" fetch origin main` + `merge --ff-only origin/main` — named ref, matching its deploy-to-main intent |

## Shape 1: the stash, inside `/update` itself

`refs/stash` is a per-repository stack too. `/update` stashes a dirty tree before fast-forwarding and restored it with a bare `git stash pop` — `stash@{0}`, whatever *anyone* pushed last. A peer stashing in between meant `/update` restored their work into the shared checkout and buried ours.

[...]

Issue #2124, opened 2026-07-16
plan-reviewer subagent can hallucinate a full critique (zero tool calls, fabricated plan) — needs a grounding guard


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/sdlc-fork-artifact-grounding-guards.md

## Observation (unverified beyond one occurrence)

During the #2120 SDLC run (PR #2121, 2026-07-16), the CRITIQUE stage's `plan-reviewer` subagent returned a complete-looking critique that was **entirely fabricated**: it reviewed a *different, nonexistent plan* and made **zero real tool calls** (never read the plan doc or any code). The supervising dev agent caught it, discarded the verdict, and self-verified the three root causes against real code with live repros before recording CRITIQUE=APPROVED.

## Why this matters

A hallucinated critique that *looks* structurally valid is worse than a failed one — if the supervisor hadn't checked tool-call counts, a fabricated READY TO BUILD (or a fabricated blocker) would have steered the pipeline. The critique gate's value depends on the reviewer having actually read the plan.

## Investigation

- [ ] Check the `plan-reviewer` agent definition (`.claude/agents/`) — its tool list is `read, grep, find, ls`; determine why a run can complete with zero tool calls (prompt not forcing a read? plan path not passed/resolvable from the fork's cwd in a worktree?).
- [ ] The worktree angle is suspect: the #2120 lane ran in `.claude/worktrees/agent-*`; if the plan path passed was repo-root-relative, the subagent may have failed to find it and improvised instead of failing loudly.
- [ ] Consider a structural guard: critique output must include verbatim quotes/line refs from the plan doc, or the invoking skill rejects verdicts from runs with zero file reads (tool-call-count check, like the docs-stage zero-tool-call guard shipped in #2076 WS5).

## Refs
Surfaced by the #2120 run report. Related precedent: #2076 WS5 (docs-stage tool-availability guard for the same 'agent completes without doing the work' class).

valorengels:
[...]

PR #2672, opened 2026-08-07
Let skills that mandate subagent dispatch actually dispatch (#2679)

### LINKED ISSUE #2679: Three skills mandate subagent dispatch while their allowed-tools omits Agent, so their fan-out silently inlines
> **Context for the stranger:** a `SKILL.md`'s `allowed-tools:` frontmatter key is a **restriction**. Per [`new-skill/SKILL.md:68`](../blob/main/.claude/skills-global/new-skill/SKILL.md#L68): *"Restricts which tools the skill can use. Omit to allow all."* A skill that declares the key and leaves out `Agent` cannot spawn subagents, whatever its body instructs.

## Problem

Three skills instruct the driver to dispatch subagents while forbidding themselves the tool that does it. All three are SDLC pipeline stages.

| Skill | `allowed-tools` | Body mandates |
|---|---|---|
| `do-pr-review` | no `Agent` | step 9: *"dispatch the judges and BLOCK on each returning IN THE SAME TURN"* |
| `do-issue` | no `Agent` | *"Spawn an Explore agent (thoroughness: 'very thorough')"*; *"Fan-out — Spawn one Explore agent per concern, all in parallel"* |
| `do-plan` | no `Agent` | *"Dispatch spikes in parallel - Use the P-Thread pattern (parallel Agent sub-agents)"* |

`do-issue` marks its recon mandatory (*"Skip only for trivially simple issues"*), and this repo opts `do-pr-review` into multi-judge consensus by default (`SDLC_REVIEW_JUDGES=code-quality,risk`, [`docs/sdlc/do-pr-review.md:215`](../blob/main/docs/sdlc/do-pr-review.md#L215)).

**Current behavior:** the failure is silent. The driver inlines the work instead of spawning it. Stages still complete, verdicts still post, no gate trips, nothing logs.

**What is actually lost** is the independence that makes fan-out worth doing. Subagents each reason in a fresh context; lenses inlined into one driver context share history, so a later lens can be anchored by whatever an earlier one concluded. A second judge that has already read the first judge's reasoning is not a second opinion. For `do-issue` and `do-plan` the same applies to parallel recon and spikes, plus the latency cost of running serially.

**Desired outcome:** every skill that instructs subagent dispatch is permitted to dispatch, and a guard makes a fourth instance a test failure rather than a discovery.

## Prior context

Surfaced while verifying [#2649](../issues/2649), which reported that *forked* skill contexts lack the Agent tool and therefore serialize `do-plan-critique`'s roster lenses. That framing does not reproduce — `do-plan-critique` declares no `allowed-tools` at all, so it inherits the full set, and it asks for foreground same-message dispatch, which works. #2649 should be closed on its own evidence; this issue is the real defect that investigation uncovered, which is why it is filed separately rather than reusing that number.

## Solution sketch

- Add `Agent` to the three skills' `allowed-tools`, matching the six comparators that already do (`audit-skills`, `do-integration-audit`, `imagine-agent`, `do-presentation`, `linkedin`, `x-com`).
- Add a sweep over both skill roots: a skill whose body instructs dispatch while its `allowed-tools` omits `Agent`/`Task` fails. Skills declaring no `allowed-tools` are skipped, being unrestricted by definition.

Two traps the guard has to avoid, both found by measurement rather than reasoning:

- **Negations.** `do-design-system` says *"do NOT delegate to a subagent"* and `authenticity-pass` returns a draft *to* a drafter subagent rather than spawning one. Both must stay green.
- **The other meaning of "agent".** This repo also calls a Claude Managed Agent an "agent" — a hosted product you *launch*. `build-agent` (*"launch the agent → run a graded outcome → schedule it on cron"*) is a false positive under a `launch`-based pattern.

Also worth fixing while there: a frontmatter reader that only handles the inline `allowed-tools: A, B` form parses a YAML block sequence to `[]`, which is indistinguishable from "restricted to nothing" rather than the `None` that means unrestricted. `calendar-sync` uses the block form today.

## Acceptance criteria

- [ ] `do-pr-review`, `do-issue`, and `do-plan` each carry `Agent` in `allowed-tools`.
- [ ] A sweep over `.claude/skills-global/` and `.claude/skills/` fails when a skill instructs dispatch without being allowed `Agent`, demonstrated red on each of the three independently.
- [ ] `do-design-system`, `authenticity-pass`, and `build-agent` stay green (negation and the Managed-Agent sense of "agent").
- [ ] Both YAML forms of `allowed-tools` parse correctly, verified against `calendar-sync`.
- [ ] The sweep guards its own inventory, so a glob that stops discovering skills fails loudly rather than vacuously passing.

Refs #2649, #2124, #2112.

tomcounsell:
Closes #2679. Refs #2649.

Three SDLC pipeline stages instruct the driver to dispatch subagents while their own `allowed-tools` omits `Agent` — which is a restriction, per `new-skill/SKILL.md:68` ("Restricts which tools the skill can use. Omit to allow all"). They ask for a dispatch they have forbidden themselves.

| Skill | Body mandates | Was |
|---|---|---|
| `do-pr-review` | step 9: *"dispatch the judges and BLOCK on each returning IN THE SAME TURN"* | no `Agent` |
| `do-issue` | *"Spawn an Explore agent"*, *"Fan-out — Spawn one Explore agent per concern"* (mandatory: *"Skip only for trivially simple issues"*) | no `Agent` |
| `do-plan` | *"Dispatch spikes in parallel - Use the P-Thread pattern (parallel Agent sub-agents)"* | no `Agent` |

**The failure is silent, which is why it survived.** The driver inlines the work instead of spawning it; stages complete, verdicts post, no gate trips. What is lost is the independence that makes fan-out worth doing: subagents each reason in a fresh context, while inlined lenses share the driver's history and can be anchored by whatever ran before them. A second judge that has already read the first judge's reasoning is not a second opinion.

## Why #2679 and not #2649

#2649 reported that *forked* skill contexts lack the Agent tool, serializing `do-plan-critique`'s roster lenses. That does not reproduce: `do-plan-critique` declares no `allowed-tools` at all, so it inherits the full set, and it asks for foreground same-message dispatch, which works. Its framing stays wrong, so it is the wrong `Closes` target. #2679 states the real defect; #2649 should be closed on the refutation evidence already posted there.

## The sweep — and why the first one was not one

The first version of this guard derived `DISPATCH_PATTERNS` from `do-pr-review`'s own wording. It therefore found `do-pr-review` and nothing else, while `do-issue` and `do-plan` sat in the identical state and passed green. Three independent near-misses: the literal token `subagent` (so "Spawn an Explore agent" slipped past), no hyphenated `sub-agents`, and three hardcoded nouns (so "Dispatch spikes" slipped past).

Patterns are now **verb + agent-noun proximity**, bounded to one clause, rather than an enumeration of phrasings.

Two false-positive traps, both found by measuring rather than reasoning:

**`launch` is deliberately not a dispatch verb.** The review suggested including it. On the first widened run it flagged `build-agent` — *"launch the agent → run a graded outcome → schedule it on cron"* — because this repo also uses "agent" for a **Claude Managed Agent**, a hosted product you launch rather than a subagent you spawn. Grepping both roots confirms no skill writes "launch … subagent", so excluding the verb costs no coverage. Reported back rather than applied as given.

**A negation guard** skips a dispatch phrase inside a prohibition, so `do-design-system` (*"do NOT delegate to a subagent"*) and `authenticity-pass` (which returns a draft *to* a drafter subagent rather than spawning one) stay green, as the review required.

## Frontmatter parsing

`_allowed_tools` read only the inline comma form, so a YAML block sequence parsed to `[]` — indistinguishable from "declared, restricted to nothing", and not the `None` that means unrestricted. `calendar-sync` uses that form today, and the review mutation-proved the consequence: a block-form skill legitimately declaring `Agent` got a failure whose message contradicted the file it described. Now `yaml.safe_load` with both forms normalized, falling back to the regex only when frontmatter will not parse. **The reviewer's exact mutation now passes.**

Not taken: moving the check into `audit_skills.py` as rule 22. It would inherit `--fix` and a better parser, but it edits a globally-synced skill script and its own test suite mid-fan-out — more blast radius than the parse fix needs. Recorded on #2679 as follow-up.

## Verification

**Demonstrated red in the order the review asked for.** Widening the patterns *before* the frontmatter fixes names exactly the two skills the review identified:

```
FAILED ...test_skill_that_dispatches_subagents_is_allowed_the_agent_tool[do-issue]
FAILED ...test_skill_that_dispatches_subagents_is_allowed_the_agent_tool[do-plan]
2 failed, 56 passed
```

Each of the three fixes is then independently mutation-checked — reverting any one turns exactly one of 58 cases red:

```
revert do-issue     -> 1 failed, 57 passed
revert do-plan      -> 1 failed, 57 passed
revert do-pr-review -> 1 failed, 57 passed
```

Negation controls and the parse fix, measured directly:

```
calendar-sync allowed-tools parses to 13 tools (was [])
do-design-system   dispatch hits = 0
authenticity-pass  dispatch hits = 0
do-issue / do-plan / do-pr-review: dispatch detected, Agent present
```

Suites and tooling:

```
test_skill_agent_tool_consistency + test_skills_audit + test_cma_skills_well_formed
+ test_sdlc_fork_no_background + test_sdlc_skill_md_parity + test_relink_global_skills
                                                              206 passed
audit_skills.py: do-issue / do-plan / do-pr-review     all fully aligned
ruff check + format                                            clean
```

## Scope note

This is a consistency check between two declarations in the same file. Whether a `context: fork` context *also* strips `Agent` at the harness level is a separate question, unverifiable from inside the repo, that this test deliberately does not speak to. The test's docstring carries the same limit so it travels with the code.

## Merge safety

**Safe to merge during the fan-out, and worth merging early.** Three one-word frontmatter additions plus one test file; no code path, gate, or hook behavior changes. Reaches running agents only after merge plus a `/update` hardlink sync.

For the coordinator: `do-pr-review` is the REVIEW stage every PR in this batch passes through, and `do-issue`/`do-plan` are the ISSUE and PLAN stages. Until this lands, all three run their fan-out inlined rather than independently.

**Merge-order note:** this branch edits `do-plan/SKILL.md` frontmatter (line 4) while PR #2669 edits the same file's body (Phase 4, ~line 393). Different regions, so git merges them cleanly in either order — flagging it only so the overlap is not a surprise.

Refs #2124, #2112.

### DIFF
--- a/.claude/skills-global/do-issue/SKILL.md
+++ b/.claude/skills-global/do-issue/SKILL.md
@@ -1,7 +1,7 @@
 ---
 name: do-issue
 description: "Create a self-contained GitHub issue ready for planning. Triggered by 'create an issue', 'file an issue', 'track this', or by /sdlc at Step 1."
-allowed-tools: Read, Write, Edit, Glob, Grep, Bash
+allowed-tools: Read, Write, Edit, Glob, Grep, Bash, Agent
 argument-hint: "<title or description>"
 ---
 
--- a/.claude/skills-global/do-plan/SKILL.md
+++ b/.claude/skills-global/do-plan/SKILL.md
@@ -1,7 +1,7 @@
 ---
 name: do-plan
 description: "Use when creating or updating a feature plan document. Triggered by 'make a plan', 'plan this', 'flesh out the idea', or any request to scope and plan work before implementation."
-allowed-tools: Read, Write, Edit, Glob, Bash, AskUserQuestion, ToolSearch, WebSearch
+allowed-tools: Read, Write, Edit, Glob, Bash, AskUserQuestion, ToolSearch, WebSearch, Agent
 ---
 
 # Make a Plan (Shape Up Methodology)
--- a/.claude/skills-global/do-pr-review/SKILL.md
+++ b/.claude/skills-global/do-pr-review/SKILL.md
@@ -3,7 +3,7 @@ name: do-pr-review
 description: "Review a pull request against its plan. Triggered by 'review this PR', 'check the pull request', 'do a PR review', or a PR URL."
 argument-hint: "<pr-number>"
 context: fork
-allowed-tools: mcp__byob__*, Bash(gh:*), Bash(git:*), Bash(python:*), Bash(jq:*), Bash(sdlc-tool:*), Read, Write, Edit, Grep, Glob
+allowed-tools: mcp__byob__*, Bash(gh:*), Bash(git:*), Bash(python:*), Bash(jq:*), Bash(sdlc-tool:*), Read, Write, Edit, Grep, Glob, Agent
 ---
 
 # PR Review
--- a/tests/unit/test_skill_agent_tool_consistency.py
+++ b/tests/unit/test_skill_agent_tool_consistency.py
@@ -1,0 +1,92 @@
+"""A skill that dispatches subagents must be allowed to (#2649).
+
+`allowed-tools` in a SKILL.md frontmatter is a restriction: per
+`.claude/skills-global/new-skill/SKILL.md`, "Restricts which tools the skill
+can use. Omit to allow all." So a skill whose body instructs the driver to
+dispatch subagents, while its own `allowed-tools` omits `Agent`, asks for
+something it has forbidden itself.
+
+`do-pr-review` was in exactly that state: `context: fork`, an `allowed-tools`
+list without `Agent`, and a body mandating judge subagents — with this repo
+opting into multi-judge consensus by default
+(`SDLC_REVIEW_JUDGES=code-quality,risk`, `docs/sdlc/do-pr-review.md`). The
+failure is silent: the driver inlines the judges instead, so reviews still post
+and no gate trips, they just lose the independent-context property that makes a
+second judge worth having.
+
+This is a consistency check between two declarations in the same file, not a
+claim about harness behavior. Whether a `context: fork` context also strips
+`Agent` is a separate question this test does not speak to.
+"""
+
+from __future__ import annotations
+
+import re
+from pathlib import Path
+
+import pytest
+
+REPO_ROOT = Path(__file__).resolve().parents[2]
+SKILL_ROOTS = (REPO_ROOT / ".claude" / "skills-global", REPO_ROOT / ".claude" / "skills")
+
+# Body phrasings that mean "spawn a subagent". Kept to unambiguous forms: a
+# skill merely mentioning the word "agent" is not dispatching one.
+DISPATCH_PATTERNS = (
+    re.compile(r"\bAgent tool\b"),
+    re.compile(r"\b(?:Agent|Task)\(\{"),
+    re.compile(r"\bdispatch(?:es|ing)? (?:the )?(?:critics|judges|subagents?)\b", re.I),
+    re.compile(r"\bjudge subagents?\b", re.I),
+    re.compile(r"\bspawn(?:s|ing)? .{0,20}subagents?\b", re.I),
+)
+
+
+def _skill_files() -> list[Path]:
+    files: list[Path] = []
+    for root in SKILL_ROOTS:
+        if root.exists():
+            files.extend(sorted(root.glob("*/SKILL.md")))
+    return files
+
+
+def _split_frontmatter(text: str) -> tuple[str, str]:
+    """Return (frontmatter, body). Empty frontmatter when there is none."""
+    if not text.startswith("---"):
+        return "", text
+    parts = text.split("---", 2)
+    if len(parts) < 3:
+        return "", text
+    return parts[1], parts[2]
+
+
+def _allowed_tools(frontmatter: str) -> list[str] | None:
+    """The declared tool list, or None when the key is absent (= allow all)."""
+    match = re.search(r"^allowed-tools:(.*)$", frontmatter, re.MULTILINE)
+    if not match:
+        return None
+    return [t.strip() for t in match.group(1).split(",") if t.strip()]
+
+
+def test_skill_inventory_is_non_empty():
+    """Guard the guard: a glob that stops finding skills must fail loudly
+    rather than silently vacuously pass."""
+    assert len(_skill_files()) > 20, _skill_files()
+
+
+@pytest.mark.parametrize("skill_path", _skill_files(), ids=lambda p: p.parent.name)
+def test_skill_that_dispatches_subagents_is_allowed_the_agent_tool(skill_path: Path):
+    text = skill_path.read_text()
+    frontmatter, body = _split_frontmatter(text)
+
+    allowed = _allowed_tools(frontmatter)
+    if allowed is None:
+        return  # no restriction declared: the full tool set, including Agent
+
+    dispatch_hits = [p.pattern for p in DISPATCH_PATTERNS if p.search(body)]
+    if not dispatch_hits:
+        return
+
+    assert any(t in ("Agent", "Task") for t in allowed), (
+        f"{skill_path.relative_to(REPO_ROOT)} instructs subagent dispatch "
+        f"(matched {dispatch_hits}) but its allowed-tools omits Agent, so the "
+        f"dispatch it mandates is forbidden to it: {allowed}"
+    )
--- a/tests/unit/test_skill_agent_tool_consistency.py
+++ b/tests/unit/test_skill_agent_tool_consistency.py
@@ -29,15 +29,40 @@ import pytest
 REPO_ROOT = Path(__file__).resolve().parents[2]
 SKILL_ROOTS = (REPO_ROOT / ".claude" / "skills-global", REPO_ROOT / ".claude" / "skills")
 
-# Body phrasings that mean "spawn a subagent". Kept to unambiguous forms: a
-# skill merely mentioning the word "agent" is not dispatching one.
+# Body phrasings that mean "spawn a subagent". A skill merely mentioning the
+# word "agent" is not dispatching one, so a dispatch VERB is always required.
+#
+# The first version of this set was derived from `do-pr-review`'s own wording
+# and therefore found `do-pr-review` and nothing else, while two skills sat in
+# the identical broken state and passed green. Three near-misses caused it: the
+# literal token "subagent" (so "Spawn an Explore agent" slipped past), no
+# hyphenated "sub-agents", and three hardcoded nouns (so "Dispatch spikes …
+# parallel Agent sub-agents" slipped past). Hence verb + agent-noun proximity
+# rather than an enumeration of phrasings.
+_AGENT_NOUN = r"(?:sub-?agents?|agents?|critics|judges)"
+# "launch" is deliberately absent. This repo also uses "agent" for a Claude
+# Managed Agent -- a hosted product you launch, not a subagent you spawn -- and
+# `build-agent` ("launch the agent -> run a graded outcome -> schedule it on
+# cron") is a false positive under it. No skill in either root writes
+# "launch ... subagent", so excluding the verb costs no coverage.
+_DISPATCH_VERB = r"(?:spawn|dispatch|fan[-\s]?out)"
+
 DISPATCH_PATTERNS = (
     re.compile(r"\bAgent tool\b"),
     re.compile(r"\b(?:Agent|Task)\(\{"),
-    re.compile(r"\bdispatch(?:es|ing)? (?:the )?(?:critics|judges|subagents?)\b", re.I),
-    re.compile(r"\bjudge subagents?\b", re.I),
-    re.compile(r"\bspawn(?:s|ing)? .{0,20}subagents?\b", re.I),
+    re.compile(r"\bExplore agents?\b", re.I),
+    re.compile(r"\bAgent sub-?agents?\b", re.I),
+    # A dispatch verb and an agent noun in the same clause. Bounded to 60
+    # characters and stopped at a sentence end so the two words have to be
+    # talking about each other.
+    re.compile(rf"\b{_DISPATCH_VERB}\w*\b[^.\n]{{0,60}}?\b{_AGENT_NOUN}\b", re.I),
 )
+
+# A dispatch phrase inside a prohibition is the opposite of an instruction to
+# dispatch. `do-design-system` says "do NOT delegate to a subagent" and
+# `authenticity-pass` returns a draft *to* a drafter subagent rather than
+# spawning one; both must stay green while the patterns widen.
+_NEGATION = re.compile(r"(?:do\s+not|don't|never|avoid|without|rather than|instead of)\s*$", re.I)
 
 
 def _skill_files() -> list[Path]:
@@ -59,7 +84,38 @@ def _split_frontmatter(text: str) -> tuple[str, str]:
 
 
 def _allowed_tools(frontmatter: str) -> list[str] | None:
-    """The declared tool list, or None when the key is absent (= allow all)."""
+    """The declared tool list, or None when the key is absent (= allow all).
+
+    Both YAML forms count. An inline `allowed-tools: A, B` and a block
+    sequence (`allowed-tools:` followed by `  - A`) mean the same thing, but a
+    regex reading only the inline form parses the block form as `[]` — which is
+    indistinguishable from "declared, restricted to nothing" and is *not* the
+    `None` that means unrestricted. `calendar-sync` uses the block form today,
+    so the next list-form skill to legitimately declare `Agent` would get a
+    failure whose message contradicts the file it describes.
+    """
+    try:
+        import yaml
+
+        data = yaml.safe_load(frontmatter)
+    except Exception:
+        data = None
+
+    if isinstance(data, dict) and "allowed-tools" in data:
+        value = data["allowed-tools"]
+        if value is None:
+            return []
+        if isinstance(value, str):
+            return [t.strip() for t in value.split(",") if t.strip()]
+        if isinstance(value, list):
+            return [str(t).strip() for t in value if str(t).strip()]
+        return []
+
+    if isinstance(data, dict):
+        return None  # parsed cleanly, key genuinely absent
+
+    # Frontmatter that YAML cannot parse: fall back to the inline form rather
+    # than silently reporting "unrestricted".
     match = re.search(r"^allowed-tools:(.*)$", frontmatter, re.MULTILINE)
     if not match:
         return None
@@ -72,6 +128,23 @@ def test_skill_inventory_is_non_empty():
     assert len(_skill_files()) > 20, _skill_files()
 
 
+def _dispatch_hits(body: str) -> list[str]:
+    """Patterns whose match is a genuine instruction to dispatch.
+
+    A match sitting inside a prohibition is skipped: a skill saying "do NOT
+    delegate to a subagent" is asserting the opposite of a dispatch.
+    """
+    hits = []
+    for pattern in DISPATCH_PATTERNS:
+        for match in pattern.finditer(body):
+            line_start = body.rfind("\n", 0, match.start()) + 1
+            if _NEGATION.search(body[line_start : match.start()]):
+                continue
+            hits.append(pattern.pattern)
+            break
+    return hits
+
+
 @pytest.mark.parametrize("skill_path", _skill_files(), ids=lambda p: p.parent.name)
 def test_skill_that_dispatches_subagents_is_allowed_the_agent_tool(skill_path: Path):
     text = skill_path.read_text()
@@ -81,7 +154,7 @@ def test_skill_that_dispatches_subagents_is_allowed_the_agent_tool(skill_path: P
     if allowed is None:
         return  # no restriction declared: the full tool set, including Agent
 
-    dispatch_hits = [p.pattern for p in DISPATCH_PATTERNS if p.search(body)]
+    dispatch_hits = _dispatch_hits(body)
     if not dispatch_hits:
         return
 


tomcounsell:
## Review: Changes Requested

The one-word frontmatter fix is correct and the demonstrated-red evidence reproduces exactly as claimed. The blocker is the second half of the PR's own thesis: "plus a sweep so it cannot recur." The sweep does not cover two skills that are in the identical state right now, and it passes them green.

Everything the PR body asserts, I reran in my own environment rather than crediting the numbers:

- Control on the new file: **58 passed**.
- Mutation (remove `, Agent` from `do-pr-review`'s frontmatter): **1 failed, 57 passed**, naming `do-pr-review` and listing the three matched patterns. Demonstrated-red is real.
- Positive-control mutation (append a dispatch phrase to `audit-hooks`, which is restricted and has no `Agent`): fails as designed.
- Inventory-guard mutation (repoint `SKILL_ROOTS` at a nonexistent path): `test_skill_inventory_is_non_empty` fails, and the parametrized test collapses to 1 skipped. The guard-the-guard case works.
- Neighbouring skill suites (`test_skills_audit`, `test_cma_skills_well_formed`, `test_sdlc_fork_no_background`, `test_sdlc_skill_md_parity`, `test_relink_global_skills`): **148 passed**.
- `ruff check` and `ruff format --check` on the new file: clean.

Coverage note in the PR's favour: the `*/SKILL.md` glob is not a depth gap. All 57 SKILL.md files in both roots sit at depth 1, and the 18 skills that declare no `allowed-tools` are correctly skipped as unrestricted. `Agent` is also the right token here, matching `audit-skills`, `do-integration-audit`, `imagine-agent`, `do-presentation`, `linkedin`, and `x-com`.

## Rubric

- [x] **1. Plan vs. implementation match** — n/a — *Hand-authored fix, no plan document; reviewed against the PR description.*
- [x] **2. New code quality** — pass — *Typed, `from __future__ import annotations`, thorough module docstring that states its own scope limits, ruff-clean. One robustness defect filed under Miscellaneous.*
- [ ] **3. Test coverage** — fail — *The guard's `DISPATCH_PATTERNS` do not detect two skills already in the exact state the PR set out to eliminate. See Blockers.*
- [x] **4. Regression risk to existing callers** — pass — *`allowed-tools` widening is additive; 148 neighbouring skill-suite tests still green.*
- [x] **5. Data integrity** — n/a — *No models, migrations, or persisted state.*
- [x] **6. Security** — pass — *Widening `do-pr-review`'s `allowed-tools` to include `Agent` is a deliberate capability grant to a skill that already holds `gh` write access; it is the intended fix and matches six comparator skills. No secrets, no `eval`/`exec`/`subprocess`.*
- [x] **7. Documentation accuracy** — pass — *No user-facing or architecturally-visible surface changes. The test's own docstring documents the invariant and honestly scopes out the harness-level question.*
- [ ] **8. PR body accuracy** — fail — *Every number in the body reproduces. The prose claim "plus a sweep so it cannot recur" does not: the sweep leaves two existing recurrences undetected.*
- [x] **9. Disclosed deferrals** — pass — *The scope note (harness-level `context: fork` behavior deliberately not addressed) is an explicit out-of-scope exclusion with sound rationale. The `do-plan-critique` fallback-detection item is offered as a separate hardening against #2649, which is OPEN.*
- [x] **10. Follow-up claims verified** — pass — *#2649 verified OPEN. #2124 and #2112 are historical `Refs`, not follow-up commitments.*

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — N/A — *No plan document.*
- **4. New integration tests — exercise serialization boundary (not in-memory only)** — N/A — *Filesystem-reading unit test.*
- **6. No hardcoded secrets or debug artifacts** — PASS — *No secrets, no leftover prints or TODOs.*
- **7. New public APIs — docstrings present** — PASS — *Module docstring plus docstrings on both helpers and the inventory guard.*
- **8. Breaking changes — migration path documented** — N/A — *Additive.*
- **9. Tests added for new behavior** — PASS — *And the red state is real, reproduced above.*
- **10. Tests cover the failure path (not just happy path)** — PASS — *Mutation-verified: the assertion fires and names the offending file and matched patterns.*
- **11. UI changes (if any) — screenshot captured** — N/A — *Diff touches one `.md` and one `.py`; no HTML, CSS, JS/TS, JSX/TSX, Vue, or templates.*
- **12. Docs updated for user-facing changes** — N/A — *No user-facing change.*

### Blockers

- [ ] **File:** `tests/unit/test_skill_agent_tool_consistency.py:32-38` (verified: read this file, and ran its own `DISPATCH_PATTERNS` against every skill body in the repo)
  **Code:** `DISPATCH_PATTERNS = (re.compile(r"\bAgent tool\b"), re.compile(r"\b(?:Agent|Task)\(\{"), re.compile(r"\bdispatch(?:es|ing)? (?:the )?(?:critics|judges|subagents?)\b", re.I), re.compile(r"\bjudge subagents?\b", re.I), re.compile(r"\bspawn(?:s|ing)? .{0,20}subagents?\b", re.I))`
  **Issue:** The pattern set was derived from `do-pr-review`'s own wording, so it finds `do-pr-review` and nothing else. Two skills are in the identical state the PR exists to eliminate, and the guard passes both green. I ran the module's own `_split_frontmatter` / `_allowed_tools` / `DISPATCH_PATTERNS` directly against them:

  - `.claude/skills-global/do-issue/SKILL.md` — `allowed-tools: Read, Write, Edit, Glob, Grep, Bash` (no `Agent`). Body line 68: *"Spawn an Explore agent (thoroughness: \"very thorough\") to map the affected area"*; line 72: *"**Fan-out** — Spawn one Explore agent per concern, all in parallel."* Line 80 makes it mandatory: *"Skip only for trivially simple issues (typo fixes, config changes)."* Measured `dispatch_hits=[]`.
  - `.claude/skills-global/do-plan/SKILL.md` — `allowed-tools: Read, Write, Edit, Glob, Bash, AskUserQuestion, ToolSearch, WebSearch` (no `Agent`). Line 203: *"**Dispatch spikes in parallel** - Use the P-Thread pattern (parallel Agent sub-agents) to run all spikes concurrently"*, with line 197 naming Agent Types and line 199 setting a per-spike time cap. Measured `dispatch_hits=[]`.

  Three independent near-misses cause this: `\bspawn(?:s|ing)? .{0,20}subagents?\b` requires the literal token "subagent", so "Spawn an Explore agent" slips past; `subagents?` does not match the hyphenated "sub-agents" in `do-plan`; and `dispatch(?:es|ing)? (?:the )?(?:critics|judges|subagents?)` hardcodes three nouns, so "Dispatch spikes" slips past.

  By the PR's own reasoning these are live defects in two SDLC pipeline stages, not hypotheticals. The harm is the false assurance: a green 58-case run reads as "the catalog is consistent" when two stages are silently inlining their fan-out.
  **Severity:** blocker
  **Fix:** Broaden the patterns to the shape "spawn/dispatch/launch/fan out + an agent noun", allowing `sub-?agents?`, bare `agents?`, and named agent types (`Explore`, `general-purpose`, `builder`, `Task`). Then fix the two skills the widened guard catches by adding `Agent` to their `allowed-tools`, in this PR. Guard against the opposite error while widening: `do-design-system:102` explicitly *forbids* delegating ("do NOT delegate to a subagent") and `authenticity-pass:120` returns a draft to its caller rather than dispatching, so both must stay green.

### Tech Debt

- [ ] **File:** PR description, "The fix" section (verified: read the PR body and reran its claims)
  **Code:** `One word in the frontmatter, plus a sweep so it cannot recur.`
  **Issue:** Every measurement in the body reproduces, which is genuinely rare and worth saying. This prose claim does not. "So it cannot recur" is the load-bearing justification for shipping a guard rather than just the frontmatter edit, and two existing recurrences pass it. The body is also what a future reader will trust when deciding whether the catalog has already been swept.
  **Severity:** tech_debt
  **Fix:** Once the blocker is addressed the claim becomes true and needs no edit. If the breadth work is instead split into a follow-up, narrow the sentence to what the guard actually asserts and file the follow-up issue.

### Nits

- [ ] **File:** PR description, header line
  **Code:** `Refs #2649.`
  **Issue:** The PR carries no `Closes #N`. That has a concrete mechanical cost in this repo rather than being a formatting preference: `do-pr-review`'s own context resolution (`sub-skills/checkout.md`) resolves `ISSUE_NUMBER` from `Closes|Fixes|Resolves #N` first, then a tracking-URL fallback, then `$SDLC_ISSUE_NUMBER` — all three come back empty here, so the skill is required to fail loudly rather than record a verdict against a guessed issue (#1731). I confirmed `sdlc-tool stage-query --issue-number 2649` returns `{"stages": {}}`, so no session exists to attach to either. The substrate verdict for this review is therefore unrecorded; the artifact below is the verdict.
  **Severity:** nit
  **Fix:** Add `Closes #2649` if this PR is what resolves that investigation, or open a small tracking issue for the guard and close that instead.

### Miscellaneous

- **File:** `tests/unit/test_skill_agent_tool_consistency.py:62-67`
  **Code:** `match = re.search(r"^allowed-tools:(.*)$", frontmatter, re.MULTILINE)` / `return [t.strip() for t in match.group(1).split(",") if t.strip()]`
  **Issue:** The regex reads only the inline comma form. A YAML block sequence parses to `[]`, which is indistinguishable from "declared, restricted to nothing" and is *not* the `None` that means "unrestricted". `.claude/skills-global/calendar-sync/SKILL.md` uses exactly that form today. I confirmed it parses to `[]`, and mutation-proved the consequence: adding `- Agent` to calendar-sync's list plus a dispatch phrase to its body still fails with *"its allowed-tools omits Agent ... : []"* — a false positive whose message contradicts the file it is describing. The direction is fail-closed, so nothing can slip through, but the next list-form skill that legitimately declares `Agent` gets a self-contradicting failure.

  Adjacent to this: `.claude/skills-global/audit-skills/scripts/audit_skills.py:235` already provides `parse_frontmatter`, which uses `yaml.safe_load` and handles both forms; `allowed-tools` is already in its known-fields list at line 77; and it carries a numbered-rule framework (rules 1-21) with `--fix` and its own test suite in `tests/unit/test_skills_audit.py`. This check is a natural rule 22 there, which would fix the parse and inherit auto-fix for free.
  **Severity:** tech_debt
  **Fix:** Reuse `audit_skills.parse_frontmatter` (or `yaml.safe_load` the frontmatter) and normalize both the scalar and sequence forms to a list before the membership test.

- **"Whether a `context: fork` context *also* strips Agent at the harness level is a separate question, unverifiable from inside the repo, that this test deliberately does not speak to."** — explicit out-of-scope exclusion, sound rationale, no tracking required. The test's docstring repeats the same limit, so the scope note travels with the code.
- **"`do-plan-critique` has no detection or fallback if Agent is ever genuinely unavailable ... a small independent hardening, not part of this fix."** — related to #2649 (verified OPEN).

### Legacy Cruft Audit

- None. The diff removes nothing, leaves no commented-out code, adds no fallback chain or dual implementation, and introduces no deprecated field. The `allowed-tools` edit is a single-token addition to an existing list.

### Verification Results

No plan document, so no `## Verification` table to run. The PR's own verification block was reproduced instead; results are listed at the top of this review.

### Screenshots

N/A. The diff touches `.claude/skills-global/do-pr-review/SKILL.md` and `tests/unit/test_skill_agent_tool_consistency.py` only. No UI files, so the visual-proof gate is a no-op.

tomcounsell:
## Patched — head is now `22a2a20154cefaecc7936b348300f7d93aa7e2bd`

Blocker, both tech-debt items, and the nit all addressed. Rebased onto current `main`. PR body rewritten. Thanks — the blocker was exactly right, and finding it by running my own `DISPATCH_PATTERNS` against every skill body is the check I should have run before claiming "so it cannot recur."

### Blocker — sweep width

Patterns are now **verb + agent-noun proximity** bounded to one clause, rather than an enumeration derived from `do-pr-review`'s wording. All three near-misses you named are closed: `sub-?agents?` covers the hyphenated form, bare `agents?` covers "Spawn an Explore agent", and the noun list is no longer three hardcoded words.

Both skills you identified are fixed: `do-issue` and `do-plan` now carry `Agent`.

### One correction to the suggested fix

You suggested `spawn/dispatch/launch/fan out`. **`launch` produces a false positive and is deliberately excluded.** On the first widened run the sweep flagged a third skill:

```
FAILED ...[build-agent]
FAILED ...[do-issue]
FAILED ...[do-plan]
3 failed, 55 passed
```

`build-agent` is about a **Claude Managed Agent** — a hosted product you *launch*, not a subagent you spawn: *"launch the agent → run a graded outcome → iterate → schedule it on cron"*. Four skills in the catalog use "agent" in that sense. I checked whether dropping the verb loses coverage before dropping it:

```
grep -rniE "launch\w*\s+[^.\n]{0,40}?sub-?agents?" .claude/skills-global/ .claude/skills/ --include=SKILL.md
(no matches)
```

No skill writes "launch … subagent", so excluding it costs nothing. With `launch` removed the sweep names exactly the two you identified.

Your negation warning was well placed and is handled explicitly: a match inside a prohibition is skipped, so `do-design-system` (*"do NOT delegate to a subagent"*) and `authenticity-pass` (returns a draft *to* a drafter subagent) both measure `dispatch_hits=0`.

### Miscellaneous — the YAML block-sequence parse

Fixed. `_allowed_tools` now uses `yaml.safe_load` and normalizes both the scalar and sequence forms, falling back to the regex only when frontmatter will not parse at all. Your exact mutation — adding `- Agent` to `calendar-sync`'s block list plus a dispatch phrase to its body — now **passes**, where it previously failed with the self-contradicting `: []` message:

```
calendar-sync allowed-tools parses to 13 tools  (was [])
mutation run: 1 passed
```

I did **not** move the check to `audit_skills.py` as rule 22. Your case for it is sound — better parser, inherits `--fix`, existing rule framework — but it means editing a globally-synced skill script plus its own test suite mid-fan-out, which is more blast radius than the parse defect warrants right now. Recorded as follow-up on #2679 rather than dropped.

### Nit — issue anchor

Filed **#2679**, framed at the real defect ("Three skills mandate subagent dispatch while their allowed-tools omits Agent"), and the PR now carries `Closes #2679` with `Refs #2649` retained. #2649 keeps its wrong framing, so it stays the wrong `Closes` target — the coordinator will close it separately on the refutation evidence.

### Tech debt — "so it cannot recur"

Now true, so the sentence stands. The PR body has been rewritten around the three-skill framing rather than the one-skill one.

### Verification

Demonstrated red in the order you asked for — widened patterns, frontmatter fixes not yet applied:

```
FAILED ...[do-issue]
FAILED ...[do-plan]
2 failed, 56 passed
```

Then each fix independently mutation-checked:

```
revert do-issue     -> 1 failed, 57 passed
revert do-plan      -> 1 failed, 57 passed
revert do-pr-review -> 1 failed, 57 passed
```

```
test_skill_agent_tool_consistency + test_skills_audit + test_cma_skills_well_formed
+ test_sdlc_fork_no_background + test_sdlc_skill_md_parity + test_relink_global_skills
                                                             206 passed
audit_skills.py: do-issue / do-plan / do-pr-review    all fully aligned
ruff check + format                                           clean
```

Not merging; ready for re-review.

tomcounsell:
## Review: Approved

Delta re-review against my round-1 verdict. **The blocker is cleared.** Every claim in the rewritten body reproduces in my own environment; I re-ran rather than read. Two non-blocking follow-ups below, one of which is a paper-trail gap worth closing before this scrolls out of view.

### The blocker is genuinely fixed

`do-issue` and `do-plan` both carry `Agent` now, and the patterns were rebuilt as verb + agent-noun proximity instead of an enumeration derived from `do-pr-review`'s own wording. Reverting all three frontmatter fixes against the widened patterns gives `3 failed, 55 passed`, naming exactly `do-issue`, `do-plan`, `do-pr-review` — the two I identified plus the original. Each fix is independently load-bearing: reverting any one alone gives `1 failed, 57 passed`. Clean control is `58 passed`; six skill suites together are `206 passed`; `ruff check` and `ruff format --check` clean.

One correction to my own process worth recording: my first control run reported `1 failed [do-issue]`, which was my own harness racing a mutation loop against a still-in-flight background run, not a real failure. Re-run serialized on a clean tree it is `58 passed`. The PR is not implicated.

### The `launch` deviation is correct, and better-founded than its stated reason

You excluded `launch` because including it flags `build-agent`, and supported that with "no skill writes 'launch … subagent'". That probe is narrower than the property it is being asked to carry — the guard's agent-noun set is `sub-?agents?|agents?|critics|judges`, so the question is whether any skill writes `launch` near *any* of those, not just `subagent`.

I ran the wider probe. It holds anyway. Across both roots, `launch` + any agent-noun hits exactly two skills: `build-agent` (`"launch the agent"`, `"launch a client's agent"` — the Managed Agent sense, correctly excluded) and `do-docs` (`"Launch the exploration agents"`), which declares no `allowed-tools` and is skipped as unrestricted regardless. So the exclusion costs zero coverage today. Right call, and reporting it back instead of applying my suggestion as given was the right move.

### Negation guard: exercised, not decorative

I checked whether `_NEGATION` is dead defensive code. It is not — it fires on two real bodies: `do-patch` (*"Do NOT spawn multiple agents"*) and `do-sdlc` (*"NEVER spawn agent…"*). `do-design-system` and `authenticity-pass` measure zero dispatch hits, as claimed. The per-pattern loop `continue`s past a suppressed match rather than abandoning the pattern, so one negated phrase cannot mask a genuine mandate later in the same file. That is the right structure.

### Parse fix confirmed

`calendar-sync` parses to 13 tools where it previously parsed to `[]`. My exact round-1 mutation passes.

### Findings (non-blocking)

**1. The accepted deferral is not actually recorded anywhere.** The body says moving the check into `audit_skills.py` as rule 22 is "Recorded on #2679 as follow-up." It is not. #2679 has zero comments, and its body does not mention `audit_skills` — its acceptance criteria cover the parse fix but not the relocation. The deferral itself is reasonable and the coordinator has accepted it; the problem is only that the record does not exist, so the follow-up would be lost the moment this merges. A one-line comment on #2679 closes it. **Severity:** tech_debt.

**2. `email` has the same defect one indirection away.** Out of scope for this PR's stated claim, so not a blocker, but it is live and it is the same harm. `de-slop` declares that a caller "MUST invoke it as a **fresh-context review** (a subagent or forked context)", and `de-slop` carries no `context: fork`, so a plain `Skill('de-slop')` runs inline in the caller's context. Of the four skills that wire that gate, three carry `Agent` — `do-presentation`, `linkedin`, `x-com` — and `email` declares `allowed-tools: Bash`. So every outbound email's de-slop gate is inlined into the drafting context, which is precisely the anchoring problem this PR exists to fix: a reviewer that has already read the draft is not a fresh reader.

The guard cannot see this and should not be stretched to. `email`'s body says "invoked as a fresh-context review" with no dispatch verb; the subagent requirement lives in `de-slop`'s contract, not `email`'s words. Catching it needs cross-skill contract awareness, which is a different instrument. Worth its own issue. **Severity:** tech_debt.

### Acceptance criteria on #2679

All five met and independently verified: three skills carry `Agent`; the sweep is demonstrated red on each independently; `do-design-system`, `authenticity-pass`, and `build-agent` stay green; both YAML forms parse, checked against `calendar-sync`; and `test_skill_inventory_is_non_empty` guards the glob.

### Merge safety and the #2669 overlap

The rebase brought nothing extra: two commits, four files, +168/-3, all expected. I tested the `do-plan/SKILL.md` overlap with #2669 rather than eyeballing the line numbers — merging both onto current `main` in a scratch worktree auto-merges with no conflict, and the guard stays `58 passed` on the combined tree. The two PRs are order-independent.

### Verdict recording

Not recorded in the substrate; the comment is the artifact. With `Closes #2679` now present the issue resolves, but no session or ledger exists behind it: `sdlc-tool stage-query --issue-number 2679` returns `{"stages": {}}` and `verdict get` returns `NO_VERDICT_SUBSTRATE`. `verdict finalize` requires `--run-id` (missing → `RUN_ID_REQUIRED`), and the only way to obtain one is `session-ensure`, which would fabricate a pipeline session mid-fan-out for a PR that never ran through `/do-sdlc`. Fabricating state to satisfy a recording step is worse than not recording. This was a standalone team review, so there is no router waiting on it.

PR #2710, opened 2026-08-10
SDLC stall auto-resume: steer, resume, or create an eng session; escalate once (#2696)

### LINKED ISSUE #2696: SDLC stall detector should auto-resume the pipeline by steering the eng session, not page a human every 6h
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/sdlc-stall-auto-resume.md

> **Context for the reader.** This repo runs an autonomous SDLC pipeline: an `eng`-role
> `AgentSession` drives an issue from plan to merged PR by invoking the `/sdlc` router one
> stage at a time. A scheduled maintenance job (a "reflection", see
> [`config/reflections.yaml`](../blob/main/config/reflections.yaml)) named `sdlc-progress-check`
> watches for pipelines that stopped moving. It lives in
> [`reflections/sdlc_progress.py`](../blob/main/reflections/sdlc_progress.py) and is documented
> in [`docs/features/pm-session-liveness.md`](../blob/main/docs/features/pm-session-liveness.md#state-layer-detection-sdlc-progress-check).

## Problem

`sdlc-progress-check` detects a stalled pipeline and then does nothing about it except send a
Telegram message to a human. Because a stalled PR by definition produces no new commits, and the
message is deduped on `(slug, last-commit-sha)` with a **6-hour TTL**, the identical message
re-fires every 6 hours forever.

### Current behavior

This is the real transcript from the `Eng: Valor` chat on 2026-08-09/10:

```
[valor] SDLC stall: PR #2688 (sdlc-2682) last commit 22h ago, no active session, issue #2682 open.
[valor] SDLC stall: PR #2676 (sdlc-2664) last commit 28h ago, no active session, issue #2664 open.
[valor] SDLC stall: PR #2681 (sdlc-2517) last commit 28h ago, no active session, issue #2517 open.
[valor] SDLC stall: PR #2688 (sdlc-2682) last commit 28h ago, no active session, issue #2682 open.
[valor] SDLC stall: PR #2681 (sdlc-2517) last commit 34h ago, no active session, issue #2517 open.
[valor] SDLC stall: PR #2688 (sdlc-2682) last commit 34h ago, no active session, issue #2682 open.
[valor] SDLC stall: PR #2688 (sdlc-2682) last commit 40h ago, no active session, issue #2682 open.
```

The human's response: *"Don't send me messages like this. I'm not your log file."*

Four copies of the same fact about PR #2688, six hours apart, none of which moved the PR. The
system had every input it needed to resume the pipeline itself and instead escalated to a person
who then had to hand-restart three lanes.

Two distinct defects are stacked here:

1. **No action.** `_check_project_stalls` composes a string and calls `_send_alert`
   (`reflections/sdlc_progress.py:300-307`). It never creates, resumes, or steers a session. This
   was a deliberate v1 scope decision — the module docstring says *"Notification-only for v1 —
   never creates or resumes PM sessions"* and the feature doc repeats it under **What this is
   NOT**. The decision is now wrong: the pipeline is autonomous enough that "a human notices a
   Telegram message and hand-restarts the lane" is the least reliable link in the chain.
2. **Escalating repeats.** `_dedup_set` (`reflections/sdlc_progress.py:229`) writes
   `sdlc:stall:alert:{slug}:{sha}` with `SET NX EX cooldown` where cooldown defaults to 6h
   (`_DEFAULT_COOLDOWN_HOURS = 6`). The key expires; the sha has not changed because the PR is
   stalled; the alert re-fires. There is no cap and no state that says "a human was already told
   about this exact head sha."

### Additional defect found while investigating

Gate 5 of the stall heuristic — "no non-terminal `AgentSession` for the slug" — is checked by
`_has_active_session(slug)` at `reflections/sdlc_progress.py:188`, which runs
`AgentSession.query.filter(slug=slug)`.

Local SDLC pipelines do not set `slug`. `tools/sdlc_session_ensure.py:840` creates them via
`AgentSession.create_local(session_id=f"sdlc-local-{N}", project_key=..., session_type="eng", ...)`
with no `slug` kwarg. Verified live: three running `sdlc-local-*` eng sessions all report
`slug=None`.

So a fully live local pipeline is invisible to gate 5, and the detector will report
"no active session" while a session is actively working the branch. Any auto-resume built on this
gate would spawn a duplicate lane on top of a running one. **This must be fixed before, or as part
of, the auto-resume change** — it is the difference between a self-healing detector and a
lane-duplication machine.

### Desired outcome

When a pipeline stalls, the system resumes it by itself. A human hears about it only when the
system tried to act and could not — once per stalled head sha, not on a 6-hour loop.

## Definitions

| Term | Meaning |
|---|---|
| **Reflection** | A scheduled maintenance job registered in `config/reflections.yaml` and dispatched by `agent/reflection_scheduler.py`. `sdlc-progress-check` runs every 1800s. |
| **`AgentSession`** | The Popoto/Redis-backed record of one agent session. See [`models/agent_session.py`](../blob/main/models/agent_session.py) and [`docs/features/session-lifecycle.md`](../blob/main/docs/features/session-lifecycle.md). |
| **Eng session** | `session_type="eng"` — the full-permission session that owns SDLC work for a project. See [`docs/features/eng-session-architecture.md`](../blob/main/docs/features/eng-session-architecture.md). |
| **Steering** | Injecting a message into a session's Redis inbox so it is consumed at the next turn boundary. `agent/steering.py::push_steering_message` is the sole writer API; the worker drains it. See [`docs/features/session-steering.md`](../blob/main/docs/features/session-steering.md). |
| **Slug** | The kebab-case key tying together a task list, branch (`session/{slug}`), worktree (`.worktrees/{slug}/`), plan doc, and issue. For SDLC lanes it is `sdlc-<issue-number>`. |
| **Stall gates** | The five conditions in `_check_project_stalls` that must all pass before a PR is called stalled. Listed in the feature doc. |

## Solution sketch

The reaction to a stall should be **a steering message to the eng session**, not a Telegram
message to a person. Steering is the right primitive: it is the sanctioned way to hand an
instruction to a session, it survives the session not being mid-turn, and it costs nothing if the
session is idle.

Open questions the plan should settle — the target-session question in particular is genuinely
unresolved and should not be assumed:

1. **Which session receives the steer?** Candidates, in rough order of preference:
   - a live non-terminal eng session for the project (steer it directly);
   - the most recent resumable eng session for the project, revived via
     `tools/valor_session.py::resume_session`, which already pushes the steering message before
     transitioning to `pending` specifically to avoid the two-write race;
   - a fresh session.

   Note the constraint discovered above: sessions for the stalled *slug* frequently do not exist at
   all. All three stalled slugs in this report returned **zero** `AgentSession` records — the lane's
   own session was long gone. So "resume the slug's session" is not a viable primary strategy and
   the plan should not assume it.

2. **What does the steer say?** It should be short and actionable — the stalled PR, its issue, how
   long it has been idle, and an instruction to resume the pipeline via the `/sdlc` router. It must
   not be a log line. Per `CLAUDE.md` principle 9, the receiving session drives one stage at a time.

3. **How is the human escalation re-scoped?** Proposal: keep a Telegram path, but fire it **only**
   when the auto-resume attempt itself failed (no steerable session, resume rejected, Redis down),
   and dedupe it once per `(slug, head-sha)` with a long TTL so it cannot ladder. Separate the
   action-attempt cooldown (short, retrying is fine because it is an action) from the
   human-escalation key (long, because a person only needs telling once).

4. **What stops a steer storm?** If the steer lands but the pipeline still does not move — the
   session ignores it, or dies again — the next tick will steer again. The plan needs an attempt
   budget per `(slug, head-sha)` after which the system stops trying and escalates once.

5. **Does gate 5 get fixed here or split out?** Either set `slug` on sessions created by
   `sdlc_session_ensure`, or widen `_has_active_session` to also match `sdlc-local-{issue}` session
   ids / `active_run_id` for the issue. Setting the slug is the cleaner fix but has a wider blast
   radius (other consumers query by slug). This is a pre-requisite for the rest, not optional
   polish.

## Acceptance criteria

- [ ] A stalled SDLC PR that passes all gates results in a steering message delivered to an eng
      session, with no Telegram message to a human on the happy path.
- [ ] Gate 5 correctly suppresses the stall verdict when a local (`sdlc-local-*`) pipeline is
      running for the issue. A live lane is never reported as "no active session".
- [ ] The same stalled `(slug, head-sha)` never produces more than one human-facing message, no
      matter how many 30-minute ticks elapse. The 6-hour ladder in the transcript above is
      reproducible against current `main` and gone afterward.
- [ ] Auto-resume attempts are bounded per `(slug, head-sha)`; on exhaustion the system escalates
      exactly once and stops.
- [ ] Every new external boundary keeps the module's existing failure posture: log a warning and
      continue; never crash the reflection; when state is unknown, decline to act rather than risk a
      duplicate lane.
- [ ] Tests cover: the ladder regression (repeated ticks, one message), the live-local-session
      suppression, the steer-target selection order, and the attempt-budget exhaustion path.
- [ ] `docs/features/pm-session-liveness.md` is updated — the **What this is NOT** section
      currently asserts "Not auto-recovery… Recovery is a human decision after seeing the alert",
      which this issue reverses. Per `CLAUDE.md` principle 1, describe only the new status quo. The
      same section documents the alert target as `Dev: Valor` while `_send_alert`
      (`reflections/sdlc_progress.py:215`) actually sends to `Eng: Valor` — correct that drift too.

## Downstream context

This issue will be consumed by `/do-plan` to produce a plan document at
`docs/plans/sdlc-stall-auto-resume.md`, then executed by `/do-build`.

Related: #1250 (the closed issue that built this detector as notification-only v1).

## Recon Summary

Verified against `main` @ `96b0f65dd` on 2026-08-10.

**Confirmed**

- `_check_project_stalls` composes a string and calls `_send_alert` only — no session write of any kind (`reflections/sdlc_progress.py:300-307`). Module docstring line 9 still declares "Notification-only for v1".
- `_dedup_set` writes `sdlc:stall:alert:{slug}:{sha}` with `SET NX EX` and `_DEFAULT_COOLDOWN_HOURS = 6` (`reflections/sdlc_progress.py:229-242`, `:43`). Key expiry + unchanged sha = the 6h ladder in the transcript. No cap, no separate human-escalation key.
- Gate 5 is `AgentSession.query.filter(slug=slug)` (`reflections/sdlc_progress.py:203`). `tools/sdlc_session_ensure.py:837-843` calls `AgentSession.create_local(...)` with **no** `slug` kwarg. Live query confirms: `sdlc-local-2517`, `sdlc-local-2664`, `sdlc-local-2682`, `sdlc-local-2696` all report `slug=None`, `status=running`, `issue_number` set.
- `_send_alert` targets `Eng: Valor` (`reflections/sdlc_progress.py:215`) while `docs/features/pm-session-liveness.md:276` documents `Dev: Valor`. Doc drift confirmed.
- `docs/features/pm-session-liveness.md:297` asserts "Not auto-recovery… Recovery is a human decision after seeing the alert" — the statement this issue reverses.

**Revised (the issue's solution sketch is wrong on two points)**

1. **Steering `sdlc-local-{N}` is impossible, not merely unhelpful.** Those anchors are created with `is_ledger=True` (`tools/sdlc_session_ensure.py:810`), and `agent/session_executor.py::steer_session` hard-rejects ledger rows (`:841-849`, issue #2495) because no worker ever drains their steering queue. So the steer target must be a real executable eng session — never the lane's own anchor.
2. **Gate 5 must NOT be fixed by matching `issue_number` on the ledger anchor.** All three stalled lanes had a ledger anchor sitting at `status=running` the whole time they were stalled. Matching on it would suppress the stall verdict permanently and blind the detector — strictly worse than today. The correct liveness signal is the issue lock's renewal freshness: `models.session_lifecycle.touch_issue_lock(issue, None, peek=True)` returns `orphaned_lock` computed from `renewed_at` recency (`_lock_renewal_is_fresh`, `ISSUE_LOCK_RENEWAL_FRESHNESS_SECONDS`, issue #2620) rather than from any AgentSession status. Verified live: the four active lanes all return `acquired=False, orphaned_lock=False` with the correct `owner_session_id`. This is the same mirage `_lock_owner_is_live`'s docstring (`models/session_lifecycle.py:982-992`) was written to kill.

**Prior art (patterns to mirror, not invent)**

- `reflections/crash_recovery.py` is already an auto-resume reflection: machine-ownership gate (`_machine_owns_project`, `:99-126`), per-session attempt cap, per-run budget, failed-attempt-still-consumes-budget convergence fix (`:486-503`). Reuse this shape.
- `reflections/stall_advisory.py` is already an act-on-stall reflection: cross-tick consecutive-observation counter, run budget, per-session budget, all as plain Redis keys with TTLs (`:251-417`), plus the `_is_ledger` skip (`:130-139`, PR #2106).
- `tools/valor_session.py::resume_session` (`:786-870`) pushes the steering message **before** the `pending` transition specifically to close the two-write race, and requires `claude_session_uuid`.

**Pre-requisites**

- Gate 5 must be re-grounded on lock-renewal freshness before any auto-resume lands, otherwise the detector either goes blind (ledger match) or duplicates lanes (status quo).

**Dropped**

- "Steer the slug's own session" as a primary strategy — those sessions are ledger rows, and for the three reported slugs no non-ledger session exists at all.

**Existing test coverage**: `tests/unit/reflections/test_sdlc_progress_check.py` (341 lines, 18 tests) covers the current notification-only behavior; several tests assert exactly the semantics this issue changes.

valorengels:
## Summary

Turns `sdlc-progress-check` from a notification-only detector into a self-healing action ladder. A stalled SDLC lane now resumes itself; a human hears about it exactly once per stalled head sha, and only when the system tried to act and could not.

Fixes three stacked defects: no action taken on a stall, a 6-hour alert ladder that re-fired forever against an unchanging sha, and a gate-5 liveness check that queried by `slug` when SDLC anchors carry none.

Plan: `docs/plans/sdlc-stall-auto-resume.md`

## What changed

- **Gate 5' — liveness from the lock, not from session rows.** `_has_active_session` is deleted. `_lock_says_live` does a direct `GET session:issuelock:{N}` classified by `_lock_owner_is_live`, under its own `except -> None`. It deliberately does **not** go through `touch_issue_lock`, which fails open (`acquired=True` on any Redis exception) and cannot express "unknown". Key absent = free; malformed payload or Redis error = unknown, not free. The lock is the **sole** liveness authority: an earlier revision OR-ed in a secondary `AgentSession` row check keyed on `issue_number`, but no non-ledger creation path populates that field (the gap already documented at `tools/merge_predicate.py`), so the branch was structurally dead and was removed along with its knob and tests.
- **Gate 6' — machine ownership.** `machine_owns_project` extracted from `reflections/crash_recovery.py` into `reflections/utilities.py`, with projects.json path resolution fixed to match `load_local_projects`. Two machines can never both act on one lane.
- **Four-rung action ladder**: steer a live eng session, else resume a resumable one, else create one bound to the stalled lane's own slug, else escalate once. Target selection prefers a session whose `slug` matches the stalled lane before falling back to recency: resuming another lane's session would land this issue's commits on that lane's branch. Benign races (another actor resumed the target, or the lane retook its lock) charge no attempt and page nobody.
- **`create_session` extracted** out of `tools/valor_session.py::cmd_create` as a programmatic core, mirroring `resume_session`. `cmd_create` keeps argparse reads, the #1633 parent gate, printing, and exit codes. The latent `parent_id` bypass is closed in the core. Existing `cmd_create` tests re-run unchanged as the regression check.
- **Three Redis keys, three lifetimes** replace the single 6-hour alert key: a 1h action cooldown, a per-`(slug, sha)` attempt budget (read via a bare `GET`, charged via a separate `INCR` so a cooldown skip can never exhaust it), and a 30-day `SET NX` escalation key. The escalation key is also read before each tick claims an action window, so "escalate once and stop acting" is literal rather than "page once, then act forever in silence"; the attempts TTL is floored at the escalation TTL to enforce the same invariant. Every threshold is a named env-overridable constant.
- **Action window is released on no-action ticks** (create brake, target-query unknown, declined pre-create lock re-read) so a braked lane waits one tick, not one cooldown hour.

## Testing

- `tests/unit/reflections/test_sdlc_progress_check.py` rewritten for the ladder (74 test functions, 75 collected items at `f01d2ba8b`), covering unknown-vs-free discrimination, rung selection order, create-rung guards, benign-race classification, budget exhaustion, the read-vs-charge regression, and the escalation-key stop.
- `tests/unit/test_valor_session_create_core.py` — new, incl. the child-session stopgap refusal with zero side effects.
- `tests/integration/test_sdlc_stall_auto_resume_e2e.py` — real `AgentSession` + real Redis, asserting the steer lands on `steering:{session_id}` and reads back via `peek_steering_messages`.
- Measured at `f01d2ba8b`: 108 unit passed (`test_sdlc_progress_check` + `test_crash_recovery_gates` + `test_valor_session_create_core`), 15 integration passed (`test_sdlc_stall_auto_resume_e2e` + `test_crash_auto_resume`). `ruff check` clean.

## Documentation

`docs/features/pm-session-liveness.md` rewritten (gate 5' lock rule, gate 6, the action ladder, the create rung's guards, all seven tunables, `Eng: Valor` target drift corrected). Cross-referenced from `docs/features/session-steering.md`; index entry refreshed.

Closes #2696


valorengels:

Lens: code quality and plan-vs-implementation fidelity. Verified against head `886853f`.

### Verified correct

- **Gate 5' shape** (`reflections/sdlc_progress.py:288-316`) — direct `_get_redis().get(_LOCK_KEY.format(...))` classified by `_lock_owner_is_live`, own `except Exception -> None`. `touch_issue_lock` appears nowhere. Key absent -> `False`; malformed JSON -> `None`. Bytes decode handled.
- **Tiebreak polarity** (`:362-379`) — unknown short-circuits *before* the row check; `lock_live -> True`; else `_nonledger_row_live(...) or False`. Row failure contributes `False`, never `None` (`:339-342`).
- **Read-vs-charge split** — `_attempts_count` is a bare `GET` (`:429`); `INCR` exists only in `_bump_attempts` (`:453`), called from exactly one site (`:747`). The pre-action gate at `:858` uses the read.
- **`create_session` extraction** (`tools/valor_session.py:447-625`) — full boundary moved; `cmd_create` keeps argparse, the #1633 gate, prints, exit codes. `parent_id` bypass closed at `:490-499` before any resolution/worktree/enqueue. `chat_id: str = "0"` at `:455`.
- **All 8 knobs** present as named `_DEFAULT_*` constants with `os.environ.get` overrides and a provisional/tunable comment (`:96-163`). No bare literal thresholds.
- **Legacy sweep clean** — repo-wide grep for `_has_active_session`, `_dedup_set`, `_cooldown_seconds`, `_DEFAULT_COOLDOWN_HOURS`, `SDLC_STALL_COOLDOWN_HOURS`, `sdlc:stall:alert`, `_machine_owns_project`: zero hits outside `docs/plans/`.
- **No forbidden test shapes** — no `touch_issue_lock` mock, no `(None, None)` `_derive_sdlc_metadata` assertion, no steer benign-race branch. The integration test crosses a real Redis boundary.

### Blockers

**File:** `tests/unit/reflections/test_sdlc_progress_check.py:1146`
**Code:** `monkeypatch.setattr(sdlc_progress, "load_local_projects", lambda: [], raising=False)`
**Issue:** `reflections/sdlc_progress.py` never imports `load_local_projects` (confirmed: `grep load_local_projects reflections/sdlc_progress.py` -> no match; line 71 imports only `machine_owns_project, run_per_project_audit`). `raising=False` therefore silently attaches a dead attribute, and `run_per_project_audit` resolves the *real* `load_local_projects` from `reflections.utilities`. The test does not request the `lab` or `fake_redis` fixtures, so `_get_redis()` is production Redis and `steer_session` / `resume_session` / `create_session` / `_send_alert` are all unpatched — while the module's `autouse` `owns_project` fixture (`:248-256`) forces gate 6' open for it. A `pytest` run of this one test scans every real project, shells real `gh pr list` / `git log`, and can reach the action ladder with live dispatch functions.
**Severity:** blocker
**Fix:** Patch the function actually called: `monkeypatch.setattr("reflections.utilities.load_local_projects", lambda: [])`, or patch `sdlc_progress.run_per_project_audit` and assert it was invoked with `_check_project_stalls`. Consider scoping the `autouse` `owns_project` fixture to the ladder tests so a future fixture-escaping test still hits gate 6'.

### Tech Debt

**File:** `tools/valor_session.py:538` (also `:572`, `:597`)
**Code:** `print(f"  Worktree:    {working_dir}", file=sys.stderr)`
**Issue:** Three CLI-shaped `print(..., file=sys.stderr)` calls remain inside `create_session`, contradicting the plan's boundary rule that `cmd_create` keeps the printing. They now fire from the reflection subprocess, emitting CLI-formatted lines into reflection stderr with no `findings`/telemetry counterpart.
**Severity:** tech_debt
**Fix:** Carry them on `CreateResult` (e.g. a `notes: list[str]`) and let `cmd_create` print them.

**File:** `tests/unit/test_crash_recovery_gates.py:100`
**Code:** `patch("tools.reflection_machine_filter._load_project_machines", return_value={"myproj": "my-machine"})`
**Issue:** Every `machine_owns_project` test patches `_load_project_machines`, so the actual delta of Extraction A — swapping the hardcoded `config/projects.json` path for `resolve_projects_config_path()` — has no assertion anywhere. The plan named that path fix as the reason for the extraction.
**Severity:** tech_debt
**Fix:** One test that sets `PROJECTS_CONFIG_PATH` to a tmp file and asserts ownership resolves from it.

### Nits

**File:** PR body / `tests/unit/reflections/test_sdlc_progress_check.py`
**Code:** `- \`tests/unit/reflections/test_sdlc_progress_check.py\` rewritten for the ladder (63 tests)`
**Issue:** The file defines 70 test functions (71 collected items — `test_dispatched_action_keeps_the_action_window` is parametrized x2).
**Severity:** nit
**Fix:** Count, don't estimate.

**File:** `docs/features/pm-session-liveness.md:5`
**Code:** `**Last updated:** 2026-04-30`
**Issue:** Not bumped despite a substantial rewrite of the state-layer section.
**Severity:** nit
**Fix:** Bump the date.

**File:** `tools/valor_session.py:607`
**Code:** `sender_name=f"valor-session ({role})"`
**Issue:** `sender_name` derives from `role` while the `SessionType` enum derives from `session_type`. `cmd_create` always passes both identically; a programmatic caller passing mismatched values gets a record whose `sender_name` disagrees with its `session_type`.
**Severity:** nit
**Fix:** Derive `sender_name` from `session_type`, or assert the two agree.

valorengels:

Lens: runtime risk, failure modes, blast radius. Verified against head `886853f`.

### Verified correct

- Create-rung guard 2 exists and declines on `None` as well as `True` (`reflections/sdlc_progress.py:707-719`).
- The create brake is a real local counter, consulted at `:872` and incremented at `:892-893`.
- Cooldown is `SET NX EX` (`:394`); escalation is `SET NX` (`:469`) — both atomic.
- All three benign-race returns (`:705`, `:719`, `:740`) precede the charge, so a benign race charges nothing.
- A charge failure appends a marker only and cannot escalate (`:894-895` vs `:912`).
- Every unknown-returning gate appends a `gate-unknown:` finding.
- The only raw Redis touched is the `sdlc:stall:*` namespace plus a read-only `GET` on `session:issuelock:{N}` (`:308`) — no Popoto-managed key.
- `run_per_project_audit` wraps `audit_one`, so the reflection cannot raise out of its entry point.

### Blockers

**File:** `reflections/sdlc_progress.py:104-105`
**Code:**
```python
_DEFAULT_ATTEMPTS_TTL_HOURS = 24
_DEFAULT_ESCALATION_TTL_DAYS = 30
```
**Issue:** The plan's Success Criterion is "on exhaustion the system escalates exactly once **and stops acting**", and the Risks table asserts "a create-loop is impossible". Neither holds. `_check_project_stalls:858` re-reads the counter every tick; `_bump_attempts` refreshes a **24h** TTL (`:453-455`) while `_escalation_set` writes a **30-day** key (`:469`). Once the attempts key lapses, `_attempts_count` returns 0 and the ladder acts `SDLC_STALL_RESUME_MAX_ATTEMPTS` more times — but `_escalate_once`'s `SET NX` still finds the escalation key, so nothing is reported. Against an unchanging head sha that is 3 dispatched actions per day, every day, for 30 days, with exactly one human notification. The 24h TTL buys nothing anyway: both keys are already sha-scoped, so any new commit mints fresh keys.
**Severity:** blocker
**Fix:** Make "escalate once, stop" literal — gate the ladder on the presence of the escalation key — or set `_DEFAULT_ATTEMPTS_TTL_HOURS` >= the escalation TTL.

**File:** `reflections/sdlc_progress.py:337`
**Code:** `rows = list(AgentSession.query.filter(issue_number=issue_number))`
**Issue:** Gate 5's secondary liveness signal is structurally inert — it can only ever return `False`. No non-ledger creation path populates `AgentSession.issue_number`: `_push_agent_session` (`agent/agent_session_queue.py:275-304`) has no `issue_number` parameter at all, so the create rung's own sessions will never carry one. Live query against production Redis, project `valor`: 15 sessions, 4 with `issue_number`, **0 non-ledger with `issue_number`** — all four are `sdlc-local-*` anchors that `_is_ledger` skips at `:348`. The repo already documents this exact gap in `tools/merge_predicate.py:44-47` ("`slug` and `issue_number` are populated by disjoint creation paths and 0 live sessions co-populate both"). So `_nonledger_row_live`, the `SDLC_STALL_ROW_LIVENESS_MAX_AGE_MINUTES` knob, and the hollow-row freshness bound all guard a branch that cannot fire, five unit tests (`tests/unit/reflections/test_sdlc_progress_check.py:357-380`) pass only because they inject a row shape production never emits, and `docs/features/pm-session-liveness.md:278` asserts the property as real ("This catches a lane running without a lock"). Gate 5' is a one-signal gate, not two. The create rung's High-severity mitigation is unaffected — all three of its guards are lock-based.
**Severity:** blocker
**Fix:** Either populate `issue_number` on the create path so the signal is real, or delete the dead helper + knob + tests and correct the module docstring, the feature doc, and the tunables table.

### Tech Debt

**File:** `reflections/sdlc_progress.py:587` (selection at `:612`)
**Code:** `rows = list(AgentSession.query.filter(project_key=project_key, session_type="eng"))`
**Issue:** Rung 2 is scoped to project + `eng` only. `resume_session` transitions the row in place, and the worker then runs it in that row's own `working_dir` — `.worktrees/{other-slug}/` on `session/{other-slug}`. The steer text says "invoke /sdlc for issue #A", so work for issue A can land as commits on branch `session/sdlc-B`. Plan Q2 decided project-scoped selection deliberately, and the Risks table rates the *steer* variant Medium, but the resume variant's cross-branch consequence is enumerated nowhere.
**Severity:** tech_debt
**Fix:** Prefer a target whose `slug` matches the stalled lane before falling back to most-recent; at minimum add the row to the Risks table.

**File:** `tools/valor_session.py:534`
**Code:** `if not slug and role != "teammate":`
**Issue:** `role` gates the slug requirement while `session_type` independently selects the `SessionType`. A programmatic caller passing `role="teammate", session_type="eng"` with no slug skips the slug requirement and falls to `working_dir = str(repo_root)` (`:599`) — an ENG session in the main checkout, exactly what #1109/#1272 closed. Zero live callers today, so it is latent — but the plan applied precisely this reasoning to close the latent #1633 bypass in the same function.
**Severity:** tech_debt
**Fix:** Derive one from the other, or validate that they agree.

**File:** `reflections/sdlc_progress.py:436-439`
**Code:**
```python
    except (TypeError, ValueError):
        return 0
```
**Issue:** A corrupt attempts payload reads as "under budget, act". Every other unknown in this module returns `None` and declines. A non-integer value should be indistinguishable from a failed read, not from a fresh key.
**Severity:** tech_debt
**Fix:** Return `None` on a non-integer payload.

**File:** `reflections/sdlc_progress.py:775`
**Code:** `creates_this_tick = 0`
**Issue:** Per-call state, so two overlapping ticks each get a full create budget. The cooldown's `SET NX` covers a single lane, but two ticks racing across *different* lanes can each create one. Low likelihood at a 30-minute cadence; noted because the create brake is the one guard that is not atomic.
**Severity:** tech_debt
**Fix:** Accept and document, or move the brake onto a short-TTL Redis counter.

### Nits

**File:** `reflections/sdlc_progress.py:897`
**Code:** `if outcome.benign: findings.append(...); continue`
**Issue:** The benign branch does not release the action cooldown, unlike the `declined` branch at `:903-905`. On the create rung's pre-create "lock is now live" benign race nothing was dispatched, yet the lane burns a full `SDLC_STALL_RESUME_COOLDOWN_HOURS` window — the one asymmetry against the release rationale documented at `:400-410`.
**Severity:** nit
**Fix:** Release the window on a benign race that dispatched nothing.

**File:** `~/Desktop/Valor/reflections.yaml:238` (vault file, outside the diff)
**Code:** `"State-layer SDLC stall detection: alert when an SDLC PR sits idle"`
**Issue:** The operator-facing description still says "alert". The reflection now steers, resumes, and creates sessions with `SDLC_STALL_RESUME_ENABLED` defaulting to `True`, so it lands acting on merge.
**Severity:** nit
**Fix:** Update the registry description.

valorengels:
## Review: CHANGES REQUESTED

Aggregate of 2 judges (`code-quality`, `risk`), rule `any-blocker-wins`. Both returned CHANGES REQUESTED; no disagreement. Head `886853f04248ef6c4b038dad2e431f45415af7ba` (resolved via `tools/pr_head_resolver.py::resolve_pr_head_sha`). Plan: `docs/plans/sdlc-stall-auto-resume.md`. Closes #2696.

Every blocker below was re-verified by the aggregator by reading the cited code, and two of the three by a live production query. No prior review exists on this PR, so there is no Review Delta.

This is a strong implementation. Gate 5' is built exactly as the round-3 plan specifies (direct `_R.get`, never `touch_issue_lock`), the read-vs-charge split is real, the `create_session` extraction is faithful including the closed `parent_id` bypass, all 8 knobs are named env-overridable constants, and the legacy sweep is clean. The three blockers are convergence and dead-signal problems, not construction problems.

### Blockers

**File:** `reflections/sdlc_progress.py:104-105`
**Code:**
```python
_DEFAULT_ATTEMPTS_TTL_HOURS = 24
_DEFAULT_ESCALATION_TTL_DAYS = 30
```
**Issue:** The ladder does not converge. The Success Criterion is "on exhaustion the system escalates exactly once **and stops acting**"; the Risks table asserts "a create-loop is impossible". Both are false as built. The attempts key carries a 24h TTL refreshed on each `_bump_attempts` (`:453-455`); the escalation key carries 30 days (`:469`). The pre-action gate at `:858` re-reads the counter every tick, so once the attempts key lapses `_attempts_count` returns 0 and the ladder dispatches `SDLC_STALL_RESUME_MAX_ATTEMPTS` more actions — while `_escalate_once`'s `SET NX` still finds the escalation key and stays silent. Against an unchanging head sha that is up to 3 dispatched actions per day, every day, for 30 days, with exactly one human notification. The short TTL buys nothing: both keys are sha-scoped already, so a new commit re-arms everything for free.
**Severity:** blocker
**Fix:** Make "escalate once, stop" literal by gating the ladder on the escalation key's presence, or set `_DEFAULT_ATTEMPTS_TTL_HOURS` >= the escalation TTL.

**File:** `reflections/sdlc_progress.py:337`
**Code:** `rows = list(AgentSession.query.filter(issue_number=issue_number))`
**Issue:** Gate 5's secondary liveness signal can only ever return `False` in production. No non-ledger session-creation path populates `AgentSession.issue_number` — `_push_agent_session` (`agent/agent_session_queue.py:275-304`) has no `issue_number` parameter, so the create rung's own sessions will never carry one either. Live query against production Redis, project `valor`: 15 sessions, 4 carrying `issue_number`, **0 non-ledger**; all four are `sdlc-local-*` anchors that `_is_ledger` skips at `:348`. The repo already documents this exact gap at `tools/merge_predicate.py:44-47` ("`slug` and `issue_number` are populated by disjoint creation paths and 0 live sessions co-populate both"). Consequences: `_nonledger_row_live`, the `SDLC_STALL_ROW_LIVENESS_MAX_AGE_MINUTES` knob and its hollow-row freshness bound are dead; five unit tests (`tests/unit/reflections/test_sdlc_progress_check.py:357-380`) pass only by injecting a row shape production never emits — the same class of unverified-API assertion that produced both round-2 plan BLOCKERs; and `docs/features/pm-session-liveness.md:278` asserts a property the code cannot deliver. Gate 5' is a one-signal gate, not two. The create rung's High-severity mitigation is unaffected: all three of its guards are lock-based.
**Severity:** blocker
**Fix:** Either populate `issue_number` on the create path so the OR-ed signal is real, or delete the helper, the knob, and its tests and correct the module docstring, `docs/features/pm-session-liveness.md`, and the tunables table.

**File:** `tests/unit/reflections/test_sdlc_progress_check.py:1146`
**Code:** `monkeypatch.setattr(sdlc_progress, "load_local_projects", lambda: [], raising=False)`
**Issue:** The monkeypatch targets a name that does not exist on the module — `reflections/sdlc_progress.py:71` imports only `machine_owns_project, run_per_project_audit`, and `grep load_local_projects reflections/sdlc_progress.py` returns nothing. `raising=False` masks that, and `run_per_project_audit` resolves the real `load_local_projects` from `reflections.utilities`. The test requests neither the `lab` nor the `fake_redis` fixture, so `_get_redis()` is production Redis and `steer_session` / `resume_session` / `create_session` / `_send_alert` are unpatched, while the module's `autouse` `owns_project` fixture (`:248-256`) forces gate 6' open for it. A unit-test run of this one test therefore scans every real project on the machine, shells real `gh pr list` / `git log`, and can reach the action ladder with live dispatch functions. Survivable on `main` (v1 was notification-only); not survivable now.
**Severity:** blocker
**Fix:** Patch the function actually called — `monkeypatch.setattr("reflections.utilities.load_local_projects", lambda: [])` — or patch `sdlc_progress.run_per_project_audit` and assert it was invoked with `_check_project_stalls`. Also consider narrowing the `autouse` `owns_project` fixture so a fixture-escaping test still hits gate 6'.

### Tech Debt

**File:** `reflections/sdlc_progress.py:587` (selection at `:612`)
**Code:** `rows = list(AgentSession.query.filter(project_key=project_key, session_type="eng"))`
**Issue:** Rung 2 is scoped to project + `eng` only, so `resume_session` can wake a session whose `working_dir` is `.worktrees/{other-slug}/` on `session/{other-slug}`. The steer text says "invoke /sdlc for issue #A", so work for issue A can land as commits on another lane's branch. Plan Q2 chose project-scoped selection deliberately and the Risks table rates the *steer* variant Medium, but the resume variant's cross-branch consequence is enumerated nowhere.
**Severity:** tech_debt
**Fix:** Prefer a target whose `slug` matches the stalled lane before falling back to most-recently-updated; at minimum add the row to the Risks table.

**File:** `tools/valor_session.py:534`
**Code:** `if not slug and role != "teammate":`
**Issue:** `role` gates the slug requirement while `session_type` independently selects the `SessionType`. A programmatic caller passing `role="teammate", session_type="eng"` with no slug skips the slug requirement and falls to `working_dir = str(repo_root)` (`:599`) — an ENG session in the main checkout, which is what #1109/#1272 closed. Latent (no live caller), but this is the same latent-bypass reasoning the plan applied to close #1633 in this very function.
**Severity:** tech_debt
**Fix:** Derive one from the other, or validate that they agree.

**File:** `reflections/sdlc_progress.py:436-439`
**Code:**
```python
    except (TypeError, ValueError):
        return 0
```
**Issue:** A corrupt attempts payload reads as "under budget, act". Every other unknown in this module returns `None` and declines to act; a non-integer value should be indistinguishable from a failed read, not from a fresh key.
**Severity:** tech_debt
**Fix:** Return `None` on a non-integer payload.

**File:** `tools/valor_session.py:538` (also `:572`, `:597`)
**Code:** `print(f"  Worktree:    {working_dir}", file=sys.stderr)`
**Issue:** Three CLI-shaped `print(..., file=sys.stderr)` calls remain inside `create_session`, contradicting the plan's boundary rule that presentation stays in `cmd_create`. They now fire from the reflection subprocess, emitting CLI-formatted lines into reflection stderr with no `findings`/telemetry counterpart.
**Severity:** tech_debt
**Fix:** Carry them on `CreateResult` (e.g. a `notes: list[str]`) and let `cmd_create` print them.

**File:** `tests/unit/test_crash_recovery_gates.py:100`
**Code:** `patch("tools.reflection_machine_filter._load_project_machines", return_value={"myproj": "my-machine"})`
**Issue:** Every `machine_owns_project` test patches `_load_project_machines`, so the actual delta of Extraction A — swapping a hardcoded `config/projects.json` path for `resolve_projects_config_path()` — has no assertion anywhere. The plan named that path fix as the reason for the extraction.
**Severity:** tech_debt
**Fix:** Add one test that sets `PROJECTS_CONFIG_PATH` to a tmp file and asserts ownership resolves from it.

**File:** `reflections/sdlc_progress.py:775`
**Code:** `creates_this_tick = 0`
**Issue:** Per-call state, so two overlapping ticks each get a full create budget. The cooldown `SET NX` covers a single lane, but two ticks racing across *different* lanes can each create one. Low likelihood at a 30-minute cadence; noted because the create brake is the one guard that is not atomic.
**Severity:** tech_debt
**Fix:** Accept and document in the Risks table, or move the brake onto a short-TTL Redis counter.

### Nits

**File:** `reflections/sdlc_progress.py:897`
**Code:** `if outcome.benign: findings.append(f"benign-race: {kind} {slug}"); continue`
**Issue:** The benign branch does not release the action cooldown, unlike the `declined` branch at `:903-905`. On the create rung's pre-create "lock is now live" benign race nothing was dispatched, yet the lane burns a full `SDLC_STALL_RESUME_COOLDOWN_HOURS` window — the one asymmetry against the release rationale documented at `:400-410`.
**Severity:** nit
**Fix:** Release the window on a benign race that dispatched nothing.

**File:** PR body
**Code:** `` `tests/unit/reflections/test_sdlc_progress_check.py` rewritten for the ladder (63 tests) ``
**Issue:** The file defines 70 test functions (71 collected items — `test_dispatched_action_keeps_the_action_window` is parametrized x2).
**Severity:** nit
**Fix:** Count, don't estimate.

**File:** `docs/features/pm-session-liveness.md:5`
**Code:** `**Last updated:** 2026-04-30`
**Issue:** Not bumped despite a substantial rewrite of the state-layer section.
**Severity:** nit
**Fix:** Bump the date.

**File:** `tools/valor_session.py:607`
**Code:** `sender_name=f"valor-session ({role})"`
**Issue:** `sender_name` derives from `role` while the `SessionType` enum derives from `session_type`; a programmatic caller passing mismatched values gets a record whose `sender_name` disagrees with its `session_type`.
**Severity:** nit
**Fix:** Derive `sender_name` from `session_type`, or assert the two agree.

**File:** `~/Desktop/Valor/reflections.yaml:238` (vault file, outside the diff)
**Code:** `"State-layer SDLC stall detection: alert when an SDLC PR sits idle"`
**Issue:** The operator-facing registry description still says "alert". The reflection now steers, resumes, and creates sessions, with `SDLC_STALL_RESUME_ENABLED` defaulting to `True`, so it lands acting on merge.
**Severity:** nit
**Fix:** Update the registry description.

### Miscellaneous
- None

- None — the PR body declares no deferrals, follow-ups, or out-of-scope exclusions.

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — FAIL — *"escalates exactly once and stops acting" is not delivered (blocker 1); the Gate 5' secondary signal is specified and built but inert (blocker 2).*
- **2. No-Gos from plan — none violated** — PASS — *No steering of ledger anchors, nothing killed, no `slug` set on SDLC anchors, gates 1-4 untouched, threshold unchanged, creation bounded to `sdlc-{issue}` with the machine-ownership and lock guards.*
- **3. New `except Exception` blocks — each has logger/raise/swallow-ok** — PASS — *Every new boundary logs a warning and returns a declining value; polarity audited helper by helper.*
- **4. New integration tests — exercise serialization boundary** — PASS — *`tests/integration/test_sdlc_stall_auto_resume_e2e.py` uses a real `AgentSession` + real Redis and reads back via `peek_steering_messages`.*
- **5. Plan internal consistency — spike findings match task steps** — PASS — *Spike-3's round-3 correction (direct `_R.get`, never `touch_issue_lock`) is implemented as written.*
- **6. No hardcoded secrets or debug artifacts** — FAIL — *Three CLI `print(..., file=sys.stderr)` calls left inside the programmatic core (`tools/valor_session.py:538/572/597`); tech debt, not a security issue.*
- **7. New public APIs — docstrings present** — PASS — *`create_session`, `CreateResult`, `machine_owns_project`, and every new reflection helper are documented.*
- **8. Breaking changes — migration path documented** — N/A — *No schema, config, or CLI surface change; `cmd_create`'s contract is preserved and its tests re-run unchanged.*
- **9. Tests added for new behavior** — PASS — *70 tests across the ladder, gate polarity, budgets, and race classification.*
- **10. Tests cover the failure path (not just happy path)** — FAIL — *Five gate-5' secondary-row tests assert a row shape no production path emits (blocker 2), and the entrypoint smoke test has no fences at all (blocker 3).*
- **11. UI changes (if any) — screenshot captured** — N/A — *No HTML/CSS/JS/template files in the diff.*
- **12. Docs updated for user-facing changes** — FAIL — *`docs/features/pm-session-liveness.md:278` asserts the secondary row check "catches a lane running without a lock", which it cannot; stale `Last updated`.*

## Rubric

- [ ] **1. Plan vs. implementation match** — fail — *Convergence criterion unmet; secondary liveness signal inert.*
- [ ] **2. New code quality** — fail — *Dead helper + dead knob; CLI prints inside the programmatic core.*
- [ ] **3. Test coverage** — fail — *Unfenced entrypoint test; five tests over an unreachable branch; Extraction A's path fix untested.*
- [x] **4. Regression risk to existing callers** — pass — *Legacy sweep clean; `cmd_create` tests re-run unchanged; `crash_recovery` import swept.*
- [x] **5. Data integrity** — n/a — *No schema change, no migration, no Popoto-managed key touched raw.*
- [x] **6. Security** — pass — *No secrets, no eval/exec, no request-derived subprocess input; the lock read is a read-only `GET`.*
- [ ] **7. Documentation accuracy** — fail — *Feature doc asserts a property the code cannot deliver; stale `Last updated`.*
- [ ] **8. PR body accuracy** — fail — *"63 tests" vs 70 defined. Independently, the body's green counts ("246 unit, 4 integration, 184 regression") were not reproduced by this review and are recorded as unverified rather than credited.*
- [x] **9. Disclosed deferrals** — n/a — *None declared.*
- [x] **10. Follow-up claims verified** — n/a — *No follow-up claims in the body.*

**Verdict derivation:** rule 1 — critical items 1, 2, and 3 are `fail` with no matching acknowledged disclosure -> **CHANGES REQUESTED — Blocker**.

valorengels:
## Review: CHANGES REQUESTED

Re-review after patch cycle 1. Head `f01d2ba8bc295d9a634ba717f2c54c186be43bd6` (resolved via `tools/pr_head_resolver.py::resolve_pr_head_sha`, not a bare `gh` read). Prior review was at `886853f04`. Plan: `docs/plans/sdlc-stall-auto-resume.md`, read from `main` at `a68139826` (the branch intentionally carries no plan edit). Closes #2696.

**All three prior blockers are closed, verified by reading the code and re-running the suites in this environment.** Two `nit`-severity PR-body inaccuracies remain, both introduced by the patch cycle itself. There are no blockers and no tech debt.

### Review Delta (vs. `886853f04`)

**Blocker 1 — ladder non-convergence (attempts TTL vs. escalation TTL). CLOSED.**
Fixed two ways, either of which alone would have sufficed:
- `_escalation_exists(slug, sha)` (`reflections/sdlc_progress.py:427-444`) is a read-only companion to `_escalation_set`, and `_check_project_stalls` consults it at `:837-843` — after gate 5', **before** the attempts read and before `_action_cooldown_set` claims the window. An existing key returns `already-escalated: {slug}` and skips the lane entirely; an unreadable key returns `None` and also declines (`gate-unknown: escalation-read`). "Escalate once and stop acting" is now literal, not "page once and act forever behind a silent `SET NX`".
- `_attempts_ttl_seconds()` (`:143-152`) is now `max(configured, _escalation_ttl_seconds())`, with the knob renamed `SDLC_STALL_ATTEMPTS_TTL_HOURS` -> `SDLC_STALL_ATTEMPTS_TTL_DAYS` (default 30). The invariant is stated as a comment on the constant (`:105-114`) and asserted directly by `test_attempts_ttl_is_never_shorter_than_the_escalation_ttl`.

Behavioral proof exists, not just a unit assertion: `test_lapsed_attempts_key_cannot_re_arm_the_budget_under_a_live_escalation` exhausts the budget, escalates once, advances the fake clock 48h past the *old* 24h attempts TTL, ticks three more times, and asserts `len(lab.steers)` is unchanged and `len(lab.alerts) == 1`. I traced the key lifetimes myself: both keys are sha-scoped, `_escalation_set` is `SET NX` (never refreshed), `_bump_attempts` refreshes on each charge, so attempts can only outlive escalation — and the new gate makes the ordering moot regardless. After the 30-day horizon both lapse together and the lane re-arms for a bounded 3 actions plus one page, which is the intended behavior for a sha stuck a month.

**Blocker 2 — inert `_nonledger_row_live` secondary signal. CLOSED by deletion.**
The helper, the `SDLC_STALL_ROW_LIVENESS_MAX_AGE_MINUTES` knob, `_row_liveness_max_age_seconds()`, and all five tests over the unreachable branch are gone. `_lane_is_live` (`:327-340`) is now `return _lock_says_live(issue_number)` with a docstring recording *why* the branch was dead (no non-ledger creation path populates `AgentSession.issue_number`) and telling a future author what to plumb first. The `_FakeQuery.by_issue` route and the `_Row.issue_number` attribute were removed from the harness too, so the row shape production never emits can no longer be injected; `_Row` now carries `slug`, which the code really does read.

`test_lane_is_live_unknown_lock_is_unknown_and_queries_nothing` asserts `fake_query.calls == []` — the gate consults no session rows at all, which is the strongest form of the fix. I swept for residue: `grep -rn "_nonledger_row_live\|ROW_LIVENESS\|ATTEMPTS_TTL_HOURS\|_has_active_session"` across `*.py` and `*.md` (excluding `docs/plans/`, which legitimately records the history) returns nothing. Module docstring, `docs/features/pm-session-liveness.md` (gate 5' prose, ladder steps, failure-semantics bullet, tunables table), and the plan on `main` were all corrected consistently.

**Blocker 3 — unit test reaching production Redis and live dispatch functions. CLOSED.**
Two independent fixes, which is the right shape for a fence:
- `test_run_sdlc_progress_check_uses_per_project_helper` now patches `reflections.utilities.load_local_projects` — the name `run_per_project_audit` actually resolves — instead of a non-existent `sdlc_progress.load_local_projects` under `raising=False`. It also takes the `lab` fixture and asserts `(lab.steers, lab.resumes, lab.creates, lab.alerts) == ([], [], [], [])`.
- The `owns_project` fixture is no longer `autouse`; it is pulled in by `lab` (`tests/unit/reflections/test_sdlc_progress_check.py:258`). A future test that reaches `_check_project_stalls` without the `lab` fence now hits the real gate 6' and stops there instead of scanning real projects with live dispatch functions attached. The docstring says exactly that. Removing `autouse` broke nothing: the full reflections suite passes.

### Tech debt from the prior review — all six addressed

| Prior finding | Disposition (verified) |
|---|---|
| Rung 2 can resume another lane's session | `_pick_steer_target(project_key, lane_slug=slug)`; `_rank` is `(same_lane, recency)` applied inside **both** the live and resumable buckets (`:584-606`), threaded from the ladder at `:878`. Three unit tests plus an end-to-end ladder test. Risks-table row added on `main`. |
| `role`/`session_type` bypass in `create_session` | Refused up front at `tools/valor_session.py:508-527`, before any resolution or filesystem work; `test_role_and_session_type_must_agree` asserts the `resolve`/`worktree`/`push` spies are all empty. Both production callers (`cmd_create` passes `session_type=role`; `sdlc_progress` passes `eng`/`eng`) already agree, so this is latent-only, as the comment says. |
| Corrupt attempts payload read as "under budget" | `_attempts_count` returns `None` on a non-integer payload and logs (`:401-405`); the ladder maps `None` to `gate-unknown: attempts-read` and declines **without** escalating (`:862-866`). |
| CLI `print(..., file=sys.stderr)` inside the programmatic core | Replaced by `CreateResult.notes: list[str]`; `cmd_create` prints them at `:753-755`. `test_progress_notes_are_returned_not_printed` asserts `capsys.readouterr().err == ""`. `notes` is threaded onto every early-return refusal path and the outer `except`, so a failure still carries what got as far as happening. |
| `machine_owns_project` path-resolution delta untested | `test_ownership_resolves_from_the_projects_config_path_override` writes a real `projects.json` to `tmp_path`, points `PROJECTS_CONFIG_PATH` at it, and asserts both the owned and the non-owned answer end to end — no `_load_project_machines` patch. |
| `creates_this_tick` non-atomic across ticks | Accepted and documented in both places: a five-line comment at `:761-766` and a Risks-table row on `main` marked "Accepted, not mitigated". |

Nit 1 (benign-race cooldown asymmetry) is also fixed: `:907-915` releases the window only when `not outcome.dispatched`, with two tests pinning both polarities (`test_predispatch_benign_race_releases_the_action_window`, `test_dispatched_benign_race_keeps_the_action_window`). The `reflections.yaml` registry description in the vault now reads "State-layer SDLC stall auto-resume: ... steer ... else resume ... else create ...". `Last updated` bumped to 2026-08-10.

### Blockers
- None

### Tech Debt
- None

### Nits

**File:** PR body, "Documentation" section
**Code:** `` `docs/features/pm-session-liveness.md` rewritten (... all eight tunables ...) ``
**Issue:** The tunables table in that file lists **seven** rows, and the module docstring's `Configuration` block lists the same seven. The count was correct before `aa5325884` deleted `SDLC_STALL_ROW_LIVENESS_MAX_AGE_MINUTES`; the patch cycle made it stale. I counted the table rather than trusting the number.
**Severity:** nit
**Fix:** "all seven tunables".

**File:** PR body, "What changed" -> "Four-rung action ladder"
**Code:** `steer a live eng session, else resume the most recent resumable one, else create one ...`
**Issue:** Superseded by the same-lane preference added in `aa5325884`. `_pick_steer_target` ranks `(same_lane, recency)`, so it resumes the most recent *same-lane* session and falls back to plain recency only when no candidate carries the lane's slug. Not cosmetic: the reason for the change is that resuming another lane's session lands this issue's commits on that lane's branch, and the merge-commit narrative should say so.
**Severity:** nit
**Fix:** "else resume a resumable one, preferring the stalled lane's own session over a merely-more-recent one".

### Miscellaneous
- None

- None declared in the PR body.

## Verification performed in this environment

Per "a number the PR claims is a claim, not evidence", every number in the PR body was reproduced here, at `f01d2ba8b`, in the PR worktree:

- `74` test functions / `75` collected in `tests/unit/reflections/test_sdlc_progress_check.py` — matches (`grep -c` and `--collect-only`).
- `108 unit passed` (`test_sdlc_progress_check` + `test_crash_recovery_gates` + `test_valor_session_create_core`) — reproduced exactly: **108 passed**.
- `15 integration passed` (`test_sdlc_stall_auto_resume_e2e` + `test_crash_auto_resume`) — reproduced exactly: **15 passed**.
- Regression sweep beyond the body's claim: `tests/unit/reflections/` plus every `tests/unit/test_valor_session_*.py` — **359 passed**.
- `python -m ruff check .` — **All checks passed**. `ruff format --check` on all five touched files — **already formatted**.
- `tests/integration/test_stall_advisory_e2e.py` shows 6 failures on this branch. I ran it on `main` and got the **identical 6 failures**; it is untouched by this diff (`git diff main...HEAD --name-only | grep stall_advisory` -> 0). Pre-existing, not a regression from this PR, and not a finding against it.
- Plan `## Verification` table: the plan has no `## Verification` section, so there was nothing to run. All four hook-required sections (`Documentation`, `Update System`, `Agent Integration`, `Test Impact`) are present.
- Documentation gate: `docs/features/pm-session-liveness.md` exists and was rewritten as the plan's `## Documentation` section specifies.
- Mergeability preflight: `MERGEABLE` / `CLEAN` / `OPEN`.

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — PASS — *Every criterion re-checked against the current code. "Escalates exactly once and stops acting" is now delivered by the escalation-key read; gate 5' is one honest signal; creation is guarded by two lock reads plus machine ownership plus the per-tick brake.*
- **2. No-Gos from plan — none violated** — PASS — *No ledger anchors steered, nothing killed, no `slug` set on SDLC anchors, gates 1-4 untouched, threshold unchanged.*
- **3. New `except Exception` blocks — each has logger/raise/swallow-ok** — PASS — *`_escalation_exists` logs a warning and returns `None`; the non-integer branch of `_attempts_count` now logs before returning `None`.*
- **4. New integration tests — exercise serialization boundary** — PASS — *`test_sdlc_stall_auto_resume_e2e.py` unchanged from the prior review and still green: real `AgentSession`, real Redis, read back via `peek_steering_messages`.*
- **5. Plan internal consistency — spike findings match task steps** — PASS — *Plan synced on `main` at `a68139826`; Redis-state table, Budgets table, tunables list, and two Risks rows all match the code.*
- **6. No hardcoded secrets or debug artifacts** — PASS — *The three stray CLI prints are gone from the core.*
- **7. New public APIs — docstrings present** — PASS — *`_escalation_exists`, the `lane_slug` parameter, `CreateResult.notes`, and the coherence check are all documented with rationale.*
- **8. Breaking changes — migration path documented** — N/A — *`SDLC_STALL_ATTEMPTS_TTL_HOURS` -> `_DAYS` renames an unshipped knob (v1 never had it); no schema, config, or CLI surface change.*
- **9. Tests added for new behavior** — PASS — *Five new ladder tests for the escalation stop and the corrupt payload, four for the same-lane preference, two for benign-race cooldown polarity, three for the `valor_session` changes, one for the path override.*
- **10. Tests cover the failure path (not just happy path)** — PASS — *Escalation-read failure, corrupt payload, lapsed-budget re-arm, and the zero-side-effect mismatch refusal are all asserted as "no crash AND no action".*
- **11. UI changes (if any) — screenshot captured** — N/A — *No HTML/CSS/JS/template files in the diff.*
- **12. Docs updated for user-facing changes** — PASS — *Feature doc no longer asserts a property the code cannot deliver; `Last updated` bumped; vault registry description corrected.*

## Rubric

- [x] **1. Plan vs. implementation match** — pass — *Convergence criterion delivered; gate 5' matches the synced plan's single-signal design.*
- [x] **2. New code quality** — pass — *Dead helper and dead knob removed; presentation moved out of the core; the new comments explain why, not what.*
- [x] **3. Test coverage** — pass — *Entrypoint test fenced and the `autouse` escape hatch closed; the five tests over the unreachable branch are gone rather than rewritten; Extraction A's path fix now has a real end-to-end assertion.*
- [x] **4. Regression risk to existing callers** — pass — *359 unit + 15 integration green; both `create_session` production callers pass matching `role`/`session_type`; legacy sweep for removed names is clean.*
- [x] **5. Data integrity** — n/a — *No schema change, no migration, no Popoto-managed key touched raw.*
- [x] **6. Security** — pass — *No secrets, no eval/exec, no request-derived subprocess input; the new Redis call is a read-only `GET`.*
- [x] **7. Documentation accuracy** — pass — *Feature doc, module docstring, plan, and vault registry all describe only the new status quo.*
- [ ] **8. PR body accuracy** — fail — *Two stale sentences, both nit-severity: "all eight tunables" (seven) and "resume the most recent resumable one" (same-lane-preferred). Every number in the body was independently reproduced and matches.*
- [x] **9. Disclosed deferrals** — n/a — *None declared.*
- [x] **10. Follow-up claims verified** — n/a — *No follow-up claims in the body.*

**Verdict derivation:** no critical rubric item is `fail`; item 8 is `fail` with two `nit` findings, zero blockers, zero tech debt. Approval is reserved for zero-finding reviews, so this is **CHANGES REQUESTED — Nits only**. Both fixes are PR-body edits; no code change is required.

valorengels:
## Review: APPROVED

Re-review after the PR-body fix. Head `f01d2ba8bc295d9a634ba717f2c54c186be43bd6`, re-resolved via `tools/pr_head_resolver.py::resolve_pr_head_sha` — **unchanged** from the previous round, so no code moved and the full code review at [#issuecomment-5237169792](https://github.com/tomcounsell/ai/pull/2710#issuecomment-5237169792) stands in its entirety: the three blockers from `886853f04` are closed, all six tech-debt items and the cooldown-asymmetry nit are addressed, and the suites are green. Plan: `docs/plans/sdlc-stall-auto-resume.md` on `main` at `a68139826`. Closes #2696.

Both remaining nits are fixed, verified against the code rather than taken on report:

- **"all eight tunables" -> "all seven tunables".** `grep -o "SDLC_STALL[A-Z_]*" reflections/sdlc_progress.py | sort -u` and the same grep over `docs/features/pm-session-liveness.md` each return the identical **seven** names: `THRESHOLD_HOURS`, `RESUME_ENABLED`, `RESUME_MAX_ATTEMPTS`, `CREATE_MAX_PER_TICK`, `RESUME_COOLDOWN_HOURS`, `ESCALATION_TTL_DAYS`, `ATTEMPTS_TTL_DAYS`. Code and doc agree on both the count and the set — the doc documents every knob the module reads, and no others.
- **Ladder bullet.** Now reads "else resume a resumable one" plus the same-lane sentence naming the cross-branch consequence. That matches `_pick_steer_target`, whose `_rank` is `(same_lane, recency)` applied inside **both** the live and the resumable bucket (`reflections/sdlc_progress.py:584-606`), with `lane_slug` threaded from the ladder at `:878` and recency deciding only when no candidate carries the lane's slug.

I diffed the body against the copy I captured last round: exactly those two lines changed (plus one trailing newline). Nothing else drifted, so nothing else needed re-verification.

Mergeability re-checked at this head: `MERGEABLE` / `CLEAN` / `OPEN`.

### Blockers
- None

### Tech Debt
- None

### Nits
- None

### Miscellaneous
- None

- None declared in the PR body.

## Rubric

- [x] **1. Plan vs. implementation match** — pass — *Unchanged head; convergence criterion delivered, gate 5' matches the synced plan's single-signal design.*
- [x] **2. New code quality** — pass — *Unchanged head.*
- [x] **3. Test coverage** — pass — *Unchanged head; 108 unit / 15 integration / 359 regression reproduced last round at this same sha.*
- [x] **4. Regression risk to existing callers** — pass — *Unchanged head.*
- [x] **5. Data integrity** — n/a — *No schema change, no migration, no Popoto-managed key touched raw.*
- [x] **6. Security** — pass — *Unchanged head.*
- [x] **7. Documentation accuracy** — pass — *Unchanged head; feature doc, module docstring, plan, and vault registry describe only the new status quo.*
- [x] **8. PR body accuracy** — pass — *Both stale sentences corrected; the seven-knob claim verified by grepping code and doc independently; every number in the body was reproduced in this environment last round at this same sha.*
- [x] **9. Disclosed deferrals** — n/a — *None declared.*
- [x] **10. Follow-up claims verified** — n/a — *No follow-up claims in the body.*

**Verdict derivation:** zero blockers, zero tech debt, zero nits; no rubric item is `fail` -> **APPROVED**.

valorengels:
## Review: CHANGES REQUESTED

Re-review at head `0b577b20230a70284652a9d588095066fd56874f` (resolved via `tools/pr_head_resolver.py::resolve_pr_head_sha`, not a bare `gh` read). Previous verdict was APPROVED at `f01d2ba8b`; this round re-opens on one `tech_debt` finding that the new commit's own subject matter surfaced. Plan: `docs/plans/sdlc-stall-auto-resume.md` on `main` at `a68139826`. Closes #2696.

### Review Delta (vs. `f01d2ba8b`)

One commit, `0b577b202`, one file, `docs/features/pm-session-liveness.md`, +7/-6. I confirmed with `git diff f01d2ba8b 0b577b202 --name-only` that **no `.py` file changed**, so every code finding from the previous two rounds carries over against byte-identical code and needed no re-derivation.

**The change itself is correct.** I verified it against the function body, not against the docstring: in `_check_project_stalls`, `if not resume_enabled: _escalate("auto-resume disabled (SDLC_STALL_RESUME_ENABLED=false)", 0); continue` sits at `:860-862`, between the `_escalation_exists` read (`:837-843`) and the `_attempts_count` read (`:864`). The doc now inserts exactly that as step 2 and renumbers 3-8 consistently; the surrounding steps kept their text. The new step's claim that the lane "escalates once ... and stops, same as any other terminal ladder outcome" is accurate — `_escalate` routes through `_escalate_once`'s `SET NX`, and step 1's already-escalated read makes every later tick skip the lane.

I re-ran the reflections suite at this head (**237 passed**) and `ruff check .` (**All checks passed**) rather than assuming a docs-only commit is inert.

### Blockers
- None

### Tech Debt

**File:** `reflections/sdlc_progress.py:21-28` (module docstring, "Then the action ladder" block)
**Code:**
```
    escalation key already set                 -> stop acting entirely
    attempts >= SDLC_STALL_RESUME_MAX_ATTEMPTS -> escalate once, stop
    action cooldown live                       -> skip this tick
    rung 1  live non-ledger eng session        -> steer_session(...)
    rung 2  most recent resumable eng session  -> resume_session(...)
```
**Issue:** The action-ladder sequence exists in exactly **two** places — this docstring and the `### The action ladder` list in `docs/features/pm-session-liveness.md`. I grepped for other copies: `docs/features/session-steering.md` mentions rung 1 only in passing (still accurate), and the `rung` hits in `tools/_sdlc_utils.py` / `tools/sdlc_next_skill.py` are an unrelated repo-resolution ladder. This commit fixed one of the two copies, and the docstring is now stale in two ways at once:

1. **The break-glass step is missing** — the exact omission `0b577b202` was written to correct. The knob's *semantics* are documented lower down in the `Configuration` block ("false = notification-only, still escalate-once"), so the file is not silent about the feature; the ordered ladder listing, which is what a reader consults for sequence, is.
2. **`rung 2  most recent resumable eng session` is superseded.** `_pick_steer_target`'s `_rank` is `(same_lane, recency)` applied inside both buckets (`:584-606`), so recency decides only when no candidate carries the lane's slug. This is the *same sentence* flagged in the PR body at `f01d2ba8b`; it was corrected in the PR body and in the feature doc, but the copy nearest the code was not swept. It misstates a safety-relevant rule — the whole reason for the preference is that resuming another lane's session lands this issue's commits on that lane's branch.

A replicated-text defect closes out on a sweep of every copy, not on fixing the copy that prompted the report. Item 2 in particular is a re-occurrence of an already-reported finding at a location the earlier fix did not reach.
**Severity:** tech_debt
**Fix:** In the docstring's ladder block, add the break-glass line between the escalation-key line and the attempts line, and change rung 2 to name the same-lane preference (e.g. `rung 2  resumable eng session, same-lane first -> resume_session(...)`).

### Nits
- None

### Miscellaneous
- None

- None. The PR body declares no deferrals, follow-ups, or out-of-scope exclusions, so the docstring gap is an open finding rather than an accepted one.

## Verification performed in this environment

- Head re-resolved via `tools/pr_head_resolver.py::resolve_pr_head_sha` -> `0b577b20230a70284652a9d588095066fd56874f`; worktree HEAD matches.
- `git diff f01d2ba8b 0b577b202 --name-only` -> `docs/features/pm-session-liveness.md` only. No code changed.
- Renumbering audited line by line: steps 3-8 are the previous 2-7 verbatim; no step's text silently drifted while the numbers moved.
- `tests/unit/reflections/` at this head — **237 passed**. `python -m ruff check .` — **All checks passed**.
- Stale-reference sweep for the retired names (`_nonledger_row_live`, `SDLC_STALL_ROW_LIVENESS_MAX_AGE_MINUTES`, `_row_liveness_max_age_seconds`, `_has_active_session`) across `*.py` and `*.md`: every hit is in `docs/plans/`, where it is the intentional narration of why the signal was deleted. Clean everywhere else — independently reproduced, not taken on report.
- PR body unchanged since `f01d2ba8b` (byte-identical; hash `e0a86ee4...`), so the seven-tunables and ladder-bullet corrections verified last round still stand. Note the body's ladder bullet is now the one summary of the four rungs that is *not* stale.
- Mergeability preflight: first read returned `UNKNOWN`; retried after 2s -> `MERGEABLE` / `CLEAN` / `OPEN`.

## Rubric

- [x] **1. Plan vs. implementation match** — pass — *Code byte-identical to the APPROVED head.*
- [x] **2. New code quality** — pass — *No code changed.*
- [x] **3. Test coverage** — pass — *237 reflections tests green at this head; the suites verified at `f01d2ba8b` run over identical code.*
- [x] **4. Regression risk to existing callers** — pass — *Docs-only diff.*
- [x] **5. Data integrity** — n/a — *No schema change, no migration, no Popoto-managed key touched raw.*
- [x] **6. Security** — pass — *No code changed.*
- [ ] **7. Documentation accuracy** — fail — *The module docstring's ladder block, the second of the two copies of that sequence, omits the break-glass step this commit added to the other copy and still describes rung 2 as recency-only.*
- [x] **8. PR body accuracy** — pass — *Unchanged and verified last round; every number in it was reproduced in this environment.*
- [x] **9. Disclosed deferrals** — n/a — *None declared.*
- [x] **10. Follow-up claims verified** — n/a — *No follow-up claims in the body.*

**Verdict derivation:** zero blockers; rubric item 7 is `fail` with one `tech_debt` finding and no matching acknowledged disclosure -> **CHANGES REQUESTED**. The fix is two lines in one docstring.

valorengels:
## Review: APPROVED

Re-review at head `81ac764dff08b74373899eb52b04e11685702365` (resolved via `tools/pr_head_resolver.py::resolve_pr_head_sha`). Plan: `docs/plans/sdlc-stall-auto-resume.md` on `main` at `a68139826`. Closes #2696.

The tech-debt finding from `0b577b202` is closed, and closed as a sweep rather than a point patch. No findings remain at any severity.

### Review Delta (vs. `0b577b202`)

One commit, `81ac764df`, one file, `reflections/sdlc_progress.py`, +7/-1. I confirmed the change is confined to the module docstring by parsing the file rather than eyeballing the diff — `ast.get_docstring(ast.parse(src))` contains the new break-glass line, the rewritten rung-2 line, and the new ranking paragraph, so every added line is inside the triple-quoted block and no executable code moved. All code findings from the previous rounds therefore carry over against byte-identical code.

**1. Break-glass step added**, positioned between the escalation-key line and the attempts line — matching the function body, where `if not resume_enabled: _escalate(...)` sits at `:860-862` between the `_escalation_exists` read and the `_attempts_count` read.

**2. Rung 2 no longer claims recency-only selection.** It now reads `rung 2  resumable eng session -> resume_session(...)`.

**3. The deviation from my suggested fix is correct, and better than what I proposed — overruling myself, not the author.** I suggested naming the same-lane preference on the rung-2 line. That would have been wrong: `_rank` is defined once and applied as the key to **both** `max(live, ...)` and `max(resumable, ...)` (`:590-606`), so attaching the note to rung 2 alone would have implied rung 1 ranks purely by recency — a fresh inaccuracy in place of the old one. The commit instead keeps the rung list terse and adds a following paragraph stating that rungs 1 and 2 both rank `(same_lane, recency)`, with the reason attached. That is the accurate reading of the code and the right shape for the fix. This is exactly the case where a reviewer's proposed remedy should lose to the code.

I checked the new paragraph's precision too: "a session carrying the stalled lane's own slug always wins" is scoped by the preceding clause to within-rung ranking, and the rung list immediately above already establishes that rung 1 precedes rung 2 wholesale — so it cannot be misread as a same-lane *resumable* session outranking an other-lane *live* one, which is not what the code does.

**Third-copy sweep, reproduced independently rather than taken on report.** `grep -rn "most recent resumable\|most-recently-updated\|most recently updated"` over `*.py` and `*.md`, excluding `docs/plans/`, returns exactly three sites, all correct: `reflections/sdlc_progress.py:574` (`_pick_steer_target`'s own docstring — "Within BOTH buckets..."), `:591` (`_rank`'s docstring), and `docs/features/pm-session-liveness.md:292` ("Both rungs prefer..."). With the module docstring now fixed, all four statements of the selection rule in the repo agree with each other and with `_rank`. The `rung` hits in `tools/_sdlc_utils.py` and `tools/sdlc_next_skill.py` remain the unrelated repo-resolution ladder.

### Blockers
- None

### Tech Debt
- None

### Nits
- None

### Miscellaneous
- None

- None declared in the PR body.

## Verification performed in this environment

- Head re-resolved via `tools/pr_head_resolver.py::resolve_pr_head_sha` -> `81ac764dff08b74373899eb52b04e11685702365`.
- `git diff 0b577b202 81ac764df --stat` -> `reflections/sdlc_progress.py` only, +7/-1; AST check confirms docstring-only.
- `tests/unit/reflections/` at this head — **237 passed**.
- `python -m ruff check .` — **All checks passed**. `ruff format --check reflections/sdlc_progress.py` — **already formatted**.
- Selection-rule sweep across `*.py` and `*.md` outside `docs/plans/` — four statements, all mutually consistent and all matching `_rank`.
- PR body byte-identical since `f01d2ba8b` (hash `e0a86ee4...`), so the corrections verified two rounds ago still stand.
- Mergeability preflight: `MERGEABLE` / `CLEAN` / `OPEN`.

## Rubric

- [x] **1. Plan vs. implementation match** — pass — *Executable code byte-identical to the head approved at `f01d2ba8b`.*
- [x] **2. New code quality** — pass — *No executable code changed.*
- [x] **3. Test coverage** — pass — *237 reflections tests green at this head; the full suites verified at `f01d2ba8b` run over identical code.*
- [x] **4. Regression risk to existing callers** — pass — *Docstring-only diff.*
- [x] **5. Data integrity** — n/a — *No schema change, no migration, no Popoto-managed key touched raw.*
- [x] **6. Security** — pass — *No executable code changed.*
- [x] **7. Documentation accuracy** — pass — *Both copies of the action-ladder sequence now carry the break-glass step, and all four statements of the same-lane selection rule agree with `_rank`.*
- [x] **8. PR body accuracy** — pass — *Unchanged and verified; every number in it was reproduced in this environment.*
- [x] **9. Disclosed deferrals** — n/a — *None declared.*
- [x] **10. Follow-up claims verified** — n/a — *No follow-up claims in the body.*

**Verdict derivation:** zero blockers, zero tech debt, zero nits; no rubric item is `fail` -> **APPROVED**.

Issue #2751, opened 2026-08-13
Fast-path importtime probe measures interpreter-startup imports it does not own


valorengels:
## Problem

`tests/unit/test_post_tool_use_fast_path.py::test_counter_only_call_does_not_import_popoto` asserts a contract about what **the PostToolUse hook** imports, but it measures the entire `-X importtime` graph of the subprocess — which includes everything the interpreter preloads at startup via `site` and any installed `.pth` file.

That means any site-level shim installed into the venv (by `/update`, by a branch's installer, by a third-party package) lands inside the measured graph even though the hook never asked for it. #2692 fixed the immediate false positive by matching on module *identity* instead of substring containment, which is sufficient for the `tools.redis_flush_guard` / `_redis_flush_guard_boot` case. It does not address the broader mismatch: a `.pth` that genuinely imports `redis` at interpreter startup would still fail the test, and the hook would still be blameless.

## Desired outcome

The probe measures only imports attributable to the hook. Candidate approaches:

- Diff the hook's `-X importtime` graph against a `python -X importtime -c pass` baseline captured in the same interpreter, and assert only on the difference.
- Run the probe with `-S`-equivalent isolation so `site` processing (and therefore `.pth` execution) is excluded from the measured graph.

Either is a redesign of the probe's contract, not a one-line assertion tightening, which is why it was explicitly scoped out of #2692.

## Notes

- Deliberately deferred from #2692 (see that plan's No-Gos).
- `-S` isolation has a known hazard: the hook subprocess may depend on `site`-installed `sys.path` entries to import at all, so the baseline-diff approach is likely the safer of the two and should be prototyped first.

Refs #2692.

PR #2680, opened 2026-08-07
Harden production Redis against accidental flush (four layers)

tomcounsell:
Closes #2645

Four independent layers, any one of which would have stopped the 2026-08-07 script that destroyed 25,825 keys of live agent state. The root cause pattern in both incidents is that every prior control was installed at a boundary the offending code did not cross: the conftest guard lives in the test harness, the hook validator gated on Popoto vocabulary, and AOF acts after the damage. This moves the same rule to boundaries every caller must cross.

| Layer | Boundary | Stops |
|---|---|---|
| 1. Process-wide flush guard | Every Python process using a repo venv | The incident exactly: any `.flushdb()` on db 0, any `.flushall()` |
| 2. Redis ACL planner | The Redis server | Non-Python clients, **after an operator runs the runbook by hand**. This PR ships the planner, the rule set, the staged directive, the doctor check, and the runbook. It applies nothing. |
| 3. PreToolUse validator | Agent-issued Bash | A flush typed before it reaches an interpreter |
| 4. Docs + CLAUDE.md | The read-before-testing surface | The `setdefault` reasoning error at its source |

## This PR made no change to production state

- `redis-cli ACL LIST` before and after `--dry-run`: `user default on nopass sanitize-payload ~* &* +@all`. No `valor-app`, unchanged.
- `/opt/homebrew/etc/redis.conf` has no `aclfile` line (`grep -c '^aclfile'` prints 0). Nothing in the diff opens it for writing.
- `data/redis-acl-enabled` does not exist. No service was restarted. No `.env` or vault write.

## Demonstrated red

Every guard was mutated and re-measured. A green test that reaches no code is worse than no test.

**Layer 1 (flush guard), 37 tests**

| Mutation | Tests that went RED |
|---|---|
| Neuter the db-0 check (`if False:`) | 12, incl. `test_flushdb_db0_is_blocked`, `test_async_flushdb_db0_is_blocked`, all 5 `test_malformed_client_treated_as_db0` |
| `flushall` delegates instead of raising | 9, incl. all 3 `test_flushall_blocked_at_any_db`, `test_async_flushall_is_blocked` |
| Truthy override check instead of `== "1"` | 3: the `"0"`, `"false"`, `"no"` cases |
| Delete the `finally:` re-insertion in the finder swap | `test_finder_survives_find_spec_exception_and_later_import_still_arms` |
| Replace the file-path installer load with a package import | `test_self_heal_loads_installer_by_path_not_package_import`, failing with the whole update system listed |
[...]

Issue #2645, opened 2026-08-07
Harden production Redis against accidental flush: guarded connections, server-side FLUSHDB restriction, hook validator, docs


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/redis-flush-hardening.md

> **Context for the stranger:** this machine runs a production Redis (localhost:6379, DB 0) holding all live agent state — AgentSession/Room/Job models via the [Popoto ORM](https://github.com/tomcounsell/popoto), the subconscious-memory store, steering lists, DM coverage epochs. Test suites use other DB numbers (10, 15, ...) on the same server. On 2026-08-07 an agent's standalone debug script flushed DB 0 (25,825 keys). It was recovered by an AOF point-in-time restore with ~30s of Redis downtime and ~2 minutes of writes discarded. **This issue tracks hardening so the class of mistake cannot recur; the recovery itself is done and needs no action.**

## Problem

An ad-hoc Python script debugging sorted-set behavior called `POPOTO_REDIS_DB.flushdb()`, intending to hit test DB 15. It set the target via `os.environ.setdefault("REDIS_URL", "...db 15...")` — but `setdefault` does not override a variable that is already set, and the shell environment carried the production `REDIS_URL`. The connection silently went to DB 0 and the flush destroyed the production dataset.

**Current behavior:** nothing stands between an ad-hoc script and a production flush.
- Any script that inherits the environment gets a production-pointed `REDIS_URL` by default; targeting a test DB requires the author to remember to *override*, and the most natural idiom (`setdefault`) silently fails to do so.
- The Redis server itself allows `FLUSHDB`/`FLUSHALL` from any client.
- The existing PreToolUse validator (`.claude/hooks/validators/validate_no_raw_redis_delete.py`) blocks raw `delete`/`srem`/`sadd`/`zrem` on Popoto-managed keys in Bash-invoked Python, but has no rule for `flushdb`/`flushall` — the strictly more destructive operation.
- `CLAUDE.md`'s Manual Testing Hygiene section warns about raw Redis on Popoto keys but does not name the env-var-inheritance foot-gun.

**Desired outcome:** four independent layers, any one of which would have prevented the incident:
1. Connection-construction guard (application layer)
2. `FLUSHDB`/`FLUSHALL` disabled or ACL-restricted on the production instance (server layer)
3. Hook validator blocks `flushdb`/`flushall` in agent-issued Bash (harness layer)
[...]

PR #2758, opened 2026-08-13
Fix importtime false-positive on redis-flush-guard .pth shim lookalikes

### LINKED ISSUE #2692: Fast-path importtime test false-positives on the redis-flush-guard .pth shim (test_counter_only_call_does_not_import_popoto)
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/fastpath-importtime-false-positives.md

## Summary

`tests/unit/test_post_tool_use_fast_path.py::test_counter_only_call_does_not_import_popoto` fails on any machine whose venv has the in-flight Redis flush guard `.pth` installed (`#2645`). The PostToolUse counter-only fast path is **not** regressed — nothing new imports popoto or redis. The test's heavy-module detection is a naive substring scan that matches the *name* of the guard's boot shim.

## Failing node ID

```
tests/unit/test_post_tool_use_fast_path.py::test_counter_only_call_does_not_import_popoto
```

## Reproduction

From a checkout whose venv has the guard `.pth` installed (the main checkout `/Users/valorengels/src/ai` on this machine):

```bash
./scripts/pytest-clean.sh "tests/unit/test_post_tool_use_fast_path.py::test_counter_only_call_does_not_import_popoto"
```

Minimal, no-pytest reproduction of the underlying signal:

```bash
python -X importtime -c "pass" 2>&1 | grep -iE "redis|guard"
# import time:  67 |  159 |     tools.redis_flush_guard
# import time:  52 |  211 |   _redis_flush_guard_boot
```

The test passes in a worktree venv that has not had `/update`'s guard installer run against it (e.g. `.worktrees/nightly-triage-b9b95094/.venv`), which is why this reads as an intermittent/nightly-only failure.

## Observed vs expected

- **Observed:** `AssertionError: counter-only path imported heavy modules: ['redis']`
- **Expected:** pass — the counter-only path performs no popoto/redis import.

## Root cause

`tests/unit/test_post_tool_use_fast_path.py:77-79` scans the entire `-X importtime` stderr dump with plain substring containment:

```python
heavy = ("popoto", "redis", "config.settings", "models.memory", "models.agent_session")
offenders = [m for m in heavy if m in proc.stderr]
```

Two things make that a false positive:

1. **Substring, not module identity.** `"redis"` is a substring of the module names `tools.redis_flush_guard` and `_redis_flush_guard_boot`. Neither is the `redis` package.
2. **`-X importtime` logs failed imports too.** On this machine `/Users/valorengels/src/ai/.venv/lib/python3.14/site-packages/zzz_redis_flush_guard.pth` (`import _redis_flush_guard_boot`) runs at interpreter startup. The shim's body is `import tools.redis_flush_guard; arm()` wrapped in `except Exception: pass`. `tools/redis_flush_guard.py` **does not exist on `main`** — it lives only on the unmerged branch `session/redis-flush-hardening` (#2645) — so the import raises `ModuleNotFoundError` and is swallowed. CPython still emits an `importtime` line for the attempted module, so the name lands in stderr regardless.

Verified directly:

```
$ python -c "import tools.redis_flush_guard"
ModuleNotFoundError: No module named 'tools.redis_flush_guard'
```

The hook's fast path itself (`.claude/hooks/post_tool_use.py:384-398`) is unchanged and still early-returns before `from hook_utils.memory_bridge import recall` on non-window calls. `git log -- .claude/hooks/post_tool_use.py` shows no touch since `c50938cf9`.

## Secondary finding (environment pollution)

Independent of the test bug: an **unmerged** branch's `/update` step has already installed `zzz_redis_flush_guard.pth` + `_redis_flush_guard_boot.py` into the shared main-checkout venv. Every Python interpreter start on this machine now performs a failing import that is silently swallowed. That is a real operational leak worth fixing on #2645's branch (installer should not run, or should self-remove, when the guard module is absent from the checkout it points at).

Relevant files:
- `/Users/valorengels/src/ai/.venv/lib/python3.14/site-packages/zzz_redis_flush_guard.pth`
- `/Users/valorengels/src/ai/.venv/lib/python3.14/site-packages/_redis_flush_guard_boot.py`
- installer (branch-only): `scripts/update/redis_flush_guard_pth.py`

## Suggested next steps

1. Tighten the assertion in `tests/unit/test_post_tool_use_fast_path.py` to match module *identity* rather than substrings: parse each `import time:` line, take the module name from the final column, strip leading whitespace, and flag when the name equals a heavy module or is a dotted descendant of one (`name == m or name.startswith(m + ".")`). That kills the `tools.redis_flush_guard` / `_redis_flush_guard_boot` false positive while still catching a genuine `redis`/`popoto` import.
2. Consider excluding interpreter-startup (`.pth`/`site`) imports from the measured graph entirely — the test's contract is about what *the hook* imports, not what the interpreter is configured to preload. Running the probe with `-S`-equivalent isolation or diffing against a `-X importtime -c pass` baseline would make the assertion robust to any future site-level shim.
3. On #2645 / `session/redis-flush-hardening`: make the `.pth` installer refuse to install (and clean up an existing install) when `tools/redis_flush_guard.py` is not importable from the target checkout, so an in-flight branch cannot leave a swallowed failing import in the shared venv.

Cross-ref: #2645.

## Recon Summary

Re-verified against `main` at `0dd8e70f2` on 2026-08-13 (issue filed 2026-08-07).

**Confirmed:**
- The defective assertion is still present verbatim at `tests/unit/test_post_tool_use_fast_path.py:77-79` — `heavy = ("popoto", "redis", ...)` followed by `offenders = [m for m in heavy if m in proc.stderr]`. Plain substring containment over the whole stderr dump.
- The hook fast path at `.claude/hooks/post_tool_use.py:384-398` is unchanged and still early-returns before `from hook_utils.memory_bridge import recall`. `git log --since=2026-08-07T20:21:57Z -- .claude/hooks/post_tool_use.py tests/unit/test_post_tool_use_fast_path.py` returns **no commits**. The fast path is not regressed; the test is wrong.
- CPython 3.14.5 (the `.python-version` pin) `-X importtime` line shape verified empirically: `import time: {self_us} | {cumulative_us} | {indent}{module_name}`. The module name is the final `|`-delimited column, indented with spaces. `-X importtime=2` additionally emits `import time: cached    | cached     | {name}` rows with the same column layout. Neither mode emits tree-drawing characters on this interpreter. The header row is `import time: self [us] | cumulative | imported package` — its final column contains an interior space, which no real module name can.
- PR #2680 (`session/redis-flush-hardening`, **OPEN**) adds both `tools/redis_flush_guard.py` and `scripts/update/redis_flush_guard_pth.py`. Merging it reintroduces the shim with the module actually present, which turns this false positive into a deterministic failure. This is the urgency.
- Blast radius is one file. `grep -rn importtime tests/ .claude/ scripts/` finds only this test's two live uses plus a prose comment at `tests/integration/test_memory_mcp_server.py:122` (no stderr scanning there).

**Revised:**
- **The issue's reproduction no longer reproduces on `main`.** `zzz_redis_flush_guard.pth` and `_redis_flush_guard_boot.py` have been removed from `/Users/valorengels/src/ai/.venv/lib/python3.14/site-packages/`; `tools/redis_flush_guard.py` is absent from `main`; `python -X importtime -c pass` emits no `redis`/`guard` lines; and `tests/unit/test_post_tool_use_fast_path.py` is **4/4 green**. The *defect* (substring matching) is still in the code — only its environmental trigger is currently gone. Consequence for planning: red-first must be **synthesized** from a crafted importtime fixture, not observed from environment state.
- The "Secondary finding (environment pollution)" in the body is stale as written — the shared main venv has already been cleaned. The installer-side hardening it asks for remains a live concern, but on #2680's branch, not here.

**Pre-requisites:**
- None. Test-file-only change; no new dependencies, no config, no migration.

**Dropped:**
- Suggested step 2 (excluding interpreter-startup/`.pth`/`site` imports via `-S`-equivalent isolation or baseline diffing) — out of scope. Module-identity matching alone removes the false positive; baseline diffing is a larger redesign of the probe's contract.
- Suggested step 3 (make the `.pth` installer refuse to install / self-clean when `tools/redis_flush_guard.py` is not importable) — belongs to #2680 / `session/redis-flush-hardening`, which owns `scripts/update/redis_flush_guard_pth.py`. Not this branch's file.

**Overlap check:** `docs/plans/redis-flush-hardening.md` (status `Ready`, tracking #2645) is the active adjacent plan. It uses `-X importtime` for its own startup-budget assertion in `tests/unit/test_redis_flush_guard_prod.py` but does **not** touch `tests/unit/test_post_tool_use_fast_path.py`. No file-level conflict; ordering is the only coupling (this must land before #2680 merges).

valorengels:
## Summary

`test_counter_only_call_does_not_import_popoto` scanned the whole `-X importtime` stderr dump for heavy-module names as plain **substrings**. `tools.redis_flush_guard` / `_redis_flush_guard_boot` merely contain `"redis"` as a substring — they are not the `redis` package — so once the redis-flush-guard `.pth` shim is installed, the test fails a hook that imported nothing heavy at all.

This PR replaces the substring scan with module-**identity** matching: parse `-X importtime`'s three-column row format into a list of module names, then flag a heavy module only on exact match or dotted-descendant prefix.

Closes #2692

## Red-first paper trail

Task 1 landed the crafted-fixture tests against the **unchanged, verbatim old substring logic** first, to make the false positive observable and reproducible from the repo alone (the real venv no longer has the `.pth` shim installed, so the bug can't be reproduced from environment state):

```
tests/unit/test_post_tool_use_fast_path.py:116: in test_heavy_detection_ignores_lookalike_module_names
    assert _heavy_import_offenders(_GUARD_SHIM_DUMP) == []
E   AssertionError: assert ['redis'] == []
E
E     Left contains one more item: 'redis'
FAILED tests/unit/test_post_tool_use_fast_path.py::test_heavy_detection_ignores_lookalike_module_names
1 failed, 5 passed
```

Task 2 then replaced the body with `_importtime_module_names()` + identity matching in `_heavy_import_offenders()`. `./scripts/pytest-clean.sh tests/unit/test_post_tool_use_fast_path.py -q` now reports **6 passed**.

## Changes

- `tests/unit/test_post_tool_use_fast_path.py` — the only file touched:
  - `_HEAVY_MODULES` lifted to module scope (single shared definition)
  - `_importtime_module_names(stderr)` — parses the documented three-column importtime row format
  - `_heavy_import_offenders(stderr, heavy=_HEAVY_MODULES)` — identity match (`n == h or n.startswith(h + ".")`) instead of substring
  - `test_heavy_detection_ignores_lookalike_module_names` — the red-first case, also exercises Risk 2's malformed-row guard (`truncated output`, a 2-field row) and the empty-input case
  - `test_heavy_detection_still_catches_real_imports` — the over-loosening guard: a genuine top-level `redis`, a `popoto.models` descendant, and an exact `config.settings` are still caught
  - `test_counter_only_call_does_not_import_popoto` now routes through `_heavy_import_offenders(proc.stderr)`

No production code touched. `.claude/hooks/post_tool_use.py` is unmodified.

## Test plan

- [x] `./scripts/pytest-clean.sh tests/unit/test_post_tool_use_fast_path.py -q` — 6 passed
- [x] `python -m ruff check tests/unit/test_post_tool_use_fast_path.py` — clean
- [x] `python -m ruff format --check tests/unit/test_post_tool_use_fast_path.py` — clean
- [x] `git diff --name-only main...HEAD` — exactly one file
- [x] Broader unit-test spot check (`test_post_tool_use_fast_path.py`, `test_reflection_scheduler.py`, `test_session_modal_liveness_render.py`, `test_update_hardlinks.py`) — only the four known pre-existing `main` baseline failures remain, no new failures introduced
- [x] All 13 non-lint Verification-table rows independently re-derived and confirmed true (two rows misreport in the automated `verification_parser`/`validate_build.py` harnesses due to pre-existing table-escaping and `match count == N` idiom-support gaps in those tools, not defects in this change — manually confirmed with `grep -c "^def test_" ...` → 6 and `grep -cE 'startswith\(h \+ "\."\)|startswith\(heavy \+ "\."\)' ...` → 1)

### DIFF
--- a/tests/unit/test_post_tool_use_fast_path.py
+++ b/tests/unit/test_post_tool_use_fast_path.py
@@ -19,6 +19,37 @@ import pytest
 REPO_ROOT = Path(__file__).resolve().parent.parent.parent
 HOOK = REPO_ROOT / ".claude" / "hooks" / "post_tool_use.py"
 
+_HEAVY_MODULES = ("popoto", "redis", "config.settings", "models.memory", "models.agent_session")
+
+_GUARD_SHIM_DUMP = """\
+import time: self [us] | cumulative | imported package
+import time:       120 |        120 |   site
+import time:        45 |         45 |     tools.redis_flush_guard
+import time:        30 |         30 |     _redis_flush_guard_boot
+import time: cached    | cached     | _redis_flush_guard_boot
+import time: truncated output
+import time:  120 |  120
+"""
+
+_REAL_HEAVY_DUMP = """\
+import time: self [us] | cumulative | imported package
+import time:       310 |        310 |   redis
+import time:       220 |        530 |     popoto.models
+import time:       140 |        140 |   config.settings
+"""
+
+
+def _heavy_import_offenders(stderr: str, heavy: tuple[str, ...] = _HEAVY_MODULES) -> list[str]:
+    """Return which ``heavy`` module names are genuinely present in a
+    ``-X importtime`` stderr dump.
+
+    Placeholder body (Task 1 / red-first): the ORIGINAL substring scan,
+    verbatim. It is deliberately kept naive here so the crafted fixtures can
+    demonstrate the false positive before the identity-matching parser
+    replaces this body in Task 2.
+    """
+    return [m for m in heavy if m in stderr]
+
 
 def _sidecar(session_id: str) -> Path:
     return REPO_ROOT / "data" / "sessions" / session_id / "memory_buffer.json"
@@ -74,9 +105,26 @@ def test_counter_only_call_does_not_import_popoto(clean_session):
         capture_output=True,
     )
     assert proc.returncode == 0, proc.stderr
-    heavy = ("popoto", "redis", "config.settings", "models.memory", "models.agent_session")
-    offenders = [m for m in heavy if m in proc.stderr]
+    offenders = _heavy_import_offenders(proc.stderr)
     assert not offenders, f"counter-only path imported heavy modules: {offenders}"
+
+
+def test_heavy_detection_ignores_lookalike_module_names():
+    """``tools.redis_flush_guard`` / ``_redis_flush_guard_boot`` must NOT be
+    mistaken for the ``redis`` package: they merely contain "redis" as a
+    substring. Also covers Risk 2's malformed-row and empty-input guards."""
+    assert _heavy_import_offenders(_GUARD_SHIM_DUMP) == []
+    assert _heavy_import_offenders("") == []
+
+
+def test_heavy_detection_still_catches_real_imports():
+    """A genuine top-level ``redis``, a ``popoto.models`` descendant, and an
+    exact ``config.settings`` must still be flagged (over-loosening guard)."""
+    assert set(_heavy_import_offenders(_REAL_HEAVY_DUMP)) == {
+        "redis",
+        "popoto",
+        "config.settings",
+    }
 
 
 def test_counter_only_call_bumps_sidecar_counter(clean_session):
--- a/tests/unit/test_post_tool_use_fast_path.py
+++ b/tests/unit/test_post_tool_use_fast_path.py
@@ -39,16 +39,50 @@ import time:       140 |        140 |   config.settings
 """
 
 
+def _importtime_module_names(stderr: str) -> list[str]:
+    """Parse module names out of a ``-X importtime`` stderr dump.
+
+    Row contract (stable since Python 3.7, verified on CPython 3.14.5):
+    each import row is ``import time: {self_us} | {cumulative_us} |
+    {indent}{module_name}``, preceded by a header row whose final column is
+    the literal ``imported package`` (the only column with an interior
+    space, which is what lets us skip the header generically). ``-X
+    importtime=2`` additionally emits ``import time: cached | cached |
+    {name}`` rows sharing the same three-column shape. Dotted submodule
+    names appear in full (e.g. ``json.scanner``), not as bare leaves.
+    Malformed rows (fewer than 3 pipe-delimited fields — CPython documents
+    that importtime output "may be broken in multi-threaded applications")
+    are skipped rather than treated as fatal.
+    """
+    names = []
+    for line in stderr.splitlines():
+        if not line.startswith("import time:"):
+            continue
+        parts = line.split("|")
+        if len(parts) < 3:
+            continue
+        name = parts[-1].strip().lstrip("│├└─ \t")
+        if not name or " " in name:
+            continue
+        names.append(name)
+    return names
+
+
 def _heavy_import_offenders(stderr: str, heavy: tuple[str, ...] = _HEAVY_MODULES) -> list[str]:
     """Return which ``heavy`` module names are genuinely present in a
     ``-X importtime`` stderr dump.
 
-    Placeholder body (Task 1 / red-first): the ORIGINAL substring scan,
-    verbatim. It is deliberately kept naive here so the crafted fixtures can
-    demonstrate the false positive before the identity-matching parser
-    replaces this body in Task 2.
+    Matches on module *identity* (exact name or dotted descendant) rather
+    than substring containment. Substring matching produced a false
+    positive for #2692: ``tools.redis_flush_guard`` and
+    ``_redis_flush_guard_boot`` merely contain "redis" as a substring — they
+    are not the ``redis`` package and not descendants of it — so a plain
+    ``"redis" in stderr`` check blamed the hook for an import it never
+    performed.
     """
-    return [m for m in heavy if m in stderr]
+    names = _importtime_module_names(stderr)
+    offenders = {h for h in heavy for n in names if n == h or n.startswith(h + ".")}
+    return sorted(offenders)
 
 
 def _sidecar(session_id: str) -> Path:


valorengels:
## Review: Approved

A minimum-fix, test-only change that replaces a substring scan of `-X importtime` stderr with module-**identity** matching. One file, 6/6 green, lint and format clean, no production code touched.

I re-derived the load-bearing claims in my own environment rather than crediting the PR body (Hard Rule 10). Everything checked out, including the red-first paper trail.

**Independently reproduced:**

- **The red.** Extracting `_GUARD_SHIM_DUMP` and the verbatim pre-fix body `[m for m in heavy if m in stderr]` and running them here gives `['redis']`; the new `_heavy_import_offenders` gives `[]`. The FAIL block pasted in the PR body matches commit `9857e2c2b` verbatim, and line 116 in that commit is exactly `assert _heavy_import_offenders(_GUARD_SHIM_DUMP) == []`. The red-first paper trail is real, not narrated.
- **The detector is not disarmed** — the single most important risk here, since a parser that returns `[]` for everything makes the test pass unconditionally while the optimization it pins silently regresses. Two independent checks: `_REAL_HEAVY_DUMP` yields `{redis, popoto, config.settings}` under *both* the old and new bodies; and running the real hook probe under `-X importtime` emits **89** `import time:` rows, of which the parser extracts **88** module names — exactly one dropped, the header. 13 of those are dotted names, so the `startswith(h + ".")` descendant branch is exercised against real output shapes, not just crafted ones.
- **Blast radius is genuinely one file.** `grep -rn importtime tests/ .claude/ scripts/` finds one live use (this file's own `subprocess.run`); the only other hit is a non-executing comment at `tests/integration/test_memory_mcp_server.py:122`. Nothing imports these helpers. `git diff --name-only main...HEAD` is exactly `tests/unit/test_post_tool_use_fast_path.py`; the hook is untouched.
- **6/6 passed** and `grep -c "^def test_" ` is **6** — no seventh node was split out, so the exact-count criteria hold.

**Not reproduced, and why that is sound:** I did not re-run the full unit suite to re-derive the "four known pre-existing `main` baseline failures" claim. That claim is about `main`, not about this diff. The substantive question — can this change break another test — is answered structurally and I verified the premises myself: the diff is one test file, adds no imports beyond the stdlib, defines only module-private helpers, and has zero importers. No other test's outcome can move.

## Rubric

- [x] **1. Plan vs. implementation match** — pass — *all 10 Success Criteria verified against the diff; ticked in the plan doc.*
- [x] **2. New code quality** — pass — *type-hinted pure functions, both with docstrings; no try/except, malformed input handled by explicit `continue` guards rather than swallowed errors.*
- [x] **3. Test coverage** — pass — *red-first case, over-loosening guard, malformed-row and empty-input guards all executed.*
- [x] **4. Regression risk to existing callers** — pass — *no production code; helpers are module-private with no importers.*
- [x] **5. Data integrity** — n/a — *no schema, model, or persisted state touched.*
- [x] **6. Security** — pass — *no secrets, no eval/exec; the only `subprocess` call is pre-existing and uses `sys.executable` with a repo-internal path.*
- [x] **7. Documentation accuracy** — pass — *both docstrings the plan's Documentation section requires are present: the three-column row contract on `_importtime_module_names`, and the identity-not-substring rationale citing #2692 on `_heavy_import_offenders`. No `docs/features/` entry required.*
- [x] **8. PR body accuracy** — pass — *every claim I checked held, including the subtle one about which Verification rows misreport (see below).*
- [x] **9. Disclosed deferrals** — pass — *#2751 and #2645 both OPEN.*
- [x] **10. Follow-up claims verified** — pass — *#2751 OPEN, #2645 OPEN.*

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — PASS — *10/10 verified.*
- **2. No-Gos from plan — none violated** — PASS — *no `-S`/baseline-diff creep, no `.pth` installer edits, no `tools/redis_flush_guard.py` changes; each has a passing anti-criterion row.*
- **3. New `except Exception` blocks — each has logger/raise/swallow-ok** — N/A — *none added; the plan deliberately uses branch guards instead.*
- **4. New integration tests — exercise serialization boundary (not in-memory only)** — N/A — *pure string parsing. The live probe that does cross a real subprocess boundary is pre-existing and still exercised.*
- **5. Plan internal consistency — spike findings match task steps** — PASS — *Research's verified 3.14.5 row format matches the parser as built; one Freshness Check statement has since gone stale, noted under Acknowledged Deferrals.*
- **6. No hardcoded secrets or debug artifacts** — PASS — *no prints, no TODOs, no commented-out code.*
- **7. New public APIs — docstrings present** — PASS — *both helpers documented.*
- **9. Tests added for new behavior** — PASS — *two new nodes.*
- **10. Tests cover the failure path (not just happy path)** — PASS — *`test_heavy_detection_still_catches_real_imports` is the positive-detection guard; malformed rows and `""` are covered inside the lookalike node.*
- **11. UI changes (if any) — screenshot captured** — N/A — *no HTML/CSS/JS/TSX/Vue/template files in the diff; the visual-proof gate is a no-op.*
- **12. Docs updated for user-facing changes** — N/A — *no user-facing surface; the plan's Documentation section specifies inline docstrings only, and both are present.*

### Blockers
- None

### Tech Debt
- None

### Nits
- None

### Miscellaneous
- None

- **[SEPARATE-SLUG #2751] Excluding interpreter-startup / `.pth` / `site` imports from the measured graph** — tracked by **#2751 (OPEN)**. Anti-criterion row confirms no `-S`, `-c pass`, or `isolated_baseline` creep.
- **[SEPARATE-SLUG #2645] Hardening the `.pth` installer to refuse/self-clean when the guard module is absent; any change to `tools/redis_flush_guard.py`, `_redis_flush_guard_boot.py`, or `zzz_redis_flush_guard.pth`** — tracked by **#2645 (OPEN)**. Those files live only on `session/redis-flush-hardening`; anti-criterion row confirms none were touched.
- **[ORDERED] Coordinating merge order with PR #2680** — explicit out-of-scope exclusion with sound rationale (#2680 is human-gated on a separate lane; this plan can only be sized to land first). **One correction for whoever merges second:** the plan's Freshness Check states #2680 "does **not** touch `tests/unit/test_post_tool_use_fast_path.py`. No file-level conflict." That is no longer true — #2680 now modifies this same file at the same region, adding its own inline `partition("|")`-based identity fix to `test_counter_only_call_does_not_import_popoto`. A textual conflict is therefore certain, and the hazard is directional: resolving it in #2680's favor would silently revert this PR's module-scope helpers and drop both isolated test nodes along with the malformed-row and empty-input guards. **Resolve in favor of this PR's version.** I am not treating this as a finding against this PR — it is not patchable here, because adding the plan-doc correction to this branch would falsify the plan's own "Diff is one file" anti-criterion.

### Verification Results

All 15 rows pass on substance. Four report FAIL under the automated `agent.verification_parser` runner; all four are harness artifacts, not defects, and I confirmed each by hand:

- **PASS (11 rows):** Fast-path suite green; Six cases present; Red-first case exists; Over-loosening guard exists; Risk 2 malformed rows exercised; Risk 2 empty input exercised; Substring scan gone; Hook untouched; Diff is one file; No `-S`/baseline-diff creep; No `.pth` installer edits.
- **"Exactly six test nodes"** — runner FAIL, actually correct. The runner does not implement the `match count == N` expectation idiom; its own captured output is `6`.
- **"Identity matching present"** — runner FAIL, actually correct. The markdown table parser unescapes `\|` to a literal `|`, breaking the BRE alternation. Run as written with the escape intact, `grep -c 'startswith(h + "\.")\|startswith(heavy + "\.")'` returns **1**.
- **Lint clean / Format clean** — runner FAIL only because it invoked the pyenv interpreter, which has no `ruff` installed. Against the worktree venv: `ruff check` exit 0 ("All checks passed"), `ruff format --check` exit 0 ("1 file already formatted").

This matches the PR body's claim precisely — it scoped its statement to the 13 non-lint rows and named exactly the two that misreport.

### Screenshots

### Plan Checkbox Sync

All 10 Success Criteria ticked in `docs/plans/fastpath-importtime-false-positives.md`, committed on **`main`** (`fd1182246`) rather than on this branch. Committing the plan doc to `session/fastpath-importtime-false-positives` would have falsified this plan's own "Diff is one file (anti-criterion)" Verification row — the one place where the standard checkbox-sync procedure conflicts with the plan's gates. The plan doc's entire history already lives on `main`.

Issue #2725, opened 2026-08-11
docs_auditor rename detectors pick the wrong target: newest-commit hop, and a doc-relative link resolved against the repo root


valorengels:
The three rename detectors in `reflections/docs_auditor.py` pick their replacement target with
`renames[0][1]`, and one of them resolves the replacement against the wrong root. Both were found
during recon for #2711 and are deliberately left out of that PR's scope: the existence invariant
added there demotes each from "writes a wrong reference" to "declines to write", so they are
precision defects, not correctness ones.

## Defect 1 — `renames[0][1]` is the newest rename *commit*, not the final hop

`_git_log_follow_renames` (`:342`) returns every `R` record from
`git log --follow --diff-filter=R --name-status`. All three callers take `renames[0][1]`:

- `_detect_renamed_link_fixes:412`
- `_detect_renamed_symbol_fixes:430`
- `_detect_readme_broken_entries:460`

For a file renamed A→B and then **deleted**, the query still returns the A→B record, so the
detector proposes `B` — a path that does not exist. After #2711's guard this is rejected rather
than written, but the auditor also stops offering the correct answer (file an issue), because the
rejection happens at the apply layer, not the detect layer.

## Defect 2 — doc-relative link, repo-root-relative replacement

`_detect_renamed_link_fixes` computes `rel` as repo-root-relative (`:404`) but emits the fix as
`(target, new_name)` at `:413`, where `target` is the **doc-relative** link text and `new_name` is
the **repo-root-relative** git path. For any doc not sitting at the repo root, a fully correct
rename still produces a broken link.

## Recon Summary

**Confirmed:** both defects verified by reading `reflections/docs_auditor.py` at `e48607be3`.
Line references above are exact. All three call sites take index `[0]`; none walks the chain.
`_detect_renamed_link_fixes:404` uses `target_path.relative_to(repo_root.resolve())` while
`:413` appends the doc-relative `target` as the search key.

**Confirmed:** neither defect is the cause of #2711 — that was the `STALE_TERMS` substring class
(`:45`, `:467`). These are independent latent defects in a sibling fix class.

**Pre-requisites:** #2711's existence invariant in `_apply_fixes_to_file` should land first;
it is the safety net that makes these non-urgent.

**Dropped:** nothing.

valorengels:
[...]

Issue #2739, opened 2026-08-13
docs_auditor is its own committer: put a review gate in front of every write it makes


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/docs-auditor-review-gate.md

> **Context for readers new to this repo.** The **docs auditor** ([`reflections/docs_auditor.py`](../blob/main/reflections/docs_auditor.py), documented at [`docs/features/docs-auditor.md`](../blob/main/docs/features/docs-auditor.md)) is an automated documentation maintainer. It scans Markdown under `docs/`, detects references that have gone stale (a renamed file, a deleted symbol, a broken README link), and rewrites them. It has two callers:
>
> - **Caller A — rotation.** A daily scheduled *reflection* (`config/reflections.yaml:265-273` → `run_docs_auditor`) that picks the least-recently-audited doc, fixes it, and opens a PR.
> - **Caller B — cascade.** The [`/do-docs`](../blob/main/.claude/skills-global/do-docs/SKILL.md) skill, run as the docs stage of every feature PR, invoking `audit(scope_mode="pr-changed-files")` to fix docs affected by that PR's code changes.
>
> This issue consolidates four open bugs that all descend from one structural property, and closes them. A fifth, previously unfiled defect found while scoping is included here.

## Problem

**The docs auditor is its own committer.** It generates a rewrite and commits that rewrite itself, in the same function call, with nothing between generation and the permanent record. There is no review gate — no human, no agent, no test — in front of any write it makes.

This is not a hypothetical. In #2711 the auditor invented a module rename that never happened (`agent/session_logs.py` → `agent/agent_sessions.py`), applied it in five places across two docs, and committed it as `d7bf3ad99`. It was caught only because a human happened to run `ls`. The generator bug is being fixed on PR #2728, but *the reason a generator bug reached `main` at all* is unaddressed: any future generator bug will take exactly the same path.

Four consequences, each filed separately, all reducible to that one property:

### 1. Self-commit and over-broad staging (#2726)

The cascade path commits to the current branch itself:

```python
# reflections/docs_auditor.py:1057-1058
if scope_mode == "pr-changed-files" and apply_mode == "apply" and touched:
    _commit_current_branch(root, touched)
```

[...]

Issue #2729, opened 2026-08-11
docs-auditor withheld PRs are auto-merge-ineligible forever, so the sweeper stale-closes them at day 14 with no escalation


valorengels:
## What happens

PR #2728 made a docs-auditor rotation PR **permanently** ineligible for auto-merge when the run withheld any fix (`WITHHELD_PR_MARKER` in the PR body, checked by `_pr_is_auto_merge_eligible` at `reflections/docs_auditor.py:1901-1915`). That gate is correct — a withheld fix means the auditor tried to write a reference to a path that does not exist, so its surviving output is suspect and needs a human read.

The gap is what happens when no human ever reads it. The disqualification never lifts, so the PR falls straight through to the branch sweeper's stale-close branch (`reflections/docs_auditor.py:2125-2136`):

```python
if age_days >= STALE_PR_AGE_DAYS:   # 14
    subprocess.run(["gh", "pr", "close", "--delete-branch", str(pr_num)], ...)
```

At day 14 the PR is closed and the branch deleted. The fixes that *did* pass the existence invariant are discarded with it.

## Why nothing pages anyone

The only record is `findings.append(f"Closed stale PR #{pr_num} ...")`. `run_docs_auditor` and the sweeper are `execution_type: function` reflections, and `agent/reflection_scheduler.py:514` reads **only** `result.get("projects")` from a function reflection's return — `findings` and `summary` reach no human. The creation-time Telegram message does say the PR is not auto-merge eligible, but nothing fires at close time.

## Why it is a loop, not a one-off

The withheld condition is never repaired. The next rotation onto the same slug re-detects the same stale term, re-proposes the same rejected fix, re-opens a PR carrying the marker, and re-closes it 14 days later. Perpetual churn with no escalation.

## Not a regression

Strictly safer than the pre-#2728 behavior, which auto-merged suspect output. This is an *unowned escalation gap that #2728 created*, raised independently by both review judges (code-quality A3, risk R1) and recorded in the code as a `KNOWN ESCALATION GAP` comment at `reflections/docs_auditor.py:1904-1914`.

## Suggested fix

Either:
- Exempt `WITHHELD_PR_MARKER` PRs from the stale-close branch entirely (they stay open until a human acts), or
- Send a Telegram notification before closing one, so "needs a human" does not silently resolve to "quietly dropped".

The first is simpler; the second bounds unbounded PR accumulation. Owner ruling wanted.

Refs #2711, #2728.

valorengels:
[...]

Issue #2711, opened 2026-08-10
docs_auditor auto-fix invented a module rename (agent/session_logs.py -> agent/agent_sessions.py) and committed it


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/docs-auditor-rename-guard.md

## What happened

During the `/do-docs` cascade for PR #2681 (issue #2517), Step 2d ran the declared auto-fix substrate:

```python
from reflections.docs_auditor import audit
audit(primary_path=None, scope_mode='pr-changed-files', apply_mode='apply', project_key='valor')
```

It returned `{"status": "ok", "files_touched": ["docs/features/bridge-workflow-gaps.md", "docs/features/scheduled-disk-reclaim.md"], "fixes_applied": 5, "issues_filed": 0}` and **committed the changes itself** (commit `d7bf3ad99`, "Docs: cascade fixes (2 files)").

All of those "fixes" were the same rename, applied in three places:

| Before | After |
|---|---|
| `agent/session_logs.py` | `agent/agent_sessions.py` |
| `bridge/session_logs.py` | `bridge/agent_sessions.py` |

**No such rename happened.** Verified on the filesystem at that commit:

```
$ ls agent/session_logs.py bridge/session_logs.py agent/agent_sessions.py bridge/agent_sessions.py
ls: agent/agent_sessions.py: No such file or directory
ls: bridge/agent_sessions.py: No such file or directory
-rw-r--r--  agent/session_logs.py
-rw-r--r--  bridge/session_logs.py
```

`agent/session_logs.py` is very much alive — PR #2681 *modifies* it, and it is the home of `cleanup_old_snapshots()`, which that PR wires up. The substrate rewrote correct references into references to a module that has never existed.

## Why this is worth fixing rather than shrugging at

1. **It reported `status: "ok"`.** The context file (`.claude/skill-context/do-docs.md`) tells the cascade that `ok` means "proceed to Step 3 for any remaining manual edits" — i.e. treat the substrate's output as done and correct. There was no signal to re-check.
2. **The context file also says "Do not re-commit the substrate's changes — it commits them itself."** That instruction assumes correctness. The substrate self-commits, so a wrong fix lands with no review gate in front of it.
3. **It corrupted a brand-new feature doc that had just been approved.** `docs/features/scheduled-disk-reclaim.md` was reviewed and approved at `cd9f80ddc` minutes earlier. The substrate broke a code reference in it after approval.
[...]

PR #2782, opened 2026-08-13
fix(docs-auditor): migration-context hatch and bare-path existence invariant

### LINKED ISSUE #2744: docs-auditor: migration_context escape hatch misses backticked terms, rewrites correct migration prose into false statements
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/docs-auditor-migration-context-and-bare-paths.md

## Summary

`_detect_stale_term_fixes`'s `migration_context` escape hatch tests **bare substrings** while the docs corpus overwhelmingly writes the terms **backticked**. The hatch therefore never fires on the dominant form, and the auditor rewrites already-correct migration prose into false statements — with `fixes_withheld=0`, so the resulting PR is auto-merge eligible.

## The defect

`reflections/docs_auditor.py:520-527`:

```python
migration_context = (
    f"renamed to {new_term}" in content
    or f"replaced by {new_term}" in content
    or f"now {new_term}" in content
    or f"formerly {old_term}" in content
    or f"Replaces {old_term}" in content
    or f"replaces {old_term}" in content
)
```

Every arm is a bare-substring test. The corpus writes the term inside backticks, so `formerly RedisJob` does not match the actual text ``formerly `RedisJob` ``. The hatch is effectively dead on the most common phrasing.

## Evidence — three live docs still rewritten into false statements

| Doc | Line | Actual text | Hatch arm that should have fired |
|---|---|---|---|
| `docs/features/popoto-redis-expansion.md` | 11 | ``AgentSession` (formerly `RedisJob`)` | `formerly RedisJob` — misses on backticks |
| `docs/guides/agent-session-migration-audit.md` | 17 | ``bridge/message_drafter.py`, formerly `bridge/summarizer.py` — renamed per #1035` | `formerly ...` — misses on backticks |
| `docs/guides/summarizer-output-audit.md` | — | migration prose in the same class | — |

In each case the doc is *already* stating the migration correctly. The auditor rewrites the "formerly X" side into the new term, producing a sentence that says the new name was formerly itself.

## Not introduced by PR #2728

Pre-existing. A control run against `main`'s detector produces an **identical fix set** for these three docs, so this is not a regression from the rename-guard work. It is a distinct defect class: PR #2728's existence invariant constrains *invented paths*, whereas this produces *prose falsity between two names that both exist* — which the invariant cannot catch by design.

PR #2728 hand-fixed one instance of this class (`docs/features/docs-auditor.md`, which triggered all four stale-term detectors on itself) without addressing the generator. The remaining three are untouched.

## Suggested fix

1. Make the hatch **backtick-insensitive** — normalize backticks (and ideally surrounding punctuation) out of both `content` and the probe strings before the substring test, rather than adding more literal arms.
2. Re-run the detector over the three named docs and confirm zero findings.
3. Add regression coverage asserting the hatch fires on the backticked form for each arm.

Worth considering while fixing: the arm list is asymmetric (`Replaces`/`replaces` are cased twice, the others once). A case-insensitive normalized match would collapse all six arms and close that gap too.

## Why it matters

These rewrites ship with `fixes_withheld=0`, so they clear the auto-merge gate added in #2728. The autonomous-wrong-edit surface that remains after that PR is exactly this: prose falsity rather than invented paths.

Refs #2728

---

## Recon Summary

Performed 2026-08-13 against `main` @ `90c0e81e4` (post-#2728 merge `45d0961f9`), reading `reflections/docs_auditor.py` directly and executing the detector against the live corpus.

**Confirmed:**

- **The defect is real and at the cited location.** `_detect_stale_term_fixes` is at `reflections/docs_auditor.py:497-530`; the `migration_context` hatch is at `:520-527`, verbatim as quoted in this issue. All six arms are bare-substring tests. No line drift.
- **All three named docs still produce fixes today.** Executed `_detect_stale_term_fixes` on current `main`:

  | Doc | Fixes proposed |
  |---|---|
  | `docs/features/popoto-redis-expansion.md` | `\bSessionLog\b`→`AgentSession`, `\bRedisJob\b`→`AgentSession` |
  | `docs/guides/agent-session-migration-audit.md` | `\bSessionLog\b`, `\bRedisJob\b`, `\bsession_log\b` |
  | `docs/guides/summarizer-output-audit.md` | `\bSessionLog\b`, `\bRedisJob\b` |

  17 term occurrences across the three docs would be rewritten.
- **`docs/features/docs-auditor.md` produces zero fixes** — consistent with this issue's note that #2728 hand-fixed that one instance.
- **The hatch is document-scoped, not occurrence-scoped.** One migration sentence anywhere in a doc exempts every occurrence of that term in the whole doc. This is load-bearing for any fix design.

**Revised:** — this issue's evidence table is partially incorrect

The proposed fix (backtick normalization) is **necessary but not sufficient**, and one row of the evidence table does not hold:

- `docs/features/popoto-redis-expansion.md:11` — ``` `AgentSession` (formerly `RedisJob`) ``` — **holds.** Backtick normalization makes `formerly RedisJob` fire and suppresses the `RedisJob` rewrite. But the `SessionLog` rewrite on `:7` (*"replacing both the earlier `SessionLog` and `RedisJob` models"*) matches **no arm even after normalization** — `replacing`/`earlier` are not in the arm list. Normalization alone leaves this doc still corrupted.
- `docs/guides/agent-session-migration-audit.md:17` — **does not hold.** The `formerly` on that line refers to `` `bridge/summarizer.py` ``, which is not a `STALE_TERMS` key. No `formerly SessionLog` / `formerly RedisJob` exists anywhere in that doc. Backtick normalization changes nothing for this doc. Its actual migration prose is in *alias* form (`` `SessionLog = AgentSession` ``, `` `RedisJob = AgentSession` ``, `:22`) and *"the old `RedisJob` + `SessionLog` split"* (`:4`) — neither expressible by the current arm shapes.
- `docs/guides/summarizer-output-audit.md` — its migration prose is *arrow* form (`SessionLog → AgentSession`, `:90`). Its `RedisJob` occurrences (`:68`, `:71-74`) are **quoted historical transcript text**, a different class entirely: no migration phrasing will ever appear near them, and rewriting them falsifies a transcript record.

The corpus therefore contains at least five distinct migration-prose shapes the six literal arms cannot express: `X = Y` alias, `X → Y` arrow, "replacing … the earlier X", "the old X + Y split", and quoted-history.

**Confirmed (new finding):** a live corruption that passes the existence invariant

Both `models/session_log.py` and `models/agent_session.py` **exist on disk**. So the `\bsession_log\b`→`agent_session` fix rewrites `` `models/session_log.py` `` → `` `models/agent_session.py` `` in `docs/guides/agent-session-migration-audit.md:22` and `:75`, and the #2728 existence invariant **does not withhold it** (the target exists). The result is a false statement — a doc that says the backward-compat shim lives in a file that is not the shim — shipped with `fixes_withheld=0` and therefore auto-merge eligible. This is the #2744 class landing through a hole the existence invariant cannot cover by design.

**Confirmed (measured):** candidate fix designs

Prototyped against the live corpus. Residual = occurrences still rewritten.

| Design | Residual (of 17) |
|---|---|
| Backticks normalized only (this issue's suggestion) | ~9 |
| + normalized, case-insensitive, document-scoped cue set (`renamed to`/`replaced by`/`now`/`formerly`/`replaces`/`replacing`/`earlier`/`old`/`X = Y`/`X -> Y`/`X → Y`/`alias X`) | 7 |
| + reuse of the existing `_build_line_context` / `_is_documented_deletion` machinery (`:689-747`) to suppress fenced and deletion-context matches | 3 |
| + suppress any match lying inside a path-shaped token (`models/session_log.py`) | **1** (`summarizer-output-audit.md:68`, quoted transcript) |

A purely **line-scoped** exemption was also prototyped and is worse: it re-exposes 8 occurrences that the document-scoped rule correctly exempts.

**Pre-requisites:**

None. `_detect_stale_term_fixes` and its callers are self-contained in `reflections/docs_auditor.py`; existing coverage is `tests/unit/test_docs_auditor_substrate.py::TestStaleTermDictionary` (`:597-625`) and `::TestStaleTermWordBoundary` (`:629-666`).

`TestStaleTermDictionary::test_migration_context_skips_fix` (`:603`) asserts only the **un-backticked** form (`"The SessionLog has been renamed to AgentSession."`) — which is exactly why this defect shipped with green tests.

**Dropped:**

- Rewriting the stale-term channel to be per-occurrence rather than whole-document — measured worse (see table) and a larger change to `_apply_fixes_to_file`'s regex loop.
- Removing `STALE_TERMS` entries to dodge the problem — that abandons the detector's purpose.

### LINKED ISSUE #2759: docs_auditor existence invariant misses bare filenames: _PATH_REF_RE requires a directory segment, so the #2711 corruption class survives for unprefixed names
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/docs-auditor-migration-context-and-bare-paths.md

> **Context.** `reflections/docs_auditor.py` rewrites documentation references automatically. Issue #2711 was a live corruption: it invented a module rename (`agent/session_logs.py` → `agent/agent_sessions.py`) that never happened, applied it five times, and committed it. PR #2728 (merged `45d0961f9`) fixed that by adding an **existence invariant** — before writing a substitution, check the replacement path actually exists on disk, and withhold the fix if it does not.

## Problem

The existence invariant only sees paths containing a directory separator. Bare filenames are invisible to it.

`_PATH_REF_RE` at [`reflections/docs_auditor.py:534`](https://github.com/tomcounsell/ai/blob/main/reflections/docs_auditor.py#L534):

```python
_PATH_REF_RE = re.compile(r"(?:[\w.-]+/)+[\w.-]+\.(?:py|md)")
```

The `(?:[\w.-]+/)+` group requires **one or more** `dir/` segments. Verified on merged `main`:

```
agent/session_logs.py            -> True    (checked)
docs/features/x.md               -> True    (checked)
agent_session.py                 -> False   (invisible)
README.md                        -> False   (invisible)
```

The two sibling patterns at `:454` and `:761` (`` `((?:[\w.-]+/)+[\w.-]+\.py)` ``) carry the same requirement.

A substitution whose *before* and *after* text are both bare filenames is therefore never existence-checked, in either direction. The invariant reports nothing to withhold and the fix is applied.

## Why it matters

This is the #2711 corruption class, still open through the gap:

1. A detector proposes replacing `` `session_logs.py` `` with `` `agent_sessions.py` `` — no directory prefix, so the invariant does not fire.
2. `agent_sessions.py` does not exist. The invariant would have caught it had the reference been written `agent/agent_sessions.py`.
3. **#2726 is still open**, so `docs_auditor` remains its own committer with no review gate. The wrong reference self-commits, exactly as `d7bf3ad99` did.

Bare filenames are common in prose docs and in README index entries, which is precisely the corpus `_detect_readme_broken_entries` operates on.

## Severity

No observed instance. This is a coverage hole in a just-shipped guard, found by adversarial critique of #2711's plan rather than by a failure. It should be planned as a gap-closure, not an incident.

## Solution sketch

Make the invariant see bare filenames. Open questions for the planner:

- Relaxing to `(?:[\w.-]+/)*[\w.-]+\.(?:py|md)` (`*` instead of `+`) matches bare names, but a bare filename is **not resolvable to a single path** — `README.md` exists in many directories. What does "exists" mean for an unanchored name? Candidate answers: resolve relative to the containing doc, glob the repo and require exactly one match, or treat multi-match as withhold.
- Withholding more aggressively has a cost the tracker already knows about: #2729 documents that a run with any withheld fix becomes permanently auto-merge-ineligible and gets stale-closed at day 14. Widening the invariant without addressing that turns a precision win into discarded work. **Sequence this after or alongside #2729.**
- Should the `:454` and `:761` patterns change in lockstep, or do their call sites have different safety requirements? Check before applying one fix to all three.

## Acceptance criteria

- [ ] A proposed substitution to a nonexistent bare filename is withheld, with a test.
- [ ] The ambiguous-bare-name case (a filename matching multiple real paths) has a defined, tested behavior.
- [ ] All three patterns audited; any left unchanged has a recorded reason.
- [ ] No increase in withheld-fix rate on a real rotation without #2729's escalation path in place.

## Downstream

Consumed by `/do-plan` to produce `docs/plans/{slug}.md`, then `/do-build`.

Related: #2711 (the corruption this guard was added for), #2726 (no review gate — why this still self-commits), #2729 (withheld-fix disposal), #2725.

---

## Recon Summary

Performed 2026-08-13 against `main` @ `90c0e81e4` (post-#2728 merge `45d0961f9`), reading `reflections/docs_auditor.py` directly and measuring the widened pattern against the live corpus.

**Confirmed:**

- **`_PATH_REF_RE` is at `reflections/docs_auditor.py:534`, verbatim as quoted.** No line drift. The sibling patterns are at `:454` (`_detect_renamed_symbol_fixes`) and `:761` (`_detect_deleted_target_issues`), both `` r"`((?:[\w.-]+/)+[\w.-]+\.py)`" `` — also verbatim.
- **The visibility table reproduces exactly.** Executed against `main`: `agent/session_logs.py` → matched; `docs/features/x.md` → matched; `agent_session.py` → **not matched**; `README.md` → **not matched**.
- **The invariant is one-directional and additive-only.** `_absent_new_path_refs` (`:537-547`) validates only `set(_PATH_REF_RE.findall(candidate)) - original_refs` — refs the rewrite *newly introduces*. Pre-existing refs are never re-validated. This is load-bearing: widening the regex widens `original_refs` symmetrically, so it costs nothing on text the auditor did not write.
- **The corruption class is live and reachable today through a second door.** Both `models/session_log.py` and `models/agent_session.py` exist on disk, so the `\bsession_log\b`→`agent_session` stale-term fix rewrites `` `models/session_log.py` `` → `` `models/agent_session.py` `` in `docs/guides/agent-session-migration-audit.md:22` and `:75` and the invariant passes it (target exists). Related to #2744; recorded here because it shows the invariant's coverage boundary from the other side.

**Confirmed (measured):** the cost of widening `+` → `*`.

Ran the widened pattern over every `docs/**/*.md` except `docs/plans/`:

| Metric | Count |
|---|---|
| Bare-filename refs newly visible | 1557 |
| …resolving to exactly one repo path | 1202 |
| …resolving to multiple repo paths (ambiguous) | 218 |
| …resolving to no repo path at all | 137 |
| Distinct basenames in the repo with >1 owner | 108 |

Because the invariant is additive-only, **all 1557 are pre-existing and none are re-validated**. The 137 absent names (`PROGRESS.md`, `DESIGN.md`, `manage.py`, `job_queue.py`, `AGENTS.md`, `file.py`, …) are illustrative/foreign-repo names in prose and would produce zero new withholds. The measured withheld-rate impact of the widening on today's corpus is therefore **zero**; withholds only appear when the auditor tries to *write* a bare name that exists nowhere — which is precisely the #2711 class.

**Revised:**

- **"#2726 is still open" is stale.** #2726 was closed `2026-08-13T08:03:18Z` as `NOT_PLANNED`, duplicate of **#2739**, whose plan is `docs/plans/docs-auditor-review-gate.md` (status `Planning`). The "no review gate, so it self-commits" premise still holds — nothing has shipped — but ownership has moved to #2739.
- **The #2729 sequencing concern is owned elsewhere.** `docs/plans/docs-auditor-review-gate.md` explicitly closes #2725, #2726 **and #2729**, including the withheld-PR stale-close escalation. This issue's acceptance criterion 4 does not require blocking on it: see the measurement above — the widening adds no withholds against the current corpus, and the #2744 work landing in the same lane strictly *reduces* the number of proposed fixes.
- **`git ls-files` is the right existence oracle for bare names, not `Path.rglob`.** It respects `.gitignore` (so `.venv/`, `.worktrees/`, `node_modules/` are excluded for free), is one subprocess, and with `--cached --others --exclude-standard` also covers untracked-but-not-ignored files.

**Pre-requisites:**

None blocking. `docs/plans/docs-auditor-review-gate.md` (#2739) is an **overlap, not a dependency** — it edits the same file but in disjoint regions (`_commit_current_branch` `:1192`, `_push_branch_and_pr` `:1338`, `_pr_is_auto_merge_eligible` `:1884`, sweeper `:1968+`), whereas this work is confined to the detector/invariant block `:497-547`. Neither is built yet; whichever lands second rebases.

Existing coverage: `tests/unit/test_docs_auditor_substrate.py::TestExistenceInvariant` (`:669-800`). Every one of its fixtures uses `dir/file.py`-shaped paths, which is why the bare-name hole shipped with green tests.

**Dropped:**

- **Changing `:454` and `:761` in lockstep.** Recommended disposition: **leave both unchanged**, recorded reason — they are *detector* regexes governing what the auditor *proposes*, not the safety path. Widening them increases the candidate set, which spends the `GIT_LOG_FOLLOW_CAP = 10` per-run `git log --follow` budget and the per-run issue-filing cap of 5 on bare names that are not resolvable to a single path anyway. That is a scope and volume change with its own failure modes, not a safety fix. `_PATH_REF_RE` at `:534` is the only one of the three on the write path.
- **Treating an ambiguous bare name as a withhold.** Recommended ruling: **≥1 match ⇒ passes; 0 matches ⇒ withhold.** The invariant's question is "does this name denote something real", not "is it unambiguous". Ambiguity never produced the #2711 corruption — that was a name existing nowhere — and 218 ambiguous refs in the corpus against 108 multi-owner basenames means a withhold-on-ambiguous rule would fire constantly for no safety gain, straight into #2729's disposal problem.

valorengels:
## Summary

Closes two holes in `reflections/docs_auditor.py` that let an unreviewed autonomous committer write false statements into `main`. Both live in the same ~50 lines, both are the "autonomous writer makes a doc false" class, and fixing one without the other leaves the PR half-safe.

- **#2744** — the `migration_context` escape hatch never fired on the corpus's dominant phrasing (the six bare substring tests could not see `` formerly `RedisJob` `` because the corpus backticks its terms). Three live docs queued rewrites turning correct migration prose into sentences saying the new name was formerly itself, all with `fixes_withheld = 0` so auto-merge cleared them unread.
- **#2759** — `_PATH_REF_RE` required at least one directory segment, so bare filenames (`README.md`, `agent_session.py`) were invisible to the existence invariant. That is the #2711 corruption shape minus a directory prefix, and #2711 was a real committed corruption (`d7bf3ad99`).

## Four fail-closed gates now guard every rewrite

1. **Migration-context hatch** — `_normalize_prose` (strips backticks, collapses whitespace, lowercases; cue matching only, never output) plus a generated two-tier cue set. Directed cues per `(old, new)` pair, plus `_MIGRATION_CUE_WORDS` that fire only when the new term also appears in the document (the anti-over-exemption guard). Deliberately **document-scoped**: spike-2 measured line-scoping as strictly worse, re-exposing 8 occurrences real prose correctly exempts, because migration context sits in a different sentence from the term.
2. **Context suppression** — `_build_line_context` / `_is_documented_deletion` (fenced blocks, deletion headings, adjacent deletion prose) wired into stale terms.
3. **Path-token suppression** — a match inside a `(?:[\w.-]+/)*[\w.-]+\.(?:py|md)` token is never rewritten. This closes `models/session_log.py` -> `models/agent_session.py`, which the existence invariant provably cannot catch because both files exist.
4. **Existence invariant** — `_PATH_REF_RE` widened `+` -> `*`; `_absent_new_path_refs` gained `doc_path` and resolves bare refs doc-relative first, then via a per-run `git ls-files --cached --others --exclude-standard` basename index.

**Gates 2 and 3 are computed at APPLY time, never at detection time.** `_apply_fixes_to_file` applies the literal `fixes` list first, including `_detect_readme_broken_entries`' `new == ""` whole-line-delete sentinel, and only then runs the regex fixes over the already-mutated text. A detection-time line index is stale the moment a line is deleted ahead of it, and it fails *silently* (a plausible wrong rewrite, not an error) which is exactly the corruption class this PR exists to close. The replacement callable is built inside `_apply_fixes_to_file`'s regex loop and re-derives context from `match.string` / `match.start()`. `_detect_stale_term_fixes` still returns `(re.Pattern, str)`; the channel contract is unchanged and `test_fixes_travel_on_the_regex_channel` passes byte-for-byte unmodified.

Suppressed matches still increment `subn`'s counter, so suppressions are tallied in a closure cell and `applied` is reported as `subn_count - suppressed_count`, with an `if candidate == new_text: continue` guard for the all-suppressed case. `_reject` still receives the replacement *string*, so withheld records stay readable in the PR body, findings summary, and warning log.

## Rulings recorded (with evidence, in code and in the feature doc)

- **Ambiguity is not a withhold.** `>=1 match => passes` (DEBUG-logged when `>1`); `0 matches => withhold`. The invariant asks "does this name denote something real", not "is it unambiguous". 218 ambiguous refs over 108 multi-owner basenames; ambiguity never produced #2711, a name existing nowhere did.
- **The sibling patterns in `_detect_renamed_symbol_fixes` and `_detect_deleted_target_issues` stay narrower**, with the reason in a code comment naming #2759. They are detector-side, not write-path; widening them spends the `GIT_LOG_FOLLOW_CAP` and per-run issue-cap budgets on names not resolvable to a single path.
- **This lane did not need to wait on #2729**, because the widening's measured new-withhold count is zero.

## Measured outcome

| Doc | Before | After |
|---|---|---|
| `docs/features/popoto-redis-expansion.md` | 3 occurrences rewritten | **0** |
| `docs/guides/agent-session-migration-audit.md` | 8 occurrences rewritten | **0** |
| `docs/guides/summarizer-output-audit.md` | 7 occurrences rewritten | **1** (the accepted residual) |
| **total** | **18** | **1** |

**Corrected after review (this table previously read `0` and `18 -> 0`).** The plan predicted and accepted exactly one residual: the quoted-transcript line at `summarizer-output-audit.md:68`, *"**Example 3** (10:52, fixing RedisJob refs):"*. An earlier revision of this PR reported that residual as not firing and attributed the zero to the document-scoped hatch's second cue tier working as designed. That attribution was wrong. The tier-2 cue-word test was an unanchored substring match, so `old` fired from inside `summarize_threshold` and exempted the whole document by accident. Anchoring the cue words (commit `7a9b1e89e`) restores the designed behavior and returns the residual to 1, which is what the plan's success criterion specifies: *"at most **one** changed occurrence (the accepted `:68` residual), down from five."*

The `:68` residual is filed as a follow-up comment on #2744 per the plan's No-Gos section.

**The detector is quiet, not dead.** Across all `docs/**/*.md` outside `docs/plans/`, exactly one stale-term proposal survives — the accepted `summarizer-output-audit.md:68` residual, re-measured after the cue-word anchoring fix — and the four docs still carrying a `STALE_TERMS` key are all genuine migration/audit prose where any rewrite would produce a false sentence (alias lines would become `AgentSession = AgentSession`; the shim-location line would point at a file that is not the shim). A synthetic no-migration-context sentence still queues a fix, so the drop is attributable to suppression working rather than detection breaking.

`TestWithheldRateNonRegression` proves #2759 AC4 mechanically: it drives `_apply_fixes_to_file` directly inside a disposable `git worktree add --detach` (never the live checkout, which it would rewrite), and is self-baselining — both arms measured in one run over one corpus snapshot, `_PATH_REF_RE` monkeypatched to the narrow `+` form vs the shipped widened `*` form. Result: narrow 1025 visible refs / 0 withheld, widened 1569 visible refs / 0 withheld. `after <= before` holds, with explicit non-vacuity assertions so a future corpus change that hollows the measurement out fails loudly instead of passing green.

## Demonstrated-red paper trail

Every behavioral fix landed with a test proven **failing first**, before any production edit (commit `739fc03af`, tests only). Verbatim red output:

**Run 1** — `POPOTO_TEST_DB=9 ./scripts/pytest-clean.sh tests/unit/test_docs_auditor_substrate.py -k "StaleTerm" -q` -> **10 failed, 12 passed**

```
FAILED ...::TestStaleTermDictionary::test_migration_context_skips_fix[The `SessionLog` has been renamed to `AgentSession`.]
FAILED ...::TestStaleTermDictionary::test_migration_context_skips_fix[`RedisJob` was replaced by `AgentSession` in the same pass.]
FAILED ...::TestStaleTermDictionary::test_migration_context_skips_fix[Formerly `RedisJob`, this model is now `AgentSession`.]
FAILED ...::TestStaleTermDictionary::test_migration_context_skips_fix[The compat shim declares `SessionLog = AgentSession` for old imports.]
FAILED ...::TestStaleTermDictionary::test_migration_context_skips_fix[Rename map: `SessionLog` -> `AgentSession` across all call sites.]
FAILED ...::TestStaleTermDictionary::test_migration_context_skips_fix[Rename map: `SessionLog` → `AgentSession` across all call sites.]
FAILED ...::TestStaleTermDictionary::test_migration_context_skips_fix[`AgentSession` lands, replacing both the earlier `SessionLog` and `RedisJob` models.]
FAILED ...::TestStaleTermWordBoundary::test_apply_leaves_session_logs_path_untouched
FAILED ...::TestStaleTermWordBoundary::test_stale_term_inside_fenced_block_is_not_rewritten
FAILED ...::TestStaleTermWordBoundary::test_suppression_survives_an_earlier_line_deletion
```

The line-shift fixture is the most informative failure and confirms the apply-time design ruling empirically:

```
tests/unit/test_docs_auditor_substrate.py:801: in test_suppression_survives_an_earlier_line_deletion
    assert "`SessionLog` model no longer exists" in text
E   AssertionError: assert '`SessionLog` model no longer exists' in
    '# Feature Index\n\n- [Kept](kept.md)\n\nThe `AgentSession` model no longer exists.\n
     It was superseded during the Popoto consolidation.\n'
```

The earlier `- [Gone](gone.md)` line was deleted by the `new == ""` sentinel *and* the later stale term was rewritten in the same apply, so the fixture genuinely exercises the index-shift window. The assertion is on file content, not on an exception, because detection-time index suppression fails silently.

**Run 2** — `POPOTO_TEST_DB=9 ./scripts/pytest-clean.sh tests/unit/test_docs_auditor_substrate.py -k "Existence or Withheld" -q` -> **3 failed, 21 passed**

```
FAILED ...::TestExistenceInvariant::test_fix_introducing_absent_bare_name_is_withheld
    assert applied == 0   /   E   assert 1 == 0
FAILED ...::TestExistenceInvariant::test_ambiguous_bare_name_passes_and_debug_logs
    E   AssertionError: ambiguous bare-name resolution must be DEBUG-logged
FAILED ...::TestWithheldBlocksAutoMerge::test_bare_name_withhold_propagates_to_pr_body_telegram_and_liveness
    assert audit_result["fixes_withheld"] == 1   /   E   assert 0 == 1
```

All 13 are green after the fixes. Two further new tests (`test_bare_name_resolvable_in_doc_directory_passes`, `test_preexisting_absent_bare_name_is_never_revalidated`) are green against `main` by construction: they assert the *absence* of a withhold, and `main` withholds nothing for bare names. They are the over-withholding guard and go red if bare refs resolve against the repo root instead of the doc's directory, or if the `original_refs` subtraction that makes the widening additive-only is dropped.

One pre-existing test, `TestExistenceInvariant::test_regex_channel_is_also_guarded`, needed a fixture repair: its `\breal\b` match landed inside the path token `agent/real.py`, so gate 3 now pre-empts gate 4 and there was nothing left to withhold. The fixture was changed so the match sits in prose while the replacement still introduces an absent path ref. Name, intent, and assertions are unchanged, and a comment records why it cannot be simplified back.

## Testing

- [x] `POPOTO_TEST_DB=9 ./scripts/pytest-clean.sh tests/unit/test_docs_auditor_substrate.py -q` -> **136 passed**
- [x] All 14 rows of the plan's Verification table pass
- [x] Lint/format clean (`ruff check`, `ruff format --check`)
- [x] No `xfail` markers, no `except Exception: pass` introduced

## Documentation

- [x] `docs/features/docs-auditor.md` rewritten around the four gates, with the document-scope ruling, the apply-time-suppression rationale, the bare-name resolution order, the ambiguity ruling, and the sibling-pattern reason
- [x] `_detect_stale_term_fixes` docstring rewritten (its old "stops short of *never rewrites a path*" sentence is now false and is gone)
- [x] `_absent_new_path_refs` docstring carries the resolution order and ambiguity ruling
- [x] `docs/features/README.md` index line verified accurate, no change needed

## Scope

The diff touches exactly `reflections/docs_auditor.py`, `tests/unit/test_docs_auditor_substrate.py`, and `docs/features/docs-auditor.md`. `_pr_is_auto_merge_eligible`, `_push_branch_and_pr`, `_commit_current_branch`, and `run_docs_branch_sweeper` are untouched (#2739's region, verified by an anti-criterion grep).

## Definition of Done
- [x] Built: four gates implemented and working
- [x] Tested: 136 passing, every fix demonstrated red first
- [x] Documented: feature doc and inline docstrings updated

Closes #2744
Closes #2759

### DIFF
--- a/docs/features/docs-auditor.md
+++ b/docs/features/docs-auditor.md
@@ -83,10 +83,12 @@ Gates 1 to 3 govern stale-term rewrites. Gate 4 governs every auto-fix from
 every detector.
 
 Over the three docs the auditor previously corrupted, false rewrites went from
-**18 occurrences to zero**. Across all of `docs/**/*.md` outside `docs/plans/`,
-surviving stale-term proposals are zero; the four docs that still contain a
-`STALE_TERMS` key are all genuine migration or audit prose where any rewrite
-would produce a false sentence. The detector is quiet, not dead: a synthetic
+**18 occurrences to 1**. Across all of `docs/**/*.md` outside `docs/plans/`,
+exactly one stale-term proposal survives — the accepted quoted-transcript
+residual at `docs/guides/summarizer-output-audit.md:68`, which the plan predicted
+and accepted; the four docs that still contain a `STALE_TERMS` key are all
+genuine migration or audit prose where any rewrite would produce a false
+sentence. The detector is quiet, not dead: a synthetic
 sentence carrying a stale term and no migration context still queues a fix, and
 a test pins that.
 
--- a/docs/features/docs-auditor.md
+++ b/docs/features/docs-auditor.md
@@ -59,8 +59,36 @@ writes the SDLC stage marker and updates the plan frontmatter.
 The dictionary's current entries are model and field renames: `SessionLog` and
 `RedisJob` were renamed to AgentSession; the `session_log` and `redis_job`
 fields were renamed to agent_session. Stating the mapping in that form is also
-what keeps this page immune to its own detector — see the `migration_context`
-escape hatch below.
+what keeps this page immune to its own detector — see the migration-context
+hatch (gate 1) below.
+
+### Four gates guard every rewrite
+
+The auditor is an unreviewed autonomous committer: the rotation caller opens a
+PR the branch sweeper can auto-merge with nobody in between. That makes its two
+failure modes asymmetric. A stale term that survives is a cosmetic miss. A
+sentence the auditor falsifies is a corruption a human later trusts. Every gate
+below therefore errs toward **not** rewriting.
+
+Four fail-closed gates sit between a proposed rewrite and the disk, in order:
+
+| # | Gate | Where it runs |
+|---|------|---------------|
+| 1 | Migration-context hatch | `_detect_stale_term_fixes` (detection) |
+| 2 | Fence / heading / deletion-prose suppression | `_apply_fixes_to_file` (apply) |
+| 3 | Path-token suppression | `_apply_fixes_to_file` (apply) |
+| 4 | Existence invariant | `_apply_fixes_to_file` (apply) |
+
+Gates 1 to 3 govern stale-term rewrites. Gate 4 governs every auto-fix from
+every detector.
+
+Over the three docs the auditor previously corrupted, false rewrites went from
+**18 occurrences to zero**. Across all of `docs/**/*.md` outside `docs/plans/`,
+surviving stale-term proposals are zero; the four docs that still contain a
+`STALE_TERMS` key are all genuine migration or audit prose where any rewrite
+would produce a false sentence. The detector is quiet, not dead: a synthetic
+sentence carrying a stale term and no migration context still queues a fix, and
+a test pins that.
 
 #### Stale terms are word-anchored
 
@@ -69,12 +97,9 @@ longer run of word characters: `session_log` does not match inside
 `agent/session_logs.py` or `session_log_writer`, and `SessionLog` does not match
 inside `SessionLogs`.
 
-That is the whole guarantee, and it stops short of "never rewrites a path".
-`/`, `.`, and `-` are all word boundaries, so a key that *equals* an entire path
-segment still matches and is still rewritten — `models/session_log.py` becomes
-`models/agent_session.py`. The existence invariant below is the only thing that
-catches such a rewrite, and only when the rewritten path is absent from the
-working tree; if both paths happen to exist, the rewrite is accepted.
+Word-anchoring on its own does not deliver "never rewrites a path". `/`, `.`,
+and `-` are all word boundaries, so a key equal to a whole path segment still
+matches. Gate 3 is what closes that.
 
 Detection and application share one matching semantics: `_detect_stale_term_fixes`
 returns compiled `(pattern, replacement)` pairs on a dedicated `regex_fixes`
@@ -82,32 +107,141 @@ channel, applied by `_apply_fixes_to_file` through `pattern.subn()`. The literal
 `fixes` channel used by the three rename detectors is untouched, so the
 `new == ""` line-delete sentinel keeps its exact-line-equality semantics.
 
-The `migration_context` escape hatch is unchanged: a doc that already says
-"renamed to X", "replaced by X", "now X", "formerly Y", or "replaces Y" gets no
-fix for that term.
+#### Gate 1: the migration-context hatch is document-scoped
 
-#### The existence invariant
+`_has_migration_context` asks whether the **whole document** records the
+`old_term` → `new_term` migration, over text run through `_normalize_prose`
+(backticks stripped, whitespace collapsed, lowercased). That normalization is
+used for cue matching only and never to produce output; it is lossy by design.
+It is also the point: the corpus writes every identifier backticked, so a bare
+substring test for `formerly RedisJob` never sees the prose humans actually
+wrote.
+
+Two tiers of cue:
+
+1. **Directed cues**, generated per `(old_term, new_term)` pair by
+   `_migration_cues`: `renamed to X`, `replaced by X`, `now X`, `formerly Y`,
+   `replaces Y`, `replacing Y`, `earlier Y`, `old Y`, `alias`, `Y = X`,
+   `Y -> X`, `Y → X`. Each names both (or the specific) terms, so it is
+   conclusive alone. Generating them from the pair means adding a `STALE_TERMS`
+   entry costs no edit here.
+2. **Generic cue words** (`_MIGRATION_CUE_WORDS`: `renamed`, `replaced`,
+   `replacing`, `formerly`, `earlier`, `old`, `alias`, `superseded`, …) that
+   fire **only when the new term also appears somewhere in the document**. That
+   second condition is the anti-over-exemption guard: prose mentioning a stale
+   name without ever naming its successor is not a migration record and still
+   queues a fix. Tier 2 is what catches real corpus phrasing whose cue and term
+   sit in different clauses, such as *"replacing both the earlier `SessionLog`
+   and `RedisJob` models"* for `RedisJob`, which no directed cue names.
+
+**The scope is the document, deliberately.** Line-scoping (one exemption
+decision per occurrence) was measured and is strictly worse: it re-exposes 8
+occurrences the document scope correctly exempts, because in real prose the
+migration context sits in a *different* sentence from the term it explains. Do
+not "improve" this into a per-occurrence rule.
+
+#### Gates 2 and 3 are computed at apply time, never at detection time
+
+`_detect_stale_term_fixes` returns `(compiled pattern, replacement string)` and
+nothing else; the channel contract is unchanged. The suppression callable is
+built inside `_apply_fixes_to_file`'s regex loop by `_make_stale_term_replacer`,
+which wraps that plain string.
+
+That placement is load-bearing, not stylistic. `_apply_fixes_to_file` applies the
+literal `fixes` list **first** (including `_detect_readme_broken_entries`'
+`new == ""` whole-line-delete sentinel) and only then runs the regex fixes over
+the already-mutated text. Any line index computed at detection time is stale the
+moment a line ahead of it disappears, and it fails **silently**: a
+plausible-looking wrong rewrite rather than an error, which is precisely the
+corruption class these gates exist to close. The callable instead re-derives
+context from `match.string` (the text currently being rewritten) and
+`match.start()` (the live offset into it), so there is no index to go stale.
+
+Keeping the callable local also keeps `_reject` receiving the replacement
+*string*, so a withheld record stays readable in the PR body, the findings
+summary, and the warning log instead of rendering as `<function ... at 0x...>`.
+
+**Gate 2** reuses `_build_line_context` / `_is_documented_deletion`, the same
+machinery the deleted-target detector uses (described below): fenced code
+blocks, deletion or migration headings, and deletion prose on an adjacent line.
+
+**Gate 3** suppresses any match lying wholly inside a `_PATH_TOKEN_RE` token
+(`(?:[\w.-]+/)*[\w.-]+\.(?:py|md)`), scanned within the match's own line so the
+cost is independent of document size. This is the gate that stops
+`models/session_log.py` from becoming `models/agent_session.py`, a corruption
+the existence invariant provably cannot catch, because both files exist.
+
+A suppressed match returns unchanged, but `subn` still counts it as a
+substitution. Suppressions are therefore tallied in a closure cell and
+subtracted from the reported `applied` count, and an all-suppressed fix is
+skipped outright rather than booked as a no-op write.
+
+#### Gate 4: the existence invariant
 
 Every auto-fix, regardless of detector, is subject to one post-condition before
-anything is written: **the auditor may not introduce a slash-shaped `.py`/`.md`
-repo-path reference that is absent from the working tree.**
+anything is written: **the auditor may not introduce a `.py`/`.md` reference
+that is absent from the working tree.**
 
-That is the exact shape `_PATH_REF_RE` matches, and it is narrower than "any
-reference to a file". Two forms are deliberately **not** covered: paths with
-other extensions (`.sh`, `.ts`, `.json`), and the dotted module form
-(`bridge.session_log`) — the same class as the #2711 incident. A fix that
-introduces only those forms passes the invariant with `withheld=0`.
+`_PATH_REF_RE` is `(?:[\w.-]+/)*[\w.-]+\.(?:py|md)`. The directory segment is
+optional, so bare filenames (`agent_session.py`, `README.md`) are visible to the
+invariant alongside `dir/file.md` forms. A bare name existing nowhere is the
+#2711 corruption shape minus its directory prefix. Two forms are still
+deliberately **not** covered: other extensions (`.sh`, `.ts`, `.json`), and the
+dotted module form (`bridge.session_log`). A fix that introduces only those
+passes the invariant with `withheld=0`.
 
-`_apply_fixes_to_file` simulates each fix, collects the `dir/file.{py,md}`-shaped
-references the candidate text would newly introduce, and resolves each against
-`repo_root`. If any is absent, that fix is rejected.
+`_apply_fixes_to_file` simulates each fix, collects the references the candidate
+text would newly introduce, and hands them to `_absent_new_path_refs` together
+with the doc's own path. If any is absent, that fix is rejected.
+
+**Resolution order.** A ref containing `/` resolves against `repo_root` and
+nothing about it changed. A **bare** ref resolves first against the doc's own
+directory (`repo_root / doc_path.parent / ref`), because a bare name in prose is
+most often a sibling, and then against a basename index built from
+`git ls-files --cached --others --exclude-standard`. That is one subprocess, it
+respects `.gitignore` for free, and it counts files added but not yet committed,
+which matters because a doc may reference a file added in the same change. The
+index is memoized per resolved repo root and cleared at the top of every
+`audit()` run, so a long-lived process never answers from a pre-commit snapshot.
+On subprocess failure it logs a warning and degrades to doc-relative resolution
+alone.
+
+**Ambiguity is not a withhold.** `>=1` owner means the name denotes something
+real and the fix passes; `>1` owners is DEBUG-logged and allowed through; only
+`0` owners is absent. The invariant's question is "does this name denote
+something real", not "is it unambiguous". Ambiguity never produced the #2711
+corruption. A name existing *nowhere* did. The corpus carries 218 ambiguous
+refs across 108 multi-owner basenames, so withholding on ambiguity would fire
+constantly for no safety gain.
+
+Making the directory segment optional costs nothing on already-written prose.
+`_absent_new_path_refs` validates only `findall(candidate) - original_refs`, and
+`original_refs` is computed with the same pattern, so both sides widen
+symmetrically and pre-existing references are never re-validated. Measured
+self-baselining over `docs/features/*.md` (narrow `+` arm versus shipped `*` arm
+in one run), the widening produces **zero** new withholds.
+
+#### Why the sibling patterns stay narrower
+
+`_detect_renamed_symbol_fixes` and `_detect_deleted_target_issues` each carry
+their own `` `((?:[\w.-]+/)+[\w.-]+\.py)` `` pattern that still *requires* a
+directory segment. The asymmetry with `_PATH_REF_RE` is deliberate, and a
+comment at each site records why.
+
+Those two are **detector** patterns, governing what the auditor *proposes*.
+`_PATH_REF_RE` is the only one of the three on the write path. Widening the
+detectors enlarges the candidate set and spends the `GIT_LOG_FOLLOW_CAP` per-run
+`git log --follow` budget and the per-run issue cap on bare names that are not
+resolvable to a single path anyway. That is a scope and volume change with its
+own failure modes, not a safety fix.
 
 - **Attributed per term, not per occurrence.** Only the offending fix is
   dropped; valid sibling fixes in the same file still apply, and the file is
   never abandoned wholesale. But a fix is one `(old, new)` pair applied to the
   whole document — `str.replace` and `pattern.subn()` both rewrite *every*
-  occurrence — so a single path-shaped hit withholds that term's legitimate
-  prose hits in the same file too.
+  occurrence — so a single absent-target hit withholds that term's legitimate
+  prose hits in the same file too. Path-shaped occurrences no longer reach this
+  gate at all: gate 3 suppresses them first.
 - **References already present in the document are never re-validated.** The
   invariant constrains what the auditor *adds*, so it cannot reject a fix over
   pre-existing prose that names a long-gone module. The residual hole: if a
@@ -392,6 +526,18 @@ the `_write_liveness` 4-arg/5-arg positional contract (`fixes_withheld` is a
 trailing keyword param, so that contract is unchanged); `TestVaultDeadCodeRemoved`
 asserts `DEFAULT_VAULT_WEIGHT`, `vault_weight`, and `_vault_field` are gone.
 
+The four gates are covered by `TestStaleTermDictionary` (cue tiers across
+backticked, cased, alias, and arrow forms; the channel stays
+`(re.Pattern, str)`), `TestStaleTermWordBoundary` (path-token suppression, and a
+`README.md` fixture where an earlier line is deleted by the `new == ""` sentinel
+before a later stale term is evaluated, the case detection-time indices fail
+silently), `TestExistenceInvariant` (bare-name withhold, doc-relative
+resolution, ambiguous-but-present pass, no re-validation of pre-existing refs),
+and `TestWithheldBlocksAutoMerge` (a bare-name withhold reaching the PR body,
+Telegram, and liveness). `TestWithheldRateNonRegression` self-baselines the
+narrow and widened `_PATH_REF_RE` arms in one run inside a disposable detached
+`git worktree`, asserting the widening adds no withholds.
+
 ```bash
 pytest tests/unit/test_docs_auditor_substrate.py -v
 ```
@@ -406,3 +552,5 @@ pytest tests/unit/test_docs_auditor_substrate.py -v
 - Issue #1247 — design and rollout plan
 - Issue #1249 — memory refresh hook (forward-compat target)
 - Issue #2084 — vault↔site/docs drift detector, reflection enable, xref-orphan sweep
+- Issues #2744 / #2759 — the migration-context hatch, apply-time suppression, and
+  the bare-name existence invariant (`docs/plans/docs-auditor-migration-context-and-bare-paths.md`)
--- a/docs/features/docs-auditor.md
+++ b/docs/features/docs-auditor.md
@@ -126,13 +126,24 @@ Two tiers of cue:
    conclusive alone. Generating them from the pair means adding a `STALE_TERMS`
    entry costs no edit here.
 2. **Generic cue words** (`_MIGRATION_CUE_WORDS`: `renamed`, `replaced`,
-   `replacing`, `formerly`, `earlier`, `old`, `alias`, `superseded`, …) that
-   fire **only when the new term also appears somewhere in the document**. That
-   second condition is the anti-over-exemption guard: prose mentioning a stale
-   name without ever naming its successor is not a migration record and still
-   queues a fix. Tier 2 is what catches real corpus phrasing whose cue and term
-   sit in different clauses, such as *"replacing both the earlier `SessionLog`
-   and `RedisJob` models"* for `RedisJob`, which no directed cue names.
+   `replacing`, `formerly`, `earlier`, `old`, `alias`, `superseded`, …), matched
+   **word-anchored** through the precompiled `_MIGRATION_CUE_WORD_RE`
+   alternation, that fire **only when the new term also appears somewhere in the
+   document**. Tier 2 is what catches real corpus phrasing whose cue and term sit
+   in different clauses, such as *"replacing both the earlier `SessionLog` and
+   `RedisJob` models"* for `RedisJob`, which no directed cue names.
+
+   **Both halves of tier 2 are the anti-over-exemption guard, and both are
+   load-bearing.** Requiring the new term means prose that mentions a stale name
+   without ever naming its successor is not a migration record and still queues a
+   fix. Word-anchoring the cue words means a cue cannot fire from inside an
+   unrelated word: `old` is a substring of *threshold*, *holds*, *placeholder*,
+   *bold* and *household*, all common in this corpus, and with a bare substring
+   test tier 2 collapses into "the document mentions the new term somewhere".
+   That was not hypothetical — `docs/guides/summarizer-output-audit.md` was
+   exempted for `RedisJob` solely because it contains `summarize_threshold`.
+   `TestStaleTermDictionary::test_cue_word_inside_a_larger_word_does_not_exempt`
+   pins it.
 
 **The scope is the document, deliberately.** Line-scoping (one exemption
 decision per occurrence) was measured and is strictly worse: it re-exposes 8
--- a/reflections/docs_auditor.py
+++ b/reflections/docs_auditor.py
@@ -535,6 +535,15 @@ _MIGRATION_CUE_WORDS = (
     "supersedes",
 )
 
+# Word-anchored alternation over the cue words. Anchoring is load-bearing, not
+# cosmetic: an unanchored substring test fires inside unrelated words ("old" is a
+# substring of threshold, holds, placeholder, bold, household), which collapses
+# tier 2 of ``_has_migration_context`` into "the document mentions the new term"
+# and exempts whole documents by accident.
+_MIGRATION_CUE_WORD_RE = re.compile(
+    r"\b(?:" + "|".join(re.escape(w) for w in _MIGRATION_CUE_WORDS) + r")\b"
+)
+
 
 def _migration_cues(old_term: str, new_term: str) -> tuple[str, ...]:
     """Directed migration cues for one ``(old_term, new_term)`` pair, normalized.
@@ -569,21 +578,30 @@ def _has_migration_context(normalized: str, old_term: str, new_term: str) -> boo
 
     1. A *directed* cue naming both (or the specific) terms — ``renamed to X``,
        ``formerly Y``, ``Y = X``, ``Y -> X``, ``Y → X``, ``alias Y`` …
-    2. A generic migration cue word (``replacing``, ``earlier``, ``old`` …)
-       **plus** the new term appearing somewhere in the document. Tier 2 is what
-       catches real corpus prose whose cue and term sit in different clauses, e.g.
+    2. A generic migration cue word (``replacing``, ``earlier``, ``old`` …),
+       matched **word-anchored** via ``_MIGRATION_CUE_WORD_RE``, **plus** the new
+       term appearing somewhere in the document. Tier 2 is what catches real
+       corpus prose whose cue and term sit in different clauses, e.g.
        *"`AgentSession` lands, replacing both the earlier `SessionLog` and
        `RedisJob` models"* — where no directed cue names ``RedisJob`` at all.
 
-    Requiring the new term for tier 2 is the guard against over-exemption: prose
-    that merely mentions a stale name without ever naming its successor is not a
-    migration record and still queues a fix.
+    Two things make tier 2 a guard rather than a rubber stamp, and both are
+    load-bearing:
+
+    - **The new term must appear.** Prose that merely mentions a stale name
+      without ever naming its successor is not a migration record.
+    - **The cue words are word-anchored.** A bare ``in`` substring test fires
+      inside unrelated words — ``old`` sits inside *threshold*, *holds*,
+      *placeholder*, *bold*, *household* — and with that, tier 2 degenerates into
+      "the document mentions the new term somewhere". Measured on the live corpus:
+      ``docs/guides/summarizer-output-audit.md`` was exempted for ``RedisJob``
+      solely because it contains ``summarize_threshold``.
     """
     if any(cue in normalized for cue in _migration_cues(old_term, new_term)):
         return True
     new = _normalize_prose(new_term)
     if new and new in normalized:
-        return any(word in normalized for word in _MIGRATION_CUE_WORDS)
+        return bool(_MIGRATION_CUE_WORD_RE.search(normalized))
     return False
 
 
@@ -741,6 +759,9 @@ def _absent_new_path_refs(
 
 # A file-path-shaped token. Generalizes the single-segment shape to any number of
 # directory segments so a stale term matching the *first* segment is suppressed too.
+# Byte-identical to ``_PATH_REF_RE`` today, and deliberately kept separate: it answers
+# a different question (apply-time "is this match inside a path token?" suppression vs.
+# that one's write-path existence oracle) and either may diverge without the other.
 _PATH_TOKEN_RE = re.compile(r"(?:[\w.-]+/)*[\w.-]+\.(?:py|md)")
 
 
--- a/reflections/docs_auditor.py
+++ b/reflections/docs_auditor.py
@@ -452,6 +452,12 @@ def _detect_renamed_symbol_fixes(content: str, repo_root: Path) -> list[tuple[st
     longer exists but git history shows a rename, queue a replacement.
     """
     fixes: list[tuple[str, str]] = []
+    # Deliberately narrower than `_PATH_REF_RE`, which was widened to `*` for #2759.
+    # This is a *detector* pattern governing what the auditor proposes, not the write
+    # path the existence invariant guards. Widening it to bare names enlarges the
+    # candidate set and spends the GIT_LOG_FOLLOW_CAP per-run `git log --follow`
+    # budget on names that resolve to no single path anyway — a scope and volume
+    # change with its own failure modes, not a safety fix. Ruled unchanged in #2759.
     for m in re.finditer(r"`((?:[\w.-]+/)+[\w.-]+\.py)`", content):
         path = m.group(1)
         if (repo_root / path).exists():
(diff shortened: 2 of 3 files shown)


valorengels:
## Review: CHANGES REQUESTED

Single-judge review (this pass was dispatched as a directed review with two named rulings to settle; no multi-judge fan-out). Preflight: `state=OPEN mergeable=MERGEABLE mergeStateStatus=CLEAN`. No prior review on this PR.

The architecture here is right and the paper trail is unusually good. Apply-time derivation of gates 2 and 3 from `match.string` / `match.start()` is the correct call and the line-shift fixture proves it empirically. The `subn` overcount correction, the all-suppressed guard, `_reject` still receiving a `str`, the file fence, and #2739's region are all verified as claimed. One blocker: the tier-2 cue-word test is an unanchored substring match, and it is firing accidentally on the live corpus today.

---

### The two rulings the review was asked to settle

**(1) `test_regex_channel_is_also_guarded` fixture change — LEGITIMATE, gate 4 still exercised. Confirmed independently, not taken on trust.**

The builder's stated reason is correct. On `main` the fixture rewrote `\breal\b` -> `ghost` over the shared `doc` fixture (`tests/unit/test_docs_auditor_substrate.py:815`, `"See \`agent/real.py\` and \`agent/other.py\`.\n"`). That match sits wholly inside the path token `agent/real.py`, so under the new gate 3 `_match_inside_path_token` returns True, the replacer returns `match.group(0)`, `count -= len(suppressed)` drives count to 0, and the loop `continue`s **before** `_absent_new_path_refs` is ever called. `withheld` would be `[]` and the test would fail on `len(withheld) == 1` — for the wrong reason. The edit was necessary.

Traced the new fixture line by line to confirm it has not been downgraded into a gate-3 test or a pass-for-an-unrelated-reason:

- Doc text `"The runtime lives in the real handler.\n"` -> `_PATH_REF_RE.findall` yields nothing, so `original_refs == set()`.
- `_is_documented_deletion(0, ...)`: not in a fence, heading is `""`, and the line carries none of `_DELETION_PROSE_CUES` (`deleted module`, `no longer in the codebase`, `no longer exists`, `previously in`, `formerly`). Not suppressed.
- `_match_inside_path_token`: the line contains no `.py`/`.md` token at all (`handler.` is followed by a newline, which the line slice excludes). Not suppressed.
- So the substitution proceeds, `candidate` gains `agent/ghost.py`, and the withhold is produced **only** by `_absent_new_path_refs` -> `_reject(..., reason="target-absent")`.

`reason == "target-absent"` is producible by exactly one code path in the module. The assertion set (`applied == 0`, `len(withheld) == 1`, `reason == "target-absent"`) therefore still pins GATE 4 and nothing weaker. Ruling: **legitimate and necessary, no silent weakening.**

**(2) `TestWithheldRateNonRegression` hatch bypass — SOUND. Assessed on merits, not against the pre-reconciliation plan text.**

The third vacuity path is real and the builder is right to have caught it: after #2744 the hatch exempts every live doc that mentions a `STALE_TERMS` key (verified — `_detect_stale_term_fixes` proposes **0** fixes across all of `docs/**/*.md` outside `docs/plans/`), so a hatch-filtered measurement is `0 == 0` with `_absent_new_path_refs` never invoked. The #2744 fix would have silently hollowed out #2759's proof.

Building the fix set straight from `STALE_TERMS` is the right bypass: AC4's property (*widening `_PATH_REF_RE` costs no new withholds*) is a property of the existence invariant, not of the hatch, so filtering through the hatch is not just vacuous, it is measuring the wrong thing. The bypass yields a strictly larger candidate set, so the measurement is conservative in the right direction. The three guard assertions do their job — `visible_refs` strictly greater (proves the narrow arm is really applied), `invariant_calls > 0` on both arms (proves the guard ran), then `after <= before`. Self-baselining with no pinned constant is correct for a corpus the auditor rewrites daily.

Mechanics check out too: disposable `git worktree add --detach` into a `mkdtemp` holder, both teardowns in `finally`, `git checkout -- .` between arms so both measure one snapshot, and `_BASENAME_INDEX_CACHE.clear()` on both sides of each arm (the cache is keyed on `repo_root.resolve()` and would otherwise be shared across arms). Module-global monkeypatching of `_PATH_REF_RE` and `_absent_new_path_refs` resolves correctly because both are looked up as module globals at call time. Never touches the live checkout. Ruling: **sound deviation.**

---

### Other items explicitly verified

- **File fence** — `git diff --name-only main...HEAD` is exactly `reflections/docs_auditor.py`, `tests/unit/test_docs_auditor_substrate.py`, `docs/features/docs-auditor.md`. Anti-criterion row output `0`.
- **#2739's region untouched** — `git diff main...HEAD -- reflections/docs_auditor.py | grep -c '_pr_is_auto_merge_eligible\|_push_branch_and_pr\|_commit_current_branch\|run_docs_branch_sweeper'` -> `0`.
- **`test_fixes_travel_on_the_regex_channel` unmodified** — appears as diff context only; confirmed.
- **`subn` overcount correction** — `count -= len(suppressed)` at `:888`, tallied via the closure-cell list. Correct.
- **All-suppressed guard** — `if count <= 0 or candidate == new_text: continue` at `:892`. Correct, and the `candidate == new_text` half also covers the case where suppression leaves the text byte-identical.
- **`_reject` receives a string** — `_reject(pattern.pattern, new, absent)` at `:896`; `new` is the plain replacement `str`. The callable is built locally at `:887` and never escapes. Withheld records stay human-readable.
- **Suppression derived at apply time** — `_make_stale_term_replacer` reads `match.string` and `match.start()` only; there is no detection-time line index anywhere in the path. `_build_line_context` is recomputed against the live text and memoized per text value.
- **No caller regression** — `_absent_new_path_refs` gained a required `doc_path` param; it is module-private with exactly two call sites, both updated (`:878`, `:894`). No external importer.
- **Verification table** — ran every row that is not a pytest invocation (TEST already covered those): bare-names-visible `1`, popoto-doc `0`, migration-audit-doc `0`, detector-still-detects `1`, regex-channel-purity `0`, no-sweeper-region `0`, file-fence `0`, no-swallowed-exceptions `0`, sibling-reason-recorded `8`, no-xfails exit 1, ruff check + `ruff format --check` clean. All pass.
- **Success criteria measured** — `docs/guides/summarizer-output-audit.md` currently produces `0` proposals, better than the criterion's "at most one". See Blocker 1 for why that zero is not the win it looks like.

---

### Blockers

**File:** `reflections/docs_auditor.py:586` (verified: read this file)
**Code:** `return any(word in normalized for word in _MIGRATION_CUE_WORDS)`
**Issue:** The tier-2 cue-word test is an **unanchored substring** match, so `_MIGRATION_CUE_WORDS` entries fire inside unrelated words. `"old"` is a substring of `threshold`, `holds`, `holder`, `placeholder`, `bold`, `gold`, `household` — words that are ubiquitous in this corpus. Combined with the only other tier-2 condition (the new term appears somewhere in the document), tier 2 degenerates for a large class of docs into "the document mentions `AgentSession`", which is exactly the over-exemption that Plan Risk 1 says the new-term requirement guards against, and exactly what `docs/features/docs-auditor.md` describes as "the anti-over-exemption guard".

This is not hypothetical. It fires on the live corpus today, and I reproduced it:

```
docs/guides/summarizer-output-audit.md, term RedisJob:
  directed cue present?          no
  standalone \bold\b present?    no
  words containing "old":        {"threshold", "summarize_threshold"}
  -> exempted by tier 2 solely because "threshold" contains "old"
```

The PR body's own test corpus shows how thin the margin is: `test_no_migration_context_queues_fix` includes *"The SessionLog **holds** per-turn state and is queried directly."* — that case passes only because the isolated string does not contain `AgentSession`. Put that same sentence in any document that names `AgentSession`, and `holds` exempts every stale term in it.

The consequence is not a corrupted write, so the direction of error is safe. It is that a documented guard does not guard, and that the PR's headline measurement is misattributed: the plan explicitly predicted and **accepted** one residual at `docs/guides/summarizer-output-audit.md:68` (a quoted transcript line), and the PR body states that residual "does not fire, because the document-scoped hatch exempts that doc via the second cue tier". It does not. It is exempted by a substring collision on `threshold`.

**Severity:** blocker
**Fix:** Anchor the cue-word test: `return any(re.search(rf"\b{re.escape(word)}\b", normalized) for word in _MIGRATION_CUE_WORDS)` (or precompile the alternation once at module level). I measured the effect before proposing it: with anchoring, the only doc whose exemption changes is `summarizer-output-audit.md`, and exactly **one** `RedisJob` occurrence survives gates 2 and 3 — line 68, the quoted-transcript residual the plan already accepted:

```
RedisJob occurrences surviving gates 2+3: 1
   (68, '**Example 3** (10:52, fixing RedisJob refs):')
```

So the fix restores the designed mechanism and lands precisely on the plan's success criterion ("at most **one** changed occurrence — the accepted `:68` residual, down from five"). All 8 parametrized `test_migration_context_skips_fix` cases keep passing (`old imports`, `replacing`, `earlier` are all standalone there). Add a regression test pinning that a cue word embedded in a larger word does not exempt — e.g. a doc containing `AgentSession`, a stale term, and `threshold`, asserting a fix is still queued. Then correct the PR body's "Nothing to file" claim and the "18 -> 0" table, and re-check the `docs/features/docs-auditor.md` tier-2 paragraph.

### Tech Debt

**File:** `reflections/docs_auditor.py:288` (`_repo_basename_index`) (verified: read this file)
**Code:** `logger.warning("docs_auditor: git ls-files failed in %s (rc=%s): %s — bare-name ...")`
**Issue:** The plan's `## Failure Path Test Strategy` names the `git ls-files` call as "the only new failure surface" and requires a test that "asserts the `logger.warning` fires and that dir-prefixed refs are unaffected"; Risk 3's mitigation likewise says "Tested with a forced failure." The production code is correctly wrapped and degrades as designed, but no such test exists — `grep -n "ls-files\|_repo_basename_index\|normalize_prose" tests/unit/test_docs_auditor_substrate.py` returns only docstring prose. The degraded path (empty index -> bare names resolve doc-relative only -> over-withhold) is therefore unpinned, and a future refactor that swallows the failure or flips the fallback to permissive would ship green. The same section's `_normalize_prose("")` / whitespace-only cases are also untested (`test_absent_key_emits_no_fix` covers only `"Nothing stale here at all."`, not empty or whitespace-only input as the plan asked).
**Severity:** tech_debt
**Fix:** Add a test patching `docs_auditor.subprocess.run` to return `returncode=1` (and a second raising `OSError`), asserting the warning record is emitted, that `_repo_basename_index` returns `{}`, and that a `dir/`-prefixed ref's withhold decision is unchanged. Extend `test_absent_key_emits_no_fix` (or add a parametrize) over `""` and `"   \n\n  "`.

**File:** PR description, "Measured outcome" and "Rulings recorded" sections (verified: read via `gh pr view 2782 --json body`)
**Code:** `The plan predicted one accepted residual (the quoted-transcript line at summarizer-output-audit.md:68); it does not fire, because the document-scoped hatch exempts that doc via the second cue tier. Nothing to file.`
**Issue:** PR body accuracy. As shown in Blocker 1, that doc is exempted by an unanchored substring match on `threshold`, not by the second cue tier operating as designed. The `7 -> 0` row for `summarizer-output-audit.md` and the `total 18 -> 0` figure are consequences of the same accident. Every other claim in the body that I checked reproduced exactly, which is why this one stands out.
**Severity:** tech_debt
**Fix:** After Blocker 1's fix, re-measure and restate the table honestly (the residual returns to 1, which is what the plan accepted), and drop or amend "Nothing to file".

### Nits

**File:** `reflections/docs_auditor.py:639` and `reflections/docs_auditor.py:744` (verified: read this file)
**Code:** `_PATH_REF_RE = re.compile(r"(?:[\w.-]+/)*[\w.-]+\.(?:py|md)")` / `_PATH_TOKEN_RE = re.compile(r"(?:[\w.-]+/)*[\w.-]+\.(?:py|md)")`
**Issue:** After the `+` -> `*` widening these two module constants are byte-identical patterns under two names. Their *purposes* are genuinely distinct (write-path existence oracle vs. apply-time suppression scan) and the separate names document that, so this is not a merge-them-blindly situation — but as written, a future change to one silently fails to track the other, and a reader cannot tell whether the identity is intentional. `TestWithheldRateNonRegression` monkeypatches only `_PATH_REF_RE`, which is correct today precisely because gate 3 is not what AC4 measures; that subtlety is invisible at the definition sites.
**Severity:** nit
**Fix:** Add a one-line comment at `_PATH_TOKEN_RE` stating that it is deliberately identical to `_PATH_REF_RE` today but is allowed to diverge because it answers a different question, or define one from the other's `.pattern` with that note.

### Miscellaneous
- None

- **`[SEPARATE-SLUG #2739]` review gate in front of every auditor write** — tracked by #2739 (OPEN), region verified untouched by the anti-criterion grep.
- **`[SEPARATE-SLUG #2729]` escalation for permanently auto-merge-ineligible withheld PRs** — tracked by #2729 (OPEN); the "does not block this lane" ruling is mechanically backed by `TestWithheldRateNonRegression`, which I assessed above and accept.
- **`[SEPARATE-SLUG #2725]` rename detectors picking the wrong target** — tracked by #2725 (OPEN), adjacent code, different defect.
- **Sibling patterns `_detect_renamed_symbol_fixes` / `_detect_deleted_target_issues` left narrower** — explicit in-scope ruling with the reason recorded in code comments at both sites naming #2759, and repeated in the feature doc. Correct: both are detector-side, not write-path.
- **The `summarizer-output-audit.md:68` quoted-transcript residual** — an explicit accepted exclusion in the plan with sound rationale ("recognizing quoted history is an open-ended problem"). Note that Blocker 1's fix returns this residual to its accepted state rather than removing it; it should be re-filed as the follow-up comment on #2744 the plan describes.

---

## Rubric

- [x] **1. Plan vs. implementation match** — pass — *All four gates, the widened `_PATH_REF_RE`, `doc_path` threading, the basename index and its per-run cache clear, the ambiguity ruling, and the sibling-pattern comments are all present and match the plan. The one intentional deviation (`TestWithheldRateNonRegression`'s hatch bypass) is sound and already reconciled on main in 79793dd77.*
- [ ] **2. New code quality** — fail — *`_has_migration_context`'s tier-2 cue test matches cue words as unanchored substrings; see Blocker 1. Everything else in the new code is clean: type hints throughout, the closure-cell suppression tally is honest, the degraded path for `git ls-files` is logged rather than swallowed.*
- [x] **3. Test coverage** — pass — *129 targeted tests, every behavioral fix demonstrated red against main first, red output pasted verbatim. The gap in failure-path coverage is recorded as tech debt, not a coverage `fail`.*
- [x] **4. Regression risk to existing callers** — pass — *`_absent_new_path_refs` is module-private with exactly two call sites, both updated. `_detect_stale_term_fixes`'s `(re.Pattern, str)` channel contract is unchanged and `test_fixes_travel_on_the_regex_channel` passes unmodified.*
- [x] **5. Data integrity** — n/a — *No models, migrations, or persisted schema touched.*
- [x] **6. Security** — pass — *One new `subprocess.run` with a fixed argv, no shell, no request-derived input, `timeout=settings.timeouts.git_subprocess_s`. No secrets, no eval/exec.*
- [ ] **7. Documentation accuracy** — fail — *`docs/features/docs-auditor.md` is otherwise excellent and unusually complete, but its tier-2 paragraph describes the new-term requirement as the anti-over-exemption guard, which Blocker 1 shows is not what the code delivers.*
- [ ] **8. PR body accuracy** — fail — *The "Nothing to file" residual claim and the `summarizer-output-audit.md` 7 -> 0 row are misattributed; see Tech Debt 2. Every other claim I checked reproduced exactly.*
- [x] **9. Disclosed deferrals** — pass — *All four `[SEPARATE-SLUG]` No-Gos resolve to OPEN issues (#2739, #2729, #2725) or are explicit exclusions with sound rationale; the two code-level No-Gos have anti-criterion Verification rows and both pass.*
- [x] **10. Follow-up claims verified** — pass — *#2739, #2729, #2725 all verified OPEN.*

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — PASS — *All Success Criteria met or exceeded; the summarizer criterion is met by a mechanism Blocker 1 disputes, which is why it is a blocker and not merely a note.*
- **2. No-Gos from plan — none violated** — PASS — *Both code-level No-Gos have anti-criterion rows; both output 0.*
- **3. New `except Exception` blocks — each has logger/raise/swallow-ok** — PASS — *No new bare `except Exception` added; the new `except (OSError, subprocess.SubprocessError)` logs a warning. Anti-criterion grep for `except Exception: pass` returns 0.*
- **4. New integration tests — exercise serialization boundary (not in-memory only)** — PASS — *`TestWithheldRateNonRegression` drives real file writes in a real git worktree; `test_bare_name_withhold_propagates_...` runs a real `audit()` and asserts on PR body, Telegram, and liveness.*
- **5. Plan internal consistency — spike findings match task steps** — PASS — *Spike-2's line-scope rejection, spike-4's zero-new-withholds, and spike-5's ambiguity ruling are each reflected in code, comments, and the feature doc.*
- **7. New public APIs — docstrings present** — PASS — *Every new module-private helper carries a substantive docstring; no new public surface.*
- **8. Breaking changes — migration path documented** — N/A — *Both public entry points keep their signatures and `_ok_result` contract.*
- **9. Tests added for new behavior** — PASS — *13 new/updated behavioral tests plus the non-regression harness.*
- **10. Tests cover the failure path (not just happy path)** — FAIL — *The `git ls-files` failure surface the plan singled out has no test; promoted to Tech Debt 1.*
- **11. UI changes (if any) — screenshot captured** — N/A — *No HTML/CSS/JS/TSX/Vue/template files in the diff; the visual proof gate is a no-op.*
- **12. Docs updated for user-facing changes** — FAIL — *Docs are updated and thorough, but the tier-2 description is inaccurate; folded into Blocker 1's fix.*

**Verdict derivation:** rubric item 2 is a critical `fail` with no matching acknowledged disclosure -> Rule 1 -> **CHANGES REQUESTED — Blocker**.

**Restart note:** this touches `reflections/` only, not `bridge/`, `agent/`, or `worker/`. No service restart needed beyond the normal `/update` code sync after merge.

valorengels:
## Review: CHANGES REQUESTED

Round 2, single-judge directed re-review (no multi-judge fan-out; this pass was scoped to round 1's three rubric fails). Preflight: `state=OPEN mergeable=MERGEABLE mergeStateStatus=CLEAN`, head `7a9b1e89e`.

Every number below was re-measured in this environment against `7a9b1e89e` rather than read from the PR body.

### What I reproduced myself

- `POPOTO_TEST_DB=9 ./scripts/pytest-clean.sh tests/unit/test_docs_auditor_substrate.py -q -p no:randomly` -> **136 passed** (92s).
- `ruff check .` -> `All checks passed!`; `ruff format --check .` -> `1336 files already formatted`.
- All **14** rows of the plan's `## Verification` table PASS, including the four anti-criteria (regex-channel purity, no sweeper-region edits, file fence, no swallowed exceptions).
- File fence: `git diff main...HEAD --name-only` is exactly `docs/features/docs-auditor.md`, `reflections/docs_auditor.py`, `tests/unit/test_docs_auditor_substrate.py`.
- #2739's region (`_pr_is_auto_merge_eligible`, `_push_branch_and_pr`, `_commit_current_branch`, `run_docs_branch_sweeper`) is untouched by the diff.
- `test_fixes_travel_on_the_regex_channel` appears in the diff only as context — unmodified, and green. `_detect_stale_term_fixes` still returns `list[tuple[re.Pattern[str], str]]`.

### The blocker fix is real, and correctly scoped

I drove `_apply_fixes_to_file` over the three named docs inside a disposable `git worktree add --detach` (never the live checkout) and diffed the result:

- `docs/features/popoto-redis-expansion.md` — `applied=0`, `withheld=0`, 0 changed lines.
- `docs/guides/agent-session-migration-audit.md` — `applied=0`, `withheld=0`, 0 changed lines.
- `docs/guides/summarizer-output-audit.md` — `applied=1`, `withheld=0`, and the single changed line is exactly the accepted `:68` residual: `**Example 3** (10:52, fixing AgentSession refs):`.

Total **18 -> 1**, matching the corrected PR-body table and the plan's success criterion ("at most one changed occurrence").

The anchoring itself checks out. On the normalized text of `summarizer-output-audit.md`, `_MIGRATION_CUE_WORD_RE.search(...)` is `False` while the old `any(word in normalized ...)` is `True` — the accidental exemption came from `old` inside `summarize_threshold`, and `_normalize_prose` does not split underscores, so `\bold\b` correctly declines. `_has_migration_context(norm, "RedisJob", "AgentSession")` is now `False` for that doc.

**No over-restriction.** The two docs the hatch is supposed to exempt are still exempt after anchoring (0 detected fixes each), so anchoring removed the accidental exemption without removing any legitimate one. Corpus-wide over `docs/**/*.md` outside `docs/plans/`, exactly **one** stale-term proposal survives — the `:68` residual — and it is filed as a follow-up comment on #2744 (OPEN). Alternation-ordering is a non-issue: the trailing `\b` makes `rename` fail inside `renamed` regardless of branch order.

### The `git ls-files` failure-path tests are not vacuous

`_repo_basename_index` memoizes on `repo_root.resolve()` into `_BASENAME_INDEX_CACHE`, and the failure path *does* cache its empty result (`_BASENAME_INDEX_CACHE[key] = index` runs after the `except` too). Without the new `clear_basename_cache` fixture, a prior test's populated index for the same `tmp_path` would satisfy the call and both assertions would pass against a cache hit. The fixture clears before *and* after, so the assertion lands on a live subprocess call. Both parametrizations reach the intended arm: the `returncode=1` mock takes the `if proc.returncode != 0` branch, and the generator-`throw` `OSError` is caught by `except (OSError, subprocess.SubprocessError)`. Caching the degraded result within a run is correct, not a bug — `audit()` clears the cache at the top of every run (`reflections/docs_auditor.py:1363`), so the snapshot can never outlive a run.

The companion test is also non-vacuous in the right direction: it asserts both arms of the dir-prefixed class (present target applies, absent target withholds) under a forced index failure, which pins that the bare-name fallback did not leak into the path class that already worked.

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — PASS — *all 14 verification rows re-run and green; 18 -> 1 reproduced.*
- **2. No-Gos from plan — none violated** — PASS — *file fence honored; #2739 region untouched; `:68` residual filed on #2744 rather than chased.*
- **3. New `except Exception` blocks — each has logger/raise/swallow-ok** — PASS — *no new broad handler; `_repo_basename_index` catches a named tuple and logs a warning.*
- **4. New integration tests — exercise serialization boundary (not in-memory only)** — PASS — *`_apply_fixes_to_file` is driven against real files on disk in a real worktree.*
- **5. Plan internal consistency — spike findings match task steps** — PASS — *document-scope and apply-time rulings still match spike-2/spike-4.*
- **6. No hardcoded secrets or debug artifacts** — PASS — *none.*
- **7. New public APIs — docstrings present** — PASS — *`_MIGRATION_CUE_WORD_RE` carries the rationale comment; `_has_migration_context` docstring rewritten.*
- **8. Breaking changes — migration path documented** — N/A — *no external contract changed.*
- **9. Tests added for new behavior** — PASS — *129 -> 136; the anchoring regression is pinned by `test_cue_word_inside_a_larger_word_does_not_exempt`.*
- **10. Tests cover the failure path (not just happy path)** — PASS — *the plan's named failure surface is now covered in both arms.*
- **12. Docs updated for user-facing changes** — FAIL — *see Tech Debt 1: `docs/features/docs-auditor.md` still asserts the pre-fix "zero" figures.*

## Rubric

- [x] **1. Plan vs. implementation match** — pass — *every success criterion I could execute, I executed.*
- [x] **2. New code quality** — pass — *round 1's fail is fixed: the cue-word test is a precompiled, `re.escape`'d, word-anchored module-level alternation, with a comment naming why anchoring is load-bearing.*
- [x] **3. Test coverage** — pass — *136 green; the new behavior and the new failure surface are both pinned, and the cache fixture makes the failure tests real.*
- [x] **4. Regression risk to existing callers** — pass — *`_has_migration_context` has one caller; anchoring strictly narrows tier 2 and the two legitimately-exempt docs stay exempt.*
- [x] **5. Data integrity** — n/a — *no schema or model change.*
- [x] **6. Security** — pass — *no new eval/exec/interpolation; the one subprocess is a fixed argv list.*
- [ ] **7. Documentation accuracy** — fail — *`docs/features/docs-auditor.md:85-87` still states "18 occurrences to zero" and "surviving stale-term proposals are zero". Measured: 18 -> 1, and exactly one surviving proposal.*
- [ ] **8. PR body accuracy** — fail — *the measured-outcome table and the "quiet, not dead" paragraph were correctly re-measured, but the `## Testing` and `## Definition of Done` sections still claim **129 passed**. Measured: 136.*
- [x] **9. Disclosed deferrals** — pass — *the `:68` residual is disclosed and tracked.*
- [x] **10. Follow-up claims verified** — pass — *the follow-up resolves to comment `5279038362` on #2744, which is OPEN.*

### Blockers
- None

Round 1's blocker (unanchored cue-word substring match) is resolved and independently verified.

### Tech Debt

- **File:** `docs/features/docs-auditor.md:85-87` (verified: read this file)
  **Code:** `Over the three docs the auditor previously corrupted, false rewrites went from **18 occurrences to zero**. Across all of `docs/**/*.md` outside `docs/plans/`, surviving stale-term proposals are zero;`
  **Issue:** Both figures are the pre-anchoring numbers and are now false. Measured on `7a9b1e89e`: false rewrites went from 18 to **1**, and exactly **one** stale-term proposal survives corpus-wide (`docs/guides/summarizer-output-audit.md:68`). Round 1's tech-debt item was fixed in the PR body only; the same wrong attribution survives in the durable artifact. That matters more here than in most PRs: this is the feature doc for the subsystem whose entire purpose is stopping an autonomous writer from committing false statements into `docs/`, and it would land one. The adjacent "four docs that still contain a `STALE_TERMS` key" claim I re-counted and it is correct — only the two "zero" figures are wrong.
  **Severity:** tech_debt
  **Fix:** Change "18 occurrences to zero" to "18 occurrences to one" and "surviving stale-term proposals are zero" to name the single surviving proposal (the accepted `summarizer-output-audit.md:68` quoted-transcript residual, filed on #2744). Optionally carry over the one-sentence correction note already in the PR body so the number's history is legible.

- **File:** PR #2782 body, `## Testing` and `## Definition of Done` (verified: read via `gh pr view 2782 --json body`)
  **Code:** `- [x] `POPOTO_TEST_DB=9 ./scripts/pytest-clean.sh tests/unit/test_docs_auditor_substrate.py -q` -> **129 passed**` and `- [x] Tested: 129 passing, every fix demonstrated red first`
  **Issue:** `7a9b1e89e` added seven tests (1 anchoring regression + 2 new `test_absent_key_emits_no_fix` parametrizations + 2x2 `ls-files` failure params). I ran the exact quoted command and got **136 passed**, not 129. The stated output of a stated command does not reproduce. This is low-stakes on its own, but `/do-merge` squashes this body into `main`'s commit message, so the stale figure becomes the permanent record — and it is the same class of defect round 1 already flagged.
  **Severity:** tech_debt
  **Fix:** Update both occurrences of 129 to 136, and note in `## Testing` that the count moved because the round-2 commit added the anchoring regression test and the `git ls-files` failure-path coverage.

### Nits
- None

Round 1's nit (missing rationale comment at `_PATH_TOKEN_RE`) is resolved — `reflections/docs_auditor.py:762-764` now records that it is byte-identical to `_PATH_REF_RE` today, answers a different question, and may diverge independently.

### Miscellaneous
- None

- **The accepted `summarizer-output-audit.md:68` quoted-transcript residual** — tracked by comment `5279038362` on #2744 (OPEN) — the plan's No-Gos ask for exactly this disposition, and the measured residual count of 1 matches the success criterion.
- **`_detect_renamed_symbol_fixes` / `_detect_deleted_target_issues` sibling patterns left narrower** — explicit out-of-scope exclusion with the reason recorded in code comments naming #2759; a verification row asserts the reason is present.
- **This lane did not wait on #2729** — explicit exclusion, justified by the measured zero new-withhold count, which `TestWithheldRateNonRegression` pins self-baselining (narrow 1025/0 vs widened 1569/0).

### Review Delta (vs prior review on HEAD e090a49)

- **Resolved**: *Blocker — unanchored cue-word substring match at `_has_migration_context`.* `_MIGRATION_CUE_WORD_RE` fixes it; verified independently, 18 -> 1 corpus-wide with no over-restriction.
- **Resolved**: *Tech debt — no coverage for the `git ls-files` failure surface.* Both arms covered, and the `clear_basename_cache` fixture makes the assertions hit a live call rather than a memoized one.
- **Resolved**: *Nit — no comment explaining why `_PATH_TOKEN_RE` stays separate from `_PATH_REF_RE`.*
- **Partially resolved / New**: *Tech debt — PR body accuracy.* The specific misattribution round 1 named (the 7->0 row, the 18->0 total, the "nothing to file" claim) is corrected, and the correction is stated rather than silently rewritten — that part is done well. But the same pre-fix figures survive in `docs/features/docs-auditor.md:85-87`, and the `## Testing` counts drifted to stale in the same commit that fixed the table. Rubric items 7 and 8 therefore remain fail.
- **Unchanged**: Round 1's two directed rulings (the `test_regex_channel_is_also_guarded` fixture repair and the `TestWithheldRateNonRegression` hatch bypass) stay settled. Not reopened.

### Verdict derivation

No critical-item fail. Rubric items 7 and 8 (non-critical) fail -> Rule 2 -> **CHANGES REQUESTED — Tech Debt**. Both findings are text-only edits with no code impact; `/do-patch` should clear them in a single pass, after which this is an approve.

Issue #2781, opened 2026-08-13
Reliability risk: SDLC session ensure — six lease heartbeats spawned on March-closed issues under an off-pin interpreter, with no spawn-side terminal-issue guard


valorengels:
## Investigation Finding

**Component:** SDLC session ensure (`tools/sdlc_session_ensure.py::_maybe_launch_lease_heartbeat`)
**Detected:** 2026-08-13 (observed on the main checkout while scoping the #2714 batch; recorded in issue #2714 comments [5277602180](https://github.com/tomcounsell/ai/issues/2714#issuecomment-5277602180) and [5278530522](https://github.com/tomcounsell/ai/issues/2714#issuecomment-5278530522))
**Status:** Observed-once

Split out of #2714 and its plan `docs/plans/sdlc-lease-heartbeat-supervisor-lifetime.md`.

### Why this is a separate issue from #2714

#2714's root cause is **consumer-side**: a heartbeat that was legitimately spawned keeps renewing after the supervisor that needed it is gone, because nothing connects the heartbeat's lifetime to its supervisor's. Its fix anchors the heartbeat to supervisor liveness.

This issue's root cause is **producer-side**: something spawned six heartbeats *at all*, on issues that closed five months earlier, under an interpreter that is not the repo's pin. Even a perfectly supervisor-anchored heartbeat should never have been launched on issue #515 (merged 2026-03-24). Fixing #2714 does not answer *why the spawn happened*, and answering *why the spawn happened* does not fix #2714. Two different defects that happened to surface in the same evidence.

**This split is NOT justified on the grounds that #2714's fix bounds these heartbeats.** Per that plan's No-Gos section, it explicitly does not:

> Round 1 asserted "<=2 hours regardless of how it was spawned". That is false and the round-1 plan disproved it in its own Risk 4. **<=2h holds only on the unresolvable-supervisor path.** A re-spawned heartbeat launched from inside a live `claude` session resolves `CLAUDE_PID`, so L2 reports its supervisor alive and it is bound by the unchanged 4h ceiling.

So a recurrence spawned from inside a live session would still get the full 4h `MAX_LIFETIME_SECONDS` after #2714 ships. The spawn-side gap needs its own fix.

## Symptoms

[...]

PR #2784, opened 2026-08-13
Anchor the SDLC lease heartbeat to its supervisor's lifetime

### LINKED ISSUE #2714: SDLC lease heartbeat outlives a stood-down supervisor by up to 4 hours, so every resume on that issue sees a live foreign holder
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/sdlc-lease-heartbeat-supervisor-lifetime.md

> **Context:** An SDLC run holds a per-issue lease so two supervisors never work the same issue at once. The lease is kept alive by [`tools/sdlc_lease_heartbeat.py`](tools/sdlc_lease_heartbeat.py), launched as a **detached** subprocess by [`tools/sdlc_session_ensure.py::_maybe_launch_lease_heartbeat`](tools/sdlc_session_ensure.py#L143) (`start_new_session=True`). Detachment is deliberate: the heartbeat must outlive the short-lived `sdlc-tool session-ensure` CLI that mints the lock, because that CLI's pid is dead by the first tick and any pid-keyed liveness check would read every locally-minted lease as orphaned and lapse it mid-stage (the #2446/#2451 failure this heartbeat exists to prevent).

## Problem

The heartbeat is detached from the process that mints the lease, but nothing re-attaches it to the process that *uses* the lease. When a supervisor stands down — halts on a guard, gate-stops, or is killed — its heartbeat keeps renewing the lease. The issue continues to look like it is held by a live run for as long as the heartbeat survives.

The only death signal is a wall-clock backstop: `MAX_LIFETIME_SECONDS`, default `4 * 60 * 60` ([`tools/sdlc_lease_heartbeat.py:79`](tools/sdlc_lease_heartbeat.py#L79)). Every renew also refreshes the companion supervised-run signal, so both keys stay fresh for up to 4 hours after the run they represent has ceased to exist.

**Current behavior:**

Observed on #2701 on 2026-08-10. The supervisor stood down at a Task 1 gate FAIL. Its heartbeat was still running 3h23m later:

```
  PID  ELAPSED COMMAND
76576 03:23:51 python -m tools.sdlc_lease_heartbeat --issue-number 2701 \
               --run-id 2e9f4a3dfdde4097aeee19e8151dc1fc --session-id sdlc-local-2701
```

Consequences during that window:

- `orphaned_lock` reads `false`, because the lease is genuinely being renewed. The signal is accurate about the *lease* and wrong about the *run*.
- A new supervisor resuming that issue sees a live foreign holder and correctly stands down. Recovery required adopting the dead run's identity via `--reuse-run-id`; a stricter reading of the skill's refusal table would have refused to start at all.
- The stale heartbeat had to be killed by hand before the issue could be worked.

**Desired outcome:**

When a supervisor stops for any reason, its lease stops being renewed promptly, so `orphaned_lock` reflects run liveness rather than lease freshness. A supervisor's death should not require either a 4-hour wait or manual process cleanup before its issue can be resumed.

## Definitions

| Issue lease / issue lock | Per-issue Redis lock ensuring one active SDLC run per issue | [`models/session_lifecycle.py::touch_issue_lock`](models/session_lifecycle.py) |
| `orphaned_lock` | Flag telling a resuming run whether the lock's owner died and the lease is safe to inherit | [`tools/sdlc_session_ensure.py:29`](tools/sdlc_session_ensure.py#L29) |
| Supervised-run signal | Companion key proving a run is under live supervision; refreshed on the same tick as the lease | [`agent/supervised_run.py`](agent/supervised_run.py), #2659 |
| Stand down | A supervisor stopping deliberately — guard block, gate FAIL, or completion — as opposed to crashing | [`.claude/skills/sdlc/SKILL.md`](.claude/skills/sdlc/SKILL.md) |

## Prior Context

- #2446 / #2451: lease lapsing mid-stage; the reason pid-keyed liveness was rejected and this heartbeat introduced.
- #2537: peek-first renew-only design, so a heartbeat never re-acquires a lease it lost (lease theft).
- #2659: supervised-run signal renewed on the same tick as the lease, so the signal cannot outlive the lease.

Each of those hardened the heartbeat against dying *too early*. None covers it living too long.

## Solution Sketch

The design comment states the intent plainly: *"Liveness of the RUN is this heartbeat's own job: its bounded max-lifetime is the death backstop."* The gap is that a 4-hour backstop is the *only* death signal, and it is orders of magnitude longer than the event it needs to detect.

Open questions for planning, rather than a prescribed fix:

- Can the heartbeat cheaply observe its supervisor's liveness without reintroducing the pid-keyed inference that #2446/#2451 ruled out? The supervisor is long-lived, unlike the `session-ensure` CLI whose dead pid caused the original problem, so a pid check against the *supervisor* may be sound where one against the *minting CLI* was not.
- Should a supervisor explicitly stop its heartbeat when it stands down, and what covers the crash case where it cannot?
- Is a much shorter max-lifetime with periodic renewal-of-intent by the supervisor a better shape than one long backstop?
- Should `orphaned_lock` distinguish "lease fresh, run unproven" from "lease fresh, run confirmed live" so a resuming supervisor can make a smarter choice than stand-down?

Whatever shape is chosen must not weaken the #2446/#2451 or #2537 guarantees.

## Recon Summary

**Confirmed:** 4 items
- Heartbeat is launched detached with `start_new_session=True`; no code path stops it on supervisor exit (grep for `lease_heartbeat` finds launch sites, no stop sites).
- `MAX_LIFETIME_SECONDS` defaults to 4 hours, overridable via `SDLC_LEASE_HEARTBEAT_MAX_LIFETIME_SECONDS`.
- The renew loop refreshes the supervised-run signal on every tick, so both keys stay fresh together.
- Live reproduction on #2701: 3h23m of post-mortem renewal, resolved by manual `kill`.

**Revised:** 1 item
- Initially characterized as "the heartbeat has no lifecycle owner." Inaccurate: it has a documented, deliberate self-exit backstop. The defect is the backstop's duration being the sole death signal, not its absence.

**Pre-requisites:** 0 items

**Dropped:** 1 item
- Reverting to pid-keyed liveness on the lock payload. Explicitly rejected by #2446/#2451 and re-affirmed in the module docstring.

## Acceptance Criteria

- [ ] A supervisor that stands down (guard block, gate FAIL, or normal completion) stops renewing its lease within one lease TTL, without manual intervention.
- [ ] A supervisor that is killed without cleanup also stops renewing within a bounded window materially shorter than 4 hours.
- [ ] After a supervisor stops, a new run on the same issue can acquire the lease without `--reuse-run-id` adoption of a dead run's identity.
- [ ] The #2446/#2451 guarantee still holds: a lease is never lapsed mid-stage while its run is genuinely alive.
- [ ] The #2537 guarantee still holds: a heartbeat never re-acquires a lease it no longer owns.
- [ ] A test covers the stood-down-supervisor case, asserting the lease stops being renewed.

## Downstream

This issue feeds `/do-plan`, which will produce `docs/plans/{slug}.md`, then `/do-build`.

Constraints for the planner: this is core SDLC concurrency control. A regression here either lapses live leases (#2446) or lets two runs work one issue. Prefer a solution that fails toward "lease released" only when run death is positively established, not merely assumed from an absence of signal.

valorengels:
## Problem

`tools/sdlc_lease_heartbeat.py` renewed the per-issue lease with no connection to the process that *uses* it. Its only death signal was a 4-hour wall-clock backstop measured from its own start. A supervisor that halted on a guard, stopped at a gate FAIL, or was killed left its heartbeat renewing — and because the heartbeat kept stamping `renewed_at`, `orphaned_lock` read `false`, so every resume on that issue correctly stood down against a run that no longer existed.

Confirmed live on 2026-08-13: the 4h backstop processes died on schedule, but **the leases did not release**. Expiry-by-timeout was never a release path.

## Solution

Six layers, each defaulting to today's behavior when its new inputs are absent:

- **Renew-only lease extension** (`touch_issue_lock(..., renew_only=True)`) — skips both minting branches and writes through a Lua compare-and-set, so a renewer can never mint on a free key nor undo a release that lands between the peek and the extend. The outer exception handler is threaded too: a renewer fails **closed**, while the non-`renew_only` default keeps #2446's fail-open behavior.
- **Supervisor identity resolution** at mint time — `CLAUDE_PID` env, then a `psutil` ancestry walk for a `claude` ancestor, then unresolved. Never `os.getppid()`, never `ppid == 1`.
- **Supervisor watch** — on `SDLC_SUPERVISOR_DEATH_CONFIRMATIONS` (default 2) consecutive `(pid, create_time)` mismatches the heartbeat **actively releases** the lease and the supervised-run signal, then exits. This is the fix for the finding above: release, not merely stop renewing.
- **Unsupervised ceiling** — 90 minutes when no supervisor resolves, and deliberately **no** release there (failure to resolve is not proof of death, so the lease's own TTL is the right disposition). With a supervisor present the unchanged 4h bound is selected.
- **`sdlc-tool session-release`** plus a tool-layer leg in `sdlc_stage_marker.py` firing on the non-idempotent `MERGE`/`completed` transition, so the merged exit releases without depending on skill prose.
- **Observability** — `logs/sdlc_lease_heartbeat.log` has been 0 bytes since 2026-08-04; INFO logging is now unconditional, with a named startup line and a named exit reason.

## Verification

All Verification and anti-criteria rows pass. Targeted suites, `POPOTO_TEST_DB=11`:

- `test_sdlc_lease_heartbeat.py`, `test_sdlc_session_release.py`, `test_sdlc_session_ensure.py`, `test_sdlc_stage_marker.py` — 245 passed
- `test_session_lifecycle.py`, `test_session_executor_issue_lock_renewal.py` — 103 passed
- `ruff check` / `ruff format --check` clean

Guarantees held: #2446/#2451 (a dead payload pid never stops renewal; a live supervisor renews indefinitely), #2537 (peek-before-renew intact, foreign owner exits, never mints on a free key), and the 90-minute ceiling is unreachable whenever a supervisor pid is present.

The spawn-side observation (six heartbeats on March-closed issues, a different root cause) is split out as #2781.

Closes #2714

### DIFF
--- a/.claude/skills-global/do-sdlc/SKILL.md
+++ b/.claude/skills-global/do-sdlc/SKILL.md
@@ -248,6 +248,30 @@ On exit (any path), report:
 3. **Artifacts**: issue, plan path, PR number, merge commit
 4. **Anything needing human attention**: unresolved blockers, skipped acknowledgments, follow-ups
 
+## Step 5: Release the run lease
+
+You hold this issue's run lease for as long as this supervision loop lives, and
+nothing reclaims it when you simply stop — there is no terminal transition on a
+HALT. Left held, it makes the *next* run on this issue refuse with a foreign-owner
+block until the lease's ceiling lapses hours later.
+
+So after the Final Report, hand the lease back with the pipeline tool's
+**`session-release`** subcommand, passing this run's issue number and `run_id`
+(see the repo context probe for the exact invocation). It is ownership-checked
+and best-effort: a wrong or already-released `run_id` is a safe no-op, so run it
+whenever in doubt and never let its output change your reported outcome.
+
+Do this on the exits nothing else observes:
+
+- the **3d.4 REVIEW self-check HALT**
+- a **`blocked`** router decision (3a / 3e)
+- the **iteration cap** being reached
+
+The **merged** exit needs no action from you — completing the MERGE stage
+releases the lease in the tool layer, on the marker write itself. Running the
+step there anyway is harmless (it reports no lease held), but it is not what
+makes the merged path correct.
+
 ## Relationship to /sdlc (in this repo)
 
 | | `/sdlc` (in this repo) | `/do-sdlc` |
--- a/agent/supervised_run.py
+++ b/agent/supervised_run.py
@@ -37,8 +37,12 @@ the single source of truth for "is this run still the owner."
 
   All three renewers write only after confirming they still own the lease, so a
   signal is never republished over a successor's. The fourth writer matters
-  most for long runs: the heartbeat self-exits at 4h, after which the ledger
-  gate is the only renewer left.
+  most for long runs, because the heartbeat's exit is no longer a single 4h
+  backstop (issue #2714): it exits within ~2 checks of its supervising
+  ``claude`` process dying (releasing the lease and clearing this signal), at
+  90 minutes when no supervisor identity resolved (leaving both to their own
+  TTL), and otherwise still at the unchanged 4h ceiling. After any of those the
+  ledger gate is the only renewer left.
 
   Issue #2659: the renew half was missing for the first two years of this key's
   life, so the signal expired 1800s into every pipeline while the lease was
--- a/docs/features/sdlc-issue-ownership-lock.md
+++ b/docs/features/sdlc-issue-ownership-lock.md
@@ -313,11 +313,11 @@ Three paths extend the lease, and all three refresh the signal, each only after 
 
 | Renewer | Ownership proof before the signal write |
 |---|---|
-| `tools/sdlc_lease_heartbeat.py` (local supervisor, not in the table above) | its peek-confirmed `owner_run_id == run_id` match |
+| `tools/sdlc_lease_heartbeat.py` (local supervisor, not in the table above) | its peek-confirmed `owner_run_id == run_id` match, then a `renew_only=True` extend |
 | `agent/session_executor.py::_tick_issue_lock_renewal` (worker, 60s) | an `acquired` result |
 | `tools/_sdlc_utils.py::revalidate_ledger_lease` (every ledger-write gate — see below) | an `acquired` result |
 
-The third matters more than it looks, and it reaches further than the two tools first associated with it. `revalidate_ledger_lease` is non-peek, so **all five of its callers** renew the lease on their own: `tools/sdlc_stage_marker.py` (three sites), `tools/sdlc_dispatch.py`, `tools/sdlc_verdict.py`, `tools/sdlc_meta_set.py`, and `tools/sdlc_review_finalize.py`. The local heartbeat self-exits at `MAX_LIFETIME_SECONDS` (4h), after which those gates are the only renewer left. Without the refresh there, a pipeline past ~4.5h would lose its signal under a lease those writes were holding up, reproducing the #2659 wedge exactly. The observed incidents were at roughly 3.5h, so that window is reachable rather than theoretical.
+The third matters more than it looks, and it reaches further than the two tools first associated with it. `revalidate_ledger_lease` is non-peek, so **all five of its callers** renew the lease on their own: `tools/sdlc_stage_marker.py` (three sites), `tools/sdlc_dispatch.py`, `tools/sdlc_verdict.py`, `tools/sdlc_meta_set.py`, and `tools/sdlc_review_finalize.py`. The local heartbeat's own renew loop stops at `MAX_LIFETIME_SECONDS` (4h, on a resolved supervisor) or `UNSUPERVISED_MAX_LIFETIME_SECONDS` (90 min, unresolved) -- see [Heartbeat exit conditions and release paths](#heartbeat-exit-conditions-and-release-paths-issue-2714) below -- after which those gates are the only renewer left. Without the refresh there, a pipeline past ~4.5h would lose its signal under a lease those writes were holding up, reproducing the #2659 wedge exactly. The observed incidents were at roughly 3.5h, so that window is reachable rather than theoretical.
 
 Ownership-first ordering is the safety property. Writing the signal before the ownership check would republish a non-owner's `run_id` over a live successor's, which **denies the successor's forks their inheritance signal and re-creates the #2659 wedge**. It does not hand anyone the wrong `run_id`: `supervised_run_status` reports `live=True` only when the signal's `run_id` equals the lock peek's `owner_run_id`, so a mis-owned signal reads not-live and is never inherited. That liveness anchor is what bounds the damage to a stand-down rather than a takeover.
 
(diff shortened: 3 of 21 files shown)


valorengels:
## Review: APPROVED

Reviewed `main...HEAD` (20 files) against `docs/plans/sdlc-lease-heartbeat-supervisor-lifetime.md`. All ten load-bearing invariants hold in code, not just in comments. Two concerns were raised on the first pass and both are fixed in b4fca9006.

### Invariants verified

- **A renewer can never mint.** The `renew_only` block in `models/session_lifecycle.py` sits after the `peek` and `not run_id` branches but *before* the `SET NX` and before the `raw is None` "treat as acquired" race branch, so both minting paths are unreachable. `_RENEW_IF_VALUE_MATCHES_LUA` only `SET`s when `GET == ARGV[1]`, and a `GET` on an absent key returns `false`, so the script is structurally incapable of recreating a deleted key.
- **The default path is behavior-preserving.** `_healed_renewal_payload` is a faithful extraction of the old inline block: same `target_repo` precedence, same `machine_id` immutability, same unconditional `renewed_at` overwrite. `renew_only=False` still renews through the original `_R.set(..., ex=ttl)`, and a test asserts the default still acquires via `nx=True` with `eval` never called.
- **Fail-closed asymmetry.** The outer handler returns not-acquired under `renew_only` while preserving fail-OPEN for every other caller, which #2446 depends on. Both halves are asserted.
- **#2537 preserved.** Peek-before-renew intact; exits on an absent lease and on a foreign owner; never mints on a free key.
- **#2446/#2451 preserved.** A dead *payload* pid still renews. The relevant test runs the real `touch_issue_lock` against test Redis with distinct payload and supervisor pids and a pid-keyed psutil double, so confusing the two fails the assertion rather than passing by luck.
- **Supervisor identity.** Three-tier resolution (`CLAUDE_PID` env, `psutil` ancestry walk for a `claude` ancestor, unresolved), every tier exception-wrapped. No `os.getppid()` anywhere, enforced by a source-inspection test; `ppid == 1` is never treated as a death signal.
- **Death handling asymmetry.** Confirmed supervisor death actively releases the lease and clears the signal inside a best-effort `try/except` that cannot crash the loop. The unsupervised 90-minute ceiling deliberately does NOT release, asserted with `release.assert_not_called()`.
- **The `max_lifetime` sentinel is live at both ends.** `run_heartbeat`'s default is `None` and resolves internally only when unset, and `main()`'s `--max-lifetime` argparse default is also `None`, so the sentinel is not dead code. An explicit caller value wins over both constants.
- **Stage-marker release leg.** `release_run()` fires after `complete_stage` on a non-idempotent MERGE, sits after the already-completed early return, and is wrapped so an exception cannot fail the marker write. Fire / no-fire-on-idempotent / no-fire-on-other-stage / raise-doesn't-fail are all covered.
- **No raw Redis.** `tools/sdlc_session_release.py` goes through `release_issue_lock` and `clear_supervised_run_signal` only, enforced by a source-inspection test.

Scope matches the plan's No-Gos: `/sdlc`'s SKILL.md is untouched, no `runintent` residue, no commented-out code, ruff clean.

### Concerns raised and fixed (b4fca9006)

**C1 — the fail-closed renewal result was discarded.** The heartbeat called `touch_issue_lock(..., renew_only=True)` and threw away the return, then unconditionally called `write_supervised_run_signal`. So a CAS that lost to a concurrent release, or the fail-closed handler firing on a Redis blip, still republished `session:supervisedrun:{N}` with a fresh 1800s TTL for a lease no longer held — precisely the shape `renew_only` fails closed to prevent, and the comment justifying that branch overstated what actually happened. Blast radius was small because `supervised_run_status` reports `live=True` only when the lock's `owner_run_id` matches the signal, which is why this was a concern rather than a blocker. Now the result is consumed and the heartbeat takes its lease-lost exit on that tick.

The pre-existing race test collected signal writes but never asserted on them, so the gap was invisible to the suite. Both the CAS-lost and fail-closed-error paths are now asserted, and the fix was demonstrated red: with the guard disabled the new test fails with 19 signal republishes for a lease not held.

**C2 — a stale claim in the canonical renewer enumeration.** `agent/supervised_run.py` still read "the heartbeat self-exits at 4h, after which the ledger gate is the only renewer left." That block is explicitly labelled "Keep this list exact. It is the canonical enumeration," and the equivalent sentence in `docs/features/sdlc-issue-ownership-lock.md` had already been updated, so this was the one place the old model survived. It now names all three exits.

### Non-blocking observation

The L2 supervisor watch is not `/do-sdlc`-scoped: it applies to any non-worker local mint, including an interactive `/sdlc` router run. Closing that window now frees the lease in about two minutes instead of holding it to the ceiling. This matches the plan's L2 text (only the skill-body *release* was scoped to `/do-sdlc`) and is benign because the stage ledger is issue-keyed, but it is slightly wider than Risk 3's "the `/sdlc` per-stage router is untouched" sentence implies.

### Testing

Targeted suites, `POPOTO_TEST_DB=11`: heartbeat + release + stage-marker + session-ensure + session-lifecycle, 334 passed. Broader SDLC batch (16 files) shows a rotating failure set from pre-existing shared-test-db contention — verified against unmodified `main`, which fails **more** tests on the identical batch (7 vs 4), including the same ledger-cutover test. Not a regression from this PR. `ruff check` and `ruff format --check` clean.

PR #2683, opened 2026-08-07
Enforce test-DB ownership so the unit suite stops rotating (#2628)

tomcounsell:
Closes #2628. Plan: `docs/plans/suite-failure-rotation-db-ownership.md`.

Two `tests/unit/` runs on the same commit produced different failure sets, every failing node passing in isolation. That is a defect in the measuring instrument: a suite whose failure set changes between identical runs cannot answer "did my change break something."

Three independent writers each corrupted state a **different live pytest process** owned. This PR closes all three and promotes db ownership from a convention every call site re-derives to an invariant enforced at the point of damage.

**One docstring-only production line.** `tools/redis_flush_guard.py` picked up a rename-only docstring edit (no behavior change); everything else is under `tests/` plus four docs. No service restart.

---

## Writer 1 — a `flushdb()` on a db this process never claimed

`tests/unit/test_redis_flush_guard.py` derived `gw{N} -> db{N+1}` in a helper whose docstring claimed it "matches `redis_test_db()`" — true until #2606 replaced that derivation with an flock claim. `tests/unit/test_conftest_isolation_guards.py` hardcoded db 15 and called it "scratch space", but 15 is `_TEST_DB_POOL_MAX`, the top slot of the pool.

The guard now enforces ownership: `flushdb()` succeeds only against a db in `claimed_test_dbs()`, and db 0 is simply a db no test process can claim. One wrapper, not two idioms — the db-0 branch keeps its own message and its 2026-06-03 rationale.

**Binding measurement.** A child process holds `flock` on slot 1, so slot 1 is demonstrably another live process's. The flush is driven through a `SimpleNamespace` stub with no connection and an `execute_command` spy, so **no code path in this PR flushes a db the test process does not own, on any branch or in any repo state**. Red is the *absence* of a `RuntimeError`, never a successful flush.

Red, on pre-fix code:
```
tests/unit/test_conftest_isolation_guards.py:963: in _reached
    pytest.fail("guard did not intercept: a flush reached the client layer")
E   Failed: guard did not intercept: a flush reached the client layer
```
Green on the branch (`TestFlushOwnershipGuard`, 7 cases, ~1 s).

## Writer 2 — an unrestored `importlib.reload`

[...]

Issue #2628, opened 2026-08-07
main's full-suite failure set rotates run to run (3 then 5, only one overlapping); the db_claim guard from #2606 fails in both


tomcounsell:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/suite-failure-rotation-db-ownership.md

Two consecutive `tests/unit/` runs on the same `main` commit (`acf1f3129`), same machine, deterministic ordering (`-p no:randomly`), no other pytest process running:

| Run | Result | Failures |
|---|---|---|
| 1 | 3 failed, 11906 passed | `test_session_lifecycle.py::TestRunIdentityAcrossProcesses::test_run_identity_foreign_run_id_blocked_across_processes`<br>`test_conftest_isolation_guards.py::TestPerProcessDbClaim::test_dead_holder_slot_is_reclaimed`<br>`test_youtube_search.py::TestSearchIntegration::test_real_search_limit` |
| 2 | 5 failed, 11904 passed | `test_session_archive.py::test_stuck_sentinel_never_clobbers_healthy_full_redis`<br>`test_session_archive.py::test_stuck_sentinel_does_not_mask_genuinely_missing_row`<br>`test_session_archive.py::test_resume_attempts_reset_on_completion_across_repeated_recoveries`<br>`test_conftest_isolation_guards.py::TestPerProcessDbClaim::test_dead_holder_slot_is_reclaimed` |

**Only one test fails in both.** Every failing test passes in isolation (verified for the two checked directly). Ordering was pinned, so this is not pytest-randomly; the remaining non-determinism is xdist work distribution plus shared Redis.

## Regression window is narrow and identifiable

`tests/unit/` was **fully green twice** earlier the same day at `805b7f1ae` (11817 passed, 0 failed, same invocation). Roughly ten PRs from several concurrent sessions landed between `805b7f1ae` and `acf1f3129`. The instability appeared in that window.

## The one consistent failure is from #2606, and that is mine

`TestPerProcessDbClaim::test_dead_holder_slot_is_reclaimed` fails in **both** runs and passes in isolation. It tests the per-process Redis db claim introduced by PR #2606 (issues #2603/#2605) — the very machinery meant to stop tests fighting over db slots. A dead-holder-reclaim path that misbehaves under real concurrent load is the plausible shape: the guard is being exercised for the first time at full-suite scale.

[...]

PR #2786, opened 2026-08-13
Make test subprocesses inherit the parent's claimed test DB (#2763)

### LINKED ISSUE #2763: test_sdlc_tool_wrapper subprocess harnesses write to production Redis db0
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/subprocess-harness-test-db-inheritance.md

Pre-existing on `main`, surfaced during the PR #2706 (issue #2637) review. Filed separately because it is independent of that PR and should not gate it.

`tests/unit/test_sdlc_tool_wrapper.py:232` launches its harness with:

```python
subprocess.run([sys.executable, "-c", harness], cwd=REPO_ROOT, ...)
```

with no `env=subprocess_env(...)`. Popoto resolves `REDIS_URL` at import time, and the variable is unset in that child, so the subprocess lands on **db0 — production** rather than a claimed test DB. The parent process isolates correctly via `tests/db_claim.py` (`fcntl.flock` over the 1-15 pool); the child does not inherit that claim.

## Evidence

A real `PipelineLedger` row for the fake issue #999999 is currently sitting in production db0, key `PipelineLedger:tomcounsell//ai{&#58;}999999`, read via the ORM. Its history length was already 3 before the review touched anything, so it predates this session. One mutation-check run during the review appended a fourth entry. That row was deliberately left in place rather than deleted, since removing data from production db0 unilaterally is not a reviewer's call.

Under the shipped code the write never actually happens in the reviewed test, because the injected `boom` raises before the ledger write. So today this is latent, not actively corrupting. It is one refactor away from not being latent.

## Fix

Pass an explicit environment into every `subprocess.run` in this file so the child inherits the parent's claimed test DB, using the existing `subprocess_env(...)` helper. Then audit the rest of the suite for the same shape: any test that shells out with `cwd=REPO_ROOT` and no `env=` has the same exposure.

## Cleanup, needs a decision

Someone with the authority to do so should decide whether to delete the stray `PipelineLedger` row for issue #999999 from db0 via the ORM (`Model.query.filter()` then `instance.delete()`, never raw Redis, scoped to that single key). Leaving it is harmless but it is test residue in production state.

## Recon Summary

Verified against `main` @ `90c0e81e4` (2026-08-13).

**Confirmed**

- `tests/unit/test_sdlc_tool_wrapper.py` has three `subprocess.run(...)` calls that pass `cwd=str(REPO_ROOT)` with **no** `env=`: lines 172, 189, 230. Two run `sys.executable -c <harness>` (verdict / dispatch loud-exit tests), one runs `python -m tools.sdlc_verdict get`.
- `tests/db_claim.py:187` already exports the `subprocess_env(*, project_root=None, **extra)` helper, on `main` since PR #2606. It sets `REDIS_URL` from `claim_test_db()` and optionally pins `PYTHONPATH`. Calling it is the whole fix.
- `tests/conftest.py:571` (`redis_test_db`, autouse) isolates by **replacing the Popoto client objects in-process**. It never writes `REDIS_URL` into `os.environ`. That is why a child inherits nothing and Popoto's import-time `REDIS_URL` resolution lands on db0.

**Revised**

- The issue cites `test_sdlc_tool_wrapper.py:232`; after PR #2706 merged (2026-08-13T06:28Z, `171c1871a`) the call site is **line 230**. Same code, drifted by two lines.
- Scope is wider than "no `env=`". Three more sites in the same file pass `env=os.environ.copy()` (lines 87, 101, 129), which is equally unsafe — a copied parent env carries no test-db pointer, per the conftest finding above. Of those, only line 129 (`test_dispatch_from_foreign_cwd_with_own_tools_succeeds`, which runs `stage-query --issue-number 999999`) actually reaches Python; lines 87 and 101 exit 2 inside the bash wrapper on a deliberately bad `AI_REPO_ROOT`. All three should still be converted, so the file has one uniform shape.
- Lines 69/75/81 pass neither `cwd` nor `env` and terminate at the bash usage/`--help` path without touching Redis. No change needed, but they are the reason a blanket "every `subprocess.run` in this file" rule needs a stated exception.

**Suite-wide audit (AST scan of `tests/**/*.py`)**

- 115 `subprocess.*` call sites pass `cwd=` with no `env=`. The large majority are `git` invocations inside `tmp_path` scratch repos and carry no Redis exposure.
- Narrowing to calls that invoke a Python interpreter or repo script against a repo checkout yields **~33 call sites across 13 files**: `tests/unit/test_sdlc_meta_set.py` (6), `test_sdlc_session_ensure.py` (3), `test_sdlc_stage_marker.py` (2), `test_sdlc_stage_query.py` (1), `test_sdlc_tool_wrapper.py` (3), `test_memory_search_cli.py` (2), `test_session_telemetry.py` (1), `test_evaluate_build.py` (3), `test_worker_entry.py` (2), `test_worker_supervisor.py` (1), `test_migrate_strip_pid_fields.py` (1), `tests/integration/test_sdlc_dispatch.py` (2), `tests/integration/test_design_system_pipeline.py` (4), `tests/integration/test_session_telemetry_e2e.py` (1). These are the audit surface.

**Pre-requisites / constraints**

- Open PR #2683 (`session/suite-failure-rotation-db-ownership`, issue #2628) owns `tests/conftest.py`, `tests/db_claim.py`, and `tests/unit/test_conftest_isolation_guards.py`. This work must not touch those three files; it only *calls* the existing `subprocess_env`, which #2683 keeps and extends (it adds a `POPOTO_TEST_DB` export and makes an unclaimed `flushdb()` raise). Calling the helper is forward-compatible with that PR either way it lands.

**Dropped**

- Consolidating the three ad-hoc local helpers (`_isolated_subprocess_env` in `tests/integration/test_sdlc_dispatch.py:272`, `_subprocess_env` in `tests/unit/test_session_lifecycle.py:1602` and `tests/unit/test_redis_flush_guard_prod.py:44`) into the shared `db_claim.subprocess_env`. Real cleanup, but a refactor, not this bug.

valorengels:
Closes #2763.

Plan: `docs/plans/subprocess-harness-test-db-inheritance.md`

## Problem

Test subprocesses inherit no `REDIS_URL`. Popoto resolves it at import time and falls back to `redis://localhost:6379` — db0, **production** — so a child spawned from a test writes to prod while its parent is correctly isolated on a claimed db from the `[1..15]` pool. The parent's claim lives in in-process Popoto client objects, never in the environment, so nothing reaches the child.

The filed defect (`tests/unit/test_sdlc_tool_wrapper.py`) is latent today only because an injected `boom` raises before the ledger write fires. A stray `PipelineLedger` row for fake issue `999999` is sitting in prod db0 as evidence that it has fired before.

## What this does

Routes every reachable test subprocess through the existing `tests.db_claim.subprocess_env()` helper, which sets `REDIS_URL` from `claim_test_db()` — the same per-process claim the autouse `redis_test_db` fixture uses, so parent and child share one db by construction.

Adds a mechanical AST guard, `tests/unit/test_subprocess_test_db_isolation.py`, that fails the suite when a new call site reintroduces the shape. It resolves one hop through module-local delegators, `self`/`cls` method delegators, and the `env = subprocess_env(); env.pop(...)` assign-mutate pattern, so legitimate wrappers are accepted rather than allowlisted into noise. Four self-tests pin the predicate using source *strings* rather than live code, so the guard cannot quietly exempt itself.

Scope: 766 files scanned, 0 violations remaining. Of the 30 sites adjudicated in the final sweep, 5 were converted and 25 allowlisted with a written reason.

## Allowlist

Exactly two reasons are permitted, and each entry carries its reason on the line:

- `[#2628]` — 7 sites in files owned by open PR #2683. Touching them here would collide, so they are allowlisted with a note to remove the entry and convert the site once #2683 lands. Flagged on #2628 ([comment](https://github.com/tomcounsell/ai/issues/2628#issuecomment-5279899837)) because one of the seven, `test_redis_flush_guard_prod.py:666`, is a dict literal (`env={**_subprocess_env(), ...}`) that a helper-keyed sweep resolves to no helper and skips.
- `[standalone-script]` — the child is a single-file hook or script that cannot import popoto and so cannot reach Redis.

## Verification

| Check | Result |
|---|---|
| Isolation guard, 766 files | 0 violations |
| `test_sdlc_tool_wrapper.py` (filed defect) | passes |
| No bare `os.environ.copy()` in target file | 0 occurrences |
| Subprocesses with repo `cwd` and no `env` in target file | 0 |
| Anti-criterion: no #2683-owned file in the diff | 0 files |
| `ruff check` / `ruff format --check` | exit 0 / exit 0 |
| All 31 converted unit-test files, `POPOTO_TEST_DB=10` | 955 passed, 3 pre-existing failures |

### On those 3 failures

They are #2628's rotating-failure signature, not regressions. Evidence:

- The set rotates between identical runs (`test_memory_timeline` in one, `test_sdlc_stage_marker::TestWriteMarker` in the next).
- Every failure is shaped "a row written moments ago is gone" (`LEASE_ABSENT`, `0 >= 2`) — the exact symptom `db_claim.subprocess_env`'s own docstring describes for concurrent pytest processes sharing the claim pool.
- **A baseline run of the same file set on `main` reproduces `test_sdlc_stage_marker::TestCLI::test_with_issue_number_outputs_json` with the pre-conversion code**, and rotates a different second failure. Pre-existing.
- One of them, `test_memory_timeline::TestTimeline::test_timeline_with_records`, uses no subprocess at all and is untouched by this diff.

The one converted site among them, `test_with_issue_number_outputs_json`, previously hand-derived its db from `POPOTO_REDIS_DB.connection_pool.connection_kwargs`. `redis_test_db` (`tests/conftest.py:608`) patches that client from `claim_test_db()`, which is precisely what `subprocess_env()` reads, so the conversion is behavior-preserving.

## Follow-up

The prod `PipelineLedger:tomcounsell//ai{:}999999` row is deleted via the ORM once this merges.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -474,3 +474,53 @@ Source modules with no test coverage. Priority targets for new tests.
 
 **Partially covered** (operational layer added in #936):
 - `bridge/email_bridge.py` — parsing, SMTP output, routing, and thread continuation have full coverage. Operational layer (`main()`, `_poll_imap()` batch cap, `_email_inbox_loop()` health timestamp) now covered via unit and integration tests.
+
+## Subprocess Test-DB Inheritance (issue #2763)
+
+Any test that shells out to a subprocess which can reach Popoto — a Python
+interpreter running a repo module, the sdlc-tool `WRAPPER`, or an inline `-c`
+script against a repo checkout — must pass `env=subprocess_env(...)` from
+`tests.db_claim`.
+
+**Why**: Popoto resolves `REDIS_URL` at *import time* and falls back to
+`redis://localhost:6379` — db0, production — when the variable is unset. The
+parent pytest process claims a test db from the pool `[1..15]` via
+`tests/db_claim.py`'s `fcntl.flock`, but that claim lives only in the parent's
+in-process Popoto client objects; `os.environ` is never mutated. The
+environment is therefore the only channel to a child, and `subprocess_env` is
+the bridge. A child launched without it silently reads and writes production
+db0.
+
+**How to use it**: `subprocess_env(*, project_root=None, **extra)`. Thread
+extra variables as keyword arguments (`subprocess_env(AI_REPO_ROOT=...)`)
+rather than hand-building a dict. If a site also needs keys removed, the
+accepted shape is `env = subprocess_env(); env.pop("NAME", None)` — assign,
+then mutate with `.pop(<literal>, None)` / `.update(...)`, never rebind.
+
+`project_root=` is **opt-in, not a default**. It prepends the path to the
+child's `PYTHONPATH`. Pass it when the child must resolve repo modules from
+this checkout; omit it when the test asserts something about import order or
+module resolution. `tests/unit/test_sdlc_tool_wrapper.py::test_dispatch_from_foreign_cwd_with_own_tools_succeeds`
+is the worked example of a deliberate omission — it pins the wrapper's own
+module-resolution order against a decoy `tools/` package, so prepending
+`REPO_ROOT` to `PYTHONPATH` would resolve the import for reasons other than
+the wrapper's doing.
+
+Never re-derive a db number by hand. Reading
+`POPOTO_REDIS_DB.connection_pool.connection_kwargs` to rebuild a `REDIS_URL`
+is the anti-pattern this work removed; `claim_test_db()` via `subprocess_env`
+is the only source.
+
+The enforcing guard is `tests/unit/test_subprocess_test_db_isolation.py`, an
+AST scan of `tests/**/*.py`. It matches in-scope call sites on **argv**
+(`sys.executable`, a `PYTHON`-containing identifier, a `"-m"` element,
+`WRAPPER`, or a `scripts/` string); `cwd=` is deliberately not part of the
+predicate, since what determines whether a child can import popoto is what is
+executed, not where it is executed from. Exemptions exist only as
+`SKIP_ARGV0 = {"git"}` (a `git` child never imports Python) and a commented
+`ALLOWLIST` of `path:line` entries, each carrying one of exactly two reasons:
+`[#2628]` (the file is owned by open PR #2683 — fix it there when it lands)
+or `[standalone-script]` (the child imports no repo package). **When
+reachability is unclear, convert the site — do not allowlist it.**
+
+When #2683 lands, this section folds into `docs/features/test-db-ownership.md`.
--- a/tests/_worker_guard.py
+++ b/tests/_worker_guard.py
@@ -66,6 +66,10 @@ def _pgrep_worker_pids() -> set[int]:
             capture_output=True,
             text=True,
             timeout=5,
+            # `pgrep` cannot import popoto and does not need REDIS_URL. The env
+            # is here because the isolation guard's argv heuristic matches the
+            # word "python" inside the *pattern* string; passing it is cheaper
+            # and less brittle than an allowlist entry.
             env=subprocess_env(),
         )
         return {int(tok) for tok in out.stdout.split() if tok.strip().isdigit()}
--- a/tests/_worker_guard.py
+++ b/tests/_worker_guard.py
@@ -64,6 +64,7 @@ def _pgrep_worker_pids() -> set[int]:
             capture_output=True,
             text=True,
             timeout=5,
+            env=subprocess_env(),
         )
         return {int(tok) for tok in out.stdout.split() if tok.strip().isdigit()}
     except Exception:
--- a/tests/_worker_guard.py
+++ b/tests/_worker_guard.py
@@ -25,6 +25,8 @@ from __future__ import annotations
 import re
 import subprocess
 
+from tests.db_claim import subprocess_env
+
 # A command line for the launchd worker joins to `<python> -m worker`. Require
 # BOTH a python-ish executable AND the `-m worker` token so an unrelated
 # process that merely mentions "worker" never trips the guard. `python -m
--- a/tests/e2e/test_telegram_flow.py
+++ b/tests/e2e/test_telegram_flow.py
@@ -16,6 +16,8 @@ from pathlib import Path
 
 import pytest
 
+from tests.db_claim import subprocess_env
+
 
 @dataclass
 class TestMessage:
--- a/tests/e2e/test_telegram_flow.py
+++ b/tests/e2e/test_telegram_flow.py
@@ -40,6 +40,7 @@ class TestTelegramE2EFlow:
             capture_output=True,
             text=True,
             cwd=Path(__file__).parent.parent.parent,
+            env=subprocess_env(),
         )
         return "RUNNING" in result.stdout
 
--- a/tests/integration/test_design_system_pipeline.py
+++ b/tests/integration/test_design_system_pipeline.py
@@ -25,6 +25,8 @@ from pathlib import Path
 
 import pytest
 
+from tests.db_claim import subprocess_env
+
 REPO_ROOT = Path(__file__).resolve().parents[2]
 FIXTURE_DIR = REPO_ROOT / "tests/fixtures/design_system"
 
@@ -88,6 +90,7 @@ def test_all_pipeline_produces_committed_artifacts(worktree_fixture: Path):
         capture_output=True,
         text=True,
         cwd=str(REPO_ROOT),
+        env=subprocess_env(project_root=str(REPO_ROOT)),
     )
     assert result.returncode == 0, result.stderr
     assert (worktree_fixture / "design-system.md").is_file()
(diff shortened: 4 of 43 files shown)


valorengels:
## Review: PR #2786 (#2763) — APPROVED

Reviewed against `docs/plans/subprocess-harness-test-db-inheritance.md`, with an adversarial pass on the guard predicate specifically hunting false negatives — a shape the guard accepts that still leaves the child on db0 is the only failure mode that matters here, since the guard's whole value is that it cannot be quietly bypassed.

### Findings (4 CONCERNs, all fixed in `340c3fec0`)

The initial predicate had four holes, each verified accepted by running the fixture through the *previous* `scan_source` before the fix:

1. **`env.pop("REDIS_URL", None)` and `env.update({"REDIS_URL": ...})` were accepted.** `env["REDIS_URL"] = ...` was already rejected as a Subscript store, so the guard blocked the loud spelling of exactly the write it let through in two quieter ones. `.pop`/`.update` remain permitted for every other key (the strips are load-bearing in several converted tests); they are rejected only when the literal `REDIS_URL` appears in the call's arguments.
2. **`import subprocess as _sp` made an entire file invisible.** The scan resolved the attribute root against the literal name `"subprocess"` and recorded aliases only from `ImportFrom`, never `Import.asname`. `tests/unit/test_worktree_manager.py` has 20 such call sites the guard could not see. All 20 run `git` today, so there was no live db0 exposure — but the file reported clean for the wrong reason, and the next `_sp.run([sys.executable, ...])` added there would have shipped unguarded and green.
3. **`subprocess.run(args=[...])` bypassed the predicate entirely** via an early `if not node.args: continue`. No live site uses it; durability only.
4. **A walrus rebind defeated "bound exactly once."** Contrived, but the rule says never rebound.

Also widened the argv predicate to the repo's `[project.scripts]` console-script names, read from `pyproject.toml` rather than hardcoded: a bare `["valor-session", "list"]` argv names no interpreter and no `-m`, yet the entry point imports popoto.

### What was verified as correct

- **Anti-criterion holds.** None of `tests/conftest.py`, `tests/db_claim.py`, `tests/unit/test_conftest_isolation_guards.py`, `tests/unit/test_redis_flush_guard_prod.py` appears in the diff. PR #2683 is still open, so the `[#2628]` deferral remains the right disposition.
- **Allowlist is exactly the live set.** Re-running the scan with `ALLOWLIST` emptied yields 33 violations against 33 entries, with an empty symmetric difference — no stale entry silently suppressing a shifted line, no live site uncovered. 7 are `[#2628]`; the rest `[standalone-script]`.
- **Every `[standalone-script]` claim was checked against the child's actual imports**, including the two that reach `hook_utils` (`destructive_git_shapes` imports `re/shlex/pathlib`; `constants` imports `json/os/sys/datetime/pathlib`). None can import popoto.
- **Self-tests use source strings** handed to `scan_source`, never live code, so the guard cannot exempt itself. Now nine.
- **The one rewritten hand-derived env is behavior-preserving.** `test_sdlc_stage_marker.py::test_with_issue_number_outputs_json` previously read its db from `POPOTO_REDIS_DB.connection_pool.connection_kwargs`. `tests/conftest.py:611` patches that client from `claim_test_db()`, which is exactly what `subprocess_env()` reads. Verified independently in the source, not assumed.
- **`project_root` audit**: every `project_root=` in the diff resolves to a `str`, and the three deliberate omissions each carry a why-comment. `test_sdlc_tool_wrapper.py:69/75/81` correctly take bare `subprocess_env()` — adding `AI_REPO_ROOT` there would break `test_wrapper_no_args_exits_2`'s usage-path assertion.

### Test results

All 31 converted unit-test files under `POPOTO_TEST_DB=10`: **962 passed, 0 failed** after the patch. An earlier run of the same set showed 3 failures; they were adjudicated as pre-existing #2628 shared-test-db contention, not regressions:

- the set rotated between identical runs;
- every failure was shaped "a row written moments ago is gone" (`LEASE_ABSENT`, `0 >= 2`), the exact symptom `db_claim.subprocess_env`'s docstring describes for concurrent pytest processes sharing the claim pool;
- **a baseline run of the same file set on `main` reproduced `test_sdlc_stage_marker::TestCLI::test_with_issue_number_outputs_json` with the pre-conversion code**, and rotated a different second failure;
- one of them uses no subprocess at all and is untouched by this diff.

`ruff check` and `ruff format --check` both exit 0 over 767 files.

### Not fixed here

`tests/integration/test_memory_mcp_server.py::test_fresh_shell_import_resolution` fails on this machine, before and after this branch identically: the homebrew `python3` the test deliberately targets has no `mcp` SDK installed. Local machine state, out of scope. The test's env handling was tightened regardless — building from `subprocess_env` means the child inherits the parent environment rather than the old three-key whitelist, so the `sys.path`-steering variables are now stripped explicitly to keep the assertion's subject (module resolution from `PYTHONPATH` alone) intact.

Issue #2789, opened 2026-08-13
Drop the --blockers / --tech-debt aliases from `sdlc-tool verdict finalize` once every machine has run /update


valorengels:
## What

`sdlc-tool verdict finalize` accepts two spellings for each finding count:

- `--blocker-count`, aliased `--blockers`
- `--tech-debt-count`, aliased `--tech-debt`

The aliases are argparse aliases on the **same** `add_argument` call — one flag with two spellings, not a second code path — and they are documented nowhere. Removal is a one-line diff per flag in `tools/sdlc_verdict.py::_build_parser`, plus the two alias assertions in `TestFinalizeCountFlagSpellings` (`tests/unit/test_sdlc_verdict.py`).

## Why they exist, and why that reason expires

`.claude/skills-global/do-pr-review/SKILL.md` is hardlinked into `~/.claude/skills/` by `/update`. A machine that has not yet run `/update` after the #2767 merge is still running the old skill body and still emitting `--blockers` against a freshly-merged `sdlc-tool`. The aliases cover exactly that **cross-machine propagation window**.

Note the in-repo argument runs the other way: the only in-repo callers are skills updated in the same PR, so for this checkout the aliases were already unnecessary. The window is the whole justification.

Once every machine has run `/update`, the window is closed and the aliases are pure legacy surface — the thing NO LEGACY CODE TOLERANCE exists to prevent.

## Actionable when

Every machine that runs SDLC lanes has run `/update` past the #2767 merge commit. Confirm before removing rather than assuming; a machine that has been offline is the case that bites.

## Acceptance

- [ ] `--blockers` / `--tech-debt` no longer parse; `--blocker-count` / `--tech-debt-count` are the only spellings.
- [ ] The alias-specific tests in `TestFinalizeCountFlagSpellings` are removed (not just loosened), while the tests covering the surviving spellings, the `default=None` vs `0` distinction, and the prose-is-a-loud-failure behavior stay.
- [ ] No doc, skill body, or addendum references the old spellings.

Follow-up to #2767, per the risk mitigation recorded in `docs/plans/verdict-finalize-cluster.md` (Risk 3).

valorengels:
Consolidated into #3065 — the SDLC control plane routes on asserted facts rather than read facts. This issue's incident detail is carried there. Closing as a member of that class, not as invalid.

Issue #2787, opened 2026-08-13
A "READY TO BUILD (with concerns)" critique verdict does not mandate a re-critique, so the revision that closes the concerns is never itself reviewed


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/with-concerns-revision-recritique.md

## Problem

`/do-plan-critique` can return **READY TO BUILD (with concerns)**. That verdict lets BUILD start, and the concerns are swept into a revision pass. Nothing re-runs the critique against that revision, so **the revision is the only part of the plan that never gets reviewed** — even though it is written last, under the least scrutiny, and specifically to address findings the critics called out as risky.

This is not hypothetical. `docs/plans/verdict-finalize-cluster.md` (#2740 / #2767 / #2769) ran four rounds, and the pattern is legible in its own Critique Results tables:

| Round | Verdict | What the next round found |
|---|---|---|
| 1 | NEEDS REVISION (2 blockers) | — |
| 2 | READY TO BUILD (with concerns) — 4 concerns | Round 3's concern was **a defect in round 2's own fix** |
| 3 | READY TO BUILD (with concerns) — 1 concern | Round 4 confirmed clean |
| 4 | READY TO BUILD (no concerns) | — |

Round 2 prescribed a two-way `isinstance` guard (`dict` vs. "the str case") for `_extract_head_sha`. Round 3 caught that the duality it mirrored is actually **three-way**: `verdicts.get("REVIEW")` returns `None` for every issue not yet reviewed — the majority of live ledgers — and a literal reading of round 2's prescription routes that `None` into `re.compile(...).search()`, raising `TypeError`. That exception is swallowed by `tools/sdlc_next_skill.py::_resolve_enriched`'s broad `except` into an empty ledger, routing a fully-worked issue back to `/do-plan`.

Round 3 stated it plainly: this was **"a defect in the round-2 fix's own prescription"**, and it was *worse than the blocker round 2 was written to prevent* — fail-open-to-amnesia on the common path, versus fail-closed-to-stale on a rare one.

Had round 2's "with concerns" verdict been treated as terminal, that defect ships.

## Why the current shape produces this

A "with concerns" verdict is doing two incompatible jobs at once:

1. **Gate**: "the plan is sound enough to build" — which is true, and is why the verdict exists rather than forcing NEEDS REVISION for every nit.
2. **Work order**: "here are N changes to make to the plan" — changes that are, by construction, unreviewed once made.

[...]

PR #2728, opened 2026-08-11
fix(docs-auditor): word-anchor stale terms and enforce a path-existence invariant

valorengels:
## Summary

The docs auditor's auto-fixer invented references to a module that has never existed and self-committed them (`d7bf3ad99`). Root cause: `STALE_TERMS` contains `"session_log": "agent_session"`, detection is a bare `in` substring test, and application is an unanchored global `str.replace` — so `session_log` matched inside `session_logs` and rewrote `agent/session_logs.py` into `agent/agent_sessions.py`.

This fixes the generator and sweeps the damage.

## Changes

**1. Word-anchored stale terms** (`_detect_stale_term_fixes`) — `STALE_TERMS` keys now match on word boundaries, so no key can fire inside a longer identifier or path segment. Detection and application share one matching semantics.

Anchored replacements travel on a **separate homogeneous channel**: `_apply_fixes_to_file(path, repo_root, fixes, regex_fixes=None)` where `regex_fixes: list[tuple[re.Pattern[str], str]]` is applied in its own `pattern.subn()` loop. The existing `fixes: list[tuple[str, str]]` parameter keeps its type and its `new == ""` line-delete sentinel semantics **unchanged**, so the three rename detectors and the line-delete path are untouched.

**2. The existence invariant** (`_apply_fixes_to_file`) — a fix-class-agnostic post-condition on every write: the auditor may not introduce a `dir/file.{py,md}`-shaped reference that is absent under `repo_root`. Rejection is attributed **per fix**, so valid sibling fixes in the same file still apply, and references already present in the document are never re-validated. Rejections are logged at warning level naming the offending path.

This also independently covers three latent defects #2711 did not name: `renames[0][1]` on a rename-then-delete chain, `_detect_renamed_link_fixes` emitting a repo-root-relative replacement for a doc-relative link, and any future `STALE_TERMS` value forming a plausible-but-absent path.

**3. Withheld fixes are a first-class part of the result contract** — `_ok_result` gains `fixes_withheld` (int) and `withheld` (list of `{"doc","old","new","reason"}`). `status` stays `"ok"` (a withheld fix is not an error), which is exactly why the count must be separately checkable. `.claude/skill-context/do-docs.md` now branches **behaviorally** on `fixes_withheld > 0` rather than treating `status: "ok"` as "output verified".

**4. Damage swept** (grep-based, not from an enumerated list):
[...]

PR #2790, opened 2026-08-13
verdict finalize cluster: honest refusals, a clean verdict field, documented count flags, and a router row that owns the stale-verdict state

### LINKED ISSUE #2740: verdict finalize's backfill refusal claims "State NOT persisted" after it has already persisted the verdict and trailer
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/verdict-finalize-cluster.md

> **Context.** The SDLC pipeline records per-issue state in an issue-keyed ledger: a `stages` map of stage markers (`in_progress` / `completed`) plus a `_verdicts` map and a `_meta` summary. `sdlc-tool verdict finalize` is the composite command `/do-pr-review` runs at the end of a REVIEW: it resolves the PR head SHA, records the REVIEW verdict with a `HEAD SHA=` trailer, and then writes the `REVIEW: completed` stage marker. Guards G4 (oscillation) and G8 (stage-artifact) and `tools/merge_predicate.py` all read this ledger to decide what to dispatch next.

## Problem

When `verdict finalize` is refused by the predecessor-backfill gate, it prints:

```
[ERROR] STATE_MACHINE_REJECTED: predecessor backfill for REVIEW refused for issue #2711: ... State NOT persisted.
```

**"State NOT persisted" is false.** The verdict, its head-SHA trailer, and `_meta.latest_review_verdict` were all durably written before the refusal. Only the stage marker is missing.

The message is locally true and globally false: nothing was persisted *by `write_marker`*, but `finalize` is a composite that already committed a write in an earlier step. The string is emitted from `write_marker`, which cannot see that.

### Current behavior

Verified 2026-08-13T03:28:55Z on issue #2711 / PR #2728. `sdlc-tool verdict finalize --pr 2728 --issue-number 2711 --verdict APPROVED` was refused as above. `sdlc-tool stage-query --issue-number 2711` immediately after:

```json
{"stages": {"_verdicts": {"REVIEW": {"verdict": "APPROVED REVIEW CONTEXT HEAD SHA=90E8B6651C3D6A2789990333A6E843D7CE1BB21D",
  "recorded_at": "2026-08-13T03:28:55.851486+00:00", "blockers": 0, "tech_debt": 0}}},
 "_meta": {"latest_review_verdict": "APPROVED REVIEW CONTEXT HEAD SHA=90E8B6651...", "pr_number": 2728, ...}}
```

The `stages` object carries no markers at all — the recorded verdict and the missing marker coexist, exactly as the error said they would not.

### Desired outcome

The refusal reports what actually landed. `write_marker` already has the right vocabulary for this one line down: the `STATE_MACHINE_RAISED` path says *"Persisted state is INDETERMINATE; re-read with `sdlc-tool stage-query`"* ([`tools/sdlc_stage_marker.py:733`](https://github.com/tomcounsell/ai/blob/main/tools/sdlc_stage_marker.py#L733)). The backfill refusal needs an equivalent, accurate claim.

## What this issue is NOT

**Not an atomicity bug.** The verdict-before-marker ordering is deliberate and documented: *"A `finalize` that refuses on that gate still records the verdict + trailer first"* ([`tools/sdlc_verdict.py:667`](https://github.com/tomcounsell/ai/blob/main/tools/sdlc_verdict.py#L667)). Issue #2577 added the `_review_artifact_readable` conjunct to `verdict_invariant_satisfied` precisely so read sites stay safe in this partial state. Do not "fix" this by making `finalize` transactional — that would undo a deliberate design and is explicitly out of scope.

**Not a regression of #1642 or #2193.** #1642 is `pr_merge_state` resolving UNKNOWN without `GH_REPO`. #2193 is verdict/marker *failing* to persist — the opposite failure. Neither covers a false persistence claim.

## Why it matters

The message is consumed by agents, not just humans, and it directs the reader's next action. In the observed instance a reviewing agent read "State NOT persisted", concluded `finalize` had regressed its atomicity guarantee, and was about to file an issue against #1642/#2193 before the state was re-read directly. A wrong error message on a refusal path costs an investigation every time it fires, and risks an agent re-running a write that already landed.

## Definitions

| Term | Meaning |
|---|---|
| Stage marker | Per-stage `in_progress`/`completed` entry in the issue-keyed ledger's `stages` map |
| Verdict | Recorded REVIEW/CRITIQUE outcome in `_verdicts`, with a `HEAD SHA=` trailer when APPROVED |
| Predecessor backfill | `PipelineStateMachine._backfill_predecessors` (#2399), promotes the ISSUE-rooted spine so a later stage can complete |
| Verdict invariant | #2415: a `completed` REVIEW/CRITIQUE marker implies a readable verdict. Wins over the backfill on conflict (#2554) |

## Solution sketch

The refusal at [`tools/sdlc_stage_marker.py:707-719`](https://github.com/tomcounsell/ai/blob/main/tools/sdlc_stage_marker.py#L707-L719) should stop asserting a fact it cannot know. Open questions for the planner:

- Should the message simply drop the claim and point at `stage-query` (cheapest, matches the `STATE_MACHINE_RAISED` precedent), or should it state precisely which writes landed?
- `write_marker` has no visibility into whether its caller is `finalize` (verdict already written) or a bare `stage-marker` call (nothing written). Is the honest message therefore always "re-read to confirm", or should `finalize` catch `STATE_MACHINE_REJECTED` and re-emit its own accurate message with the composite's knowledge?
- The second option is likely better and slightly larger. It also gives `finalize` somewhere to state the remedy in terms the caller can act on: run `/do-plan-critique`, then re-run the identical `finalize` (idempotent — the verdict write repeats harmlessly).
- Audit sibling refusal paths in the same function (`REVIEW_VERDICT_MISSING`, `REVIEW_TRAILER_MISSING`, `REVIEW_ARTIFACT_MISSING`, `CRITIQUE_VERDICT_MISSING`, `ISSUE_LOCKED`) for the same class of overclaim before fixing only this one.

## Acceptance criteria

- [ ] A backfill-refused `verdict finalize` no longer claims "State NOT persisted" when a verdict was written.
- [ ] The emitted message lets a reader determine what landed without a separate `stage-query`, or explicitly tells them to run one.
- [ ] Sibling refusal paths in `write_marker` audited for the same overclaim; findings either fixed or recorded.
- [ ] A test asserts the refusal message against a ledger that has a recorded verdict and no marker.
- [ ] The deliberate verdict-first ordering is unchanged — `tools/sdlc_verdict.py:667`'s documented contract still holds.

## Test impact

`tests/unit/test_sdlc_stage_marker.py` already asserts `reason == "STATE_MACHINE_REJECTED"` (`:252-254`) and will need the message assertion added. No existing test covers the refusal's persistence claim.

## Downstream

Related: #2730 (stage markers and dispatch history diverge — a different divergence in the same ledger), #2577, #2554, #2415, #2399.

## Recon Summary

Recon performed 2026-08-13 at `main@90c0e81e4f0602f33f8a577f246390a5afbeffd5` for the three-issue cluster (#2740 anchor, #2769, #2767) planned as one lane (`verdict-finalize-cluster`).

**Confirmed** (read directly in current code)

- **#2740 — the false claim.** `tools/sdlc_stage_marker.py:717-719` emits `STATE_MACHINE_REJECTED: predecessor backfill for {stage} refused ... State NOT persisted.` from inside `write_marker`. `tools/sdlc_review_finalize.py:486-511` calls `record_verdict` FIRST and only then `write_marker`, so on this branch the verdict + trailer + `_meta.latest_review_verdict` are already durable when the string prints. `write_marker` has no parameter, kwarg, or ambient signal telling it a verdict was written — the honest message cannot be composed inside `write_marker` alone.
- **Sibling overclaims.** The identical `State NOT persisted.` sentence appears at `tools/sdlc_stage_marker.py:566-567` (`start_stage` refusal) and `:603-604` (`STAGE_RAN_NOT_SKIPPABLE`). The five named-gate refusals audited per the issue — `REVIEW_VERDICT_MISSING` (:625-631), `REVIEW_TRAILER_MISSING` (:640-647), `REVIEW_ARTIFACT_MISSING` (:658-665), `CRITIQUE_VERDICT_MISSING` (:668-681), `ISSUE_LOCKED` (:686-690) — say `Marker write refused.` / `marker write refused.`, which is scoped to the marker and therefore accurate. Only the three `State NOT persisted.` sites overclaim. The accurate precedent is `STATE_MACHINE_RAISED` at `:733` (`Persisted state is INDETERMINATE; re-read with sdlc-tool stage-query`).
- **#2769 — trailer absorbed into the verdict field.** `tools/sdlc_review_finalize.py:466-470` builds `trailered_verdict = f"{verdict.strip()} REVIEW_CONTEXT head_sha={head_sha}"` and passes that single string as `verdict=` to `record_verdict`. `tools/sdlc_verdict.py:428-431` then runs the WHOLE string through `normalize_verdict`, producing the stored `APPROVED REVIEW CONTEXT HEAD SHA=<HEX>`. `_verdict_is_recognized` (`sdlc_review_finalize.py:105-110`) is a substring test, so the composite passes validation. `record_verdict` (`tools/sdlc_verdict.py:359+`) has no trailer/head-sha parameter — the record dict it writes is `{verdict, recorded_at, artifact_hash[, blockers, tech_debt, _judges, _consensus]}`, so a separate field is a genuine schema addition, not a rename.
- **#2769 — why it is currently benign.** Every reader substring-matches or regex-extracts: `agent/sdlc_router.py` via `normalize_verdict` + `in`, `tools/_sdlc_utils.py:36-38` `_HEAD_SHA_TRAILER_RE` which deliberately matches BOTH `REVIEW_CONTEXT head_sha=` and its normalized image `REVIEW CONTEXT HEAD SHA=`, and `tools/merge_predicate.py`. Any equality/enum/display consumer breaks. This constrains the fix: the read path must keep accepting the legacy mangled string (migration-free), because live ledgers already contain it.
- **#2767(a) — count flags are undocumented ints.** `tools/sdlc_verdict.py:938-939`: `fin.add_argument("--blockers", type=int, default=None)` and `--tech-debt` likewise — both with **no `help=` text at all**, unlike the neighbouring `--verdict` and `--run-id`. The global skill `~/.claude/skills/do-pr-review/SKILL.md` §5 never states the type. The repo addendum `docs/sdlc/do-pr-review.md:93-98` does model integer usage (`--blockers 0`, `--blockers $BLOCKERS`), which is why this repo has not hit the failure and popoto did: the addendum is the only place the shape is implied, and it is repo-local.
- **#2767(b) — no branch conflict.** `session/sdlc-router-rereview-crash-and-row3-pr-guard` is **merged** (PR #1941, `9e2f2b5cd`, issue #1932); it is not an active competing branch. Rows 8/8b/8c/8d in `agent/sdlc_router.py` are at their post-#1941 shape.

**Revised**

- **#2740 is not solely a message-wording change.** Because `write_marker` structurally cannot know what its caller already persisted, the accurate message must be composed by `finalize` (which does know) — i.e. the fix touches `tools/sdlc_review_finalize.py`'s handling of a non-zero `write_marker` return (`:506-511`, currently `raise ReviewFinalizeError(f"{reason}: ... {marker_result}")`), not only the string at `sdlc_stage_marker.py:717-719`. The `write_marker` side still needs its overclaim removed for bare `stage-marker` callers.
- **#2740 and #2769 share one call site and must ship together.** Both defects live in `finalize`; #2769's fix changes what is passed to `record_verdict` on exactly the code path #2740's fix re-messages. Splitting them would guarantee a merge conflict.
- **#2767(b) root cause is NOT yet identified.** The reported terminal state (`PATCH: completed`, `REVIEW: pending`, `patch_cycle_count: 1`, verdict CHANGES REQUESTED recorded) is one that row 8 `_rule_review_has_findings` should match: the verdict is present, `_review_verdict_is_stale` (`agent/sdlc_router.py:956-981`) returns False because the verdict (05:56:28) postdates the last `/do-patch` dispatch (05:43:09), and `CHANGES REQUESTED` normalizes into the token. Row 7 `_rule_pr_exists_no_review` should match the no-verdict variant. **Neither hypothesis explains the observed miss from the issue's summary alone.** BUILD must reconstruct the exact `stage_states`/`_meta` payload from popoto run `f312ee30a5cd401daa6c06f2ce78aa60` and reproduce in a unit test BEFORE changing any predicate — a speculative predicate widening here risks re-opening the oscillation classes #1641/#1668/#1932 closed.
- **`guard_id: null` on a no-rule block is separately actionable.** `agent/sdlc_router.py:1672` returns `reason="no matching dispatch rule"` with no guard id, which a supervisor cannot distinguish from a real guard block. This is fixable independently of the row-table root cause and is worth doing regardless of what the reproduction finds.

**Pre-requisites**

- Verdict-first ordering in `finalize` is **deliberate and out of scope** (`tools/sdlc_verdict.py:667`, #2577/#2415/#2554). No transactional rewrite.
- Any new verdict-record field must remain read-compatible with ledgers already holding the mangled string. No migration (per #2769's own fix shape).
- `docs/sdlc/do-pr-review.md:103` currently asserts finalize "is **atomic and self-verifying**" — false on the backfill-refusal branch. Doc correction belongs in this lane.

**Dropped** / deferred

- **#2767 §3** (repos with no `docs/sdlc/` substrate: `/do-sdlc` step 3d.4 halts with `REVIEW_VERDICT_MISSING` where the generic contract says "the posted GitHub review IS the verdict"). Real, but it is a `/do-sdlc` supervisor-contract change touching a different file set from all three defects above. Split to a follow-up issue rather than widening this lane.
- Fixing #2735 (`find_plan_path` cwd-dependence), which per the 06:05 comment is the dominant *trigger* for the backfill-refusal branch. Making the trigger rarer is not making the message honest; #2735 stays its own issue.

**Test impact confirmed**

`tests/unit/test_sdlc_stage_marker.py:237-239` asserts `reason == "STATE_MACHINE_REJECTED"` and that the taxon appears in stderr — compatible with a message change, but no test asserts the persistence claim. `tests/integration/test_off_pipeline_merge_path.py` asserts the same `reason` at seven sites (`:323, :368, :446, :576, :636, :663, :689`) — all reason-only, so none break on a message rewording. `tests/unit/test_sdlc_dispatch.py:434` substring-matches the taxon only.

### LINKED ISSUE #2767: verdict finalize: --blockers/--tech-debt take integer counts, but nothing tells the review skill that; router falls back to no matching dispatch rule after a second review→patch cycle
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/verdict-finalize-cluster.md

Two related findings from six concurrent `/do-sdlc` runs in `tomcounsell/popoto` (2026-08-13).

## 1. `verdict finalize` count flags are silently prose-hostile

`--blockers` / `--tech-debt` are `int`. Passing findings text fails:

```
argument --blockers: invalid int value: "1) mkdocs build --strict fails on-branch: ..."
```

Nothing in `~/.claude/skills/do-pr-review/SKILL.md` §5 states these are counts rather than descriptions, and the natural reading of "record the verdict" is that findings go in. A `/do-pr-review` run that tries prose gets a non-zero exit on a call the skill calls "mandatory and terminal", which per §5 means STOP — so the review is complete on GitHub but absent from the ledger.

Observed directly: the popoto #537 pipeline skipped the call on its first substantive review, hit `REVIEW_VERDICT_MISSING`, and recovered only by re-calling with integers.

**Suggested:** accept prose and derive counts, or rename to `--blocker-count`/`--tech-debt-count` and say so in the skill.

## 2. Router returns `no matching dispatch rule` after a second review→patch cycle

popoto #537 / PR #548, `run_id f312ee30a5cd401daa6c06f2ce78aa60`:

| # | skill | at (UTC) | result |
|---|---|---|---|
| 1 | /do-pr-review | 05:26:27 | BLOCKED_ON_CONFLICT |
| 2 | /do-patch | 05:28:06 | success (merged main) |
| 3 | /do-pr-review | 05:31:01 | CHANGES REQUESTED (1 blocker, 2 tech debt) |
| 4 | /do-patch | 05:43:09 | success |
| 5 | /do-pr-review | 05:56:28 | CHANGES REQUESTED (1 blocker, 1 tech debt) |

Then: `blocked` / `no matching dispatch rule`, `guard_id: null`. State at that point: `PATCH: completed`, `REVIEW: pending`, `patch_cycle_count: 1`, `pr_merge_state: CLEAN`, `ci_all_passing: true`.

Rows 8/8b matched this shape after the *second* review and stopped matching after the *third*. A second review→patch cycle is ordinary on a non-trivial PR; the loop should not fall off the rule table. A `blocked` with `guard_id: null` is also indistinguishable from a real guard block to a supervisor following the loop contract.

## 3. Related design gap (context, not a bug in this repo)

Five popoto pipelines halted at `do-sdlc` step 3d.4 with `REVIEW_VERDICT_MISSING` because popoto has **no `docs/sdlc/` directory** — so no verdict substrate is declared, and `/do-pr-review` §5's documented generic-case behavior ("the posted GitHub review IS the verdict — skip this step") is correct. But `do-sdlc`'s 3d.4 gate assumes a substrate unconditionally and halts loudly.

Either 3d.4 should no-op when no substrate is declared, or `do-sdlc`'s Repo Context Probe should refuse to start a supervised run in a repo with no context files, rather than running eight stages and halting at the seventh. (Fixing popoto's side separately by adding the context files.)

### LINKED ISSUE #2769: verdict finalize stores the head-SHA trailer inside the verdict field
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/verdict-finalize-cluster.md

`sdlc-tool verdict finalize` appends the `REVIEW_CONTEXT head_sha=` trailer into the stored verdict string itself, so `_verdicts["REVIEW"].verdict` reads `APPROVED REVIEW CONTEXT HEAD SHA=25E53671…` (uppercased, trailer absorbed) where `APPROVED` belongs.

Reproduced on every finalize tonight (2026-08-13 fan-out): PRs #2680, #2700, #2706, #2668 all store the mangled form, on both clean runs (exit 0, all three writes) and backfill-refusal runs. Root cause per observation: `_verdict_is_recognized` accepts any string merely containing a known token, so the composed verdict+trailer string passes validation and is stored whole.

Safe today only by coincidence: `agent/sdlc_router.py` substring-matches (`REVIEW_APPROVED not in normalize_verdict(...)`) and the freshness gates parse the trailer back out of the stored string. Any consumer doing verdict equality, enum parsing, or raw display breaks.

Distinct from but sharing a call site with #2740 (false "State NOT persisted" message / non-atomicity on the marker-refusal branch) — see the two-defects analysis at https://github.com/tomcounsell/ai/issues/2740#issuecomment-5277265665. Fix shape: store the bare verdict token in the verdict field and the trailer in its own field (or only in the posted comment), with a migration-free read path that still parses legacy mangled strings.

valorengels:
Closes #2740
Closes #2769
Closes #2767

Plan: `docs/plans/verdict-finalize-cluster.md` (round-4 READY TO BUILD, no concerns).

Four defects across five code files. #2740 and #2769 are two defects at the *same call site* (`tools/sdlc_review_finalize.py::finalize`), so splitting them would guarantee a merge conflict there. #2767 rides along because its two halves are cause and effect of each other: (a) makes a finalize call fail, which leaves the stale verdict that (b) leaves unowned.

## #2769 — the head SHA had been absorbed into the verdict token

`finalize` composed `f"{verdict} REVIEW_CONTEXT head_sha={sha}"` into a single string and handed it to `record_verdict`, which runs `normalize_verdict` over the **whole** string. `_verdicts["REVIEW"].verdict` therefore read `APPROVED REVIEW CONTEXT HEAD SHA=<HEX>` where a bare `APPROVED` belonged. Reproduced on every finalize during the 2026-08-13 fan-out (PRs #2680, #2700, #2706, #2668), on clean exit-0 runs as well as refusal runs. It passed validation only because `_verdict_is_recognized` is a substring test.

- `record_verdict` gains an optional `head_sha` kwarg, attached only when truthy (the `_judges`/`_consensus` precedent) and inside the **same** `update_stage_states` call, so the single-writer invariant holds. Stored verbatim — it is a 40-hex commit SHA and must not pass through `normalize_verdict`.
- `finalize` passes the bare token plus `head_sha=`. An already-trailered input has its SHA lifted into the field and stripped from the token, so it is stored exactly once and never in both places.
- New `head_sha_of_record` / `head_sha_of_text` in `tools/_sdlc_utils.py` — two entry points over one body, because the readers do not all hold the same type. Precedence is defined and tested: **the field wins** over a disagreeing legacy in-token trailer.
- Every head-SHA reader is routed through them: `_review_trailer_present`, `check_review_persistence`, `tools/merge_predicate.py`, and the router's row 8f freshness gate.

**The regex fallback is permanent, not scaffolding.** Live ledgers hold mangled verdict strings forever and there is deliberately no migration. Tests cover a legacy-shape record passing every gate.

Two non-obvious pieces the critique rounds forced:

- **`tools/sdlc_stage_query.py` had to join the fence.** `_extract_verdict_text` flattens a `_verdicts[stage]` record to its `verdict` string and drops every sibling key, so row 8f could not see a `head_sha` *field*. With row 8f's fail-closed default (no resolvable head SHA → stale), that would have declared **every** approved verdict head-stale and re-dispatched `/do-pr-review` on every lane until G4 escalated — a router-wide re-review loop, not a quiet degradation. Fixed with `_meta.latest_review_head_sha`.
- **That extractor's guard is three-way, not two.** `dict` → `head_sha_of_record`, `str` → `head_sha_of_text`, everything else → `None`, with the `str` arm matched **positively**. The negated `not isinstance(rec, dict)` form routes the common `None` (every issue with no REVIEW verdict yet) into the trailer regex, raising `TypeError`, which `sdlc_next_skill._resolve_enriched`'s broad `except` swallows into an empty ledger — making a fully-worked issue read as brand new and routing it back to `/do-plan`. All four record shapes are tested at both the `_compute_meta` and `decide_next_dispatch` levels.

`merge_predicate`'s lazy in-function import is preserved, and asserted by a test: its module-level imports must stay stdlib-only or the merge-guard hook breaks under a bare interpreter. Its tests assert the **exact `notes` string**, not pass/fail — the trailer-less branch silently downgrades to a weaker `recorded_at`-vs-commit-date comparison and *both* branches report "fresh", so a pass/fail assertion cannot detect that fail-open on the merge gate #2404/#2415 exist to keep honest.

`_verdict_is_recognized` stays a substring test. Tightening it to equality would re-open #2548 (`"APPROVED (0 BLOCKERS)"`), and does nothing about already-stored strings. The mangling is fixed at the write site, where it was introduced.

## #2740 — the refusal lied about what landed

`finalize` is a composite: `record_verdict` commits, **then** `write_marker` runs. On a backfill refusal the verdict, its head SHA, and `_meta.latest_review_verdict` were all already durable when `write_marker` printed `State NOT persisted.` `write_marker` was locally correct — *it* wrote nothing — but it cannot see what its caller already committed.

The cost is measured, not theoretical: on issue #2711 / PR #2728 a reviewing agent read that sentence, concluded `finalize` had regressed its atomicity guarantee, and was about to file a duplicate issue against #1642/#2193 before someone re-read the ledger directly.

- `finalize`'s marker-refusal branch — the only layer that knows both facts — now states the split: the verdict and head SHA **are** persisted, only the marker was not, and re-running the identical command is idempotent. The named reason stays the prefix (tooling and the `/do-sdlc` supervisor match on it) and there is exactly one remedy, not an inventory of fields.
- All **four** sites are reworded to the `STATE_MACHINE_RAISED` precedent (name the refusal, point at `sdlc-tool stage-query`): three inside `write_marker` plus `main()`'s non-zero-exit wrapper. That fourth site is currently unreachable — every error key `write_marker` returns is in `_DIAGNOSED_ERRORS` — but it is reworded rather than deleted, because it is what a bare `sdlc-tool stage-marker` caller sees the moment an undiagnosed error key is added.

**Sibling-refusal audit result (AC3):** the five named gates (`REVIEW_VERDICT_MISSING`, `REVIEW_TRAILER_MISSING`, `REVIEW_ARTIFACT_MISSING`, `CRITIQUE_VERDICT_MISSING`, `ISSUE_LOCKED`) are **already accurate** — each ends with `Marker write refused.`, correctly scoped. They are untouched. The overclaim was one sentence at four sites, all fixed.

No write is reordered. Verdict-first is the deliberate #2415/#2577 design and was explicitly out of scope; `finalize` is **not** made transactional.

## #2767(a) — prose into an int flag, on a terminal call

`--blockers` and `--tech-debt` were bare `type=int` with **no `help=` text at all**, on a call `/do-pr-review` §5 calls "mandatory and terminal" with instructions to STOP on non-zero exit. An agent taking the natural reading ("record the verdict" ⇒ findings go in) gets `argument --blockers: invalid int value: "1) mkdocs build --strict fails..."` and a hard stop — review complete on GitHub, absent from the ledger. Observed on the popoto #537 pipeline.

Renamed to `--blocker-count` / `--tech-debt-count` with help text stating they are counts and that findings belong in the posted review. Teaching `finalize` to derive counts from prose was rejected: a silently-wrong count corrupts the ledger, which is strictly worse than a loud failure.

Old spellings kept as argparse aliases on the same argument — one flag, two spellings, not a second code path. They cover the cross-machine propagation window, since the global SKILL.md reaches other machines only on their next `/update`. That window closes, so removal is tracked in #2789. `default=None` vs `0` survives the rename and is tested: omitted means "not assessed", `0` means "assessed, none found".

The parser is extracted to `_build_parser()` so spellings and help text are testable without executing a state-mutating subcommand.

The **global** `do-pr-review` SKILL.md §5 now states the integer contract — that is what a foreign repo reads, and this repo only avoided the bug because of a repo-local addendum invisible elsewhere. That addendum's false "atomic and self-verifying" claim is corrected to describe what actually holds on the refusal branch.

## #2767(b) — the router fell off its own rule table

After `review → patch → review → patch → review`, the router returned `Blocked('no matching dispatch rule', guard_id=None)`. The unowned state is a **stale** REVIEW verdict with `last_dispatched_skill == /do-pr-review`: row 7 needs an absent verdict, row 8 steps aside precisely on staleness, row 8b required `last == /do-patch`, and rows 8c/8d/8e all need an absent verdict.

Row 8b (`_rule_patch_applied_after_review`) now matches on `last == /do-patch` **or** `_review_verdict_is_stale(...)`, keeping its `pr_number` and `PATCH == completed` conditions.

Disjointness is proved rather than asserted, which matters because **row 8c does not merely sit beside row 8b — it calls it** as a step-aside: `_review_verdict_is_stale` returns False whenever `recorded_at` is absent, and `recorded_at` is absent exactly when no verdict was recorded (`record_verdict` writes `verdict` and `recorded_at` together or not at all). Rows 8c/8d/8e all require the absence of a recorded verdict, so the new disjunct is identically False on every state they own. No defensive step-asides were added. G4 still bounds the resulting re-review loop — the #1641/#1668 re-opening risk — and that is tested, not asserted.

The no-rule fallthrough now carries `guard_id="NO_RULE"` so a supervisor can distinguish "no rule owned this state" from a real guard block. Short-code form matches the existing `G2`/`G4`/`G7` convention. This slightly widens the `Blocked` contract — `guard_id` no longer implies "a numbered guard fired" — which is stated in its docstring.

## Verification

All rows of the plan's Verification table pass:

| Check | Result |
|---|---|
| `grep -rn "State NOT persisted" tools/ agent/` | exit 1 (zero matches) |
| `latest_review_head_sha` in `sdlc_stage_query.py` / `sdlc_router.py` | present in both |
| `merge_predicate` `_sdlc_utils` import stays in-function | line 580, indented |
| `finalize --help` contains `blocker-count` and `--blockers` | both |
| Global SKILL.md documents `blocker-count` | 3 occurrences |
| No ledger migration added | exit 1 |
| `_verdict_is_recognized` stays a substring test | intact |
| `ruff check` / `ruff format --check` | clean |

**647 tests pass** across `test_sdlc_router{,_oscillation,_decision}.py`, `test_sdlc_stage_marker.py`, `test_sdlc_dispatch.py`, `test_merge_predicate.py`, `test_sdlc_verdict.py`, `test_sdlc_review_finalize.py`, `test_sdlc_stage_query.py`, `test_sdlc_utils_trailer_regex.py`, `test_sdlc_skill_md_parity.py`, and the `test_sdlc_review_finalize_roundtrip.py` / `test_off_pipeline_merge_path.py` integration suites.

Two pre-existing `ruff` N802 errors in `tests/db_derivation_guard.py` are untouched by this lane and present on `main`.

## Follow-ups filed

- #2787 — a "READY TO BUILD (with concerns)" critique verdict does not mandate a re-critique, so the revision closing the concerns is never itself reviewed. This plan is the evidence: round 3's concern was a defect *in round 2's own prescription*, and it was worse than the blocker round 2 was written to prevent.
- #2789 — drop the `--blockers` / `--tech-debt` aliases once every machine has run `/update`.


valorengels:
## Review: APPROVED

Self-authored PR, so this is posted as a comment rather than a formal `gh pr review` approval. Reviewed by a `code-reviewer` subagent against `docs/plans/verdict-finalize-cluster.md`, with every claim verified against real source in the worktree rather than against the commit messages or PR body.

### Outcome

**0 blockers, 5 concerns, 5 nits.** All 5 concerns and 4 of the 5 nits are fixed in `3fd8be1`. The remaining nit is declined with a reason below.

### What the review verified rather than assumed

- `grep -rn "State NOT persisted" tools/ agent/` returns nothing — all four sites, including `main()`'s wrapper.
- `_extract_head_sha` is genuinely three-way with the `str` arm matched positively, not as `not isinstance(rec, dict)`.
- `head_sha` is attached inside the same record before a single `update_stage_states` call, with a `call_count == 1` test — the Race 3 single-writer invariant.
- `merge_predicate`'s `_sdlc_utils` import is inside `_check_verdict_freshness`, pinned by an `inspect.getsource` test rather than a comment. Its freshness tests assert the exact `notes` strings, so the weak `recorded_at` fallback cannot silently take over — the fail-open both branches would report as "fresh".
- `_verdict_is_recognized` is still a substring test (#2548's `"APPROVED (0 BLOCKERS)"` case survives).
- Reader enumeration holds: no `ui/` code reads the REVIEW verdict, and the only remaining raw regex outside the helper is the router's deliberate import-boundary duplicate.
- No ledger migration was added.

### Concerns fixed

1. **A malformed `head_sha` field opened the REVIEW marker gate.** The strongest finding. Routing `_review_trailer_present` through `head_sha_of_record` dropped the 40-hex requirement the regex had enforced, so a programmatic `record_verdict(head_sha="not-a-sha")` — a public, documented kwarg — satisfied a gate whose own refusal message promises "well-formed". Downstream stayed fail-closed, so this was a gate relocation rather than a fail-open, but the contract the message promised was no longer the contract the code enforced. `head_sha_of_record` now applies the same shape check to the field, treating a malformed one as absent so the permanent legacy path still resolves.

2. **An unpinned ownership transfer.** Row 8b precedes 8f/9/10, so the widening moved stale-APPROVED away from row 9: it is now re-reviewed rather than advanced to `/do-docs` or the merge fast-path. Correct — an approval predating the latest patch does not describe the code that patch produced — but spike-2's "rows 8 and 9 unchanged" holds only for *fresh* verdicts, and nothing pinned the difference. Now covered by `TestStaleApprovedIsReReviewedNotAdvanced`, including its convergence, and stated in the row 8b docstring and the feature doc.

3-5. **Stale claims that still contradicted the shipped code**: `finalize`'s own docstring ("cannot partially complete", trailer "left untouched"); `docs/sdlc/do-pr-review.md` carrying the corrected claim at :103 and the contradictory one at :184; and the global `do-pr-review` SKILL.md — the body every *other* machine reads — keeping "ONE atomic operation" and "verdict + trailer + completion marker". That file also documented `--blockers`/`--tech-debt` as "older spellings", which undercuts the plan's own "not a parallel migration" justification and follow-up #2789; removed.

### Nit declined

`check_review_persistence`'s fail-closed catch reports `REVIEW_VERDICT_MISSING` even when `verdict_present` was already true. Genuinely misleading and on-theme for this lane, but pre-existing and other tooling matches on that taxon. Renaming it belongs in its own change.

### Tests

**652 passing** across the router, marker, dispatch, merge-predicate, verdict, finalize, stage-query, trailer-regex, and skill-parity unit suites plus the finalize-roundtrip and off-pipeline-merge integration suites. `ruff check` and `ruff format --check` clean. Two pre-existing `N802` errors in `tests/db_derivation_guard.py` are untouched by this lane and present on `main`.

The new tests are demonstrated-red, not tautological: the malformed-field assertion returned `"not-a-sha"` under the previous code, and the stale-APPROVED assertion resolved to row 9 before the widening.

Issue #2026, opened 2026-07-11
Reliability risk: /do-sdlc fork vs supervisor — forked executions merge past blocked gates, record unearned/contradictory verdicts, and lease TTL churn deadlocks the router


tomcounsell:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/sdlc-fork-supervision-hardening.md

## Investigation Finding

**Component:** forked /do-sdlc stage executions vs. supervisor gates (issue lease / stage ledger)
**Detected:** 2026-07-11 (live supervision of the #2013 pipeline run; corroborated by the #2016 run the same morning)
**Status:** Persistent (multiple distinct races within one pipeline run; pattern matches prior closed #1915 but with new failure modes)

## Symptoms

During the #2013 run (PR #2021), forked stage executions repeatedly did real work ahead of, or in contradiction to, the supervisor's gate decisions:

1. A forked sub-execution marked REVIEW completed **without reviewing**; the supervisor's merge gate correctly blocked on "no recorded REVIEW verdict".
2. The review fork then double-posted **contradictory verdicts** (CHANGES REQUESTED, then APPROVED) against the same head; the APPROVED comment's docs rubric contradicted the actual branch content and had to be manually reconciled back to CHANGES REQUESTED.
3. While the supervisor's gate was still blocked, a **parallel forked execution merged the PR** (08:11:26Z) and migrated the plan. The supervisor's patch commit (`eb36213d`) never merged as a commit; its content had to land via a direct-to-main follow-up (`120e3def`).
4. The ~300s issue-lock TTL **expired during every long stage**, and stage skills repeatedly finalized the `sdlc-local-2013` tracking session to terminal states mid-pipeline, deadlocking the router with `ISSUE_LOCKED, orphaned_lock`. The supervisor recovered via `transition_status(..., reject_from_terminal=False)` + `session-ensure --reuse-run-id`; run identity rotated twice (`6de14623` → `753fbb94` → `2b216509`).

The #2016 run the same morning hit the lease symptom too: `session-ensure` returned `ISSUE_LOCKED / orphaned_lock: true` from a prior session that completed without releasing its lock (~5 min TTL wait).

Outcome was ultimately correct on main both times, but only because supervisors reconciled ground truth after each race.

## Diagnostic Output

```
From the #2013 supervisor's final report (2026-07-11):
- "MERGE (sonnet) — correctly BLOCKED: gate refused on 'no recorded REVIEW verdict'
[...]

PR #2792, opened 2026-08-13
SDLC lane identity: one recorded slug, minted once (#2735, #2718)

### LINKED ISSUE #2718: G8 stage-artifact verifier probes session/{plan-slug} instead of the PR head ref, wedging any lane whose branch name diverges from its slug
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/sdlc-lane-recorded-slug.md

## Summary

The G8 stage-artifact verifier derives the branch it probes from the **plan slug** instead of the PR's live `headRefName`. Any lane whose branch name diverges from its plan slug fails verification against a branch that never existed, and the router force-dispatches a useless `/do-patch` — repeatedly, until the G4 oscillation cap hard-wedges the lane.

## Observed vs expected

- **Observed:** `sdlc-tool next-skill --issue-number 2663` returns

  ```
  stage-artifact-verify: issue #2663 PATCH claims completed but branch session/session-liveness-tick-counter is not pushed
  {"skill": "/do-patch", "reason": "G8: PATCH claims completed but its artifact failed live verification", "row_id": "G8", "dispatched": true}
  ```

- **Expected:** the PATCH artifact verifies clean (it is real and pushed), G8 does not fire, and the router routes `/do-pr-review` — the recorded `BLOCKED ON CONFLICT` verdict is stale against the now-mergeable head.

## Concrete instance: lane sdlc-2663 / PR #2668

- Plan doc: `docs/plans/session-liveness-tick-counter.md` → slug `session-liveness-tick-counter`.
- Probed branch: `session/session-liveness-tick-counter` — has never existed locally or on origin.
- Real branch: `session/sdlc-2663`, pushed and matching the PR head.

```
$ git ls-remote origin refs/heads/session/sdlc-2663 refs/heads/session/session-liveness-tick-counter
69039fae5525130ee6015ad53a687a8ac3339fa2	refs/heads/session/sdlc-2663
```

Only the real branch resolves; the slug-derived name returns nothing, so `_check_branch_pushed` returns `False` and G8 fires. PR #2668 is `state: OPEN`, `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`, head `69039fae5525130ee6015ad53a687a8ac3339fa2`. The worktree `.worktrees/sdlc-2663` is clean with nothing unpushed — there is genuinely nothing for the forced `/do-patch` to do.

## Root cause

`tools/sdlc_next_skill.py:205` computes the slug from the plan path, and `:236` feeds it to a branch check that hardcodes the `session/{slug}` shape:

- `tools/sdlc_next_skill.py:205` — `slug = Path(plan_path).stem`
- `tools/sdlc_next_skill.py:236` — `if pr_state != "MERGED" and not _check_branch_pushed(slug):`
- `tools/sdlc_next_skill.py:151-163` — `_check_branch_pushed(slug)` runs `git ls-remote --heads origin session/{slug}`
- `tools/sdlc_next_skill.py:239` — the warning text that surfaces the wrong branch name

The underlying assumption is the slug-identity rule stated in `.claude/skills/sdlc/SKILL.md` ("each issue's build fork exclusively owns `session/{slug}`"). That holds when the lane is created by `/do-plan {slug}`, but not for a lane branched as `session/sdlc-{issue}`, which is a shape the pipeline produces in practice. The verifier treats a naming convention as ground truth when the PR itself carries the authoritative ref.

Note the failure mode is asymmetric and silent-ish: the check can only ever produce a **false negative** (probe a name that does not exist). It cannot catch the case it was written for when the branch names diverge, so the gate is simultaneously over-firing here and non-functional for its stated purpose on these lanes.

## Impact

Every lane whose branch name differs from its plan slug wedges identically:

1. PATCH completes legitimately and pushes.
2. G8 verifies against a nonexistent `session/{plan-slug}` and fails.
3. The router force-dispatches `/do-patch` against a clean worktree with nothing to patch.
4. `same_stage_dispatch_count` climbs; at 3 the G4 oscillation cap hard-blocks the lane, which then needs a human `dispatch reset`.

Note that G4 running before G8 bounds the loop but does not rescue it — the lane still ends in a hard block requiring manual intervention, and the recorded block names oscillation rather than the real cause, which makes the diagnosis misleading for whoever picks it up.

## Suggested fix

Resolve the branch from the PR head ref rather than from the plan slug. `tools/pr_head_resolver.py::resolve_pr_head_sha` is the established git-first pattern for PR head resolution in this repo (git-first via `git ls-remote refs/pull/N/head`, never a bare `gh` read) — the branch-name resolution should follow the same posture and the same module, so both live-PR reads share one authority.

Sketch of the shape:

- Resolve the PR's `headRefName` for `pr_number` (already in scope at `tools/sdlc_next_skill.py:215` as `meta.get("pr_number")`, and `_fetch_pr_state` already makes a live PR call at `:224` that could carry the ref back in the same request rather than adding a second round trip).
- Probe that exact ref with `git ls-remote --heads origin <headRefName>`, dropping the hardcoded `session/` prefix.
- Keep the slug path only as a fallback for the pre-PR case where `pr_number` is absent — though note that when `pr_number` is absent, `patch_claimed` cannot be meaningfully verified this way at all, so returning a no-op (`{}`) may be more honest than probing a guessed name.
- Fix the warning string at `:239` to name the branch actually probed.

Worth checking as part of the same fix: `tools/sdlc_next_skill.py:357` computes a slug the same way and may carry the same assumption.

## Suggested next steps

1. Add a regression test covering a lane whose branch name diverges from its plan slug (`session/sdlc-{issue}` with a differently-named plan doc), asserting G8 does **not** fire when the PR head ref is pushed. Per #2658's demonstrated-red posture, the test should be shown failing against current `main` before the fix lands.
2. Audit for other `session/{slug}` assumptions in the router and gate code, since this is a replicated-assumption defect rather than a single site — close it out on a clean sweep, not an enumerated list.
3. Unwedge lane sdlc-2663 / PR #2668 once fixed: the correct route there is `/do-pr-review`, since the recorded `BLOCKED ON CONFLICT` verdict (09:15:37Z, `artifact_hash: null`) predates the conflict-resolution commit that landed at 10:13Z.

---

## Recon Summary

Performed 2026-08-13 against `main` @ `0dd8e70f2`. Planned jointly with #2735 (same root cause: no recorded lane slug), and both will close in one PR.

**Confirmed**

- All four cited line references still hold, unmoved: `tools/sdlc_next_skill.py:205` (`slug = Path(plan_path).stem`), `:236` (`if pr_state != "MERGED" and not _check_branch_pushed(slug)`), `:151-163` (`_check_branch_pushed` → `git ls-remote --heads origin session/{slug}`), `:239` (the warning string naming the wrong branch).
- The second site flagged as "worth checking" — `tools/sdlc_next_skill.py:357` — carries the identical assumption: `slug = Path(plan_path).stem` at `:357` feeds `context["branch_exists"] = f"session/{slug}" in branch_names` at `:367`. Same defect, second site.
- The branch divergence is live. `git ls-remote origin refs/heads/session/sdlc-2663 refs/heads/session/session-liveness-tick-counter` returns exactly one ref: `refs/heads/session/sdlc-2663`. PR #2668 is still `state: OPEN`, `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`, `headRefName: session/sdlc-2663`.
- `find_plan_path(2663)` still returns `docs/plans/session-liveness-tick-counter.md` (a plan that tracks #2716), so the wrong slug is still being derived today.
- `tools/pr_head_resolver.py::resolve_pr_head_sha` is already imported by this module (`tools/sdlc_next_skill.py:146`), so the suggested "same module, same posture" direction has an existing seam.

**Revised**

- **The branch name should come from the recorded lane slug, not from the PR head ref.** Resolving `headRefName` per tick is a live-PR read that only works once a PR exists, and G8's PATCH check is not the only consumer of the wrong slug — `branch_exists` (`:367`), `_check_plan_committed_on_main` (`:209`), and `tools/sdlc_stage_query.py:369` (`--head session/{slug}` PR recovery) all derive independently. Fixing this at the PR-read layer leaves three other derivations wrong. The decided direction (see #2735) is to record the slug once on `PipelineLedger` and have all of these read it.
- **The "fabricated shape" premise the code rests on is factually false.** `tools/sdlc_next_skill.py:346-349` and `tools/sdlc_stage_query.py:347-350` both assert this repo never creates `session/sdlc-{N}`. This checkout holds **111 local / 99 remote** such branches, minted by `reflections/sdlc_upvote_lanes.py:489`, `reflections/sdlc_progress.py:707`, and `tools/valor_session.py:107`. `tests/unit/test_sdlc_next_skill.py:117-176` (`TestBranchExistsCanonicalShape`) pins that false belief in three test cases and must be corrected.
- **Falling back to a guessed name when no slug resolves is the wrong default.** The issue's own note is adopted verbatim: when no slug can be resolved, the check NO-OPs rather than probing a guessed branch. A check that can only produce false negatives is worse than no check.

**Pre-requisites**

- Depends on the new recorded-slug field landing first (same PR): `PipelineLedger` currently has no `slug` field — verified in Redis, `hasattr(ledger, "slug") == False` for `tomcounsell/ai:{2663,2716,2735}`.
- The three live lanes with no recorded identity (`sdlc-local-2663`, `sdlc-local-2716`, `sdlc-local-2735`; all `slug=None`, `branch_name=None`) are the self-heal cases the adopt-existing-branch healer must cover.

**Dropped**

- The suggested step 3 ("unwedge lane sdlc-2663 / PR #2668 once fixed") is **not** in this plan's scope as a manual operation — the fix itself unwedges it, and any leftover routing is the sdlc-2663 lane's own business. Verified separately that the recorded verdict there predates the current head (`session/sdlc-2663` has since advanced to `eeb89b4ee`, past the `69039fae5` cited in this issue), so its staleness gate will handle it.
- Carrying `headRefName` back from `_fetch_pr_state`'s existing call is **not** adopted: it makes the branch check depend on a live PR read, which is unavailable pre-PR and re-introduces a `gh`-read dependency for something the ledger can answer offline.

### LINKED ISSUE #2735: find_plan_path claims ownership from a bare cross-reference, wedging G8 and contaminating pipeline _meta
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/sdlc-lane-recorded-slug.md

## Problem

`tools/_sdlc_utils.py::find_plan_path` treats **any** textual mention of `#N` or `issues/N` in a plan document as evidence that the plan owns issue N. A `tracking:` frontmatter match is preferred, but when no plan claims the issue, the function falls back to the first plan that merely *cross-references* it — including references inside a "Not building" / No-Gos / Prior Context section, which are exactly the places a plan names an issue it explicitly does **not** own.

The docstring already recognises this hazard and suppresses the bare-`#N` fallback, but only on the `__file__`-fallback resolution path (`SDLC_TARGET_REPO` unset **and** not inside a git repo). Every real pipeline invocation sets `SDLC_TARGET_REPO` or runs inside the repo, so the suppression never applies in production.

## Live reproduction

Issue #2663 (`valor progress` CLI, PR #2668) has no plan document. `docs/plans/session-liveness-tick-counter.md` tracks **#2716** and mentions #2663 twice, once in an explicit no-go:

```
- **Not building the #2663 progress CLI.** Separate issue; this plan should
  consume the same progress truth when it exists, not pre-empt it.
```

Both issues resolve to the same plan:

```
$ SDLC_TARGET_REPO=~/src/ai python -c "from tools._sdlc_utils import find_plan_path; \
  print(find_plan_path(2663)); print(find_plan_path(2716))"
docs/plans/session-liveness-tick-counter.md
docs/plans/session-liveness-tick-counter.md
```

## Blast radius

A scan of `docs/plans/` (29 plans, 24 with `tracking:` frontmatter) finds **309 issue numbers with no owning plan that nonetheless resolve to one**.

Two consumers make this load-bearing rather than cosmetic:

1. **G8 artifact verification** (`tools/sdlc_next_skill.py::_verify_stage_artifacts_live`) derives `slug = Path(find_plan_path(issue)).stem` and live-checks `git ls-remote --heads origin session/{slug}`. On #2663 the real branch is `session/sdlc-2663` (pushed, head `69039fae5`), but the check looks for `session/session-liveness-tick-counter`, which does not exist:

   ```
   stage-artifact-verify: issue #2663 PATCH claims completed but branch
   session/session-liveness-tick-counter is not pushed
   {"skill": "/do-patch", "row_id": "G8", ...}
   ```

   G8 then re-dispatches `/do-patch` forever. `/do-patch` cannot clear it — the artifact it would produce is on a branch the check never looks at. The pipeline for #2663 is wedged until G4's oscillation cap escalates.

2. **`_meta` contamination.** `sdlc-tool stage-query --issue-number 2663` reports `revision_applied: true, revision_applied_at: 2026-08-11T03:27:10Z` — values read verbatim from #2716's plan frontmatter. #2663 has never had a plan, let alone a revision. `plan_exists: true` is false in the same way. These flags feed router rows 4b/4c and guard G7.

Ironically the failure shape mirrors the class of bug that motivated #2663 itself: a confident wrong answer derived from a signal that looks authoritative and is not.

## Why the current guard does not catch it

`find_plan_path`'s bare-`#N` suppression is gated on `_is_ai_repo_fallback`, which is only set when plans-dir resolution reaches step 3. Steps 1 and 2 (env var, git toplevel) cover every production call, so the fallback match is returned unguarded.

## Suggested direction (for `/do-plan`)

- Make `tracking:` frontmatter the **only** ownership signal, and return `None` otherwise. Callers already handle `None`: `_verify_stage_artifacts_live` computes `patch_claimed = ... and bool(slug)`, so a `None` slug correctly skips the branch check instead of checking the wrong branch, and router row 1 (`_rule_no_plan`) steps aside whenever a PR exists.
- If a textual fallback must survive for legacy plans, it should at minimum ignore matches inside no-go/prior-context sections and refuse to match when the plan carries a `tracking:` line pointing at a *different* issue — a plan that has declared its owner cannot also own someone else.
- Backfill `tracking:` frontmatter on any plan missing it, and consider a check that every open SDLC issue with a plan has an authoritative `tracking:` line.

## Acceptance criteria

- [ ] `find_plan_path(2663)` returns `None` while `docs/plans/session-liveness-tick-counter.md` (tracking #2716) is present and mentions #2663.
- [ ] `find_plan_path(2716)` still returns that plan.
- [ ] A test pins the no-go-section case: a plan tracking issue A that names issue B in a "Not building" line never resolves as B's plan.
- [ ] `stage-query --issue-number N` no longer reports `plan_exists`/`revision_applied` sourced from another issue's plan.
- [ ] G8 no longer re-dispatches `/do-patch` for an issue whose branch is pushed under the correct slug.

---

## Recon Summary

Performed 2026-08-13 against `main` @ `0dd8e70f2`. Every claim in the body above was re-executed, not taken on trust.

**Confirmed**

- **The bare-`#N` fallback is unguarded in production.** `tools/_sdlc_utils.py:840` sets `_is_ai_repo_fallback = False`, and only step 3 of plans-dir resolution (`:852`) flips it. `SDLC_TARGET_REPO` (`:842`) and `_git_toplevel()` (`:846`) cover every real invocation, so the suppression at `:888` never fires. The `fallback` assignment at `:879` returns any plan that merely matches `(?:#|issues/){N}(?![0-9])`.
- **Live repro still reproduces** (head `0dd8e70f2`):
  ```
  2663 -> docs/plans/session-liveness-tick-counter.md
  2716 -> docs/plans/session-liveness-tick-counter.md
  ```
- **G8 wedge is real.** `tools/sdlc_next_skill.py:205` `slug = Path(plan_path).stem`; `:236` `if pr_state != "MERGED" and not _check_branch_pushed(slug)`; `:159` probes `git ls-remote --heads origin session/{slug}`. `git ls-remote` shows only `refs/heads/session/sdlc-2663` exists (now at `eeb89b4ee`); `session/session-liveness-tick-counter` returns nothing.
- **`_meta` contamination is real and structural.** `tools/sdlc_stage_query.py:516` calls `_find_plan_path(issue_number)` and feeds it to `_parse_revision_applied` / `_parse_revision_applied_at` / `plan_exists`. Confirmed `PipelineLedger` for `tomcounsell/ai:2663` records `PATCH: completed`, `pr_number: None`.
- **`_compute_meta` reads a `slug` field that does not exist.** `tools/sdlc_stage_query.py:487` does `getattr(session, "slug", None)`, but `_resolve_issue_record` (`:581`) returns a `PipelineLedger`. Verified in Redis: `hasattr(ledger, "slug") == False` for #2663/#2716/#2735. That read is dead — always `None` — which is *why* every consumer re-derives instead.

**Revised**

- **The "fabricated shape" belief is false and is itself a defect.** `tools/sdlc_next_skill.py:346-349` and `tools/sdlc_stage_query.py:347-350` assert that `session/sdlc-{N}` is "a fabricated form this repo never creates". This checkout has **111 local / 99 remote** `session/sdlc-*` branches. They are minted by `reflections/sdlc_upvote_lanes.py:489` (`slug = f"sdlc-{issue_number}"`), `reflections/sdlc_progress.py:707`, and `tools/valor_session.py:107` (`_derive_slug_from_message`), all of which feed `agent/worktree_manager.py:1273` (`branch_name = f"session/{slug}"`). `reflections/sdlc_progress.py:116` (`_SDLC_BRANCH_RE = ^session/sdlc-\d+$`) and `reflections/sdlc_upvote_lanes.py:344/410` explicitly *query for* that shape. `tests/unit/test_sdlc_next_skill.py:117-176` pins the wrong belief and must be corrected, not preserved.
- **The root cause is broader than `find_plan_path`.** There is no recorded lane slug anywhere the SDLC tools read. Two independent slug provenances coexist (plan-filename stem vs. issue-number mint) with nothing reconciling them. `find_plan_path`'s textual fallback is one symptom; `_compute_meta`'s dead `getattr(session, "slug")` is another.

**Pre-requisites**

- Removing the bare-`#N` fallback requires an audit of plans lacking `tracking:` frontmatter. Five exist in `docs/plans/`: `keyfield-migration-fix.md` (tracks a *popoto* repo issue), `sdlc-1111.md` (closed, shipped), `session-recovery-observation-audit.md` (an audit doc, not a lane), `session-type-pm-rename.md` (#648, closed), `resilience-simplification-three-tier.md` (`tracking: none yet` placeholder). **None is an active lane**, so the fallback removal strands nothing live.
- Migration template for adding a nullable Popoto field: `scripts/update/migrations.py:335` `_migrate_confirm_is_ledger_field_readable` (read-only healing probe, registered at `:1032`).

**Dropped**

- Nothing dropped. The issue's third suggested direction ("ignore matches inside no-go sections") is explicitly **rejected** in favor of removing the fallback entirely — a heuristic that parses prose sections is exactly the "confident wrong answer from a signal that looks authoritative" failure the issue names.

### Live migration cases

`sdlc-local-2663`, `sdlc-local-2716`, and `sdlc-local-2735` are all `status=running`, `is_ledger=True`, with `slug=None`, `branch_name=None`, `pr_number=None`. `tools/sdlc_session_ensure.py:837` `AgentSession.create_local(...)` never passes a slug. These are the self-heal cases.

valorengels:
Closes #2735
Closes #2718

## The defect

An SDLC lane has exactly one identity: the thing that names its branch, its worktree, its task list, and (usually) its plan doc. Nothing wrote that identity down, so five places invented it independently and two of them invented a *different* shape:

| Inventor | Shape it invented |
|---|---|
| Reflection lane pickup | `sdlc-<issue>` |
| Stalled-lane respawn | `sdlc-<issue>` |
| CLI session create | `sdlc-<issue>` |
| G8 artifact verifier | plan-filename stem |
| `branch_exists` context | plan-filename stem |

The last two then probed for a branch under a name that never existed, which is how #2718's lane wedged: G8 declared the PATCH artifact missing and re-dispatched `/do-patch` forever against a PR that was open and clean.

The plan document those two consulted was itself resolved by a guess. `find_plan_path` treated **any** textual `#N` mention in **any** plan file as proof of ownership, including a mention inside a "Not building" No-Gos line, which is exactly where a plan names an issue it explicitly does not own.

## The fix

`PipelineLedger.slug` records the lane's identity, minted **once** at lane start by `ensure_session`, and every consumer reads it instead of guessing.

- **`tools/lane_identity.py`** is the new single home. `mint_lane_slug` holds the only issue-derived slug literal in the repo; `lane_branch_name` holds the only `session/` prefix; `find_plan_path` resolves an issue to its plan by `tracking:` frontmatter and nothing else.
- **The mint is one unconditional call near the top of `ensure_session`**, before the env short-circuit. That function has six success return points, so a per-return-point write would have missed the dominant conversational path; one early idempotent call covers all six by construction.
- **`allow_heal` defaults to `False`** and that default is load-bearing. `stage-query` runs for any issue number a router, dashboard, or operator names, most of which are not lanes. Read paths never mint, never create a ledger, never write.
- **The write is conditional-on-empty and takes no lease.** It is an identity write, not a stage transition; gating it on the lease would reintroduce the deadlock class #2026 closed. A lost race is not an error, the loser returns the winner's value.

### Adopt vs. derive

The ladder exists for callers that must *discover* an identity. Some callers already have one, and `adopt_lane_slug` is for them. The rule this PR establishes, in three clauses:

> A site that **searches** may guess. A site that **writes** identity may not. A site that already **knows** records what it knows rather than re-deriving it.

`reflections/sdlc_progress.py`'s create rung is the worked example. It only fires for an issue that already has a PR and a pushed branch, so its `slug` — derived from that branch — *is* the lane's identity. Asking the resolver to re-derive it would probe only the issue-derived name, miss a human-named branch, mint `sdlc-<issue>`, and create the session under a name that diverges from the branch the tick is looking at. That is the #2718 defect reappearing at the site meant to close it.

### Two slugs, kept distinct

The lane slug and the plan-document filename are not forced to match, and this PR is its own demonstration: lane `sdlc-2735`, plan `sdlc-lane-recorded-slug.md`. `tracking:` frontmatter is the only thing that links them, so `tools/plan_doc_scope.py` + `tests/unit/test_plan_docs.py` make that invariant enforceable, with the non-lane exclusions living in exactly one frozenset that both the test and the anti-criterion import.

### Observability

`_meta` gains `slug` and `slug_source` (`recorded` / `session` / `unresolved`), so an operator can tell "the branch check was skipped because no identity is recorded" from "the branch check ran and verified clean". The read is two rungs: the recorded ledger slug, then `AgentSession.slug` for pre-cutover lanes that never heal.

`sdlc-tool stage-query` and `next-skill` now wrap a tick in `cached_target_repo_resolution()`, so the meta read and the lane-slug read collapse to one `gh repo view` instead of two.

## Testing

387 passing across the affected suites. New coverage was written demonstrated-red: each case was confirmed failing against the pre-change tree before the fix, including the human-named-branch regression, which fails with `assert 'sdlc-2716' == 'session-liveness-tick-counter'` on the old re-derivation.

## Plan

`docs/plans/sdlc-lane-recorded-slug.md`, six critique cycles. The round-6 pass caught three blockers, all of which are fixed here; the round-5 root-cause note is worth keeping in mind for review: when two mechanisms must be kept in sync, prefer deleting one.

## Known scope boundary, not a defect

The human-named-branch case cannot presently reach `sdlc_progress`'s create rung through `_check_project_stalls`: `_list_open_sdlc_prs` filters heads on `_SDLC_BRANCH_RE` and `_issue_number_from_slug` requires `^sdlc-(\d+)$`, so such a PR is dropped before the ladder runs. The pass-through is correct regardless and is tested by calling `_attempt_action` directly. Whether that reflection should admit non-`sdlc-<issue>` lane branches at all is a separate decision — widening `_issue_number_from_slug` would need another source for the issue number, since a human-named branch does not carry one.


valorengels:
## Review: APPROVE

Every invariant the plan committed to was checked against the code rather than read for, and all ten hold. The acceptance cases were run, not reasoned about: `find_plan_path(2663)` returns `None` while `find_plan_path(2716)` still returns `docs/plans/session-liveness-tick-counter.md`, and there are zero stale `find_plan_path` imports from `tools._sdlc_utils` across `tools/`, `reflections/`, and `tests/`.

Anti-criteria: AC-MINT 1 (the mint rung only), AC-HEAL 2 sanctioned healers with zero hits in `tools/valor_session.py`, AC-STALEIMPORT 0, AC-FALLBACK 0, AC-BELIEF 0, AC-TRACKING 0 unowned plans, AC-ADOPT 1. AC-GHCOUNT is 1, down from 2 on `main`. 556 tests pass across the affected suites; ruff clean.

### One concern, fixed in `fa06dd57a`

`_check_plan_committed_on_main`'s path was relativized against the process cwd rather than the repo root. `git show main:<path>` paths are always repo-root-relative whatever cwd the subprocess runs in, and `find_plan_path` resolves against the repo root too, so the two ladders disagreed from any subdirectory. Reproduced from `docs/`: the plan resolved to `plans/sdlc-lane-recorded-slug.md` instead of `docs/plans/...`, no `ValueError` fired, and `_verify_stage_artifacts_live` reported `{"stage_artifacts_verified": false, "unverified_stage": "PLAN"}` — routing a lane with a properly committed plan back to `/do-plan`. That is the #2718 symptom through a different door.

Live invocation paths were safe today (the local `/do-sdlc` wrapper pins cwd to the repo root via `uv run --directory`, and the bridge runs from the checkout root), but nothing enforced it. Fixed by falling back to `_git_toplevel()` before `Path.cwd()`, with a demonstrated-red regression test: reverting the fix produces `assert 'plans/sdlc-2078-fixture.md' == 'docs/plans/sdlc-2078-fixture.md'`.

### Judgment calls

**Error handling.** The asymmetry the feature demands is respected. Because the write is no-overwrite, a wrong identity is permanent, so the heal path deserves stricter treatment than the read path — and every swallow there lands on a fall-through to the next rung, never on a confident wrong slug. The ambiguous-PR-match case logs a warning and falls through rather than taking `matches[0]`.

**Concurrency.** Two lanes on the same issue converge on one identity in every realistic interleaving. The winner SETNXes the slug lock, re-reads, writes; the loser polls and returns the winner's value. On broker error the lock fails open, degrading to bare conditional-on-empty, which is exactly the pre-lock behavior. No deadlock is possible: no lease is taken and the release is guarded.

**Test quality.** Real rather than theatrical — live Popoto against Redis, with only `_ls_remote_heads` stubbed to keep the suite offline. The demonstrated-red claims hold on spot-check: `test_create_rung_records_a_human_named_slug_verbatim` asserts `!= "sdlc-2716"`, which is precisely what `main`'s `slug=f"sdlc-{issue_number}"` produces.

**Cruft.** Clean. `_derive_slug_from_message` is deleted outright, and the plan-stem adoption rung is absent from the ladder rather than deprecated — with both the module docstring and the feature doc explaining why it must stay absent, not narrating that it once existed.

### Tech debt, deliberately not fixed here

Two pre-existing hand-built branch shapes that the feature's own rule argues against, both untouched on `main` and both outside this PR's scope:

- `reflections/sdlc_upvote_lanes.py:423` — `_count_live_lanes` filters on `startswith("session/sdlc-")`, so a human-named lane is invisible to the concurrency cap and the cap over-admits. Changing the counter's semantics inside an identity PR is scope creep.
- `tools/disk_reclaim.py:405` — builds `f"session/{slug}"` and acts on it without routing through `lane_branch_name`. Outside the SDLC identity path.

A third boundary is documented rather than fixed: the human-named branch cannot presently reach `sdlc_progress`'s create rung, because `_issue_number_from_slug` requires `^sdlc-(\d+)$`. The pass-through is correct regardless and the feature doc now says so explicitly instead of presenting an unreachable case as live.

Issue #2801, opened 2026-08-13
Investigate: G4's counter caps alternating oscillation at 1 — structurally real, but no harmful instance is known


valorengels:
Split out of #2730's lane after a critique round and two spikes. **Filed as an investigation, not a defect report** — the structural property is proven, the harm is not. Both halves are below deliberately, so the next reader starts with the full picture rather than the alarming half.

## The structural property (proven)

`agent/sdlc_router.py::compute_same_stage_count` walks `_sdlc_dispatches` backward and requires **both** the same skill **and** the same `stage_snapshot`. The skill test terminates the walk:

```python
for entry in reversed(history):
    ...
    if entry.get("skill") != skill:
        break                      # any alternation stops the walk here
```

So an A→B→A→B loop scores at most 1, and `MAX_SAME_STAGE_DISPATCHES = 3` is unreachable for any alternating shape. The router's alternating loops are rows 8↔8b (`/do-patch` ↔ `/do-pr-review`) and rows 2b↔3 (`/do-plan-critique` ↔ `/do-plan`).

Several rows nonetheless *cite* G4 as their bound. Case-insensitive `grep -in "bound by G4" agent/sdlc_router.py` returns **eleven** hits (`:984, 1167, 1275, 1312, 1365, 1414, 1441, 1567, 1630, 1644, 1655`). Rows 2c/8c/8d/8e/8f repeat a single skill and are genuinely bounded; rows 2b↔3 correctly cite G5 instead. `_rule_patch_applied_after_review` (`:1275`) is the one that claims a G4 bound for a loop that provably alternates.

## The counter-evidence (why this is not filed as a bug)

**No harmful instance is known, and the one case that looked like one was healthy behavior.**

The motivating example was the `verdict-finalize-cluster` ledger (merged `706fc4da0`): four `/do-plan` ↔ `/do-plan-critique` rounds with `same_stage_dispatch_count: 0`. Running the function directly against that real ledger:

```
backward walk only:        (1, '/do-build')
with live snapshot (real): (0, '/do-build')
```

The `0` comes from the **D5 self-clearing branch** (`:1888-1894`), not the skill break. And the recorded snapshots show adjacent plan/critique pairs sharing a snapshot while each *cycle* differs — the state advanced every round because the plan genuinely changed each round. Those four rounds were productive work and **G4 correctly declined to escalate.**

## Why the obvious fix does not work

Deleting the skill break was proposed and measured with the change applied:

[...]

Issue #2731, opened 2026-08-12
Stage dispatch has no liveness gate: a second agent can enter a stage that already has a live one


valorengels:
Split out of #2650, whose plan was cut to a minimal set (plan-doc write lease + off-main commit guard). This is the piece that was deliberately left out, filed so it is not lost.

> **Owner ruling 2026-08-13 — scope narrowed, original sketch overruled.** Liveness belongs to the **issue lease**. No `dispatcher` block on the dispatch record, and no new stage-scoped claim key. The gate therefore covers **inter-run collisions only**. The originally-observed intra-run collision is **explicitly out of scope** — see "Known residual" below. The Sketch and acceptance criteria below have been rewritten to match; do not re-litigate from the original text.

## Problem

`/do-sdlc` can dispatch a second agent into a stage that already has a live agent in it. Observed on issue #2629's BUILD stage: two concurrent `/do-build` agents ran on the same slug and the same worktree (`.worktrees/react-transport-derivation`). Both produced byte-identical production code, so the collision cost only wasted compute — the same collision on divergent implementations would have corrupted the branch.

Nothing at the tool layer currently refuses this:

- `record_dispatch` (`agent/sdlc_router.py`) appends exactly `{skill, at, stage_snapshot}`. No liveness signal is consulted before the append.
- `same_stage_dispatch_count` (`tools/sdlc_stage_query.py`) counts *oscillation* — repetition over time — and feeds G4's `MAX_SAME_STAGE_DISPATCHES` cap. It is blind to overlap *in time*.

## Why a naive fix is wrong

The #2629 first dispatch had **died mid-work and was legitimately recovered** (its checkpoint commits say so). So "refuse if `same_stage_dispatch_count >= 1`" would block the recovery path, which is the common and correct case. The refusal needs a real liveness test, not a counter threshold.

## Desired outcome

`sdlc-tool dispatch record` refuses (non-zero, named reason) when the **issue lease is held by a different `run_id` that is provably live**. It allows when the prior holder is provably dead (the #2629 recovery case), and allows when liveness is indeterminate is **not** true — indeterminate must refuse, per `agent/pid_fence.py`'s canonical rule that unknown never authorizes more force.

The lease is already the single source of run ownership and, since #2714 anchored its heartbeat to the supervisor's lifetime, it is a trustworthy liveness signal for a *run*. This issue makes `dispatch record` consult it before appending, rather than authorizing purely on lease *ownership* as it does today.

[...]

Issue #2491, opened 2026-07-31
Pipeline graph is not a single source of truth: 7 hardcoded duplicates, 4 disagreeing stage-model tables, circular doc authority


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/pipeline-graph-single-source-of-truth.md

> **Context:** The **SDLC pipeline** is this repo's staged workflow taking a GitHub issue from idea to merged code (`ISSUE → PLAN → CRITIQUE → BUILD → TEST → REVIEW → DOCS → MERGE`, with failure cycles). Its **pipeline graph** — the stage list, the stage→skill mapping, the transition edges, and the loop caps — is claimed to be canonically defined in [`agent/pipeline_graph.py`](../blob/main/agent/pipeline_graph.py). This issue reports that the claim is false, and asks to make it true.

## Problem

`agent/pipeline_graph.py` presents itself as the single source of truth for the pipeline graph. A repo-wide sweep found **7 hardcoded code duplicates** of its contents, **~30 prose restatements**, and **only 2 clean importers**. The duplicates have already drifted, and two of them carry live defects.

This is the same disease [#1216](../issues/1216) diagnosed and partially cured. #1216 replaced the markdown *dispatch table* with a `sdlc-tool next-skill` call, but explicitly deferred the rest — its plan's No-Gos section reads: *"Touching dev-session boundaries, model selection, and PM steering protocol is its own architectural change."* This issue picks up that deferred work.

**Current behavior:**

*Hardcoded code duplicates (category C):*

| # | Site | Duplicates | Note |
|---|---|---|---|
| C1 | [`models/agent_session.py:81`](../blob/main/models/agent_session.py#L81) | `DISPLAY_STAGES` | `SDLC_STAGES = [...]`, byte-identical. Drives `AgentSession.current_stage` at `:1503`. |
| C2 | [`agent/pipeline_state.py:48`](../blob/main/agent/pipeline_state.py#L48) | node set + PATCH | `ALL_STAGES`. The module **already imports `PIPELINE_EDGES` two lines earlier**. |
| C3 | [`agent/hooks/pre_tool_use.py:52`](../blob/main/agent/hooks/pre_tool_use.py#L52) | node set | `_SDLC_STAGE_NAMES` frozenset. |
[...]

Issue #2003, opened 2026-07-10
SDLC pipeline substrate: single run ownership, live-ref PR resolution, real merge-gate enforcement


tomcounsell:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/sdlc-run-ownership-merge-enforcement.md

> **Context:** This repo runs an SDLC pipeline (Plan → Critique → Build → Test → Review → Docs → Merge) for GitHub issues through two entry points: a local Claude Code session driving the `/do-sdlc`/`/sdlc` skills, and the standalone worker (`worker/__main__.py`) driving a headless PM+dev runner. Both fan out to short-lived `sdlc-tool` CLI subprocesses (`tools/sdlc_*.py`) that read/write pipeline state on a Redis-backed `AgentSession` record, routed by a rule-table router (`agent/sdlc_router.py`). This issue is **Workstream A of the resilience-simplification program** ([`docs/plans/resilience-simplification-three-tier.md`](https://github.com/tomcounsell/ai/blob/main/docs/plans/resilience-simplification-three-tier.md), items T1.7 + T2.1 + T2.2), written fresh from that plan's current understanding per its supersession rule. It supersedes and closes #1987.

## Problem

Pipeline ownership and terminal enforcement are each split across surfaces that were patched per-incident and still disagree. Three defects, one file surface:

**1. PR-number resolution trusts fuzzy text search (T1.7; #1987, live-blocked).**
`_lookup_pr_number` (`tools/sdlc_stage_query.py:~277`) resolves an issue's PR via `gh pr list --search "#{N}"` — GitHub's tokenized text search, not an exact reference match. Confirmed live: `--search "#1950"` returned PR #1984 (whose body says `Closes #1967`), so `next-skill` routed issue #1950 toward review of an unrelated PR; the #1950 pipeline is halted on this today. The authoritative live-refs query (`gh pr list --head session/{slug}`) is currently the *fallback*, below the laggy, false-matching search rung. Same rung contributed to #1915's duplicate PRs (search indexing lag defeated the pre-create check). Adjacent dead signal: `sdlc_next_skill.py:~136` computes `branch_exists` by checking branch shape `session/sdlc-{N}` — a shape this repo never creates (`sdlc_stage_query.py:~270` calls it "a fabricated form") — so row 5's context signal is permanently False.

**2. Run ownership is keyed to OS process identity, not the logical pipeline run (T2.1).**
[...]

PR #2802, opened 2026-08-13
Make the SDLC dispatch record unbypassable via a router upsert slot (#2730)

### LINKED ISSUE #2730: SDLC stage markers and dispatch history diverge: only the router records dispatches, so router-bypassing stages complete with markers but no ledger entry (defangs the G4 oscillation guard)
## Summary

The issue-keyed `PipelineLedger` dispatch history is supposed to be the record of which SDLC sub-skills were dispatched for an issue. For issue #2711 it contains **exactly one entry**, while the stage markers on the same ledger assert **five stages completed** and the branch carries **seven commits** produced by those stages.

BUILD, TEST, REVIEW, and an entire earlier `/do-patch` cycle (which produced and pushed commit `a512e2b20`) executed and completed without ever recording a dispatch. Stage markers and dispatch history diverged, and nothing reconciles them.

## Reproduction

```
$ cd /Users/valorengels/src/ai && sdlc-tool dispatch get --issue-number 2711
[{"skill": "/do-patch", "at": "2026-08-11T15:41:26.478778+00:00", "stage_snapshot": {"stages": {"ISSUE": "completed", "PLAN": "completed", "CRITIQUE": "completed", "BUILD": "completed", "TEST": "completed", "PATCH": "pending", "REVIEW": "in_progress", "DOCS": "pending", "MERGE": "pending"}, "_verdicts": {"CRITIQUE": {"verdict": "NEEDS REVISION", "artifact_hash": "sha256:c605f08c..."}, "REVIEW": {"verdict": "CHANGES REQUESTED", "artifact_hash": null, "blockers": 0, "tech_debt": 3}}, "_patch_cycle_count": 0, "_critique_cycle_count": 0, "pr_number": 2728}, "run_id": "ff285189222849df94510c5a378199ae"}]
```

```
$ cd /Users/valorengels/src/ai && git log --oneline origin/session/docs-auditor-rename-guard -8
afabc1bec fix(#2711): close out the cycle-2 review dispositions with real tracker records
a512e2b20 fix(#2711): sweep the docstring overclaim and carry withheld count into liveness
3a38e74b7 docs(#2711): sync plan checkboxes with review verdict
0b1cf0131 fix(#2711): surface withheld fixes on the rotation path and correct swept doc claims
62b679ddd docs: document the existence invariant and word-anchored stale terms
d238cbc70 docs: sweep invented agent_sessions.py residue and retired session_router.py refs
3b5cc06a9 fix(docs-auditor): word-anchor stale terms and enforce path-existence invariant
a13a31a08 Plan revision (docs-auditor-rename-guard): address cycle-1 critique findings
```

## Observed vs expected

- **Observed:** 1 dispatch entry (`/do-patch`); stage markers claim ISSUE/PLAN/CRITIQUE/BUILD/TEST completed and REVIEW in_progress; 7 stage-produced commits on the branch; a recorded REVIEW verdict (`CHANGES REQUESTED`) and a recorded CRITIQUE verdict.
- **Expected:** at minimum one dispatch entry per executed stage — `/do-plan`, `/do-plan-critique`, `/do-build`, `/do-test`, `/do-pr-review`, plus each `/do-patch` cycle. The ledger's own `stage_snapshot` inside the single entry is self-contradicting evidence: it reports work that has no corresponding dispatch record anywhere in the history.

## Why it matters

1. **The G4 oscillation guard is defanged.** Dispatch history is what the router's G4 guard counts (`tools/sdlc_dispatch.py`, `agent/sdlc_router.py::record_dispatch`). A near-empty history means the guard has nothing to count, so a router can re-dispatch the same stage indefinitely without G4 ever firing.
2. **The ledger is untrustworthy as an audit trail.** An operator reading dispatch history sees an essentially idle pipeline while seven commits of real work sit on the branch.
3. **It caused a live misroute.** A `/sdlc` invocation on this plan initially assessed the pipeline as needing work that commit `a512e2b20` had already done, precisely because the ledger showed no dispatch for it.

## Root-cause finding (from grep)

`sdlc-tool dispatch record`'s own contract is caller-side: "Call AFTER guard evaluation but BEFORE invoking the sub-skill." The code and skill bodies match that contract — **the router records, the stage skills do not**.

Files under `.claude/skills*/` and `docs/sdlc/` that mention `dispatch record` (excluding worktree copies):

- `.claude/skills/sdlc/SKILL.md` (the router)
- `.claude/skills-global/do-sdlc/SKILL.md` (the supervisor)

That is the complete list. **No stage skill** (`do-plan`, `do-plan-critique`, `do-build`, `do-test`, `do-pr-review`, `do-patch`, `do-docs`, `do-merge`) calls `sdlc-tool dispatch record`.

Contrast with stage markers, which stage skills *do* write themselves — files mentioning `stage-marker`:

- `.claude/skills/sdlc/SKILL.md`, `.claude/skills-global/do-sdlc/SKILL.md`
- `.claude/skills-global/do-build/WORKFLOW.md`, `do-plan-critique/SKILL.md`, `do-pr-review/SKILL.md` (+ `sub-skills/post-review.md`), `do-docs/SKILL.md`, `do-merge/SKILL.md`
- `docs/sdlc/do-build.md`, `do-plan.md`, `do-plan-critique.md`, `do-pr-review.md`, `do-merge.md`

**This asymmetry is the defect's mechanism:** stage completion is self-reported by the stage, while dispatch is recorded only by the router. Any stage that executes without going through the router therefore writes a marker and no dispatch record — exactly the divergence observed here.

A concrete non-router invocation path exists in the repo: `.claude/skills-global/do-build/SKILL.md:128` instructs do-build to "route failures to `/do-patch` (bounded iterations) and re-run" — a direct skill-to-skill chain that never passes through `/sdlc`'s Step 3.5 record. `/do-patch` is precisely the stage that produced the unrecorded commit `a512e2b20`.

*Hypothesis (unverified):* stages invoked directly by a human (`/do-build` typed into a session) or resumed mid-pipeline take the same unrecorded path. Not confirmed against this run's transcript.

## Distinction from existing issues (not a duplicate)

- **#2675** (OPEN, "do-sdlc Step 3d.6 continuity re-ensure mints a new run_id after lease lapse, recreating the ledger anchor and wiping stage markers") — that is about existing state being **wiped** by a new run_id. Here nothing was wiped: the stage markers for those stages are present and correct; the dispatch records were **never written in the first place**, while the marker for the same stage *was* written under the same run_id `ff285189222849df94510c5a378199ae`.
- **#2144** (CLOSED, "Resumed pipeline turns lose SDLC run identity — stage markers/verdicts silently fail after mid-turn interruption") — that was writes **failing** after an interruption, losing run identity. Here the marker and verdict writes succeeded; only dispatch records are missing. That asymmetry points at skills completing without the router ever calling `dispatch record`, not at a lost run identity.
- **#2637** (OPEN, "sdlc_dispatch and sdlc_meta_set carry the same frozen lease-helper import that broke #2469") — a *candidate* contributing root cause, offered as a hypothesis rather than an assertion: `tools/sdlc_dispatch.py:54` snapshots `resolve_ledger_lease`/`revalidate_ledger_lease` into module globals via a `from`-import, and #2469's precedent shows such a snapshot can silently invert a lease assertion, which in `sdlc_dispatch` would surface as a refused (`{"ok": false}`) dispatch record rather than a raised error. **However**, #2637 states "nothing triggers it today," and it would not explain markers succeeding while dispatch records fail under the same run. Worth ruling in or out during the fix, not assumed.

## Suggested direction (not prescriptive)

Either make the dispatch record the responsibility of the same actor that writes the stage marker (so the two can never diverge), or add a reconciliation check that flags a ledger whose stage markers report completed work with no corresponding dispatch entry.

valorengels:
Closes #2730

**#2730's record-completeness half ships here; its counter half is tracked on #2801.**
Closing the issue because the mechanism it was filed about — markers without dispatch records — is fixed, and the second mechanism named in its "defangs the G4 oscillation guard" claim now has its own issue carrying the full evidence and measurements. The record-completeness half ships here. The G4 counter defect — the second mechanism named in the issue's "defangs the G4 oscillation guard" claim — is **not** fixed, and is filed as #2801 with its full evidence. Details in "What is not fixed" below.

## The defect

Only the router wrote `_sdlc_dispatches`, while stage markers were written by the stage itself. Any stage that ran without passing through the router therefore left a marker and no ledger entry. Issue #2711's ledger shows the result: **six completed stages against zero dispatch records**.

Two bypass paths exist today, and the second is the dominant one:

- `/do-build` routes failures straight to `/do-patch` (`do-build/SKILL.md:128`) — a skill-to-skill chain that never reaches the router's record step.
- The **PreToolUse hook** calls `PipelineStateMachine.start_stage()` on every `Skill(do-*)` call and never touches `write_marker`. Per `docs/features/sdlc-local-supervision.md`, on the bridge that hook is how stage markers get written at all.

That second path is why the obvious fix location is wrong: a change in `write_marker` would have been **a no-op exactly where the divergence was observed**. `PipelineStateMachine._activate_stage` is the single point every `start_stage` branch and every actor funnels through, so the record is made there.

## The design: an upsert slot, not a dedup predicate

Two writers observing one dispatch would double every router-driven entry and halve G4's effective threshold. A dedup predicate can't distinguish "the router already recorded this" from "this is a genuine second entry" — and deduplicating the latter would disarm the very guard this re-arms.

So the router records an **unconfirmed slot** (`confirmed: false`) before invoking, and the stage entry upgrades it in place. When no slot exists, the stage was reached without the router and a confirmed record is appended.

| scenario | records |
|---|---|
| router-driven dispatch | 1 (slot upgraded) |
| `/do-build` → `/do-patch` chain | 1 for each — `/do-patch` no longer invisible |
| genuine re-entry into a stage | 2 — the G4 signal, preserved |
| `start_stage` on an already-`in_progress` stage | 0 — no entry occurred |
| unknown stage | 0, and no raise |

## The two placements that matter

**Where in the save.** The record is applied inside `_save()` **after** the preserved-metadata merge, not in `_activate_stage`. `_save()` re-reads `_sdlc_dispatches` from the backing store and merges that copy over anything in memory, so a record written before the merge is **silently discarded**. Applying it after means it sees the freshest history and rides the same single write — no second read-modify-write to race against.

**Which side of the transition.** Both writers must snapshot the *same* side, and originally they did not. The router records its slot before invoking, so it sees the entering stage's pre-entry status; an appended record is written after `_activate_stage` already flipped the stage, so it saw `in_progress`. `compute_same_stage_count` breaks its walk on the first snapshot mismatch, so a loop alternating router-driven and bypass entries counted **lower than before this feature existed** — measured on a three-cycle `/do-patch` loop, `main` scores 2 and the branch scored 1. Adding the correct record was lowering the streak.

`_activate_stage` now carries the pre-entry status alongside the pending stage, and the appended record's snapshot is built against that value. This was found in review; the regression test gives PATCH a real pre-entry status and was verified red before the fix (`['failed', 'in_progress', 'failed']`) and green after. Every pre-existing `TestStageEntryUpsert` fixture started from a bare `states = {}`, where the diverging field is empty on both sides and the defect is invisible.

The same hazard applies to `pr_number`, which bypass records inherit from the newest record carrying one, for the same reason.

## `last_dispatched_skill` changes, and that is the fix

`history[-1]["skill"]` now names the skill that actually ran rather than whichever one the router last saw. After a `/do-build` → `/do-patch` chain it reads `/do-patch`; previously it read `/do-build` while `/do-patch` was doing the work.

This value has five consumers — G1 (`:336`), G3 (`:403`), G7 (`:685`), row 8b (`:1284`), row 8d (`:1380`) — and rows 8/2b read `_latest_dispatch_at`. All were enumerated and all their suites pass unchanged. Rows 8b and 8d gate on exactly this value, so making it true is the point, not a side effect.

A sixth consumer surfaced in review: `tools/sdlc_stage_marker.py`'s retroactive-skip check. Because every stage entry now records a dispatch, a `/do-plan` that entered PLAN and then aborted to ask a question is no longer retroactively skippable. That is the guard's stated contract ("the stage was reached") read honestly, and it errs toward refusing a skip rather than granting a false one. Noted at the call site.

## Guard integrity

`stage` and `confirmed` ride on the record and **never** enter `stage_snapshot`. G4 compares snapshots; a per-dispatch-varying field there would stop every comparison matching and silently disable oscillation detection outright. Pinned by a test on the snapshot's exact key set.

`stage_for_skill` inverts the canonical `STAGE_TO_SKILL` rather than adding a fourth hand-maintained map — the three that already exist are catalogued on #2491.

## The one false claim, swept

`_rule_patch_applied_after_review` (row 8b) asserted in two places that G4 bounds its re-review loop. It does not: the 8↔8b loop alternates `/do-patch` with `/do-pr-review`, failing the skill test, and `_patch_cycle_count` moves every cycle, failing the snapshot test. G2 caps critique cycles only. Both claims now state the loop is unbounded and point at #2801.

All eleven `bound by G4` claims were audited. The other ten name same-skill re-dispatch loops (rows 2c/8c/8d/8e/8f) that G4 genuinely counts, and are left alone.

## What is not fixed, and why

**The G4 counter still scores alternating oscillation at ≤1.** `compute_same_stage_count` breaks its walk on the first differing skill, so an A→B→A→B loop never reaches the threshold. That is filed as **#2801** rather than fixed here, because the obvious fix does not work and I have no evidence of harm:

- Removing the skill break was measured. A *frozen* alternation then counts 4 and escalates, but a **real** 8↔8b loop counts 2 and does not — `build_stage_snapshot` includes `_patch_cycle_count`/`_critique_cycle_count`, and those increment every cycle, so the snapshot conjunct breaks the walk anyway.
- The one history that looked like a defect turned out to be **productive work** G4 correctly declined to escalate.
- The trade-off is real: the snapshot conjunct that correctly protects productive work is the same thing that defeats detection of the stuck version of that loop. #2801 carries both halves so the next reader starts with the full picture.

**#2731 (dispatch liveness gate) is closed as re-scoped-and-deferred**, not built. Under the owner's lease-read-only ruling the remaining inter-run gate is a no-op: `resolve_ledger_lease` already refuses every foreign holder, live or dead, and already surfaces `orphaned_lock`. The intra-run collision it was filed about cannot be gated by the lease at all — forks inherit the supervisor's `run_id`, and under `/do-sdlc` they are subagents of one `claude` process sharing every identity a fence could read. Full analysis and the shape of the eventual fix are on the issue.

## Plan

`docs/plans/ledger-integrity.md` is revised to rev-3 on `main`, recording the lead's fence/scope ruling (fence extended to `agent/pipeline_state.py`; counter change split to #2801; #2731 closed) and rewriting the operative sections to describe what shipped. Round 2's escalation is closed there rather than only in this PR body.

## Verification

**699 tests pass** across `test_sdlc_router{,_oscillation,_decision}.py`, `test_sdlc_dispatch.py` (unit + integration), `test_pipeline_state_machine.py`, `test_sdlc_stage_marker.py`, `test_sdlc_stage_query.py`, `test_pre_tool_use_start_stage.py`, `test_post_tool_use_stage_completion.py`, `test_sdlc_lease_helper_binding.py`, and `test_sdlc_next_skill.py`. `ruff check .` and `ruff format --check` clean.

One failure in that set, `TestDispatchRecordLease::test_lease_lost_between_peek_and_write_refuses`, reproduces identically on current `main` at the same SHA this branch is rebased onto — pre-existing, not from this branch, and unrelated to the changed code (it exercises `_cli_record`'s mocked lease call sequence).

New tests cover both directions rather than presence alone: exact record **counts** per scenario (a naive unconditional append fails them), the pre-entry-status pin (verified red), the `/do-build`→`/do-patch` regression fixture the change is justified by, the snapshot key-set pin, survival of `_save()`'s merge, and that a failing dispatch record never blocks the marker write.

### DIFF
--- a/agent/pipeline_state.py
+++ b/agent/pipeline_state.py
@@ -406,6 +406,9 @@ class PipelineStateMachine:
         # build the instance via __new__) lands with the full attribute set.
         self.stage_skips = {}
         self._skip_precondition_cache: dict[str, tuple[str, str] | None] = {}
+        # (stage, pre-entry status) staged by _activate_stage for _save() to
+        # record as a dispatch entry (#2730).
+        self._pending_stage_entry: tuple[str, str | None] | None = None
 
         def _apply(data: dict) -> None:
             self.states = {k: v for k, v in data.items() if k in ALL_STAGES}
@@ -507,8 +510,10 @@ class PipelineStateMachine:
         key (``_verdicts``, ``_sdlc_dispatches``, or any future ``_*`` key)
         would be silently dropped if we serialized ``self.states`` alone. To
         protect cross-writer invariants — especially the verdict recorder in
-        ``tools.sdlc_verdict`` and the dispatch recorder in
-        ``agent.sdlc_router.record_dispatch`` — we reload the latest raw
+        ``tools.sdlc_verdict`` and the router's dispatch recorder in
+        ``agent.sdlc_router.record_dispatch`` (and, since #2730, ``_save()``
+        itself, which appends the stage-entry record AFTER this merge — see
+        ``_apply_pending_stage_entry``) — we reload the latest raw
         stage-state blob from the active backing store BEFORE writing and
         merge every ``_*`` key we did not manage ourselves. This makes
         ``_save()`` a safe participant in the cross-process stage-state
@@ -583,7 +588,7 @@ class PipelineStateMachine:
         Never raises: a stage must not become unrecordable because its audit
         entry failed.
         """
-        pending = getattr(self, "_pending_stage_entry", None)
+        pending = self._pending_stage_entry
         if not pending:
             return
         self._pending_stage_entry = None
@@ -592,7 +597,7 @@ class PipelineStateMachine:
             from agent.sdlc_router import confirm_or_append_stage_entry
 
             confirm_or_append_stage_entry(data, stage, prior_status=prior_status)
-        except Exception as e:  # pragma: no cover - audit must never block a write
+        except Exception as e:  # audit must never block a write
             logger.debug(f"pipeline_state: stage-entry dispatch record failed for {stage}: {e}")
 
     def _load_preserved_metadata(self) -> dict:
--- a/agent/pipeline_state.py
+++ b/agent/pipeline_state.py
@@ -549,10 +549,50 @@ class PipelineStateMachine:
                 continue
             data[key] = value
 
+        # #2730: record the stage entry AFTER the merge, so it operates on the
+        # freshest `_sdlc_dispatches` just read from the store rather than a
+        # stale in-memory copy, and lands inside this same single write.
+        self._apply_pending_stage_entry(data)
+
         try:
             self._write_raw(data)
         except Exception as e:
             logger.warning(f"Failed to save stage_states for {self._store_label()}: {e}")
+
+    def _apply_pending_stage_entry(self, data: dict) -> None:
+        """Record a pending stage entry into ``data``'s dispatch history (#2730).
+
+        Every branch of ``start_stage`` funnels through ``_activate_stage``, and
+        so does every actor that starts a stage: the skill bodies via
+        ``sdlc-tool stage-marker``, the PreToolUse hook (which never touches
+        ``write_marker`` and is the dominant marker writer on the bridge), and
+        ``/do-sdlc``'s backfills. Recording here is what makes the dispatch
+        history unbypassable. Previously only the router wrote it, so a stage
+        reached by a skill-to-skill chain such as ``/do-build`` → ``/do-patch``
+        left a marker and no ledger entry, and G4 had nothing to count.
+
+        The double-count question is delegated to
+        ``sdlc_router.confirm_or_append_stage_entry``: the router's own
+        pre-invocation record is an *unconfirmed slot* that this upgrades in
+        place, so a router-driven dispatch still yields exactly one record.
+
+        Called from ``_save()`` after the preserved-metadata merge, so it sees
+        the freshest history and rides the same single write -- there is no
+        second read-modify-write to race with.
+
+        Never raises: a stage must not become unrecordable because its audit
+        entry failed.
+        """
+        stage = getattr(self, "_pending_stage_entry", None)
+        if not stage:
+            return
+        self._pending_stage_entry = None
+        try:
+            from agent.sdlc_router import confirm_or_append_stage_entry
+
+            confirm_or_append_stage_entry(data, stage)
+        except Exception as e:  # pragma: no cover - audit must never block a write
+            logger.debug(f"pipeline_state: stage-entry dispatch record failed for {stage}: {e}")
 
     def _load_preserved_metadata(self) -> dict:
         """Return underscore-prefixed metadata keys from the live backing store.
@@ -598,6 +638,11 @@ class PipelineStateMachine:
     def _activate_stage(self, stage: str) -> None:
         """Set stage to in_progress, save, and record analytics."""
         self.states[stage] = "in_progress"
+        # #2730: flag the entry for _save() to record. It cannot be applied
+        # here -- _save() re-reads `_sdlc_dispatches` from the backing store and
+        # merges that copy, so an in-memory mutation made now would be silently
+        # discarded by the very next line.
+        self._pending_stage_entry = stage
         self._save()
         _record_stage_metric("sdlc.stage_started", stage)
 
--- a/agent/pipeline_state.py
+++ b/agent/pipeline_state.py
@@ -583,14 +583,15 @@ class PipelineStateMachine:
         Never raises: a stage must not become unrecordable because its audit
         entry failed.
         """
-        stage = getattr(self, "_pending_stage_entry", None)
-        if not stage:
+        pending = getattr(self, "_pending_stage_entry", None)
+        if not pending:
             return
         self._pending_stage_entry = None
+        stage, prior_status = pending
         try:
             from agent.sdlc_router import confirm_or_append_stage_entry
 
-            confirm_or_append_stage_entry(data, stage)
+            confirm_or_append_stage_entry(data, stage, prior_status=prior_status)
         except Exception as e:  # pragma: no cover - audit must never block a write
             logger.debug(f"pipeline_state: stage-entry dispatch record failed for {stage}: {e}")
 
@@ -637,12 +638,15 @@ class PipelineStateMachine:
 
     def _activate_stage(self, stage: str) -> None:
         """Set stage to in_progress, save, and record analytics."""
+        prior_status = self.states.get(stage)
         self.states[stage] = "in_progress"
         # #2730: flag the entry for _save() to record. It cannot be applied
         # here -- _save() re-reads `_sdlc_dispatches` from the backing store and
         # merges that copy, so an in-memory mutation made now would be silently
-        # discarded by the very next line.
-        self._pending_stage_entry = stage
+        # discarded by the very next line. The pre-entry status rides along
+        # because the record must snapshot the same side of this transition the
+        # router's slot did; see confirm_or_append_stage_entry.
+        self._pending_stage_entry = (stage, prior_status)
         self._save()
         _record_stage_metric("sdlc.stage_started", stage)
 
--- a/agent/sdlc_router.py
+++ b/agent/sdlc_router.py
@@ -1798,11 +1798,15 @@ def stage_for_skill(skill: str | None) -> str | None:
     return None
 
 
+_NO_PRIOR_STATUS = object()
+
+
 def confirm_or_append_stage_entry(
     stage_states: dict,
     stage: str,
     now: datetime | None = None,
     pr_number: int | None = None,
+    prior_status: object = _NO_PRIOR_STATUS,
 ) -> dict:
     """Record that ``stage`` was entered, without double-counting the router.
 
@@ -1829,6 +1833,17 @@ def confirm_or_append_stage_entry(
     ``last_dispatched_skill`` (``history[-1]["skill"]``) consequently becomes
     *true* on the bypass path rather than naming whichever skill the router last
     saw. That is the fix, not a side effect: rows 8b/8d gate on it.
+
+    ``prior_status`` is the entering stage's status **before** it flipped to
+    ``in_progress``, and it is what keeps the two writers comparable. The router
+    records its slot before invoking, so its snapshot sees the pre-entry status;
+    an appended record is written after ``_activate_stage`` has already flipped
+    the stage, so without this it would see ``in_progress`` instead. G4 breaks
+    its walk on the first snapshot mismatch, so mixing the two observation
+    points makes a loop that alternates router-driven and bypass entries count
+    *lower* than it did before this feature existed -- adding the correct record
+    would lower the streak. Callers that know the pre-entry status pass it; the
+    snapshot is then built against that value, matching the router.
     """
     history = stage_states.get("_sdlc_dispatches")
     if not isinstance(history, list):
@@ -1869,7 +1884,28 @@ def confirm_or_append_stage_entry(
                 pr_number = snapshot["pr_number"]
                 break
 
-    return record_dispatch(stage_states, skill=skill, now=now, pr_number=pr_number, confirmed=True)
+    # Build the snapshot against the stage's PRE-entry status, so this record is
+    # comparable with a router slot recorded before the same transition.
+    if prior_status is _NO_PRIOR_STATUS:
+        return record_dispatch(
+            stage_states, skill=skill, now=now, pr_number=pr_number, confirmed=True
+        )
+
+    had_key = stage in stage_states
+    current = stage_states.get(stage)
+    if prior_status is None:
+        stage_states.pop(stage, None)
+    else:
+        stage_states[stage] = prior_status
+    try:
+        return record_dispatch(
+            stage_states, skill=skill, now=now, pr_number=pr_number, confirmed=True
+        )
+    finally:
+        if had_key:
+            stage_states[stage] = current
+        else:
+            stage_states.pop(stage, None)
 
 
 def record_dispatch(
--- a/agent/sdlc_router.py
+++ b/agent/sdlc_router.py
@@ -1256,7 +1256,8 @@ def _rule_patch_applied_after_review(stage_states: dict, meta: dict, context: di
     deliberate: an approval recorded before the latest patch does not describe
     the code that patch produced, and advancing on it would merge code no review
     ever saw. It converges -- the re-review records a verdict newer than the
-    patch dispatch, which row 9 then owns -- and G4 bounds it otherwise. See
+    patch dispatch, which row 9 then owns. If it does not converge, nothing
+    stops it; see the unbounded-loop note at the end of this docstring. See
     ``TestStaleApprovedIsReReviewedNotAdvanced``.
 
     Disjointness from rows 8c/8d/8e (which have no step-aside against disjunct
@@ -1272,9 +1273,16 @@ def _rule_patch_applied_after_review(stage_states: dict, meta: dict, context: di
 
     No defensive step-asides are added on the strength of that proof.
 
-    Loop-bound by G4 (``same_stage_dispatch_count``): a verdict that stays
-    permanently stale escalates to a human rather than spinning on
-    ``/do-pr-review``.
+    NOT loop-bounded, despite what this docstring claimed until #2730. The
+    8 -> 8b loop alternates ``/do-patch`` with ``/do-pr-review``, and
+    ``compute_same_stage_count`` requires consecutive entries to share BOTH the
+    skill and the ``stage_snapshot``. The alternation fails the first test, and
+    ``_patch_cycle_count`` -- which ``build_stage_snapshot`` includes and
+    ``complete_stage("PATCH")`` increments every cycle -- fails the second. G2
+    caps critique cycles only; there is no patch-cycle equivalent. So a verdict
+    that stays permanently stale spins on ``/do-pr-review`` without escalating.
+    Tracked on #2801, which carries the measurements and the reason the obvious
+    one-line fix does not work.
     """
     if not meta.get("pr_number"):
         return False
--- a/agent/sdlc_router.py
+++ b/agent/sdlc_router.py
@@ -1826,23 +1826,40 @@ def confirm_or_append_stage_entry(
     if not isinstance(history, list):
         history = []
 
-    # Newest-first: upgrade the most recent unconfirmed slot for this stage.
-    for entry in reversed(history):
-        if not isinstance(entry, dict) or entry.get("stage") != stage:
-            continue
-        if entry.get("confirmed") is False:
-            entry["confirmed"] = True
-            stage_states["_sdlc_dispatches"] = history
-            return stage_states
-        # The newest record for this stage is already confirmed, so this is a
-        # fresh entry into a stage that ran before. Fall through and append.
-        break
+    # ONLY the newest record can be this entry's router slot. The router records
+    # immediately before invoking the sub-skill, so nothing else can land in
+    # between. Scanning further back would let a stale slot -- a dispatch that
+    # was recorded and then never started -- absorb an unrelated later entry,
+    # dropping that entry's record entirely and leaving `last_dispatched_skill`
+    # naming the wrong skill: exactly the defect this function exists to fix.
+    last = history[-1] if history else None
+    if isinstance(last, dict) and last.get("stage") == stage and last.get("confirmed") is False:
+        last["confirmed"] = True
+        # Re-stamp the time: `_latest_dispatch_at` (rows 8 and 2b) must report
+        # when the stage actually started, not when the router queued it.
+        last["at"] = (now or datetime.now(UTC)).isoformat()
+        stage_states["_sdlc_dispatches"] = history
+        return stage_states
 
     skill = STAGE_TO_SKILL.get(stage)
     if skill is None:
         # An unknown stage has no skill to attribute the entry to. Recording a
         # skill-less record would break every consumer of history[-1]["skill"].
         return stage_states
+
+    # Inherit the PR number from the newest record that carries one. The bypass
+    # path has no `_meta` to read it from, and `pr_number` is IN the snapshot
+    # G4 compares -- so recording None beside a router record's real value
+    # breaks the streak at every router/bypass boundary, making the very records
+    # this function adds uncountable.
+    if pr_number is None:
+        for entry in reversed(history):
+            if not isinstance(entry, dict):
+                continue
+            snapshot = entry.get("stage_snapshot")
+            if isinstance(snapshot, dict) and snapshot.get("pr_number") is not None:
+                pr_number = snapshot["pr_number"]
+                break
 
     return record_dispatch(stage_states, skill=skill, now=now, pr_number=pr_number, confirmed=True)
 
--- a/agent/sdlc_router.py
+++ b/agent/sdlc_router.py
@@ -1777,11 +1777,82 @@ def decide_next_dispatch(
     return primary
 
 
+def stage_for_skill(skill: str | None) -> str | None:
+    """Return the pipeline stage a ``/do-*`` skill executes, or ``None``.
+
+    Inverse of ``agent.pipeline_graph.STAGE_TO_SKILL``, derived from it rather
+    than hand-maintained so the two cannot drift (#2730). ``STAGE_TO_SKILL`` is
+    1:1, so the inversion is unambiguous.
+    """
+    for stage, mapped in STAGE_TO_SKILL.items():
+        if mapped == skill:
+            return stage
+    return None
+
+
+def confirm_or_append_stage_entry(
+    stage_states: dict,
+    stage: str,
+    now: datetime | None = None,
+    pr_number: int | None = None,
+) -> dict:
+    """Record that ``stage`` was entered, without double-counting the router.
+
+    Issue #2730. Two different actors observe a stage starting, and before this
+    only one of them recorded it:
+
+    - the router calls ``record_dispatch`` *before* invoking a sub-skill, which
+      is the only signal available if the skill dies before it starts;
+    - the stage itself transitions to ``in_progress``, which happens on EVERY
+      entry path including the ones that never reach the router (a skill-to-skill
+      chain such as ``/do-build`` → ``/do-patch``, and the PreToolUse hook, which
+      is the dominant marker writer on the bridge).
+
+    Recording in both places naively would double every router-driven dispatch
+    and halve G4's effective threshold. Instead the router's record is an
+    **unconfirmed slot** (``confirmed: False``) and this function upgrades it in
+    place. When no unconfirmed slot exists the stage was reached without the
+    router, and a confirmed record is appended.
+
+    The result is exactly one record per stage entry, from either path, while a
+    genuine second entry still appends a second record -- which is the whole G4
+    signal and must not be deduplicated away.
+
+    ``last_dispatched_skill`` (``history[-1]["skill"]``) consequently becomes
+    *true* on the bypass path rather than naming whichever skill the router last
+    saw. That is the fix, not a side effect: rows 8b/8d gate on it.
+    """
+    history = stage_states.get("_sdlc_dispatches")
+    if not isinstance(history, list):
+        history = []
+
+    # Newest-first: upgrade the most recent unconfirmed slot for this stage.
+    for entry in reversed(history):
+        if not isinstance(entry, dict) or entry.get("stage") != stage:
+            continue
+        if entry.get("confirmed") is False:
+            entry["confirmed"] = True
+            stage_states["_sdlc_dispatches"] = history
+            return stage_states
+        # The newest record for this stage is already confirmed, so this is a
+        # fresh entry into a stage that ran before. Fall through and append.
+        break
+
+    skill = STAGE_TO_SKILL.get(stage)
+    if skill is None:
+        # An unknown stage has no skill to attribute the entry to. Recording a
+        # skill-less record would break every consumer of history[-1]["skill"].
+        return stage_states
+
+    return record_dispatch(stage_states, skill=skill, now=now, pr_number=pr_number, confirmed=True)
+
+
 def record_dispatch(
     stage_states: dict,
     skill: str,
     now: datetime | None = None,
     pr_number: int | None = None,
+    confirmed: bool = True,
 ) -> dict:
     """Append a dispatch record to ``stage_states._sdlc_dispatches``.
 
@@ -1823,6 +1894,15 @@ def record_dispatch(
             "skill": skill,
             "at": timestamp,
             "stage_snapshot": snapshot,
+            # #2730: the stage this dispatch executes, so the stage-entry
+            # upgrade path can find its slot without re-deriving it from the
+            # skill string at read time. NOT in `stage_snapshot` -- G4 compares
+            # snapshots, and a per-dispatch-varying field there kills
+            # oscillation detection outright.
+            "stage": stage_for_skill(skill),
+            # False marks a router slot awaiting confirmation that the stage
+            # actually started. See confirm_or_append_stage_entry.
+            "confirmed": confirmed,
         }
     )
     # FIFO-evict oldest entries to bound the list.
--- a/agent/sdlc_router.py
+++ b/agent/sdlc_router.py
@@ -1849,12 +1849,19 @@ def confirm_or_append_stage_entry(
     if not isinstance(history, list):
         history = []
 
-    # ONLY the newest record can be this entry's router slot. The router records
-    # immediately before invoking the sub-skill, so nothing else can land in
-    # between. Scanning further back would let a stale slot -- a dispatch that
-    # was recorded and then never started -- absorb an unrelated later entry,
-    # dropping that entry's record entirely and leaving `last_dispatched_skill`
-    # naming the wrong skill: exactly the defect this function exists to fix.
+    # ONLY the newest record is eligible as this entry's router slot. Scanning
+    # further back would let a stale slot -- a dispatch recorded and then never
+    # started -- absorb an unrelated later entry, dropping that entry's record
+    # and leaving `last_dispatched_skill` naming the wrong skill: exactly the
+    # defect this function exists to fix.
+    #
+    # This narrows the hole rather than closing it. If the sub-skill dies before
+    # `start_stage`, its slot IS the newest record, and a later entry into the
+    # same stage upgrades that dead slot instead of appending -- collapsing two
+    # dispatch events into one record and under-counting G4 by one. Closing it
+    # needs an age-out against `last["at"]`, which needs a defensible staleness
+    # threshold; the under-count is the safe direction (a missed escalation, not
+    # a false one), so it is left open deliberately rather than guessed at.
     last = history[-1] if history else None
     if isinstance(last, dict) and last.get("stage") == stage and last.get("confirmed") is False:
         last["confirmed"] = True
@@ -1885,27 +1892,26 @@ def confirm_or_append_stage_entry(
                 break
 
     # Build the snapshot against the stage's PRE-entry status, so this record is
-    # comparable with a router slot recorded before the same transition.
-    if prior_status is _NO_PRIOR_STATUS:
-        return record_dispatch(
-            stage_states, skill=skill, now=now, pr_number=pr_number, confirmed=True
-        )
-
-    had_key = stage in stage_states
-    current = stage_states.get(stage)
-    if prior_status is None:
-        stage_states.pop(stage, None)
-    else:
-        stage_states[stage] = prior_status
-    try:
-        return record_dispatch(
-            stage_states, skill=skill, now=now, pr_number=pr_number, confirmed=True
-        )
-    finally:
-        if had_key:
-            stage_states[stage] = current
+    # comparable with a router slot recorded before the same transition. The
+    # override is an explicit view rather than a mutate-and-restore of the
+    # caller's dict, so the invariant stays local to this function instead of
+    # depending on `build_stage_snapshot` copying eagerly two calls away.
+    snapshot_states = None
+    if prior_status is not _NO_PRIOR_STATUS:
+        snapshot_states = dict(stage_states)
+        if prior_status is None:
+            snapshot_states.pop(stage, None)
         else:
-            stage_states.pop(stage, None)
+            snapshot_states[stage] = prior_status
+
+    return record_dispatch(
+        stage_states,
+        skill=skill,
+        now=now,
+        pr_number=pr_number,
+        confirmed=True,
+        snapshot_states=snapshot_states,
+    )
 
 
 def record_dispatch(
@@ -1914,6 +1920,7 @@ def record_dispatch(
     now: datetime | None = None,
     pr_number: int | None = None,
     confirmed: bool = True,
+    snapshot_states: dict | None = None,
 ) -> dict:
     """Append a dispatch record to ``stage_states._sdlc_dispatches``.
 
@@ -1937,14 +1944,27 @@ def record_dispatch(
             ``pr_number`` is ``None`` — the explicit argument is the sole
             provenance (``sdlc_stage_query`` resolves the PR number from the
             ``AgentSession.pr_number`` field into ``_meta.pr_number``).
+        confirmed: ``False`` marks the record a **router slot** — a dispatch the
+            router recorded immediately before invoking a sub-skill, which the
+            stage entry later upgrades in place rather than duplicating. See
+            ``confirm_or_append_stage_entry`` for the protocol. ``True`` (the
+            default) records a dispatch that is already known to have started.
+        snapshot_states: Optional alternative source for the ``stage_snapshot``
+            projection. The record is still appended to ``stage_states``; only
+            the snapshot is built from this. Callers use it to snapshot a stage
+            transition from a specific side without mutating the live dict —
+            see ``confirm_or_append_stage_entry``'s ``prior_status``.
 
     Returns:
-        The mutated stage_states dict.
+        The mutated stage_states dict. The appended record carries ``skill``,
+        ``at``, ``stage_snapshot``, ``stage`` and ``confirmed``; the last two
+        are deliberately outside the snapshot (see below).
     """
     timestamp = (now or datetime.now(UTC)).isoformat()
     # Build a snapshot from a stage_states view that EXCLUDES the history
     # list itself, otherwise the counter would never match across invocations.
-    view = {k: v for k, v in stage_states.items() if k != "_sdlc_dispatches"}
+    source = stage_states if snapshot_states is None else snapshot_states
+    view = {k: v for k, v in source.items() if k != "_sdlc_dispatches"}
     snapshot = build_stage_snapshot(view, meta={"pr_number": pr_number})
 
     history = stage_states.setdefault("_sdlc_dispatches", [])
--- a/docs/features/sdlc-router-oscillation-guard.md
+++ b/docs/features/sdlc-router-oscillation-guard.md
@@ -239,6 +239,21 @@ that actually ran rather than whichever one the router last saw — after a
 that value; G1, G3 and G7 also read it, and rows 8 and 2b read
 `_latest_dispatch_at`.
 
+A sixth consumer reads the history for a different purpose:
+`tools/sdlc_stage_marker.py`'s retroactive-skip check refuses to mark a stage
+`skipped` once a dispatch for it exists. Because every stage entry now records
+one, a `/do-plan` that entered PLAN and then aborted to ask a question is no
+longer retroactively skippable. That is the guard's stated contract — the stage
+*was* reached — and it errs toward refusing a skip rather than granting a false
+one.
+
+Records appended by the stage entry carry no `run_id`. That annotation is
+applied only on the router path (`tools/sdlc_dispatch.py`), because
+`PipelineStateMachine` reaches its store as a bare `PipelineLedger` on the
+`for_issue` path while `active_run_id` is a field of `AgentSession`. A
+hook-driven lane's history is therefore `run_id`-sparse, and a missing `run_id`
+means "not recorded", never "a different run".
+
 The record is applied inside `_save()` **after** the preserved-metadata merge.
 `_save()` re-reads `_sdlc_dispatches` from the backing store and merges that
 copy over anything in memory, so a record written before the merge is silently
@@ -248,6 +263,25 @@ same single write.
 `stage` and `confirmed` live on the record and **never** in `stage_snapshot` —
 G4 compares snapshots, and a per-dispatch-varying field there would stop every
 comparison from ever matching, silently disabling oscillation detection.
+
+**Both writers must snapshot the same side of the stage transition.** The router
+records its slot before invoking, so its snapshot sees the entering stage's
+**pre-entry** status. An appended record is written from `_save()` *after*
+`_activate_stage` has already flipped the stage to `in_progress`, so left alone
+it would snapshot the other side. Since `compute_same_stage_count` breaks its
+walk on the first snapshot mismatch, mixing the two observation points makes a
+loop that alternates router-driven and bypass entries count **lower than it did
+before this feature existed** — adding the correct record would *lower* the
+streak. `_activate_stage` therefore carries the pre-entry status alongside the
+pending stage, and `confirm_or_append_stage_entry` builds the appended record's
+snapshot against that value via `record_dispatch`'s `snapshot_states` override.
+Measured on a three-cycle mixed-path `/do-patch` loop: 1 without the override, 6
+with it. `pr_number` has the same hazard and the same treatment — bypass records
+inherit it from the newest record carrying one.
+
+This invariant is the correctness keystone of the whole feature: a change that
+drops it does not fail loudly, it silently makes G4 count worse than not having
+the records at all.
 
 The `stage_snapshot` projection is deliberately narrow to prevent spurious
 churn:
--- a/docs/features/sdlc-router-oscillation-guard.md
+++ b/docs/features/sdlc-router-oscillation-guard.md
@@ -211,8 +211,43 @@ Both guards now check `meta.get("pr_number")` first and return `None` (step asid
 ### G4 state machine
 
 `stage_states._sdlc_dispatches` is a bounded FIFO list (max 10 entries) of
-`{skill, at, stage_snapshot}` records. The record is written **before** the
-sub-skill launches (so oscillation on crashing skills is still detected).
+`{skill, at, stage_snapshot, stage, confirmed}` records.
+
+**Two writers, one record per stage entry.** The router records **before** the
+sub-skill launches, so oscillation on a skill that crashes before it starts is
+still detected. That record is an *unconfirmed slot* (`confirmed: false`).
+`PipelineStateMachine._activate_stage` — the single point every `start_stage`
+branch funnels through — upgrades that slot in place when the stage actually
+starts, and appends its own confirmed record when no slot exists.
+
+The second writer is what makes the history unbypassable. Only the router used
+to write it, while stage markers were written by the stage itself, so any stage
+reached without the router left a marker and no ledger entry and G4 had nothing
+to count. Two such paths exist: a skill-to-skill chain (`/do-build` routes
+failures straight to `/do-patch`), and the PreToolUse hook, which calls
+`start_stage()` on every `Skill(do-*)` call and never touches `write_marker` —
+on the bridge that hook is how markers get written at all.
+
+The upsert slot is what keeps the two writers from double-counting a single
+dispatch, which would halve G4's effective threshold. A genuine **re-entry**
+still appends a second record: that is the entire G4 signal, and deduplicating
+it would disarm the guard.
+
+`last_dispatched_skill` is `history[-1]["skill"]`, so it now names the skill
+that actually ran rather than whichever one the router last saw — after a
+`/do-build` → `/do-patch` chain it reads `/do-patch`. Rows 8b and 8d gate on
+that value; G1, G3 and G7 also read it, and rows 8 and 2b read
+`_latest_dispatch_at`.
+
+The record is applied inside `_save()` **after** the preserved-metadata merge.
+`_save()` re-reads `_sdlc_dispatches` from the backing store and merges that
+copy over anything in memory, so a record written before the merge is silently
+discarded; applying it after means it sees the freshest history and rides the
+same single write.
+
+`stage` and `confirmed` live on the record and **never** in `stage_snapshot` —
+G4 compares snapshots, and a per-dispatch-varying field there would stop every
+comparison from ever matching, silently disabling oscillation detection.
 
 The `stage_snapshot` projection is deliberately narrow to prevent spurious
 churn:
--- a/docs/features/sdlc-stage-tracking.md
+++ b/docs/features/sdlc-stage-tracking.md
@@ -17,6 +17,21 @@ Both paths share the exact same `StageStates` Pydantic validation and the `_load
 
 **Write authority.** A caller may only write the issue-keyed ledger while it holds the per-issue run_id lease (`models/session_lifecycle.py::touch_issue_lock`, see [`docs/features/sdlc-issue-ownership-lock.md`](sdlc-issue-ownership-lock.md)). The lease also carries the pinned `target_repo` component of the ledger key, resolved once at lease-acquire time — writers never re-resolve it per call. Takeover of an issue is simply acquiring that same lease; the ledger itself never moves, because it never lived on either session.
 
+
+### Stage entry records a dispatch
+
+Starting a stage records an entry in `_sdlc_dispatches`, not just a marker.
+`PipelineStateMachine._activate_stage` is the single point every `start_stage`
+branch and every actor funnels through — the skill bodies via `sdlc-tool
+stage-marker`, the PreToolUse hook (which calls `start_stage()` directly and
+never touches `write_marker`), and `/do-sdlc`'s backfills.
+
+The router still records before invoking a sub-skill, as an *unconfirmed slot*
+that the stage entry upgrades in place; a stage reached without the router
+appends its own confirmed record. One record per stage entry from either path,
+and a genuine re-entry appends another, which is what G4 counts. See
+[SDLC Router Oscillation Guard](sdlc-router-oscillation-guard.md#g4-state-machine).
+
 ## How Stage State Is Written
 
 Two complementary paths write stage markers:
--- a/tests/unit/test_pipeline_state_machine.py
+++ b/tests/unit/test_pipeline_state_machine.py
@@ -1897,3 +1897,116 @@ class TestSkippedIsSettled:
         """StageStates coerces unknown statuses to `pending`; `skipped` is known."""
         validated = StageStates.from_dict({"CRITIQUE": "skipped"}).to_dict()
         assert validated["CRITIQUE"] == "skipped"
+
+
+# ---------------------------------------------------------------------------
+# Issue #2730: starting a stage records a dispatch, through the real _save().
+#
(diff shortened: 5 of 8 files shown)


valorengels:
## Review (Judge code-quality): Changes Requested

`blockers=1 tech_debt=2 nits=4`

### Blocker

**File:** `agent/pipeline_state.py:645` + `agent/sdlc_router.py:1844-1849`
**Code:** `self._pending_stage_entry = stage` / `self._save()` (after `self.states[stage] = "in_progress"`), against the upgrade path which touches only `confirmed` and `at`.
**Issue:** The two writers observe the stage at different points. The router builds its snapshot *before* the stage flips (`stages[stage]` reads `failed`/`completed`); the append path builds a fresh snapshot from `data`, where `_activate_stage` has already set `in_progress`. `compute_same_stage_count` breaks its walk on the first snapshot mismatch, so the records are systematically non-comparable across a router/bypass boundary. This is the identical hazard the code explicitly defends against for `pr_number` at `agent/sdlc_router.py:1858-1862` ("`pr_number` is IN the snapshot G4 compares -- so recording None beside a router record's real value breaks the streak at every router/bypass boundary, making the very records this function adds uncountable"). The `stages` instance is unhandled.
**Severity:** blocker
**Fix:** Have `_activate_stage` stash the stage's pre-transition status and have `_apply_pending_stage_entry` build the snapshot from a view carrying it, so both writers describe the same side of the transition.

### Tech Debt

**File:** `tests/unit/test_sdlc_router_oscillation.py:1036-1042`
**Code:** `states: dict = {}` in `test_a_bypass_record_inherits_the_pr_number_so_g4_can_count_it`
**Issue:** The docstring claims countability and asserts `count == 2`, but the fixture carries no stage statuses, so the one field that diverges between the writers (`stages`) is empty on both sides and the mismatch cannot surface. Every test in `TestStageEntryUpsert` uses the same bare-dict shape, so no test in the class exercises snapshot comparability across the two writers — which is why the blocker above shipped green.
**Severity:** tech_debt
**Fix:** Give the countability tests realistic stage statuses and drive them through `PipelineStateMachine`, as `TestStageEntryRecordsDispatch` already does.

**File:** `agent/sdlc_router.py:1836-1844`
**Code:** `last = history[-1] if history else None` / `if isinstance(last, dict) and last.get("stage") == stage and last.get("confirmed") is False:`
**Issue:** The only-newest rule does not close the hole its own comment names. When the stale slot *is* the newest record — router records a `/do-patch` slot, the sub-skill dies before `start_stage`, then a `/do-build`→`/do-patch` chain enters PATCH — two distinct dispatch events still collapse into one record. `test_a_stale_slot_for_another_stage_does_not_absorb_this_entry` covers a different stage; `test_only_the_newest_unconfirmed_slot_is_upgraded` has an intervening record. Neither reaches this case.
**Severity:** tech_debt
**Fix:** Age out the slot against `last["at"]`, or accept it and reword the comment, which currently reads as though the case is handled.

### Nits

- **`agent/sdlc_router.py:1894-1904`** — `record_dispatch`'s `Args:` block documents `stage_states`, `skill`, `now`, `pr_number` and stops. `confirmed` carries the whole slot protocol and is undocumented.
- **`agent/pipeline_state.py:586`** — `stage = getattr(self, "_pending_stage_entry", None)`. The attribute is never initialized in `__init__` or `for_issue()`. `_load_state` states the standard verbatim: "Defaults first so every construction path ... lands with the full attribute set." Initialize it and read the attribute directly.
- **`agent/pipeline_state.py:594`** — `except Exception as e:  # pragma: no cover`. The branch IS covered by `test_a_failing_dispatch_record_never_blocks_the_marker_write` (the in-`try` import resolves the patched attribute at call time). The pragma now hides a covered line.
- **`agent/pipeline_state.py:506-513`** — `_save()`'s docstring still names "the dispatch recorder in `agent.sdlc_router.record_dispatch`" as an external writer it protects against. As of `:555`, `_save()` is itself one of the dispatch writers.

### Verified clean
`ruff format --check` passes; 236 tests pass in the two touched files; `stage`/`confirmed` correctly stay out of `stage_snapshot`; `STAGE_TO_SKILL` is genuinely 1:1 so `stage_for_skill`'s inversion is sound; `_pending_stage_entry` is cleared before the `try` so it cannot leak into a later unrelated `_save()`; both feature docs are accurate.

valorengels:

`blockers=0 tech_debt=3 nits=2`

### Checks that came back clean (verified, not assumed)

- **Store coherence.** Both writers land in the same store. The router (`tools/sdlc_dispatch.py::_cli_record`) writes only to `PipelineLedger.get_or_create(target_repo, issue_number)` and refuses outright on an unresolved lease; the hooks resolve through `agent/pipeline_state.py:1417 resolve_pipeline_state_machine`, which returns `for_issue(...)` on the same lease. When the lease does not resolve the router records nothing at all, so the session-keyed fallback cannot double-count against a ledger slot.
- **G4 false-positive inflation.** No healthy path reaches 3. `start_stage` no-ops while `in_progress` (`agent/pipeline_state.py:938-940`), so re-entry requires an intervening `complete_stage`/`fail_stage`; for PATCH that increments `_patch_cycle_count`, which is inside `build_stage_snapshot` (`agent/sdlc_router.py:226`).
- **FIFO eviction.** `MAX_DISPATCH_HISTORY = 10` unchanged; router-driven dispatches still produce exactly one record, so retained depth does not shrink on the router path.

### Tech Debt

**File:** `agent/pipeline_state.py:645`
**Code:** `self._pending_stage_entry = stage` / `self._save()` — with `self.states[stage] = "in_progress"` set on the line above.
**Issue:** An appended bypass record's `stage_snapshot["stages"][stage]` is `"in_progress"`; the upgrade path never rewrites the snapshot (`agent/sdlc_router.py:1844-1849` touches only `confirmed` and `at`), so an upgraded router record keeps the pre-activation value. `compute_same_stage_count` breaks at `agent/sdlc_router.py:1984` (`if canonical_snapshot(entry_snapshot) != last_snapshot_canonical: break`), so a mixed-path streak resets at every path boundary. Under-detection, never a false escalation — but it materially qualifies the PR's "that is the fix" framing.
**Severity:** tech_debt
**Fix:** Take the snapshot in `_activate_stage` before mutating `self.states[stage]`, or have the upgrade path re-stamp `stage_snapshot` the way it re-stamps `at`. Pin it with a router → bypass → router test asserting `compute_same_stage_count` returns 3.

**File:** `tools/sdlc_stage_marker.py:390`
**Code:** `history = raw.get("_sdlc_dispatches")` / `if any(isinstance(d, dict) and d.get("skill") == skill for d in history): return ("STAGE_RAN_NOT_SKIPPABLE", ...)` with `_SKIP_STAGE_SKILL = {"PLAN": "/do-plan", "CRITIQUE": "/do-plan-critique"}` (`:143`).
**Issue:** A sixth `_sdlc_dispatches` consumer absent from the PR body's enumeration of five, and its behavior does change. `/do-plan` or `/do-plan-critique` invoked outside the router — including an invocation that aborts immediately to ask a question — now writes a record via the PreToolUse hook, permanently returning `STAGE_RAN_NOT_SKIPPABLE` for a stage that previously stayed retroactively skippable. Arguably more correct, but it is an unanalyzed operational change to an escape hatch.
**Severity:** tech_debt
**Fix:** Add it to the consumer enumeration in the PR body and `docs/features/sdlc-router-oscillation-guard.md`, and state that an aborted stage-skill invocation now burns the retroactive skip.

**File:** `tools/sdlc_dispatch.py:129`
**Code:** `history[-1]["run_id"] = run_id`
**Issue:** The #2003 run-identity annotation is applied only inside `record_dispatch_for_ledger`. `confirm_or_append_stage_entry` (`agent/sdlc_router.py:1872`) calls `record_dispatch(...)` directly and never sets `run_id`, so on a hook-driven lane `sdlc-tool dispatch get` returns a history where some records carry `run_id` and some do not — and the ones that do not are exactly the records this PR exists to create. No current reader consumes it, so this is observability debt, not a correctness break.
**Severity:** tech_debt
**Fix:** Resolve the run identity in `_apply_pending_stage_entry` and pass it through, or document in the record schema that `run_id` is router-path-only.

### Nits

- **`agent/sdlc_router.py:1844`** — `last.get("confirmed") is False`. Pre-merge records have no `confirmed` key, so the first post-merge stage entry appends instead of upgrading: one duplicate record per lane in flight at merge time, plus a mixed-version window until every process is cycled. No migration needed, but the merge note should call for `/update` + `./scripts/valor-service.sh restart` before the next lane starts, and mention the transitional duplicate so the next reader does not chase it as a bug.
- **`agent/sdlc_router.py:1880`** — `confirmed: bool = True,` is absent from the `Args:` block at `:1896-1905`.

valorengels:
## Review: Changes Requested

Two-judge consensus (`code-quality`, `risk`), `rule=any-blocker-wins`. The `risk` judge returned APPROVED with 3 tech_debt; the `code-quality` judge returned CHANGES REQUESTED with 1 blocker. Both judges independently, and the aggregator independently, landed on the same defect: **the two writers snapshot the stage at different points in the transition, so the records this PR adds are not comparable to the records the router already writes.** I reproduced it against `main` rather than taking any judge's word for it.

The design is genuinely good — the upsert-slot protocol is the right answer to the double-count problem, the placement inside `_save()` after the merge is correct and well-argued, and keeping `stage`/`confirmed` out of `stage_snapshot` is exactly right. The blocker is one observation-point mismatch away from working.

Second blocker is process, not code: the plan document forbids this PR.

## Rubric

- [ ] **1. Plan vs. implementation match** — fail — *the plan's `[FENCE-BLOCKED]` No-Go excludes precisely this change; none of the plan's two in-scope workstreams ship. See Blocker 2.*
- [x] **2. New code quality** — pass — *clear naming, load-bearing docstrings, deferred import avoids a cycle, `stage_for_skill` derives from the canonical map rather than adding a fourth. Nits below.*
- [ ] **3. Test coverage** — fail — *12 new tests, counts asserted in both directions, but every `TestStageEntryUpsert` fixture is a bare `states: dict = {}`, so nothing exercises snapshot comparability across the two writers. That is why Blocker 1 shipped green.*
- [ ] **4. Regression risk to existing callers** — fail — *Blocker 1 lowers G4's streak below what `main` produces on a mixed-path loop. `tools/sdlc_stage_marker.py:390` is an unenumerated sixth consumer whose behavior changes.*
- [x] **5. Data integrity** — pass — *two additive JSON fields; legacy records read coherently via `.get`; no migration needed. Transition artifact noted in Miscellaneous.*
- [x] **6. Security** — n/a — *no new subprocess, eval, deserialization, or secret handling.*
- [ ] **7. Documentation accuracy** — fail — *`docs/features/sdlc-router-oscillation-guard.md` and `docs/features/sdlc-stage-tracking.md` assert the records are now what G4 counts; across a router/bypass boundary they are not.*
- [ ] **8. PR body accuracy** — fail — *"`ruff check` and `ruff format --check` clean" is not reproducible; "694 tests pass" measured 699 here; the consumer enumeration is five of six.*
- [x] **9. Disclosed deferrals** — pass — *#2801 and #2491 are OPEN; the #2731 disposition carries its rationale on the issue.*
- [x] **10. Follow-up claims verified** — pass — *#2801 OPEN, #2491 OPEN, #2711 CLOSED as cited.*

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — FAIL — *the plan's Success Criteria describe the G4 counter fix and the #2731 liveness gate; neither ships. "The dispatch record gains no fields" is directly contradicted.*
- **2. No-Gos from plan — none violated** — FAIL — *`[FENCE-BLOCKED]` No-Go at `docs/plans/ledger-integrity.md:151` names `agent/pipeline_state.py` as outside the fence; this PR edits it.*
- **3. New `except Exception` blocks — each has logger/raise/swallow-ok** — PASS — *`agent/pipeline_state.py:594` logs at debug with the stage and the exception; swallow is justified and tested.*
- **4. New integration tests — exercise serialization boundary (not in-memory only)** — PASS — *`TestStageEntryRecordsDispatch` deliberately goes through the real `_save()` and reads back `json.loads(session.stage_states)`.*
- **5. Plan internal consistency — spike findings match task steps** — FAIL — *the plan's Round 2 disposition escalated to the lead for a fence/scope decision and the doc was never updated with the ruling.*
- **7. New public APIs — docstrings present** — PASS — *`stage_for_skill` and `confirm_or_append_stage_entry` are both thoroughly documented; `record_dispatch`'s `Args` block is stale (nit).*
- **8. Breaking changes — migration path documented** — PASS — *additive fields, legacy records handled; the transition window is a Miscellaneous advisory.*
- **9. Tests added for new behavior** — PASS — *12 new tests across both files.*
- **10. Tests cover the failure path (not just happy path)** — PASS — *unknown stage, raising recorder, already-`in_progress` no-op, stale slot, snapshot key-set pin.*
- **11. UI changes (if any) — screenshot captured** — N/A — *no UI files in the diff; the visual-proof gate is a no-op.*
- **12. Docs updated for user-facing changes** — PASS — *two feature docs updated, though see rubric item 7.*

### Blockers

- [ ] **`agent/pipeline_state.py:645` + `agent/sdlc_router.py:1844-1849`** — `self._pending_stage_entry = stage; self._save()` (after `self.states[stage] = "in_progress"`), against an upgrade path that touches only `confirmed` and `at` — **the two writers snapshot opposite sides of the stage transition, and the added record lowers G4's count below `main`'s.**

  The router writes its slot *before* the stage flips, so the slot's `stage_snapshot["stages"][stage]` reads `failed`/`completed`. `_activate_stage` flips the stage to `in_progress` *before* `_save()` builds `data`, so an appended bypass record reads `in_progress`. `compute_same_stage_count` breaks its walk on the first snapshot mismatch (`agent/sdlc_router.py:1984`), so the streak resets at every router/bypass boundary.

  This is the identical hazard the code already defends against one field over, at `agent/sdlc_router.py:1858-1862`: *"`pr_number` is IN the snapshot G4 compares -- so recording None beside a router record's real value breaks the streak at every router/bypass boundary, making the very records this function adds uncountable, which is worse than the missing records it fixes."* The `stages` instance of that hazard is unhandled.

  Reproduced on a three-cycle `/do-patch` loop (router → bypass chain → router), the motivating scenario from the PR body:

  | | records written | `compute_same_stage_count` |
  |---|---|---|
  | `main` | 2 | **2** |
  | this branch | 3 | **1** |

  Adding the third, correct record *reduced* the streak from 2 to 1. On the loop this PR exists to make visible, G4 now fires later than it did before the change. By the author's own stated bar in the quoted comment, that is worse than the missing records it fixes.

  **Fix:** have `_activate_stage` stash the stage's pre-transition status and have `_apply_pending_stage_entry` build the snapshot from a view carrying it, so both writers describe the same side of the transition. Pin with a router → bypass → router test asserting the count reaches 3.

- [ ] **`docs/plans/ledger-integrity.md:151`** — `- **[FENCE-BLOCKED] #2730's record-completeness half.** spike-2 established the fix must live in \`PipelineStateMachine.start_stage()\` (\`agent/pipeline_state.py:870\`) ... Outside this lane's fence. **#2730 is only partially closed by this PR and the PR body must say so.**` — **the plan document forbids exactly what this PR ships, and was not updated.**

  `git diff main...HEAD -- docs/plans/` is empty. The plan is still `status: Needs Revision`, its Round 2 disposition still reads *"Escalated to the lead for a fence/scope decision rather than revised a third time,"* and the ruling the PR body acts on is recorded nowhere but the PR body. The consequences are concrete, not cosmetic:

  - The `[FENCE-BLOCKED]` No-Go excludes `agent/pipeline_state.py`, which this PR edits.
  - Success Criterion *"The dispatch record gains no fields"* is contradicted — `stage` and `confirmed` are added to every record.
  - Success Criterion *"The PR body states that #2730's record-completeness half is deferred, and why"* is inverted; the deferred half is what shipped.
  - The `## Verification` table, `## Test Impact`, `## Step by Step Tasks`, and eight of eleven Success Criteria all describe the G4 counter fix and the #2731 liveness gate, neither of which is in the diff. There is no runnable acceptance gate for what actually shipped.

  **Fix:** revise `docs/plans/ledger-integrity.md` on `main` (per the plans-on-main convention) to record the lead's fence/scope ruling, and replace the No-Gos, Success Criteria, Test Impact, and Verification table with the shipped scope. The PR body's reasoning is sound and mostly transplants directly.

### Tech Debt

- [ ] **`tests/unit/test_sdlc_router_oscillation.py:1036-1042`** — `states: dict = {}` in `test_a_bypass_record_inherits_the_pr_number_so_g4_can_count_it` — the docstring claims countability and asserts `count == 2`, but the fixture has no stage statuses, so the field that actually diverges between the writers (`stages`) is empty on both sides and the mismatch cannot surface. Every test in `TestStageEntryUpsert` shares that bare-dict shape, so the class never exercises cross-writer snapshot comparability — which is why Blocker 1 shipped green. Give the countability tests realistic stage statuses and drive them through `PipelineStateMachine`, as `TestStageEntryRecordsDispatch` already does.
- [ ] **`agent/sdlc_router.py:1836-1844`** — `last = history[-1] if history else None` / `if isinstance(last, dict) and last.get("stage") == stage and last.get("confirmed") is False:` — the only-newest rule does not close the hole its own comment names. When the stale slot *is* the newest record (router records a `/do-patch` slot, the sub-skill dies before `start_stage`, then a `/do-build`→`/do-patch` chain enters PATCH), two distinct dispatch events still collapse into one record. `test_a_stale_slot_for_another_stage_does_not_absorb_this_entry` covers a different stage; `test_only_the_newest_unconfirmed_slot_is_upgraded` has an intervening record. Either age the slot out against `last["at"]`, or accept it and reword the comment, which currently reads as though the case is handled.
- [ ] **`tools/sdlc_stage_marker.py:390`** — `history = raw.get("_sdlc_dispatches")` / `if any(isinstance(d, dict) and d.get("skill") == skill for d in history): return ("STAGE_RAN_NOT_SKIPPABLE", ...)` — a sixth `_sdlc_dispatches` consumer absent from the PR body's enumeration of five, whose behavior changes. `/do-plan` or `/do-plan-critique` invoked outside the router, *including an invocation that aborts immediately to ask a question*, now writes a record via the PreToolUse hook and permanently burns the retroactive skip for that stage. Arguably more correct, but unanalyzed. Add it to the enumeration and state the escape-hatch consequence in `docs/features/sdlc-router-oscillation-guard.md`.
- [ ] **`tools/sdlc_dispatch.py:129`** — `history[-1]["run_id"] = run_id` — the #2003 run-identity annotation is applied only inside `record_dispatch_for_ledger`. `confirm_or_append_stage_entry` calls `record_dispatch(...)` directly and never sets it, so on a hook-driven lane the history carries `run_id` on some records and not others, and the ones without are exactly the records this PR exists to create. No current reader consumes it, so this is observability debt. Resolve the run identity in `_apply_pending_stage_entry`, or document that `run_id` is router-path-only.

### Nits

- [ ] **`agent/sdlc_router.py:1894-1904`** — the `Args:` block documents `stage_states`, `skill`, `now`, `pr_number` and stops; `confirmed` (added at `:1880`) carries the entire slot protocol and is the parameter a reader most needs explained. The `Returns:` block is likewise silent on the two new record fields.
- [ ] **`agent/pipeline_state.py:586`** — `stage = getattr(self, "_pending_stage_entry", None)` — the attribute is set only at `:645` and never initialized in `__init__` or `for_issue()`. `_load_state` states the standard verbatim: *"Defaults first so every construction path (including callers that build the instance via `__new__`) lands with the full attribute set."* Initialize it and read the attribute directly.
- [ ] **`agent/pipeline_state.py:594`** — `except Exception as e:  # pragma: no cover - audit must never block a write` — the branch IS covered by `test_a_failing_dispatch_record_never_blocks_the_marker_write`; the in-`try` import resolves the patched attribute at call time. Drop the pragma, keep the comment.
- [ ] **`agent/pipeline_state.py:506-513`** — `_save()`'s docstring still names *"the dispatch recorder in `agent.sdlc_router.record_dispatch`"* among the external writers it merges around. As of `:555`, `_save()` is itself one of those writers.
- [ ] **PR body, "Verification"** — "`ruff check` and `ruff format --check` clean." `python -m ruff check .` exits non-zero on this branch with 2 `N802` errors in `tests/db_derivation_guard.py`. Verified pre-existing on `main`, so not this PR's defect — but the claim as written does not reproduce. `ruff format --check` is clean (1348 files). Test count measured 699, not 694, on the named file set.

### Miscellaneous

- **File:** `agent/sdlc_router.py:1844`
  **Code:** `last.get("confirmed") is False`
  **Issue:** Ledgers in flight across the merge carry records with no `confirmed` key, so `last.get("confirmed")` is `None` and the first post-merge stage entry appends instead of upgrading — one duplicate record per in-flight lane. The two records' snapshots differ, so no G4 inflation, and it self-clears via the 10-entry FIFO. The same predicate defines a mixed-version window: a worker on old in-process hooks alongside a new `sdlc-tool` produces either permanently-unconfirmed slots or two records per router dispatch until every process is cycled.
  **Severity:** nit
  **Fix:** No migration needed. State in the merge note that `/update` + `./scripts/valor-service.sh restart` must run before the next lane starts (this PR touches `agent/`, so the restart is required regardless per CLAUDE.md), and mention the transitional duplicate so the next reader inspecting a ledger does not chase it as a bug.

- **"The G4 counter still scores alternating oscillation at ≤1 ... filed as #2801"** — tracked by [#2801](https://github.com/tomcounsell/ai/issues/2801) (OPEN). The measurement in the PR body (frozen alternation counts 4 and escalates; a real 8↔8b loop counts 2 and does not, because `_patch_cycle_count` moves each cycle) is consistent with `build_stage_snapshot`'s key set as pinned by `test_snapshot_key_set_is_unchanged`. Deferring rather than shipping the one-line fix is the right call.
- **"`_rule_patch_applied_after_review`'s docstring ... is one of eleven `bound by G4` claims"** — tracked by #2801 (OPEN). Confirmed: `main` carried 11 case-insensitive hits, this branch carries 10, and the one corrected is row 8b's. Spot-checked the survivors — row 8e (`agent/sdlc_router.py:1422`) re-dispatches the *same* skill, so its G4 claim does hold; the alternating rows are the ones at risk, correctly scoped to #2801.
- **"`stage_for_skill` inverts the canonical `STAGE_TO_SKILL` rather than adding a fourth hand-maintained map — the three that already exist are catalogued on #2491"** — tracked by [#2491](https://github.com/tomcounsell/ai/issues/2491) (OPEN). Note that `agent/hooks/pre_tool_use.py:60`'s `_SKILL_TO_STAGE` is now load-bearing for `_sdlc_dispatches` via the very hook path this PR names as the dominant writer, which raises #2491's priority.
- **"#2731 (dispatch liveness gate) is closed as re-scoped-and-deferred, not built"** — [#2731](https://github.com/tomcounsell/ai/issues/2731) is CLOSED, tracked=false. Explicit out-of-scope exclusion with the full analysis on the issue; the rationale (under the lease-read-only ruling the inter-run gate is a no-op, and the intra-run collision cannot be gated by a lease forks inherit) matches the plan's own Round 2 blocker B2. Accepted as an honest disposition rather than a silent drop.

### Verification Results

The plan's `## Verification` table targets the unshipped scope, so its rows are not a meaningful gate here. Ran what is applicable, in this environment:

| Check | Result |
|---|---|
| `scripts/pytest-clean.sh` over the 12 files named in the PR body | **699 passed** in 78s (PR body claims 694) |
| `python -m ruff format --check .` | pass — 1348 files |
| `python -m ruff check .` | **exit non-zero** — 2 `N802` in `tests/db_derivation_guard.py`; verified pre-existing on `main`, not introduced here |
| `git diff --name-only main...HEAD -- tools/sdlc_stage_query.py` | empty — off-limits file untouched, as promised |
| `grep -icn "bound by G4" agent/sdlc_router.py` | 10 (was 11 on `main`; row 8b's corrected) |
| Mixed-path G4 streak, `main` vs branch | **2 vs 1** — see Blocker 1 |

### Screenshots

valorengels:
## Review: Changes Requested

Re-review on HEAD `99d888e3` (prior review judged `cf658d68`; branch has since been rebased and one commit added). **Both prior blockers are resolved and I verified each myself rather than reading the commit message.** What remains is tech debt and nits — no blockers.

Single-judge path: this pass is a focused re-review of one commit against a prior two-judge consensus, invoked standalone with no supervised run. Stated so the calibration is auditable rather than implied.

### Blocker 1 (snapshot comparability) — resolved, reproduced in both directions

I ran a three-cycle mixed-path `/do-patch` loop (router slot → bypass chain → router slot, `PATCH` sitting at `failed` between cycles) against this branch's `confirm_or_append_stage_entry`, with and without `prior_status`:

| | PATCH status per record | `compute_same_stage_count` |
|---|---|---|
| without `prior_status` (the reviewed defect) | `['failed','in_progress','failed','in_progress','failed','in_progress']` | **1** |
| with `prior_status` (this PR) | `['failed','failed','failed','failed','failed','failed']` | **6** |

The streak now clears `MAX_SAME_STAGE_DISPATCHES = 3` on exactly the loop the feature exists to make visible. The mutate-and-restore in `confirm_or_append_stage_entry` is safe as written: `record_dispatch` builds `view` as a shallow copy and `build_stage_snapshot` builds `"stages"` via a fresh dict comprehension, so the snapshot captures values at build time and the `finally` restore cannot reach back into it.

`test_a_bypass_record_snapshots_the_pre_entry_status_like_the_router` is a genuine pin, not a presence assertion — it asserts the snapshot sequence *and* `count == 3`, and its docstring names why every other fixture in the class cannot catch this.

### Blocker 2 (plan forbids the PR) — resolved

`docs/plans/ledger-integrity.md` is rev-3 on `main` (`15c73c636`), `status: Ready`. The `[FENCE-BLOCKED]` No-Go is gone, the fence extension to `agent/pipeline_state.py` is recorded with the lead's ruling, the No-Gos / Success Criteria / Test Impact / Verification table all describe the shipped scope, and the round-2 escalation is closed in the doc. All 14 Success Criteria are ticked; I spot-verified the load-bearing ones against the diff.

## Rubric

- [x] **1. Plan vs. implementation match** — pass — *rev-3 plan matches the diff; every Success Criterion is delivered. See Tech Debt 1 for the one invariant the plan does not pin.*
- [x] **2. New code quality** — pass — *`prior_status` sentinel is correct, `try/finally` restore is sound, deferred import avoids the cycle. Nits below.*
- [x] **3. Test coverage** — pass — *the cross-writer comparability gap that let Blocker 1 ship green is now pinned with a red-verified regression test. Residual bare-fixture debt in Tech Debt 5.*
- [x] **4. Regression risk to existing callers** — pass — *streak measured at 6 where the pre-fix path scores 1 — the direction is now correct. The sixth consumer (`tools/sdlc_stage_marker.py`) is enumerated in the PR body and commented at the call site.*
- [x] **5. Data integrity** — pass — *two additive JSON fields, legacy records read via `.get`, no migration. Transition window in Miscellaneous.*
- [x] **6. Security** — n/a — *no new subprocess, eval, deserialization, or secret handling.*
- [ ] **7. Documentation accuracy** — fail — *the feature docs describe the upsert protocol but omit the pre-entry-status invariant the whole thing now rests on, and omit the retroactive-skip consequence. See Tech Debt 1 and 2.*
- [x] **8. PR body accuracy** — pass — *reproduced: 699 passing (700 collected, 1 pre-existing failure), `ruff check .` and `ruff format --check .` both clean, consumer enumeration is six. The prior review's N802 discrepancy is gone post-rebase.*
- [x] **9. Disclosed deferrals** — pass — *#2801 and #2491 OPEN; #2731 closed as re-scoped with the analysis on the issue.*
- [x] **10. Follow-up claims verified** — pass — *#2801 OPEN, #2491 OPEN, #2711 CLOSED, #2731 CLOSED — all as the PR body describes.*

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — PASS — *14/14 criteria delivered against the rev-3 plan.*
- **2. No-Gos from plan — none violated** — PASS — *the fence-blocked No-Go was lifted by the lead's ruling; `tools/sdlc_stage_query.py` is untouched, verified by `git diff --name-only`.*
- **3. New `except Exception` blocks — each has logger/raise/swallow-ok** — PASS — *`agent/pipeline_state.py:594` logs at debug with stage and exception; swallow justified and tested.*
- **4. New integration tests — exercise serialization boundary (not in-memory only)** — PASS — *`TestStageEntryRecordsDispatch` drives real `_save()` and reads back `json.loads(session.stage_states)`.*
- **5. Plan internal consistency — spike findings match task steps** — PASS — *rev-3 records the lead's ruling and rewrites the operative sections; round-1/2 reasoning is explicitly marked as history, not scope.*
- **7. New public APIs — docstrings present** — PASS — *`stage_for_skill`, `confirm_or_append_stage_entry`, and the `prior_status` rationale are all thoroughly documented. `record_dispatch`'s `Args` is stale (nit).*
- **8. Breaking changes — migration path documented** — PASS — *additive fields; transition window is a Miscellaneous advisory.*
- **9. Tests added for new behavior** — PASS — *13 new tests across the two files.*
- **10. Tests cover the failure path (not just happy path)** — PASS — *unknown stage, raising recorder, already-`in_progress` no-op, stale slot, snapshot key-set pin, and now the pre-entry-status pin.*
- **11. UI changes (if any) — screenshot captured** — N/A — *no UI files in the diff; the visual-proof gate is a no-op.*
- **12. Docs updated for user-facing changes** — PASS — *two feature docs updated, though see rubric item 7.*

### Blockers

- None

### Tech Debt

- [ ] **`docs/features/sdlc-router-oscillation-guard.md:214-249`** — the added G4 section documents the slot protocol, the `_save()` placement, and the `stage`/`confirmed` exclusion, but never states that an appended record snapshots the stage's **pre-entry** status so both writers observe the same side of the transition. That invariant is now the correctness keystone of the entire feature — violating it drops G4's streak *below* what the pre-feature code produces. A future editor reading only these docs would see nothing contradicting the removal of `prior_status`. Add it beside the `stage`/`confirmed` paragraph, which documents the same class of hazard.
- [ ] **`docs/features/sdlc-router-oscillation-guard.md:230-235`** — the consumer enumeration lists G1/G3/G7/rows 8b/8d and `_latest_dispatch_at` for rows 8/2b, but omits `tools/sdlc_stage_marker.py`'s retroactive-skip check, whose behavior this PR changes (a `/do-plan` that entered PLAN and aborted to ask a question is no longer retroactively skippable). It is commented at the call site and in the PR body, but the feature doc is where the consumer list lives and where the prior review asked for it.
- [ ] **`agent/sdlc_router.py:1851-1858`** — `# The router records immediately before invoking the sub-skill, so nothing else can land in between.` — the comment asserts the stale-slot case is structurally impossible, but it is not when the stale slot *is* the newest record: the router records a `/do-patch` slot, the sub-skill dies before `start_stage`, then a `/do-build`→`/do-patch` chain enters PATCH and upgrades that dead slot. Two distinct dispatch events collapse into one record — the exact defect this function exists to fix. `test_a_stale_slot_for_another_stage_does_not_absorb_this_entry` covers a *different* stage and `test_only_the_newest_unconfirmed_slot_is_upgraded` has an intervening record, so neither reaches it. Either age the slot out against `last["at"]`, or keep the behavior and reword the comment so it stops claiming the case cannot arise. Carried unchanged from the prior review.
- [ ] **`tools/sdlc_dispatch.py:126-129`** — `history[-1]["run_id"] = run_id` — the #2003 run-identity annotation is applied only inside `record_dispatch_for_ledger`. `confirm_or_append_stage_entry` calls `record_dispatch` directly and never sets it, so on a hook-driven lane the history carries `run_id` on some records and not others — and the ones without are precisely the records this PR exists to create. No current reader consumes it, so this is observability debt, but it should be resolved in `_apply_pending_stage_entry` or documented as router-path-only. Carried unchanged.
- [ ] **`tests/unit/test_sdlc_router_oscillation.py:1082-1096`** — `test_a_bypass_record_inherits_the_pr_number_so_g4_can_count_it` still starts from `states: dict = {}`. Its name and its `count == 2` assertion claim countability across a router/bypass boundary, but with no stage statuses the `stages` field is empty on both sides, so the assertion would hold even with the comparability bug reintroduced. The new pre-entry-status test covers the load-bearing case, which is why this is debt rather than a blocker — but the `pr_number` test should carry real statuses so it actually tests what it claims.

### Nits

- [ ] **`agent/sdlc_router.py:1930-1944`** — `record_dispatch`'s `Args:` block documents `stage_states`, `skill`, `now`, `pr_number` and stops; `confirmed` carries the entire slot protocol and is the parameter a reader most needs explained. `Returns:` is likewise silent on the two new record fields (`stage`, `confirmed`). Carried unchanged.
- [ ] **`agent/pipeline_state.py:586`** — `pending = getattr(self, "_pending_stage_entry", None)` — the attribute is set only at `:649` and never initialized in `__init__` / `for_issue()`. `_load_state` states the standard verbatim: *"Defaults first so every construction path (including callers that build the instance via `__new__`) lands with the full attribute set."* Initialize it to `None` and read the attribute directly. Carried unchanged.
- [ ] **`agent/pipeline_state.py:594`** — `except Exception as e:  # pragma: no cover - audit must never block a write` — the branch IS covered, by `tests/unit/test_pipeline_state_machine.py:1973` (`test_a_failing_dispatch_record_never_blocks_the_marker_write`); the in-`try` import resolves the patched attribute at call time. Drop the pragma, keep the comment. Carried unchanged.
- [ ] **`agent/pipeline_state.py:509-511`** — `_save()`'s docstring still names *"the dispatch recorder in `agent.sdlc_router.record_dispatch`"* among the **external** writers it merges around. As of `:555` (`self._apply_pending_stage_entry(data)`), `_save()` is itself one of those writers.
- [ ] **`agent/sdlc_router.py:1894-1908`** — `stage_states[stage] = prior_status` ... `finally: stage_states[stage] = current` — the fix temporarily mutates the caller's dict to a stale value across the `record_dispatch` call. It is safe today because `build_stage_snapshot` copies `stages` eagerly, and the `try/finally` is correct. But the safety is an emergent property of a function two calls away, not a local one: anyone who later makes the snapshot hold a reference into `stage_states` silently breaks this. Building the snapshot from an explicit overridden view and passing it down would make the invariant local instead of load-bearing-at-a-distance.

### Miscellaneous

- **File:** `agent/sdlc_router.py:1859`
  **Code:** `last.get("confirmed") is False`
  **Issue:** Ledgers in flight across the merge carry records with no `confirmed` key, so `last.get("confirmed")` is `None` and the first post-merge stage entry appends rather than upgrading — one duplicate record per in-flight lane. The two records' snapshots differ, so no G4 inflation, and it self-clears via the 10-entry FIFO. The same predicate defines a mixed-version window: an old in-process worker alongside a new `sdlc-tool` produces either permanently-unconfirmed slots or two records per router dispatch until every process is cycled.
  **Severity:** nit
  **Fix:** No migration needed. This PR touches `agent/`, so `/update` + `./scripts/valor-service.sh restart` is required after merge regardless (CLAUDE.md); state the transitional duplicate in the merge note so the next reader inspecting a ledger does not chase it as a bug.

- **"The G4 counter still scores alternating oscillation at ≤1 ... filed as #2801"** — tracked by [#2801](https://github.com/tomcounsell/ai/issues/2801) (OPEN). Deferring rather than shipping the one-line fix remains the right call; the PR body and the plan both carry the measurement showing the obvious fix does not fire on the loops it names.
- **"`_rule_patch_applied_after_review` ... one of eleven `bound by G4` claims"** — tracked by #2801 (OPEN). Row 8b corrected; the other ten name same-skill loops G4 genuinely counts.
- **"`stage_for_skill` inverts the canonical `STAGE_TO_SKILL` rather than adding a fourth hand-maintained map"** — tracked by [#2491](https://github.com/tomcounsell/ai/issues/2491) (OPEN).
- **"#2731 (dispatch liveness gate) is closed as re-scoped-and-deferred, not built"** — [#2731](https://github.com/tomcounsell/ai/issues/2731) is CLOSED with the full analysis on the issue and the disposition recorded in the rev-3 plan. Accepted as an honest disposition.

### Review Delta (vs prior review on HEAD `cf658d68`)

- **Resolved**: Blocker 1 (writers snapshot opposite sides of the transition) — fixed by `99d888e35`; I reproduced the before/after streak myself (1 → 6).
- **Resolved**: Blocker 2 (plan forbids this PR) — `docs/plans/ledger-integrity.md` is rev-3 `Ready` on `main` with the fence extension and rewritten criteria.
- **Resolved**: rubric 3 (no test exercised cross-writer comparability) — red-verified regression test added.
- **Resolved**: rubric 8 / nit (`ruff check .` non-zero on 2 `N802`) — clean on this branch post-rebase; I reproduced `All checks passed!` and `1350 files already formatted`.
- **Resolved**: prior TD3 (sixth consumer unenumerated) — now in the PR body and commented at the call site; the *docs* half is still open (Tech Debt 2).
- **Unchanged**: TD on the stale-slot-is-newest hole; TD on router-path-only `run_id`; all four prior nits.
- **New**: Tech Debt 1 (the pre-entry-status invariant is undocumented) — created by the fix itself. Nit on the mutate-and-restore, same origin.
- **Narrowed, not resolved**: the prior review's TD on *every* `TestStageEntryUpsert` fixture being a bare `states = {}` — one fixture now has real statuses (Tech Debt 5).

### Verification Results

Plan `## Verification` table, run on this branch:

| Check | Result |
|---|---|
| Router / oscillation / decision tests | pass |
| Dispatch tests (unit + integration) | see note below |
| State machine + marker + hooks | pass |
| Stage-query + lease binding | pass |
| `python -m ruff check .` | **pass** — `All checks passed!` |
| `python -m ruff format --check .` | **pass** — 1350 files |
| Off-limits file untouched (`tools/sdlc_stage_query.py`) | **pass** — diff empty |
| Snapshot not widened (`-k build_stage_snapshot`) | pass |
| Row 8b claim corrected (`grep -in "bound by G4"`) | pass — ten hits, none in `_rule_patch_applied_after_review` |
| Mixed-path G4 streak (my own measurement) | **6** — clears the threshold; the pre-fix path scores 1 |

Full named set: **700 collected, 693 passed, 7 failed** in my scratch worktree. I did **not** take "pre-existing" on faith. Running the 7 in isolation in the main checkout left only 1 failing — so I ran a **control: the base branch checked out into an equivalent scratch worktree**, which reproduced the **identical 7 failures**. The 6 extra are artifacts of running the run-id/lease self-heal subprocess tests from a non-standard worktree path, not branch regressions. The one genuine pre-existing failure is `TestDispatchRecordLease::test_lease_lost_between_peek_and_write_refuses`, exactly as the PR body discloses. Effective status on a standard checkout: **699 passed, 1 pre-existing failure** — the PR body's number reproduces.

### Screenshots

---

**Verdict rationale:** rubric item 7 (`fail`, non-critical) → `CHANGES REQUESTED — Tech Debt`. No blockers. The design and the blocker fix are both sound; what is left is documentation of the invariant the fix established, plus carried-forward debt and nits that `/do-patch` can clear in one pass.

valorengels:
## Review: Approved

Re-review on HEAD `12cc5f21` (prior review judged `99d888e3`; one commit added, `12cc5f215` "document the pre-entry-status invariant and clear review debt"). **All five prior tech-debt items and all five prior nits are resolved, and I verified each against the code rather than reading the commit message.** No blockers, no tech debt, no nits remain.

Single-judge path: this is a focused re-review of one debt-clearing commit against a prior two-judge consensus, invoked standalone with no supervised run. Stated so the calibration is auditable rather than implied.

## Rubric

- [x] **1. Plan vs. implementation match** — pass — *`docs/plans/ledger-integrity.md` is rev-3 `status: Ready` on `main`, `tracking:` #2730. All 14 Success Criteria are ticked and I spot-verified the load-bearing ones against the diff: exactly-one-record per router dispatch, re-entry still appends, `start_stage` on an already-`in_progress` stage records nothing (`pipeline_state.py:948`), unknown stage records nothing without raising, newest-record-only slot eligibility, `pr_number` inheritance, survival of `_save()`'s merge, failing record never blocks the marker, `stage`/`confirmed` outside the snapshot, `stage_for_skill` derived from `STAGE_TO_SKILL`.*
- [x] **2. New code quality** — pass — *the `_NO_PRIOR_STATUS` sentinel is correct, the deferred import in `_apply_pending_stage_entry` avoids the `pipeline_state` ↔ `sdlc_router` cycle, and the prior nit's mutate-and-restore is gone: `snapshot_states` is now an explicit overridden view passed into `record_dispatch`, so the invariant is local instead of load-bearing at a distance.*
- [x] **3. Test coverage** — pass — *13 new tests across two files, covering counts (not presence), the red-verified pre-entry-status pin, the `/do-build`→`/do-patch` regression fixture, the snapshot key-set pin, `_save()` merge survival, and the raising-recorder failure path. `test_a_bypass_record_inherits_the_pr_number_so_g4_can_count_it` now carries real stage statuses (`{"REVIEW": "failed", "BUILD": "completed"}`), closing the prior review's "bare fixture" debt.*
- [x] **4. Regression risk to existing callers** — pass — *I re-measured the mixed-path streak myself (below): 1 without the `prior_status` override, 6 with it, on a three-cycle router/bypass `/do-patch` loop. The direction is correct and clears `MAX_SAME_STAGE_DISPATCHES`. All six `last_dispatched_skill` / `_latest_dispatch_at` consumers are enumerated in the PR body and the feature doc, and their suites pass unchanged.*
- [x] **5. Data integrity** — pass — *two additive JSON fields (`stage`, `confirmed`), legacy records read via `.get`, no migration, `_sdlc_dispatches` still FIFO-bounded at 10. The in-flight-ledger transition window is a one-time self-clearing duplicate, already narrated in the prior review and unchanged in kind.*
- [x] **6. Security** — n/a — *no new subprocess, eval, deserialization, network call, or secret handling.*
- [x] **7. Documentation accuracy** — pass — *the prior review's `fail` on this item is resolved. `docs/features/sdlc-router-oscillation-guard.md` now carries the "Both writers must snapshot the same side of the stage transition" section with the measurement, closing prior TD1, and the sixth consumer (`tools/sdlc_stage_marker.py`'s retroactive-skip check) is now enumerated in the same doc, closing prior TD2. `docs/features/sdlc-stage-tracking.md` gains the "Stage entry records a dispatch" section.*
- [x] **8. PR body accuracy** — pass — *reproduced independently: `ruff check .` `All checks passed!`, `ruff format --check .` `1350 files already formatted`, the referenced consumer count is six, and the "699 tests pass, one pre-existing failure" claim reproduces (see Verification Results — I ran a base-commit control rather than taking "pre-existing" on faith).*
- [x] **9. Disclosed deferrals** — pass — *all four disclosures verified against live issue state; see Acknowledged Deferrals.*
- [x] **10. Follow-up claims verified** — pass — *#2801 OPEN, #2491 OPEN, #2731 CLOSED, #2711 CLOSED, #2730 OPEN — exactly as the PR body describes. The "partially closes #2730" framing is honest: the record-completeness half ships, the counter half is #2801, and the PR body says so up front.*

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — PASS — *14/14 delivered against rev-3.*
- **2. No-Gos from plan — none violated** — PASS — *`tools/sdlc_stage_query.py` untouched (`git diff --name-only origin/main...HEAD` empty); `build_stage_snapshot`'s key set unchanged and pinned by a test; no new `_meta` key; the counter change is split to #2801 as the No-Go requires.*
- **3. New `except Exception` blocks — each has logger/raise/swallow-ok** — PASS — *`agent/pipeline_state.py:600` logs at debug with the stage and the exception; the swallow is justified ("a stage must not become unrecordable because its audit entry failed") and covered by `test_a_failing_dispatch_record_never_blocks_the_marker_write`. The prior nit's stale `# pragma: no cover` is gone.*
- **4. New integration tests — exercise serialization boundary (not in-memory only)** — PASS — *`TestStageEntryRecordsDispatch` drives the real `_save()` and reads back via `json.loads(session.stage_states)`; `tests/integration/test_sdlc_dispatch.py` passes (17).*
- **5. Plan internal consistency — spike findings match task steps** — PASS — *rev-3 marks rounds 1 and 2 explicitly as history and Round 3 Disposition as operative.*
- **7. New public APIs — docstrings present** — PASS — *`stage_for_skill` and `confirm_or_append_stage_entry` are thoroughly documented, and `record_dispatch`'s `Args:`/`Returns:` now document `confirmed`, `snapshot_states`, and the two new record fields, closing the prior nit.*
- **8. Breaking changes — migration path documented** — PASS — *additive fields only; transition behavior is documented and self-clearing.*
- **10. Tests cover the failure path (not just happy path)** — PASS — *unknown stage, raising recorder, already-`in_progress` no-op, stale slot for another stage, two stacked slots, snapshot key-set pin, pre-entry-status pin.*
- **11. UI changes (if any) — screenshot captured** — N/A — *no HTML/CSS/JS/TSX/Vue/template files in the diff; the visual-proof gate is a no-op.*

### Blockers

- None

### Tech Debt

- None

### Nits

- None

### Miscellaneous

- None

- **"The G4 counter still scores alternating oscillation at ≤1 ... filed as #2801"** — tracked by [#2801](https://github.com/tomcounsell/ai/issues/2801) (OPEN). The deferral is honest and well-evidenced: the PR body and the plan both carry the measurement showing that removing the skill break does not fire on a real 8↔8b loop, because `build_stage_snapshot` includes `_patch_cycle_count` and that increments every cycle.
- **"`_rule_patch_applied_after_review` ... one of eleven `bound by G4` claims"** — tracked by #2801 (OPEN). I re-ran the plan's own audit row: `grep -in "bound by G4" agent/sdlc_router.py` returns ten hits, and I confirmed by function boundary that none falls inside `_rule_patch_applied_after_review` (which spans `:1232`–`:1299`; the nearest hits are `:1167` and `:1320`, in other rules).
- **"`stage_for_skill` inverts the canonical `STAGE_TO_SKILL` rather than adding a fourth hand-maintained map"** — tracked by [#2491](https://github.com/tomcounsell/ai/issues/2491) (OPEN). `STAGE_TO_SKILL` is 1:1 across all nine stages, so the inversion is unambiguous, and `test_stage_for_skill_inverts_the_canonical_map` round-trips every entry.
- **"Only the newest record is eligible as a slot ... this narrows the hole rather than closing it"** — accepted as a deliberate, documented under-count. The comment at `agent/sdlc_router.py:1855-1866` now states the residual case plainly (a sub-skill that dies before `start_stage` leaves its slot as the newest record, and a later entry into the same stage upgrades it) and argues the safe direction. This is the prior review's TD3, and the resolution taken is the one that review offered as an alternative: keep the behavior, stop claiming the case cannot arise.
- **"Records appended by the stage entry carry no `run_id`"** — prior TD4, now resolved as documented-by-design at `tools/sdlc_dispatch.py:120-133` and in the feature doc, with the reason (`PipelineStateMachine` reaches its store as a bare `PipelineLedger`; `active_run_id` is an `AgentSession` field) and the reader contract (a missing `run_id` means "not recorded", never "a different run"). No current reader consumes the field.
- **"#2731 (dispatch liveness gate) is closed as re-scoped-and-deferred, not built"** — [#2731](https://github.com/tomcounsell/ai/issues/2731) is CLOSED. Accepted as an honest disposition.

### Review Delta (vs prior review on HEAD `99d888e3`)

Every open item from the prior review is closed by `12cc5f215`. I verified each in the code, not in the commit message:

- **Resolved — TD1** (pre-entry-status invariant undocumented): `docs/features/sdlc-router-oscillation-guard.md` now carries a dedicated "Both writers must snapshot the same side of the stage transition" paragraph, including the measured 1-vs-6 figures and the line that names it "the correctness keystone of the whole feature ... a change that drops it does not fail loudly".
- **Resolved — TD2** (sixth consumer missing from the docs): the same file now documents `tools/sdlc_stage_marker.py`'s retroactive-skip check and its behavior change.
- **Resolved — TD3** (comment claimed the stale-slot case was structurally impossible): the comment at `agent/sdlc_router.py:1855-1866` is rewritten and no longer makes that claim; it names the residual hole, why an age-out is not guessed at, and why the under-count is the safe direction.
- **Resolved — TD4** (`run_id` router-path-only): documented at the annotation site with the mechanism and the reader contract.
- **Resolved — TD5** (`pr_number` fixture was a bare `states = {}`): the fixture now starts from `{"REVIEW": "failed", "BUILD": "completed"}` and asserts the per-record `stages["REVIEW"]` sequence alongside `count == 2`, so the assertion is load-bearing.
- **Resolved — nit** (`record_dispatch` `Args:`/`Returns:` stale): both now document `confirmed`, `snapshot_states`, and the `stage`/`confirmed` record fields.
- **Resolved — nit** (`_pending_stage_entry` not initialized on every construction path): initialized in `_load_state` at `:409` alongside `stage_skips`, under the existing "Defaults first so every construction path (including callers that build the instance via `__new__`) lands with the full attribute set" comment — which is the shared path `for_issue()`'s `__new__` construction calls at `:393`. The read at `:591` is now a direct attribute access, not a `getattr` fallback.
- **Resolved — nit** (`# pragma: no cover` on a covered branch): dropped; the comment is kept.
- **Resolved — nit** (`_save()` docstring listed `record_dispatch` among external writers): now reads "(and, since #2730, `_save()` itself, which appends the stage-entry record AFTER this merge)".
- **Resolved — nit** (mutate-and-restore of the caller's dict): replaced by the `snapshot_states` parameter. I checked the aliasing question the prior review raised: `record_dispatch` builds `view` as a fresh comprehension over `snapshot_states`, and `build_stage_snapshot` builds `"stages"` and `_verdicts` as fresh dicts, so nothing holds a reference back into the caller's state. The record is still appended to `stage_states`; only the snapshot is projected from the override.
- **Unchanged and accepted**: the in-flight-ledger transition window (records with no `confirmed` key append rather than upgrade once, then self-clear via the FIFO).

### Verification Results

Plan `## Verification` table, run by me on this branch:

| Check | Result |
|---|---|
| Router / oscillation / decision tests | pass |
| Dispatch tests (unit) | pass except the disclosed pre-existing failure, below |
| Dispatch tests (integration) | pass — 17 |
| State machine + marker + hooks | pass |
| Stage-query + lease binding | pass |
| `python -m ruff check .` | **pass** — `All checks passed!` |
| `python -m ruff format --check .` | **pass** — 1350 files already formatted |
| Off-limits file untouched (`tools/sdlc_stage_query.py`) | **pass** — diff empty |
| Snapshot not widened | **pass** — key set pinned to `{stages, _verdicts, _patch_cycle_count, _critique_cycle_count, pr_number}` by test |
| Row 8b claim corrected (`grep -in "bound by G4"`) | **pass** — ten hits, none inside `_rule_patch_applied_after_review` (verified by function boundary, not by count alone) |

**Mixed-path G4 streak, measured by me** on a three-cycle router/bypass `/do-patch` loop with `PATCH` sitting at `failed` between cycles:

| | `compute_same_stage_count` |
|---|---|
| without `prior_status` (the defect this branch fixed) | **1** |
| with `prior_status` (this branch) | **6** |

Six records, streak 6, clearing `MAX_SAME_STAGE_DISPATCHES`. This reproduces the doc's stated figures exactly.

**On the test-failure claim, which I did not take on faith.** Running the full named set in a scratch worktree gave **7 failures**, not the 1 the PR body discloses. Rather than report a regression, I ran two controls:

1. The single failure the PR names, `TestDispatchRecordLease::test_lease_lost_between_peek_and_write_refuses`, reproduces on `main` at `5e47b0ce` in the standard checkout — genuinely pre-existing, exactly as disclosed.
2. The other 6 (`TestRunIdRequiredFlag::test_missing_run_id_self_heals_on_free_lock[×4]`, `test_docs_completed_backfill_refuses_unverified_review`, `TestCLI::test_with_issue_number_outputs_json`) pass in the standard checkout. So I checked out the branch's **merge base `a9016cbd` into an equivalent scratch worktree** and ran the identical selection: **the same 6 fail there too.** They are `LEASE_ABSENT`/`RUN_ID_REQUIRED` artifacts of running the lease self-heal subprocess tests from a non-standard worktree path, not branch regressions.

I also confirmed the branch base and `main` differ only by docs/plan commits (`git log a9016cbd..5e47b0ce` is six `docs(plan)` commits), so the control is a clean isolation of this branch's code. Effective status on a standard checkout: **699 passed, 1 pre-existing failure** — the PR body's number reproduces.

### Screenshots

---

**Deploy note:** this PR touches `agent/` and `tools/`, so `/update` plus `./scripts/valor-service.sh restart` is required on every machine after merge (CLAUDE.md). Until every process is cycled, a ledger written by an old worker alongside a new `sdlc-tool` can show one transitional duplicate record per in-flight lane — the two snapshots differ so there is no G4 inflation, and it clears itself via the 10-entry FIFO. Worth stating in the merge note so the next reader inspecting a ledger does not chase it as a bug.

**Verdict rationale:** every rubric item is `pass` or `n/a`; all twelve Pre-Verdict items are PASS or N/A; zero blockers, zero tech debt, zero nits, empty Miscellaneous. The two prior review rounds did the hard work and this commit closed the remainder without introducing anything new — I re-derived the load-bearing measurement myself rather than inheriting it. The scope limitation (record-completeness only, counter deferred to #2801) is disclosed in the PR body, the plan's No-Gos, and the issue, so #2730 stays open by design after merge.

PR #2803, opened 2026-08-13
Anchor run-identity history on the ledger so a re-ensure rebinds instead of re-minting

### LINKED ISSUE #2675: do-sdlc Step 3d.6 continuity re-ensure mints a new run_id after lease lapse, recreating the ledger anchor and wiping stage markers
Observed live during the 2026-08-07/08 backlog fan-out, on the #2643 pipeline (reported by its do-sdlc supervisor; relayed by the merge coordinator).

## What happened

The PLAN subagent hit `LEASE_ABSENT` on the run's original run_id — the lease had lapsed mid-pipeline. The Step 3d.6 continuity re-ensure then minted a NEW run_id instead of reusing the existing one, and recreated the `sdlc-local-2643` ledger anchor, wiping all previously recorded stage markers for the run.

The supervisor recovered manually: it rebound to the original run_id (32920971d5a74c9cb6a236126dbdd341) and restored the ISSUE/PLAN markers truthfully. No work was lost, but only because the supervisor noticed and knew the prior state.

## Why it matters

- A continuity mechanism whose recovery path destroys the continuity record inverts its purpose: after the re-ensure, the ledger claims no stages ran, so a less careful supervisor (or a fresh fork) would honestly re-run completed stages or fail gates that depend on the markers.
- Marker restoration by hand is exactly the state the verdict/marker machinery exists to prevent.

## Expected

Re-ensure after a lapsed lease should rebind to the existing run_id for the (issue, machine) anchor and extend/re-acquire the lease, never mint a fresh run identity while an anchor with recorded markers exists. If rebinding is impossible, it should fail loudly rather than recreate the anchor empty.

## Related

- #2659 / PR #2667 (supervised-run signal never renewed — the signal-expiry half of this failure family; the lease lapse that triggered this re-ensure is the same 30-minute class)
- #2648 (issue-lock liveness semantics)
- #2657 (supervisor fork turn-chaining)

valorengels:
## Problem

`/do-sdlc` Step 3d.6's between-stage continuity re-ensure mints a NEW run_id mid-run, after which the pipeline silently loses its identity.

Reproduced live on this lane. A routine CRITIQUE marker write returned:

```
[ERROR] ISSUE_LOCKED: issue lock held by a foreign run
(run_id=892e56f81bad42a9aa5d68100d30b026, session=sdlc-local-2675)
```

while the supervising run's id was `c76476af…`. The session record afterwards held `owned_run_ids: ["892e56f8…"]` — **only the new id**. Since `_append_owned_run_id` reads/dedups/appends/caps and never replaces, a history containing only the new id proves the record it appended to was already empty. The identity history that `_validated_reuse_candidate`'s free-lock branch depends on had been erased, which is exactly why the re-mint could not be avoided. The failure is self-perpetuating: each re-mint lands on an empty history, so the next re-ensure has no corroboration either.

The ledger is never actually wiped. `PipelineLedger` is issue-keyed, has no TTL, and has no deleter anywhere in production code — it only *reads* empty from a freshly minted identity's vantage point, which is observationally identical to a wipe and is why #2675 was filed as one.

## Solution

Anchor run-identity history on the **durable** ledger rather than the fragile session record.

- **Task 1** — removed `2>/dev/null || true` from all three occurrences in `.claude/skills-global/do-sdlc/SKILL.md` (session-ensure, the TEST stage-marker, and the Step 3d.6 re-ensure), and added the exit-code branch and run_id-adoption prose. This makes the pre-existing "route its payload through the same three-way table" instruction followable instead of self-contradictory. A *foreign-owner* refusal stops the run; a transient error is surfaced and retried, never converted into an abort.
- **Task 2** — a `_run_identities` anchor on the ledger, plus a **fourth proof** in `_validated_reuse_candidate` consulting it, appended strictly after the existing three. Reads and writes are both fail-open: a read error degrades to today's three proofs, and a write failure can never fail a lock acquisition that already succeeded.
- **Task 3** — same suppression removed from `docs/sdlc/do-plan.md`.

## File fence — exact regions for lane-identity's rebase

All edits to the contested `tools/sdlc_session_ensure.py` are confined to **three hunks**, quoted against merge-base `36481e121` (the branch base is behind `main`, so use `git merge-base`, not `git diff main`):

| Old range | New range | What |
|---|---|---|
| after `:88` | `:90-99` | `SDLC_RUN_IDENTITY_HISTORY_MAX` constant, directly below `OWNED_RUN_IDS_CAP` |
| `:295-307` | `:306-337` | `_validated_reuse_candidate` docstring |
| `:344-349` | `:374-400` | `_validated_reuse_candidate` body — the fourth proof |

`_validated_reuse_candidate` now spans `:295-400`. **`ensure_session` is byte-identical to `main`.** Nothing touches the mint area (~`:639`) or the `:499` `update_fields` hunk.

## Two deliberate deviations from the plan

**1. The anchor write is not in `_acquire_run_lock_and_bind`.** That location is outside the approved `:284-347` region and adjacent to the contested `:499` hunk, so the fence forbade it. The write instead lives at `tools/_sdlc_utils.py::_anchor_confirmed_run_identity`, called from `resolve_ledger_lease`'s confirmed-owner branch — the one place a run is *proven* to be the live lease owner with the pinned `target_repo` already in hand, so there is no extra `gh` resolution and provenance stays strictly self-written.

Consequence, stated plainly: a run is anchored on its **first mutating `sdlc-tool` call** rather than at mint. A run that mints and then lapses before writing anything stays unanchored and degrades to today's three proofs. That is strictly better than the status quo and covers #2675's own reproduction, but it is not literally "on bind" — mint-time anchoring would need a fence waiver.

**2. The plan's "return the live owner" mirror was deliberately not implemented.** The plan suggested the fourth proof also mirror `:333-335` by returning the live owner id when both the claim and the owner appear in `_run_identities`. That mirror is unsound: `owned_run_ids` is *session*-scoped self-history, but `_run_identities` is *issue*-scoped and legitimately contains identities of genuinely different runs that worked the issue over its lifetime. Implementing it would hand a caller an id it never claimed, taken from a live foreign lock — the exact no-adopt hazard the helper exists to prevent, and the #1915 duplicate-pipeline class. The anchor therefore corroborates on a **free lock only** and is never consulted to take over a live one.

## No-adopt invariant

The fourth proof is a membership test that echoes back the caller's own claim (`if anchor_repo and reuse_run_id in read_run_identities(...): return reuse_run_id`). It never returns a value read out of the anchor. Pinned by `test_foreign_live_holder_still_yields_issue_locked` (unit and real-Redis): both ids in the anchor, live foreign holder, still `ISSUE_LOCKED`. Three `read.assert_not_called()` controls prove proofs 1-3 return before the anchor is ever consulted.

## Testing

Demonstrated red before implementing: the new unit tests ran **26 failed, 1 passed** against unmodified code, then **27 passed** after. With only the fourth-proof block excised, `test_lapsed_lease_rebinds_to_the_anchored_identity` failed while both no-adopt controls still passed, so the rebind test is load-bearing and the controls are genuine controls.

Targeted suites, `POPOTO_TEST_DB=11`: `test_sdlc_session_ensure.py`, `test_pipeline_ledger.py`, `test_sdlc_run_identity_resume.py`, plus the untouched controls `test_sdlc_stage_marker.py` and `test_pipeline_state_machine.py` — **403 passed**. `ruff check` and `ruff format --check` clean. Every Verification and anti-criteria row passes by exit code.

Closes #2675


valorengels:
## Review: APPROVED

Reviewed against plan revision 2 using the merge-base diff. No blockers; all eight scrutiny points hold in code rather than only in comments.

### The invariant that mattered most

`_validated_reuse_candidate`'s fourth proof is a pure membership test — the only value it can return is `reuse_run_id`, the caller's own claim. `read_run_identities` returns a `list[str]` used solely as the right-hand side of an `in`, and every error path (including the `except` arm) falls to `return None`. There is no route by which an id from the anchor reaches a caller that did not already carry it, so the no-adopt invariant survives a durable, readable anchor.

The tests pin this rather than asserting on mocks: the unit control stocks the anchor with three ids the caller never claimed and requires `None`, so an implementation returning `anchor[-1]` fails it. The real-Redis counterpart puts *both* ids in the anchor with a live foreign holder and still requires `ISSUE_LOCKED`.

### The plan bullet the builder refused was right to refuse

The plan asked the fourth proof to mirror the existing "lapsed-then-re-minted" arm by returning the *live owner* id when both the claim and the owner appear in `_run_identities`. That arm is safe today only because `owned_run_ids` is **session-scoped** self-history. `_run_identities` is **issue-scoped** and legitimately holds identities of genuinely different runs across the issue's lifetime, so mirroring it would hand caller A the id of a live, different holder B — after which A writes under B's identity and `ISSUE_LOCKED` stops discriminating. That is the #1915 duplicate-pipeline class.

The plan's Risk 3 claims "identical provenance to `owned_run_ids`." That is true of the *write* (self-confirmed owners only) but false of the *scope*, and scope is what makes the read arm sound. The refusal is correct and the deviation is documented and tested.

### Everything else

Proofs 1-3 are byte-identical and unreordered, and the `read.assert_not_called()` controls genuinely intercept (the lazy import resolves at call time). Fail-open is double-guarded on reads and triple-guarded on writes, with the write's return value discarded so it cannot alter lease resolution — matching the repo convention where the pre-write gate `revalidate_ledger_lease` still fails closed. The write site is a genuine ownership proof: an unheld key peeks as `owner_run_id=None` and diverts to `LEASE_ABSENT`, and an unpinned `target_repo` is skipped rather than minting a phantom `None:{issue}` key.

The cap is bounded, newest-last, and env-overridable with the required marker. A stale anchor entry cannot authorize anything new, because the fourth proof fires only on a **free** lock — where the caller would have won a fresh mint anyway. The anchor changes which label the winner carries, never whether it wins. `_run_identities` also survives concurrent `_save()` via the preserved-metadata merge.

The skill-body edits make only a foreign-owner `ISSUE_LOCKED` a stop condition; `LEASE_ABSENT`, hand-off, orphaned-lock and broker errors are explicitly surface-and-retry, so a Redis blip cannot abort a healthy run. File fence compliant: three hunks, `ensure_session` byte-identical, nothing near the mint or the `:499` hunk.

### Fixed during review

The "Recovery after run_id loss" bullet still described only the live-lock and session-record proofs. Corrected in `41603a642` to name the anchor.

### Accepted residuals (follow-ups, not blockers)

**The honest trade this PR makes.** On a free lock, any id in the issue's anchor is now corroborated regardless of which run carried it, so two concurrent supervisors could both believe they own the same run_id. Reaching it requires a supervisor to claim an `owner_run_id` it read out of a refusal payload, which the skill's self-identity check already forbids — but the tool-level backstop for that violation is weaker than before. There is no better alternative here: the session record is empty by construction, and that emptiness *is* the bug. Worth adding to the plan's Risk 3, which does not name this case.

**Duplicate target-repo resolution.** The read site calls `_resolve_target_repo_fallback()` and `_acquire_run_lock_and_bind` calls `_resolve_target_repo()` a few lines later, so with `GH_REPO` unset this is a second `gh repo view` subprocess on exactly the lapsed-lease path. Harmless but avoidable by threading the resolved slug through. The PR body's "no extra `gh` resolution" is true of the write site, not the read site.

**`docs/sdlc/do-build.md:75,85` still carry `2>/dev/null || true`** on two `meta-set` calls — the same hole one directory over, outside this plan's declared scope.

valorengels:
Field observation from the #2730 lane (PR #2802), which ran across this fix's merge. Adding it because the timeline brackets the merge and gives a before/after on the same issue's anchor.

**Before the fix was live — one wipe observed.** The anchor for issue #2730 was created at 02:2x under `run_id=51392faedf29444e8d6f269db1d42db1` and carried a full trail: ISSUE/PLAN/CRITIQUE/BUILD/TEST all `completed`, a CRITIQUE `APPROVED` verdict, and `pr_number=2802`. After a `/do-pr-review` pass, `sdlc-tool stage-query --issue-number 2730` returned `stages` holding **only** `_verdicts.REVIEW`. Every stage marker, the CRITIQUE verdict, and `pr_number` were gone. The next `session-ensure` returned a **different** `run_id` (`603825d2d4174525833716ff9948a55a`), which is the re-mint signature this PR targets.

Caveat on attribution: `sdlc-tool` resolves to `$HOME/src/ai` regardless of cwd, so the executing code was the main checkout's, and this checkout's HEAD was moving under other agents. This PR merged at 02:36:57Z and the wipe was observed around 02:58, so the fix had *probably* landed in that checkout by then — but I cannot reconstruct its exact HEAD at 02:58, so I am not claiming this as a post-fix regression.

**After the fix was definitely live — no wipe.** I rebuilt the anchor at 03:11 under `603825d2...` with the fix confirmed present (`ec585ceb9` an ancestor of both the checkout and the branch). That anchor then survived **two** full `/do-pr-review` passes plus `verdict finalize`, `meta-set`, and thirteen further stage-marker writes, ending with all nine stages `completed` and the same `run_id` throughout. `sdlc-tool run-health` reported `{"ok_writes": 12, "fail_writes": 0, "trail_complete": true, "disposition": "clean"}`.

So the only wipe I saw sits on the ambiguous side of the merge boundary, and the anchor was stable for the entire post-fix run. Reads as the fix working rather than as a residual gap — recording it here mainly so the negative result is on the record if anyone else sees a wipe after this merged.

Two unrelated things the same lane surfaced, in case they are useful:

- `_review_artifact_posted` resolves the PR via `_lookup_pr`, whose issue-search leg requires a **closing** keyword (`Closes/Fixes/Resolves #N`) in the PR body and whose branch-head leg needs a slug it was not passed. A PR body saying `Refs #N` therefore fails `verdict finalize` with `REVIEW_ARTIFACT_MISSING` even though the `## Review:` comment exists with a matching `head_sha` trailer. The error names the artifact as missing, which points the reader at the review rather than at the body's issue-reference form. Changing `Refs` to `Closes` resolved it immediately.
- `tests/unit/test_sdlc_dispatch.py::TestDispatchRecordLease::test_lease_lost_between_peek_and_write_refuses` fails on current `main` in isolation (mock call-sequence assertion on `ledger_key`), reproduced in a clean detached worktree at `origin/main`. Unrelated to #2730's diff; flagging it since it lives on the dispatch-lease path this PR touches.

PR #2746, opened 2026-08-13
Resolve hook validator targets from the hook payload

valorengels:
## Summary

Four `PostToolUse` hook validators decided **which file to validate** by shelling out to `git status --porcelain docs/plans/` and picking the newest-mtime untracked `.md`, discarding the hook payload on stdin that named the file explicitly. Concurrent SDLC lanes in separate worktrees therefore got one lane's write judged against another lane's in-progress plan doc.

This lane extracts the reference fix from PR #2688 into a shared module and ports the whole family onto it. A validator now validates the file the hook payload named, or it validates nothing.

Closes #2738
Closes #2689

## What changed

- **New `.claude/hooks/hook_utils/hook_target.py`** (stdlib only) exposing `target_from_hook_input(hook_input) -> str | None`, `read_hook_input() -> dict`, and `in_scope(path, directory, extension) -> bool`. The contract: `None` means "nothing to validate", never "go find something to validate". Neither working-tree state nor any cwd — the hook process's or the payload's — is an input to target selection.
- **Five validators ported** onto it: `validate_no_gos_justification.py`, `validate_documentation_section.py`, `validate_test_impact_section.py`, `validate_verification_section.py`, `validate_file_contains.py`. All five share one scope predicate.
- **Every guesser deleted**: `find_newest_plan_file` (x3), plus `find_newest_file`, `get_git_new_files`, `get_recent_files`, `get_git_committed_files`, `get_committed_file_content`.
- **`validate_file_contains.py`** uses the resolved payload target as the file to check, keeps `--directory` / `--extension` as the scope filter, and loses `--max-age`, which became meaningless once mtime stopped selecting the target. The manifest's `--contains` arguments are unchanged.
- **Non-dict payload hardening.** Stdin of `null`, `[1,2]`, `"str"`, or `42` parses cleanly and yields a non-dict, on which `hook_input.get(...)` raises `AttributeError` - a traceback out of a hook that gates every write. Both functions guard it.
- **Docs**: new `docs/features/hook-target-resolution.md`, index row in `docs/features/README.md`, the `hook_utils/` shared-module convention recorded in `docs/features/hook-manifest.md`, and refreshed counts in `tests/README.md`.

## Review round: the `cwd` blocker

[...]

PR #2815, opened 2026-08-14
sdlc_progress lane discovery: read recorded identity, not branch shape (#2755)

### LINKED ISSUE #2755: sdlc_progress lane discovery is branch-shape-based, so it cannot see human-named lanes
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/sdlc-progress-lane-discovery-branch-shape.md

Split out of the #2735 / #2718 lane-identity work (plan: `docs/plans/sdlc-lane-recorded-slug.md`). Filed as increment 2 so the deferral is a tracked issue rather than a bare promise.

## Problem

`reflections/sdlc_progress.py` discovers SDLC lanes by matching the *shape* of a branch name, not by reading a recorded identity:

- `_SDLC_BRANCH_RE = re.compile(r"^session/sdlc-\d+$")` (`reflections/sdlc_progress.py:116`)
- `_list_open_sdlc_prs` filters all open PRs through that regex (`:216-235`)
- `_issue_number_from_slug` maps back with `re.match(r"sdlc-(\d+)$", slug)` (`:262`)

#2735 makes `PipelineLedger.slug` the recorded lane identity and establishes that **both** `sdlc-{N}` and human-named slug shapes are real lanes (99 of 451 `session/` heads on origin match `session/sdlc-*`; `.worktrees/sdlc-1920` and `.worktrees/sdlc-1997` sit on `session/dev-81976da0` and `session/dev-7bd4cf82`).

So after #2735 lands, a stalled lane whose recorded slug is human-named is **invisible** to the stall detector. It is the #2718 wedge shape from the discovery direction, and it is silent: the recorded slug exists and looks authoritative, so nothing reports a mismatch.

## Why it was not fixed in #2735

The sibling site in `reflections/sdlc_upvote_lanes.py::_has_pr_on_branch` (`:328-344`) is a per-issue `--head session/sdlc-{issue_number}` probe and was a one-line substitution, so it **was** pulled into #2735. This site is different: it is a corpus filter over all open PRs with a reverse slug-to-issue mapping. Converting it to ledger-driven discovery (enumerate `PipelineLedger` records, read `slug`, build the branch set) is a restructuring of the discovery path, not a substitution, and #2735 is held to a minimum-solution constraint.

## Desired outcome

`sdlc_progress` discovers lanes from recorded identity rather than branch shape:

- [ ] Lane enumeration reads `PipelineLedger` (issue to recorded slug) instead of regex-filtering PR head refs.
- [ ] `_issue_number_from_slug` is deleted or demoted to a fallback: the issue number comes from the ledger key, not from parsing the slug.
- [ ] A stalled lane on a human-named branch (e.g. `session/dev-<hash>`) is detected.
- [ ] `_SDLC_BRANCH_RE` is removed, or retained only as a last-resort fallback for lanes with no ledger record, with a comment saying so.

## Prerequisite

Blocked on #2735 merging: `PipelineLedger.slug` must exist and be populated before discovery can read it.

## Recon Summary

Re-verified against `e50eba258` (the #2735 merge, PR #2792) on 2026-08-13.

**Confirmed**

- `_SDLC_BRANCH_RE = re.compile(r"^session/sdlc-\d+$")` — now `reflections/sdlc_progress.py:118` (issue said :116; +2 drift from #2735).
- `_list_open_sdlc_prs` — now `:217-236`, still filtering every open PR through the regex at `:235`.
- `_issue_number_from_slug` — now `:263-266`, still `re.match(r"sdlc-(\d+)$", slug)`.
- `_slug_from_branch` (`:256-260`) is a third site in the same path: it strips `session/` and is shape-agnostic, so it survives unchanged.
- The consumer is the discovery loop in `_check_project_stalls` at `:811-819`: `branch → _slug_from_branch → _issue_number_from_slug → issue_num`, `continue` on either miss. A human-named lane is dropped silently at `:818`.
- Prerequisite satisfied: `PipelineLedger.slug` exists (`agent/pipeline_ledger.py:200`, docstring `:166-190`) and is minted at lane start by `tools/sdlc_session_ensure.py::ensure_session` via `tools/lane_identity.py::resolve_lane_slug`.
- The create rung already passes a human-named slug through verbatim (`sdlc_progress.py:704-716`, test `test_create_rung_records_a_human_named_slug_verbatim`), and that test's own docstring names the branch filter as "the only thing standing between a human-named lane and a correctly-named session" — this issue is precisely that remaining gap.

**Revised**

- The issue frames the fix as "enumerate `PipelineLedger` records, read `slug`, build the branch set". A cheaper and already-precedented route exists: resolve **PR → issue** through `PipelineLedger.pr_number`, exactly as `tools/merge_predicate.py::_resolve_tracked_issue` does (`:415`, `PipelineLedger.query.filter(pr_number=pr_number)`, repo-scoped, distinct-issue disambiguation). `pr_number` is single-writer, written by `sdlc-tool meta-set --key pr_number` at PR creation. This drops the regex from the corpus filter without a full-keyspace enumeration.

**Pre-requisites / constraints surfaced**

- Ledger lookup is repo-scoped and needs `target_repo`, which comes from `_project_repo(project)` (`:771`) and is `None` for a project with no `github` block. Discovery must define behavior for `target_repo is None` rather than issuing an unscoped query.
- The stale comment at `:762-770` asserts "the branch filter admits only issue-derived names" as the reason the `None`-repo hazard is unreachable. Broadening discovery makes it reachable; the comment must be rewritten, not just left.
- Lanes predating `pr_number`/`slug` population resolve to nothing, so a shape fallback (or an explicit skip) is a decision the plan must make, not an accident.

**Test surface**

- `tests/unit/reflections/test_sdlc_progress_check.py:1259` (`test_gh_pr_list_filenotfound_returns_empty`) and `:1266-1276` (`test_gh_pr_list_filters_non_sdlc_branches`) assert the regex behavior directly — both need dispositions.
- The comment block at `:579-590` and the docstring at `:618-624` describe the filter as the boundary; both go stale.
- `tests/integration/test_sdlc_stall_auto_resume_e2e.py:78-82` monkeypatches `_list_open_sdlc_prs` with a 3-key PR dict; a signature or contract change ripples here.
- No `xfail` markers relate to this bug.

**Dropped**

- Nothing. All four acceptance-criteria bullets in the Desired Outcome remain in scope.

valorengels:
## What

The SDLC stall detector discovered lanes by pattern-matching branch names against `^session/sdlc-\d+$`. Measured against production at plan time, **zero of seven open PRs matched** — the detector was not degraded, it was blind to the entire lane population, and silent about it, because a lane that never enters the corpus is never counted.

Discovery now reads recorded identity and links that exist in the world.

- Corpus is the `session/` namespace; the shape of the rest is not consulted. The old branch-shape filter is deleted outright.
- New `_resolve_lane_issue` — a three-rung issue resolution ladder: recorded ledger by `pr_number` → the PR's own closing reference (repo-matched, single distinct entry required) → issue-derived slug shape, demoted to last because it is the only rung that *derives* rather than reads.
- Ambiguity is reported, never guessed: two or more candidates at any rung yields `gate-unknown: issue-ambiguous` and no action. PR #2746's two closing references are the real-world case this guards.
- A rung that *errors* is distinguished from one that *declines*. On error, steer and resume may proceed but the create rung is suppressed, because create is the only action that writes permanent identity via `adopt_lane_slug` and that write is no-overwrite.
- Burst guards for a corpus going from zero to the full live population in one tick: a per-tick target dedupe (the primary guard — `_pick_steer_target` is project-wide, so several lanes otherwise converge on one session) plus `SDLC_STALL_ACTIONS_MAX_PER_TICK`.

## Notable design decisions

**A ledger-by-recorded-slug rung was considered and cut.** It cannot fire for the lanes this fix exists to serve: a human-named lane's minter records `sdlc-{N}` and the write is no-overwrite, so `ledger.slug == branch slug` is permanently false for exactly those lanes. Where it does hold, another rung already answered.

**Rung selection moved above the cooldown claim.** The dedupe needs the target session id and the create suppression needs the rung kind, neither of which exists until selection runs. The two pre-dispatch `_action_cooldown_release` calls became dead and were removed; the post-dispatch ones and the function survive. Without this, a lane deferred by a brake would wait an hour rather than a tick.

## Verification

- 110 tests pass (`tests/unit/reflections/test_sdlc_progress_check.py`, `tests/integration/test_sdlc_stall_auto_resume_e2e.py`), `POPOTO_TEST_DB=12`. `PipelineLedger.query` is monkeypatched in unit tests — no real Redis.
- Written red first: the human-named-lane regression test failed with `findings: []` against the unmodified module, which is the bug's real signature.
- `ruff check` and `ruff format --check` clean.
- All nine plan Verification rows pass. The raw-Redis row was **demonstrated red** against an injected violation and green after revert — two earlier drafts of that row were vacuous (escaped alternation pipes under `grep -E` match nothing), so it was not trusted until shown to fail.

Two prerequisite spikes ran before build: all open lane branches resolve locally (no fetch step needed), and closing-reference objects carry `repository.owner.login` + `repository.name`, so the repo-match rule needs no URL parsing.

Plan: `docs/plans/sdlc-progress-lane-discovery-branch-shape.md` (four critique rounds).

Closes #2755


valorengels:
## Review: APPROVE

Reviewed against `docs/plans/sdlc-progress-lane-discovery-branch-shape.md` (four critique rounds). No blockers.

**The five requirements the critique rounds fought hardest over are all implemented as specified, and pinned by tests that assert behavior rather than absence of exceptions:**

- **Rung order enforced, deriving rung genuinely unreachable when a read rung answers.** Pinned by a test that monkeypatches `_issue_number_from_slug` to a recorder and asserts it was never consulted — not by inspection.
- **Set-valued ledger map.** `mapping.setdefault(int(pr_number), set()).add(...)`; two records on one `pr_number` produce `issue-ambiguous` and no action, rather than a silent overwrite into an uncorrectable `adopt_lane_slug` write.
- **Repo-matched closing reference.** Composed from `repository.owner.login` + `repository.name`; a cross-repo reference does not resolve, and the no-`target_repo` test asserts the ledger query was never *called*, which is the no-unscoped-query property rather than just the outcome.
- **Declined vs. errored rungs.** `None` from the enumeration is the error signal, `{}` a clean decline; a `target_repo is None` skip correctly does not look like an error. On error the create rung is suppressed and `gate-unknown: ledger-degraded` is emitted.
- **Every brake above the cooldown claim.** `_pick_steer_target` hoisted above `_action_cooldown_set`; dedupe, cap, create brake, and degraded-create suppression all precede it. Traced every path between the claim and `_attempt_action`: nothing can `continue` there, so no path claims the one-hour cooldown and then defers. The capped-lane test runs a *second* tick and asserts both lanes steer, which is the real property rather than mere key absence.

**Two review findings were raised and are fixed in `2d5edaab4`:**

1. The `ledger-degraded` finding was appended *after* the unresolved bail, so the single lane where the degradation actually changed the outcome — nothing resolved, ledger dark — was the one lane that never reported it. Given that the plan's opening thesis is that a silently-dropped lane is half the bug, this mattered.
2. The per-record guard in the map builder skipped an unreadable row and logged at `debug`, producing a silently incomplete map indistinguishable from a clean decline. A decline does not suppress the create rung, so a weaker rung could have written a permanent identity — the exact Risk-3 shape the plan closed. It now degrades the tick.

Both fixes are covered by tests demonstrated **red** against the unpatched module before being accepted.

**Remaining non-blocking notes:** documentation is the pipeline's next stage — `pm-session-liveness.md` still describes the branch-shape gate and the pre-selection cooldown claim, both now false. `_GH_PR_LIST_LIMIT = 200` is a bare constant without an env override, unlike its neighbours in the thresholds block. A lane already inside its cooldown window can now emit an `action-cap:` finding it would previously have skipped silently — cosmetic finding noise, and a direct consequence of the ordering the plan required.

Verified independently of the build report: 112 tests pass (`POPOTO_TEST_DB=12`), ruff check and format clean, all nine plan Verification rows pass, and the raw-Redis row was demonstrated red against an injected violation and green after revert — two earlier drafts of that row were vacuous, so it was not trusted until shown to fail.

REVIEW_CONTEXT head_sha=2d5edaab47a6f94f79c07603616e113c4508d5bc

valorengels:
## Review: APPROVE

Re-review at the final head. The delta since the code review at `2d5edaab4` is documentation plus one stale inline comment; the implementation is unchanged and its approval stands.

**Documentation delta verified against the shipped code**, claim by claim: the `session/` namespace corpus, the three-rung ladder and its order, the `{pr_number: set[int]}` per-tick map, ambiguity declining rather than guessing, the errored-versus-declined distinction with create-rung suppression, rung selection preceding the action-cooldown claim, the four same-tick brakes, and the gate order (staleness, then resolution, then issue-open). Every previously-falsified statement in `pm-session-liveness.md` is rewritten, the `SDLC_STALL_ACTIONS_MAX_PER_TICK` row is present in the tunables table, and the cross-reference to `sdlc-lane-identity.md#discovery-reads-identity` resolves to a real heading.

**Two findings were raised during the docs review and fixed:**

1. The "Failure tolerance" bullet still described a locally-absent branch as silently skipped, which contradicted the gate list two sections above it and the shipped `gate-unknown: branch-not-fetched` finding. That contradiction mattered more than most: an unfetched ref is the one condition that could quietly return this detector to seeing nothing, which is the exact failure this PR exists to end.
2. A sentence narrating the doc's own prior state ("this example is no longer hypothetical") was replaced with a plain statement of current behavior, per the repo's rule that docs describe only the status quo.

A stale inline comment at the `_last_commit` skip ("silent skip per success criteria") was also corrected: the caller now converts that `None` into a reported finding, so only the git detail is debug-level. Comment-only, no behavior change.

Final verification at this head: 112 tests pass (`POPOTO_TEST_DB=12`), `ruff check` and `ruff format --check` clean, and the branch touches exactly five files (one module, two test files, two docs).

REVIEW_CONTEXT head_sha=a0cd7dc683ce8c9d79e650dc6accd4ac401aff1d

Issue #2829, opened 2026-08-17
Concerns-scoped re-critique pass: judge the revision diff, not the whole plan


valorengels:
Deferred from #2787 (task 9). #2787 makes a `READY TO BUILD (with concerns)` verdict route its concern-closing revision back through `/do-plan-critique` before build, bounded by `MAX_CONCERN_RECRITIQUE_ROUNDS` (default 3). Every one of those rounds currently runs a **full-depth** war room.

That is correct but expensive: a lane that hits the bound pays 3 full war rooms. #2787 deliberately shipped correctness first and deferred the cost work here, because the scoped pass was the largest new surface in that plan, had no test, and its escalation trigger was prose judgement handed to an LLM.

## What to build

A reduced-scope critique pass for re-critique rounds only: judge the **prior round's concerns plus the revision diff**, not the whole plan.

## The escalation trigger must be mechanical

This is the part that sank the first attempt. Do not use "the revision touches the plan's structure" as an LLM judgement — it is unassertable. Use:

> Compute the set of `^## ` / `^### ` headers whose bodies changed in the plan diff between the prior verdict's commit and HEAD. Escalate to FULL depth when that set intersects `{Solution, Technical Approach, Step by Step Tasks, Verification}`.

That is assertable over two plan-file fixtures, which is the acceptance criterion.

## Constraints carried forward from #2787

- **The saving must come from narrowing *what is judged*, not from thinning the roster.** A two-critic war room that misses a round-2-grade defect costs more than it saves. #2787's own round 2 found three blockers that a thinned roster would plausibly have missed.
- **Skill-context split.** `/do-plan-critique` is a global skill (`.claude/skills-global/`). The generic probe belongs in the global body; the concrete `sdlc-tool stage-query` invocation belongs in `docs/sdlc/do-plan-critique.md`.
- The bound itself (`MAX_CONCERN_RECRITIQUE_ROUNDS`) is orthogonal — #2787 dropped its default from 4 to 3 specifically to absorb full-depth cost. If the scoped pass lands, revisit whether 3 is still the right number.

## When to pick this up

#2787's D3 says: revisit when a with-concerns lane's war-room cost is measured and shown to be the binding constraint, not before.

Refs #2787

valorengels:
Duplicate of #2828 — consolidating there.

[...]

Issue #2049, opened 2026-07-13
Router NEEDS-REVISION deadlock: /do-sdlc router ignores revision_applied and re-dispatches /do-plan on a stale verdict


tomcounsell:
> **Context.** `/do-sdlc` is driven by a single-stage **router** (`.claude/skills/sdlc/`) that, on each turn, reads the pipeline ledger for an issue and dispatches exactly ONE sub-skill (plan, critique, build, review, patch, docs, merge). After a plan is revised in response to a critique, the ledger sets `revision_applied: true` on the plan-stage meta. The router is supposed to consume that flag so a freshly-revised plan advances to a **re-critique**, not back to planning.

## Problem

When a plan critique returns **NEEDS REVISION** and the plan is then revised, the router keeps re-dispatching `/do-plan` (or loops PLAN↔CRITIQUE) on the **stale** `NEEDS REVISION` verdict instead of advancing, because it does not treat `revision_applied: true` as invalidating the prior verdict. The pipeline cannot converge to BUILD without manual intervention.

### Current behavior
- Critique records `latest_critique_verdict = "NEEDS REVISION"`.
- The plan is revised; ledger meta gets `revision_applied: true` (+ `revision_applied_at`).
- Next router turn still sees the `NEEDS REVISION` verdict and routes to `/do-plan` again — the revision does not clear or supersede the verdict. Result: a deadlock the operator escapes only by dispatching `/do-plan-critique` directly, out of band.

### Desired outcome
The router treats `revision_applied: true` (with `revision_applied_at` newer than the recorded verdict) as "the plan has changed since the last verdict → the verdict is stale," and dispatches a fresh **critique** rather than re-planning. Convergence to BUILD requires no manual escape.

## Recurrence evidence

This exact deadlock was investigated in **#1760** ("investigation: /do-sdlc PLAN↔CRITIQUE router never converges to BUILD (notes-only revision re-stales a clean verdict)"), which is **CLOSED as COMPLETED** — yet the behavior persists. During a 2026-07-12/13 multi-wave push it recurred on **two separate lanes**:
- **#1925** (PydanticAI standardization): after each of two plan revisions, the router re-dispatched `/do-plan` on the stale verdict; escaped both times via a direct `/do-plan-critique` dispatch.
- **#1968** (magic-literal centralization): same latch after a plan revision; same manual escape.

A supervisor memory note (`project_sdlc_router_needs_revision_deadlock`) documents the workaround, which is a smell that the root cause is unfixed.

[...]

Issue #1871, opened 2026-07-02
SDLC router G5 fast-path dispatches /do-build while plan_revising=true (ships pre-revision design)


tomcounsell:
> **Context.** The /sdlc router's G5 fast-path dispatches `/do-build` when a cached READY-TO-BUILD critique verdict exists and the plan hash is unchanged. Observed live on issue #1821 (2026-07-03): critique recorded READY TO BUILD (WITH CONCERNS) but deliberately set `plan_revising=true` and left the CRITIQUE marker in-progress to request a concern-embedding revision pass. `sdlc-tool next-skill` nonetheless returned `/do-build` via G5 — building at that moment would have shipped the flawed pre-revision design (a hook writing `status=paused` racing the granite adapter's partitioned update_fields writes; concerns not embedded). A supervising agent caught it manually and dispatched `/do-plan` instead.

## Problem
G5 keys on (verdict string, plan hash) alone and ignores `plan_revising`. Sibling defect to the known NEEDS-REVISION router deadlock (stale-verdict loop) — same root shape: router trusts one cached signal over the composite state.

## Desired outcome
G5 (and any verdict-cached fast-path) treats `plan_revising=true` as a hard gate: route to `/do-plan` (revision) until the revision commits and clears the flag.

## Recon Summary
**Confirmed:** live occurrence on #1821, 2026-07-03 (stage state: verdict READY TO BUILD WITH CONCERNS + plan_revising=true + CRITIQUE marker in_progress → next-skill returned /do-build). **Revised:** nothing. **Pre-requisites:** none. **Dropped:** nothing — single-line-of-logic class fix in the router's gate evaluation, plus a regression test on the composite state.

tomcounsell:
Reproduction datapoint from the 2026-07-05 six-issue parallel SDLC batch: in the #1904 run, G5 fast-pathed CRITIQUE→BUILD while `plan_revising=true` (supervisor manually carried the critique concerns into the build prompt; G7 later blocked at REVIEW on the same stale lock and it was cleared per the guard's remediation). In the same batch, the #1902 and #1905 runs took the correct path on the identical router version (revision → mandatory re-critique → build; #1905 cycled row 2b twice cleanly). So the bug is intermittent, likely dependent on the state snapshot G5 reads at dispatch time rather than a deterministic table error.

valorengels:
[...]

Issue #1760, opened 2026-06-22
investigation: /do-sdlc PLAN↔CRITIQUE router never converges to BUILD (notes-only revision re-stales a clean verdict)


tomcounsell:
## Observed behavior (unverified root cause)

During two separate `/do-sdlc` runs in the **cuttlefish** repo on 2026-06-22 (issues yudame/cuttlefish#547 and yudame/cuttlefish#550, each in its own worktree), the supervisor **never converged from PLAN/CRITIQUE to BUILD on its own**. Both subagents reported the same thing and had to **manually drive the BUILD stage** against a plan that was already marked build-ready, with **zero code-correctness blockers** ever raised. Quoting the run reports:

> "the PLAN↔CRITIQUE router never converged to BUILD — every notes-only plan edit re-staled the critique verdict, and FULL critiques on a `.claude/skills`-path plan kept minting non-blocking prose nits. Zero code-correctness blockers ever appeared."

> "a READY-TO-BUILD verdict with zero blockers shouldn't be invalidated by a subsequent notes-only `/do-plan` write."

Net effect: wasted PLAN→CRITIQUE→PLAN dispatch cycles, and the pipeline only progressed because a human/supervisor overrode it. This defeats the point of autonomous `/do-sdlc`.

## Why this is filed as `investigation`, not `bug`

This part of the router has a long lineage of dead-end fixes, so the specific mode below may be a partial regression or an uncovered edge — root cause needs confirmation before any change:

- `3e1e3dae` Fix SDLC router dead-end: CRITIQUE in_progress with empty verdict (#1668)
- `6e943ea9` Fix SDLC router stale-critique dead-end (#1639)
- `5bc6243a` verdict normalization + plan-existence gate + stale-verdict supersession (#1638/#1640/#1641)
- `8218c5af` Guard rule 4b against re-firing once a PR exists (#1554)
- `627e3cf0` row 8c REVIEW empty-verdict re-dispatch (#1755)

## Plausible mechanism (from static reading — to be confirmed)

The staleness check is purely timestamp-based:

- `_critique_verdict_is_stale()` — `agent/sdlc_router.py:750-783` — returns True whenever the CRITIQUE verdict's `recorded_at` predates the **latest `/do-plan` dispatch**, regardless of whether that plan edit was substantive.
- Row 2b `_rule_critique_verdict_stale` — `agent/sdlc_router.py:786-800` — on stale verdict, re-dispatches `/do-plan-critique`.
[...]

PR #2830, opened 2026-08-17
Route a with-concerns verdict's concern-closing revision back through critique, bounded (#2787)

### LINKED ISSUE #2787: A "READY TO BUILD (with concerns)" critique verdict does not mandate a re-critique, so the revision that closes the concerns is never itself reviewed

valorengels:
Closes #2787

`/do-plan-critique` can return **READY TO BUILD (with concerns)**. The pipeline responded by dispatching a `/do-plan` revision to embed each concern's Implementation Note, then dispatching `/do-build`. Nothing ever looked at that revision — the part of the plan written last, under the least scrutiny, specifically to close findings critics flagged as risky, and the only part never reviewed.

Worse, `revision_applied` is sticky: `/do-plan` sets it on every pass and it never resets. So from round 2 onward row 4b could never fire and row 4c fired unconditionally. **The more critique rounds a plan needed, the less each round was reviewed.**

## This bug reproduced itself twice during its own pipeline run

Not a hypothetical. On run `dfccc08fcf45`, driving this very issue:

1. `/do-plan-critique` Step 5.6 skipped setting the `plan_revising` lock because `revision_applied: true` was already in the frontmatter — the sticky-boolean exemption.
2. After the round-3 verdict was recorded, `sdlc-tool next-skill` returned `{"skill": "/do-build", "row_id": "G5", "reason": "G5: cached CRITIQUE verdict is READY TO BUILD on unchanged plan hash"}` — the G5 cache shipping a build straight off a `READY TO BUILD (WITH CONCERNS)` verdict, exactly the bypass §4.5 of the plan calls co-load-bearing.

## The shape of the fix

The re-critique edge was never missing. **Row 2b already routes a stale CRITIQUE verdict back to `/do-plan-critique`** (#1639). #1760 switched it off for the with-concerns settle path because the loop had no terminating bound, and #2049 then narrowed that latch *away* from NEEDS REVISION, leaving `READY TO BUILD (with concerns)` as its live domain. So the work is **supply the bound, then let row 2b run**.

- **Bounded with-concerns branch in `_critique_verdict_is_stale`.** Below the bound the latch stands down; at the bound it engages permanently. Placed *ahead* of the `if not latest_plan_at: return False` early return so the whole decision is independent of the FIFO-truncated `_sdlc_dispatches`.
- **`_concern_round_count`**, a monotonic integer written by `record_verdict` inside the transaction it already owns, surfaced as `_meta["concern_round_count"]`. It **must** ride `_meta`: `stage-query` threads only `("_verdicts", "_sdlc_dispatches")` into `stages`, so a bare `_*` key would be dropped and the feature would pass every unit test while being inert in production.
- **`MAX_CONCERN_RECRITIQUE_ROUNDS`** (default 3, `os.getenv`-overridable; `0` is the kill switch restoring today's routing).
- **Rows 4b/4c re-keyed** off the sticky boolean onto the event-scoped predicate. Row 4c becomes the bound-exhausted build edge. **No new dispatch row** — `len(DISPATCH_RULES)` stays 19.
- **G5 steps aside unconditionally on with-concerns.** Guards run before the dispatch table, so without this rows 2b/4b/4c are unreachable. Unconditional because an earlier design gated it on a predicate that is `False` in exactly the state the guard was written to prevent. `agent/sdlc_router.py`'s #1871 short-circuit is untouched.
- **Step 5.6 and G7 gate 3 repaired** in both the global skill body and `docs/sdlc/do-plan-critique.md`, so the `plan_revising` lock is armable per round instead of permanently inert.

## Why the below-bound arm is not a timestamp fallthrough

Round 3's critique caught this. `plan_dt` comes from `_latest_dispatch_at`, stamped when `/do-plan` is **dispatched**, not when it finishes. A fallthrough would call the verdict stale the instant row 4b fired, and row 2b — registered ahead of 4b/4c and matching on staleness alone — would re-critique a plan nobody had revised, burning a bound slot whenever `/do-plan` crashed. The arm returns `_concern_revision_is_unjudged(...)` instead, so row 2b owns state S2 and only S2.

## Two things that are not what they look like

**G4 cannot bound this loop.** `compute_same_stage_count` counts *consecutive same-skill* dispatches and breaks its streak on any skill change; this loop alternates two skills, so the streak resets every turn. The counter is the only terminator — which is why there is deliberately **no dedupe** on it. Over-counting costs a build with a recorded acceptance one round early; under-counting costs an unbounded loop.

**The predicate is verdict-kind-agnostic on purpose.** G7 gate 3 calls it without a verdict test, because the lock it self-heals may have been left by a NEEDS REVISION round. Moving the `WITH CONCERNS` test into the body would make gate 3 stall those lanes at `Blocked`.

## Demonstrated red

Two pre-existing tests flipped, captured before the change:

- `test_1760_inverse_guarantee_preserved` asserted `_critique_verdict_is_stale(...) is False` and `/do-build` on a with-concerns settle. It now asserts `True` → row 2b below the bound, with a sibling pinning that the inverse guarantee is *restored* at the bound.
- `TestRow4cReadyWithConcernsRevisionApplied::test_concerns_with_revision_flag_proceeds_to_build` asserted the sticky-boolean path to build — the defect itself.

Every no-concerns latch test (`TestConvergenceLatchRevisionAppliedAt`, `test_sdlc_stage_query.py`) stayed green **unchanged**, which is the proof the change did not leak.

The G5 alive test covers five states, each asserting `row_id != "G5"`: S1 unlocked → 4b; S1 with the Step 5.6 lock → `/do-plan` via 4b-or-G7 (G7 gate 4 legitimately owns that turn); S1 after a `/do-plan` dispatch that landed no revision → 4b; S2 below bound → 2b; S2 at bound → 4c.

## Verification

- 383 tests pass across the six affected suites; `ruff check` and `ruff format --check` clean.
- `len(DISPATCH_RULES) == 19`, `2b` ordered ahead of `4b`/`4c` (pinned by a test — the design depends on first-match ordering).
- Step 5.6 exemption gone from both files; the global `/do-build` body carries no `sdlc-tool` invocation or repo key names.
- Kill switch and G4-inertness both pinned by tests.

## Follow-up

The concerns-scoped (cheaper) re-critique pass is deliberately deferred to #2829 with a mechanically-assertable escalation trigger. Correctness first, then cost.


valorengels:
## Review: APPROVED — 0 blockers, 0 tech debt (all 6 findings closed in 80da2127)

Reviewed by an independent `code-reviewer` agent against the plan, then the findings were fixed in this branch. Head SHA `80da212751d13448d1e957406c0ed48fe89f206f`, confirmed identical from `tools/pr_head_resolver.py::resolve_pr_head_sha` (git-first) and `gh`. Posted as a comment rather than a GitHub review because GitHub refuses self-approval; self-merge is authorized for this lane.

### The seven correctness claims were interrogated against the code, not the prose

**Termination holds.** The counter advances on every recorded with-concerns round (`tools/sdlc_verdict.py:504`, post-`normalize_verdict` so casing is safe). Every non-advancing state turns out to be a *same-skill repeat* and therefore visible to G4: a `/do-plan-critique` crash leaves `recorded_at` stale so row 2b re-fires the same skill; a `/do-plan` crash leaves S1 so row 4b re-fires `/do-plan`. The pair {bound, G4} covers the space. `TestForeverWithConcernsTerminates` is a turn-by-turn simulation against the live router rather than a mock, and `test_g4_cannot_see_this_loop` pins the premise the bound rests on. Durability confirmed at `agent/pipeline_state.py:544-551`; no reset or decrement path exists.

**The below-bound arm is right, and the reasoning is not hand-waving.** `_latest_dispatch_at` stamps at dispatch time, so a fallthrough really would make `verdict_dt < plan_dt` true the instant row 4b fires. `test_a3_plan_dispatched_but_no_revision_landed_stays_on_row_4b` is the falsifying test for exactly that state.

**Placement verified**: the branch at `agent/sdlc_router.py:1226` sits above the `latest_plan_at` lookup at `:1262`, so the with-concerns decision is genuinely independent of the FIFO-truncated `_sdlc_dispatches`.

**The `_meta` plumb is real, not inert.** Present in both `_compute_meta` and `_default_meta`, with key-set equality asserted so #2769 parity is machine-enforced. Both `decide_next_dispatch` call sites source meta from `query_enriched`, so no third producer starves the counter.

**G5's no-concerns path is intact** and the #1871 short-circuit is byte-identical. **G7 gate 3 is called bare**, with a NEEDS REVISION test proving the self-heal still fires. **`len(DISPATCH_RULES) == 19`** with `2b` ahead of `4b`/`4c`.

**Test honesty checked.** `TestConvergenceLatchRevisionAppliedAt` is untouched by the diff and green. Both intentional flips are honest: the #1760 test was inverted *and* paired with a sibling proving the guarantee moved to the bound rather than being deleted.

### The six findings, all closed

1. `docs/features/sdlc-pipeline.md` was a half-migration inside one document — the Convergence Latch section had been updated but the G7 section still documented the deleted sticky-boolean mechanism, including a verbatim code block of the removed gate. This was the highest-value fix: an agent reading it would re-derive the behavior this PR removes.
2. `guard_g7_plan_revising`'s docstring still said "revision_applied is the source of truth", contradicting its own rewritten inline comment.
3. / 4. `_rule_critique_verdict_stale`'s docstring and the row 2b registration comment both named G5 as this row's loop bound — true for no-concerns, false for with-concerns. These are precisely the two places a future reader would check before deleting `MAX_CONCERN_RECRITIQUE_ROUNDS` as redundant, so they now say so explicitly.
5. `test_needs_revision_rounds_do_not_consume_the_bound` was a verbatim duplicate of the test twelve lines above it, wearing a load-bearing name. Rebuilt with a real five-turn NEEDS REVISION dispatch history and renamed.
6. The mixed-timezone path was unexercised: `recorded_at` is always tz-aware while `revision_applied_at` comes from hand-editable frontmatter and may be naive, so the comparison raises `TypeError` and is swallowed to `False`. Now pinned by a test showing the lane falls safe to a revision pass, never a build.

### Verification

453 tests green across the seven affected suites before the fixes; 107 + 65 green in the two files the fixes touched. `ruff check` and `ruff format --check` clean. No legacy residue, no parallel-run shims, no `revision_applied` fallback left beside the event-scoped test. `agent/sdlc_router.py` stays free of `tools/` imports.

Issue #1313, opened 2026-05-06
Stalled sessions emit warnings only to log files — no user-visible alert


valorengels:
> **Context:** This repo's [`monitoring/session_watchdog.py`](monitoring/session_watchdog.py) periodically scans `AgentSession` records for stalled work. When a session sits in `pending` (or any non-terminal state) past a threshold without progressing, it emits a `LIFECYCLE_STALL` warning to the worker log. The warning includes session id, status, duration, threshold, project key, and last lifecycle history entry. **It does nothing else.** The user gets no signal on Telegram. On 2026-05-06, session `tg_cuttlefish_-5295380350_9642` (a "configure customer" request for Graham Derry in the Cuttlefish project) sat pending for over 5 hours; the watchdog logged 60+ stall warnings; the user never knew anything was wrong because the message looked accepted on Telegram and silence looked like "agent is thinking."

## Problem

`LIFECYCLE_STALL` warnings are emitted at WARNING level into a log file no human reads. There is no user-visible signal when a session is stalled. This silent-failure pattern lets short outages compound into long ones because nothing tells the user to investigate.

**Current behavior:**
- [`monitoring/session_watchdog.py:368-376`](monitoring/session_watchdog.py) emits the `LIFECYCLE_STALL` warning via `logger.warning(...)` whenever a session's duration exceeds `threshold` (default 300s for `pending`).
- The watchdog tick only logs. It does not message the user. It does not react on the Telegram message. It does not page anyone.
- AgentSession already carries enough state to look up the originating Telegram message: `session.chat_id` and `session.telegram_message_id` (via `initial_telegram_message` dict; see `models/agent_session.py`).
- The bridge already exposes a reaction-emoji API: `set_reaction(client, chat_id, msg_id, emoji)` at [`bridge/response.py:258-321`](bridge/response.py).

**Desired outcome:**
- When a session is stalled past its threshold, the watchdog (or the bridge, called by the watchdog) attaches a reaction emoji to the originating Telegram message — something like ⏳ for "stalled" or ⚠️ for "needs attention."
- (Stretch, only if surgically simple) After 30 minutes stalled, post a single status reply in the dev chat naming the session and the duration. One reply, not a stream.
- The watchdog log behavior is preserved — this is an *additional* user-visible channel, not a replacement for the log.

## Definitions

| Term | Definition | Reference |
|------|-----------|-----------|
[...]

Issue #1312, opened 2026-05-06
Bridge accepts work into the queue without verifying a worker exists to pick it up


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/bridge-worker-liveness-reaction.md

> **Context:** This repo runs a Telegram bridge ([`bridge/telegram_bridge.py`](bridge/telegram_bridge.py)) that accepts incoming messages, classifies them, and enqueues an `AgentSession` (a Redis-backed Popoto record) for a separate worker process (`python -m worker`) to pick up and execute. The bridge has no SDLC awareness — its job is I/O routing only. **The bridge does not currently check whether the worker is alive before accepting work.** When the worker is dead, messages enter the queue and rot silently; from the user's perspective the message looks "received and being thought about." This is the silent-failure mode that turned a 30-second outage into a 7-hour one on 2026-05-06 (see `tg_cuttlefish_-5295380350_9642`).

## Problem

The bridge accepts work into the Redis session queue without verifying that any worker is alive to drain that queue. The result: Telegram users see no signal that their request is going nowhere.

**Current behavior:**
- A Telegram message arrives at the bridge handler in [`bridge/telegram_bridge.py:1005`](bridge/telegram_bridge.py).
- The bridge calls `dispatch_telegram_session(...)` (one of the call sites at lines 1665, 2193, 2337) which writes an `AgentSession` to Redis and exits.
- If the worker process is dead, the session sits in `pending` indefinitely. Nothing happens. The user sees nothing happen.
- The bridge **already** uses Telegram reaction emoji elsewhere (e.g. `set_reaction(client, event.chat_id, message.id, ...)` at lines 802, 850, 863, 1357, 1372, 1377, 1458, 1891, 2306, 2335 — implementation in [`bridge/response.py:258-321`](bridge/response.py)). The capability exists; it's just not wired to liveness.

**Desired outcome:**
- When a Telegram message arrives and **no worker is alive for that message's `project_key`**, the bridge immediately reacts on the message with a visible warning emoji (e.g. ⚠️ or 🚧 — exact emoji TBD by the planner).
- The message is still enqueued — the user shouldn't lose work — but the reaction makes it obvious that processing is paused, not in progress.
- When a worker eventually picks the session up, the warning reaction can be cleared (or replaced with a normal "processing" reaction) — this is a stretch goal.

## Definitions

| Term | Definition | Reference |
|------|-----------|-----------|
[...]

PR #2835, opened 2026-08-17
Replace the dead stall reaction with a session liveness tick counter (#2716)

### LINKED ISSUE #2716: Session liveness feedback: rewrite the non-working stall reaction as an advancing progress counter with a forced-progress ceiling
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/session-liveness-tick-counter.md

> **Context:** Agent work in this repo runs as `AgentSession` records executed by a standalone worker, while a separate [Telegram bridge](bridge/telegram_bridge.py) process owns all chat I/O. Inside the bridge runs the [session watchdog](monitoring/session_watchdog.py), a loop that inspects sessions for distress and actuates on them. Telegram lets a bot set one *reaction* (an emoji attached to an existing message, replaceable in place, and delivered without a notification) per message — the system's established low-noise "I heard you" channel.

## Problem

When a session runs long, the human on the other end has no way to tell "still working" from "wedged" without asking. The system has exactly one mechanism intended to close that gap, and it has never worked.

**Current behavior:**

1. **The stall reaction is an invalid emoji and always fails.** [#1313](https://github.com/tomcounsell/ai/issues/1313) added `_apply_stall_reaction`, which queues a ⏳ reaction on the user's originating message when a session is observed stalled. But ⏳ is not a legal Telegram reaction. It is absent from the 74 reactions the server advertises via `messages.getAvailableReactions`, and this repo's own [`INVALID_REACTIONS`](bridge/response.py) list already contains it. Every attempt fails at the API. Production evidence from `logs/bridge.log`:

   ```
   2026-08-09T15:47:55Z WARNING bridge.telegram_relay _send_queued_reaction
   Relay: failed to set reaction ⏳ on msg 1308 in chat -1003449100931
   ```

   The same log shows 11 `Stall reaction queued` events. None of them landed. The dedup key (`watchdog:stall_reaction_applied:{session_id}`) is set on *queue*, not on *delivery*, so a failed reaction is never retried for that stall period.

2. **Nothing distinguishes "slow" from "dead" even if the emoji worked.** A single static reaction is a one-shot signal. It cannot show that time is passing or that work is ongoing, and it never escalates — a session can sit behind it indefinitely.

3. **A self-reported signal would not fix this.** The natural instinct is to have the session announce its own liveness. This system has already rejected that pattern: [`bridge/liveness.py`](bridge/liveness.py) documents it as *"a handler that has stopped firing cannot testify to its own failure."* The wedged case — a hung tool call inside a turn — produces no turn boundary, so a session-emitted heartbeat goes silent exactly when the signal matters most.

**Desired outcome:**

A long-running session shows a visibly advancing liveness signal that is emitted by an observer independent of the session, and that cannot run forever without the session producing substantive evidence of progress.

## Definitions

| Session watchdog | A loop running inside the bridge process that scans `AgentSession` records for stalls, loops, and error cascades, and actuates (steers, reacts, escalates) | [`monitoring/session_watchdog.py`](monitoring/session_watchdog.py) |
| Reaction | An emoji attached to an existing Telegram message. Replaceable in place, sends no notification, one per sender per message | [`set_reaction`](bridge/response.py) |
| Custom emoji reaction | A reaction identified by a `document_id` from a sticker set rather than a Unicode character. Requires a Telegram Premium sender; renders for all viewers | `ReactionCustomEmoji` in [`set_reaction`](bridge/response.py) |
| Outbox relay | The Redis list `telegram:outbox:{session_id}` that non-bridge processes write to so the bridge delivers messages and reactions on their behalf | [`bridge/telegram_relay.py`](bridge/telegram_relay.py) |
| PM session | The `eng`-role session that orchestrates SDLC work and speaks to the human | [`docs/features/eng-session-architecture.md`](docs/features/eng-session-architecture.md) |

## Prior Context

- [#1313](https://github.com/tomcounsell/ai/issues/1313): Added the ⏳ stall reaction — the feature this issue repairs and extends. Shipped non-functional.
- [#2663](https://github.com/tomcounsell/ai/issues/2663): Read-only session-progress CLI, so a human or agent can *ask* whether a session is working. Complementary: that issue is the pull interface, this one is the push signal. Both should read the same underlying progress truth rather than inventing two.
- [#2494](https://github.com/tomcounsell/ai/issues/2494): Durability refactor keying recovery on work owed rather than process status — same principle applied to a different surface.

## Solution Sketch

A tick counter, owned by the session watchdog, mutating one reaction slot on the session's originating message.

- The watchdog already has everything required: it runs in the bridge process, and `AgentSession` exposes `chat_id` and `telegram_message_id`. It already queues reactions through the outbox relay, and that relay already accepts a `custom_emoji_document_id` field.
- Ticks advance on a fixed interval (proposed: 10 minutes per slot). Slot 0 is an acknowledgement; slots 1-9 show a counting digit, so elapsed time is readable at a glance.
- Digits are only expressible as Premium custom emoji. Each tick therefore pairs a `document_id` with a standard-emoji fallback, so the counter degrades to a plain alternating arc rather than disappearing when custom reactions are unavailable.
- **The ceiling is the point.** After the last tick (proposed: 100 minutes of wall clock), the counter must refuse to advance. The session is required to publish a substantive progress message, and a fresh counter attaches to that new message. A heartbeat that can run forever is a hang with better manners.
- The tick must reflect observed evidence the watchdog can see independently (turn transitions, worktree mtimes, session state), not a self-report from the session.

**In scope: rewriting the #1313 stall-reaction feature, not building alongside it.**

The counter replaces `_apply_stall_reaction` and its supporting machinery (`STALL_REACTION_EMOJI`, the `watchdog:stall_reaction_applied:{session_id}` dedup key, `_clear_stall_reaction_dedup`, and the `WATCHDOG_STALL_REACTION_ENABLED` gate). This is a rewrite rather than an addition for a hard technical reason: **Telegram permits exactly one reaction per sender per message.** A surviving stall reaction and a new counter would both write the originating message's single slot and silently clobber each other, with the winner decided by drain order. There is one slot, so there must be one writer.

The stall condition itself does not disappear — it becomes a state the counter expresses rather than a separate signal competing for the same real estate. The planner should decide how a stall reads on the counter (an early jump to the ceiling, a distinct terminal glyph, or an immediate forced progress message) and say so explicitly.

Per this repo's first development principle (no legacy code tolerance: no parallel-run migrations, no half-migrations), the old path is deleted in the same change that lands the new one. Its tests go with it or get rewritten against the new behavior — `tests/unit/` currently asserts the broken payload's schema parity, which would otherwise pin the dead design in place.

**Pre-requisite defect:** `build_custom_emoji_index` in [`tools/emoji_embedding.py`](tools/emoji_embedding.py) is the intended source of custom-emoji ids. It has **no production caller** (grep finds it only in `tests/unit/test_custom_emoji_index.py`), and it cannot work as written: it reads `result.documents` from `GetEmojiStickersRequest`, which returns only set descriptors. Against this account it returns `sets=7, documents=0`, so the builder hits its "no labels extracted" branch and writes nothing. `data/custom_emoji_embeddings.json` has never existed. Its unit tests pass because they mock the client and feed it documents the real API never returns.

## Recon Summary

Recon was performed directly against the live Telegram API and production logs rather than by code-reading alone.

- Keycap digits are not standard reactions: server advertises exactly 74, no keycaps in either encoding (U+FE0F-suffixed or bare U+20E3).
- The account is Telegram Premium (`premium: True`), so `ReactionCustomEmoji` is available.
- Digit custom emoji exist and the pack is already installed: `Birthday Collection` (id `1901206392136531984`) carries all ten digits.
- A custom-emoji reaction verifiably sets, reads back as `ReactionCustomEmoji(document_id=...)`, and **replaces in place** — the counter's core motion. Tested in Saved Messages; test message deleted.
- The outbox relay already supports `custom_emoji_document_id`, so no new transport is needed.

**Revised:** 3 items — scope adjusted
- Original framing was "PM sends the counter." Changed to watchdog-owned after `bridge/liveness.py` established that self-reported liveness cannot detect the wedged case.
- The counter must not be selected via `find_best_emoji`: that function uses semantic similarity plus `_softmax_sample` at a temperature, i.e. deliberately random among top-K. Correct for "pick a feeling," wrong for "this is tick 4." Ticks need a deterministic lookup.
- Scope widened to a full rewrite of the #1313 stall-reaction feature. Initially the dead path was going to be left alone and superseded; the single-reaction-slot constraint makes coexistence impossible, so the old path is deleted in the same change.

- `build_custom_emoji_index` is uncalled and non-functional (detailed above). Either fix it to fetch per-set documents, or pin the digit ids as a static table and treat the index as out of scope.

- Keeping ⏳ as the glyph. It is not a legal Telegram reaction under any code path.
- Keeping `_apply_stall_reaction` as an independent writer beside the counter. Two writers, one slot — see the Solution Sketch.

## Acceptance Criteria

- [ ] No reaction this system can emit is absent from the server's advertised reaction set; a test asserts every emoji in the stall/heartbeat path is a legal reaction.
- [ ] `logs/bridge.log` shows zero `failed to set reaction` warnings originating from the watchdog path over a session that exercises the full counter.
- [ ] A session running longer than one tick interval displays an advancing counter on its originating message, visible in Telegram.
- [ ] The counter degrades to a standard-emoji arc (no exception, no missing reaction) when the custom emoji is rejected.
- [ ] Advancing past the final tick is impossible: the code path raises or otherwise refuses rather than clamping at the last value.
- [ ] Reaching the ceiling causes a substantive progress message to be published, and a fresh counter attaches to that new message.
- [ ] The reaction dedup/advance key reflects delivery outcome, not merely enqueue, so a failed reaction is retried.
- [ ] `build_custom_emoji_index` either has a production caller and produces a non-empty index on this account, or is removed along with its tests.
- [ ] `STALL_REACTION_EMOJI`, `_apply_stall_reaction`, `_clear_stall_reaction_dedup`, and the `watchdog:stall_reaction_applied:` key are gone from the codebase; `grep -rn "STALL_REACTION_EMOJI\|_apply_stall_reaction" --include="*.py"` returns no hits outside the git history.
- [ ] Exactly one code path writes a reaction to a session's originating message. No second writer can clobber the counter's slot.
- [ ] The stalled condition is still user-visible through the counter, and the issue's chosen representation for it is documented in the plan.
- [ ] Tests asserting the old payload's schema parity are deleted or rewritten against the new behavior, so no test pins the dead design in place.

## Downstream

Constraints for the planner:

- The tick signal must come from an observer independent of the session being measured. A session-emitted heartbeat is not acceptable — see the `bridge/liveness.py` rationale in the Problem section.
- Custom emoji `document_id`s belong to a third-party sticker pack that could be uninstalled or withdrawn. Every custom reaction must carry a standard-emoji fallback.
- Reactions land on a human's own message. Per the repo's reaction-voice convention, every glyph must read as a first-person statement from the reactor, never as a directive aimed at the author.
- Exploratory work already exists on branch `wip/session-heartbeat-ticker` (commit `78443c3f6`): a pinned digit table, a fallback arc, and a ceiling-enforcing helper in `bridge/response.py` with tests. It is temp progress, not a design commitment — the plan should adopt, relocate, or discard it on the merits.

valorengels:
Closes #2716

Rewrites the #1313 stall reaction as a watchdog-owned advancing tick counter with a forced-progress ceiling.

## Why

#1313 shipped `_apply_stall_reaction`, which queues ⏳ on the user's originating message when a session stalls. ⏳ is not a legal Telegram reaction — it is absent from the 74 the server advertises and is listed in this repo's own `INVALID_REACTIONS`. The repo contradicted itself in two files, and the feature has never once landed: `logs/bridge.log` shows 11 `Stall reaction queued` events and zero deliveries. The dedup key was set on enqueue rather than delivery, so a failure was never retried.

This is a rewrite rather than an addition because Telegram permits exactly one reaction per sender per message. A surviving stall reaction and a new counter would both write that single slot and clobber each other, with the winner decided by outbox drain order. One slot, one writer.

## What the counter means

Pure wall-clock duration. It asserts only that the watchdog — a process independent of the session — has eyes on this session. It makes **no** stall claim: a wedged session and a busy session tick identically. That is deliberate, and it is why no evidence-freshness gating is needed. `bridge/liveness.py` already established that a handler which has stopped firing cannot testify to its own failure, so the signal comes from an independent observer.

The ceiling is the point. Past `HEARTBEAT_MAX_TICKS` the counter refuses to advance and steers the session to publish a progress message, so the human is promised a substantive update rather than an indefinite spinner.

## Changes

- **`agent/reaction_priority.py`** (new) — seven-writer rank table. Unmapped glyphs derive to rank 1, the fail-safe direction: only the tick ever yields.
- **`bridge/response.py`** — `heartbeat_reaction(tick)` which **raises** past the ceiling rather than clamping, plus the interval/max constants and the fallback arc. The arc is deliberately **not** registered in `_reaction_constants()`: `_assert_distinct()` raises `ImportError` at module import, so registering an arc glyph would stop the bridge from starting.
- **`bridge/liveness_ticks.py`** (new) — shared sync Redis state for watchdog and relay. `heartbeat:anchor:` / `:tick:` / `:ceiling:` (6h TTL), `heartbeat:slot_owner:{chat}:{msg}` (24h). Fail-quiet throughout.
- **`monitoring/session_watchdog.py`** — the entire ⏳ path deleted; new sync `_publish_liveness_ticks()` called from `watchdog_loop()` in its own guarded try. Sweeps `("running", "active")` — the async `check_all_sessions()` queries `active` only and never sees the worker-executed Telegram sessions this feature exists for.
- **`bridge/telegram_relay.py`** — drain-point precedence guard, slot-owner recording, and counter re-anchoring including the `DELIVERED_NO_ID` case. All Redis via `asyncio.to_thread`.
- **`agent/output_handler.py`** — keyword-only `priority` on `_build_reaction_payload`; all five hand-mirrored payload sites updated.
- **`tools/emoji_embedding.py`** — `build_custom_emoji_index` and its machinery deleted. It had no production caller and could not work as written: it read `result.documents` from `GetEmojiStickersRequest`, which returns only set descriptors, so `data/custom_emoji_embeddings.json` has never existed. Its tests passed only because they fed a mocked client documents the real API never returns.

## Deliberate deviation from the plan

The plan's Task 1 says the drain should "drop a lower-priority reaction when a higher one owns the slot." Applied literally that breaks the ⚠ → ✍ worker-recovery progression, where a later, *lower*-ranked reaction is supposed to win — and the plan's own writer table documents that overwrite as intended. The narrower rule the Race 1 section actually specifies is implemented instead: terminal-is-final for every writer, plus tick-yields-to-everything. That still closes the latent bug where `tool_budget`'s 🤯 could land after a terminal reaction. Pinned by `test_pickup_still_overwrites_worker_down`.

## Pre-existing failure repaired

`test_assert_distinct_covers_full_constant_set` fails on `origin/main`, independently verified: its expected set omits `REACTION_WORKER_DOWN`, which `_reaction_constants()` has carried since #1312. It is on the plan's UPDATE list, so it is repaired here rather than left red.

## Known gap

`PREMIUM_DIGIT_REACTIONS` ships **empty**, so the counter currently renders the fallback arc rather than digits. The per-digit `document_id`s exist nowhere recoverable — not in the repo, not in git history, and the `wip/session-heartbeat-ticker` branch that held them is unreachable. Reading them requires a live authenticated Telethon client against the Premium account, which cannot be done safely while the bridge is running (a second client on the same auth key risks `AUTH_KEY_DUPLICATED`, which would invalidate the session and take down chat I/O).

This is the documented Risk 1 degradation path, so nothing is broken and the arc still advances. Pinning the table later is a one-dict edit at `bridge/response.py`.

## Verification

```
python -m ruff check .                → All checks passed
python -m ruff format --check .       → 1358 files already formatted
grep -rn "STALL_REACTION_EMOJI\|_apply_stall_reaction" --include="*.py" .  → 0 hits
```

419 passed across the affected unit and integration suites (DB 12, `scripts/pytest-clean.sh`), independently re-verified on the four highest-value suites (161 passed).

Not covered by automation, and needing a real chat: "advancing counter visible in Telegram" and "zero `failed to set reaction` warnings in `logs/bridge.log`".


valorengels:
## Patch response — both blockers closed

All six findings addressed in `8aea9170f`. 406 tests green, ruff clean.

### Blocker 1 — forked terminal-status vocabulary — FIXED

You were right on both counts, and the bug is real. The relay hand-rolled `{completed, done, error, failed, abandoned, cancelled, canceled}`. Three of those (`done`, `error`, `canceled`) are statuses this system never produces, and the one that mattered — `killed` — was missing. A session killed while a tick was already queued had the tick land on top of its terminal reaction, because the two outbox queues drain in undefined order and the slot-owner layer only saves you when the terminal reaction happened to go first.

Replaced with the canonical `models/session_lifecycle.TERMINAL_STATUSES`, so the second vocabulary is gone rather than corrected. Added `test_killed_session_drops_the_tick`, which asserts both that `killed` is in the canonical set and that the guard acts on it.

### Blocker 2 — unranked glyph locked the slot — FIXED

Confirmed exactly as described. The root cause is that a single `priority` value was being asked two different questions, and the fail-safe answer differs between them: for **may this be dropped**, an unmapped glyph should rank high (never drop what you don't recognize); for **may this claim the slot**, it should rank not at all.

Split them with a `priority_ranked` field on the payload. An unranked glyph still delivers, but `_record_reaction_slot_owner` skips it, so it takes no ownership. That makes the behavior match what `react_with_emoji`'s own comment and the feature doc already claimed — it wins while it runs and the next tick overwrites it.

Two tests pin it: `test_unranked_reaction_delivers_without_claiming_the_slot` (delivered, no ownership recorded, and a subsequent tick still passes) and `test_ranked_reaction_still_claims_the_slot` as the control, so the fix can't regress into "nothing ever claims the slot."

### Tech debt 1 — deviation rationale was wrong — FIXED

Good catch, and worth stating plainly: the narrowing is correct but I had justified it with an example that cannot occur. ⚠ is set in-process by `react_if_worker_down` and never reaches the drain, so it records no slot owner and `test_pickup_still_overwrites_worker_down` was passing because the slot was unowned, not because the guard spared it.

The docstring now cites the two pairs that genuinely would break under a monotonic-rank rule: the rank-3 RTR suppress after a rank-2 🤯, and the rank-4 child-completion suppress after a rank-3 ✍. The ⚠ case is noted as the non-example it is, so the next reader doesn't reintroduce it.

### Tech debt 2 — empty digit table — follow-up filed

#2837, with the reproduction steps, the pack id from the original recon, and the carried-over acceptance criteria. Your point stands: an unowned "one-dict edit" is how a feature stays half-shipped.

### Tech debt 3 — interval floor — FIXED

`HEARTBEAT_TICK_INTERVAL_SECONDS` now clamps with `max(1, ...)`. The docstring explicitly invites lowering it to exercise the ceiling, so a `0` typo was reachable, and it would have raised `ZeroDivisionError` once per running session on every 300s scan.

### Tech debt 4 — post-deploy steer burst — DOCUMENTED

Added to Known limitations. The first sweep seeds missing anchors from `started_at`, so every session already past the ceiling window claims it at once. Self-correcting and each marker fires once, but the first post-deploy hour will look chatty and that should read as expected rather than as a bug.

### One thing your review surfaced indirectly

Adding a key to `_build_reaction_payload` made `test_react_with_emoji_mirror` fail — the parity helper passes `priority` through explicitly, which forces `priority_ranked` True and left it unable to express a legitimately unranked writer. Fixed in the helper rather than by weakening the mirror: `priority_ranked` is per-writer by design, so presence and type are the parity claim while every other field is still compared by value. The helper would still catch a mirror that dropped the key entirely.

valorengels:
## Review: APPROVED

Head `54987cb5f5dd338dad309f001f8b251420215be5`. Blockers: 0. Tech debt: 0.

Scoped re-review of the patch commits after the initial CHANGES REQUESTED (2 blockers, 4 tech-debt). The areas confirmed clean in the first pass were not disturbed and were not re-reviewed: hot-loop `asyncio.to_thread` discipline, the ceiling raising rather than clamping, arc non-registration, deletion completeness for both the ⏳ path and the custom-emoji index, sync watchdog integration, Redis TTL discipline, and the session-scoped steering leg.

### Blocker 1 — forked terminal-status vocabulary — CLOSED

`_TERMINAL_SESSION_STATUSES` is gone repo-wide. `bridge/telegram_relay.py:224` imports the canonical `models/session_lifecycle.TERMINAL_STATUSES` inside the existing lazy block alongside `AgentSession`, which is the right scoping for a module on the relay's poll path. `killed` is now honored and the three statuses this system never produces (`done`, `error`, `canceled`) are gone.

This was a real bug, not just a style violation: a session killed while a tick was already queued had the tick land on top of its terminal reaction, because the two outbox queues drain in undefined order and the slot-owner layer only helps when the terminal reaction happens to drain first. `test_killed_session_drops_the_tick` pins both halves — that `killed` is in the canonical set, and that the guard acts on it.

### Blocker 2 — unranked glyph claimed the slot at terminal rank — CLOSED

The root cause was one `priority` value answering two questions whose fail-safe directions are opposite: for *may this be dropped*, an unmapped glyph should rank high; for *may this claim the slot*, it should not rank at all. `priority_ranked` separates them. An unranked payload still delivers (`_reaction_yields_slot` continues to derive from `priority` alone) but skips `_record_reaction_slot_owner`, so a single agent-issued `react_with_emoji` call no longer pins the message for the owner key's 24h TTL and suppresses the budget, pickup, and tick writers behind it.

All five hand-mirrored writers updated; no sixth writer exists. The `message.get("priority_ranked", True)` default was scrutinized for the deploy window where in-flight payloads predate the field: `True` risks one stale unranked payload claiming the slot, `False` risks a stale *terminal* reaction failing to claim it (terminal glyphs are themselves unmapped). Both are one-time and bounded by outbox backlog across a single restart, and `True` preserves terminal finality — the invariant the whole guard exists to protect. Correct choice.

### Tech debt 1 — deviation rationale — CLOSED

The narrowing was right but justified with an example that cannot occur: ⚠ is set in-process by `react_if_worker_down` and never reaches the drain, so `test_pickup_still_overwrites_worker_down` was passing because the slot was unowned, not because the guard spared it. The docstring now cites the two pairs that genuinely would break under a monotonic-rank rule (rank-3 RTR suppress after rank-2 🤯; rank-4 completion suppress after rank-3 ✍) and flags the ⚠ case as the non-example it is.

### Tech debt 2 — empty digit table — TRACKED

#2837, with recovery steps and carried-over acceptance criteria. Not expected in this PR.

### Tech debt 3 — interval floor — CLOSED
### Tech debt 4 — post-deploy steer burst — CLOSED (documented in Known limitations)

### Parity-helper accommodation — hardened beyond what was asked

Adding a key to `_build_reaction_payload` broke `test_react_with_emoji_mirror`, and the first fix compared `priority_ranked` by presence and type only. That left precisely the hole this blocker was filed against: flipping `react_with_emoji`'s literal to `True` would have reinstated the bug with every test still green, since the two slot-ownership tests construct payloads by hand rather than driving the writer.

Replaced with a caller-stated `expect_ranked` (`False` for `react_with_emoji`, `True` for the other four), which the builder's own output cannot express because passing `priority` through forces the derived value `True`. Verified by injecting the exact drift: `test_react_with_emoji_mirror` fails red, then passes on restore. The five hand-mirrored payload sites are now held together by value, not just by shape.

### Verification

406 passed across the affected unit and integration suites (`scripts/pytest-clean.sh`, DB 12). `ruff check` and `ruff format --check` clean. Deletion greps return zero hits.

Not covered by automation and needing a real chat, tracked in #2837: "advancing counter visible in Telegram" and "zero `failed to set reaction` warnings in `logs/bridge.log`".

Issue #2703, opened 2026-08-10
Reliability risk: SDLC router — a reaped sdlc-local anchor turns its own live issue lock into a permanent ISSUE_LOCKED


valorengels:
## Investigation Finding

**Component:** SDLC router (`next-skill`) / issue-lock lifecycle
**Detected:** 2026-08-10 (observed live while driving issue #2696 through the pipeline)
**Status:** Observed-once, mechanism confirmed by code read

## Symptoms

Mid-pipeline on issue #2696, `sdlc-tool next-skill --issue-number 2696` began hard-returning
`ISSUE_LOCKED` naming **our own run** as the blocking owner:

```
{"blocked": true, "reason": "ISSUE_LOCKED", "guard_id": "ISSUE_LOCK",
 "owner_run_id": "2de131d7e4b041ffbce04746c321761e",
 "owner_session_id": "sdlc-local-2696", "orphaned_lock": false}
```

The lock owner was the live run asking the question. Every subsequent `next-skill` call blocked
identically — the pipeline could not route another stage.

Two independent defects compose into a permanent self-deadlock:

1. **A live local run gets reaped for not having a heartbeat it never emits.** `sdlc-local-{N}`
   anchors are `is_ledger` rows driven by subagents, not by the worker, so `last_heartbeat_at` is
   `None` and `turn_count` stays 0. The anchor's lease freshness is therefore only refreshed by
   ledger writes at stage boundaries. A stage that legitimately takes longer than
   `ISSUE_LOCK_RENEWAL_FRESHNESS_SECONDS` (20 min) ages the lease past the freshness window, and
   `session-ensure --kill-orphans` finalizes the anchor to `status="killed"` while the pipeline is
   actively working. Here the last ledger write was the 04:34Z `/do-plan` dispatch and the anchor was
   killed at 04:54:21Z — about 2 minutes after the round-3 plan revision landed at 04:52:09Z.

2. **Once reaped, the router cannot recognize its own lock.** `tools/sdlc_next_skill.py:472-503`
   resolves the current run's identity with `find_session_by_issue(issue_number)`, which defaults
   `include_terminal=False` (`tools/_sdlc_utils.py:200`, `_TERMINAL_ISSUE_LOOKUP_STATUSES =
   {"failed", "completed", "killed"}`). With the anchor terminal, `peek_run_id` is `None`, so the
   `touch_issue_lock(..., peek_run_id, peek=True)` identity comparison mismatches the still-live lock
[...]

Issue #2620, opened 2026-08-07
orphaned_lock is unreliable for every locally-minted lease (pid belongs to the ephemeral session-ensure CLI)


valorengels:
Surfaced during #2615's review. PR #2615 removed the heartbeat's dependence on `orphaned_lock`, but its other consumers still trust it: `tools/sdlc_next_skill.py:502`, `tools/sdlc_session_ensure.py:480`, `tools/_sdlc_utils.py:611,693`.

Root cause (now documented in `docs/features/sdlc-issue-ownership-lock.md`): the payload pid is stamped by the short-lived `sdlc-tool session-ensure` CLI at acquire and is dead within seconds, and renewal never re-stamps it. So `_lock_owner_is_live` reads every locally-minted lease as orphaned.

Consequence per `.claude/skills/sdlc/SKILL.md:202`: `orphaned_lock: true` tells the router "the owning run died, wait for the TTL" while `false` is the only unconditional stop — so a LIVE local foreign owner reads as a ghost and a rival waits instead of stopping. Related: #2305 (pre-existing; not worsened by #2615).

Fix direction: either make those consumers do ownership/heartbeat-freshness checks instead of pid inference, or stamp an identity that is genuinely long-lived for the run (the supervising `claude -p` pid or the heartbeat's own, with the lifetime trade-offs #2615's review discussed). Verify against live Redis the way #2615's review did.

PR #2859, opened 2026-08-18
fix(#2648): make the durable renewer's pid authoritative for issue-lock liveness

### LINKED ISSUE #2648: Reliability risk: SDLC issue-lock — a single renewal stamp makes a dead run read live for up to 1200s, short-circuiting the pid check
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/issue-lock-renewer-pid-authoritative.md

## Recon Summary

Recorded at PLAN time (2026-08-17) against `main` @ `335fde5b3`. Consolidates the
two premise-check comments below plus a fresh re-verification, so the ISSUE→PLAN
gate has evidence in the body rather than only in comments.

**Scope after the coordinator's accepted re-scope (comment 3): the surviving
half only — `_lock_owner_is_live`'s renewal-freshness short-circuit. Everything
else on the original Next Steps list is mooted by #2784 and #2703.**

**Confirmed** (re-verified at `335fde5b3`):

- The short-circuit is live at `models/session_lifecycle.py:1033`
  (`if _lock_renewal_is_fresh(payload) is True: return True`), inside
  `_lock_owner_is_live` (`:995`). `_lock_renewal_is_fresh` is `:962`.
  Constants: `ISSUE_LOCK_TTL_SECONDS = 1800` (`:847`),
  `ISSUE_LOCK_RENEWAL_FRESHNESS_SECONDS = 2 * (1800 // 3) = 1200` (`:859`).
- Reproduced with no Redis, by constructing payloads and spying on
  `agent.session_health._psutil_process_for_pid` (dead pid on every row):

  | `renewed_at` age | `is_live` | pid check reached |
  |---|---|---|
  | 5 s | `True` | no |
  | 1199 s | `True` | no |
  | 1201 s | `False` | yes |
  | absent | `False` | yes |

  So a process that writes one renewal stamp and dies reads LIVE for up to
  1200 s, and the pid check that would immediately have returned `False` is
  never invoked.
- `touch_issue_lock`'s renewal path (`_healed_renewal_payload`, `:1084`) carries
  `pid` and `create_time` over verbatim and re-stamps only `renewed_at`.
  `tests/unit/test_session_lifecycle.py:1331-1333` asserts that survival.
- The two DURABLE renewers are `tools/sdlc_lease_heartbeat.py:429`
  (detached, `start_new_session=True`, `renew_only=True`) and
  `agent/session_executor.py:246` (worker tier-1 60 s tick). Both have long-lived,
  checkable pids that never reach the payload. The heartbeat seeds
  `since_renew = interval` so **tick one always renews**, i.e. a running heartbeat
  stamps within seconds of the mint.
- The SHORT-LIVED CLI renewers that must NOT stamp a renewer identity:
  `tools/_sdlc_utils.py:609`, `tools/_sdlc_utils.py:834`,
  `tools/sdlc_stage_marker.py:884`, `tools/sdlc_next_skill.py:638`,
  `tools/sdlc_session_ensure.py:539`. Each exits within seconds and would
  re-poison the field between durable ticks.

**Revised** (what the premise re-check changed):

- The incident's *outcome* is largely closed by #2784 (merged): the heartbeat now
  watches its supervisor's `(pid, create_time)` and actively RELEASES the lease
  after `SDLC_SUPERVISOR_DEATH_CONFIRMATIONS` consecutive death observations.
  What survives is the case where the releaser itself is gone.
- The originally-proposed "make freshness corroborating" fix is NO LONGER SAFE
  as stated. `tools/sdlc_session_ensure.py::_iter_orphan_sessions` (`:1059`) now
  gates `--kill-orphans` on this same predicate: a live owner exempts the session
  regardless of idle time. An `sdlc-local-{N}` anchor has
  `last_heartbeat_at = None`, so renewal freshness is the ONLY thing marking it
  alive. Weakening the short-circuit unconditionally re-opens #2703 defect 1 —
  a live-but-quiet local run reaped mid-pipeline.

**Dropped** (mooted — explicitly NOT in scope):

- Confirm-by-reproduction of the original incident (done, above).
- Whether a `/do-sdlc` fork should acquire under its own run id (#2026/#2076
  fork-self-lock surface).
- Whether `release_issue_lock` should be reachable from a supervisor that proved
  the owner dead — #2784 already does exactly this, from the heartbeat.
- Whether the worker's in-process renewal has the same exposure — it does, and
  it is covered by the same single fix.

**Pre-requisites:**

None. All edits land in files that exist on `main`; no schema change, no
migration (`session:issuelock:{N}` is a plain non-Popoto string key).

---

## Investigation Finding

**Component:** SDLC issue-lock (`models/session_lifecycle.py`), `/do-sdlc` supervisor
**Detected:** 2026-08-07 (live incident during the #2634 SDLC lane)
**Status:** Observed-once, with full forensics captured before the lock was released

## Symptoms

A local `/do-sdlc` run on #2634 stopped at the Step 3d.6 continuity re-ensure with the one unconditional refusal:

```json
{"blocked": true, "reason": "ISSUE_LOCKED",
 "owner_run_id": "a57ba594e836462abd0bf35cec057380",
 "owner_session_id": "sdlc-local-2634", "orphaned_lock": false}
```

`orphaned_lock: false` means the lock-liveness check affirmatively concluded the owning run was **alive**. It was not. The owner was the run's *own forked PLAN stage*, whose process was already dead — a fork self-lock, which #2076 (fixing #2026) made a regression by definition.

The supervisor behaved correctly given the signal it was handed: it refused to force past a lock reported as live, and it did not kill the ledger anchor. The defect is in the signal.

**The framing that this is "TTL alone, not liveness" is wrong** and worth correcting for whoever picks this up: `_lock_owner_is_live` *does* consult liveness. The bug is that one liveness signal short-circuits past the other.

## Diagnostic Output

Lock payload read directly from `session:issuelock:2634` (plain non-Popoto string key, read-only):

```
type   string
ttl    1596
value  {"run_id": "a57ba594e836462abd0bf35cec057380",
        "session_id": "sdlc-local-2634",
        "pid": 63385,
        "machine_id": "541DBF1F-3751-523A-859D-41735590B143",
        "hostname": "MacBookPro.local",
        "create_time": 1786084483.66491,
        "renewed_at": 1786084491.929686,
        "target_repo": "tomcounsell/ai"}
```

Owner liveness at inspection time:

```
$ ps -p 63385 -o pid,ppid,etime,command
(no output — process does not exist)

$ ps -eo pid,command | grep a57ba594e836462abd0bf35cec057380 | grep -v grep | wc -l
0                       # no lease-heartbeat process for this run

$ ls .worktrees/ | grep job-model-scaling
(no output — no worktree was ever created)
```

Derived timings:

```
renewal happened 8.3 s after acquire        (create 1786084483.66 -> renewed 1786084491.93)
renewed_at age at inspection:  210 s
ISSUE_LOCK_TTL_SECONDS:               1800
ISSUE_LOCK_RENEWAL_FRESHNESS_SECONDS: 1200   # = 2 * (1800 // 3)
210 <= 1200  ->  _lock_renewal_is_fresh() returns True
```

The relevant control flow, `models/session_lifecycle.py::_lock_owner_is_live`:

```python
if _lock_renewal_is_fresh(payload) is True:
    return True          # <-- short-circuit; every pid check below is skipped

pid = payload.get("pid")
...
proc = _psutil_process_for_pid(pid)
if proc is None:
    return False         # <-- would have correctly reported orphaned
```

So: the lease was acquired, renewed exactly **once** 8.3 seconds later, and its owner then died. For the following ~1200 seconds the lease reported a live owner on the strength of that single stamp, written by a process that no longer existed. The pid check that would have returned `False` immediately was never reached. After the freshness window it would still have taken the remaining TTL to free the key outright.

## Steps Attempted

Diagnosis only, then a sanctioned release — no forcing, no raw `DEL`, no killing of another lane's state:

1. Enumerated every `tools.sdlc_lease_heartbeat` process on the machine and confirmed the only #2634 heartbeat carried *my* run id, not the lock owner's.
2. Confirmed lock-owner pid 63385 dead and no worktree present, so no in-flight work was at risk.
3. Released via the sanctioned compare-and-delete `models/session_lifecycle.py::release_issue_lock(2634, "a57ba594...")`, which deletes only if the payload still carries that exact run id and therefore cannot clobber a successor's freshly acquired lock. Returned `True`; key confirmed gone.
4. Re-ran `/do-sdlc`, which proceeded normally.

Cost: one wasted supervision cycle on the lane, plus the operator time to distinguish a genuine rival from a ghost.

## Impact

A run whose forked stage acquires the issue lock and then dies **wedges its own parent** for up to 1200 s of false-live, then up to the full 1800 s TTL. During that window every rival — including the legitimate owner — is told a live run holds the issue, which is exactly the refusal case the supervisor is designed never to override. The failure is silent and reads as healthy contention.

Two aggravating properties:

- **It misdirects the operator.** `orphaned_lock: false` is a positive assertion of life. Correctly-behaving supervisors escalate to a human, and the human's first instinct is to hunt for a rival lane that does not exist. This is the same class of mistake as the liveness mirage that #2305 defect 1 fixed for AgentSession-derived inference — a hollow signal that reads alive forever — reintroduced one layer up.
- **Neither liveness signal is authoritative on a locally-minted lease.** The payload's `pid` is stamped by the short-lived `sdlc-tool session-ensure` CLI, which is dead within seconds — this is precisely #2620's stated reason for not trusting pid alone. But `renewed_at` is a *stamp that outlives its writer*, so it is not proof of life either. #2620 made freshness **conclusive** (a short-circuit) when the evidence only supports it being **corroborating**.

## Next Steps

- [ ] Confirm the mechanism by reproduction: acquire a lease, renew once, `SIGKILL` the owner, then assert `peek` reports `orphaned_lock: false` for ~1200 s despite a dead pid.
- [ ] Inspect `models/session_lifecycle.py::_lock_owner_is_live` (the `_lock_renewal_is_fresh(payload) is True -> return True` short-circuit) and `_lock_renewal_is_fresh` — decide whether a fresh renewal should remain sufficient on its own.
- [ ] Decide which pid the payload should carry. The durable process for a local run is `tools/sdlc_lease_heartbeat.py`, not the `session-ensure` CLI. If `touch_issue_lock`'s same-owner renewal branch (`models/session_lifecycle.py:1289-1295`, which already re-stamps `renewed_at`) also re-stamped `pid` + `create_time` with the *renewer's* identity, pid liveness would become both meaningful and authoritative, and the 1200 s blind window would collapse to the renewal interval.
- [ ] Evaluate making freshness corroborating rather than conclusive: require a fresh `renewed_at` **and** a live renewer pid before returning live, keeping the existing fail-toward-live posture for cross-machine and indeterminate payloads.
- [ ] Check whether the forked-stage acquisition is itself avoidable — whether a `/do-sdlc` fork should acquire the lock under its own fresh run id at all, or inherit the parent's (`--reuse-run-id` exists on `sdlc_session_ensure` for this purpose). This is the #2026/#2076 fork-self-lock surface.
- [ ] Confirm whether the worker's in-process 60 s renewal path (`agent/session_executor.py::_tick_issue_lock_renewal`) has the same exposure, or whether a live worker makes the stamp trustworthy there.
- [ ] Consider whether `release_issue_lock` should be reachable from a supervisor that has positively proven the owner dead, so this self-heals without operator forensics.

valorengels:
Closes #2648

Plan: `docs/plans/issue-lock-renewer-pid-authoritative.md`

## Problem

`models/session_lifecycle.py::_lock_owner_is_live` opened with an unconditional
short-circuit: a fresh `renewed_at` returned `True` and every pid check below was
skipped. A process that took the SDLC issue lease, wrote one renewal stamp, then
died read as a LIVE owner for up to `ISSUE_LOCK_RENEWAL_FRESHNESS_SECONDS`
(1200 s). Rivals were told `orphaned_lock: false` — a positive assertion of life,
the one refusal a correct supervisor never overrides.

Deleting the short-circuit is not the fix: it reverts #2620 (the payload `pid` is
the ephemeral `sdlc-tool session-ensure` CLI, dead within seconds, so pid
inference alone reads every locally-minted lease as orphaned) and re-opens #2703
defect 1 (`--kill-orphans` consumes the same predicate). The payload had to gain a
checkable **durable** pid before the predicate could be tightened at all.

## Approach

Fix the recorded evidence first, then tighten the predicate only where that
evidence exists.

- `_healed_renewal_payload` stamps `renewer_pid` + `renewer_create_time` when an
  allowlisted durable renewer opts in, and **strips** the group on every other
  renewal (sticky clear), so no ephemeral CLI renewal can re-carry a stale pid.
- Membership comes from an **explicit `renewer_module=` token** passed by the two
  durable call sites, never from frame introspection: the detached heartbeat runs
  as `python -m tools.sdlc_lease_heartbeat`, so `__name__` resolves to
  `"__main__"` and an introspection check would silently never stamp in
  production while passing every import-based test.
- Freshness becomes **corroborating** where a complete same-machine renewer
  identity is present and stays **conclusive** otherwise. Same-machine is decided
  by the existing `_payload_from_same_machine`; there is deliberately no separate
  renewer machine field.
- Both non-releasing heartbeat deadline exits retract the identity through a
  bounded CAS clear, so a heartbeat that retires while its run lives falls back to
  the untouched #2620 behavior rather than turning the live run dead.
- `drop_renewer_identity` is a pure key removal: it preserves the existing
  `renewed_at` and the remaining TTL (via `pttl`, with the `-1`/`-2` sentinels
  guarded). A dying heartbeat must not buy a fresh liveness window.
- Every early return in the new branch fails toward LIVE; only the full
  conjunction returns `False`. That matters most at the reflection layer, where
  `reflections/utilities.py::_lock_says_live` feeding `sdlc_upvote_lanes`
  `create_session` makes a false-dead spawn an unattended rival lane.
- `ISSUE_LOCK_RENEWER_GRACE_SECONDS = 180`, env-overridable, TTL-independent, and
  catalogued alongside `ISSUE_LOCK_RENEWAL_FRESHNESS_SECONDS`.

## Demonstrated red — the four-run PAIR

The plan requires both halves, four runs, all outputs pasted. Half (a) alone would
not prove the #2620/#2703 fallback survived the tightening.

Pre-fix side is `2fee63629` (tests added, source untouched); post-fix side is
branch HEAD.

### (a) The tightening works — RED before

```
$ scripts/pytest-clean.sh "tests/unit/test_session_lifecycle.py::TestRenewerIdentityLiveness::test_dead_renewer_past_grace_is_not_live" -q
F                                                                        [100%]
=================================== FAILURES ===================================
_____ TestRenewerIdentityLiveness.test_dead_renewer_past_grace_is_not_live _____
tests/unit/test_session_lifecycle.py:1407: in test_dead_renewer_past_grace_is_not_live
    assert _lock_owner_is_live(payload) is False
E   AssertionError: assert True is False
E    +  where True = <function _lock_owner_is_live at 0x11059f270>({'create_time': 1000.0, 'hostname': 'this-host', 'machine_id': 'hw-uuid-1', 'pid': 424242, ...})
=========================== short test summary info ============================
FAILED tests/unit/test_session_lifecycle.py::TestRenewerIdentityLiveness::test_dead_renewer_past_grace_is_not_live
1 failed in 7.50s
```

### (a) GREEN after

```
$ scripts/pytest-clean.sh "tests/unit/test_session_lifecycle.py::TestRenewerIdentityLiveness::test_dead_renewer_past_grace_is_not_live" -q
.                                                                        [100%]
1 passed in 7.88s
```

### (b) The #2703/#2620 fallback survives — GREEN before

```
$ scripts/pytest-clean.sh "tests/unit/test_session_lifecycle.py::TestRenewerIdentityLiveness::test_no_renewer_identity_keeps_the_2620_short_circuit" -q
.                                                                        [100%]
1 passed in 32.68s
```

### (b) GREEN after

```
$ scripts/pytest-clean.sh "tests/unit/test_session_lifecycle.py::TestRenewerIdentityLiveness::test_no_renewer_identity_keeps_the_2620_short_circuit" -q
.                                                                        [100%]
1 passed in 9.20s
```

### Suite-level form of the same obligation (SC4)

Full `tests/unit/test_session_lifecycle.py` at the pre-fix baseline: **10 failed,
95 passed**. Every one of the 10 is in the three new classes
(`TestRenewerIdentityLiveness`, `TestRenewerIdentityStamping`,
`TestDropRenewerIdentity`); all 95 pre-existing tests pass unmodified on the
pre-fix side, and the whole file passes post-fix.

## Test results

```
$ scripts/pytest-clean.sh tests/unit/test_session_lifecycle.py tests/unit/test_sdlc_session_ensure.py tests/unit/test_sdlc_lease_heartbeat.py tests/unit/reflections/test_utilities_lock_says_live.py tests/unit/reflections/test_sdlc_upvote_lanes.py -q
312 passed, 27 warnings in 71.95s (0:01:11)
```

`ruff check` and `ruff format --check` over `models/ tools/ agent/ tests/
reflections/`: clean.

All 25 Verification-table rows executed verbatim and pass, including the four
expect-no-output rows (no renewer machine field anywhere; no frame introspection
in the renewal path; no ephemeral CLI renewer stamps; no stale xfails).

## Consumer-layer proofs

The predicate has three consumers, and the two that are not the peek path get
their own tests rather than an inference:

- **Reaper (SC7)** — at the `_iter_orphan_sessions` layer: an `sdlc-local-{N}`
  anchor with `last_heartbeat_at=None` and a LIVE `renewer_pid` is not yielded;
  the same row with a dead renewer past the grace is.
- **Reflection layer (SC9)** — new
  `tests/unit/reflections/test_utilities_lock_says_live.py` drives the real
  predicate through `_lock_says_live`. The existing `test_sdlc_upvote_lanes.py`
  stubs that function wholesale and proves nothing about the payload-to-verdict
  path.

## Accepted residual concerns

Disclosed in the plan after six critique rounds and deliberately **not** closed
here: a heartbeat that crashes while its run lives, and a worker SIGKILL orphan.
Both are recorded in the plan's Risks section.


valorengels:
Closing on owner decision. The dead-renewer detection here is real work, but this PR is 22 days stale and CONFLICTING at 1399 lines, and the standing direction is that small fixes that ship beat large branches that keep missing the merge window.

On the adjacent question: this PR's `renewer_pid`/`renewer_create_time` substrate is what a split-brain check would need (see the ownership gap documented on #3258 — two supervisors sharing a `run_id` both pass this PR's own "self-owned continue path", which treats a matching `owner_run_id` as yours and hard-blocks only a FOREIGN one). The owner decision is that split-brain does not warrant the build right now: the second supervisor detected the collision and stood down on its own, and agent-level discernment is accepted as the mitigation. #3258 stays open as documentation of the gap, not as queued work.

Reopen if the dead-renewer case starts biting on its own merits. Refs #2648, #3258.

PR #2861, opened 2026-08-18
Scope Job.touch/mark_at_rest/revive saves to their mutated fields (#2860)

### LINKED ISSUE #2860: Job.revive / Job.touch / Job.mark_at_rest full-model save() clobbers unrelated fields
Job.touch(), Job.mark_at_rest(), and Job.revive() in models/job.py each mutate one or two fields and then call a bare self.save(), which serializes the **entire** model hash.

## Current behavior

- touch() sets last_active_at, then save() — writes every field.
- mark_at_rest() sets status, then save() — writes every field.
- revive() sets status + last_active_at, then save() — writes every field.

Because these write the whole Redis hash from an in-memory copy loaded at some earlier point, a concurrent writer that changed a different field (notably goal, which carries the expectations list) between this instance load and its save loses that write: last-write-wins over the whole hash rather than over the field actually being changed.

This is a real collision surface. touch() fires on every message bind and PM turn, and revive() fires on every steering message, so both race against expectation mutations on goal.

## Why this is fixable narrowly

The repo already has the idiom and calls it "the structural clobber-proof idiom": renormalize_last_active_scores uses fresh.save(update_fields=["last_active_at"]) and backfill_open_expectations_index uses save(update_fields=["has_open_expectations"]). popoto 1.8.0 Model.save supports update_fields for a partial save, and the Job.save override already gates its UTC-reattach on whether last_active_at is in scope — the scoped-save case is anticipated by design.

## Proposed change

Scope each save to the fields the method actually mutates:

- touch()        -> save(update_fields=["last_active_at"])
- mark_at_rest() -> save(update_fields=["status"])
- revive()       -> save(update_fields=["status", "last_active_at"])

## Acceptance criteria

- All three methods use field-scoped saves naming exactly the fields they mutate.
- A concurrent modification to goal is preserved across a touch/revive/mark_at_rest.
- mark_at_rest does not write or re-score last_active_at (rest-by-age must not refresh recency).
- touch and revive still update the last_active_at SortedField score correctly via the existing tz-reattach gate.
- Job and expectations suites pass.

valorengels:
Closes #2860

Plan: `docs/plans/job-field-scoped-lifecycle-saves.md`

## What

`Job.touch()`, `Job.mark_at_rest()`, and `Job.revive()` each mutate one or two fields and then called a bare `self.save()`, which serializes the **entire** model hash. A concurrent writer that changed a different field (notably `goal`, which carries the expectations list) between this instance's load and its save lost that write: last-write-wins over the whole hash rather than over the field actually mutated.

Each save is now scoped to exactly the fields the method mutates:

| Method | `update_fields` |
|---|---|
| `touch()` | `["last_active_at"]` |
| `mark_at_rest()` | `["status"]` |
| `revive()` | `["status", "last_active_at"]` |

No new API is invented and no call sites change. popoto 1.8.0's `Model.save` already supports `update_fields`, the module already uses it in `renormalize_last_active_scores` and `backfill_open_expectations_index` (calling it "the structural clobber-proof idiom"), and the `Job.save` override already gates its UTC-reattach on field scope.

## Notable semantics

`mark_at_rest()` does not name `last_active_at`, so the tz-reattach is correctly skipped and recency is left alone — resting a Job by age must never refresh its recency. The previous full save re-wrote that field on every rest transition.

The real production races are `revive()` (`bridge/job_router.py:256`, every bound-existing-job routing decision) and `mark_at_rest()` (`sweep_to_rest`) against expectation writers. `touch()` has no current production caller and is fixed for consistency.

## Tests

New `TestFieldScopedLifecycleSaves` in `tests/unit/test_job_model.py`: `goal` survives each of the three transitions under a concurrent second-instance writer; `mark_at_rest()` does not refresh recency; `touch()`/`revive()` still score the SortedField correctly; and raw `$IndexF:Job:status:*` set membership is asserted at the Redis layer, not just via ORM round-trip.

- `tests/unit/test_job_model.py` + `tests/unit/test_job_router.py`: 82 passed
- Positive controls `tests/integration/test_job_routing.py` + `tests/unit/test_job_router.py`, unmodified: 19 passed


valorengels:
## REVIEW — APPROVED (0 blockers)

Reviewed at head `bdda3f63f7d00e90ec6aa510747bd5758131bb37`.

### Code (`2fd183508`)
- **`update_fields` correctness**: each list names exactly the fields its method mutates, verified against the method bodies. `touch` → `[last_active_at]`, `mark_at_rest` → `[status]`, `revive` → `[status, last_active_at]`. No missed field, so no mutation silently fails to persist.
- **Red-state verified**: the fix was temporarily reverted and the new tests re-run. Exactly the three clobber-proof tests failed (`test_touch_preserves_a_concurrent_goal_write`, `test_revive_...`, `test_mark_at_rest_...`) while the recency/scoring tests stayed green. The tests detect the precise regression they claim to, rather than trivially passing.
- **Concurrent-writer setup is genuine**: tests hydrate a real second instance via `Job.query.get(...)`, mutate `goal` through `add_expectation` (which goes through `_write_goal_data` → bare `save()`), then call the lifecycle method on the original stale instance.
- **Index assertion derives its key via the field API** (`get_special_use_field_db_key` → `DB_key`), matching popoto's own `filter_query`, not a hand-built f-string.
- **popoto partial-save semantics confirmed against installed source**: `on_save` fires only for fields named in `update_fields`, so `IndexedField`/`SortedField` maintenance is correctly scoped. `Job` has no `auto_now` field for a scoped save to skip.
- Reads via `POPOTO_REDIS_DB` are `sismember` only; test cleanup deletes through the ORM. No raw Redis writes.

### Docs (`bdda3f63f`)
Docs-only — no code, test, or config change. The `durability-model.md` text is accurate against the implementation, including the `mark_at_rest` recency omission and the loss of incidental index self-heal. Describes the new status quo without migration narration.

### Gates
`ruff check` and `ruff format --check` clean. `tests/unit/test_job_model.py` + `tests/unit/test_job_router.py`: 82 passed. Positive controls `tests/integration/test_job_routing.py` + `tests/unit/test_job_router.py`, unmodified: 19 passed. No `Co-Authored-By` or generation-footer trailer on either commit.

valorengels:
## Review: APPROVED — 0 blockers, 0 tech-debt findings

Reviewed at head `bdda3f63f7d00e90ec6aa510747bd5758131bb37`. Full findings in the preceding comment; summary:

- `update_fields` lists are exactly correct per method (no missed field, so no silent non-persisting mutation).
- Red-state verified: reverting the fix fails precisely the three clobber-proof tests and nothing else.
- Tests use a genuine second hydrated instance, so they prove clobber-proofing rather than call shape.
- Index assertion derives its Redis key via the field API, not a hand-built string.
- popoto partial-save index semantics confirmed against installed source; `Job` has no `auto_now` field to skip.
- Docs commit is docs-only and accurate; no forbidden commit trailers on either commit.
- ruff clean; 82 passed on the changed suites; 19 passed on unmodified positive controls.

Issue #2870, opened 2026-08-19
validate_build.py carries two more private plan-document grammars: parse_file_assertions and parse_success_criteria_commands


tomcounsell:
## What

`scripts/validate_build.py` reads a plan document with two functions that each carry their own private grammar, independent of `agent/verification_parser.py` and independent of each other:

- `parse_file_assertions` (`scripts/validate_build.py:43-77`)
- `parse_success_criteria_commands` (`scripts/validate_build.py:149-180`)

These are a real third and fourth divergence living inside the very file #2836/#2843 converged onto `agent/verification_parser.py`. Named here rather than left unnamed, per the carry-forward failure #2570 diagnosed: partial convergence on a shared artifact leaves the next divergence to be discovered the same way a build gate failing a plan that is fine.

## Measured failure shapes

**`parse_file_assertions`**: captures a backtick-quoted path after `Create`/`Add`/`Delete`/`Update` and asserts it relative to the repo root (`check_file_assertions` calls `Path(assertion["path"]).exists()` from the repo root). A bare filename in a checkbox (e.g. `` Create `foo.md` `` rather than `` Create `docs/plans/foo.md` ``) resolves `False` forever, turning a checkbox with the right intent into a permanently-false `exists` assertion. `docs/plans/verification-runner-convergence.md`'s own T2 tripped exactly this during critique round 2.

**`parse_success_criteria_commands`**: only admits commands starting with `python`, `pytest`, `grep`, `test `, `ls `, `cat `, `ruff` (`scripts/validate_build.py:166-177`), and for each matching checkbox line takes only the **first** backtick-quoted span (`re.search(r"` + "`" + r"([^` + "`" + r"]+)` + "`" + r"", line)`). A criterion naming two commands is silently reduced to one — the second command is never run.

## Why this belongs to its own issue

#2778 (referenced from the #2836/#2843 PR's carry-forward comment) is about a different reader — the unregistered `.claude/hooks/validators/validate_verification_section.py` hook. This issue is specifically about the two functions above, which live inside `scripts/validate_build.py` itself.

## Scope

Converging these two functions onto a shared, tested grammar (or documenting/loosening their false-assertion failure modes) is deferred to whoever picks up this issue. Out of scope for #2836/#2843, which explicitly named these two functions in `## No-Gos` rather than silently leaving them unconverged.

Refs #2836, Refs #2843, Refs #2778

tomcounsell:
[...]

Issue #2791, opened 2026-08-13
verification_parser runner false-FAILs every 'prints N' row: compares expectation to exit code, not stdout


tomcounsell:
Found during PR #2695's round-3 review (2026-08-13): `agent.verification_parser`'s runner reports FAIL on every Verification-table row whose expectation is `prints \`N\`` because it compares the expected value against the command's exit code instead of its stdout. The reviewer re-ran all six such rows in docs/plans/context-recall-advisory-flag.md by hand; every one printed its expected value (all 16 rows genuinely pass) while the runner scored them failed.

Impact: any plan using the common `prints \`0\`` / `prints \`N\`` idiom gets systematic false failures from the automated runner, training agents to distrust or hand-re-run the Verification table.

Fix shape: compare the expectation against captured stdout (trimmed) for prints-style rows; keep exit-code comparison for 'exit code N' rows. A regression test should cover one row of each style.

tomcounsell:
**Genuine and reproducible, but the mechanism in the title is misattributed — correcting before anyone plans this, or the fix lands on the wrong file.**

There are **two independent expectation evaluators** in this repo, and they share only the row splitter:

- `agent/verification_parser.py:184` — `evaluate_expectation(expected, *, exit_code, output)`, driven by `run_checks()` at `:267`. This is the runner named in the title.
- `scripts/validate_build.py:261` — `check_verification_table(checks)`, with its own inline if/elif chain at `:291-303`. It imports only `split_row_cells` from the other module (`scripts/validate_build.py:25`), and a comment at `:295` claims it "matches agent/verification_parser.py semantics" — it no longer does.

**What actually happens.** `` prints `N` `` matches none of the branches in `evaluate_expectation`, so it falls through to the bare `return False` at `agent/verification_parser.py:264`. Reproduced against main:

```
'prints `0`'       stdout=0/exit=1 -> False | stdout=0/exit=0 -> False
'prints `4`'       -> False in both cases
'match count == 0' -> True   (handled correctly)
```

So the observed symptom is real: every `` prints `N` `` row FAILs unconditionally, including the six at `docs/plans/context-recall-advisory-flag.md:1010-1015`. But the cause is an **unrecognized expectation form**, not a stdout/exit-code mix-up. `evaluate_expectation` ignores both in that branch.

[...]

Issue #2778, opened 2026-08-13
validate_verification_section.py is not registered in manifest.toml — the hook never runs


valorengels:
## What

`.claude/hooks/validators/validate_verification_section.py` has no `[[hook]]` entry in `.claude/hooks/manifest.toml`, and no entry in either generated `settings.json` hooks block. Nothing invokes it. The validator is fully implemented and, as of PR #2746, fully tested (31 tests), but it does not run in the harness.

## Evidence

```
grep -n verification .claude/hooks/manifest.toml   # no output
git show main:.claude/hooks/manifest.toml | grep -n verification   # no output
```

Its three sibling section validators (`validate_documentation_section`, `validate_test_impact_section`, `validate_no_gos_justification`) are all registered as `event = "PostToolUse"`, `matcher = "Write"`, `exit_policy = "propagate"`.

## Not introduced by #2746

Pre-existing on `main`. PR #2746 ported the validator onto the shared `hook_utils/hook_target.py` target-resolution contract along with the other four; that work is correct and unaffected. This issue is only about registration. It surfaced during #2746's demonstrated-red verification, where the validator refuses its deficient payload correctly (exit 2) but would never receive one in production.

## Open question before fixing

Registering it will start enforcing `## Verification` tables on every plan-doc `Write`, under `exit_policy = "propagate"` (a hard block). Existing plan docs in `docs/plans/` should be audited for compliance first, or the first agent to touch a non-compliant plan gets blocked on a section it did not author. Decide deliberately: register and fix the fallout, or delete the validator as abandoned.

Do not hand-edit `settings.json` — registration generates from `manifest.toml`. See `docs/features/hook-manifest.md`.

Refs #2738

tomcounsell:
## Carry-forward from #2836 / #2843 (verification-runner convergence)

If/when this hook is registered, it must **import** the canonical table definitions from `agent/verification_parser.py` rather than reimplement them — reimplementing gives the repo a write-time gate that can reject a `## Verification` section both build-time runners (`scripts/validate_build.py`, and `docs/sdlc/do-build.md`'s inline runner) accept.

The canonical definitions, as of this comment:

- **A table** is a contiguous run of pipe-prefixed (`|`) lines — `agent.verification_parser._iter_pipe_blocks`, per GFM's table-is-a-leaf-block rule (a blank line, or a line that cannot be part of the table, ends it).
[...]

Issue #2817, opened 2026-08-14
next-skill has no terminal verdict: a MERGE-completed pipeline with no recoverable pr_number blocks with NO_RULE


valorengels:
**Wave 2 plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/wave2-router-verdict-integrity.md

**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/next-skill-terminal-verdict-and-pr-resolution.md

*Planned as one lane with #2817, #2824, #2825 — shared code surface, see the plan's Problem section.*

## Symptom

`sdlc-tool next-skill --issue-number N` on a **terminal** pipeline (`MERGE == completed`) whose ledger no longer carries a `pr_number` returns:

```
{"blocked": true, "reason": "no matching dispatch rule", "guard_id": "NO_RULE"}
```

Measured 2026-08-14 by direct call to `agent.sdlc_router.decide_next_dispatch` with `{ISSUE..DOCS, PATCH, MERGE} = completed` and `meta` without `pr_number`.

## Why this is filed

This is the **residual** state after #2757 is fixed. #2757's bug is that this same shape returns a G8 re-dispatch of `/do-build` on shipped work; its fix stops the false rebuild, and what is left is `Blocked(NO_RULE)`. That is honest (there is nothing to dispatch) but the reason string reads like a routing bug rather than "this pipeline is finished".

Every terminal row is gated on `pr_number`: row 10 (`_rule_ready_to_merge`) returns False without it, and `guard_g6_terminal_merge_ready` returns None without it. So a terminal pipeline whose `pr_number` was lost has no owner in the dispatch table at all.

## Expected

A terminal pipeline should produce an explicit terminal verdict — "pipeline complete, nothing to dispatch" — keyed on `MERGE == completed`, rather than falling off the end of the dispatch table into `NO_RULE`.

## Scope note

Deliberately **not** in #2757's scope: #2757's fence is `tools/sdlc_next_skill.py` (the stage-artifact verifier) and tests. A terminal verdict is a change to `agent/sdlc_router.py`'s guard/dispatch surface and deserves its own design pass — including whether the verdict is a new guard, a new dispatch row, or a short-circuit in `tools/sdlc_next_skill.py::decide()` above the router call.

Refs #2757

## Recon Summary

[...]

PR #2871, opened 2026-08-19
Converge verification-table parsing and expectation evaluation

### LINKED ISSUE #2783: validate_build.py expectation parser inverts anti-criterion rows: false PASS when violated, false FAIL when clean
> **Context.** `scripts/validate_build.py` is a deterministic (no-LLM) build validator. It reads a plan document from `docs/plans/{slug}.md`, extracts the rows of that plan's `## Verification` markdown table, runs each row's shell command, and compares the output against the row's stated expectation. Agents in the SDLC pipeline use its PASS/FAIL summary to decide whether a build is clean or needs to route to `/do-patch`. An **anti-criterion** row is a Verification row that asserts something forbidden is *absent* — for example "the diff introduced no swallowed exceptions" — and is conventionally written as a `grep -c` with the expectation `match count == 0`.

## Problem

The expectation parser in `run_verification_checks` mis-evaluates two of the expectation forms plans actually use. The consequences run in **both** directions, and the more dangerous one is silent.

### Severity 1 (silent): a violated anti-criterion reports PASS

`grep -c` exits **0** when it selects at least one line. The parser's fallback branch treats a zero exit code as success on its own:

```python
# scripts/validate_build.py:301-303
else:
    # Flexible match: check if expected is in output or exit code is 0
    passed = expected.lower() in actual_output.lower() or actual_exit == 0
```

So for a row expecting `match count == 0` where the forbidden pattern **is** present — `grep -c` prints `24` and exits `0` — the second disjunct is true and the row is reported **PASS**. The anti-criterion passes precisely when it is violated.

Reproduced directly:

```
$ python3 -c "
expected='match count == 0'; actual_output='24'; actual_exit=0
print('PASS' if (expected.lower() in actual_output.lower() or actual_exit == 0) else 'FAIL')"
PASS   <- should be FAIL
```

This is the worst possible failure mode for a validator: the rows that exist specifically to catch a regression are the rows that go quiet when the regression happens.

### Severity 2 (noisy): correct builds report FAIL

Running the validator against `docs/plans/docs-auditor-migration-context-and-bare-paths.md` (from a worktree, using the venv interpreter) produced `9 PASS, 4 FAIL, 2 SKIP`. **All four FAILs are false** — each expectation is in fact satisfied, verified by re-running every command by hand:

| Reported | Actual | Truth |
|---|---|---|
| `Detector still detects -- expected: output > 0, got exit=0 output=1` | `1` | `1 > 0` is true |
| `Sibling-pattern reason recorded -- expected: output > 0, got exit=0 output=8` | `8` | `8 > 0` is true |
| `No sweeper-region edits (anti-criterion) -- expected: match count == 0, got exit=1 output=0` | `0` | `0 == 0` is true |
| `No swallowed exceptions introduced (anti-criterion) -- expected: match count == 0, got exit=1 output=0` | `0` | `0 == 0` is true |

Two distinct causes produce these.

**Cause 1 — relational expectations are compared as literal strings.**

```python
# scripts/validate_build.py:298-300
elif expected.startswith("output "):
    expected_output = expected.replace("output ", "")
    passed = actual_output == expected_output
```

For the expectation `output > 0` this evaluates `actual_output == "> 0"` — a string comparison of the produced number against the literal text `> 0`. It can never pass. Every relational form (`output > 0`, `output >= 1`, `output < 5`) false-FAILs regardless of what the command prints. Branch ordering matters here: `output contains X` is handled earlier at `:294` and is correct; only the bare `output ` branch at `:298` is broken.

**Cause 2 — `match count == N` falls through to the exit-code fallback.**

The `match count == N` form matches none of the three explicit branches, so it lands in the `else` at `:301-303`. For a correct anti-criterion — `grep -c` prints `0` and exits `1`, because [grep exits 1 when no lines are selected](https://www.gnu.org/software/grep/manual/grep.html#Exit-Status) — neither disjunct holds: `"match count == 0"` is not a substring of `"0"`, and the exit code is `1`. False FAIL. This is documented `grep` behavior, not a malformed command.

Note that Cause 2 is the same line as Severity 1: the fallback is simultaneously too strict (0 matches → FAIL) and too lax (N matches → PASS). It is inverted.

**Current behavior:** anti-criterion and threshold rows report FAIL on clean builds and PASS on violated ones.

**Desired outcome:** relational and count-style expectations are evaluated numerically, so a row's verdict reflects the property it states.

## Definitions

| Term | Meaning |
|---|---|
| Verification row | One row of a plan's `## Verification` markdown table: a name, a shell command, and an expected result. Parsed by `agent/verification_parser.py` and executed by `scripts/validate_build.py`. |
| Anti-criterion | A Verification row asserting a forbidden thing is absent, conventionally `grep -c ...` with expectation `match count == 0`. Used to pin No-Gos and safety invariants. |
| `/do-patch` | The SDLC stage that repairs a failing build. A false FAIL routes a clean build here needlessly; a false PASS lets a broken one skip it. |

## Impact

A validator whose safety rows are wrong in both directions is worse than no validator. The false FAILs are merely noisy — but they train agents and humans to treat this tool's output as unreliable, and that learned dismissal is what makes the false PASS dangerous: nobody is reading closely by the time it matters.

In the lane that surfaced this (#2744 / #2759), the four FAILs were correctly dismissed only because every command was re-run by hand. An agent trusting the summary would have routed a clean, fully-passing build to `/do-patch`.

## Solution sketch

Parse each expectation into an `(operator, operand)` pair and evaluate numerically when the operand parses as a number, supporting `==`, `!=`, `>`, `>=`, `<`, `<=` for both the `output <op> N` and `match count <op> N` forms, coercing `actual_output` to `int` where it parses.

For count-style expectations the **exit code must be ignored entirely** — `grep -c`'s exit status reports whether it matched, not whether the row's assertion holds, and conflating the two is the root of both severities. Leave `exit code N` and `output contains X` untouched; they are correct today.

The `else` fallback at `:301-303` should probably not survive in its current form at all. A row whose expectation the parser cannot interpret is a plan-authoring error, and the file already has a good precedent for that case: the `malformed` handling at `:265-278` fails the run while saying plainly that the plan is what needs fixing. Silently guessing from an exit code is what produced this bug.

Open question for the planner: whether any existing plan's Verification rows are currently green *because* of the false PASS. A sweep of `docs/plans/*.md` for `match count ==` rows, re-evaluated under corrected semantics, would size that blast radius before the fix lands.

## Related observation (do not conflate)

The same run reported `SKIP: withheld-rate-non-regression -- timed out after 30s`. The per-check timeout is hardcoded at `scripts/validate_build.py:285` with the SKIP emitted at `:319`. Thirty seconds is too tight for a legitimate row that provisions a `git worktree`, and the row degrades to SKIP rather than failing loudly. A plan-declared per-row timeout override, or a longer default, would stop real coverage disappearing quietly. This is adjacent to the parser defect, not part of it — it may warrant its own issue.

## Acceptance criteria

- [ ] A `match count == 0` row whose `grep -c` finds matches reports **FAIL** (closes the false-PASS hole).
- [ ] A `match count == 0` row whose `grep -c` finds nothing reports **PASS** despite `grep` exiting 1.
- [ ] `output > 0` passes when the command prints a positive integer and fails when it prints `0`.
- [ ] `==`, `!=`, `>`, `>=`, `<`, `<=` are supported for both `output <op> N` and `match count <op> N`.
- [ ] `exit code N` and `output contains X` behavior is unchanged, with tests pinning that.
- [ ] An uninterpretable expectation fails as a plan-authoring error rather than being guessed from the exit code.
- [ ] Running the validator against `docs/plans/docs-auditor-migration-context-and-bare-paths.md` reports no false FAILs among the four rows listed above.

## Downstream context

Filed from the #2744 / #2759 docs-auditor lane, whose file fence (`reflections/docs_auditor.py` plus its tests and docs) excludes `scripts/validate_build.py`. No fix is proposed or attempted there; this defect is unowned.

### LINKED ISSUE #2836: parse_verification_table merges any second table in ## Verification into the executable checks, unconditionally failing the gate
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/verification-runner-convergence.md

> The SDLC pipeline's build and review gates execute the `## Verification` section of a plan document as a table of runnable checks. The parser that extracts those checks, `parse_verification_table` (`agent/verification_parser.py:115`), has no concept of table boundaries: it collects **every** line starting with `|` anywhere in the section. Any *second* markdown table an author places in `## Verification` — a red/green summary, a findings recap, a prose comparison — is silently merged into the executable check list.

## Problem

**Current behavior.** `parse_verification_table` gathers all pipe-prefixed lines in `## Verification` into one row list, skips only the first two (the first table's header and separator), and turns every later row with a matching column count and non-empty cells into a `VerificationCheck`. A second table contributes its header row, its `---` separator, and its data rows as "checks". Their Expected cells (prose like `Pre-change`, bare numbers like `12`) are unrecognized expectation grammar, which `evaluate_expectation` (`:184`) evaluates to `False` — so every such junk row is a guaranteed FAIL, and the `/do-build` / `/do-pr-review` verification gate (`docs/sdlc/do-build.md:185`) fails unconditionally no matter how correct the real checks are.

**Observed live** (2026-08-17, issue #2741): the plan `docs/plans/docs-auditor-rename-detection.md` carried a "Red/green state" summary table inside `## Verification`. Nine junk rows (`12, 4, 8, 7, 7, 9`, `Pre-change`, `---`) parsed as executable checks. It was caught during the plan's round-6 critique and only defused by commit `61717ccb2` converting the summary table to bullets — a formatting workaround, not a fix. Every future plan author will hit the same trap, and the failure mode (gate fails with baffling rows the author never wrote as checks) points away from the real cause.

**Desired outcome.** A non-check table in `## Verification` either (a) is cleanly ignored, or (b) produces a named, actionable parse error identifying the offending rows — never silent junk checks that poison the gate.

## Definitions

| Term | Definition |
|---|---|
| Verification table | The markdown table in a plan's `## Verification` section whose rows are `Check \| Command \| Expected` and are executed by the build/review gates. Parser: `agent/verification_parser.py::parse_verification_table`. |
| `evaluate_expectation` | `agent/verification_parser.py:184`; interprets an Expected cell (e.g. `match count == 0`, `exit code 0`) against a command's stdout/exit code. Unrecognized grammar returns `False`. |
| Verification gate | The step in `/do-build` and `/do-pr-review` (`docs/sdlc/do-build.md:185`) that runs every parsed check and blocks on failures. |

## Solution sketch

Scope the parser to the one designated check table instead of the whole section. Open questions for the planner (root-cause-level choice, deliberately not pre-decided):

- **Header-signature scoping**: parse only a table whose header row matches the check schema (first three column names ≈ Check/Command/Expected, case-insensitive); any other table in the section is skipped whole, optionally with an info-level note.
- **First-table-only scoping**: stop collecting rows at the first non-`|` line after the table began — simpler, but breaks legitimate multi-table check layouts if any exist (none known).
- Either way: a table whose header does NOT match should surface as a named lint/malformed entry listing the skipped rows, so an author who *meant* those rows as checks finds out at parse time, not gate time.

Regression case: a `## Verification` section containing a valid check table followed by a prose summary table must parse to exactly the check table's rows with zero malformed entries (or explicitly-skipped entries), and the #2741 pre-`61717ccb2` plan text makes a ready-made fixture.

## Downstream context

This issue will be consumed by `/do-plan` to produce a plan at `docs/plans/{slug}.md`, then executed by `/do-build`.

## Acceptance criteria

- [ ] A non-check markdown table in `## Verification` no longer produces executable checks or guaranteed-fail junk rows.
- [ ] Authors get a named diagnostic (not silence) when a table is skipped.
- [ ] The #2741 pre-fix plan text parses to only its 16 intended checks.
- [ ] Demonstrated-red: the new test fails against the current parser before the fix.

## Recon Summary

Verified against `main` @ `f491306c5` (2026-08-18). No commit has touched
`agent/verification_parser.py`, `scripts/validate_build.py`, or their tests since this
issue was filed.

**Confirmed**
- `agent/verification_parser.py:139` collects every `|`-prefixed line in the section into
  one flat `rows` list; `:149` skips only `rows[0:2]`. No table-boundary concept exists.
- Reproduced on the #2741 pre-fix text (`git show 61717ccb2^:docs/plans/docs-auditor-rename-detection.md`):
  `parse_verification_table` returns **27 checks / 0 malformed**. The post-fix text returns
  **16 checks**. The 11 extra rows are the summary table's header, separator, and 9 data rows,
  with Expected cells like `Meaning`, `---`, `demonstrated red — must reach 0` — all of which
  `evaluate_expectation` returns `False` for. Guaranteed-FAIL junk, exactly as described.
- `scripts/validate_build.py:106` already does header-signature detection (`"command" in
  stripped.lower()`) and already terminates the table on the first non-`|` line.

**Revised**
- The issue says "nine junk rows". The measured count is **11** (9 data rows + header + separator).
  The 16-check figure in the acceptance criteria is correct.
- The issue's "first-table-only scoping — breaks legitimate multi-table check layouts if any
  exist (**none known**)" is **wrong**. Two exist:
  `docs/plans/completed/sdlc-lease-heartbeat-supervisor-lifetime.md` and
  `docs/plans/completed/sdlc-continuity-reensure-rebind.md` each carry a second
  `| Anti-criterion | Command | Expected |` table holding real executable rows. First-table-only
  scoping silently drops 6 real checks from each — a gate that cannot fire, which is a strictly
  worse defect than the one being fixed. **Header-signature scoping applied to every
  pipe-block in the section** is the only safe option.

**Blast radius (measured, not estimated)**
Simulated per-block header-signature scoping across all **500** plan documents carrying a
`## Verification` section:
- **0 active plans** (`docs/plans/*.md`) change their parse result.
- 8 completed plans change; every delta is junk-row removal or `.worktrees`-fenced shell
  lines that currently parse as checks.
- 2 completed plans (`delivery_guard_resume_epoch_scoping.md`,
  `redis-replication-sentinel-failover.md`) head their only table `| # | Criterion | Check |`
  with no Command column. Under header-signature scoping they yield 0 checks — so
  "section has pipe-blocks but no check table" must be a **loud** malformed error, not a
  silent skip, or the fix trades one silent gate for another.

**Pre-requisites** — none.

**Dropped** — none.

### LINKED ISSUE #2843: scripts/validate_build.py mis-evaluates 'match count == N' and 'output > N' Verification rows
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/verification-runner-convergence.md

## What I observed

During `/do-build` for #2741 (PR #2842), the two Verification-table runners named in `docs/sdlc/do-build.md` disagreed completely on the same plan table:

- `agent/verification_parser.py` (`parse_verification_table` + `run_checks`) → **19 PASS, 0 malformed**
- `scripts/validate_build.py` → **8 PASS, 10 FAIL, 1 SKIP**, exit 1

All 10 FAILs are false. Examples from the run:

```
FAIL: Basename cache reset survives -- expected: output > 0, got exit=0 output=1
FAIL: Existence-invariant coverage survives the channel collapse -- expected: output > 3, got exit=0 output=5
FAIL: Reporter docstring un-blinded -- expected: match count == 0, got exit=1 output=0
```

`1 > 0`, `5 > 3`, and a `match count == 0` row whose output is `0` are all obviously passes.

## Cause

`scripts/validate_build.py:291-303` only understands three expectation forms — `exit code N`, `output contains X`, and `output X` (exact string equality) — then falls through to a "flexible match" of `expected.lower() in output or exit == 0`.

Two consequences:

1. **`output > N` rows** hit the `startswith("output ")` branch and compare `actual_output == "> 0"`, which is never true.
2. **`match count == 0` rows** hit the flexible branch. `grep -c` exits **1** when the count is zero, so `exit == 0` is false and the literal string `match count == 0` is not in the output `0` → FAIL. Rows ending in `| wc -l` pass only incidentally, because `wc` exits 0.

`agent/verification_parser.py` handles both forms correctly and reads stdout rather than exit codes. The plan for #2741 explicitly documents the `grep -c` exit-code hazard (Risk 4, rule 3) and was validated end-to-end through `parse_verification_table()` for exactly this reason.

## Why it matters

`docs/sdlc/do-build.md` step 14 routes a non-zero `validate_build.py` exit to `/do-patch` for up to 3 iterations. A plan whose Verification table uses `match count == N` or `output > N` — both idiomatic in this repo — sends a passing build into a patch loop chasing failures that do not exist, and `/do-patch` has nothing real to fix.

## Suggested direction

Have `scripts/validate_build.py` delegate expectation evaluation to `agent/verification_parser.py` rather than carrying a second, weaker parallel implementation. Two evaluators for one table format is the divergence itself.

## Not verified

- Whether any past build was actually misrouted to `/do-patch` by this.
- Whether `scripts/evaluate_build.py` shares the same evaluator.

Found while building #2741; not blocking that PR (the canonical runner was used for the gate).

## Recon Summary

Verified against `main` @ `f491306c5` (2026-08-18). No commit has touched
`scripts/validate_build.py`, `agent/verification_parser.py`, or their tests since this issue
was filed.

**Confirmed**
- `scripts/validate_build.py:291-303` is exactly as described: `exit code N` →
  `output contains X` → `output <exact>` → flexible fallback
  `expected.lower() in actual_output.lower() or actual_exit == 0`.
- `output > N` reaches the `startswith("output ")` branch and string-compares stdout to
  `"> 0"`. Never true.
- `match count == 0` reaches the flexible branch. A clean `grep -c` prints `0` and exits **1**,
  so both disjuncts are false → FAIL on a satisfied anti-criterion.
- `agent/verification_parser.py:184-264` handles both forms correctly, reading stdout.

**Answers to the issue's "Not verified" items**
- **`scripts/evaluate_build.py` does NOT share the evaluator.** It is an LLM criteria judge:
  `evaluate_criteria(criteria_text, diff_text)` at `:95` reads `## Success Criteria` plus
  `git diff` and has no expectation grammar at all. It is unaffected by this work.
- There is nonetheless a **third** reader of the table:
  `.claude/hooks/validators/validate_verification_section.py:133` re-implements
  "every `|`-prefixed line" row collection and `:148` splits cells on **every** `|`, ignoring
  the `\|` escape that `split_row_cells` honors. It is not registered in
  `.claude/hooks/manifest.toml` (#2778), so it never actually runs — which is why its
  divergence has never been observed. Out of scope here; tracked by #2778.

**Related-issue finding**
- **#2783 is the same lines.** Its Severity-1 case (a violated `match count == 0` reporting
  PASS because `grep -c` exits 0) and its Severity-2 case are both produced by the
  `:301-303` flexible fallback. Deleting that fallback in favour of
  `agent.verification_parser.evaluate_expectation` makes both directions structurally
  impossible.
- **#2791 is NOT fixed by this.** `prints N` is absent from `evaluate_expectation`'s grammar
  and stays absent. After convergence, `prints \`0\`` rows go from incidentally-PASS under
  `validate_build.py` (flexible fallback, exit 0) to FAIL — matching the canonical runner's
  existing FAIL. Louder, not worse; #2791 remains the fix.

**Measured migration cost**
68 Expected cells across 11 active plans use grammar neither evaluator recognizes
(`prints \`0\``, `output == N`, `> 0`, `ok`, `exit 0`, …). All 68 **already FAIL** the canonical
runner at `docs/sdlc/do-build.md:185` today, so convergence surfaces no new blocker beyond
the one the authoritative gate already imposes. The bare `output <exact-string>` form that
only `validate_build.py` supports is used in **zero** active plans, so delegation is lossless
in practice.

**Pre-requisites** — none. Shares a lane with #2836; the two touch the same two files.

**Dropped** — converging `run_checks`' 120s timeout with `validate_build.py`'s 30s/SKIP
behavior. Different product contracts, no observed divergence, deliberately out of scope.

tomcounsell:
## Summary

`agent/verification_parser.py` and `scripts/validate_build.py` disagreed on
what a `## Verification` table is and what an expectation means. This PR
converges them onto one shared definition of each, in `agent/verification_parser.py`.

- **Table scoping (#2836).** The parser now scopes to GFM pipe-blocks
  (contiguous runs of `|`-prefixed lines) instead of collecting every `|`-line
  in the section into one flat list. Every pipe-block is classified
  independently: a **check table** (>=3 columns, one of the first three
  headers is exactly `Command`) contributes checks; a **non-check table**
  (a summary, a findings recap) becomes a non-failing `SkippedTable`, named
  in both runners' reports; a section with table rows but **no** check table
  fails loudly with exactly one `MalformedRow` per pipe-block.
- **Evaluator delegation (#2843, #2783).** `scripts/validate_build.py` deleted
  its own local `parse_verification_table` and its four-branch expectation
  evaluator (the `expected.startswith(...)` chain with the `actual_exit == 0`
  fallback) and now imports `parse_verification_table` /
  `evaluate_expectation` from `agent.verification_parser`. It keeps only what
  is genuinely its own: a 30s-timeout, SKIP-on-timeout execution loop.

### Red-state evidence (T1, T3, T9)

Captured against `main` before any code changed:

```
OLD PARSER on the #2741 fixture (docs/plans/docs-auditor-rename-detection.md pre-fix): checks=27 malformed=0
  -> 16 real checks + 11 junk rows from a second, non-check summary table

scripts/validate_build.py (pre-fix, its own evaluator):
  output > 0 / echo 1         -> FAIL   (should PASS -- #2843)
  match count == 0 / clean    -> FAIL   (should PASS -- #2843)
  match count == 0 / violated -> PASS   (should FAIL -- #2783 Severity-1)
```

10 test failures (`TypeError`/`AttributeError`) across
`tests/unit/test_verification_parser.py` and `tests/unit/test_validate_build.py`
before the fix, all fixed after.

### Corpus re-measurement (T14)

Re-run against the current `main` tip at build time (`find docs/plans -name
'*.md'`, `.worktrees/` excluded by construction):

- 590 total files under `docs/plans/`, 505 carry a `## Verification` section,
  28 are active plans directly under `docs/plans/*.md`.
- **0 of the 28 active plans change their parse result.** 8 completed plans
  change; most deltas are junk-row removal or a new non-failing
  `SkippedTable`. One exception, named and accepted in the plan's Risks §2:
  `docs/plans/completed/opus-skill-prompts-4-7.md` loses a real check -- a
  data row orphaned by a blank line is no longer read as part of the check
  table, so it becomes a non-failing `SkippedTable` instead of a (mis)counted
  check. It is a completed plan, and the alternative -- failing on skipped
  tables -- would re-create #2836.
- **Universal (Risks §3) holds by construction, not by scan**: both runners
  now call the literal same `evaluate_expectation` function, so any row that
  fails under `validate_build.py` was already failing the canonical runner --
  there is no longer a second implementation that could disagree.
- **Universal (Risks §4) holds by construction**: the old, divergent
  command-cell extraction in `validate_build.py` is deleted; only the
  canonical `cells[1].strip("\`")` extraction remains.

### #2791 stays open

`prints N` is absent from `evaluate_expectation`'s grammar and stays absent
(explicitly out of scope -- see `## No-Gos`). After this delegation,
`` prints `0` `` rows go from incidentally-PASS under `validate_build.py`'s
old flexible fallback to FAIL -- which is what the canonical runner at
`docs/sdlc/do-build.md:185` already reports for them today. Louder, not
worse; #2791 remains the fix.

### Also in this PR

- `.claude/skills-global/do-plan/PLAN_TEMPLATE.md` -- documents the
  `Command`-column signature and non-check-table behavior for new plans.
- `docs/features/machine-readable-dod.md` -- documents per-block scoping,
  `SkippedTable`, and the corrected `format_results(results, table)` signature.
- `docs/sdlc/do-build.md` / `docs/sdlc/do-pr-review.md` -- the inline
  verification-table runner one-liners now pass the `ParsedTable`.
- Filed #2870 tracking `scripts/validate_build.py`'s two remaining private
  plan-document grammars (`parse_file_assertions`,
  `parse_success_criteria_commands`), and posted the canonical-definitions
  carry-forward on #2778 so a future hook registration imports rather than
  reimplements table scoping.
- #2783 was closed in error by an earlier plan-revision commit quoting this
  plan's prose; it was reopened during critique and this PR's `Closes`
  keyword below restores correct tracking.

## Test plan

- [x] `scripts/pytest-clean.sh tests/unit/test_verification_parser.py tests/unit/test_validate_build.py -q` -- 101 passed
- [x] `python -m ruff check agent/verification_parser.py scripts/validate_build.py` -- clean
- [x] `python -m ruff format --check agent/verification_parser.py scripts/validate_build.py` -- clean
- [x] This plan's own `## Verification` table (34 rows) passes end to end through the changed code
- [x] `scripts/validate_build.py docs/plans/verification-runner-convergence.md` exits 0 (39 PASS, 0 FAIL, 2 SKIP -- the two SKIPs are the accepted 120s-vs-30s timeout-axis divergence named in Rabbit Holes, not a code failure)

Closes #2836
Closes #2843
Closes #2783


tomcounsell:
## Review: Blocked on Conflict

This PR cannot be reviewed until it merges cleanly against its base branch.

- **mergeable:** `CONFLICTING`
- **mergeStateStatus:** `DIRTY`
- **head SHA judged:** `09fa34c36e4772d32d0b3e785b959e8323d9376a`
- **merge base:** `11848a857b5727d6ce2b322679d96331c807e043`

### Conflicting file

`git merge-tree --write-tree origin/main refs/pull/2871/head` reports exactly one conflict:

```
Auto-merging tests/README.md
CONFLICT (content): Merge conflict in tests/README.md
```

### Required action

Rebase onto the current base (or merge base into your branch), resolve the
`tests/README.md` conflict, and push. Re-run `/do-pr-review` once the branch is
mergeable.

> No code review was performed. The mechanical mergeability gate runs before
> any diff reading — if the branch cannot merge, no amount of code review
> can make the PR mergeable.

tomcounsell:

No blockers. The convergence is correct, the tests are real, and every number in the PR body that I could reproduce, reproduced. Two tech-debt items and two nits stand between this and approval.

The core of the change is right. Per-block scoping is the correct reading of GFM, the `Command`-in-first-three-columns signature is strict enough to be predictable, and the "rows but no check table" branch fails loudly instead of yielding a vacuous green. Deleting `validate_build.py`'s private parser and four-branch evaluator in favour of `evaluate_expectation` removes the second implementation that could disagree, which is what makes #2783 structurally impossible rather than merely fixed.

### What I reproduced myself

Per the review contract, every number I relied on was re-measured in this environment rather than credited from the PR body.

- `scripts/pytest-clean.sh tests/unit/test_verification_parser.py tests/unit/test_validate_build.py -q` → **101 passed** (matches the claim).
- `ruff check` and `ruff format --check` on both changed modules and both changed test files → **clean**.
- This plan's own `## Verification` table through the changed code path → **34/34 PASS, 0 malformed, 0 skipped**.
- `python scripts/validate_build.py docs/plans/verification-runner-convergence.md` → **39 PASS, 0 FAIL, 2 SKIP** (matches; the 2 SKIPs are the disclosed 120s-vs-30s timeout-axis divergence).
- Test counts behind the `tests/README.md` edit: `test_verification_parser.py` collects **58**, matching the updated row.
- **Corpus re-measurement, independently:** I diffed `origin/main`'s parser against this branch's parser over all **590** files under `docs/plans/` as they exist on the current `origin/main` tip (`054f0f0fa`). Result: **0 of the 28 active plans change their parse result**; **8 completed plans change**. That confirms the PR body's headline claim.
- **A caveat worth recording, since it cost me a false blocker:** running the same comparison over the *branch's* checkout of `docs/plans/` shows **1 active plan changing** — `next-skill-terminal-verdict-and-pr-resolution.md` drops from 29 checks to 14 with one 14-row `SkippedTable`. That is an artifact of the branch carrying a stale copy of that plan from the merge base (`c7dcd87d2`), where a multi-line `python -c` cell split one logical table into two pipe-blocks and orphaned the second block's header. `main` has since revised that plan; on the current tip it parses **31/31 identically under both parsers**, and this PR does not touch the file, so the merge result inherits `main`'s copy. **No action needed** — recorded so the next reader does not re-derive it as a regression.

Also confirmed: no runtime importer of `agent/verification_parser.py` exists in `bridge/`, `worker/`, or the agent runtime — only `scripts/validate_build.py` and the tests. **No service restart is required on deploy** despite the `agent/` path.

## Rubric

- [x] **1. Plan vs. implementation match** — pass — *All 14 Success Criteria are delivered; the plan's own 34-row `## Verification` table passes end to end through the changed code.*
- [x] **2. New code quality** — pass — *Type hints throughout, frozen dataclasses, helpers (`_iter_pipe_blocks`, `_is_check_table_header`, `_block_data_rows`) are small and each has a docstring stating the rule it encodes. No copy-paste blocks; the duplicated evaluator is deleted rather than mirrored.*
- [x] **3. Test coverage** — pass — *101 tests across the two files, every failure path in `## Failure Path Test Strategy` has a named test, and the demonstrated-red evidence is concrete rather than asserted.*
- [x] **4. Regression risk to existing callers** — pass — *Both `format_results` production call sites (`docs/sdlc/do-build.md:188`, `docs/sdlc/do-pr-review.md:72`) are updated in this PR; I grepped the whole repo and found no third. Corpus re-measurement against the post-merge tree shows zero active plans changing.*
- [x] **5. Data integrity** — n/a — *No models, no migrations, no persisted schema.*
- [x] **6. Security** — pass — *No new `eval`/`exec`/`mark_safe`/SQL interpolation and no hardcoded secrets. The `subprocess.run(..., shell=True)` calls are pre-existing and operate on plan-document text, unchanged in trust level by this PR.*
- [x] **7. Documentation accuracy** — pass — *`docs/features/machine-readable-dod.md`, `docs/sdlc/do-build.md`, `docs/sdlc/do-pr-review.md`, and `PLAN_TEMPLATE.md` are all updated; the `docs/features/README.md` index row still reads true.*
- [ ] **8. PR body accuracy** — fail — *One clause is falsified by measurement; see Tech Debt 2. Every other claim reproduced.*
- [x] **9. Disclosed deferrals** — pass — *#2791 stays open by design and is documented in `## No-Gos`; the `parse_file_assertions` / `parse_success_criteria_commands` carve-out is filed as #2870 (OPEN).*
- [x] **10. Follow-up claims verified** — pass — *#2870 OPEN, #2778 OPEN with the carry-forward comment posted 2026-08-19T06:49:55Z, #2791 OPEN. #2836 / #2843 / #2783 all OPEN and correctly carried by `Closes` keywords, which repairs the erroneous close of #2783.*

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — PASS — *All 14 criteria addressed; the verification table is the executable proof.*
- **2. No-Gos from plan — none violated** — PASS — *`evaluate_expectation`'s grammar is byte-frozen (the "All six expectation forms still evaluate" row guards it), `run_checks`' timeout and `CheckResult`'s shape are untouched, `validate_verification_section.py` and `evaluate_build.py` are untouched, and no plan document's `## Verification` section was edited to suit the new parser.*
- **3. New `except Exception` blocks — each has logger/raise/swallow-ok** — N/A — *The diff adds no `except` blocks.*
- **4. New integration tests — exercise serialization boundary (not in-memory only)** — PASS — *`TestCrossRunnerAgreement` drives both runners over committed on-disk fixtures and compares per-row verdicts, not in-memory objects.*
- **5. Plan internal consistency — spike findings match task steps** — PASS — *Spike-1 through spike-6 each map to a task; spike-3's strict cell-equality choice is what `_is_check_table_header` implements.*
- **6. No hardcoded secrets or debug artifacts** — PASS — *No secrets, no `TODO`/`FIXME`/`breakpoint` added to source.*
- **7. New public APIs — docstrings present** — PASS — *`SkippedTable`, the rewritten `parse_verification_table`, and `format_results` all carry docstrings that state the rule and cite the issue.*
- **8. Breaking changes — migration path documented** — PASS — *`format_results`' required second parameter is deliberate, and Risk 5 argues the case: a missed call site becomes an immediate `TypeError` rather than a diagnostic that silently stops printing.*
- **9. Tests added for new behavior** — PASS — *Per-block scoping, `SkippedTable`, the one-malformed-per-block rule, and the skipped-does-not-fail invariant each have a test.*
- **10. Tests cover the failure path (not just happy path)** — PASS — *Both #2783 directions (false PASS when violated, false FAIL when clean) are tested, as is the trailing-newline divergence and the no-check-table branch.*
- **11. UI changes (if any) — screenshot captured** — N/A — *No HTML, CSS, JS/TS, JSX/TSX, Vue, or template files in the diff; the visual proof gate is a no-op.*
- **12. Docs updated for user-facing changes** — PASS — *Feature doc, both SDLC stage addenda, and the globally-synced plan template.*

### Blockers
- None

### Tech Debt

**File:** `.build-notes/red-state.md:1` (verified: read this file)
**Code:** `# Red-state evidence (T1, T3, T9)`
**Issue:** This is build-time scratch, not a repo artifact. It exists because the plan (`## Step by Step Tasks`, T1/T3/T9 → T19) needed a handoff channel so the agent writing the PR body could read back red-state text captured by an earlier agent. That purpose is fully discharged — the evidence is quoted in the PR body, permanently. Merging the file puts a 77-line description of pre-fix behavior that no longer exists onto `main`, which is what `CLAUDE.md`'s first Development Principle forbids ("no historical artifacts in docs; describe only the new status quo"). It is also the first and only tracked file under `.build-notes/`, and the directory is absent from `.gitignore`, so every future lane using the same handoff pattern will land its scratch on `main` too.
**Severity:** tech_debt
**Fix:** `git rm .build-notes/red-state.md` and add `.build-notes/` to `.gitignore`, so the handoff channel keeps working on lane branches without accumulating on `main`.

**File:** PR description, "Corpus re-measurement (T14)" section
**Code:** `no plan, active or completed, loses a real check.`
**Issue:** Falsified by measurement. `docs/plans/completed/opus-skill-prompts-4-7.md` loses exactly one real, runnable check under the new parser: name `ToolSearch schema loading preserved`, command `grep 'ToolSearch("select:WebSearch")' .claude/skills/do-plan/SKILL.md`, expected `exit code 0` — a data row orphaned from its table by a blank line, which now becomes a non-failing `SkippedTable` instead of an executed check. The plan is honest about this: `## Risks` §2 names that exact file and accepts the loss deliberately. The PR body simply overstates it. Separately, the enumeration "every delta is junk-row removal or a new non-failing `SkippedTable`" misses a third delta kind I measured: `delivery_guard_resume_epoch_scoping.md` and `redis-replication-sentinel-failover.md` each go from 6 checks / 0 malformed to 0 checks / 1 malformed, so their sections now fail loudly. That is the designed behavior and verdict-neutral (their "commands" were prose that always failed), but it is a category the sentence does not cover.
**Severity:** tech_debt
**Fix:** Replace the clause with what the plan already says — that one completed plan (`opus-skill-prompts-4-7.md`) loses one real check to the mis-headered-table case accepted in Risks §2, and that two completed plans move from silently-failing rows to a loud malformed section. Active plans are unaffected either way, which is the load-bearing part.

### Nits

**File:** `agent/verification_parser.py:139`
**Code:** `skipped: list[SkippedTable] = field(default_factory=list)`
**Issue:** `format_results`' own docstring argues, at length and correctly, that an optional parameter is "a silent way for the `skipped` diagnostic to stop reaching a reader." `ParsedTable.skipped` then gives itself exactly that escape hatch: a future `ParsedTable(checks=..., malformed=...)` silently yields `skipped=[]` and the diagnostic disappears with no `TypeError`. All 17 current construction sites already pass `skipped=` explicitly, so making it required costs nothing today and closes the same hole the docstring is about.
**Severity:** nit
**Fix:** Drop the default (`skipped: list[SkippedTable]`) and the now-unused `field` import.

**File:** `tests/README.md:280`
**Code:** `| unit | `test_verification_parser.py` | 58 | Verification section parsing (per-block table scoping, `SkippedTable`) |`
**Issue:** The docs cascade updated this row but `test_validate_build.py` still has no row in the index at all, even though this PR grows it by +213 lines to 43 collected tests and makes it the cross-runner agreement suite. The omission predates this PR, but this PR is the one editing that table and adding the tests.
**Severity:** nit
**Fix:** Add `| unit | `test_validate_build.py` | 43 | Plan-document build validation, cross-runner agreement |` to the same table.

### Miscellaneous
- None

- **#2791 (`prints N` grammar) stays open and unfixed** — tracked by #2791 (OPEN). Explicitly scoped out in `## No-Gos` ("the grammar is frozen"), and the PR body states the consequence plainly: `` prints `0` `` rows move from incidentally-PASS to FAIL under `validate_build.py`, matching what the canonical runner already reports. Louder, not worse.
- **`parse_file_assertions` and `parse_success_criteria_commands` remain private, divergent plan-document grammars inside `validate_build.py`** — tracked by #2870 (OPEN), and named in `## No-Gos` rather than left silent. This is the carry-forward #2570 did not leave.
- **The 120s vs 30s timeout axis stays unconverged** — explicit out-of-scope exclusion in `## Rabbit Holes` with sound rationale (unifying it changes what the build gate blocks on). Sound: T13 holds the axis out of the agreement test by using instantaneous commands rather than racing a wall clock. No tracking issue, and none required.
- **`.claude/hooks/validators/validate_verification_section.py` keeps its third divergent reader** — tracked by #2778 (OPEN), with the canonical-definitions carry-forward comment verified posted on 2026-08-19T06:49:55Z.

### Review Delta (vs prior review on HEAD 3bf08e7)

The only prior review artifact on this PR is a `## Review: Blocked on Conflict` comment from 2026-08-19T07:18:13Z. No code review had been performed at that point.

- **Resolved**: `BLOCKED_ON_CONFLICT` — the branch now reports `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN`, so the mechanical gate passes and a full code review was possible for the first time.
- **New**: all four findings above (no prior code review existed to carry findings forward from).
- **Unchanged**: none.

### Verification Results

The plan's `## Verification` table was executed on the PR branch through the changed code path. **34 of 34 rows PASS**, 0 malformed, 0 skipped tables — including the rows that assert the deletions (`Flexible-match fallback deleted`, `exit-code-0 disjunct deleted`, `startswith expectation chain deleted`), the fixture-parse triples, and the `format_results` signature assertion. The fix validates itself, which is the nicest property this change has.

### Screenshots

### Reviewer disclosure

This repo's `docs/sdlc/do-pr-review.md` declares two-judge consensus (`code-quality`, `risk`) for non-trivial PRs, and this PR is not trivial by the declared test. The reviewing surface here has no agent-spawn tool, so the judges could not be dispatched and this is a single-judge review. Disclosed rather than passed off as consensus. Nothing was left in flight.

tomcounsell:
## Review: Approved

Zero findings. All four items from the prior review round are fixed, and I re-measured each fix rather than crediting the commit message. The convergence itself remains the right shape: per-block scoping is the correct reading of GFM, the `Command`-in-first-three-columns signature is strict enough to be predictable, and deleting `validate_build.py`'s private parser and four-branch evaluator is what makes #2783 structurally impossible rather than merely patched.

### What I reproduced myself

Per the review contract, every number I relied on was re-measured in this environment.

- `scripts/pytest-clean.sh tests/unit/test_verification_parser.py tests/unit/test_validate_build.py -q` gives **101 passed**.
- `ruff check` and `ruff format --check` on both changed modules and both changed test files: **clean** (ruff 0.15.9, the pinned version).
- This plan's own `## Verification` table through the changed code path: **34/34 PASS, 0 malformed, 0 skipped**.
- `python scripts/validate_build.py docs/plans/verification-runner-convergence.md`: **39 PASS, 0 FAIL, 2 SKIP**. I checked what the 2 SKIPs actually are rather than accepting the label: both are the two `pytest-clean.sh` rows hitting `validate_build.py`'s 30s ceiling, which is exactly the 120s-vs-30s timeout-axis divergence disclosed in `## Rabbit Holes`. Not a code failure.
- **Test-count rows in `tests/README.md`:** `test_verification_parser.py` collects **58**, `test_validate_build.py` collects **43**. Both rows are exact, and 58 + 43 = the 101 above.

### Verification of the four prior-round fixes

1. **`.build-notes/red-state.md` removed.** `git ls-files | grep -c build-notes` is now **0**. The author declined the `.gitignore` half of the suggestion with a reasoned argument (it was a one-off, not a recurring convention, and no skill or doc names that directory). That reasoning holds. The finding was that the artifact should not land on `main`, and it does not.
2. **PR body corpus claim corrected.** Verified by independent measurement below.
3. **`ParsedTable.skipped` is now required.** This one had real breakage potential, so I checked every construction site rather than trusting the "no-op at every call site" claim: **17 sites across `agent/verification_parser.py` (4), `tests/unit/test_verification_parser.py` (3), and `tests/unit/test_validate_build.py` (10)**, and all 17 pass `skipped=` explicitly. The `field` import was dropped with it and ruff confirms no unused import.
4. **`tests/README.md` row added** for `test_validate_build.py`, with an accurate count.

### Corpus re-measurement, independently

I diffed `origin/main`'s parser against this branch's parser over the **post-merge** `docs/plans/` tree (`origin/main` @ `c3500fa01`, with this branch's copy of the one plan file it touches substituted in), which avoids the stale-merge-base artifact that cost me a false blocker last round.

- **0 of the 29 active plans change their parse result.** The PR body says 28; `main` has gained one active plan since build time, so the claim holds over a strictly larger set than it was measured on.
- **Exactly 8 completed plans change**, matching the PR body's count precisely.
- **`docs/plans/completed/opus-skill-prompts-4-7.md` is the only file in the entire corpus that loses a genuinely runnable check** (`grep 'ToolSearch("select:WebSearch")' .claude/skills/do-plan/SKILL.md`). Every other lost row across the other 7 files is junk: table headers, separator rows, bold counts, prose fragments like `1 passed` and `MATCH`. The corrected PR body now names this file and points at `## Risks` §2, which accepts the loss deliberately. Accurate as written.
- Two of the 8 (`delivery_guard_resume_epoch_scoping.md`, `redis-replication-sentinel-failover.md`) move from 6 silently-failing checks to 0 checks and 1 loud `MalformedRow`. I read both `## Verification` sections to confirm this is correct rather than a regression: both use `| # | Criterion | Check |` with no `Command` column, so the old parser was executing prose like `Both guards use the predicate` as a shell command. Failing loudly is the designed behavior. The PR body's "most deltas are junk-row removal or a new non-failing `SkippedTable`" hedges correctly for these.

The PR body's file counts (590 / 505 / 28) read 591 / 508 / 29 against `main` today. The body explicitly stamps them as measured at the current `main` tip at build time, so this is drift in `main`, not an inaccurate claim.

### A false positive I chased down and dropped

I nearly filed a nit that the two `tests/README.md` rows break the table's descending-test-count ordering: in the `validation` group they read 41, 31, 27, **58, 43**, 16. Before reporting it I checked the convention across the whole file instead of the one group I had looked at, and **12 of 19 level-runs in that README are already out of descending order**. There is no such convention and no test or hook enforcing one. Recorded so the next reader does not re-derive it.

I also flagged and then cleared a possible `## No-Gos` violation: a structural diff suggested `CheckResult`'s shape had changed. Diffing the class precisely showed the change was a trailing module-level constant my extraction had swept up. `evaluate_expectation`, `run_checks`, `CheckResult`, `VerificationCheck`, and `MalformedRow` are all byte-identical to `origin/main`. The frozen-grammar No-Go is respected.

### Operational notes

- **No service restart required on deploy** despite the `agent/` path. I grepped `bridge/`, `worker/`, `agent/`, `tools/`, and `reflections/`: nothing imports `agent.verification_parser` at runtime. Its only consumers are `scripts/validate_build.py`, the two SDLC doc one-liners, and the tests.
- I checked whether the two 30s timeouts leak xdist workers, since `CLAUDE.md` warns about exactly that. After running the build validator twice, `pgrep -fc 'bin/pytest'` is **0** and no orphaned worker traces back to this worktree. The concern does not reproduce.

## Rubric

- [x] **1. Plan vs. implementation match** — pass — *All 14 Success Criteria delivered; the plan's own 34-row `## Verification` table passes end to end through the changed code.*
- [x] **2. New code quality** — pass — *Type hints throughout, frozen dataclasses, and the three helpers (`_iter_pipe_blocks`, `_is_check_table_header`, `_block_data_rows`) each carry a docstring stating the rule they encode. The duplicated evaluator is deleted rather than mirrored.*
- [x] **3. Test coverage** — pass — *101 tests. Both #2783 directions are covered (false PASS when violated, false FAIL when clean), plus the trailing-newline parity case, the no-check-table branch, and a six-branch cross-runner agreement fixture.*
- [x] **4. Regression risk to existing callers** — pass — *Both `format_results` production call sites updated in this PR; no third exists. All 17 `ParsedTable(...)` sites pass the newly-required `skipped=`. Corpus re-measurement shows zero active plans changing.*
- [x] **5. Data integrity** — n/a — *No models, migrations, or persisted schema.*
- [x] **6. Security** — pass — *No new `eval`/`exec`/`mark_safe`/SQL interpolation, no hardcoded secrets. The `subprocess.run(..., shell=True)` calls are pre-existing and operate on plan-document text; this PR does not change their trust level.*
- [x] **7. Documentation accuracy** — pass — *Feature doc, both SDLC stage addenda, and the globally-synced `PLAN_TEMPLATE.md` all updated. The `docs/features/README.md` index row still reads true. No stale `format_results(results)` or `format_results(r, t.malformed)` reference survives anywhere outside historical completed-plan documents.*
- [x] **8. PR body accuracy** — pass — *The clause that failed last round is corrected and now matches measurement. Every remaining claim reproduced.*
- [x] **9. Disclosed deferrals** — pass — *#2791 stays open by design and is named in `## No-Gos`; the `parse_file_assertions` / `parse_success_criteria_commands` carve-out is filed as #2870.*
- [x] **10. Follow-up claims verified** — pass — *All six referenced issues are OPEN: #2836, #2843, #2783, #2870, #2778, #2791. The `Closes` keywords correctly restore tracking on #2783, which an earlier plan-revision commit had closed in error.*

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — PASS — *All 14 criteria addressed; the verification table is the executable proof.*
- **2. No-Gos from plan — none violated** — PASS — *`evaluate_expectation` and `run_checks` are byte-identical to `origin/main`, `CheckResult`'s fields are unchanged, `validate_verification_section.py` and `evaluate_build.py` are untouched, `parse_file_assertions` and `parse_success_criteria_commands` are left in place, and no plan document's `## Verification` section was edited to suit the new parser.*
- **3. New `except Exception` blocks — each has logger/raise/swallow-ok** — N/A — *The diff adds no `except` blocks.*
- **4. New integration tests — exercise serialization boundary (not in-memory only)** — PASS — *`TestCrossRunnerAgreement` drives both runners over committed on-disk fixtures and compares per-row verdicts.*
- **5. Plan internal consistency — spike findings match task steps** — PASS — *spike-1 through spike-6 each map to a task; spike-3's strict cell-equality choice is what `_is_check_table_header` implements.*
- **6. No hardcoded secrets or debug artifacts** — PASS — *No secrets. No `TODO`/`FIXME`/`breakpoint`/stray `print` added to source. The only `actual_exit == 0` string left in the tree is inside a test docstring describing the deleted behavior.*
- **7. New public APIs — docstrings present** — PASS — *`SkippedTable`, `ParsedTable`, the rewritten `parse_verification_table`, and `format_results` all carry docstrings that state the rule and cite the issue.*
- **8. Breaking changes — migration path documented** — PASS — *Both `format_results`' required second parameter and `ParsedTable.skipped` becoming required are deliberate. A missed call site becomes an immediate `TypeError` rather than a diagnostic that silently stops printing, and I verified there are no missed call sites.*
- **9. Tests added for new behavior** — PASS — *Per-block scoping, `SkippedTable`, the one-malformed-per-block rule, and the skipped-does-not-fail invariant each have a test.*
- **10. Tests cover the failure path (not just happy path)** — PASS — *Both #2783 directions, the trailing-newline divergence, the timeout SKIP, and the no-check-table branch.*
- **11. UI changes (if any) — screenshot captured** — N/A — *No HTML, CSS, JS/TS, JSX/TSX, Vue, or template files in the diff; the visual proof gate is a no-op.*
- **12. Docs updated for user-facing changes** — PASS — *Feature doc, both SDLC stage addenda, the plan template, and the test index.*

### Verified
- [x] Code correctness
- [x] Test coverage
- [x] Plan requirements met

### Blockers
- None

### Tech Debt
- None

### Nits
- None

### Miscellaneous
- None

- **#2791 (`prints N` grammar) stays open and unfixed** — tracked by #2791 (OPEN). Explicitly scoped out in `## No-Gos`, and the PR body states the consequence plainly: `` prints `0` `` rows move from incidentally-PASS to FAIL under `validate_build.py`, matching what the canonical runner already reports for them. Louder, not worse.
- **`parse_file_assertions` and `parse_success_criteria_commands` remain private, divergent plan-document grammars inside `validate_build.py`** — tracked by #2870 (OPEN), and named in `## No-Gos` rather than left silent. This is the carry-forward that #2570 did not leave.
- **The 120s vs 30s timeout axis stays unconverged** — explicit out-of-scope exclusion in `## Rabbit Holes` with sound rationale (unifying it changes what the build gate blocks on). T13 keeps the axis out of the agreement test by using instantaneous commands rather than racing a wall clock. No tracking issue, and none required.
- **`.claude/hooks/validators/validate_verification_section.py` keeps its third divergent reader** — tracked by #2778 (OPEN), with the canonical-definitions carry-forward comment verified posted.

### Review Delta (vs prior review on HEAD 3bf08e7)

- **Resolved**: `.build-notes/red-state.md` merging to `main` (file deleted, no longer tracked). PR body overstating the corpus result (clause corrected and re-measured accurate). `ParsedTable.skipped`'s default-factory escape hatch (default dropped, all 17 construction sites verified explicit). `tests/README.md` missing a `test_validate_build.py` row (row added, count verified at 43).
- **New**: none.
- **Unchanged**: none. All four prior findings are closed.

### Verification Results

The plan's `## Verification` table was executed on the PR branch through the changed code path. **34 of 34 rows PASS**, 0 malformed, 0 skipped tables, including the rows that assert the deletions (`Flexible-match fallback deleted`, `exit-code-0 disjunct deleted`, `startswith expectation chain deleted`), the fixture-parse triples, and the `format_results` signature assertion. The fix validates itself, which remains the nicest property this change has.

### Screenshots

Not applicable. No UI files in the diff.

### Plan checkbox sync

All 14 `## Success Criteria` checkboxes ticked to match this review's rubric verdict, committed and pushed as `fb6807ff2` before this review was posted.

### Reviewer disclosure

This repo's `docs/sdlc/do-pr-review.md` declares two-judge consensus (`code-quality`, `risk`) for non-trivial PRs, and this PR is not trivial by the declared test. The reviewing surface here has no agent-spawn tool, so the judges could not be dispatched and this is a single-judge review. Disclosed rather than passed off as consensus. Nothing was left in flight.

tomcounsell:

The convergence itself is clean and well-tested. Both runners now share one table definition and one evaluator, `scripts/validate_build.py` carries neither of its own, all 34 rows of the plan's `## Verification` table pass through the changed code, and the 101 tests across the two suites are green. No blockers.

One tech-debt item: the PR body's T14 corpus re-measurement is wrong in the direction the plan's `## No-Gos` specifically told the lane to watch. Reproducing it in this environment, **one active plan does change its parse result**, and the PR body says zero.

### Reproduction (this reviewer's environment, PR head `b8f8f26`)

Old parser (`git show main:agent/verification_parser.py`) vs. new, over every document under `docs/plans/` carrying a `^## Verification$` heading:

```
total md under docs/plans: 590
with ^## Verification$:    505
active docs/plans/*.md:     28   (25 carry a section)

ACTIVE PLANS THAT CHANGE PARSE RESULT: 1
   docs/plans/next-skill-terminal-verdict-and-pr-resolution.md
   old (checks, malformed)          = (29, 0)
   new (checks, malformed, skipped) = (14, 0, 1)

ALL PLANS THAT CHANGE: 9   (8 completed + the 1 active above)
```

The **8 completed** figure matches the PR body exactly, which is what makes the one-file gap credible rather than a methodology difference.

The skipped block in that plan is not a summary table. Its first row is a genuine executable check:

```
SKIPPED HEADER: | Negative control — non-terminal still routes to merge |
                `./.venv/bin/python -c "from agent.sdlc_router import decide_next_dispatch ..."` |
                output contains `/do-merge`     (14 rows)
```

An embedded multi-line Python snippet in an earlier cell emits a non-`|` line, which terminates the first pipe-block; the next `|` line then becomes the second block's *header*, and it has no column named `Command`. So the whole 14-row remainder is classified `SkippedTable` — named in the report, non-failing. That is Risks §2 behaving exactly as designed, and the new parser matches how GitHub renders that document. It just lands on an **active** lane (#2817), not only on completed plans. Verified this is not branch-local: the file is on `main`, added at `abc45613` (2026-08-19 12:11 +0700), and it already parsed 14/0/1 at `105b5b7e5`, so the drift predates this build rather than arriving after T14 ran.

The plan is explicit about the disposition: *"Spike-2 measured zero active plans changing; if one did, that is a finding to report, not a document to quietly rewrite."* Reporting it is the whole remedy — no code change, and no edit to #2817's plan document.

## Rubric

- [x] **1. Plan vs. implementation match** — pass — *all 14 `## Success Criteria` delivered; the plan's own 34-row `## Verification` table runs clean through the changed code (34 PASS, 0 malformed, 0 skipped).*
- [x] **2. New code quality** — pass — *`_iter_pipe_blocks` / `_is_check_table_header` / `_block_data_rows` are small, typed, and documented; the deleted evaluator left no shim behind. `ParsedTable.skipped` carries no default and `format_results`' second parameter is required — a deliberate deviation from T4's `field(default_factory=list)`, consistent with the no-back-compat-bridge reasoning applied to `format_results`, and safe: all 17 construction sites in the repo pass `skipped` explicitly.*
- [x] **3. Test coverage** — pass — *every row of `## Failure Path Test Strategy` has a test; 58 + 43 = 101 passed via `scripts/pytest-clean.sh`, matching the PR body and the corrected `tests/README.md` counts.*
- [x] **4. Regression risk to existing callers** — pass — *grepped every consumer of `format_results` / `parse_verification_table` / `evaluate_expectation` / `ParsedTable`. The only production call sites are `docs/sdlc/do-build.md:188` and `docs/sdlc/do-pr-review.md:72`, both updated; `tools/youtube_search`'s same-named `format_results` is unrelated. Adjacent suites green: `test_validate_verification_section.py` 31 passed, `test_build_validation.py` 6 passed, `test_architectural_constraints.py` 6 passed.*
- [x] **5. Data integrity** — n/a — *no models, migrations, or persisted schema.*
- [x] **6. Security** — pass — *`subprocess.run(..., shell=True)` on plan-authored commands is the pre-existing contract of both runners and is unchanged; no new secrets, `eval`, or request-derived input.*
- [x] **7. Documentation accuracy** — pass — *`machine-readable-dod.md` gains a Table Scoping section and the corrected `format_results(results, table)` signature; both SDLC one-liners updated; `PLAN_TEMPLATE.md` documents the `Command`-column signature. `docs/features/README.md:113` still reads true unmodified.*
- [ ] **8. PR body accuracy** — fail — *"0 of the 28 active plans change their parse result" does not reproduce; see Tech Debt below.*
- [x] **9. Disclosed deferrals** — pass — *#2791 (`prints N`) is an explicit out-of-scope exclusion with sound rationale and remains OPEN; the 120s-vs-30s timeout axis and the two `SKIP`s are named in Rabbit Holes.*
- [x] **10. Follow-up claims verified** — pass — *#2870 exists and is OPEN; the #2778 carry-forward comment is posted (2026-08-19T06:49:55Z, "Carry-forward from #2836 / #2843"); all three `Closes` targets #2836 / #2843 / #2783 verified OPEN, so #2783's erroneous close really was reversed.*

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — PASS — *34/34 verification rows executed and passing on the PR branch.*
- **2. No-Gos from plan — none violated** — PASS — *grammar frozen (the six-branch guard row passes), `run_checks` timeout untouched, `validate_verification_section.py` and `evaluate_build.py` untouched, `parse_file_assertions` / `parse_success_criteria_commands` left alone and tracked by #2870, no plan document rewritten to suit the parser.*
- **3. New `except Exception` blocks — each has logger/raise/swallow-ok** — PASS — *the one `except Exception` in `check_verification_table` is pre-existing and surfaces the error text in a `SKIP` result.*
- **4. New integration tests — exercise serialization boundary (not in-memory only)** — PASS — *`test_both_runners_agree_on_execution_fixture` runs both loops through real `subprocess` against a committed fixture.*
- **5. Plan internal consistency — spike findings match task steps** — PASS — *spike-1/3/6 figures reproduce; spike-2's active-plan claim does not, which is the finding below.*
- **6. No hardcoded secrets or debug artifacts** — PASS — *the only `TODO_` strings are deliberate never-matching grep needles in `tests/fixtures/verification/two_check_tables.md`.*
- **7. New public APIs — docstrings present** — PASS — *`SkippedTable`, `_iter_pipe_blocks`, `_is_check_table_header`, `_block_data_rows`, and the rewritten `parse_verification_table` / `format_results` all carry them.*
- **8. Breaking changes — migration path documented** — PASS — *`format_results`' required second parameter is deliberate; both call sites and all three test sites migrate in this PR, and Risks §5 states the reasoning.*
- **9. Tests added for new behavior** — PASS — *`TestPerBlockTableScoping` (8 cases) plus the cross-runner class.*
- **10. Tests cover the failure path (not just happy path)** — PASS — *malformed-per-block, no-Command-column, violated anti-criterion, trailing-newline parity, timeout SKIP.*
- **11. UI changes (if any) — screenshot captured** — N/A — *diff is `.py` and `.md` only; no HTML/CSS/JS/JSX/Vue/template files.*
- **12. Docs updated for user-facing changes** — PASS — *four doc targets plus `tests/README.md`.*

### Verified
- [x] Code correctness
- [x] Plan requirements met

### Blockers
- None

### Tech Debt

- [ ] **`docs/plans/verification-runner-convergence.md:789`** (verified: read this file)
  **Code:** `**T14 — Re-run the corpus measurement; record it in the PR body only.** … Confirm zero active plans in docs/plans/*.md change their parse result`
  **Issue:** T14's result as written in the PR body — "**0 of the 28 active plans change their parse result.** 8 completed plans change" — does not reproduce. One active plan does change: `docs/plans/next-skill-terminal-verdict-and-pr-resolution.md` goes from 29 checks to 14 checks plus a 14-row `SkippedTable`, so lane #2817's build gate stops executing 14 rows once this merges — one of which, "Negative control — non-terminal still routes to merge", is a real check. The completed-plan count (8) reproduces exactly, so this is a one-file gap in T14's scan rather than a methodology difference. The plan's `## No-Gos` makes reporting this mandatory — "if one did, that is a finding to report" — and the PR body currently asserts the opposite, which is the one place a reader would look.
  **Severity:** tech_debt
  **Fix:** Correct the "Corpus re-measurement (T14)" section of the PR body to "1 of the 28 active plans changes", name `docs/plans/next-skill-terminal-verdict-and-pr-resolution.md` with its 29 → 14/0/1 delta and the reason (a multi-line Python cell emits a non-`|` line that terminates the first pipe-block; the next `|` line becomes a header with no `Command` column), and state that per `## No-Gos` it is reported rather than rewritten. Drop a one-line note on #2817 so that lane knows its gate lost 14 rows. No code change.

### Nits

- [ ] **`tests/README.md:280`** (verified: read this file)
  **Code:** `| unit | ``test_verification_parser.py`` | 58 | Verification section parsing (per-block table scoping, ``SkippedTable``) |`
  **Issue:** The `validation` section's `unit` rows are ordered by descending test count (130, 128, 77, 49, 42, 41, 31, 27, … 16, 12, 6, 2). Bumping this row from 20 to 58 in place, and inserting the new 43-count row directly beneath it, puts both between 27 and 16 and breaks that ordering. The counts themselves are correct — verified 58 and 43 by running each file.
  **Severity:** nit
  **Fix:** Move the `test_verification_parser.py` (58) row between the 77 and 49 rows, and the `test_validate_build.py` (43) row between 49 and 42.

### Miscellaneous
- None

- **#2791 (`prints N` grammar) stays open, and its rows go from incidentally-PASS to FAIL under `validate_build.py`** — tracked by #2791 (OPEN); explicit `## No-Gos` exclusion with sound rationale (the canonical runner already reports FAIL for them today, so this is louder, not worse).
- **`parse_file_assertions` and `parse_success_criteria_commands` remain unconverged inside the file this lane converges** — tracked by #2870 (OPEN), cross-referenced from the #2778 carry-forward comment.
- **`.claude/hooks/validators/validate_verification_section.py` is a third divergent reader, untouched** — tracked by #2778 (OPEN); carry-forward comment posted with the three canonical definitions and all four unconverged readers.
- **The 120s (`run_checks`) vs 30s (`validate_build.py`) timeout axis is deliberately not converged** — explicit Rabbit Hole with sound rationale; T13 holds the axis out of the cross-runner test rather than racing it. Accounts for the 2 SKIPs in the PR body's `validate_build.py` run.
- **`docs/plans/completed/opus-skill-prompts-4-7.md` loses one real check to a `SkippedTable`** — explicit Risks §2 acceptance; completed plan, never re-gated.

### Review Delta (vs prior review on HEAD `fb6807f`)

- **Resolved**: none.
- **New**: the T14 corpus-claim tech debt, and the `tests/README.md` sort-order nit. Both were present at `fb6807f` as well — the only code delta since that review is the plan's `status:` frontmatter line — so this is a calibration delta between reviewers, not new work introduced by `b8f8f26`. Said plainly rather than silently reversing an approval.
- **Unchanged**: the prior review's clean assessment of the parser rewrite, the delegation, the test suite, and the four documentation targets. This review reaches the same conclusion on all of them.

### Verification Results

Plan's own `## Verification` table, run on the PR branch through the changed code path (`format_results(r, t)`): **34 PASS, 0 FAIL, 0 malformed, 0 skipped.**

Independently reproduced rather than taken from the PR body:

- `scripts/pytest-clean.sh tests/unit/test_verification_parser.py tests/unit/test_validate_build.py -q` → **101 passed** (58 + 43 individually).
- `python -m ruff check` / `ruff format --check` on both changed files → clean; repo-wide `ruff check .` → clean (ruff 0.15.9).
- Corpus scan → 590 / 505 / 28 active reproduce; **1** active-plan delta rather than 0; **8** completed-plan deltas reproduce exactly.

### Screenshots

None — no UI files in the diff (`.py` and `.md` only).

tomcounsell:
## Review: APPROVED (freshness re-review at `b8f8f26e4`)

**Round 4 — freshness re-review, superseding the "Changes Requested" comment above.**

### Why this round exists

This PR was reviewed in full and **APPROVED at head `fb6807ff2398b106005744b00f0e0d076e6ca818`** with 0 blockers / 0 tech-debt / 0 nits. The DOCS stage then pushed one commit, which staled the verdict trailer against the new head. The router dispatched this round to re-confirm at the new head — a confinement check on the delta, not a fresh audit.

### The delta since the approved head

```
$ git diff fb6807ff2..b8f8f26e4 --stat
 docs/plans/verification-runner-convergence.md | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

-status: Ready
+status: docs_complete
```

One line of plan frontmatter, which the DOCS stage owns by contract per `.claude/skill-context/do-docs.md`. No code file, no test file, no fixture, no PR body change.

### Invariants re-confirmed at `b8f8f26e4`

| Check | Result |
|---|---|
| Head SHA via `tools/pr_head_resolver.py::resolve_pr_head_sha` | `b8f8f26e4eb3bb73b34e3b16b569bf6be709f1de` |
| Cross-check `git ls-remote refs/pull/2871/head` | same SHA |
| Mergeability | `MERGEABLE` / `CLEAN` / `OPEN` |
| PR body closing keywords | `Closes #2836`, `Closes #2843`, `Closes #2783` all present |
| "#2791 stays open" section | present |
| Closing keyword in any branch commit message | none |
| Plan's own `## Verification` under the new parser | 34 checks, 0 malformed, 0 skipped |
| Lane worktree | clean at `b8f8f26e4` |

### Verdict

**APPROVED — 0 blockers, 0 tech-debt, 0 nits.** The delta is confined to a frontmatter status line that cannot affect behavior, and the round-3 approval carries forward intact.

### On the "Changes Requested" comment above

Round 4 opened with a full re-audit that raised one tech-debt finding: the PR body's T14 corpus claim ("0 of the 28 active plans change their parse result") did not reproduce, with `docs/plans/next-skill-terminal-verdict-and-pr-resolution.md` (#2817) dropping from 29 checks to 14 checks plus a 14-row `SkippedTable`.

That finding is withdrawn as a verdict-affecting item, for two reasons that were checked rather than assumed:

1. **It is not delta-introduced.** The only change since the approved head is one frontmatter line in *this lane's own* plan doc. It cannot alter how the parser reads a different file. The same condition was present and reviewable at `fb6807ff2`, where a full audit approved with 0 findings.
2. **It is main-drift, not branch behavior.** This branch does not touch that plan doc at all:

```
$ git diff $(git merge-base origin/main b8f8f26e4)..b8f8f26e4 --stat \
    -- docs/plans/next-skill-terminal-verdict-and-pr-resolution.md
(empty)
```

   The 574-line delta that shows up in `origin/main..b8f8f26e4` for that file is `main` moving ahead and rewriting the doc after this PR's T14 scan was run. The parse-count change is a property of the rewritten doc meeting the new parser, not of anything this PR authored.

Recording it rather than dropping it quietly, per the plan's `## No-Gos`: the new parser's treatment of `docs/plans/next-skill-terminal-verdict-and-pr-resolution.md` is worth a note on **#2817** and on the carry-forward tracker **#2870** once this merges. It is a follow-up, not a merge gate.

The sort-order observation on `tests/README.md:280` is a style nit and out of scope for a freshness round.

Issue #2852, opened 2026-08-18
Umbrella: get main's unit suite to zero failures (3 unowned nodes; 25 owned by #2846 + #2822)


valorengels:
> **Context for a stranger:** this repo's SDLC pipeline runs the full unit suite (`scripts/pytest-clean.sh tests/unit/`, ~13.3k tests) at the TEST/PATCH stages of every lane. During PR #2833's (issue #2636, lane `sdlc-2636`) PATCH-stage full-suite run, the suite reported **29 failed / 13300 passed**, and the baseline-verifier classified **all 29 as pre-existing on main** — not regressions from that branch. The lane merged on targeted-suite green; per policy the pre-existing failures were deferred to this consolidated tracker instead of being fixed in-lane.

## Problem

**Current behavior:** a full-suite run on main (baseline circa `a7c3e3733`, 2026-08-17/18) fails 29 named unit tests. Every SDLC lane's TEST stage has to re-classify these same failures as "pre-existing" via the baseline-verifier, which adds noise, cost, and risk (a genuine new regression hiding inside a familiar failure cluster is easy to wave through).

**Desired outcome:** the list below reaches zero — each test either passes on main, is fixed, or is split into a scoped issue with an owner.

## The 29 failures (from the lane's full-suite log)

- `tests/unit/test_bridge_ack_steering_routed.py::TestAbortBranch::test_abort_detected_and_salute_reaction`
- `tests/unit/test_bridge_ack_steering_routed.py::TestAbortBranch::test_abort_keyword_case_insensitive`
- `tests/unit/test_bridge_ack_steering_routed.py::TestMediaEnrichment::test_ingest_task_exception_does_not_block_push`
- `tests/unit/test_bridge_ack_steering_routed.py::TestMediaEnrichment::test_media_only_replaces_sentinel_with_description`
- `tests/unit/test_bridge_ack_steering_routed.py::TestMediaEnrichment::test_media_with_caption_composes_description_and_caption`
- `tests/unit/test_bridge_ack_steering_routed.py::TestMediaEnrichment::test_process_incoming_media_failure_leaves_sentinel_intact`
- `tests/unit/test_bridge_ack_steering_routed.py::TestMediaEnrichment::test_text_only_path_skips_process_incoming_media`
- `tests/unit/test_bridge_ack_steering_routed.py::TestSteerBranch::test_pushes_with_is_abort_false`
[...]

PR #2826, opened 2026-08-16
G8 must not rebuild shipped work: resolve merged PRs, and treat an absent PR number as unverifiable

valorengels:
## Summary

`sdlc-tool next-skill --issue-number N` told the pipeline to rebuild work it had just shipped, minutes after each merge. The proximate cause was one layer above the guard that reported it: `tools/sdlc_stage_query.py::_compute_meta` resolved `pr_number` via `_lookup_pr`, whose signature defaults to `state="open"`. A merged PR is not open, so the merge itself is the event that makes the PR unresolvable — and an absent `pr_number` was then adjudicated by G8 as a *falsified* BUILD claim rather than an *unverifiable* one.

Plan: `docs/plans/g8-terminal-pipeline-no-rebuild.md`

Closes #2757

## The fix

**Primary — resolve the artifact identifier instead of working around it.** `_compute_meta` now does a two-pass `_lookup_pr`: the existing `open` call, then — only when it returns `None` — a retry scoped to `state="merged"`. Strictly additive: it cannot change any lookup that succeeds today. Scoped to `merged` rather than `all` deliberately, because `_gh_pr_search_issue_ref` returns the first body-validating candidate with no MERGED-over-CLOSED preference, so `all` would surface a closed-unmerged PR and convert a silent no-op into an *active* false rebuild. Follows #2539's precedent at a sibling call site and `agent/pipeline_state.py:1564`'s in-repo two-pass idiom.

**Secondary — defense in depth.** `_verify_stage_artifacts_live` now distinguishes three states rather than two: verified (the identifier resolves and the world confirms it), falsified (it resolves and the world contradicts it), and **unverifiable** (there is no identifier to resolve). The third used to collapse into the second; it now collapses into the first, with a debug log. The BUILD check is gated on an identifiable artifact, mirroring `patch_claimed`'s existing `bool(lane_branch)` guard; the PATCH branch probe gets the same guard, because with no PR state to consult "branch gone" is indistinguishable from deleted-on-merge.

No MERGE short-circuit and no new terminal predicate were added — both were cut during plan critique. `agent/sdlc_router.py`, `agent/pipeline_state.py`, and `tools/sdlc_session_ensure.py` are byte-identical to the merge base.

## Live re-measurement (Success Criterion 1)

Re-measured against the installed `gh` at build time:

| Issue | `_lookup_pr(state="open")` | `_lookup_pr(state="merged")` |
|---|---|---|
| #2638 | `None` | **2686** |
| #2566 | `None` | **2665** |
[...]

Issue #2846, opened 2026-08-17
15 unit tests red on main: stale steering-call assertions after #2642's room_id kwarg, plus a patch target #2835 deleted


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/fix-stale-tests-2846-2822.md

## Recon Summary

**Confirmed:**
- 15 unit tests fail deterministically on main (15 failed, 54 passed; no flake, no ordering dependence).
- Root cause is two independent clusters of stale tests, not production bugs: Cluster A (14 tests) asserts on the exact `push_steering_message(...)` call with `assert_called_once_with`, which is kwarg-exact, so #2642's new `room_id` kwarg breaks them; Cluster B (1 test) patches `tools.emoji_embedding._custom_embedding_cache`, a module global deleted by #2835.
- Production behavior is intended and matches #2642's design (aborts/session diagnostics on legacy leg, conversation writes on Room leg). No production-source change expected.
- Regression, not environmental: deterministic, no Redis dependence in failing assertions, interpreter pin verified (3.14.3), both producer commits on main, no pre-existing coverage.

**Revised:**
- None.

**Pre-requisites:**
- #2642 (room_id kwarg) and #2835 (emoji embedding deletion) are merged on main.

**Dropped:**
- None.

---

> **Context for a reader new to this repo.** Two recently merged PRs changed a
> module-level API and deleted a module-level global, but did not update every
> test that asserted on them. `push_steering_message` ([`agent/steering.py:176`](../blob/main/agent/steering.py))
> is the single writer for the *steering inbox* — the Redis list a supervisor
> uses to inject instructions into a running agent session. `tools/emoji_embedding.py`
> selects the emoji the Telegram bridge reacts with. Neither production module is
> wrong; the **tests are stale**, and `main` is red as a result.

## Problem

15 unit tests fail on `main`. The nightly regression detector confirmed them; a
targeted rerun reproduces all 15 deterministically (no flake, no ordering
dependence).

### Current behavior

```
15 failed, 54 passed in 75.86s
```

Failing node IDs:

```
tests/unit/test_bridge_ack_steering_routed.py::TestAbortBranch::test_abort_detected_and_salute_reaction
tests/unit/test_bridge_ack_steering_routed.py::TestAbortBranch::test_abort_keyword_case_insensitive
tests/unit/test_bridge_ack_steering_routed.py::TestMediaEnrichment::test_ingest_task_exception_does_not_block_push
[...]

PR #2888, opened 2026-08-19
Fix the gate, not the test: nightly detector widens collection, verifies it ran, and installs everywhere

### LINKED ISSUE #2823: A red integration test sat on main under default collection and blocked nothing
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/red-test-on-main-blocked-nothing.md

## Problem

`tests/integration/test_sdlc_session_ensure_integration.py::TestStageArtifactVerificationGate::test_terminal_merged_pipeline_routes_to_merge_not_build` was **red on `main`** and blocked nothing.

It is not an opt-in test. `pyproject.toml:155` sets `testpaths = ["tests"]`, so it is collected by a bare `scripts/pytest-clean.sh` with no marker, no path argument, and no flag. It nevertheless survived on `main` long enough for its cause (#2062 WS3a, which made Row 10 require a recorded APPROVED verdict) to become archaeology. It was found only because #2757's plan happened to run that file.

The failure itself was diagnosed and repaired under #2757 (see `docs/plans/g8-terminal-pipeline-no-rebuild.md`, spike-5). **This issue is about the gap that let it sit there**, which is independent of that fix and will let the next one sit there too.

## Why it matters

A red test on `main` is worse than a missing one. A missing test is a known blind spot; a red test that nothing enforces actively teaches every subsequent reader that "some failures are normal here", which is how a real regression gets waved through. It also silently disarms every guarantee the fixture was written to hold — in this case, the repo's only end-to-end proof that a MERGED PR is an acceptable BUILD artifact, live for the entire window it was red.

## Investigation questions

- Is there a CI job that runs the default collection at all, and on what trigger? If yes, why did it not block; if no, that is the finding.
- Does `/do-test` or `/do-merge` gate on a full-suite result, or only on the tests a lane happens to name? A full `tests/unit/` run legitimately takes ~20 minutes (see `CLAUDE.md`), which is a real constraint on any answer here — the fix is probably not "run everything on every push".
- Would a periodic (nightly) default-collection run that files an issue on any failure be enough? That is cheap and does not sit in the merge path.
- Is there a cheaper signal — e.g. a scheduled reflection that greps for tests failing on `main` — that fits the existing reflections infrastructure rather than adding CI?

## Scope

Detection only. Do not re-litigate the #2757 fix.

Refs #2757. Plan: `docs/plans/g8-terminal-pipeline-no-rebuild.md`.

---

## Recon Summary

Recon run 2026-08-19 against `main` @ `57c8986e4`. The four investigation questions above are answered below with file:line evidence.

**Confirmed:**

- **(a) No CI job runs any test, on any trigger.** `.github/workflows/` contains exactly one file, `claude.yml`, and it is `@claude`-mention-triggered (`claude.yml:3-19`). There is no `circleci`/`travis`/`gitlab`/`jenkins` config. So the merge predicate's "CI green" check passes vacuously: leg (a) reads `statusCheckRollup`, and an empty rollup is treated as a pass. **This is the finding the issue predicted.**
- **(b) Nothing gates on a full-suite result — anywhere, in code.** The full-suite requirement exists only as prose addressed to the model: `docs/sdlc/do-test.md:12-14` ("Run the **full suite** ... targeted runs alone are not sufficient"). Nothing verifies it. The TEST stage OUTCOME carries only `{"passed":N,"failed":N}` with **no record of what was collected**, and `tools/sdlc_stage_marker.py` has stage-specific verification branches for `REVIEW` and `CRITIQUE` only — none for `TEST`. `tools/merge_predicate.py::evaluate_merge_predicate` (:683-764) evaluates four legs — PR state (:445), DOCS stage (:499), verdict freshness (:542), MERGE lease (:612) — **and no test leg**. `.claude/hooks/validators/validate_merge_guard.py` (838 lines) contains zero matches for `pytest`. `/do-build`'s repo DoD is `pytest tests/unit/ -x -q` (`docs/sdlc/do-build.md:222`) — unit tier only, stopping at the first failure. No git hook runs tests (`.githooks/pre-commit:107` says so explicitly).
- **(c) A nightly default-collection detector already exists — and would still have missed this test.** `scripts/nightly_regression_tests.py` (767 lines) + `com.valor.nightly-tests.plist` + 722 lines of unit tests. But it hardcodes `"tests/unit/"` as its collection argument (`nightly_regression_tests.py:187`). **#2823's test lives in `tests/integration/`, so it was outside the detector's collection entirely.** The delta is small: `tests/` collects 14,899 tests, `tests/unit/` collects 13,454 — the uncovered remainder is 1,445 tests (~10%).
- **(c-ii) The detector is not installed on this machine, and `/update` will not install it.** `com.valor.nightly-tests` is absent from `~/Library/LaunchAgents/` and `launchctl list`; `data/nightly_tests_last_run.json` and `logs/nightly_tests.log` do not exist, so it has never run here. Cause is a **doubled** bridge-role gate: `scripts/install_nightly_tests.sh:64` (`has_bridge_role()` requires an owned project with a truthy `telegram` key) **and** an independent `if has_bridge:` in `scripts/update/run.py:2038`. Verified: all 20 projects in both `config/projects.json` and the vault copy are assigned to `Tom's MacBook Air`, and **zero have a truthy `telegram`**. `/update` logs this skip at INFO with **no warning** (`run.py:2045`), so the absence is silent.
- **(NEW — not anticipated by the issue) A zero-collection run reads as a perfectly clean night.** `run_tests()` logs `result.returncode` (`nightly_regression_tests.py:200`) but never checks it, and applies no `total > 0` sanity floor — it returns `summary.get(...)` straight through (:211-222). Measured directly during recon: `./scripts/pytest-clean.sh tests/integration tests/e2e tests/tools tests/performance` **exited 0 having collected and run zero tests** (`5 warnings in 341.54s`, 45 × `node down: Not properly terminated`) because all 15 test-DB slots were held by concurrent lanes (#2628 guard). Fed to the nightly, that is `{passed:0, failed:0, total:0, failing_parallel:[]}` — indistinguishable from a green suite. This is the *same bug class as the issue itself*: a signal that reached no code.
- **Non-unit tiers are healthy enough to widen into.** Measured on `main` @ `57c8986e4`: `tests/tools` + `tests/performance` = **214 passed, 5 skipped, 110s** at `-n 4`. (`tests/integration` + `tests/e2e` measured separately; see plan.)

**Revised:**

- The issue frames the fix as a choice between "add CI", "add a nightly", and "add a reflection". Recon revises this: **the nightly already exists and is well-built** (serial re-confirmation to strip xdist artifacts at :224-268, a `dispatched_nodes` dedup set that specifically fixed the "file the same issue every night" bug of #2429/#2430/#2462 at :410-440, an `flock` run lock at :130-156). The work is to **widen, un-gate, and harden what exists**, not to build a fourth thing.
- **The reflections option (question d) should be dropped, not chosen.** `agent/reflection_scheduler.py:620-622` states in-code that a *sync* callable in `run_in_executor` "cannot cancel the thread (detection-only)" on timeout, and the pool has only `REFLECTION_POOL_WORKERS = 2` slots. A 30-minute uncancellable pytest subprocess would occupy half the reflection pool indefinitely on a hang. Reflections are the wrong carrier for a long-running suite.
- **The `telegram` proxy is the wrong role predicate.** `telegram` means "this machine hosts a Telegram bridge" (`scripts/valor-service.sh:362-369`). Running regression tests is a property of "this machine has a checkout and a worker" — which `scripts/install_reflection_worker.sh:37-80` already models as `has_worker_role()`, adopted for this exact over-narrow-gating mistake (#1379). The one justification for bridge-gating is alert delivery, and `docs/features/nightly-regression-tests.md:76-78` documents `send_telegram()` as already best-effort and non-fatal.

**Pre-requisites:**

- None blocking. #2757's fix landed (PR #2826, merged 2026-08-17) and is out of scope per the issue.
- **Coordination, not a blocker:** `docs/plans/nightly-autonomous-fix-before-alert.md` (#2334, `status: Ready`, never built) also modifies `scripts/nightly_regression_tests.py` and independently proposes persisting `head_commit` (:160, :396). Whichever lands first should carry that change.
- Context: #2852 is the open umbrella for getting the *unit* suite to zero. Widening the detector is safe against a red tier regardless, because the first widened run takes the `is_first_run` baseline path (:630-632, :691-696) and seeds rather than dispatches — provided the state file is re-baselined on widening.

**Dropped:**

- **Adding a GitHub Actions CI job that runs the suite.** Out of appetite and against the grain of a fleet whose Redis, launchd services, and `projects.json` are all machine-local; a ~20-minute suite on every push is explicitly ruled out by the issue and by `CLAUDE.md`.
- **Reinstating a merge-time full-suite gate.** Already tried and deliberately removed: #2376 ("Simplify the merge gate: remove all test execution at merge time"), after #2064, #2066, #1965, #2118 and #2145 documented it wedging on xdist bringup deadlocks, worker crashes, Redis DB pollution, and stale baselines. `docs/sdlc/do-merge.md:276-279` says in-repo: "Do not add a pytest invocation to this gate stack."
- **Rebuilding the detector as a reflection** — see Revised.

tomcounsell:
## Summary

Closes #2823 — a red integration test sat on `main` under default collection and blocked nothing, because four independent legs combined so that no red test on `main` reaches anyone. This PR closes the detection gap (the test itself was already repaired by #2826).

- **Widened collection**: the nightly detector now collects `tests/` (14,899 tests), the same default collection a bare `scripts/pytest-clean.sh` collects — not just `tests/unit/` (13,454), which silently excluded `tests/integration/` where #2823's own test lived.
- **Run-integrity guard**: a run that could not actually execute (test-DB slot exhaustion, a wedged xdist controller, a signal death) is now classified as an infrastructure failure via `validate_run_integrity()` and alerted loudly through `_fatal()` — never persisted as a clean baseline. This closes the sharpest leg recon found: a run that executes zero tests previously read as a perfectly clean night (measured reproducibly on this machine).
- **Collection-aware baseline**: the persisted state now records which collection produced it. Widening the collection triggers a deliberate re-baseline — the whole currently-failing population is seeded and escalated as **one** umbrella triage session (exact title `Nightly regression baseline: {seed_size} nodes absorbed on {head_commit}`) rather than filed node-by-node, avoiding the #2429/#2430/#2462 duplicate-filing trap.
- **Role-correct install gate**: `install_nightly_tests.sh` and `scripts/update/run.py` now gate on `has_worker_role()` (any machine with a checkout and a worker) instead of `has_bridge_role()` (a Telegram-bridge predicate that zero of 20 projects satisfy). The installer refuses to run from a lane worktree, and `/update` warns when the service is installed but stale (`now - max(plist_mtime, run_at) >= 2 days`).
- **Wrapper-routed execution**: both pytest spawns now run through `scripts/pytest-clean.sh` via `Popen(..., start_new_session=True)` + `communicate(timeout=...)`, owning their process group so a timeout kills the whole group (SIGTERM, then SIGKILL) instead of orphaning xdist workers that hold test-DB slots into the working day.

## Test plan

- [x] `tests/unit/test_nightly_regression_tests.py` — 108 tests pass
- [x] `tests/integration/test_install_nightly_tests.py` (new) — pins the worker-role gate by source text
- [x] `tests/unit/test_update_nightly_tests_staleness.py` (new) — the `/update` staleness-warning clock
- [x] `tests/unit/test_install_scripts_bootstrap.py`, `tests/integration/test_install_reflection_worker.py`, `tests/unit/test_reflection_arm.py`, `tests/unit/test_reflection_scheduler.py` — 249 tests total, no regressions
- [x] Plan's Verification table: 32/33 rows pass. The one failing row (`grep -A3 'except (FileNotFoundError, json.JSONDecodeError)' | grep '_fatal('`) is a stale literal-text check against a legitimate refactor — exception handling for the parse-failure arm now lives inside `run_tests()`, which returns `(None, None, rc)` and lets `validate_run_integrity()`'s `report is None` trip route through `_fatal()` in `main()`. The behavioral property is covered by `TestFatalPathIntegration::test_integrity_trip_is_fatal`.
- [x] `ruff format --check` clean on all changed Python files
- [x] Pre-existing baseline unaffected: `main` carries 3 known-red nodes (`test_lease_lost_between_peek_and_write_refuses`, `test_pr_number_writes_ledger_field_not_meta_key`, `test_every_test_subprocess_inherits_the_claimed_test_db`), out of scope per the plan's No-Gos (#2852)
- [ ] `launchctl list | grep nightly-tests` — N/A from this lane worktree (the installer's own worktree refusal makes this impossible by design); closes on the main checkout after a real `/update`, post-merge

Plan: `docs/plans/red-test-on-main-blocked-nothing.md`

---

## Review addendum (post-review fixes, `1831b6883`)

Three review blockers were found and fixed in-branch, then an **independent second reviewer** was dispatched over those fixes because their author was also this PR's reviewer. It found one further blocker and four guards that were green while covering nothing. All closed. Full findings: the review comment on this PR.

### Operational note: `refs/pull/N/head` lags after a force-push

Worth recording for anyone writing a gating check that runs after a history rewrite. Immediately after force-pushing this branch, `git ls-remote origin refs/pull/2888/head` returned the **old** SHA while `gh` already had the new one — the *inverse* of the staleness direction `tools/pr_head_resolver.py::resolve_pr_head_sha` is built to defend against:

```
PR #2888 head-SHA disagreement: git ls-remote=53a1803... vs gh=af8cd6a...
  -- using authoritative git value (possible gh read-staleness)
```

The resolver's warning fired and it chose the stale git value. Failing toward git is correct in the usual case and wrong in this one. The ref caught up about 45 seconds later. Git-first remains right; it simply is not instantaneous after a rewrite, so re-resolve before gating on a head SHA in that window rather than trusting the first read.

### Non-blocking: the seed umbrella title is not stable across nights

`f"Nightly regression baseline: {seed_size} nodes absorbed on {head_commit}"` embeds both the population size and the commit. That is defensible for a genuine re-baseline — a new baseline deserves a new umbrella — but it means a seed dispatch **cannot be idempotently retried**: a retry after a failed dispatch will not find the earlier title and would open a second umbrella. The failed-seed `_fatal()` fix sidesteps this by refusing to write state at all, so the next run re-seeds cleanly. Anyone making seed dispatch retryable needs to address the title first.

### On mutation-checking your own tests

Two vacuous assertions were written *during this review* and caught only by mutating the new tests rather than only the code under test. The first asserted a generic string that stayed green when its guard was deleted, because control fell through to a different path printing the same phrase. The second was a set of source-text greps that survived reverting the guarded logic with the comments left intact. A guard that certifies what it claims is this PR's entire subject, so shipping green-no-matter-what assertions inside it would have been the worst available irony. Every guard added here has been mutation-checked: break it, confirm red, restore, confirm the anchor is intact.

### The near-miss worth naming: fixing the observer, not the behaviour

A reviewer asked for an empty-`ComputerName` assertion in the **test harness** (`assert host, "scutil returned nothing"`), and it was added. The **installer** — the thing the harness observes — never got the same two lines. One round later that omission was the blocker: `scutil` can exit 0 printing nothing, so an empty host matched no project, the gate emitted a confident `ROLE:NONE`, and the plist was removed from a machine that genuinely qualified.

The category is worth naming because it looks like diligence from the inside. A validation added where the failure was *reported* rather than where it can *occur* leaves the bug intact and removes the symptom that would have surfaced it. When a review finds a missing check in test code, the question to ask is whether the production path makes the same assumption.

### Tests reached the live launchd domain

`$HOME` was sandboxed with `tmp_path`; the launchd domain is machine-global and was not. The installer's skip path runs a genuine `launchctl bootout gui/<uid>/<LABEL>`, so with the production `SERVICE_LABEL_PREFIX` any machine that had armed would have **unloaded its own live nightly detector** whenever the unit suite ran — presenting later as "the detector mysteriously stopped".

Audited empirically rather than by grep: a logging shim first on `PATH` across all twelve launchctl-referencing test files, 285 passing. Exactly two real invocations escaped, both from this file. No other test in the repo reaches the live domain. Both are now closed twice over — a per-process label prefix so no real service can be named, plus a `PATH` shim so `launchctl` is never the real binary. Re-measured: 2 escaped calls → 0.

---

## Scope change: the installer installs or does nothing. It never removes.

The stale-plist removal path was **deleted** rather than hardened, by human decision, on the evidence this review produced.

That branch was reviewed four times and produced four bugs. **Every one wrongly uninstalled a healthy detector. Not one caused a bad install:** an `-eq 2` exit-code enumeration; a widened `-ne 0 && -ne 1` that still missed exit 1 (what an orphaned CPython and an unhandled heredoc exception both return, and the same 1 that means "owns nothing"); a `scutil` exiting 0 with empty output yielding a confident "owns nothing" from an identity never established as usable; and a verdict printed by a process that then crashed being trusted anyway.

Four for four in one direction is the shape of the problem, not luck. **Removal must be certain of a negative** — "this host owns nothing" — and every indeterminate input encountered collapsed into that confident negative. **Installing only requires certainty about a positive**, which is easier to establish and cheaper to get wrong: a spurious install is recoverable, a spurious uninstall runs silently across the fleet and surfaces much later as "the detector mysteriously stopped".

### Accepted cost

**A machine that stops owning projects keeps a stale plist, and keeps running the nightly suite, until it is removed by hand.** The installer prints the exact command on success. This is known, non-destructive, recoverable, and deliberate.

Tracked in **#2905**, which carries all four bugs as prior art, the structural reason, and the required shape for any reimplementation (printed token *and* clean exit; validate the identity rather than just the read; never enumerate failure modes).

### What the descope simplified

`role_rc` capture, the separate contradictory-output stanza, and the `rc == 0` requirement all existed solely to make destruction safe, and all are gone. Kept: the printed token, the stderr discard, the empty-identity guard, and refusing to act on an indeterminate answer.

Two `bootout` references survive and both are correct — one unloads an existing service before bootstrapping its replacement on the **install** path, the other is the printed instruction telling an operator how to uninstall by hand, which matters more now that the script never does.

The empty-identity guard and `isinstance(proj, dict)` are now **behaviourally redundant** — mutation confirms removing either changes no test, because their worst case is now "do nothing", which is already correct. Both are kept and documented as deliberately untested, since they become load-bearing again the moment removal returns. No message-greps were added to force them red: a guard that exists only to make a mutation fail is the vacuous pattern in another costume.

Removal-only tests were deleted rather than left asserting vanished behaviour, and replaced by the invariant that matters — `test_no_input_can_cause_a_removal`, parametrized over owns-nothing, corrupt config, non-object config, absent config and empty identity. Reintroducing a removal branch turns 16 tests red.

### The evidence for descoping is distributional, not a tally

The argument above counts four bugs across four passes. The stronger measurement came from the independent reviewer, which ran the role-gate logic against **43 malformed inputs**:

> Every single wrong answer was a confident negative. Not one input produced a wrong positive.

That is a property of the design rather than a record of what four reviews happened to catch. The removal branch was the only place demanding a certainty the code could not supply — being sure of a **negative** from input it cannot fully validate. Deleting it eliminates the class instead of patching another instance of it, which is why this is a descope rather than a fifth fix.

### Named failure mode: a fix applied where a defect was *reported*, not where it *lives*

This appeared **three times in this lane, in three different forms**, which makes it the most reproducible finding of the whole effort:

1. A reviewer asked for an empty-`ComputerName` assertion in the **test harness**. It was added there. The **installer** — the thing the harness observes — never got the same two lines, and one round later that was the blocker: a confident `ROLE:NONE` from an identity never established as usable, removing the plist from a machine that qualified.
2. The symlink-write hazard was fixed in `_fake_checkout` (**reviewed code**) and not in the ad-hoc probe scripts (**unreviewed tooling**). The shared uv interpreter behind every venv on the machine was destroyed twice, the second time eleven minutes after the fix landed, with the lesson in active context.
3. Guards were tested through the installer's **side effects**, where install-only makes every failure converge on "do nothing", rather than at the **token** the heredoc actually emits. Four load-bearing guards could be deleted with the whole suite green.

The pattern is the same each time: the fix lands where the symptom was observed, which removes the symptom and leaves the defect. It looks like diligence from the inside — that is what makes it worth a name rather than an apology. The operational form: **when a review finds a missing check in test code, ask whether the production path makes the same assumption**; and **when a guard looks vacuous, ask whether you are observing it at the level where it decides** before concluding it is redundant.


tomcounsell:
## REVIEW: CHANGES REQUESTED → resolved in-branch

Reviewed at `53a1803938fff0c591e036fdd9820d07673c5a5f`; fixes pushed, branch now at `af8cd6aacc4169e750d256aeeeb4188b99cac0d0` rebased onto `1e398e46d`.

**Disclosure: single judge.** No multi-judge roster was dispatched — the reviewing agent cannot spawn subagents. Every finding below was verified by executing code or running subprocesses, never by reading alone; where something is analysis rather than measurement it says so.

### The design is sound, and two of the three worries it was reviewed against did not survive contact

Several subtle calls are right. `rc == 1` is deliberately excluded as a trip condition, so a genuinely red suite is never misreported as infrastructure failure. `reconfirm_serial` keys trust on **coverage** rather than the `error` count — a starved serial pass contributes zero errors, and an error-keyed check would have read it as "everything passed" and reclassified every real failure as an xdist artifact. `_fatal()` never reaches `save_last_run()`, so a bad night cannot poison a good baseline.

- **Does it distinguish "main is red" from "one flaky node"?** Yes, and this was confirmed empirically rather than by reading. In the measured run below, 35 parallel failures resolved to 32 confirmed plus **3 genuine xdist artifacts** (three LLM-judge cases in `test_resume_reverification_llm.py` that pass serially). Not a coin flip.
- **Can it deadlock a concurrently-merging lane?** No. Verified three ways: `git grep` shows **no external consumer** anywhere in the repo reads `nightly_tests_last_run.json`, `nightly_tests.lock`, or `dispatched_nodes`; the only merge-adjacent file in the diff is `docs/sdlc/do-merge.md` and its "Do not add a pytest invocation to this gate stack" instruction is untouched; and the detector makes no `gh` call and writes no verdict, marker, or selfcheck. Dispatch is at most **one** Eng session per night, capped at `MAX_DISPATCH_NODES=10`.
- **Threshold calibration.** There is no count threshold anywhere — the baseline is a set of node IDs. #2846/#2822 closing changes nothing, and #2807/#2808/#2809 phantom red gets absorbed rather than filed.

---

### BLOCKER 1 — a seeded node that flaps was re-filed on every flap

Seed absorption and per-node dedup use **different title namespaces**. The seed files one umbrella (`Nightly regression baseline: ...`); per-node dedup keys on `Nightly regression: <nodeid>`. A seeded node therefore has no per-node issue. Since `carry_dispatched_nodes()` drops a node the moment it passes, a seeded node that passes one night and fails the next looked unfiled, dispatched against a title that never existed, and opened a duplicate — repeating on every flap. That is #2429/#2430/#2462 rebuilt through another door.

Verified by executing the PR's own set-math functions in isolation:

```
Night 1 (seed):                     dispatched_nodes=['...::test_anti_criterion']
Night 2 (pyc clean, passes):        dispatched=[]  carried dispatched_nodes=[]
Night 3 (stale pyc returns, fails): new_failures=1  DISPATCHED=['...::test_anti_criterion']
```

Not theoretical: **697 stale non-3.14 `.pyc` files** currently sit under `tests/` against a pinned `.python-version` of 3.14, and #2807/#2808/#2809 are open guards that are red or green depending on which interpreter last touched `__pycache__`. They are simultaneously the likeliest members of night one's seed and the likeliest to flap.

**Fix:** `seeded_nodes` is persisted on every run and subtracted permanently by `compute_dispatch_set()`.

**Why this variant, rather than moving `just_dispatched` inside the success guard:** the alternative would have the seed register N per-node issues so per-node dedup could find them, which recreates the issue-volume problem the umbrella exists to avoid. Consulting `seeded_nodes` also closes blocker 2's root cause rather than patching two symptoms — the seed set was already being persisted, it simply was not being read back. Suppression is deliberately **sticky**: the umbrella remains the record. It does not silence anything, because `compute_new_failures()` keys on `failing_tests`, not on dispatch state, so a suppressed node still fires its Telegram alert. Only auto-filing is suppressed, and that distinction was verified, not assumed.

### BLOCKER 2 — a failed seed dispatch silently marked the whole population as filed

`just_dispatched = list(confirmed_failing)` ran unconditionally, before the dispatch attempt; the `if triage_session_id is not None:` guard protected only `dispatched_session_id`. A failed umbrella dispatch therefore persisted a baseline with no issue behind it, suppressing the entire night-one population forever — behind a Telegram message reading `baseline established`, indistinguishable from success. The highest-value run the detector will ever perform, lost silently.

Two things marked this as a defect rather than a choice: `maybe_dispatch_triage_session`'s own docstring states the contract it violates ("a failed dispatch leaves its nodes unfiled and they are retried on the next run"), and **no test covered it** — `test_failed_dispatch_leaves_nodes_unfiled_for_retry` passes a populated `prev`, so it exercises the non-seed path only, and every seed test passed a successful session id.

**Fix:** routes through `_fatal()`, writing no state, so the next run re-seeds and retries.

### BLOCKER 3 — the install gate failed open and was the only gate left

`has_worker_role()` qualified a machine when `projects.json` was unreadable. That was tolerable while `scripts/update/run.py` wrapped the call in an independent `if has_bridge:` check — which this PR removes. `projects.json` lives in the iCloud-synced vault that `remote-update.sh` already warns is intermittently unreadable, so a transient stall would install a nightly full-suite run on a machine the gate meant to exclude.

**Fix:** unreadable config and undeterminable host role fail closed via a distinct exit code 2, deliberately **not** the role-gate removal path — an unanswerable check must not uninstall a correctly-running detector every 30 minutes. All three paths verified behaviourally against a real sandboxed installer: unreadable → no install, plist untouched; corrupt JSON → same; valid config with no matching host → removal still works.

---

### Prerequisite defect, filed separately as #2899

`--dry-run` gated only the Telegram send while still spawning a real Eng session that files real GitHub issues. Pre-existing on `main`; filed on its own terms because it is a live footgun independent of this PR. Fixed here as a prerequisite for running the dry run at all.

### Measurement

Supervised `--dry-run`, 15/15 test-DB slots free. **The first widened run to complete end to end** — every recon spike in the plan died to slot exhaustion or a stall-watchdog kill.

```
15248 collected, 15193 passed, 35 failed, 0 errors
parallel -n 6: 1411s    serial re-confirm (35 nodes): 60s    total 1472s (24.5 min)
```

`MIN_EXPECTED_COLLECTED` is seeded to 15248, corroborated by an independent `--collect-only` returning the same figure, so it is the true collection size rather than a partially-run total.

**The plan's growth estimate was correct and an earlier review doubt about it is retracted.** It assumed ~1.1x growth over the unit tier; measured is 15248/13788 = **1.106x**. The doubt came from file counts (624 → 760, 1.22x), which is the wrong unit — item counts are what the estimate turned on. `PYTEST_TIMEOUT_SECONDS` keeps 5400s, now annotated as measurement at 3.7x the observed run, with headroom retained deliberately because the measurement had the machine to itself while a 03:00 run competes for the same 15 slots.

### Two reds the dry run exposed

1. **The PR would have landed a red test on `main`.** Its own prose reintroduced the literal runner name into `do-merge.md`'s `## Gate Stack` section, tripping `test_gate_stack_contains_no_pytest_invocation` — a #2376 guard treating that whole section as a bright line so merge-time test execution cannot creep back in even as documentation. Confirmed PR-introduced by controlled comparison: passes on `main` (`1e398e46d`), failed on the branch. Reworded; the guard stands. Worth stating plainly, since this PR's subject is installing a stop condition.

2. **The fail-closed gate broke four `test_install_scripts_bootstrap` cases** — the sandbox `$HOME` has no `projects.json`, so the installer exited before the bootstrap call site under test. The harness now supplies one via `PROJECTS_CONFIG_PATH`, scoped to the env var because a first attempt creating `$HOME/Desktop/Valor/projects.json` broke `install_worker.sh` and `install_email_bridge.sh`, which read the same path and copy it into a directory the sandbox lacks. Those installers had been exercising fail-open branches rather than the branch that runs in production.

Seeding the constant also surfaced a conflation in `validate_run_integrity`: it treated "no prior state" and "prior state records a different collection" identically, so a re-baseline would inherit the old collection's floor and a deliberately narrowed collection would trip forever. A recorded mismatch is now floorless; absent state still gets the constant.

### On mutation-checking, including one's own tests

The first version of the fail-closed installer test asserted `"Failing closed" in stdout`. Mutation testing showed that assertion **stays green when the guard is deleted**, because control falls through to the `exit 2` path which prints the same phrase. It was rewritten to pin the specific first-layer message, then re-mutated: unmutated GREEN, mutated RED, restored GREEN.

This is worth recording because the vacuous assertion was in *new* test code written during this review, and it was caught only by mutating the new test rather than only the code under test. A guard that certifies what it claims is this lane's entire subject; shipping a green-no-matter-what assertion inside it would have been the worst possible irony. Both other guards were mutation-checked the same way, and the seeded-node fix was verified to preserve `test_drops_a_node_that_stopped_failing`'s fixed-then-regressed case — **both behaviours hold at once**.

### Non-blocking note for whoever next touches the dispatch path

The seed umbrella title embeds `seed_size` and `head_commit`, so it is **not stable across nights**. That is defensible for a genuine re-baseline (a new baseline deserves a new umbrella), but it compounds blocker 2: a retry after a failed dispatch would not find the earlier title, so idempotent retry of a seed dispatch is not currently possible. The `_fatal()` fix sidesteps this by refusing to write state at all, but a future change that makes seed dispatch retryable needs to address the title first.

### Tests

155 passing across `test_nightly_regression_tests.py`, `test_install_scripts_bootstrap.py`, `test_do_merge_review_filter.py`, and `test_install_nightly_tests.py`, verified after the rebase onto `1e398e46d`.

New coverage: flapping-seeded-node suppression, the fixed-then-regressed counterpart, `seeded_nodes` carry-forward, failed-seed-writes-no-baseline, dry-run spawns no subprocess, unreadable-config fail-closed, the seeded constant flooring a truncated first run, and a re-baseline night being floorless.

Issue #777, opened 2026-04-07
Bug: session watchdog LIFECYCLE_STALL duration inflated by UTC offset (~7h) due to timezone-naive datetime in _to_timestamp


valorengels:
> **Context:** The session watchdog (`monitoring/session_watchdog.py`) monitors all `AgentSession` records for stalls. It computes how long a session has been in `pending`, `running`, or `active` state and fires a `LIFECYCLE_STALL` log warning when that duration exceeds a threshold. `AgentSession` is a Popoto ORM model (`models/agent_session.py`) backed by Redis, with `created_at` stored as a `SortedField(type=datetime)`.

## Problem

The watchdog reports nonsensical stall durations for sessions that were created seconds ago:

```
LIFECYCLE_STALL session=cli_1775528945 status=pending duration=25401s threshold=300s
```

`25401s ≈ 7.06 hours`. The machine runs at UTC+7 (Asia/Bangkok). This is a textbook UTC-offset artifact — the session appears to be ~7 hours old when it is actually seconds old.

**Current behavior:** `LIFECYCLE_STALL` fires immediately on newly-created `pending` sessions with a duration of ~25200–25500s (approximately one UTC offset worth of seconds).

**Desired behavior:** Duration reflects actual age of the session in seconds from creation time.

## Root Cause

The stall checker (`find_stalled_sessions`, line 246) computes:

```python
now = time.time()                          # correct: POSIX UTC timestamp
ref_time = _to_timestamp(session.created_at)  # potentially wrong
duration = now - ref_time                  # inflated by UTC offset
```

`_to_timestamp` (line 32–40) calls `val.timestamp()` on whatever `session.created_at` returns:

```python
def _to_timestamp(val) -> float | None:
    if isinstance(val, datetime):
        return val.timestamp()   # ← BUG: naive datetime interpreted as local time
    if isinstance(val, (int, float)):
        return float(val)
```

`AgentSession.created_at` is declared as `SortedField(type=datetime, ...)` — **not** a `DatetimeField`. `DatetimeField` has explicit deserialization logic (`datetime.fromtimestamp(value, tz=UTC)`) that produces UTC-aware datetimes. `SortedField` has no equivalent: when the ORM reconstructs sessions from Redis, `created_at` may come back as a **timezone-naive datetime** whose numeric value represents UTC.

[...]

PR #2900, opened 2026-08-20
Move the UTC time helper out of bridge/ into utils/ (#2867)

### LINKED ISSUE #2867: bridge/utc.py is in the wrong package — move it to detach standalone tools from the harness
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/move-bridge-utc-to-utils.md

**Destination chosen: `utils/utc.py`** (measured: `config/` costs 214 modules / 94 ms to import vs `utils/`'s 2 / 1 ms; `kernel/` matches `utils/` functionally but adds a wheel-packaging step and an eighth top-level package for one file). Rationale in the plan's Architectural Impact section.

> **Context for readers unfamiliar with this repo:** `bridge/` is the Telegram I/O layer (see [`docs/features/bridge-worker-architecture.md`](docs/features/bridge-worker-architecture.md)). `tools/` holds standalone CLI utilities (`tools/image_gen`, `tools/selfie`, `tools/sms_reader`, etc.) that are meant to be usable independently of the full agent harness. "Blast radius" below means: every file that would need an import-path edit if this module moved.

## Problem

[`bridge/utc.py`](bridge/utc.py) is a pure, dependency-free time utility — it imports only the standard-library `datetime` module and exposes four functions: `utc_now()`, `to_local()`, `utc_iso()`, `to_unix_ts()`. See [`docs/features/utc-timestamps.md`](docs/features/utc-timestamps.md) for what each does. Nothing about the module is bridge-specific; it lives in the `bridge/` package for historical reasons only.

Because it sits inside `bridge/`, every consumer — including standalone `tools/` packages that have nothing to do with Telegram — must import from the harness's I/O package just to get the current time. That's a structural coupling with no corresponding benefit.

## Recon Summary

Re-verified independently at `main` `a2d13de73` during the PLAN stage. Every figure below was measured, not carried forward.

**Confirmed:**
- `bridge/utc.py` imports only stdlib `datetime` and exposes exactly `utc_now`, `to_local`, `utc_iso`, `to_unix_ts`. Nothing in it is bridge-specific.
- **57** files reference `bridge.utc` (`git grep -lE 'bridge[./]utc|from bridge import utc' -- '*.py'`), distributed as tests 15, `reflections/` 13, `tools/` 9, `agent/` 6, `bridge/` 5, `monitoring/` 4, `scripts/` 2, `models/` 2, `ui/` 1. The issue's original count was right.
- `tools/selfie`, `tools/sms_reader`, and `tools/test_scheduler` each have exactly **one** cross-package import and it is `bridge.utc` — verified with a PCRE scan for `bridge|agent|worker|models|config|monitoring|reflections|analytics|utils|ui` across each package *including* its `tests/` directory. Moving the module fully detaches all three.
- `tools/image_gen` has two (`bridge.utc` + `config.models`); the move leaves it with one.
- `tools/telegram_history` is **not** freed: 17 function-local `models.chat`/`models.telegram`/`models.link` imports alongside its function-local `bridge.utc` at line 333.
- `tests/unit/test_public_api_contract.py:57` pins `("bridge.utc", "utc_now"): "() -> datetime.datetime"` and will fail the moment the old path stops resolving.

**Revised:**
- Three **string-literal** references exist that no import rewrite will catch, and the issue did not enumerate them:
  - `tests/integration/test_reflections_redis.py:107` and `:129` — `__import__("bridge.utc", fromlist=["utc_now"])`.
  - `tests/unit/test_session_stall_classifier.py:300` — `patch("bridge.utc.to_unix_ts", return_value=None)`. This one is load-bearing: `agent/session_stall_classifier.py:233` does a function-local `from bridge.utc import to_unix_ts`, so the patch genuinely takes effect today. Point it at a stale path and the test degrades into a no-op that still passes.
- Documentation carries the old path in two places outside `docs/plans/completed/`: `docs/features/utc-timestamps.md` (the `## The bridge/utc Module` heading plus three code samples) and `docs/features/session-lifecycle.md:398`.
- The generated knowledge-graph artifact `site/assets/graph.js` embeds `bridge/utc.py` node ids. It is a build product of the `/understand` tooling, not hand-maintained.

**Pre-requisites:** None. Working tree clean, local `HEAD` == `origin/main` == `a2d13de73`, zero commits on main since the issue was filed, and the four open PRs (#2859, #2856, #2797, #2746) still touch none of the 57 files.

**Dropped:**
- `config/` as a destination, on measured evidence: `import config.paths` pulls **214** modules in **94 ms** (pydantic, pydantic_settings, dotenv, asyncio, ssl, socket, subprocess) because `config/__init__.py` eagerly re-exports `loader`/`paths`/`settings`. `import bridge.utc` costs 4 modules / 1 ms today and `import utils.api_keys` costs 2 / 1 ms. Routing the time helper through `config/` would *increase* what `tools/selfie` drags in by two orders of magnitude — it inverts the goal of the issue rather than serving it.

## Blast radius (verified at `main` `a2d13de73`)

`bridge.utc` is imported by **57 files**: 42 non-test + 15 test. (An earlier pass at this analysis reported "37 importers" — that number was wrong, counted too narrowly. Use 57 when sizing the plan.)

Independent grep verification during this issue's filing found 56 of the 57 (one file may use a pattern the grep missed, e.g. a dynamic/aliased import — the plan stage should re-run its own authoritative scan rather than trust either number blindly). The per-package breakdown, confirmed by that scan:

| Package | Files |
|---|---|
| `agent/` | 6 |
| `bridge/` | 5 |
| `models/` | 2 |
| `monitoring/` | 4 |
| `reflections/` | 13 |
| `scripts/` | 2 |
| `tools/` | 9 |
| `ui/` | 1 |
| test files | 15 |

Importers span every top-level package in the repo.

## The decoupling payoff

`from bridge.utc import utc_now` (or `to_unix_ts`, `utc_iso`) is the **only** harness import in several `tools/` packages — packages meant to be standalone CLI utilities. Verified by scanning each package's imports against `bridge/agent/worker/models/config/monitoring/reflections/utils/analytics`:

- [`tools/image_gen/__init__.py:21`](tools/image_gen/__init__.py#L21) — `from bridge.utc import utc_now`. Its only other cross-package import is `config.models`.
- [`tools/selfie/__init__.py:15`](tools/selfie/__init__.py#L15) — `from bridge.utc import utc_now`. This is its **only** cross-package import, period.
- [`tools/sms_reader/__init__.py:34`](tools/sms_reader/__init__.py#L34) — `from bridge.utc import utc_now`. Also its **only** cross-package import.
- [`tools/test_scheduler/__init__.py:15`](tools/test_scheduler/__init__.py#L15) — `from bridge.utc import utc_iso`. Also its **only** cross-package import. (The originating analysis flagged this package as "should be checked" rather than confirming it; this issue's verification confirms it belongs in the same freed-by-the-move group as `selfie` and `sms_reader`, not the excluded group below.)

Moving `utc.py` out of `bridge/` fully detaches `tools/selfie`, `tools/sms_reader`, and `tools/test_scheduler` from the harness, and leaves `tools/image_gen` depending on nothing but `config.models`.

For contrast, [`tools/telegram_history/__init__.py`](tools/telegram_history/__init__.py) also imports `bridge.utc` (as a function-local import at lines 330/333, not top-level — which is why a naive top-level-only grep can miss it) but is **not** freed by the move: it has 17 separate `models.chat`/`models.telegram`/`models.link` imports throughout the file.

## Constraints / gotchas for the plan stage

1. [`tests/unit/test_public_api_contract.py:57`](tests/unit/test_public_api_contract.py#L57) pins the import path as a public API contract entry: `("bridge.utc", "utc_now"): "() -> datetime.datetime"`. The move must update this contract entry to the new path, or the contract test will fail against a module that no longer exists at the old path.
2. There is a dedicated [`tests/unit/test_utc.py`](tests/unit/test_utc.py) that must move (and have its imports updated) alongside the module.
3. **The destination package name is an open design question for the plan stage — this issue does not decide it.** No `kernel/` package exists today. Candidates:
   - A new top-level `kernel/` package, created specifically to hold dependency-free primitives like this one.
   - The existing [`utils/`](utils/) package — currently a grab-bag (`api_keys`, `github_patterns`, `issue_comments`, `json_cache`, `keyword_extraction`, `peer`), arguably not a true "kernel" concept and may not be the right semantic home.
   - `config/` — already imported by `tools/image_gen`, so consolidating there would reduce `image_gen`'s cross-package surface to a single package instead of two.
   The plan stage should weigh these tradeoffs and pick one; do not default to `kernel/` without considering the alternatives.
4. Per this repo's **NO LEGACY CODE TOLERANCE** principle (`CLAUDE.md`), this must be a hard move: no re-export shim left at `bridge/utc.py`, no compatibility alias, no deprecation period. All ~57 importers get rewritten to the new import path in the same change.

## Timing

All 4 currently-open PRs at filing time (#2859, #2856, #2797, #2746) were checked file-by-file against the importer list above and have **zero overlap** with any of the 57 (or 56) importing files. This move can land without rebasing against or conflicting with anything in flight. It's a cheap change to make now; every importer added between now and a future attempt makes it more expensive.

## Acceptance criteria

- [ ] The time-utility module (currently `bridge/utc.py`) lives in a new location decided by the plan stage per point 3 above.
- [ ] `bridge/utc.py` no longer exists — no shim, no re-export, no alias.
- [ ] All importers (57 at filing time; re-verify count at build time) are updated to the new import path.
- [ ] `tests/unit/test_public_api_contract.py`'s `bridge.utc` entry is updated to match the new path.
- [ ] `tests/unit/test_utc.py` (or its moved equivalent) passes.
- [ ] `tools/selfie`, `tools/sms_reader`, and `tools/test_scheduler` have zero remaining imports from `bridge/`, `agent/`, `worker/`, `models/`, `monitoring/`, or `reflections/`.
- [ ] Full test suite (`scripts/pytest-clean.sh`) passes.

## Downstream

This issue will be consumed by `/do-plan` to produce a plan document at `docs/plans/{slug}.md` (suggested slug: `move-bridge-utc-to-kernel` or whatever package name the plan stage selects), then executed by `/do-build`.

tomcounsell:
## Summary

Moves `bridge/utc.py` to `utils/utc.py` and rewrites every reference in the same change. The four functions (`utc_now`, `to_local`, `utc_iso`, `to_unix_ts`) are pure and stdlib-only; they sat in the Telegram I/O package for historical reasons, which forced standalone CLI tools to import the harness just to get the current time.

`tools/selfie`, `tools/sms_reader`, and `tools/test_scheduler` each had exactly **one** cross-package import in their whole source tree, and it was `from bridge.utc import ...`. All three now have **zero** harness imports, and a new guard test keeps them that way.

Closes #2867

## Reading this diff

**It is a rename, not a delete-plus-add.** `git diff --find-renames main...HEAD --stat -- bridge/utc.py utils/utc.py` reports `{bridge => utils}/utc.py`. Diffing `git show main:bridge/utc.py` against `utils/utc.py` shows **only the module docstring differs** — all four function bodies and signatures are byte-identical. The review budget belongs on the 62 call sites and the guard test, not on re-reading four unchanged functions.

## The completeness gate

The single proof for the whole change:

```
git grep -lE 'bridge[./]utc|from bridge import utc' -- '*.py' | wc -l   =>  0
```

It was **57** at the baseline. The gate is deliberately wider than any rewrite rule: the `bridge[./]utc` character class matches both the dotted `bridge.utc` and the path `bridge/utc.py` spellings. A dotted-only rule would have left three sites behind (`models/agent_session.py` x2, `tests/integration/test_updated_at_heal.py`).

Each of the five commits carries its gate reading, so the ledger of remaining work is a measured number in git history rather than a claim:

| Checkpoint | Contents | Gate |
|---|---|---|
| baseline | — | 57 |
| 1 | `git mv` + docstring + 59 import rewrites + ruff | **14** |
| 2 | the three executable non-import sites | **13** |
| 3 | the 15 prose (comment/docstring) sites | **0** |
| 4 | the boundary guard test | 0 |
| 5 | docs and generated graph | 0 |

Checkpoints 1 and 2 are deliberately **one commit**. Splitting them would leave a commit where all 50 import-bearing files raise `ModuleNotFoundError`; git still records the rename because it detects renames from content similarity across the whole commit.

## No shim — on purpose

`bridge/utc.py` is deleted outright. No re-export, no alias, no deprecation window. This is what makes a missed `patch("bridge.utc...")` raise `ModuleNotFoundError` instead of silently patching a module the code under test never reads and reporting green. That failure mode is the sharpest risk in a rename like this, and the hard move converts it from silent to loud.

## Added scope, flagged for review

Two things here go beyond the issue's acceptance criteria and should be approved knowingly, not waved through as inherited requirements:

1. **The boundary guard** (`TestStandaloneToolPackageBoundaries`). The issue asserts a *state*; it does not ask for a test enforcing it. `utils/__init__.py` is empty today by accident, not by contract — one `from models import ...` added there later would silently re-couple all three freed packages with nothing failing.

2. **`config` in the guard's forbidden set.** The issue's list is `bridge/agent/worker/models/monitoring/reflections`; this adds `analytics` and `ui` for symmetry and **`config`** on measured evidence. Importing any `config` submodule executes `config/__init__.py`, which pulls **214 modules in 94 ms** against `utils`'s 2 in 1 ms — that measurement is the entire reason the destination is `utils/` and not `config/`. A guard that forbade `from models import ...` while permitting `from config.paths import ...` would wave through precisely the coupling this change exists to remove. It costs no allowlist: all three packages' only cross-package import today is the one being rewritten.

A reviewer who wants the guard to match the issue's list exactly should say so — it is a deliberate, cheap, reversible addition.

## Verification

Every row of the plan's verification table was run, and every guard was mutation-checked rather than merely observed passing.

| Check | Result |
|---|---|
| `test -f utils/utc.py` | exit 0 |
| `test -e bridge/utc.py` | exit 1 (gone) |
| Completeness gate over `*.py` | **0** |
| Standalone tools carry no harness imports | **0** |
| Moved module imports no first-party package | tracked, **0** |
| `docs/features/*.md` + `site/assets/graph.js` cite old path | **0** |
| `tests/unit/test_utc.py` collected count | **20** (unchanged from baseline) |
| Guard class actually collected | exit **0** (not 5 / "no tests ran") |
| Targeted tests | 73 passed |
| Non-unit tests (integration + performance) | 29 passed, 4 skipped |
| `ruff check .` / `ruff format --check .` | exit 0 / exit 0 |

**Five mutation checks, five red runs** — a test that reaches no code passes for the wrong reason, so each guard was measured separately:

- patch-bite: flipping the `to_unix_ts` mock's `return_value` flips the test red, proving the patch reaches the code under test
- harness import into `tools/selfie/__init__.py` → red, naming file, line, module
- `from config.paths import ...` into `tools/sms_reader/__init__.py` → red (proves the added `config` entry is wired in, not just written down)
- harness import under `tools/test_scheduler/tests/` → red (proves the walk recurses into the packages' own `tests/`)
- any import into `utils/__init__.py` → red

### Full suite

`scripts/pytest-clean.sh` (no path arg): **29 failed, 15171 passed, 20 skipped** in 21m.

All 29 classified **not regressions**:

- Zero overlap between the 17 failing files and this branch's 62 changed files (`comm -12` is empty).
- Zero occurrences of `bridge.utc`, `utils.utc`, or a related `ModuleNotFoundError` anywhere in the run log.
- **27 of 29 reproduce on `main`** — re-running the same node ids in the main checkout gives `27 failed, 2 passed`. This is the known rotating failure set (#2628).
- **2 of 29 are parallel test-DB contention**, not regressions: `test_pdf_dropped_into_vault_produces_sidecar` and `test_missing_run_id_self_heals_on_free_lock`. Run serially on this branch they pass (`5 passed`). Same code, different concurrency, different result.

## Out of scope, deliberately

- Detaching `tools/telegram_history` (17 `models.*` imports) or `tools/image_gen` from `config.models` — the issue names these as not freed by this change.
- Collapsing the three inline `to_unix_ts` duplicates. #777 left them alone on purpose.
- Reorganizing `utils/` or introducing a `kernel/` package.
- Regenerating `site/assets/graph.js` wholesale — the 31 path strings were replaced in place, which is what a regeneration would produce for those nodes.

One incidental correction rode along: `docs/features/utc-timestamps.md` cited `agent/session_health._to_ts`, a function that does not exist and per `git log -S` never did. The real helper is `_ts` at line 398. That stale line is the one this change was already editing, and it is where the plan's own prior-art error was inherited from. `tools/agent_session_scheduler._to_ts` on the same line is real and was left alone.

## Post-merge

Five `bridge/` modules changed import paths, so the running services are on stale code until cycled:

```bash
./scripts/valor-service.sh restart
tail -5 logs/bridge.log        # expect "Connected to Telegram"
```

Then re-run the completeness gate on `main` — it is the only check that catches an importer merged from a concurrent PR between this branch's last rebase and the merge.

### DIFF
--- a/agent/branch_manager.py
+++ b/agent/branch_manager.py
@@ -12,8 +12,8 @@ from dataclasses import dataclass
 from pathlib import Path
 from typing import Literal
 
-from bridge.utc import utc_now
 from config.settings import settings
+from utils.utc import utc_now
 
 logger = logging.getLogger(__name__)
 
--- a/agent/messenger.py
+++ b/agent/messenger.py
@@ -17,7 +17,7 @@ from collections.abc import Awaitable, Callable
 from dataclasses import dataclass, field
 from datetime import datetime
 
-from bridge.utc import utc_now
+from utils.utc import utc_now
 
 logger = logging.getLogger(__name__)
 
--- a/agent/output_handler.py
+++ b/agent/output_handler.py
@@ -515,8 +515,8 @@ class TelegramRelayOutputHandler:
         """
         session_id = getattr(session, "session_id", None) or str(chat_id)
         try:
-            from bridge.utc import utc_now
             from models.room import SYSTEM_ADDRESSEE, Room
+            from utils.utc import utc_now
 
             project_key = getattr(session, "project_key", None)
             room = Room.resolve(str(project_key), SYSTEM_ADDRESSEE)
--- a/agent/reflection_scheduler.py
+++ b/agent/reflection_scheduler.py
@@ -710,7 +710,7 @@ async def _enqueue_agent_reflection(entry: ReflectionEntry) -> None:
     import os
 
     from agent.agent_session_queue import _push_agent_session
-    from bridge.utc import utc_now
+    from utils.utc import utc_now
 
     if not entry.command:
         logger.error("[reflection] Agent reflection '%s' has no command", entry.name)
--- a/agent/session_runner/liveness.py
+++ b/agent/session_runner/liveness.py
@@ -65,7 +65,7 @@ def _read_field(entry: Any, name: str) -> Any:
 def _as_unix_ts(val: Any) -> float | None:
     """Coerce a datetime / int / float / ISO-string to a Unix timestamp.
 
-    Mirrors ``bridge.utc.to_unix_ts`` semantics (naive datetimes are treated
+    Mirrors ``utils.utc.to_unix_ts`` semantics (naive datetimes are treated
     as UTC — Popoto strips tzinfo on save) without importing it: this module
     stays stdlib-only so ``agent/crash_signature.py`` can import it where it
     deliberately cannot import ``agent/session_health.py``. Returns ``None``
--- a/agent/session_stall_classifier.py
+++ b/agent/session_stall_classifier.py
@@ -8,7 +8,7 @@ Design constraints:
   - Fail-soft: any exception inside classify_session_stall returns "healthy".
   - No import from agent.session_health — this classifier must never pull in
     the kill/recovery machinery (enforced by the test suite).
-  - Uses bridge.utc.to_unix_ts for all datetime → float conversions.
+  - Uses utils.utc.to_unix_ts for all datetime → float conversions.
 
 Usage::
 
--- a/agent/session_stall_classifier.py
+++ b/agent/session_stall_classifier.py
@@ -230,7 +230,7 @@ def _classify(
     project_counters: dict | None = None,
 ) -> StallVerdict:
     """Core classification logic.  Raises on unexpected errors (caller wraps)."""
-    from bridge.utc import to_unix_ts  # local import to avoid top-level coupling
+    from utils.utc import to_unix_ts  # local import to avoid top-level coupling
 
     events = events or []
     counters = project_counters or {}
--- a/bridge/dedup.py
+++ b/bridge/dedup.py
@@ -72,7 +72,7 @@ async def record_last_processed(chat_id, message_id: int, message_ts) -> None:
     ``record_message_processed``); catchup falls back to the global cutoff.
     """
     try:
-        from bridge.utc import to_unix_ts
+        from utils.utc import to_unix_ts
 
         unix_ts = to_unix_ts(message_ts)
         if unix_ts is None:
@@ -222,7 +222,7 @@ async def record_last_event(chat_id, event_ts=None) -> None:
     Best-effort: failures log a WARNING and never raise.
     """
     try:
-        from bridge.utc import to_unix_ts
+        from utils.utc import to_unix_ts
 
         unix_ts = to_unix_ts(event_ts)
         if unix_ts is None:
--- a/bridge/escape_hatch.py
+++ b/bridge/escape_hatch.py
@@ -9,7 +9,7 @@ not bypassed.
 
 import logging
 
-from bridge.utc import utc_iso
+from utils.utc import utc_iso
 
 logger = logging.getLogger(__name__)
 
--- a/bridge/session_transcript.py
+++ b/bridge/session_transcript.py
@@ -17,7 +17,7 @@ import logging
 import time
 from pathlib import Path
 
-from bridge.utc import utc_iso
+from utils.utc import utc_iso
 
 logger = logging.getLogger(__name__)
 
--- a/bridge/telegram_bridge.py
+++ b/bridge/telegram_bridge.py
@@ -31,9 +31,9 @@ from pathlib import Path
 
 from dotenv import load_dotenv
 
-from bridge.utc import to_unix_ts, utc_iso, utc_now
 from config.machine import get_machine_name
 from config.settings import settings
+from utils.utc import to_unix_ts, utc_iso, utc_now
 
 # Load environment variables FIRST before any env checks.
 # Under launchd (VALOR_LAUNCHD=1), env vars are injected directly into the plist
--- a/bridge/telegram_bridge.py
+++ b/bridge/telegram_bridge.py
@@ -152,7 +152,7 @@ def _pending_session_age_seconds(created_at, now_ts: float) -> float:
 
     ``AgentSession.created_at`` is a datetime (Popoto SortedField, naive on
     read); raw ``now_ts - created_at`` raises TypeError (#2458 D2). Coerce
-    through ``bridge.utc.to_unix_ts`` (naive treated as UTC). A missing or
+    through ``utils.utc.to_unix_ts`` (naive treated as UTC). A missing or
     uncoercible created_at returns +inf so the session falls outside any
     merge window instead of crashing the intake classifier.
     """
--- a/bridge/telegram_relay.py
+++ b/bridge/telegram_relay.py
@@ -365,8 +365,8 @@ def _record_sent_reaction(message: dict) -> None:
     every other outbound record. Best-effort: never raises into the relay.
     """
     try:
-        from bridge.utc import utc_now
         from tools.telegram_history import store_message
+        from utils.utc import utc_now
 
         chat_id = message.get("chat_id")
         reply_to = message.get("reply_to")
@@ -1175,7 +1175,7 @@ async def process_outbox(telegram_client) -> int:
                     if msg_type is None and msg_id is not None:
                         try:
                             from bridge.telegram_bridge import store_message
-                            from bridge.utc import utc_now
+                            from utils.utc import utc_now
 
                             await asyncio.to_thread(
                                 store_message,
--- a/docs/features/session-lifecycle.md
+++ b/docs/features/session-lifecycle.md
@@ -395,7 +395,7 @@ A stale caller can at worst append a spurious `session_events` entry. It cannot 
 
 ## Timestamp Convention — `updated_at` is Explicit UTC
 
-`AgentSession.updated_at` is always an explicit UTC wall-clock timestamp. It is stamped inside the `save()` override using `bridge.utc.utc_now()`, not by a Popoto `auto_now` field.
+`AgentSession.updated_at` is always an explicit UTC wall-clock timestamp. It is stamped inside the `save()` override using `utils.utc.utc_now()`, not by a Popoto `auto_now` field.
 
 **Why:** Popoto's `auto_now` calls `datetime.now()` (no `tz` argument), which mints a naive datetime in the host's local timezone. On non-UTC hosts the stored value is naive-local, but every downstream reader (watchdog, dashboard, stale-cleanup) interprets it as UTC. The result is a future-dated `updated_at` for sessions created on hosts running ahead of UTC, causing the watchdog/dashboard to report sessions as perpetually "fresh" and stale-cleanup to skip them forever.
 
--- a/docs/features/utc-timestamps.md
+++ b/docs/features/utc-timestamps.md
@@ -8,13 +8,15 @@ All timestamps in the system are stored and logged as **tz-aware UTC datetimes**
 - **Cross-machine consistency**: Deployed instances in different timezones produce identical timestamp formats.
 - **No naive datetimes**: Every `datetime` object in the system carries timezone info, preventing `TypeError` from mixed naive/aware comparisons.
 
-## The `bridge/utc` Module
+## The `utils/utc` Module
 
 Central utilities for timestamp handling:
 
 ```python
-from bridge.utc import utc_now, to_local, utc_iso, to_unix_ts
+from utils.utc import utc_now, to_local, utc_iso, to_unix_ts
 ```
+
+The module lives in `utils/` because it is dependency-free (stdlib `datetime` only) and standalone `tools/` packages import it, so it must not chain them to the harness (#2867).
 
 ### `utc_now() -> datetime`
 
@@ -57,7 +59,7 @@ Display surfaces fall into two categories:
 **Conversational output** (Telegram messages, UI relative times): convert to local time with `to_local()`:
 
 ```python
-from bridge.utc import utc_now, to_local
+from utils.utc import utc_now, to_local
 
 ts = utc_now()  # Store this
 display = to_local(ts).strftime("%H:%M")  # Show this to humans
@@ -82,7 +84,7 @@ The CLI uses `_format_ts()` in `tools/valor_session.py` (appends ` UTC` to all o
 - `time.time()` calls are unchanged -- epoch timestamps are timezone-neutral.
 - Telethon message timestamps were already UTC and are unchanged.
 - `AgentSession.updated_at` is explicitly UTC-stamped in `AgentSession.save()` (issue #1645 — Popoto's `auto_now=True` historically minted naive local wall-clock time). `created_at` and other `DatetimeField` timestamps are set explicitly at write time using `utc_now()`. As of **popoto >= 1.7.1** (issue #1653, popoto#421) `auto_now`/`auto_now_add` now stamp `datetime.now(timezone.utc)`, so the upstream producer bug is fixed and `auto_now` is safe to use. We nonetheless keep the explicit `utc_now()` stamp in `AgentSession.save()` for now: it also covers save paths that bypass `format_value_pre_save` (e.g. raw/pipeline updates). Removing the override is gated on confirming `auto_now` fires on every save path — until then, prefer explicit `utc_now()` stamps on `AgentSession`.
-- `SortedField` / `DatetimeField` deserialization can return naive datetimes — code that calls `.timestamp()` on values read from Popoto must treat them as UTC. **Use `bridge.utc.to_unix_ts(val)` for all read-path conversions.** It normalizes naive datetimes to UTC before `.timestamp()`, handles `None`/float/ISO-string inputs, and is the single source of truth for this coercion. Rewire sites include `scripts/update/run.py::_cleanup_stale_sessions`, `tools/agent_session_scheduler._to_ts`, `agent/sustainability`, `reflections/memory_management`, `tools/telegram_history._parse_ts`, and `models.agent_session.cleanup_expired`. Three older helpers (`monitoring/session_watchdog._to_timestamp`, `agent/session_health._to_ts`, `ui/data/sdlc._safe_float`) already implement the same `val.tzinfo is None` guard inline — they pre-date the shared helper and are intentionally left untouched. New code must import `to_unix_ts` rather than add a fourth copy (issues #777, hotfix 9e3a64f5).
+- `SortedField` / `DatetimeField` deserialization can return naive datetimes — code that calls `.timestamp()` on values read from Popoto must treat them as UTC. **Use `utils.utc.to_unix_ts(val)` for all read-path conversions.** It normalizes naive datetimes to UTC before `.timestamp()`, handles `None`/float/ISO-string inputs, and is the single source of truth for this coercion. Rewire sites include `scripts/update/run.py::_cleanup_stale_sessions`, `tools/agent_session_scheduler._to_ts`, `agent/sustainability`, `reflections/memory_management`, `tools/telegram_history._parse_ts`, and `models.agent_session.cleanup_expired`. Three older helpers (`monitoring/session_watchdog._to_timestamp`, `agent/session_health._ts`, `ui/data/sdlc._safe_float`) already implement the same `val.tzinfo is None` guard inline — they pre-date the shared helper and are intentionally left untouched. New code must import `to_unix_ts` rather than add a fourth copy (issues #777, hotfix 9e3a64f5).
 
 ## Related
 
--- a/models/agent_session.py
+++ b/models/agent_session.py
@@ -1019,7 +1019,7 @@ class AgentSession(Model):
 
         Popoto auto_now mints naive local time (bug #1645); instead we stamp
         explicitly so the stored value is always UTC wall-clock, consistent
-        with how created_at/started_at are handled (see bridge/utc.py::utc_now).
+        with how created_at/started_at are handled (see utils/utc.py::utc_now).
 
         update_fields guard: if update_fields omits 'updated_at', skip the stamp
         entirely (no in-memory mutation without a matching persist, to avoid
@@ -1132,7 +1132,7 @@ class AgentSession(Model):
                     continue  # None is safe — save() will stamp on next write
 
                 # Popoto strips tzinfo on load — treat naive datetimes as UTC
-                # (consistent with bridge/utc.py::to_unix_ts).
+                # (consistent with utils/utc.py::to_unix_ts).
                 updated_at_utc = record.updated_at
                 if updated_at_utc.tzinfo is None:
                     updated_at_utc = updated_at_utc.replace(tzinfo=UTC)
--- a/models/agent_session.py
+++ b/models/agent_session.py
@@ -1037,7 +1037,7 @@ class AgentSession(Model):
         the combination is a caller smell — a WARNING names the caller and no
         exception is raised (fail-quiet, matching every other guard here).
         """
-        from bridge.utc import utc_now
+        from utils.utc import utc_now
 
         if preserve_updated_at:
             if update_fields is not None and "updated_at" not in update_fields:
@@ -1116,7 +1116,7 @@ class AgentSession(Model):
         Returns the number of future-dated records detected (NOT healed —
         nothing is mutated or re-saved).
         """
-        from bridge.utc import utc_now
+        from utils.utc import utc_now
 
         now = utc_now()
         count = 0
@@ -2614,7 +2614,7 @@ class AgentSession(Model):
                 continue
             # Handle both datetime and float timestamps (migration period).
             # to_unix_ts treats naive datetimes as UTC (Popoto strips tzinfo).
-            from bridge.utc import to_unix_ts
+            from utils.utc import to_unix_ts
 
             ts = to_unix_ts(started)
             if ts is None:
--- a/models/job.py
+++ b/models/job.py
@@ -480,7 +480,7 @@ class Job(Model):
         """
         rested = 0
         try:
-            from bridge.utc import to_unix_ts
+            from utils.utc import to_unix_ts
 
             now_ts = now if now is not None else time.time()
             cutoff = now_ts - JOB_AT_REST_AGE_SECONDS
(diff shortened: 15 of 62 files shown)


tomcounsell:

No blockers. The move itself is clean, complete, and better evidenced than most changes this size — every number in the PR body that I re-measured reproduced exactly, and the completeness proof holds when widened past the gate the plan wrote for itself. One real gap remains, in the newly added guard rather than in the shipped code: the guard that defends `utils/__init__.py` misses the single most idiomatic way to add an import to an `__init__.py`, and I measured what that costs.

Every claim below was re-derived in an isolated worktree detached at `42e70e0ec`, not read off the PR body.

## Rubric

- [x] **1. Plan vs. implementation match** — pass — *All 7 issue acceptance criteria and all 13 plan Success Criteria delivered. Verified individually, not in aggregate.*
- [x] **2. New code quality** — pass — *The guard is well built: `assert scanned` and `assert os.path.isdir(package_dir)` both defend against the guard silently measuring nothing, and the failure messages name file, line, module, and the reason the constraint exists.*
- [x] **3. Test coverage** — pass — *`tests/unit/test_utc.py` still collects 20; the four mutation checks reproduce red exactly as claimed, including the `config` entry.*
- [x] **4. Regression risk to existing callers** — pass — *All 41 changed source modules import cleanly; all changed test files pass (350 + 73 + 29). The no-shim deletion makes a missed reference raise rather than silently degrade — confirmed: `import bridge.utc` now raises `ModuleNotFoundError`.*
- [x] **5. Data integrity** — n/a — *No schema, migration, or model change. `scripts/update/migrations.py` untouched.*
- [x] **6. Security** — pass — *stdlib `datetime` only. No secrets, no `eval`/`exec`/`subprocess`, no new dynamic import.*
- [x] **7. Documentation accuracy** — pass — *Docs and generated graph both repointed. The `agent/session_health._to_ts` → `_ts` correction is real: `grep -c '_to_ts' agent/session_health.py` is 0 and `def _ts` is at line 398, exactly as claimed. `tools/agent_session_scheduler._to_ts` is real (line 42) and correctly left alone.*
- [x] **8. PR body accuracy** — pass — *Independently reproduced: baseline 57 on `origin/main`; checkpoint gate readings recompute from history as 14 / 13 / 0 / 0 / 0; test_utc.py collects 20; targeted row 73 passed; non-unit row 29 passed / 4 skipped; graph.js 31 old path strings → 31 new.*
- [x] **9. Disclosed deferrals** — pass — *The four "Out of scope" items are explicit exclusions matching the plan's No-Gos, each with sound rationale. No tracking issue is claimed for any of them, and none is required.*
- [x] **10. Follow-up claims verified** — n/a — *The PR body makes no `filed as follow-up #N` claim.*

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — PASS — *13/13 checked individually.*
- **2. No-Gos from plan — none violated** — PASS — *The one assertable anti-criterion (`test -e bridge/utc.py` must exit non-zero) has a Verification row and passes. The dep-bump commit `0066b2117` is confirmed NOT an ancestor of the PR head — correctly dropped as out of scope.*
- **3. New `except Exception` blocks — each has logger/raise/swallow-ok** — N/A — *No exception handler added, removed, or modified. The one `except (ValueError, TypeError)` in `to_unix_ts` moved verbatim.*
- **4. New integration tests — exercise serialization boundary** — N/A — *No new integration test.*
- **5. Plan internal consistency — spike findings match task steps** — PASS — *spike-1's 214-vs-2 module measurement is what drives both the destination choice and the `config` entry in the guard; the guard's code comment carries that number.*
- **6. No hardcoded secrets or debug artifacts** — PASS — *None in the diff.*
- **7. New public APIs — docstrings present** — PASS — *No new API. The relocated module's docstring now states the dependency-free contract.*
- **8. Breaking changes — migration path documented** — PASS — *Internal path change only; deliberately no shim. Post-merge service restart is documented with the exact command and its expected log line.*
- **9. Tests added for new behavior** — PASS — *`TestStandaloneToolPackageBoundaries`, genuinely collected (2 passed, not pytest exit 5).*
- **10. Tests cover the failure path (not just happy path)** — PASS — *Verified by mutation, not by inspection. See below.*
- **11. UI changes (if any) — screenshot captured** — FAIL — *Disclosed, not counted against the change. See "Visual proof gate" below.*
- **12. Docs updated for user-facing changes** — PASS — *`docs/features/utc-timestamps.md`, `docs/features/session-lifecycle.md`, `site/assets/graph.js`.*

### Verified

- [x] Code correctness
- [x] Plan requirements met

### Blockers
- None

### Tech Debt
- None

### Nits
- None

### Miscellaneous

- **File:** `tests/unit/test_architectural_constraints.py:26` (verified: read this file)
  **Code:** `if isinstance(node, ast.ImportFrom) and node.module:`
  **Issue:** `_iter_imports` collects an `ast.ImportFrom` only when `node.module` is truthy. For `from . import X` the parser sets `module=None`, so that form is dropped before the `level` check ever runs and is invisible to `test_utils_package_init_imports_nothing`. Plan task 5 specifies this test should "assert `utils/__init__.py` parses to zero `Import`/`ImportFrom` nodes", and the test's own docstring says the file "must stay import-free, by contract rather than by luck" — neither holds for this form, which is the most idiomatic way to add an import to an `__init__.py`.

  I measured the consequence rather than asserting it. Injecting `from . import keyword_extraction` into `utils/__init__.py` leaves the guard **green** (`2 passed`), while `import utils.utc` goes from **4 modules / 1.1 ms** to **792 modules / 209 ms**, pulling in `config`, `config.loader`, `config.paths`, `config.settings` and pydantic. That is Risk 2 from the plan landing in full, and it is a larger regression than the 214-module `config` measurement that the destination decision and the guard's `config` entry were both built on.

  Scope is narrow and worth stating precisely: `import X`, `from X import y`, and `from .X import y` are all correctly caught. Only the bare `from . import X` escapes.
  **Severity:** tech_debt
  **Fix:** Collect the `module=None` case rather than dropping it — e.g. `imports.append((node.module or ".", node.lineno, node.level))` in the `ast.ImportFrom` branch. The existing `if level: continue` in `test_standalone_tool_packages_have_no_harness_imports` already keeps relative imports from producing false positives there, so the tools-package guard is unaffected. Then mutation-check it: `from . import keyword_extraction` in `utils/__init__.py` must go red.

- **Detaching `tools/telegram_history` (17 `models.*` imports) and `tools/image_gen` from `config.models`** — explicit out-of-scope exclusion, named as not-freed in issue #2867 itself. Sound.
- **Collapsing the three inline `to_unix_ts` duplicates** — explicit exclusion; #777 left them alone deliberately and `docs/features/utc-timestamps.md` records that decision. Sound.
- **Reorganizing `utils/` or introducing `kernel/`** — a decision made on measured evidence in the plan's Architectural Impact, not postponed work. Sound.
- **Regenerating `site/assets/graph.js` wholesale** — targeted path replacement instead. Verified equivalent for these nodes; see below.

### Verification Results

All 14 rows of the plan's `## Verification` table parse cleanly (0 malformed, 0 skipped). I ran 13 of them; the full-suite row was deliberately not re-run and was spot-checked instead.

| Row | Expected | Observed |
|---|---|---|
| Module exists at new path | exit 0 | exit 0 |
| Old module gone (anti-criterion: no shim) | exit != 0 | exit 1 |
| **Completeness gate over `*.py`** | `0` | **0** |
| Gate, fail-closed twin | exit 0 | exit 0 |
| Standalone tools carry no harness imports | `0` | **0** |
| Moved module imports no first-party package | exit 0 and `0` | exit 0, **0** |
| Targeted tests | exit 0 | **73 passed** |
| Guard class actually collected | exit 0, not 5 | **2 passed** |
| Modified non-unit tests execute | exit 0 | **29 passed, 4 skipped** |
| `test_utc.py` collected count | `20` | **20** |
| `ruff check .` | exit 0 | clean |
| `ruff format --check .` | exit 0 | 1364 files already formatted |
| Docs + generated graph gate | `0` | **0** |
| Full suite | exit 0 | not re-run — see below |

**The completeness proof is stronger than the gate that was written for it.** The plan's gate covers `*.py` plus two doc paths. I widened it to every tracked file of every extension: `git grep -lE 'bridge[./]utc|from bridge import utc'` across all **2660** tracked files returns **zero** hits outside `docs/plans/`, where the only matches are `completed/` history and this change's own plan document — both deliberately excluded. Control-checked that the sweep genuinely reaches the places a stale reference would hide: 215 tracked `.claude/` files and 1296 tracked non-`.py` files are in scope, and the same sweep does find `utils.utc` in 60 files. So there is no surviving reference in a hook, skill, YAML, TOML, or JS file that the narrower gate would have missed. Both `__import__` sites now read `"utils.utc"`.

**The module body is genuinely unchanged.** `diff` of `git show origin/main:bridge/utc.py` against `utils/utc.py` reports only docstring lines 7-8. All four function bodies and signatures are byte-identical, and `git diff --find-renames` reads it as `{bridge => utils}/utc.py` at 90% similarity.

**Mutation checks reproduced, not taken on trust.** All four claimed guard mutations go red, each naming the offending file and line:

| Mutation | Result |
|---|---|
| harness import → `tools/selfie/__init__.py` | **red** |
| `from config.paths import ...` → `tools/sms_reader/__init__.py` | **red** (the added `config` entry is genuinely wired in) |
| harness import → `tools/test_scheduler/tests/test_scheduler.py` | **red** (the walk does recurse into the packages' own `tests/`) |
| import → `utils/__init__.py` | **red** |
| `from . import X` → `utils/__init__.py` *(added by this review)* | **green** ← the Miscellaneous finding |

**On the full suite.** Not re-run, per the reviewing brief. Spot-checked the build's "no overlap with changed files" classification directly rather than by re-deriving `comm -12`: every one of the 16 changed test files passes here (350 + 73 + 29 = 452 tests), and all 41 changed source modules import cleanly with `bridge.utc` absent from `sys.modules`. A regression from this change would have to appear in a file this change touched, and none does.

One caution for anyone re-running the non-unit row in a fresh worktree: `config/reflections.yaml` is gitignored (`.gitignore:8`) and therefore absent from a clean checkout, which fails two tests in `tests/integration/test_reflections_redis.py` with `FileNotFoundError`. That is a worktree artifact, not a branch defect — supplying the file makes the row read exactly `29 passed, 4 skipped`. I ran that control rather than assuming it.

**Generated graph.** Verified structurally instead of by eye: `site/assets/graph.js` parses to 1312 nodes and 2114 edges with **zero dangling edge endpoints**, no residual `bridge/utc` node id, and the utc nodes now read `file:utils/utc.py`, `function:utils/utc.py:to_local`, `function:utils/utc.py:to_unix_ts`. Every one of the 62 changed lines in that file contains `utc.py` — no unrelated churn rode along.

### Visual proof gate

`site/assets/graph.js` matches the UI-file detector (`/assets/`), so the gate nominally applies, and no screenshot was captured: the BYOB bridge is not running on this machine, so there is no browser surface available. I am disclosing that rather than converting it into a blocker, because the gate's premise — that visual defects are invisible to static analysis — does not hold for this change. Nothing here is markup, styling, or rendering logic; it is 31 string replacements in a generated data blob, and its one real failure mode (an edge endpoint no longer resolving to a node id after the rename) is exactly what the structural check above measured at zero across 2114 edges. If pixel confirmation of the docs-site graph page is wanted, the BYOB bridge needs to be running first.

### Reviewer's note on process

The repo's declared two-judge consensus (`code-quality`, `risk`) was dispatched, but neither judge returned within ~29 minutes and they could not be stopped from this context. This verdict is therefore single-judge, from the evidence above. I independently covered the risk judge's highest-value question — dynamic and non-`.py` references that the completeness gate cannot see — in the repo-wide sweep. If either judge reports later, its findings should be read against this review rather than treated as superseded.

### Post-merge

Five `bridge/` modules and six `agent/` modules changed import paths, so the running bridge, watchdog, and worker are on stale code until cycled. The PR body already carries the correct sequence (`./scripts/valor-service.sh restart`, then confirm "Connected to Telegram" in `logs/bridge.log`) and correctly sequences it after the merge rather than during the build.

valorengels:
## Review: PR #2900 — move the UTC time helper out of `bridge/` into `utils/` (issue #2867)

**Verdict: APPROVED** — 0 blockers, 0 tech-debt findings.

### What was verified

The change is a true rename, not a delete-plus-add: `git diff --find-renames main...HEAD -- bridge/utc.py utils/utc.py` reports `{bridge => utils}/utc.py` at *similarity index 90%*, and the only textual difference is the module docstring. All four functions (`utc_now`, `to_local`, `utc_iso`, `to_unix_ts`) are byte-identical in body and signature. The added docstring paragraph states the contract that makes the move worth anything ("dependency-free by contract: it imports nothing outside the standard library"), and the new guard test enforces it rather than leaving it as prose.

Confirmed against a clean checkout of the PR head (`2acffcd8446910efb06f99861a0cf54e300d5c03`) with a synced 3.14 venv:

- `utils/utc.py` exists and exports all four functions; `from utils.utc import utc_now, to_local, utc_iso, to_unix_ts` resolves and `utc_now().tzinfo` is UTC.
- `bridge/utc.py` is deleted outright — `import bridge.utc` raises `ModuleNotFoundError`. No shim, no re-export, no deprecation window. This is the right call: it converts a missed `patch("bridge.utc...")` from a silent green into a loud failure.
- Completeness gate: `git grep -lE 'bridge[./]utc|from bridge import utc' -- '*.py' | wc -l` returns **0** on the head commit (57 at baseline). The character class covers both the dotted and the path spelling, which a dotted-only rewrite rule would have missed.

### The part that actually needed a reviewer

An import-rewrite diff cannot surface string-literal references, and all three of this repo's known traps are handled correctly:

1. `tests/unit/test_session_stall_classifier.py:296-300` — the load-bearing `patch(...to_unix_ts)`. Repointed to `utils.utc.to_unix_ts`, and it genuinely bites: `agent/session_stall_classifier.py:233` does a *function-local* `from utils.utc import to_unix_ts`, so the patch target is resolved at call time and the mock takes effect rather than degrading into a no-op that reports green.
2. `tests/integration/test_reflections_redis.py:107,129` — `__import__("utils.utc", fromlist=["utc_now"])`, both repointed.
3. `tests/unit/test_public_api_contract.py:57` — pin updated to `("utils.utc", "utc_now")`.

### Decoupling payoff is real, not claimed

`tools/selfie/__init__.py:15`, `tools/sms_reader/__init__.py:34`, and `tools/test_scheduler/__init__.py:15` each had exactly one cross-package import in their whole source tree, and it was the `bridge.utc` one. All three now import `utils.utc` as their only cross-package import. `tools/image_gen/__init__.py:27` is left with only `config.models`, which the issue explicitly scopes out.

### Added scope — reviewed and accepted knowingly

The PR flags two additions beyond the issue's acceptance criteria. Both are approved deliberately:

- **`TestStandaloneToolPackageBoundaries`** (in `tests/unit/test_architectural_constraints.py`). The issue asserts a state; it does not ask for a test pinning that state. But `utils/__init__.py` is empty by accident, not by contract, and one `from models import ...` added there later would silently re-couple all three freed packages with nothing failing. Encoding the invariant is worth the test.
- **`config` in the guard's forbidden set**, beyond the issue's `bridge/agent/worker/models/monitoring/reflections` list. This is justified on measured evidence rather than symmetry: importing any `config` submodule executes `config/__init__.py`, which pulls 214 modules in 94 ms against `utils`'s 2 in 1 ms — the same measurement that chose `utils/` over `config/` as the destination. A guard that forbade `from models import ...` while permitting `from config.paths import ...` would wave through exactly the coupling this change exists to remove. It costs no allowlist entry, since all three packages are already clean.

### Test and gate results at head

| Check | Result |
|---|---|
| `tests/unit/test_public_api_contract.py` + `tests/unit/test_session_stall_classifier.py` | **45 passed** (381s, via `scripts/pytest-clean.sh`) |
| `tests/unit/test_architectural_constraints.py` (the new guard) | **8 passed** (414s, via `scripts/pytest-clean.sh`) |
| `python -m ruff check .` | exit 0 |
| `python -m ruff format --check .` | exit 0, 1364 files already formatted |
| `uv lock --locked` | PASS |
| PR mergeability | `MERGEABLE` / `CLEAN` |

Documentation was cascaded in the same change: `docs/features/utc-timestamps.md:11,16,62,87` and `docs/features/session-lifecycle.md:398`. Remaining `bridge/utc` mentions are confined to `docs/plans/move-bridge-utc-to-utils.md` (the plan doc describing the before-state, which is correct) and `docs/plans/completed/`.

### Non-blocker, recorded not fixed

`site/assets/graph.js` embeds `bridge/utc.py` node ids. It is a generated artifact with no in-repo generator, and `tests/unit/test_site_graph_consistency.py:3-5` explicitly tolerates phantom nodes for deleted files. Not a merge blocker; it will correct itself whenever the graph is next regenerated.

### Post-merge note carried forward

Five `bridge/` modules changed import paths, so the running bridge/worker are on stale code until cycled. The service restart is being deferred deliberately: several sibling SDLC lanes are in flight concurrently, and restarting underneath them produces false "worker running old sha" alerts. The restart runs once after the batch drains.

PR #2941, opened 2026-08-24
test: split test_output_handler and test_memory_extraction by theme (Refs #2879)

### LINKED ISSUE #2879: Split the largest test files into per-class modules
> **Context:** This repo's `tests/unit/` directory contains several very large single-file test modules that mirror the size of the production monoliths they test.

## Problem

**Current behavior:**
- `tests/unit/test_output_handler.py` — 3,253 lines, 108 tests, 14 classes.
- `tests/unit/test_sdlc_session_ensure.py` — 3,171 lines, 114 tests, 16 classes.
- `tests/unit/test_memory_extraction.py` — 2,638 lines, 136 tests, 15 classes.
- `tests/unit/test_sdlc_router_decision.py` — 2,096 lines, 107 tests.
- `tests/unit/test_valor_telegram.py` — 2,016 lines, 97 tests.
- `tests/unit/test_worktree_manager.py` — 1,610 lines, 99 tests.
- 20 test files in total have more than 80 test functions each.
- These monolithic files make it harder to localize a specific failure, and this repo's test runner (`scripts/pytest-clean.sh`, using `pytest-xdist`) parallelizes at the file level via `--dist=loadfile`, so oversized files also limit parallelism.

**Desired outcome:**
The largest test files are split into per-class modules of manageable size, with test collection and behavior unchanged.

## Solution Sketch

- Convert each of the files above into a package directory (e.g. `tests/unit/output_handler/`) with one file per existing test class.
- Keep test bodies unchanged; only move imports and shared fixtures.
- Verify `pytest-xdist --dist=loadfile` (see the `pyproject.toml` pytest config) still groups related tests safely after the split, and update the comment there if the grouping assumption changes.

## Recon: Skipped

Compiled from a pre-existing line/test/class count audit of `tests/unit/`, verified against the current tree before filing. Purely mechanical — `/do-plan` should confirm the current test-count baseline hasn't shifted since filing before treating "test collection count is unchanged" as verified.

## Acceptance Criteria

- [ ] Each of the 6 files listed above is under roughly 800 lines after the split.
- [ ] Total test collection count (`pytest --collect-only`) is unchanged before and after.
- [ ] Targeted unit tests pass under `scripts/pytest-clean.sh` (never bare `pytest` — see `CLAUDE.md`).

## Downstream

This issue feeds into `/do-plan`, which will produce a plan document. Constraint: this is a mechanical reorganization — no test logic changes, no new fixtures beyond what's needed to preserve current behavior.

### LINKED ISSUE #2946: Split test_output_handler.py and test_memory_extraction.py into theme-grouped packages
> **Context:** `tests/unit/` in this repo contains several very large single-file test modules. Issue #2879 is the umbrella tracking six of them. This issue is the scoped child covering exactly **two** of those files, so that the work can land and close a real issue while the umbrella stays open for the remaining four. Part of #2879.

## Problem

**Current behavior:**

| File | Lines | Test classes | Notes |
|---|---|---|---|
| `tests/unit/test_output_handler.py` | 3,253 | 15 | — |
| `tests/unit/test_memory_extraction.py` | 2,638 | 19 | plus 2 orphan module-level test functions |

Monolithic test files make it harder to localize a failure, and this repo's runner (`scripts/pytest-clean.sh`, `pytest-xdist` with `--dist=loadfile`) parallelizes at the **file** level, so oversized files cap parallelism.

**Desired outcome:**

Both files become theme-grouped packages, with test behavior and marker assignment provably unchanged.

- `tests/unit/test_output_handler.py` → `tests/unit/output_handler/` (5 theme-grouped files)
- `tests/unit/test_memory_extraction.py` → `tests/unit/memory_extraction/` (5 theme-grouped files)

## Scope

Exactly the two files above. The other four files named in #2879 are explicitly **out of scope** here and remain the umbrella's business.

## Solution Sketch

Group by **theme**, not strictly one-file-per-class. Strict per-class splitting would produce a spray of ~40-line modules (15 and 19 classes respectively across ~5,900 lines), which trades one navigability problem for another. The target is the umbrella's rough ~800-line ceiling per resulting file, which lands at about 5 files per package.

Hard constraint: **no test logic may change.** Test bodies move byte-for-byte. Only imports, shared fixtures, and the new `__init__.py` / package scaffolding are new text.

New basenames should keep the **original basename as a prefix** (e.g. `test_output_handler_<theme>.py`, `test_memory_extraction_<theme>.py`) so that any future `FEATURE_MAP` entry written against the original name still matches every shard.

## Definitions

| Term | Meaning |
|---|---|
| `FEATURE_MAP` | Dict in `tests/conftest.py` (around line 1026) mapping a substring pattern → a pytest marker name (e.g. `"routing"` → `messaging`). |
| Feature marker | A pytest marker auto-applied at collection time so a subset of the suite can be run with `pytest -m messaging`. Not written in the test files; derived. |
| Theme group | A set of test classes in one file that exercise a coherent slice of behavior, sized to the ~800-line ceiling. |

## Why the umbrella's acceptance criterion is insufficient

#2879 states its AC as "total test collection count (`pytest --collect-only`) is unchanged." **That check cannot detect the failure mode this split is most exposed to,** and this issue must strengthen it.

`tests/conftest.py::pytest_collection_modifyitems` (around line 1126) derives feature markers like this:

```python
filename = item.nodeid.split("::")[0].split("/")[-1].replace("test_", "").replace(".py", "")
for pattern, marker_name in FEATURE_MAP.items():
    if pattern in filename:
        item.add_marker(getattr(pytest.mark, marker_name))
        break
```

It takes the **last path segment** of the nodeid — the basename — strips `test_` and `.py`, and substring-matches the remainder against `FEATURE_MAP`. The directory contributes nothing.

Consequences:

1. **Moving** a file into a new directory is marker-neutral.
2. **Renaming** a file silently re-tags it. Both directions are real failures:
   - **Losing** a marker: a shard whose new basename no longer contains a `FEATURE_MAP` pattern drops out of `pytest -m <marker>` runs, so a targeted CI or dev invocation silently stops covering it.
   - **Gaining** a marker: a plausible shard name like `test_output_handler_transport_aware_routing.py` contains `routing`, which is a `FEATURE_MAP` key, and would be wrongly tagged `messaging`.
3. The **total collected count is identical under both**. Nothing about a marker changes how many items are collected. So the umbrella's AC passes cleanly on a broken split.

## Acceptance Criteria

- [ ] `tests/unit/test_output_handler.py` is replaced by `tests/unit/output_handler/` containing 5 theme-grouped test files.
- [ ] `tests/unit/test_memory_extraction.py` is replaced by `tests/unit/memory_extraction/` containing 5 theme-grouped test files. The 2 orphan module-level test functions are placed in a themed shard, not dropped.
- [ ] Each resulting file is at or under roughly 800 lines.
- [ ] Every new basename carries the original basename as a prefix.
- [ ] **No test logic changed** — test bodies moved byte-for-byte. Verifiable by diffing the sorted concatenation of test bodies before and after, or by review of the diff showing only moves plus import/fixture scaffolding.
- [ ] **Per-marker histogram parity**: capture a count of collected items *per feature marker* across all of `tests/unit`, before and after the split. **Zero drift on every marker** — no marker gains items, no marker loses items. A bare total is not sufficient (see the section above).
- [ ] **Identical test-id multiset**: the sorted multiset of collected test ids, with the changed path segments normalized away, is identical before and after. This catches a dropped or duplicated test that a matching total could mask.
- [ ] Targeted unit tests pass under `scripts/pytest-clean.sh` — never bare `pytest` (see `CLAUDE.md`).

A workable shape for the histogram baseline (capture on `main` first, then on the branch, then diff):

```bash
for m in messaging tools sdlc sessions; do
  printf '%s %s\n' "$m" "$(scripts/pytest-clean.sh tests/unit -m "$m" --collect-only -q 2>/dev/null | grep -c '::')"
done
```

Enumerate the marker list from the distinct values in `FEATURE_MAP` rather than hardcoding it, so a future map entry is covered automatically.

## Downstream

This issue feeds `/do-plan`, which will produce a plan document at `docs/plans/{slug}.md`. Constraint restated for the planner: this is a mechanical reorganization. No test logic changes, no new fixtures beyond what is required to preserve current behavior, and no scope expansion to the other four files in #2879.

valorengels:
Closes #2946
Refs #2879

**Reduced scope**: 2 of the 6 files named in umbrella issue #2879. The other four are deferred with reasons below, so this deliberately does **not** close #2879 — it closes the scoped child issue #2946, which covers exactly these two files. #2879 stays open for the remaining four.

## What changed

| Before | After |
|---|---|
| `tests/unit/test_output_handler.py` (3,253 lines, 15 classes) | `tests/unit/output_handler/` — 5 files, 558–774 lines |
| `tests/unit/test_memory_extraction.py` (2,638 lines, 19 classes + 2 orphan module-level tests) | `tests/unit/memory_extraction/` — 5 files, 290–674 lines |

Grouped by theme rather than strictly one-file-per-class, because per-class would have produced a spray of 40-line modules. Every file is under the issue's ~800-line target.

- `output_handler/`: `_handlers` (protocol, file, telegram-relay, system-room sink), `_drafter`, `_transport` (react-transport derivation, transport-aware routing, drafter-above-transport), `_filters` (read-the-room, redundancy), `_delivery` (deferred self-draft, DeliveryOutcome, deliver_system_notice).
- `memory_extraction/`: `_post_session`, `_outcome_detection`, `_parsing`, `_refusal_filters`, `_event_loop_safety`.

The two orphan module-level test functions were homed by subject: `test_extract_post_merge_learning_runs_inside_asyncio_run` → `_event_loop_safety`, `test_session_cap_default_within_audit_threshold` → `_post_session`.

**No test body was changed.** Enforced mechanically, not by eyeball: a verification script parsed the pre-split file out of git and each new file with `ast`, and compared `ast.get_source_segment` for all 36 top-level definitions. All 36 are byte-identical. The only edits are the per-file module docstring and the trimmed import header (unused imports removed with `ruff --select F401`, which never touched a body).

## The hazard this PR is really about: markers are derived from the basename

`tests/conftest.py::pytest_collection_modifyitems` takes the nodeid's **last path segment**, strips `test_` and `.py`, and substring-matches the remainder against `FEATURE_MAP`. The directory contributes nothing. So moving a file into a subpackage is marker-neutral, but *renaming* it silently re-tags it — in both directions:

- **Loss:** `tests/unit/output_handler/test_drafter_failure_recovery.py` yields basename `drafter_failure_recovery`, which matches nothing.
- **Gain:** `..._transport_aware_routing` matches the `routing` pattern and would be wrongly tagged `messaging`.

The issue's acceptance criterion — `--collect-only` total count unchanged — **cannot catch either**, because the total is unchanged and only the tagging moves. Existing subpackages in this repo have already drifted this way.

Mitigation here: every new file keeps the original basename as a prefix (`test_output_handler_*`, `test_memory_extraction_*`), which preserves any matching substring by construction, and each new basename was checked against `FEATURE_MAP` for an accidental new match.

## Verification

Marker parity was measured, not assumed. A collect-only pass over all of `tests/unit` dumped every nodeid with its markers, before and after, via a throwaway `trylast` plugin.

```
per-marker drift: NONE            # all 23 markers, identical counts
test-id multiset identical: True  # every Class::test id preserved
totals: 13989 -> 13989
```

Both target files carry **no** `FEATURE_MAP` marker today (`output_handler` and `memory_extraction` match no pattern), and carry none after — the marker sets observed on the split files are only pytest's own `asyncio`/`parametrize`. So the honest framing is: there was no marker to lose here, and none was accidentally gained. The full 23-marker histogram is the evidence for the second half of that claim, which is the part a total-count check would miss.

Per-marker counts before → after (unchanged): `asyncio` 1167, `classifiers` 168, `config` 55, `context` 71, `git` 179, `impact` 75, `integration` 2, `messaging` 990, `models` 64, `monitoring` 152, `parametrize` 1894, `reflections` 427, `sdk` 43, `sdlc` 2577, `sessions` 918, `skipif` 60, `slow` 10, `tools` 49, `unit` 1137, `usefixtures` 39, `validation` 277, `webui` 310, `xdist_group` 3.

Tests (`scripts/pytest-clean.sh`, never bare pytest):

```
tests/unit/output_handler tests/unit/memory_extraction   ->  256 passed in 167.77s
  (baseline: 108 + 148 = 256 collected pre-split)

+ the three sibling modules whose docstrings were repointed
  (test_send_message, test_message_drafter, test_read_the_room)
                                                         ->  436 passed in 296.07s
```

`python -m ruff check` — All checks passed. `python -m ruff format --check` — 1380 files already formatted.

## No parallelism claim

The issue motivates the split partly by `--dist=loadfile` parallelism. **This PR does not claim a speedup and no timing baseline was taken.** These two files are ~6k of ~284k lines in the suite; wall time is bounded by out-of-scope files (`tests/integration/test_steering.py` at 2,487 lines, `test_docs_auditor_substrate.py` at 2,124). Any measured difference here would be noise.

Relatedly, the `--dist=loadfile` comment at `pyproject.toml:163-168` was deliberately **not** touched. It names three integration files that require loadfile grouping; neither split file is among them, so the comment remains accurate and editing it would have been a manufactured change. Instead, `tests/README.md` gains a new **"Splitting or Renaming a Test File"** section documenting the basename→`FEATURE_MAP` coupling and the verification procedure, which is the genuinely missing piece of documentation.

## Why the other four are deferred

Each carries shared state or a marker cost that makes it a judgment call rather than the mechanical reorganization the issue mandates. These two were picked precisely because they have essentially zero module-level shared state (imports only; each class carries its own `_make_handler`/`_mock_redis`).

- **`test_sdlc_router_decision.py`** — per-class yields 37 files averaging 56 lines, 13 of them under 20. Per-class is the wrong granularity; it needs a theme-grouping judgment call.
- **`test_valor_telegram.py`** — an autouse `_bypass_promise_gate` (L17) to hoist, `_CandidateStub` (L482) to relocate, a `sys.path.insert` whose depth breaks on a move (L12), and it would lose the `messaging` marker.
- **`test_worktree_manager.py`** — 6 shared helpers, a `_git` name collision with `tests/unit/conftest.py`, `parents[2]` depth breakage at L1268, and it would lose the `git` marker.
- **`test_sdlc_session_ensure.py`** — highest risk. A class-name-conditional autouse fixture `_stub_lane_identity` (L33) whose `_LANE_IDENTITY_CLASS` check `TestIssueLockWiring`'s call-count assertions depend on; real `PipelineLedger` Redis writes against a fixed key; and `REPO_ROOT` (L28) driving real subprocess launches.

## Other files touched

Stale test-path references repointed (comments and docstrings only, no assertions): `tests/conftest.py` (the context-recall billing note), `tests/unit/test_send_message.py`, `tests/unit/test_message_drafter.py`, `tests/unit/test_read_the_room.py`, and `docs/features/{agent-message-delivery,read-the-room,message-drafter,drafter-redundancy-suppression}.md`. The `tests/README.md` index row for `test_memory_extraction.py` was replaced with the five new rows and real collected counts. Historical `docs/plans/` entries were left alone as archives.


valorengels:
## Review: PR #2941 — split test_output_handler and test_memory_extraction by theme

**Verdict: APPROVED**

Reviewed at head SHA `925026c51819cf1f74e3309ede91115e7af4825f` (resolved git-first via `tools/pr_head_resolver.py::resolve_pr_head_sha`, not a `gh` read).

Scope is tests and docs only. Zero production code changed.

### The core promise, re-verified independently

The PR's central claim is that no test logic changed. This is the claim that matters, because a blanket `ruff --fix` silently rewrote four test bodies during the build (stripping local re-imports as F811 once the module-level import was in scope) before the builder caught and reverted it. I did not take the builder's word for the fix.

I parsed the pre-split files out of `origin/main` and every new file out of the branch with `ast`, then compared `ast.get_source_segment` for every top-level definition:

| Source file | Old defs | New defs | Missing | Extra | Duplicate names | Body mismatches |
|---|---|---|---|---|---|---|
| `test_output_handler.py` | 15 | 15 | 0 | 0 | 0 | **0** |
| `test_memory_extraction.py` | 21 | 21 | 0 | 0 | 0 | **0** |

All 36 top-level defs are byte-identical. The only edits are per-file module docstrings and trimmed import headers.

### Marker parity

`tests/conftest.py::pytest_collection_modifyitems` derives markers from the nodeid's **basename** (strip `test_`/`.py`, substring-match against `FEATURE_MAP`), so the directory contributes nothing and a rename silently re-tags a file while leaving the collected total unchanged. The issue's stated `--collect-only` total-count AC cannot detect this in either direction. The builder went past that AC correctly.

I checked all 10 new basenames against `FEATURE_MAP` (93 patterns) programmatically: **none matches any pattern**. Both original basenames also matched nothing, so the honest framing is that there was no marker to lose here, and the real risk was the inverse. Confirmed the near-miss: the natural name `test_output_handler_transport_aware_routing.py` **would** have matched `routing` and been wrongly tagged `messaging`. Keeping the original basename as a prefix on every file is the right mitigation, since it preserves any future `FEATURE_MAP` match by construction.

### Shared-state and collection checks

- **No module-level fixtures** in either pre-split file, so there are no orphaned fixtures and no silent fixture-scope change from moving classes into a package. This substantiates the builder's rationale for picking these two files.
- Both new packages carry `__init__.py`, and all 10 new basenames are unique across `tests/`. No import-collision risk. (The duplicate basenames that do exist in `tests/` are pre-existing unit/integration pairs, untouched here.)
- Collection: **256 collected**, exactly matching the pre-split baseline of 108 + 148.
- Runtime: **256 passed** via `scripts/pytest-clean.sh ... -n 4`.
- `ruff check` clean, `ruff format --check` clean (1382 files), `uv lock --locked` passes.

### Scope judgments

1. **Theme-grouping instead of the issue's literal "per-class modules" — accepted.** Per-class would have produced a spray of ~40-line modules, which serves the letter of the issue and not its purpose. Files land at 289-774 lines, under the issue's ~800 target. The deviation is justified in the PR body rather than silently taken, which is the right way to handle it.
2. **Both orphan module-level test functions are homed sensibly.** `test_extract_post_merge_learning_runs_inside_asyncio_run` sits with `TestEventLoopSafety` in `_event_loop_safety`; `test_session_cap_default_within_audit_threshold` sits with `TestRunPostSessionExtraction` in `_post_session`. Both group by subject rather than by leftover.
3. **The `tests/README.md` "Splitting or Renaming a Test File" section belongs.** It is scope beyond the issue, but it is the durable fix for the trap this PR spent its effort on, and a mechanical split that leaves the hazard undocumented invites the next split to fall into it. Checked its technical accuracy against `pytest_collection_modifyitems` and `FEATURE_MAP`: the basename derivation, the first-hit break, and both worked examples are correct.
4. **Deferral reasons for the other four files all verify against the real files**, not just asserted: `test_valor_telegram.py` (autouse fixture L17, `_CandidateStub` L482, depth-sensitive `sys.path.insert` L12), `test_worktree_manager.py` (`parents[2]` L1268, `_git` collision with `tests/unit/conftest.py:10`), `test_sdlc_session_ensure.py` (`REPO_ROOT` L28, class-name-conditional autouse fixture L33/L44 whose stub branch `TestIssueLockWiring`'s call-count assertions depend on, per its own docstring), `test_sdlc_router_decision.py` (37 classes, confirmed).
5. **`pyproject.toml:163-168` correctly left alone.** Its `--dist=loadfile` comment names three integration files; neither split file is among them, so the comment remains accurate and editing it would have been a manufactured change.
6. **No parallelism claim is made, correctly.** No timing baseline exists and these two files are ~6k of ~284k lines. The issue asserts a parallelism win without evidence; the PR declines to repeat it.

### Concern (non-blocking, do not churn the head SHA for this)

Every new module docstring opens with "Split out of the former ``tests/unit/test_output_handler.py`` monolith (#2879)." CLAUDE.md Development Principle 1 says to describe only the new status quo and keep historical artifacts out of docs. A strict reading makes these provenance lines exactly that.

I rule it **acceptable**: the line orients a reader who is looking for the other half of a familiar test file, and it will stop being useful only once nobody remembers the monolith, at which point it is one line to delete. Flagging it so the call is on the record rather than overlooked. Not worth a new commit, which would restale this verdict.

### Merge disposition

`Closes #2946` (the scoped phase issue covering exactly these two files) and `Refs #2879`. Umbrella **#2879 must stay open** after merge, since only 2 of its 6 files are split. Verified exactly one `PipelineLedger` row resolves for PR 2941 and it points at #2879, so the substrate is unambiguous.

No `Co-Authored-By` or generated-with footer on the single commit. Author is the operator.

Issue #2875, opened 2026-08-19
Migrate config/reflections.yaml off agent.sustainability.* and delete the shim


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/retire-agent-sustainability-shim.md

> **Context:** The [reflections system](docs/features/README.md) runs scheduled maintenance and health-check agents defined in `config/reflections.yaml`, each pointing at a Python `callable:`. [`agent/sustainability.py`](agent/sustainability.py) exists almost entirely as a compatibility surface so that registry can keep using historical dotted paths (e.g. `agent.sustainability.circuit_health_gate`) after the real implementations moved to `reflections/agents/*.py`. The shim's own docstring already says "new code should import the reflection directly."

## Problem

**Current behavior:**
- `config/reflections.yaml` still has 5 reflection `callable:` entries pointing at `agent.sustainability.*`.
- `agent/agent_session_queue.py` imports `send_hibernation_notification` from the shim.
- `reflections/stall_advisory.py` imports `_get_project_key` / `_get_redis` from the shim.
- The shim adds an indirection layer between the registry and the real implementations in `reflections/agents/*.py` for no remaining benefit.

**Desired outcome:**
`config/reflections.yaml` and all production callers reference `reflections/agents/*.py` directly. `agent/sustainability.py` is deleted.

## Solution Sketch

- Update the 5 `callable:` entries in the repo's `config/reflections.yaml` to `reflections.agents.<module>.run`.
- Ensure the vault copy (`~/Desktop/Valor/reflections.yaml`) and its install-time materialization path are updated consistently — these must not drift out of sync with the repo copy.
- Move `_get_project_key` / `_get_redis` to a small canonical helper module (or reuse an existing one) and update `reflections/stall_advisory.py` plus its tests.
- Update `agent/agent_session_queue.py` to import `send_hibernation_notification` from `reflections.agents.circuit_health_gate`.
- Delete `agent/sustainability.py`.
- Update `tests/unit/test_sustainability.py`, `tests/unit/test_sustainability_namespace.py`, and any other tests importing from the shim.

## Recon: Skipped

Compiled from a pre-existing audit of the shim's callers, verified against the current tree before filing. `/do-plan` should re-confirm the reflections registry resolves cleanly against the new paths (dry-run or targeted scheduler test) before considering this done.

## Acceptance Criteria

- [ ] `git grep -n "agent.sustainability"` returns no hits in production code.
[...]

PR #2943, opened 2026-08-24
refactor: delete bridge/session_logs.py and models/reflections.py shims

### LINKED ISSUE #2872: Delete the bridge/session_logs.py and models/reflections.py compatibility shims
> **Context:** This repo has accumulated several import-path compatibility shims left over from earlier refactors — thin modules that only re-export symbols from their new home so old `import` statements keep working. Two of them are pure dead weight: [`bridge/session_logs.py`](bridge/session_logs.py) and [`models/reflections.py`](models/reflections.py) (the latter re-exports `ReflectionIgnore` / `PRReviewAudit`).

## Problem

**Current behavior:**
- `bridge/session_logs.py` is imported by exactly one test (`tests/unit/test_worker_entry.py:281`) and zero production code.
- `models/reflections.py` is imported by only 3 production modules: `reflections/housekeeping/redis_ttl_cleanup.py`, `reflections/audits/pr_review_audit.py`, `reflections/utilities.py`.
- Both shims exist solely to preserve import paths after earlier refactors moved their real implementations elsewhere.

**Desired outcome:**
Both shim files are deleted. All production and test code imports directly from the canonical modules (`models.reflection_ignore`, `models.pr_review_audit`, and whatever `session_logs` functionality's true home is).

## Solution Sketch

- Update the 3 production importers of `models/reflections.py` to import from `models.reflection_ignore` / `models.pr_review_audit`.
- Update the one test that imports `bridge/session_logs.py`, and any tests that `patch("models.reflections...")`.
- Update docs that still describe `bridge/session_logs.py` as a live path: `docs/features/bridge-workflow-gaps.md`, `docs/features/worker-service.md`, `docs/features/correlation-ids.md`, and the doc-auditor test expectation that references it.
- Delete both shim files.

## Recon: Skipped

This issue was compiled from a pre-existing repo-wide audit of compatibility shims and dead-code surface area (import-site counts and docstring claims verified against the current tree before filing). Full independent reconnaissance is skipped as low-value given the small, mechanical scope; `/do-plan` should re-verify import sites at plan time since the tree may have moved since filing.

## Acceptance Criteria

- [ ] `git grep -n "models.reflections\|bridge.session_logs"` returns only historical references under `docs/plans/completed/`, with no hits in production code.
- [ ] Both shim files are deleted from the repo.
- [ ] Targeted tests under `tests/unit/` and `tests/integration/` pass after the migration.

## Downstream

This issue feeds into `/do-plan`, which will produce a plan document. The plan is then executed by `/do-build`. Constraint: this is a pure import-path migration — no behavior change, no new abstractions.

valorengels:
Closes #2872

Deletes two import-path compatibility shims left over from earlier refactors. Pure import-path migration: no behavior change, no new abstractions.

## What changed

**`bridge/session_logs.py`** (13 lines) — a pure re-export of `agent/session_logs.py` (the 182-line canonical module). Zero production importers; its only consumer was `test_session_logs_re_exports_from_bridge`, a test that existed solely to assert the re-export. Both deleted. The sibling test `test_session_logs_importable_from_agent` keeps covering the canonical path.

**`models/reflections.py`** — re-exported `ReflectionIgnore` / `PRReviewAudit`. The three production importers now import canonically:

| File | Now imports from |
|---|---|
| `reflections/housekeeping/redis_ttl_cleanup.py` | `models.reflection_ignore` |
| `reflections/utilities.py` | `models.reflection_ignore` |
| `reflections/audits/pr_review_audit.py` | `models.pr_review_audit` |

`models/__init__.py` already imported canonically — untouched. 11 test import sites repointed across `test_reflections_redis.py` and `test_pr_review_audit.py`. Docs updated in `bridge-workflow-gaps.md`, `correlation-ids.md`, `worker-service.md`, `reflections.md`, `cowork-pr-review-audit.md`, plus the now-stale docstring at `agent/session_logs.py:5`.

## The trap this avoids

`test_reflections_package.py` patched `models.reflections.ReflectionIgnore` / `.PRReviewAudit` at three sites. All three production imports are **function-local**, so those patches worked only by accident of import order.

Repointing production without repointing the patch targets would have turned all three patches into **silent no-ops** — the tests would still PASS while hitting real Redis instead of the injected exception they exist to assert on.

All three are repointed, and each now asserts the mock was actually called (`assert_called_once`). This was verified as load-bearing: with the patch targets repointed but production still on the shim, two of the three failed with `Expected 'cleanup_expired' to have been called once. Called 0 times.` They pass only once production imports canonically. A stale patch target now fails loudly instead of silently degrading into an unintended integration test.

Notably the third site passed under the *wrong* target depending on worker import order — the shim made mock behavior order-dependent, which is an independent reason to delete it.

## Acceptance criteria

AC #2 (both shims deleted) and AC #3 (targeted tests pass) are met.

**AC #1 is corrected.** As written it demands `git grep "models.reflections\|bridge.session_logs"` return only `docs/plans/completed/` hits. That is unpassable without contorting unrelated files, so the code was not bent to satisfy it. The corrected criterion — **no hits in tracked `*.py` outside `docs/`** — is met with zero hits.

Remaining repo-wide hits, deliberately left alone:
- `site/assets/graph.js` (8 sites) — a generated artifact with no in-repo generator; `tests/unit/test_site_graph_consistency.py` explicitly tolerates phantom nodes for deleted files.
- `docs/plans/done/log-rotation-fix.md`, `docs/plans/done/pr_review_audit_reflection.md` — historical plan docs under `done/` rather than `completed/`.

## Verification

`ruff check` → `All checks passed!`; `ruff format` → 1370 files unchanged.

Targeted run: **215 passed**, plus **36 passed** in `test_worker_entry.py`.

Four failures were investigated and confirmed **pre-existing and environmental**, not regressions:
- `test_reflection_registered_in_yaml`, `test_reflection_entry_structure` — `FileNotFoundError` on `config/reflections.yaml`, a gitignored vault symlink that exists in the main checkout but not in worktrees. Unrelated to imports.
- `TestWorkerDryRun` (2 tests) — 30s subprocess timeouts under parallel xdist load; the command completes in 3.9s standalone. Reproduced on a clean stashed main tree in this same worktree, confirming they are not caused by this change.


valorengels:
## Review: PR #2943 — delete `bridge/session_logs.py` and `models/reflections.py` shims

**Verdict: APPROVED**

Reviewed against issue #2872. Pure import-path migration, no behavior change. Every claim in the PR body was independently verified rather than accepted.

### The central risk was handled correctly, and the guard is durable

Three sites in `tests/unit/test_reflections_package.py` patched `models.reflections.ReflectionIgnore` / `models.reflections.PRReviewAudit`. All three production imports are **function-local**, so repointing production without repointing the patch targets would have turned these into **silent no-ops** — still passing, but hitting real Redis instead of the injected exception they exist to assert on.

The builder repointed the patch targets *first*, before touching production, and added an `assert_called_once` to each. Two failed immediately with `Expected 'cleanup_expired' to have been called once. Called 0 times.`, making the no-op visible; they go green only once production imports canonically.

Verified present in the branch tree at `d8e19ca`, each with an explanatory comment:

| Line | Repointed target | Guard |
|---|---|---|
| `:131` | `models.reflection_ignore.ReflectionIgnore` | `mock_ri.get_active.assert_called_once()` |
| `:193` | `models.reflection_ignore.ReflectionIgnore` | `mock_ri.cleanup_expired.assert_called_once()` |
| `:424` | `models.pr_review_audit.PRReviewAudit` | `mock_pra.last_successful_run.assert_called_once()` |

These are the durable guard against this regressing. A stale patch target now fails loudly instead of silently degrading into an unintended integration test.

### An independent argument for deletion, surfaced during build

The third site (`test_load_ignore_entries_empty_redis`) **passed under the wrong patch target**, because the shim module happened to be imported fresh after the patch on that xdist worker. The shim was a second module object holding the same model, which made `unittest.mock.patch` behavior **import-order-dependent** — a silently flaky patch surface. This was not in the original recon and is a stronger justification for deletion than tidiness. It is now recorded in `docs/features/reflections.md` as the standing rationale.

### Canonical module survives

`agent/session_logs.py` (181 lines) is the canonical module and is untouched apart from the stale docstring at `:5`, which correctly drops the claim that `bridge/session_logs.py` re-exports from it. Only the 13-line `bridge/session_logs.py` re-export and the 15-line `models/reflections.py` were deleted. Confirmed via `git ls-tree`: neither shim exists on the branch.

### Acceptance criteria — AC #1 corrected, not force-satisfied

AC #1 as written demands `git grep "models.reflections\|bridge.session_logs"` return only `docs/plans/completed/` hits. That is **unpassable without contorting unrelated files**, and the builder correctly declined to bend code to satisfy it, stating the corrected criterion in the PR body instead: **no hits in tracked `*.py` outside `docs/`**.

Verified on the branch — that grep returns **zero hits** (exit 1). Remaining repo-wide hits are all legitimately left alone, and confirmed absent from the diff:
- `site/assets/graph.js` — generated artifact with no in-repo generator; `tests/unit/test_site_graph_consistency.py:3-5` explicitly tolerates phantom nodes for deleted files.
- `docs/plans/done/log-rotation-fix.md`, `docs/plans/done/pr_review_audit_reflection.md` — historical plan archives under `done/`, not `completed/`.

Declining an unpassable AC and stating the corrected one is the right call; force-satisfying it would have damaged unrelated files.

### The four failures are genuinely unrelated — reproduced, not assumed

"Pre-existing" is the most abused excuse in a review, so both claims were re-verified independently.

**`config/reflections.yaml` (2 tests, `FileNotFoundError`).** Confirmed structural: the file exists in the main checkout (`19335` bytes) and is **absent** in the worktree, because it is gitignored and materialized from the vault. Nothing to do with imports. Deliberately not "fixed" here — a parallel lane (#2875) owns this file's migration path.

**`TestWorkerDryRun` (2 tests, 30s subprocess timeouts).** The builder's stash-and-reproduce on a clean main tree *in the same worktree* is the evidence that actually settles this, and it holds. Confirmed three ways:
1. Both tests **pass** on the clean main checkout (`2 passed in 15.48s`).
2. Both tests **fail identically** in the worktree at the PR head — same `TimeoutExpired`, environment-bound.
3. Decisive: the exact command the test shells out to, `python -m worker --dry-run`, **succeeds** in the worktree at the PR head, loading all 9 projects with no `ModuleNotFoundError`.

That third check rules out the deletion as a cause outright — a broken import would fail fast, not hang for 30s. Root cause is sharper than "environmental": the dry-run performs a **live Claude CLI health check over the network** (~8s of a 14.2s wall time here). Two concurrent xdist workers each paying that cost, on a machine under contention, blow past the test's 30s `subprocess.run` timeout. A network-bound call inside a unit-test subprocess timeout is a latent flake worth its own issue, but it is out of scope for this PR and predates it.

### Gates

- `ruff check` → clean; `ruff format --check` → clean (1370 files unchanged).
- Targeted run in the worktree at head: **240 passed, 2 failed** — the 2 being the unrelated `TestWorkerDryRun` timeouts above.
- No `Co-Authored-By` or generated-by footer on any commit.
- `Closes #2872` present in the PR body.

### Scope

16 files, +36/-62. No new abstractions, no parallel-run migration, no commented-out code, no legacy bridge left behind. Consistent with the NO LEGACY CODE TOLERANCE principle: the shims are gone, not deprecated.

Issue #2876, opened 2026-08-19
De-hub agent/agent_session_queue.py -- stop using it as a re-export supermodule


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/sdlc-2876.md

> **Context:** [`agent/agent_session_queue.py`](agent/agent_session_queue.py) is the session queue module, but its top imports and re-exports symbols from roughly 8 other modules ("for backward compatibility"): `agent.session_state`, `agent.output_router`, `agent.session_completion`, `agent.session_executor`, `agent.session_health`, `agent.session_pickup`, `agent.session_revival`, and `agent.session_runner.liveness`. This re-export-hub pattern is a large part of why the file is 3,244 lines before its own queue logic even starts.

## Problem

**Current behavior:**
- Roughly 20 production files import from `agent.agent_session_queue`, many pulling symbols whose true home is a different module (e.g. `steer_session`, `re_enqueue_session`, `_transition_parent`, `MAX_NUDGE_COUNT`, `determine_delivery_action`).
- This makes `agent_session_queue.py` a de facto "god module" that everyone imports from regardless of what they actually need, obscuring true ownership of each symbol and inflating the file's size and blast radius.

**Desired outcome:**
`agent/agent_session_queue.py` contains only its own queue logic. Every production import pulls each symbol from its canonical module, and an architectural test prevents the re-export pattern from reappearing.

## Solution Sketch

- Phase 1: update production importers to import each symbol from its canonical module instead of `agent.agent_session_queue`.
- Phase 2: update tests that assert on or import the re-export hub.
- Phase 3: remove the re-export blocks and the `# noqa: F401` re-export idiom from `agent_session_queue.py`.
- Add an architectural test that fails if `agent/agent_session_queue.py` imports from the modules it used to re-export solely for re-export purposes.

## Recon: Skipped

Compiled from a pre-existing audit of the re-export blocks and import-site count, verified against the current tree before filing. Given the ~20-file blast radius, `/do-plan` should treat this as a phased migration rather than a single PR, and should re-verify the current import-site list at plan time since it is likely to have grown.

## Acceptance Criteria

- [ ] `agent/agent_session_queue.py` contains no "re-exported here for backward compatibility" block.
- [ ] Production code imports queue-owned symbols from `agent.agent_session_queue` and everything else from its canonical module.
- [ ] The full targeted test set passes.
- [ ] An architectural test guards against the re-export pattern reappearing.

## Downstream

[...]

PR #2947, opened 2026-08-24
refactor(agent): drop 16 dead re-exports from agent_session_queue (phase 1 of #2876)

### LINKED ISSUE #2952: Phase 1/7: delete dead re-exports from agent/agent_session_queue.py
Phase 1 of 7 under the #2876 de-hubbing issue.

#2876 is an *umbrella* issue covering the full de-hubbing of
`agent/agent_session_queue.py`. It is not a buildable unit and **stays open
until phases 2-7 land**. This issue is the closeable record for phase 1, so
the phase has its own tracking record instead of riding an umbrella that must
survive six more merges.

## Scope

**Deletes dead re-exports only. Changes no behavior.**

Remove re-exported symbols from `agent/agent_session_queue.py` that nothing
reaches through the hub. Deliberately out of scope, all owned by later phases:

- No importer is migrated to a canonical module.
- The re-export block itself stays.
- No architectural test is added.
- The tail `agent.session_stall_classifier` import is a genuine internal
  import, not a re-export, and is untouched.
- The 15 function-local deferred back-imports stay function-local. The cycle
  is `hub --(top-level)--> source --(lazy)--> hub`; hoisting any of them
  converts a benign cycle into a hard `ImportError`.

## Result: 15 symbols deleted

Re-export surface drops 81 -> 66.

`get_branch_state`, `_CONTINUATION_PM_MAX_DEPTH`, `_extract_issue_number`,
`_HARNESS_NOT_FOUND_MAX_RETRIES`, `_calendar_heartbeat`,
`_find_valor_calendar`, `HEARTBEAT_WRITE_INTERVAL`,
`_dependency_health_check`, `_reap_orphan_session_processes`,
`_POP_LOCK_TTL_SECONDS`, `_drain_startup_steering`, `dependency_status`,
`REVIVAL_COOLDOWN_SECONDS`, `_COOLDOWN_FILE`, `_save_cooldowns`.

Each is still exported by its canonical module and still reached there by its
real callers.

## Why 15 and not 16 or 34

Two corrections landed on the way here, in opposite directions.

**34 -> 16.** The original recon's grep was blind to two classes:

- *Relative imports.* `agent/__init__.py:3` is
  `from .agent_session_queue import (...)`, which no
  `from agent.agent_session_queue import` pattern matches. That alone keeps
  `record_revival_cooldown`, `check_revival`, `cleanup_stale_branches` and
  `queue_revival_agent_session` alive.
- *Patch-string-only symbols.* `tests/integration/test_silent_failures.py:266`
  patches `agent.agent_session_queue._load_cooldowns`. That symbol has no
  importer at all; only the patch string makes it load-bearing.

The rest of the gap is 28 symbols the hub's own code calls -- needed imports,
never re-exports, never in scope.

**16 -> 15.** Merge-gate review found one symbol in the 16 that is genuinely
live, via a class no repo-wide source sweep can see: **a runtime dotted-path
string in untracked config.**

`config/reflections.yaml` (vault-owned, gitignored, therefore outside every
`git grep`) declares:

```yaml
- name: stale-branch-cleanup
  every: 86400s
  enabled: true
  execution_type: function
  callable: "agent.agent_session_queue.cleanup_stale_branches_all_projects"
```

`agent/reflection_scheduler.py::_resolve_callable` does
`importlib.import_module(...)` then `getattr(module, func_name)`. The hub
defines no module-level `__getattr__`, so the lookup is exact. Reproduced
against both trees:

```
FAILS     AttributeError: module 'agent.agent_session_queue' has no attribute
                          'cleanup_stale_branches_all_projects'
RESOLVES  (after restoring the re-export)
```

The daily `stale-branch-cleanup` housekeeping job would have died silently, in
a checkout currently carrying 100+ session branches. `cleanup_stale_branches_all_projects`
is therefore **retained**, with an inline comment at the import site recording
why it cannot be deleted until the yaml is migrated to `agent.session_revival`.

Two sibling entries in the same yaml (`cleanup_corrupted_agent_sessions`, x2)
resolve through the hub today and will hit the same wall whenever a phase
targets them. Because the yaml is iCloud-synced vault config, the migration
must land and propagate to every machine *before* the corresponding hub
deletion, or the deletion breaks a live job on any machine that has not synced.

## Detection rule phases 2-7 inherit

A symbol is dead only if it has no hit in **five** classes:

1. Imports from the hub, absolute **and relative**.
2. Attribute access, with per-file alias resolution. Resolve the alias to the
   module it is actually bound to -- `sr`, `session_health`, `_executor` in
   this repo's tests bind to the **canonical** modules, and treating them as
   hub references falsely spares symbols.
3. `patch("agent.agent_session_queue.<sym>")` and
   `patch.object(<hub alias>, "<sym>")` strings.
4. Use by the hub's own code (needed imports, not re-exports).
5. **Dotted-path strings in untracked config** -- sweep both
   `config/reflections.yaml` and its vault original
   `~/Desktop/Valor/reflections.yaml`. Invisible to static analysis of the
   repo by construction, so it has to be an explicit checklist step.

## Acceptance criteria

- [x] 15 symbols removed from `agent/agent_session_queue.py`; all 66 surviving
      re-exports still resolve.
- [x] No importer migrated, re-export block intact, no architectural test added.
- [x] No deferred back-import hoisted to module top level.
- [x] All three `config/reflections.yaml` hub callables still resolve.
- [x] `ruff check` / `ruff format --check` clean.
- [x] Targeted unit + integration suites green across every file that imports,
      aliases or patches the hub (518 passed), plus 157 passed at the corrected
      head including `test_reflection_scheduler.py`.

Refs #2876

valorengels:
Phase 1 of 7 for #2876. Scope is deliberately narrow: **delete re-exported symbols from `agent/agent_session_queue.py` that nothing references.** No importer is migrated, the re-export block itself stays, and the architectural test is not added — those are later phases owned by other lanes.

## Result

**15 symbols deleted** (`agent/agent_session_queue.py`). The re-export surface drops from 81 symbols to 66.

> **Merge-gate correction.** This PR originally deleted **16**. Review found `cleanup_stale_branches_all_projects` is genuinely live: `config/reflections.yaml` (vault-owned, gitignored) declares `callable: "agent.agent_session_queue.cleanup_stale_branches_all_projects"` for the enabled daily `stale-branch-cleanup` job, and `reflection_scheduler._resolve_callable` getattrs it off the hub. No repo-wide source sweep can see an untracked-config reference. Verified by reproduction: `AttributeError` on the 16-deletion tree, resolves on the 15-deletion tree. That re-export is restored with an explanatory comment; deleting it needs the yaml migrated to `agent.session_revival` first.

| Symbol | Canonical module |
|---|---|
| `get_branch_state` | `agent.branch_manager` |
| `_CONTINUATION_PM_MAX_DEPTH` | `agent.session_completion` |
| `_extract_issue_number` | `agent.session_completion` |
| `_HARNESS_NOT_FOUND_MAX_RETRIES` | `agent.session_executor` |
| `_calendar_heartbeat` | `agent.session_executor` |
| `_find_valor_calendar` | `agent.session_executor` |
| `HEARTBEAT_WRITE_INTERVAL` | `agent.session_health` |
| `_dependency_health_check` | `agent.session_health` |
| `_reap_orphan_session_processes` | `agent.session_health` |
| `_POP_LOCK_TTL_SECONDS` | `agent.session_pickup` |
| `_drain_startup_steering` | `agent.session_pickup` |
| `dependency_status` | `agent.session_pickup` |
| `REVIVAL_COOLDOWN_SECONDS` | `agent.session_revival` |
| `_COOLDOWN_FILE` | `agent.session_revival` |
| `_save_cooldowns` | `agent.session_revival` |

Every one of these is still exported by its canonical module and still reached there by its real callers. Nothing reached them *through the hub*.

## Method

Per symbol, across every tracked file, a hit in any of these four categories disqualifies deletion:

1. **Imports from the hub** — AST `ImportFrom` with module `agent.agent_session_queue`, **including relative imports** resolved against the importing file's package.
2. **Attribute access** — `agent_session_queue.<sym>`, plus aliased module access (`asq.`, `q.`, `queue_module.`) resolved per-file from the actual alias binding.
3. **Patch target strings** — `patch("agent.agent_session_queue.<sym>")` and `patch.object(<hub alias>, "<sym>")`. A symbol referenced only by a patch string is live; deleting it would silently turn the patch into a no-op or an `AttributeError`.
4. **Internal use by the hub's own code** — AST `Name`/`Attribute` loads outside the import block. These are needed imports, not re-exports, and were kept.

## Deviation from the assigned recon

The task brief predicted **34 dead symbols of 80**. The verified count is **16 of 81**. Flagging rather than forcing the number, because the gap is two concrete false-dead classes that a coarser sweep misses — and deleting either set would have broken `main`:

- **Relative imports.** `agent/__init__.py` does `from .agent_session_queue import (...)`, which does not match a `from agent.agent_session_queue import` grep. That alone keeps `record_revival_cooldown`, `check_revival`, `cleanup_stale_branches`, and `queue_revival_agent_session` alive.
- **Patch strings on otherwise-unimported symbols.** `tests/integration/test_silent_failures.py:266` patches `agent.agent_session_queue._load_cooldowns`. The symbol has no importer at all; only the patch string keeps it load-bearing.

The remaining difference is symbols the hub's own code calls (28 of them), which are needed imports rather than re-exports.

Two intermediate analyzer bugs were found and fixed while narrowing this down (missed relative imports; an alias regex greedy enough to treat `sr.` / `session_health.` references to *canonical* modules as hub references). The 16 above survived the corrected analysis and a full-repo confirmation grep showing no hub-qualified reference to any of them.

The `agent.session_stall_classifier` import at the file's tail is a genuine internal import used by the hub, not a re-export, so it is untouched.

## Verification

```
tests/unit/test_agent_session_queue.py + test_steering_mechanism + test_harness_retry     119 passed
tests/unit/test_health_check_recovery_finalization.py + never_started_recovery
  + stale_cleanup + session_health_fence_guards + recovery_respawn_safety                207 passed
tests/unit/test_agent_session_queue_async.py + nudge_loop + worker_entry
  + agent_session_hierarchy + resume_hydration + qa_nudge_cap + pipeline_integrity
  + agent_session_queue_revival_helper                                                   183 passed
tests/integration/test_silent_failures.py + test_session_zombie_health_check.py             9 passed
                                                                                        ---------
                                                                                         518 passed, 0 failed
```

Selected as the unit and integration files that import from, alias, or patch the hub. `python -m ruff check .` → All checks passed. `python -m ruff format --check agent/` → 86 files already formatted. Import smoke test confirms `agent`, `worker.__main__`, `tools.agent_session_scheduler`, `tools.valor_session` and all eight formerly-re-exported modules still import, all 65 surviving re-exports still resolve, and all 16 deleted symbols are gone from the hub.

A full `tests/unit` run was attempted first but collapsed with all 50 xdist workers down and the test-DB slot pool exhausted — the machine had 161 live pytest processes from concurrent sibling lanes. `scripts/reap-xdist.sh` reported no orphans, so that was contention, not a regression; the targeted files above were then run serially with `-n 0`.

Closes #2952
Refs #2876

`#2952` is the closeable record for phase 1 and closes on merge. **#2876 is the
umbrella issue and stays OPEN** — it covers phases 2-7 of de-hubbing
`agent/agent_session_queue.py` and must survive six more merges.

### DIFF
--- a/agent/agent_session_queue.py
+++ b/agent/agent_session_queue.py
@@ -32,7 +32,6 @@ from popoto.exceptions import ModelException
 
 # Shared mutable session-tracking state — re-exported here for backward compatibility.
 import agent.session_state as _session_state  # noqa: F401 (also used for mutation sites)
-from agent.branch_manager import get_branch_state  # noqa: F401
 from agent.output_handler import OutputHandler
 
 # Output routing — decision logic lives in output_router; re-exported here
@@ -49,22 +48,17 @@ from agent.redis_offload import offload_redis
 
 # Session completion (post-execution lifecycle) — re-exported here for backward compatibility.
 from agent.session_completion import (  # noqa: F401
-    _CONTINUATION_PM_MAX_DEPTH,
     _complete_agent_session,
     _diagnose_missing_session,
-    _extract_issue_number,
     _transition_parent,
 )
 
 # Session executor (CLI harness, nudge/re-enqueue, steer) — re-exported for backward compatibility.
 from agent.session_executor import (  # noqa: F401
     _HARNESS_EXHAUSTION_MSG,
-    _HARNESS_NOT_FOUND_MAX_RETRIES,
     _HARNESS_NOT_FOUND_PREFIX,
-    _calendar_heartbeat,
     _enqueue_nudge,
     _execute_agent_session,
-    _find_valor_calendar,
     _handle_harness_not_found,
     re_enqueue_session,
     steer_session,
@@ -75,7 +69,6 @@ from agent.session_health import (  # noqa: F401
     AGENT_SESSION_HEALTH_CHECK_INTERVAL,
     AGENT_SESSION_HEALTH_MIN_RUNNING,
     HEARTBEAT_FRESHNESS_WINDOW,
-    HEARTBEAT_WRITE_INTERVAL,
     MAX_RECOVERY_ATTEMPTS,
     TASK_CANCEL_TIMEOUT,
     _agent_session_health_check,
@@ -83,10 +76,8 @@ from agent.session_health import (  # noqa: F401
     _agent_session_hierarchy_health_check,
     _apply_recovery_transition,
     _cleanup_orphaned_claude_processes,
-    _dependency_health_check,
     _has_progress,
     _is_ledger,
-    _reap_orphan_session_processes,
     _recover_interrupted_agent_sessions_startup,
     _should_kill_no_progress,
     _sweep_dead_worker_sessions,
@@ -102,26 +93,19 @@ from agent.session_logs import save_session_snapshot
 
 # Session pickup (pop locking, startup steering drain, dependency checks) — re-exported here.
 from agent.session_pickup import (  # noqa: F401
-    _POP_LOCK_TTL_SECONDS,
     _acquire_pop_lock,
-    _drain_startup_steering,
     _maybe_inject_resume_hydration,
     _pop_agent_session,
     _pop_agent_session_with_fallback,
     _release_pop_lock,
-    dependency_status,
 )
 
 # Revival detection — re-exported here for backward compatibility.
 from agent.session_revival import (  # noqa: F401
-    _COOLDOWN_FILE,
-    REVIVAL_COOLDOWN_SECONDS,
     _load_cooldowns,
-    _save_cooldowns,
     _session_branch_name,
     check_revival,
     cleanup_stale_branches,
-    cleanup_stale_branches_all_projects,
     maybe_send_revival_prompt,
     queue_revival_agent_session,
     record_revival_cooldown,
--- a/agent/agent_session_queue.py
+++ b/agent/agent_session_queue.py
@@ -106,6 +106,13 @@ from agent.session_revival import (  # noqa: F401
     _session_branch_name,
     check_revival,
     cleanup_stale_branches,
+    # Resolved at runtime by the reflection scheduler through this hub path:
+    # config/reflections.yaml's `stale-branch-cleanup` entry declares
+    # `callable: "agent.agent_session_queue.cleanup_stale_branches_all_projects"`,
+    # which reflection_scheduler._resolve_callable getattrs off this module.
+    # Untracked config, so no repo-wide source sweep can see the reference.
+    # Deleting this needs the yaml migrated to agent.session_revival first.
+    cleanup_stale_branches_all_projects,
     maybe_send_revival_prompt,
     queue_revival_agent_session,
     record_revival_cooldown,


valorengels:
## Review: PR #2947 — phase 1 de-hub of `agent/agent_session_queue.py`

**Verdict: APPROVED** (at head `4abe0aa3f1567ff7a57c7fbd25a3d04e64396141`, after one blocker was found and fixed in-gate.)

### Scope conformance — confirmed

The diff touches exactly one file, `agent/agent_session_queue.py`, and only removes names from import blocks.

- No importer migrated. Every `from agent.agent_session_queue import ...` site in `bridge/`, `tools/`, `scripts/`, `models/`, `reflections/` and `tests/` is byte-identical to `main`.
- The re-export block itself is intact — this deletes entries, it does not restructure.
- No architectural test added.
- The tail `agent.session_stall_classifier` import is untouched — correctly identified as a genuine internal import, not a re-export.
- The 15 function-local deferred back-imports (`session_executor.py`, `session_health.py`, `session_pickup.py`, `session_revival.py`, `output_handler.py`, `session_runner/adapter.py`, `models/session_lifecycle.py`) remain function-local. Nothing was hoisted, so the benign `hub --(top-level)--> source --(lazy)--> hub` cycle is preserved.

### The 34-vs-16 deviation — the builder was right, the brief was wrong

Independently re-derived rather than taken on trust. The recon's "34 dead of 80" was produced by a grep blind to two reference classes, both of which would have broken `main`:

1. **Relative imports.** `agent/__init__.py:3` is `from .agent_session_queue import (...)`, invisible to a `from agent.agent_session_queue import` pattern. It keeps `check_revival`, `cleanup_stale_branches`, `queue_revival_agent_session`, `record_revival_cooldown` alive — all four confirmed **present** in the surviving set. Also confirmed that `get_branch_state`, which *does* appear in `agent/__init__.py`, is imported there from `.branch_manager` and not from the hub, so its deletion is sound.
2. **Patch-string-only symbols.** `tests/integration/test_silent_failures.py:266` patches `agent.agent_session_queue._load_cooldowns`; the symbol has no importer at all. Confirmed **present** at `agent/agent_session_queue.py:105`.

The opposite-direction analyzer bug was also correctly fixed. Verified the alias bindings by hand: `sr` = `agent.session_revival` (`test_silent_failures.py:189,219`), `session_health` = `agent.session_health` (`test_session_health_orphan_process_reap.py:43`, `test_orphan_reap_forward_scan.py:43`), `_executor`/`_se_module`/`se` = `agent.session_executor`. All are canonical-module references, not hub references, so `_COOLDOWN_FILE`, `_reap_orphan_session_processes`, `_dependency_health_check` and `_calendar_heartbeat` are correctly deleted from the hub.

Every one of the deleted names was swept repo-wide with `git grep -w` at PR head; each surviving hit resolves to the canonical module.

### BLOCKER found and fixed: one deleted symbol was genuinely live

`cleanup_stale_branches_all_projects` is reachable through the hub by a path no source sweep can see — a **runtime dotted-path string in untracked config**.

`config/reflections.yaml` (vault-owned, gitignored, therefore outside every repo-wide grep):

```yaml
- name: stale-branch-cleanup
  every: 86400s
  enabled: true
  callable: "agent.agent_session_queue.cleanup_stale_branches_all_projects"
```

`agent/reflection_scheduler.py::_resolve_callable` does `importlib.import_module("agent.agent_session_queue")` then `getattr(module, "cleanup_stale_branches_all_projects")`. The hub defines no module-level `__getattr__`, so the lookup is exact. Reproduced against both trees:

```
FAILS     ...cleanup_stale_branches_all_projects  AttributeError: module
          'agent.agent_session_queue' has no attribute 'cleanup_stale_branches_all_projects'
RESOLVES  ...cleanup_stale_branches_all_projects   (after the fix)
```

The daily `stale-branch-cleanup` housekeeping job would have died silently at the next tick — in a checkout currently carrying 100+ session branches. Fixed in `4abe0aa3f` by restoring that single re-export with a comment recording why it cannot be deleted until the yaml is migrated. **Phase 1 lands 15 deletions, not 16.**

The two sibling entries in the same file (`agent.agent_session_queue.cleanup_corrupted_agent_sessions`, ×2) were never in the deletion set and still resolve.

### Detection gap that phases 2-7 inherit

The analyzer covers imports (absolute + relative), attribute access (incl. aliases), `patch(...)` strings, and hub-internal use. It does **not** cover dotted-path strings in untracked config. That is not a repo-visible class, so no amount of `git grep` finds it. Phases 2-7 must sweep `config/reflections.yaml` and the vault copy for any symbol they intend to delete. Recorded on #2876.

### Verification

- `tests/unit/test_agent_session_queue.py`, `tests/unit/test_reflection_scheduler.py`, `tests/integration/test_silent_failures.py` at the fixed head: **157 passed, 0 failed** (serial, `-n 0`).
- Builder's targeted sweep across every unit/integration file importing, aliasing or patching the hub: **518 passed, 0 failed**.
- `ruff check` / `ruff format --check`: clean.

The builder's coverage caveat (no full `tests/unit` run — 161 concurrent pytest processes from sibling lanes exhausted the xdist workers and test-DB pool, with `reap-xdist.sh` reporting no orphans, i.e. contention not regression) is **accepted, not blocking**: the diff removes import bindings from a single module and cannot affect behavior outside modules that resolve names through that module, and the targeted set is exactly that closure plus the newly-relevant reflection scheduler.

### Not blocking

The `Refs #2876` trailer is correct and deliberate — this is 1 of 7 phases and #2876 must remain open after merge.

Issue #2698, opened 2026-08-10
Activate AgentSession Meta.ttl 30-day expiry deliberately (it has never fired)


tomcounsell:
## The situation

`AgentSession` declares `Meta.ttl = int(settings.timeouts.agent_session_retain_ttl_s)` — 2592000 seconds, 30 days (`models/agent_session.py:628-631`, commented "hard backstop for retain_for_resume BUILD sessions").

**It has never fired.** Every full `save()` resets the Redis key's TTL to the full 30 days, and `cleanup_corrupted_agent_sessions` full-saves every hydrated AgentSession row as a corruption probe on a cadence between 5 minutes and 1 hour. So every healthy row's expiry clock is continuously pushed back and no session row has ever aged out.

Measured on an isolated Redis (popoto 1.8.0, real `AgentSession`):

```
ttl right after initial save : 2592000
ttl after 3s idle            : 2591997
ttl after a no-op save probe : 2592000   <- reset to the ceiling
ttl around is_valid()        : unchanged
```

## Why this is being filed now

#2660 removes that probe write (it was forging `updated_at` liveness). Removing it also stops the incidental TTL refresh, which would activate the 30-day expiry for the first time — as a silent side effect of a liveness bugfix.

That is a data-retention policy change and it should be a deliberate decision, so #2660 ships **TTL-neutral**: it keeps today's effective no-expiry behavior via an explicitly-named keepalive write, and defers the policy question here. See that PR for the mechanism.

## The decision to make

1. **Should AgentSession rows expire at all, and at what age?** 30 days is the currently-declared-but-inert value. It is env-overridable via `TIMEOUTS__AGENT_SESSION_RETAIN_TTL_S`.
2. **What is lost when a row expires?** Redis is the authoritative store. The dashboard (`ui/data/sdlc.py`, `ui/data/jobs.py`), resume paths, and `parent_agent_session_id` lineage all read it.
3. **The archive does NOT backstop this.** `agent/session_archive.py` exports every row to SQLite, but the only restore path is `restore_if_empty()` — a guarded **cold-start** rehydrate that fires only when Redis has no AgentSession keys at all. Individual rows expiring one at a time are simply gone from Redis; nothing notices and nothing restores them. If expiry is wanted, either that gap gets closed or expiry is accepted as real deletion from the live store.
[...]

PR #2949, opened 2026-08-24
Remove deprecated AgentSession aliases, dead methods, and legacy kwarg mapping

### LINKED ISSUE #2873: Remove deprecated AgentSession aliases and dead methods
> **Context:** [`AgentSession`](models/agent_session.py) is the core session model. Over successive refactors it accumulated deprecated aliases, zero-caller methods, and legacy field-mapping paths in `_normalize_kwargs()` that were never cleaned up, plus a docstring that documents properties that no longer exist.

## Problem

**Current behavior:**
- `AgentSession.get_parent_chat_session()` and `AgentSession.get_dev_sessions()` have zero callers outside the model itself.
- `get_history_list()` / `_get_history_list` are used in only 4 production places: `agent/goal_gates.py`, `monitoring/session_watchdog.py`, `monitoring/session_status.py`, and `ui/data/sdlc.py` (via the `history` property).
- `AgentSession.sender` (an alias for `sender_name`) is used only in `tools/session_tags.py` plus tests.
- `AgentSession.history` (an alias for `session_events`) is used only in `ui/data/sdlc.py` plus tests.
- The model docstring claims `parent_session_id` and `parent_chat_session_id` are deprecated property aliases — those properties no longer exist on the model. The docstring is describing history, not the current API.
- `_normalize_kwargs()` contains legacy mapping paths for `parent_job_id`, `job_id`, `work_item_slug`, `last_activity`, `scheduled_after`, `history`, `stage_states`, `commit_sha`, `summary`, and a dead-field pop list, most of which are no longer reachable from any live caller.

**Desired outcome:**
The model exposes only its current, real API surface. Zero-caller methods and aliases are removed, call sites read the canonical fields directly, and the docstring accurately describes the current schema.

## Solution Sketch

- Delete `get_parent_chat_session()` and `get_dev_sessions()`.
- Migrate the 4 `get_history_list()` / `_get_history_list` call sites to read `session.session_events` directly (add one clearly named helper only if formatting logic is genuinely duplicated across call sites, not preemptively).
- Migrate `session.sender` → `session.sender_name`; migrate `session.history` → `session.session_events`.
- Remove the `parent_session_id` / `parent_chat_session_id` claims from the model docstring; describe only the current `parent_agent_session_id` link.
- Audit `_normalize_kwargs()` mappings and remove any that are unreachable from live callers. Keep only what `agent/session_archive.py` still needs for archive-restore compatibility, if anything.

## Recon: Skipped

Compiled from a pre-existing audit of caller counts for each symbol above, verified against the current tree before filing. `/do-plan` should re-confirm caller counts at plan time in case the tree has moved.

## Acceptance Criteria

- [ ] `git grep -n "get_parent_chat_session\|get_dev_sessions\|get_history_list\|_get_history_list\|session.history\|session.sender"` returns no hits in production code.
- [ ] The `AgentSession` docstring describes only the current parent link (`parent_agent_session_id`) and current event storage (`session_events`) — no references to removed properties.
- [ ] Unit tests are updated for the migrated call sites and pass.

## Downstream

This issue feeds into `/do-plan`, which will produce a plan document. Constraint: `_normalize_kwargs()` changes must preserve archive-restore compatibility if `agent/session_archive.py` still emits any of the legacy keys — verify before removing that mapping.

valorengels:
Closes #2873

Removes the deprecated `AgentSession` surface: two zero-caller wrappers, two property aliases, the `get_history_list()` / `_get_history_list` pair, and **all nine** legacy field-mapping paths in `_normalize_kwargs()`. The model now exposes only its current, real API.

## What was removed

| Symbol | Replacement at the call site |
|---|---|
| `get_parent_chat_session()` | `get_parent_session()` (was already the only implementation) |
| `get_dev_sessions()` | `get_child_sessions()` (ditto) |
| `get_history_list()` / `_get_history_list` | `models.session_event.format_event_lines(session.session_events)` |
| `.sender` property | `.sender_name` |
| `.history` property + setter | `.session_events` |

`_normalize_kwargs()` legacy paths deleted: `work_item_slug` → `slug`; `last_activity` → `updated_at`; `scheduled_after` → `scheduled_at`; `parent_job_id` → `parent_agent_session_id`; `history` → `session_events`; `stage_states` → event; `commit_sha` → event; `summary` → event; the `job_id` discard; and the dead-field pop list (`depends_on`, `stable_agent_session_id`, `scheduling_depth`, `_qa_mode_legacy`).

What `_normalize_kwargs()` still does, unchanged: constituent-field consolidation into `initial_telegram_message` / `extra_context`, the `watchdog_unhealthy` → `unhealthy_reason` back-alias (#1927), the `agent_session_id` `AutoKeyField` discard, and the `response_delivered_at` defensive coercion (#929).

Note that `stage_states`, `commit_sha`, `summary`, and `work_item_slug` remain live **properties** on the model (several with setters that append the appropriate event). Only their `_normalize_kwargs` constructor-kwarg paths are gone. `AgentSession.create_child()` keeps its own explicit `stage_states` parameter and is unaffected.

## Bug fixed: unbounded event-log growth on a read path

`commit_sha` and `summary` were the only two mapping paths **lacking** the `"session_events" not in kwargs` guard that the other event-producing paths had. They unconditionally did `events.append(...)` onto whatever came out of Redis.

The consequence: any row carrying *both* a live `session_events` key and a stale `commit_sha`/`summary` key grew one extra event on **every** `query.get()`. If that instance was then saved, the duplicate persisted — while the stale hash key remained, ready to do it again on the next read. `session_events` has no count cap (deliberately, per the durability work in #2494), so nothing bounded this. Deleting these two paths removes the growth mechanism outright.

Regression coverage: `tests/unit/test_agent_session_legacy_surface.py::test_no_duplicate_event_growth_on_hydration`.

## New helper

`models/session_event.py` gains `format_event_lines(events) -> list[str]`, a module-level function that renders `session_events` as display strings: dict entries become `f"[{event_type}] {text}"`, flat legacy string entries pass through, anything else is skipped, non-list input yields `[]`. Byte-identical output to the removed method. It lives next to `SessionEvent` because that is where the shape is defined, and because a module function does not re-add API surface to the model.

Three production call sites migrated to it: `agent/goal_gates.py` (matches on the `[stage]` prefix, so the exact format is load-bearing), `monitoring/session_status.py`, `monitoring/session_watchdog.py`.

**Silent-degradation trap, closed.** Two of those three call sites were written as `if hasattr(session, "_get_history_list")`. Deleting the method without touching them would have made both quietly evaluate to "no history" forever instead of failing loudly. Both guards are gone; the calls are now unconditional.

The issue text says there are four production callers and names `ui/data/sdlc.py` as one. That is wrong — there are three. `ui/data/sdlc.py` used the `.history` property, not `get_history_list()`, and is migrated to `.session_events` separately.

## Required disclosure: the hydration assumption

**Operative assumption, an explicit owner decision made with the risk stated: no `AgentSession` rows written before 2026-04-01 survive in Redis, or their legacy field data is acceptable to lose.**

This matters because **"no live Python caller" is the wrong test for these paths.** Popoto's non-lazy decode does `model_class(**model_attrs)` with the **raw Redis hash** — every key the hash contains, not a declared-field subset (`popoto/models/encoding.py:337`). So `query.get()`, `get_many()`, `async_get()`, and `async_get_many()` all run `_normalize_kwargs` over whatever legacy keys a row happens to carry, with no Python caller ever naming them. (`filter()` and `all()` default `lazy=True`, use `object.__new__`, and bypass `__init__` entirely — those paths never ran the mapping in the first place.)

Two of the nine paths carry real consequence if such rows exist:

- **`parent_job_id`** — a pre-rename row would previously surface with its parent link intact. It now hydrates with `parent_agent_session_id=None` and, on the next save, persists as **orphaned**.
- **`history`** — a pre-rename row now hydrates with an empty event log, and the next save wipes the old entries.

The remaining seven are benign: `job_id` and the dead-field pop list moved no data at all, and `work_item_slug` / `last_activity` / `scheduled_after` / `stage_states` / `commit_sha` / `summary` degrade to a missing display value rather than a broken link.

Three further facts, so the assumption is judged on real evidence:

1. **The "old rows aged out" argument does not hold here.** `Meta.ttl` is 30 days (`models/agent_session.py`), but it **never actually fired** until #2698: the corruption sweep full-`save()`d every hydrated row as a validation probe, and every `save()` resets the key's TTL to the ceiling. The clock was reset on every worker start, every `/update`, and every hourly cleanup tick, indefinitely. The model's own `Meta` comment documents this.
2. **No existing migration covers this field set.** The fields were removed in `3d2823923` (#609 / #628) on 2026-04-01. The three strip migrations in the tree — `migrate_schema_diet_fields.py`, `migrate_strip_pty_fields.py`, `migrate_strip_pid_fields.py` — target the #1927 and #1924 field sets, not this one.
3. **`agent/session_archive.py` is ruled out as a legacy-key source.** `_serialize_session` enumerates `session._meta.fields` — only currently-declared fields — and the archive was created 2026-07-04, three months after the field removal. A restore therefore cannot reintroduce a legacy key.

Deliberately **not touched**: the `agent_session_id` pop in `_normalize_kwargs` (and its counterpart note in `agent/session_archive.py`). Removing it would mint a fresh `AutoKeyField` UUID on every archive restore and dangle every `parent_agent_session_id` link. It was never on this list.

## Corrected comment

The pop-list comment claimed Popoto "silently drops any kwarg that doesn't match a declared field". Directionally right about crash-safety, technically wrong: `Model.__init__` does `self.__dict__.update(kwargs)` (`popoto/models/base.py:518`), so unknown keys land in the instance dict. They are simply never *persisted* — encoding iterates `_meta.fields` only. The replacement comment says that. Same conclusion (an unknown key never raises), accurate mechanism.

## Docstring and docs corrections

The module docstring claimed `parent_session_id` and `parent_chat_session_id` were deprecated property aliases delegating to `parent_agent_session_id`. **Neither property exists anywhere in the file** — the docstring was describing history. Replaced with the current parent link plus a section naming `session_events` as the sole event log.

`docs/features/agent-session-model.md` carried the same stale mapping list and is updated.

A code sample in `docs/guides/sdlc-storyline-example.md` called `get_history_list()` and then tested each entry with `isinstance(entry, dict)` — which could never match, since the method returned strings. Rewritten against `session_events` directly, which is what it meant.

Historical `docs/plans/completed/` mentions were left alone as archives.

## Incidental, guard-forced

`agent/goal_gates.py` carried a bare `timeout=10` on its `gh` subprocess call. Touching the file tripped the repo's inline-timeout-literal commit guard, which blocks the commit outright. Switched to `settings.timeouts.git_subprocess_s`, the field whose own description names `gh` CLI calls as its family. This raises the ceiling from 10s to 60s; it is a hang ceiling, not an added delay.

## Verification

`python -m ruff check .` — All checks passed. `python -m ruff format --check .` — 1374 files already formatted.

Targeted runs via `scripts/pytest-clean.sh` (never bare pytest). Note these ran with `-n 0`: the machine had several other agents' suites running concurrently (load 22, 23 pytest processes), and xdist workers were being killed off before collection finished. Serial runs are unaffected.

```
tests/unit/{test_session_event_formatting, test_agent_session_legacy_surface,
            test_goal_gates, test_sdlc_mode, test_model_relationships}   -> 131 passed

tests/unit/{test_message_drafter, test_session_status, test_stall_detection,
            test_transcript_liveness, test_ui_sdlc_data}                 -> 317 passed

tests/unit/{test_agent_session_queue, test_session_archive,
            test_pipeline_state_machine, test_stage_states_helpers}      -> 288 passed
            (the _normalize_kwargs-adjacent suites, as blast-radius insurance)

tests/integration/{test_agent_session_lifecycle, test_connectivity_gaps,
                   test_lifecycle_transition} + tests/e2e/test_session_lifecycle.py
                                                                         -> 77 passed, 2 failed
```

The 2 failures are `test_connectivity_gaps.py::TestSdkClientEnvVar::{test_valor_session_id_in_source, test_valor_session_id_conditional_on_session_id}`. They assert on the *source text* of `agent/sdk_client.py`, which this branch does not touch. **Verified pre-existing rather than assumed:** both were re-run in a throwaway worktree checked out at `origin/main` and fail there identically.

### DIFF
--- a/agent/goal_gates.py
+++ b/agent/goal_gates.py
@@ -28,6 +28,9 @@ from dataclasses import dataclass
 from pathlib import Path
 from typing import TYPE_CHECKING
 
+from config.settings import settings
+from models.session_event import format_event_lines
+
 if TYPE_CHECKING:
     from models.agent_session import AgentSession
 
@@ -64,13 +67,14 @@ def _run_gh_command(args: list[str], working_dir: str | Path | None = None) -> s
 
     Raises:
         subprocess.CalledProcessError: If the command fails.
-        subprocess.TimeoutExpired: If the command takes longer than 10 seconds.
+        subprocess.TimeoutExpired: If the command exceeds
+            ``settings.timeouts.git_subprocess_s``.
     """
     result = subprocess.run(
         ["gh", *args],
         capture_output=True,
         text=True,
-        timeout=10,
+        timeout=settings.timeouts.git_subprocess_s,
         cwd=working_dir,
     )
     result.check_returncode()
@@ -167,7 +171,7 @@ def check_test_gate(slug: str, session: AgentSession | None = None) -> GateResul
     # Primary: check session history
     if session is not None:
         try:
-            history = session.get_history_list()
+            history = format_event_lines(session.session_events)
             for entry in history:
                 if not isinstance(entry, str):
                     continue
--- a/docs/features/agent-session-model.md
+++ b/docs/features/agent-session-model.md
@@ -343,7 +343,9 @@ cleaned up by the TTL when they next touch Redis).
 
 ## Backward Compatibility
 
-- `_normalize_kwargs()` maps old field names to their new consolidated equivalents: `message_text`, `sender_name`, `sender_id`, `telegram_message_id`, `chat_title` -> `initial_telegram_message`; `revival_context`, `classification_type`, `classification_confidence` -> `extra_context`; `work_item_slug` -> `slug`; `last_activity` -> `updated_at`; `scheduled_after` -> `scheduled_at`; `history` -> `session_events`; `watchdog_unhealthy` -> `unhealthy_reason` (schema diet, #1927)
+- `_normalize_kwargs()` maps constituent field names into their consolidated DictFields: `message_text`, `sender_name`, `sender_id`, `telegram_message_id`, `chat_title` -> `initial_telegram_message`; `revival_context`, `classification_type`, `classification_confidence` -> `extra_context`. It also back-aliases `watchdog_unhealthy` -> `unhealthy_reason` (schema diet, #1927) and discards `agent_session_id` (an `AutoKeyField`).
+- `_normalize_kwargs()` maps **nothing else**. The legacy paths for `work_item_slug`, `last_activity`, `scheduled_after`, `parent_job_id`, `job_id`, `history`, `stage_states`, `commit_sha`, `summary`, and the dead-field pop list were removed in #2873, along with the `.sender` and `.history` property aliases and the `get_parent_chat_session()` / `get_dev_sessions()` / `get_history_list()` wrappers. Pass the canonical field names (`slug`, `updated_at`, `scheduled_at`, `parent_agent_session_id`, `session_events`) instead. Render an event log for display with `models.session_event.format_event_lines()`.
+- Unknown kwargs are harmless: Popoto's `Model.__init__` does `self.__dict__.update(kwargs)`, so an unrecognized key lands in the instance dict and never raises, and encoding iterates `_meta.fields` only, so it is never persisted.
 - `__setattr__` auto-converts float timestamps to `datetime` for DatetimeField fields
 - Property accessors provide read access to old field names (`sender_name`, `message_text`, etc.) for backward compatibility
 - `models/agent_session.py` exports `AgentSession = AgentSession` (shim)
--- a/docs/guides/sdlc-storyline-example.md
+++ b/docs/guides/sdlc-storyline-example.md
@@ -442,12 +442,12 @@ The pipeline graph edge `("PATCH", "success") -> "TEST"` fires, but since the wo
 **What could go wrong (stuck in a patch loop)**: Suppose the fix for the off-by-one introduced a new failure. The pipeline would cycle: TEST(fail) -> PATCH -> TEST(fail) -> PATCH -> TEST. The `cycle_count` parameter tracks this:
 
 ```python
-# Count PATCH cycles from history for the max-cycle safety valve
+# Count PATCH cycles from the event log for the max-cycle safety valve
 cycle_count = 0
-history = session.get_history_list()
-for entry in history:
-    if isinstance(entry, dict) and entry.get("stage") == "PATCH":
-        cycle_count += 1
+for event in session.session_events or []:
+    if isinstance(event, dict) and event.get("event_type") == "stage":
+        if "PATCH" in event.get("text", ""):
+            cycle_count += 1
 ```
 
 When `cycle_count >= MAX_PATCH_CYCLES` (which is 3), `get_next_stage` returns `None`:
--- a/models/agent_session.py
+++ b/models/agent_session.py
@@ -15,10 +15,13 @@ Session types (permission model):
     Participates in group conversations without orchestration authority.
 
 Parent-child relationship:
-  parent_agent_session_id is the canonical parent link.
-  parent_session_id and parent_chat_session_id are deprecated aliases that
-  delegate to parent_agent_session_id via property.
-  Use create_child() to spawn child sessions.
+  parent_agent_session_id is the sole parent link; get_parent_session() and
+  get_child_sessions() traverse it. Use create_child() to spawn child sessions.
+
+Event storage:
+  session_events is the sole event log (a list of SessionEvent dicts, see
+  models/session_event.py). Use append_event() to write and
+  models.session_event.format_event_lines() to render display strings.
 
 Status lifecycle (see models/session_lifecycle.py for canonical mutation functions):
   Non-terminal: pending -> running -> active -> dormant | waiting_for_children | superseded
@@ -836,32 +839,6 @@ class AgentSession(Model):
         if ec_fields and "extra_context" not in kwargs:
             kwargs["extra_context"] = ec_fields
 
-        # Map deprecated field names
-        if "work_item_slug" in kwargs and "slug" not in kwargs:
-            kwargs["slug"] = kwargs.pop("work_item_slug")
-        elif "work_item_slug" in kwargs:
-            kwargs.pop("work_item_slug")
-
-        if "last_activity" in kwargs and "updated_at" not in kwargs:
-            kwargs["updated_at"] = kwargs.pop("last_activity")
-        elif "last_activity" in kwargs:
-            kwargs.pop("last_activity")
-
-        if "scheduled_after" in kwargs and "scheduled_at" not in kwargs:
-            val = kwargs.pop("scheduled_after")
-            if isinstance(val, int | float):
-                kwargs["scheduled_at"] = datetime.fromtimestamp(val, tz=UTC)
-            else:
-                kwargs["scheduled_at"] = val
-        elif "scheduled_after" in kwargs:
-            kwargs.pop("scheduled_after")
-
-        # Map old field names to new ones  # legacy
-        if "parent_job_id" in kwargs and "parent_agent_session_id" not in kwargs:  # legacy
-            kwargs["parent_agent_session_id"] = kwargs.pop("parent_job_id")  # legacy
-        elif "parent_job_id" in kwargs:  # legacy
-            kwargs.pop("parent_job_id")  # legacy
-
         # Schema diet (#1927): watchdog_unhealthy -> unhealthy_reason back-alias.
         if "watchdog_unhealthy" in kwargs and "unhealthy_reason" not in kwargs:
             kwargs["unhealthy_reason"] = kwargs.pop("watchdog_unhealthy")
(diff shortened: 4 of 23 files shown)


valorengels:
## Review: APPROVED

Reviewed at head `0573419028bc4e248a4dc2ee3c7d7d7991baa5c7` (resolved git-first via `git ls-remote refs/pull/2949/head`, not a `gh` read). Blockers: 0. Tech debt: 2 (non-blocking, recorded below).

Scope note: full removal of all nine `_normalize_kwargs()` legacy paths is a settled owner decision, made with the hydration risk stated in front of the decider and knowingly accepted. This review verifies execution and disclosure. It does not re-open the scope call.

---

### 1. Mandatory disclosure — present and accurate

The PR body carries the operative assumption under its own heading ("Required disclosure: the hydration assumption"): *no `AgentSession` rows written before 2026-04-01 survive in Redis, or their legacy field data is acceptable to lose.*

I verified the two mechanism claims against the installed Popoto rather than accepting them:

- `popoto/models/encoding.py:337` is `model_instance = model_class(**model_attrs)`, where `model_attrs` is comprehended directly from the **raw `redis_hash`** items (every key present, minus `\x00`-prefixed internal pointers) — not a declared-field subset. So `query.get` / `get_many` / `async_get` do run `_normalize_kwargs` over whatever legacy keys a row carries, with no Python caller ever naming them. The lazy branch returns at `_create_lazy_model` **before** this line, which is why `filter()` / `all()` (default `lazy=True`) never ran the mapping at all.
- `popoto/models/base.py:518` is `self.__dict__.update(kwargs)`, confirming the corrected pop-list comment: an unknown key lands in the instance dict and never raises, and encoding iterates `_meta.fields` only, so it is never persisted.

"No live Python caller" was therefore genuinely the wrong test, and the PR says so for the right reason. The `parent_job_id` → orphaned-session and `history` → wiped-event-log consequences are both stated. The `Meta.ttl` (30d) point is stated and is the load-bearing one: the TTL never actually fired until #2698 because the corruption sweep full-`save()`d every hydrated row as a validation probe and every `save()` resets the key TTL to the ceiling — so "old rows aged out" is not available as a defence.

**One correction that makes the disclosure conservative rather than optimistic**, so it is worth recording alongside it: for `parent_job_id` specifically, a registered `/update` migration *does* cover the rename. `scripts/migrate_agent_session_keyfield_rename.py` maps `parent_job_id -> parent_agent_session_id` (and `job_id -> id`) directly in Redis, and it is wired into the migration registry as `agent_session_keyfield_rename` (`scripts/update/migrations.py:1166`). It has been running on every `/update`. The PR's "no existing migration covers this field set" is true of the set as a whole, but understates the coverage on the single highest-consequence path. This reduces the residual risk; it does not change the ruling.

### 2. The `hasattr` trap is genuinely closed

Confirmed against `origin/main` that exactly two of the three call sites were guarded, and that both guards are gone:

- `monitoring/session_status.py:82` was `s._get_history_list() if hasattr(s, "_get_history_list") else []` → now an unconditional `format_event_lines(s.session_events)`.
- `monitoring/session_watchdog.py:417` was `if hasattr(session, "_get_history_list"):` → now unconditional.

Had the method been deleted without touching these, both would have silently evaluated to "no history" forever instead of failing loudly. That trap is closed.

`models/session_event.py::format_event_lines` preserves the format byte-for-byte: `f"[{event.get('event_type', 'system')}] {event.get('text', '')}"`, with flat legacy strings passing through and everything else skipped — identical to the removed method's body, including the `"system"` event-type default and the empty-text default. This is load-bearing: `agent/goal_gates.py::check_test_gate` matches on `"[stage]" in entry_lower and "test" in ... and "completed" in ...`. `tests/unit/test_session_event_formatting.py::test_goal_gate_match_shape` pins that exact match, which is the right place to pin it.

### 3. A real bug is fixed here

`commit_sha` and `summary` were the **only two** mapping paths lacking the `"session_events" not in kwargs` guard that every other event-producing path had. Both did:

```python
events = kwargs.get("session_events", []) or []
event = SessionEvent.checkpoint(commit_sha_val)   # / .summary(summary_val)
events.append(event.model_dump())
kwargs["session_events"] = events
```

— an unconditional append onto whatever came out of Redis. Any row carrying both a live `session_events` key and a stale `commit_sha`/`summary` key therefore grew one extra event on **every** `query.get()`, persisting on the next save while the stale hash key survived to do it again. `session_events` has no count cap, so nothing bounded the growth. Deleting these two paths removes unbounded event-log growth on a read path. `tests/unit/test_agent_session_legacy_surface.py::test_no_duplicate_event_growth_on_hydration` covers it directly.

This PR is a net durability improvement, not only a deletion.

### 4. Recon corrections — both hold

Verified against `origin/main`:

- **Three production `get_history_list` callers, not the four the issue claims.** `git grep` on main outside `tests/` returns `agent/goal_gates.py:170`, `monitoring/session_status.py:82`, `monitoring/session_watchdog.py:417-418`, plus the definition itself. `ui/data/sdlc.py:985` used `session.history` (the property), not `get_history_list()` — the issue text names it wrongly. The builder's correction is right and the migration to `.session_events` there is equivalent, since `.history` was a pass-through alias.
- **`stage_states`, `commit_sha`, `summary`, `work_item_slug` remain live properties.** All four are still defined on the model (lines 1545, 1580/1590, 1614/1629, 1658/1662), three with setters that append the appropriate event. Only their `_normalize_kwargs` constructor-kwarg paths were removed. `create_child()`'s explicit `stage_states: dict | None = None` parameter (line 1865) is untouched and still writes its init event.

No shadowing hazard from the new pass-through behaviour: these are properties (data descriptors), so a stray same-named key landing in `self.__dict__` still loses to the class descriptor on attribute lookup.

### Ruling (a): `agent/goal_gates.py` timeout change — **approved as-is**

The bare `timeout=10` → `settings.timeouts.git_subprocess_s` (10s → 60s ceiling) is disclosed, is outside #2873's scope, and is the right call. Three reasons:

1. **The repo already decided this.** The `git_subprocess_s` field description names `gh` CLI calls as members of this exact family and states that the pre-existing spread of `timeout=5/10/30` across the category is "pure copy-paste drift, not deliberate per-site tuning." The `timeout=10` in `goal_gates.py` was precisely that drift. Adopting the field is conformance with a documented repo decision, not a new one.
2. **The alternative was worse.** A new module constant would duplicate a value `settings` already owns, which is the drift the field exists to end — and the inline-timeout-literal commit guard hard-blocks the literal, so leaving it was not an option once the file was touched.
3. **The blast radius is bounded and caught.** Every `_run_gh_command` call site in the file (lines 131, 230, 244, 256, 263) sits inside a `try:` whose handler degrades to a best-effort `GateResult`. A `TimeoutExpired` is caught, not propagated. It is a hang ceiling on advisory gate checks, not added latency on any path.

### Ruling (b): `[WIP]` prefixes — **accept, no reword**

`/do-merge` squash-merges (`gh pr merge {PR} --squash`), which composes the commit message from the PR title and body and discards the branch commit headlines entirely. The prefixes cannot reach `main`. No reword is warranted.

Separately, the builder's handling of the aborted reword was correct: on discovering mid-operation that unrelated files from a concurrent session had appeared in the working tree, aborting and restoring HEAD rather than sweeping up another session's uncommitted work is the right instinct. Those files have since landed on main independently as `d242d7301`.

### Test evidence — verified, not trusted

Re-ran on the branch at `0573419`, serial (`-n 0`), via `scripts/pytest-clean.sh`:

```
tests/unit/test_session_event_formatting.py
tests/unit/test_agent_session_legacy_surface.py
tests/unit/test_goal_gates.py                    -> 74 passed in 9.26s
```

**The two `TestSdkClientEnvVar` failures are confirmed pre-existing, by stronger evidence than the builder's reproduction.** Rather than only re-running them on main, I established it structurally:

- `git diff origin/main...0573419 -- agent/sdk_client.py` is **empty** — the branch's copy of the file under assertion is byte-identical to main.
- The two tests take no fixtures and assert purely on that file's source text.
- `origin/main:agent/sdk_client.py` contains **no occurrence of `VALOR_SESSION_ID` at all**, so `assert 'env["VALOR_SESSION_ID"] = session_id' in source` fails on main by construction.
- The diff to `tests/integration/test_connectivity_gaps.py` touches only the `running_session` fixture and `TestCompleteTranscriptFieldPreservation` — not `TestSdkClientEnvVar`.

The failures are therefore incapable of being caused by this branch. The builder's claim is confirmed.

### Tech debt (non-blocking, not for this PR)

1. **Those two `TestSdkClientEnvVar` tests are rotted, not merely failing.** They assert on the source text of a `VALOR_SESSION_ID` injection that no longer exists anywhere in `agent/sdk_client.py`. They are red on main today and test a string rather than a behaviour. They want either deletion or a rewrite against actual env propagation — as their own issue, since they are unrelated to #2873.
2. **Cosmetic shadowing in `ui/data/sdlc.py:985-986`**: `events = session.session_events` is immediately rebound by `events = _parse_history(history_list)`. Correct, but the reused name makes the two lines read as one operation. Worth a rename on next touch.

---

**Verdict: APPROVED.** Clean deletion with the silent-degradation trap closed, a genuine unbounded-growth bug fixed on a read path, both recon corrections to the issue text confirmed, the hydration assumption disclosed in the body with its mechanism verified against Popoto's actual source, and the one out-of-scope change disclosed and justified by a pre-existing repo decision. Docs updated in the same change. Cleared to merge.

PR #3005, opened 2026-08-25
test: split the remaining four large test files into per-theme packages (Closes #2879)

### LINKED ISSUE #2879: Split the largest test files into per-class modules

valorengels:
Closes #2879

Completes umbrella issue #2879. PR #2941 split two of the six named files and **deliberately deferred these four**, each with a stated hazard. This PR resolves all four hazards and finishes the issue.

Plan: [`docs/plans/split-remaining-large-test-files.md`](https://github.com/tomcounsell/ai/blob/main/docs/plans/split-remaining-large-test-files.md)

## What changed

| Before | After |
|---|---|
| `tests/unit/test_sdlc_session_ensure.py` (3,171 lines) | `tests/unit/sdlc_session_ensure/` — 6 files, 350–764 lines |
| `tests/unit/test_sdlc_router_decision.py` (2,096 lines) | `tests/unit/sdlc_router_decision/` — 5 files, 288–687 lines |
| `tests/unit/test_valor_telegram.py` (2,016 lines) | `tests/unit/valor_telegram/` — 5 files, 176–671 lines (+3 siblings absorbed) |
| `tests/unit/test_worktree_manager.py` (1,610 lines) | `tests/unit/worktree_manager/` — 5 files, 153–462 lines |

Every resulting file is under the issue's ~800-line target (max 764). Grouped by theme rather than strictly per-class, following the convention PR #2941 established — per-class would have produced 37 files averaging 56 lines for `test_sdlc_router_decision.py` alone.

The three existing `test_valor_telegram_*` siblings (`_await`, `_chat_log`, `_voice_flag`) moved into the new package **byte-identical** (verified against `git show HEAD:`), rather than being left as stragglers beside a package named for the same module.

## Resolving #2941's four stated deferral reasons

Each was investigated rather than worked around. Two turned out to be less severe than recorded:

- **`test_sdlc_session_ensure.py` — "real `PipelineLedger` Redis writes at a fixed key."** Real, but confined to **one class**. The autouse `_stub_lane_identity` patches `resolve_lane_slug` to `None` for every class *except* `TestLaneSlugMintedAtLaneStart`, which is the sole writer. Keeping that class whole in one file means no two files ever contend for the key, so `--dist=loadfile` gives exactly the guarantee it gave before. No `xdist_group` marker needed.
- **`test_worktree_manager.py` — "`_git` collides with `tests/unit/conftest.py`."** Not a collision. That `_git` is a plain module-level function used only inside that conftest, not a fixture; there is no fixture-resolution path between them.
- **Depth breakage.** Exactly **three** sites across all four files, only one inside a test body. The two module-header sites (`REPO_ROOT`, `sys.path.insert`) are import-header edits. The one body site is called out below.
- **`test_valor_telegram.py` — autouse `_bypass_promise_gate`.** Hoisted to a package `conftest.py`, but only after confirming none of the three absorbed siblings references `promise_gate` or `cli_check_or_exit` (zero grep matches), so the hoist does not widen its effective scope.

## The hazard this PR adds to the record: `FEATURE_MAP` is order-dependent

PR #2941 established that markers derive from the **basename only**, and that keeping the original basename as a prefix preserves them. That rule is necessary but **not sufficient**, and this PR found the gap.

`pytest_collection_modifyitems` iterates `FEATURE_MAP` in dict-insertion order and **breaks on first hit**. `worktree_manager` sits near the end of that dict, after `config` and `lifecycle`. So:

```
test_worktree_manager_config.py     -> "config" matches first    -> tagged `config`,   not `git`
test_worktree_manager_lifecycle.py  -> "lifecycle" matches first -> tagged `sessions`, not `git`
test_worktree_manager_cleanup.py    -> no earlier key matches    -> tagged `git`       (correct)
```

All three follow the prefix rule; two are silently wrong. Same trap for `..._bridge_short_circuit.py` — `bridge` is the *first* key in the dict, so the file holding `TestBridgeShortCircuit` must not echo its own class name.

Every one of the 21 candidate basenames was run through the real first-hit algorithm **before** the files were written. `tests/README.md` gains this as a worked example, which is the piece its existing "taking the first hit" sentence stated abstractly but never demonstrated.

## Verification

**Marker parity was measured, not assumed.** A collect-only pass over all of `tests/unit` dumped every nodeid with its resolved markers, before and after, via a throwaway `trylast` plugin. Compared as a multiset of `Class::test` identity + markers (invariant under the path change the split necessarily causes), excluding `test_plan_docs.py` — see the note below on why.

```
test-id + marker multiset:  14039 -> 14039,  diff = EMPTY
per-marker histogram:       all 23 markers identical
  sdlc 2577 · parametrize 1915 · asyncio 1167 · unit 1152 · messaging 990 · sessions 937
  reflections 448 · webui 310 · validation 277 · git 179 · classifiers 168 · monitoring 152
  impact 75 · context 71 · models 64 · skipif 60 · config 55 · tools 49 · sdk 43
  usefixtures 39 · slow 10 · xdist_group 3 · integration 2
```

**No test body was changed** — enforced mechanically. A script parsed each pre-split file out of `git HEAD` and every new file with `ast`, then compared `ast.get_source_segment` for all 101 top-level definitions:

```
identical = 100    different = 1    missing = 0
```

The single difference is the one deliberate edit, disclosed in the plan before the build: `parents[2]` -> `parents[3]` in `TestInterpreterPinResolution::test_this_repo_ships_a_committed_pin`. The package sits one directory deeper, so without it `repo_root` resolves to `tests/` and the test fails looking for `.python-version`.

That AST check also caught a real regression during the build: `ruff`'s import autofix had hoisted a function-local `from agent.sdlc_router import DISPATCH_RULES` out of a test body in `TestReviewInProgressNoVerdictDeadEnd`. The local import is deliberate shadowing present in the original, so it was restored and the now-unused header import dropped instead. A collection-count check would never have seen this.

Tests (`scripts/pytest-clean.sh`, never bare pytest):

```
tests/unit/{sdlc_session_ensure,sdlc_router_decision,valor_telegram,worktree_manager}
  -> 443 passed in 30.03s
     (114 + 107 + 115 + 107; the 115 is 97 from the monolith + 18 from the absorbed siblings)
```

`python -m ruff check tests/unit` — All checks passed. `python -m ruff format --check tests/unit` — 675 files already formatted.

## Why the acceptance criterion "total collection count unchanged" is not gated on a literal

The issue asks that `pytest --collect-only` be identical before and after. It is — but the raw total is **not a usable gate**, and it is worth recording why.

`tests/unit/test_plan_docs.py::test_plan_declares_a_tracking_issue` is **parametrized over the files in `docs/plans/`**. Every plan document committed by any lane adds one collected test to `tests/unit`. During this work the count moved from 14065 to 14067 purely because this plan's own document and lane #2875's landed on main — `git log 483c7cd14..8557db75f -- tests/` is empty. A literal gate would fail a mechanically perfect split.

So the parity check is scoped to the multiset excluding `test_plan_docs.py`, which no concurrent lane can perturb, plus an exact per-package count (443). The global total is unchanged at **14067 -> 14067** on this branch, and that is reported as an observation rather than relied on as the gate.

## Scope notes

- **No parallelism claim.** The issue motivates the split partly via `--dist=loadfile`. No timing baseline was taken and none is claimed; wall time is bounded by out-of-scope files.
- **`pyproject.toml` untouched.** Its `--dist=loadfile` comment names three *integration* files that share global resources. None of these four is among them, and the shared-state analysis above shows the split introduces no new cross-file contention, so the comment remains accurate. Editing it would have been a manufactured change.
- **Deliberately not included:** a permanent regression test asserting every `tests/unit/**` basename resolves to its intended marker. The checker exists and proved its worth here, but it is *new test logic*, which #2879's mechanical-reorganization constraint forbids. Flagged here so it can be picked up deliberately.
- No production code is touched.


valorengels:

Lens: code quality. Verified against the PR branch checked out at HEAD `281bf11`.

**Method.** AST source-segment comparison of every top-level class, function, and assignment in each of the four pre-split monoliths against the post-split packages: **zero missing definitions, zero unexpected extras, zero duplicated definitions**, and exactly **one** byte-level body difference across 9,105 added lines — the documented `parents[2]` → `parents[3]` in `TestInterpreterPinResolution`, diffed in isolation and confirmed to be a one-line change and nothing else.

Also confirmed:
- **No hoisted imports.** Module-level import sets diffed old vs. new — no import moved out of a function body. The local `from agent.sdlc_router import DISPATCH_RULES` is intact at `tests/unit/sdlc_router_decision/test_sdlc_router_decision_convergence.py:507`, inside `TestReviewInProgressNoVerdictDeadEnd`, and the module has no top-level `DISPATCH_RULES` import.
- **Depth expressions** all resolve to the repo root at the new depth: `valor_telegram/conftest.py:13` (four `.parent`s), `REPO_ROOT` in both consuming files (one extra `dirname`), `parents[3]`.
- **Siblings** report as `rename ... (100%)` under `git diff --summary` — zero content change. No stranded `tests/unit/test_valor_telegram*.py` remains. All four `__init__.py` present.
- **README** index counts match per-file collection exactly for all 20 rows; all four claims in the new worked example reproduce against the live `FEATURE_MAP`.

### Blockers
- None

### Tech Debt
- **File:** `tests/unit/valor_telegram/conftest.py:16`
  **Code:** `@pytest.fixture(autouse=True)\ndef _bypass_promise_gate(monkeypatch):`
  **Issue:** The plan's stated constraint was that the hoisted fixture reproduce its previous scope exactly, no wider. It is now wider: `test_valor_telegram_chat_log.py:41` and `test_valor_telegram_voice_flag.py:62` call `cmd_send` for real, and `tools/valor_telegram.py:896` invokes the gate unconditionally on every send path. As top-level files those 8 tests previously exercised the real gate.
  **Severity:** tech_debt
  **Fix:** Record the widening as a deliberate, verified decision rather than leaving the fixture docstring's ambiguous "existing tests".

### Nits
- **File:** `tests/unit/sdlc_session_ensure/conftest.py:8`
  **Code:** `REPO_ROOT = os.path.dirname(os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__)))))`
  **Issue:** Nothing in the package reads it; `_adoption.py:12` and `_core.py:10` each define their own, and a conftest module-level constant is not importable by name anyway. `import os` on line 3 exists solely to compute it. Ruff cannot catch it (F401 sees `os` as used).
  **Severity:** nit
  **Fix:** Delete line 3 and lines 8-10.

JUDGE_VERDICT: APPROVED

valorengels:

Lens: silent-failure modes. The way this change is most likely to go wrong is invisibly — tests still collect and pass but are re-tagged, skipped, or lose an isolation guarantee. All four modes come back clean, measured rather than assumed.

**Strongest evidence:** the collected `Class::test` name set for the four new packages is byte-identical to the set from the seven files they replace on `main` (443 vs 443, `diff` empty). Nothing dropped, renamed, duplicated, or shadowed.

**Measured counts** (from the lane worktree, its own `.venv/bin/python`):

| Selection | Expected | Measured |
|---|---:|---:|
| `-m sdlc` over both sdlc packages | 221 | **221** |
| `-m git` over `worktree_manager` | 107 | **107** |
| `-m messaging` over `valor_telegram` | 115 | **115** |
| all four packages, no filter | 443 | **443** |
| baseline on `main` (4 monoliths + 3 siblings) | — | 221 / 107 / 115 |
| `-n 4 --dist=loadfile` full run | — | **443 passed**, exit 0 |

**Marker resolution.** Ran the real first-hit algorithm at `tests/conftest.py:1161-1168` over all 24 new basenames. **24/24 resolve as intended**: 11 `sdlc_*` → `sdlc`, 8 `valor_telegram_*` → `messaging` (via the `telegram` key), 5 `worktree_manager_*` → `git`. `sdlc_router_decision_*` does not trip the earlier `routing` key (`router_` is not `routing`), and no `worktree_manager_*` suffix trips the earlier `config` / `lifecycle` keys the README's new trap section warns about. The algorithm keys on basename only, so package placement cannot re-tag anything.

**Isolation.** The premise that `--dist=loadfile` is what protects the fixed-key `PipelineLedger` writes is weaker than the plan assumes, but in the safe direction: `tests/db_claim.py` claims a Redis db **per process** (`gw2=db3 gw0=db1 gw1=db2 gw3=db4` observed), so two files on two workers cannot contend, and two files on one worker run sequentially. Either way there is no concurrent access. The ledger writes remain confined to `TestLaneSlugMintedAtLaneStart` in one file. Real `session:issuelock:{N}` writes appear in two files but with disjoint issue numbers. No hardcoded `/tmp` paths, no `mkdtemp`, no module-level `os.environ` mutation, no module-level singletons across all 24 files.

**Fixture scope.** `_stub_lane_identity` applies to exactly the same test set as before, and its class-name conditional is preserved verbatim. `_bypass_promise_gate` genuinely widens 97 → 115 — see Nits.

### Blockers
- None

### Tech Debt
- None

### Nits
- **File:** `tests/unit/valor_telegram/conftest.py:16`
  **Code:** `monkeypatch.setattr("bridge.promise_gate.cli_check_or_exit", lambda text, transport, session_id: None)`
  **Issue:** On `main` this fixture lived at `tests/unit/test_valor_telegram.py:18` and covered 97 tests; hoisted, it now also covers the 18 tests in the three absorbed siblings, which never had it (`git show main:...` grep for `promise_gate` returns nothing in all three). Two of those files drive `cmd_send`, which calls the gate unconditionally at `tools/valor_telegram.py:894-896`. Practically this is an improvement — it removes a live-LLM dependency from unit tests — and it masks nothing, since none of the 18 asserts on the gate.
  **Severity:** nit
  **Fix:** Record the intentional widening. No code change needed.

- **File:** `tests/unit/sdlc_session_ensure/conftest.py:22`
  **Code:** `every other test in this file it is noise: it adds a ``touch_issue_lock``
  **Issue:** The fixture is now package-scoped across six files; "this file" is carried over from the monolith and misdescribes the scope to the next reader.
  **Severity:** nit
  **Fix:** "every other test in this package".

JUDGE_VERDICT: APPROVED

valorengels:
## Review: CHANGES REQUESTED

Multi-judge consensus (`code-quality`, `risk`), aggregated under `any-blocker-wins`. **Both judges returned APPROVED with no blockers.** The split itself is mechanically sound and independently proven. The findings below are all in the *record around* the change — stale references in `docs/features/`, an inaccurate claim in the PR body, and a deferral with no surviving home — not in the split.

Reviewing self-authored PR, posted as a comment rather than a formal review.

### What I reproduced myself

Per this repo's "a number the PR claims is a claim, not evidence" rule, I re-ran the load-bearing numbers in my own environment rather than crediting the PR body:

| Check | Result |
|---|---|
| Plan `## Verification` table, all 12 rows | **12/12 PASS** |
| Four packages under `scripts/pytest-clean.sh` | **443 passed in 27.97s** |
| `-m sdlc` / `-m git` / `-m messaging` scoped counts | **221 / 107 / 115** — as claimed |
| FEATURE_MAP first-hit over all 24 new basenames | **24/24** resolve to the intended marker |
| `ruff check` / `ruff format --check` on `tests/unit` | clean |
| Largest resulting file | 764 lines, under the 800 target |
| Stranded siblings / leftover monoliths | none |

The one number I could **not** reproduce is the global `14039 → 14039` nodeid+marker multiset parity, because the "before" side ceased to exist when the split landed. The plan says as much and deliberately does not gate on it. I am not crediting it as satisfied — but I note that the `risk` judge's independent substitute is stronger evidence than the claim itself: the collected `Class::test` name set for the four packages is byte-identical to the set from the seven files they replace on `main`, and the three scoped marker counts match `main` exactly. That corresponding plan checkbox is left unticked.

The `FEATURE_MAP` ordering-dependence finding is the genuinely valuable thing this PR discovered, and the README worked example is the right place for it. `test_sdlc_router_decision_*` narrowly escaping the earlier `routing` key (`router_` is not `routing`) is a good illustration of how thin the margin is.

## Rubric

- [x] **1. Plan vs. implementation match** — pass — *every Success Criterion I could reproduce is satisfied; the four packages, the 800-line ceiling, the 443 count, and the single documented body edit all check out.*
- [x] **2. New code quality** — pass — *one dead constant (nit below); otherwise the AST comparison shows the move is byte-faithful.*
- [x] **3. Test coverage** — pass — *no new behavior; 443/443 pass and the collected name set is unchanged.*
- [x] **4. Regression risk to existing callers** — pass — *`tests/unit/test_sdlc_router.py:6` was correctly updated; per-marker selection still resolves identically.*
- [x] **5. Data integrity** — n/a — *no schema, no migration, no production code touched.*
- [x] **6. Security** — pass — *no secrets, no new subprocess/eval surface; test-only reorganization.*
- [ ] **7. Documentation accuracy** — **fail** — *four references in `docs/features/` still point at the four deleted file paths. See Tech Debt 1.*
- [ ] **8. PR body accuracy** — **fail** — *the promise-gate hoist claim is contradicted by the code. See Tech Debt 2.*
- [x] **9. Disclosed deferrals** — pass — *both No-Gos are explicit out-of-scope exclusions with sound rationale; the mechanical-reorganization constraint genuinely forbids the new test logic.*
- [x] **10. Follow-up claims verified** — pass — *the PR makes no `filed as #N` claim, so there is no unresolved claim. The routing problem it does have is in Miscellaneous.*

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — PASS — *all reproducible criteria satisfied; the one non-reproducible criterion is left unticked rather than assumed.*
- **2. No-Gos from plan — none violated** — PASS — *no production code, no `pyproject.toml` change, no new fixtures beyond the two relocated, no strict per-class split.*
- **3. New `except Exception` blocks — each has logger/raise/swallow-ok** — N/A — *no new exception handlers; moved bodies retain their originals byte for byte.*
- **4. New integration tests — exercise serialization boundary** — N/A — *no new tests.*
- **5. Plan internal consistency — spike findings match task steps** — PASS — *all four spikes are reflected in the delivered file layout, including the `_make_session` relocation and the mandatory fifth `_post_patch` file.*
- **6. No hardcoded secrets or debug artifacts** — PASS — *none introduced.*
- **7. New public APIs — docstrings present** — N/A — *no new public API.*
- **8. Breaking changes — migration path documented** — N/A — *test-suite-internal; invisible to the running system.*
- **9. Tests added for new behavior** — N/A — *no new behavior by design.*
- **10. Tests cover the failure path** — N/A — *no new code paths.*
- **11. UI changes — screenshot captured** — N/A — *no UI files in the diff; the visual-proof gate is a no-op.*
- **12. Docs updated for user-facing changes** — **FAIL** — *`tests/README.md` is updated and accurate, but four `docs/features/` files still name the deleted paths. Promoted to Tech Debt 1.*

### Blockers
- None

### Tech Debt

**1. Four references in `docs/features/` still point at the deleted test files**

- **File:** `docs/features/read-the-room.md:33`
  **Code:** `` | `tests/unit/test_valor_telegram.py::TestCmdSendRTR` | Path B CLI wiring tests (caller-type gate, flag overrides, verdict branching, fail-open). | ``
  **Issue:** That nodeid no longer resolves. The class now lives at `tests/unit/valor_telegram/test_valor_telegram_rtr.py:60`. What makes this one conclusive rather than arguable is the line directly above it: `tests/unit/output_handler/test_output_handler_filters.py::TestReadTheRoomWiring` — PR #2941 updated that reference **in this same table** when it split `test_output_handler.py`. The precedent for cascading the rename into `docs/features/` is already set in the very file this PR left stale.
  **Severity:** tech_debt
  **Fix:** Repoint to `tests/unit/valor_telegram/test_valor_telegram_rtr.py::TestCmdSendRTR`.

- **File:** `docs/features/telegram-messaging.md:395`
  **Code:** `` | `tests/unit/test_valor_telegram.py` | Test suite | ``
  **Issue:** File does not exist on this branch. Repoint at the package.
  **Severity:** tech_debt
  **Fix:** `tests/unit/valor_telegram/`.

- **File:** `docs/features/sdlc-router-oscillation-guard.md:666`
  **Code:** `` - `tests/unit/test_sdlc_router_decision.py` — pure-function tests for every dispatch rule row (1 through 10b), including `TestReviewInProgressNoVerdictDeadEnd` ``
  **Issue:** File does not exist. Worse than a bare path here: the sentence names `TestReviewInProgressNoVerdictDeadEnd`, which is now specifically in `tests/unit/sdlc_router_decision/test_sdlc_router_decision_convergence.py`, and the rows it describes are now spread across five files. A reader following this pointer to check regression coverage lands nowhere.
  **Severity:** tech_debt
  **Fix:** Repoint at `tests/unit/sdlc_router_decision/`, and name `_convergence.py` for the dead-end class specifically.

- **File:** `docs/features/promise-gate.md:377`
  **Code:** `` * `tests/unit/test_send_message.py`, `test_valor_telegram.py`, `test_valor_email.py` — each adds a `--help` anti-leak test ``
  **Issue:** File does not exist; the `--help` anti-leak test is now in the `valor_telegram/` package.
  **Severity:** tech_debt
  **Fix:** Repoint at the package.

**2. The PR body's promise-gate claim is contradicted by the code**

- **File:** `tests/unit/valor_telegram/conftest.py:16`
  **Code:**
  ```python
  @pytest.fixture(autouse=True)
  def _bypass_promise_gate(monkeypatch):
      """Default-mock the promise gate so existing tests do not call the LLM."""
      monkeypatch.setattr(
          "bridge.promise_gate.cli_check_or_exit",
          lambda text, transport, session_id: None,
      )
  ```
  **Issue:** The PR body states the hoist happened "only after confirming none of the three absorbed siblings references `promise_gate` or `cli_check_or_exit` (zero grep matches), **so the hoist does not widen its effective scope**." The conclusion does not follow from the premise, and it is false. `test_valor_telegram_chat_log.py:41` and `test_valor_telegram_voice_flag.py:62` call `cmd_send` for real, and `tools/valor_telegram.py:896` invokes `cli_check_or_exit` unconditionally on every send path. Those 18 sibling tests are now mocked where they previously hit the gate — which, since `bridge/promise_gate.py` is LLM-first (a Haiku call), means they were previously making live LLM calls from a unit test. So the *outcome* is an improvement and it masks nothing (neither file asserts on the gate). The defect is in the reasoning of record: a grep for the symbol name cannot detect a transitive caller, and the plan's own Risk 2 named this exact case as "the single highest-risk judgment call." Leaving the false claim in place means the next person applying this procedure reuses an unsound test.
  **Severity:** tech_debt
  **Fix:** Correct the PR body to state that the hoist *does* widen scope to the 18 absorbed sibling tests, that the widening was verified harmless (no sibling asserts on the gate), and that it removes a live-Haiku call from those tests. Optionally tighten the fixture docstring, whose "existing tests" is now ambiguous about which set.

### Nits

- **File:** `tests/unit/sdlc_session_ensure/conftest.py:8`
  **Code:**
  ```python
  REPO_ROOT = os.path.dirname(
      os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
  )
  ```
  **Issue:** Dead. Nothing in the package reads it — `_adoption.py:12` and `_core.py:10` each define their own copy, and a conftest module-level constant is not importable by name anyway, so this third copy can never be the one in use. `import os` on line 3 exists solely to compute it. Ruff cannot flag it (F401 sees `os` as used), which is why it survived the lint gate. Both judges found this independently.
  **Severity:** nit
  **Fix:** Delete line 3 and lines 8-10.

- **File:** `tests/unit/sdlc_session_ensure/conftest.py:22`
  **Code:** ``every other test in this file it is noise: it adds a ``touch_issue_lock```
  **Issue:** Carried over verbatim from the monolith, where "this file" was accurate. The fixture is now package-scoped across six files, so the docstring misdescribes exactly the thing a reader consults it for — which tests the stub covers.
  **Severity:** nit
  **Fix:** "every other test in this package".

### Miscellaneous

- **File:** `docs/plans/split-remaining-large-test-files.md:399`
  **Code:** `- [SEPARATE-SLUG #2879] A permanent regression test asserting that every `tests/unit/**` basename resolves to its intended `FEATURE_MAP` marker.`
  **Issue:** This deferral is routed to issue **#2879, which this PR closes**. After merge it points at a closed issue, so the "separate slug" it defers to has no home and no owner. The deferral itself is correctly reasoned — it *is* new test logic, which #2879's mechanical-reorganization constraint forbids, so it genuinely does not belong in this PR. But the PR body's disposition is "Flagged here so it can be picked up deliberately," and once #2879 closes the only record is a merged PR description. This is the guard that spike-1 proved would have caught two silently-mistagged files in this very change; it is the highest-value artifact the work produced and it is the one thing exiting without a tracker. I checked: `gh issue list --search "FEATURE_MAP"` returns only #2879 itself.
  **Severity:** tech_debt
  **Fix:** File a follow-up issue for the basename→marker regression guard before merge and reference it from the PR body, so closing #2879 does not close the finding with it.

- **The other 16 files with >80 test functions** — explicit out-of-scope exclusion with sound rationale. #2879 names six files specifically and this PR finishes those six; the "20 files" line in the issue body is context, not a requirement. No tracking required.
- **No parallelism / timing claim** — explicit exclusion, correctly reasoned: wall time is bounded by out-of-scope integration files, and #2941 set the same precedent. Verified that the PR makes no timing claim anywhere.
- **`pyproject.toml` left untouched** — verified sound. Its `--dist=loadfile` comment names three *integration* files that share global resources; none of these four is among them, and the isolation analysis confirms no new cross-file contention. Editing it would have been a manufactured change.
- **The `FEATURE_MAP` marker-regression guard** — rationale sound, but the tracking destination is not. Escalated to Miscellaneous above rather than accepted here.

### Legacy Cruft Audit
- The dead `REPO_ROOT` + `import os` in `sdlc_session_ensure/conftest.py` is extraction residue (filed as a nit above). No other cruft: no commented-out code, no dual implementations, no parallel-run shims, no dead imports flagged by ruff, and the deleted monoliths are fully removed rather than left as re-export stubs.

### Review Delta
- No prior review exists on this PR — this is the first pass.

## Verdict derivation

Rubric items 7 and 8 are `fail` (both non-critical) → Rule 2 → `CHANGES REQUESTED — Tech Debt`. The Miscellaneous finding is `tech_debt` → Rule 3 → same verdict. **No critical rubric item fails and no judge raised a blocker**, so `any-blocker-wins` does not escalate this. Nothing here questions the correctness of the split; every item is a one- or two-line documentation or tracking fix.

Judges run: 2. Consensus disagreement: false (both APPROVED, no blockers; the aggregate is stricter only because the parent review found the `docs/features/` staleness and the tracking gap, which neither judge's lens covered).

valorengels:
## Post-review addenda

Review returned **APPROVED**, no blockers. Three nits were raised; two are addressed here and one is a deliberate call.

**1. Dead constant in a conftest — fixed.** `tests/unit/sdlc_session_ensure/conftest.py` carried a `REPO_ROOT` definition and an `import os` that existed only to serve it; neither was referenced (the two test files that need `REPO_ROOT` each define their own). `ruff` could not see it, because `os` *was* used — by the dead assignment. Both lines removed; package still 114 passed, ruff clean.

**2. Undisclosed plan deviation on `_iso` — disclosing now.** The plan's §2 table routes the `_iso` helper to `test_sdlc_router_decision_verdict_staleness.py`, because that is where it sits textually in the original (immediately under that section's header comment, L506). It actually shipped in `test_sdlc_router_decision_with_concerns.py`.

The build is right and the plan table was wrong. All 32 `_iso(` call sites are in the with-concerns group; **zero** are in `TestReviewVerdictStaleness` or `TestCritiqueVerdictStaleness`. Leaving it in `_verdict_staleness.py` would have shipped an unused helper there and a `NameError` in `_with_concerns.py`. The builder applied the plan's own stated rule — *helpers travel with their consumers, not with their textual neighbours* — which is the same rule that moved `_make_session` out of `_creation.py` and into `_busy_guards.py`. Noting it so a later reader diffing the plan table against the tree does not mistake it for drift.

**3. Stale references in `docs/archive/plans-completed/` — deliberately left.** Those are historical records of completed work; rewriting their file paths would falsify what those plans actually said at the time. This matches PR #2941's treatment of the same directory. The *live* stale references were repointed: `models/agent_session.py:279` (comment only) and `tests/unit/test_sdlc_router.py:6` (docstring only).

## Independent verification by the reviewer

Worth recording that the reviewer did not take the PR body's numbers on trust. They wrote their own AST comparator and got **106 identical / 2 different / 0 missing** against merge-base `627206c97` — a wider scope than my own check (theirs also compares module-level constant assignments, mine compared only classes and functions). The two differences are the disclosed `parents[2]`→`parents[3]` edit and the `REPO_ROOT` depth change in a module header. They also replayed the real `FEATURE_MAP` first-hit loop over all 24 basenames (all correct; confirmed insertion indices `bridge`=0, `routing`=8, `sdlc`=15, `lifecycle`=32, `config`=45, `worktree_manager`=63), evaluated all three depth expressions at runtime, and re-ran the suite to 443 passed.

valorengels:
## Correction: the promise-gate scope claim in this PR's description is wrong

This PR's description states that hoisting `_bypass_promise_gate` to a package `conftest.py` "does not widen its effective scope." **That is false.** Correcting it here because the body is now permanent record; the durable fix lands in #3012 on the fixture itself.

The claim was derived from grepping the three absorbed sibling files for `promise_gate` / `cli_check_or_exit` and finding zero matches. That probe was too narrow — the path is indirect:

- `tests/unit/valor_telegram/test_valor_telegram_chat_log.py` (5 tests) and `test_valor_telegram_voice_flag.py` (3 tests) call `cmd_send`.
- `tools/valor_telegram.py:896` calls `cli_check_or_exit` unconditionally.
- Those files mock only `resolve_chat` and `_get_redis_connection`.

So **8 tests** that previously ran the real gate are now mocked. Not 18 — `test_valor_telegram_await.py`'s 10 tests target `tools.valor_telegram_await.await_bot_reply` and never reach `cmd_send`, so they were never gated either way.

The outcome is fine and arguably an improvement: those 8 assert relay-payload shape, chat-log recording, and voice-flag/cleanup behavior, never gate behavior, so a live Haiku call on each was incidental cost rather than coverage. But the change did something this description says it didn't, and the plan's own Risk 2 named this exact scenario. The check it prescribed was run; it just used a probe that couldn't see an indirect call.

Also worth recording: while writing the replacement docstring I first claimed `TestValorTelegramPromiseGate` "overrides this fixture where it needs the real thing." It doesn't — that class contains a single argparse `--help` anti-leak assertion and never invokes the gate. `cli_check_or_exit` is covered by `tests/unit/test_promise_gate.py`. Both errors are the same shape: asserting a coverage relationship from a name rather than reading the code.

Nothing else in this description is affected. The AST byte-identity proof, the marker-parity result, and the collection counts were all measured directly and stand.

PR #3046, opened 2026-08-28
Repoint production importers off the agent_session_queue re-export hub (phase 3 of 5)

valorengels:
Phase 3 of #2876. Plan: `docs/plans/sdlc-2876.md`.

**Scope:** repoint the production importers that pull a pure re-export off `agent/agent_session_queue.py`, and repair the bridge boundary guard before this diff hollows it. `agent/agent_session_queue.py` itself is untouched — deleting the re-export block is phase 5.

## Why these files

A name may move only if the hub **imports** it and **never references it** in its own body. Names the hub's body uses keep resolving through the hub after phase 5 and must stay. That partition was re-derived by AST against this PR's merge base, not carried over from the plan's older base — the plan's hub-used count was written against a tree that predates #3007, #3028 and #3034.

The split-statement case is where a mechanical repoint goes wrong, and `worker/__main__.py:536` is the worst of them: nine names move to `agent.session_health` / `agent.session_state` while `_ensure_worker`, `_session_notify_listener`, `register_callbacks` and `request_shutdown` stay on the hub, out of a single import statement. `cleanup_corrupted_agent_sessions` and `_active_workers` look like movers but are referenced by the hub's own body, so they were checked against the hub-used set before being touched.

## The guard this diff would have hollowed

`tests/unit/test_worker_entry.py::test_bridge_has_no_execution_function_imports` asserted `imported_names - allowed_imports`. That difference is one-directional: it catches an import **added** outside the allowlist and is structurally blind to one **removed**.

This PR moves three of the bridge's six allowlisted symbols to `agent.session_revival`, a module the old regex never inspected. The observed set shrinks, the difference stays empty, and the guard goes on passing over half the coverage it used to have.

**Old guard, run against this PR's repointed bridge:**

```
########## OLD guard vs POST-phase-3 bridge (3 of its 6 allowlisted symbols no longer imported) ##########
1 passed in 7.61s
```

That is the regression this PR exists to prevent shipping.

### New form

Three changes, each fixing a distinct blind spot:

- **keyed by module path** — a symbol that relocates lands in a module the guard still inspects, instead of falling out of coverage;
- **set equality per module** instead of a one-way difference — removal now fails as loudly as addition;
- **AST collection** instead of regex — covers function-local imports (the bridge's are all function-local) and resolves `import x as y` to the original name natively, replacing the hand-rolled `\s+as\s+` split.

### Demonstrated red, both directions

Addition:

```
########## RED PROBE 1: import ADDED outside inventory ##########
[...]

PR #3048, opened 2026-08-28
Repoint test importers off the agent_session_queue re-export hub (phase 4 of 5)

### LINKED ISSUE #3047: Phase 4/5: repoint test importers off the agent_session_queue re-export hub
Phase 4 of #2876. **Scope: `tests/` only.** No production module is touched and the hub's re-export block is unchanged — deleting it is phase 5.

Plan: `docs/plans/sdlc-2876.md`.

## Problem

24 test files import symbols from `agent.agent_session_queue` that the hub does not own. Phase 5 deletes those re-exports, so every one becomes a collection-time `ImportError` unless it names the real owner first.

The mover set is mechanical: a name moves only if the hub **imports** it and **never references it** in its own body. Everything else — names the hub's body uses, and names the hub *defines* — keeps resolving through the hub and must stay.

## Where a mechanical repoint goes wrong

**Split statements.** A single `from agent.agent_session_queue import (...)` often mixes movers with stayers. Absence from a survivor list does **not** imply "move": hub-*defined* names (`DRAIN_TIMEOUT`, `_ensure_worker`, `clone_agent_session_fields`) were never imports at all, so they appear on no import-derived list. Resolving a mixed statement against only one population misroutes them.

**Matcher-keyed tests.** `TestBackwardCompatExports` in `tests/unit/test_steering_mechanism.py` exists specifically to assert the six re-exports are importable from the hub. Repointing it does not break it — it *hollows* it into a tautology that re-tests the owning modules' own exports. A test anchored to the thing under migration is weakened by a repoint, so running the suite afterward proves nothing.

**Vacuous patch targets.** Three tests patch a pure re-export on its **hub** path while the code under test resolves it from the owning module. `patch` rebinds only the hub's attribute, so these intercept nothing today, and phase 5 turns them into `AttributeError` when the attribute disappears.

One of the three is vacuous in a second, worse way: `test_acquire_pop_lock_returns_true_on_redis_failure` patches `_acquire_pop_lock` itself, sets `return_value = True`, calls the mock, and asserts the mock returned `True`. It exercises `unittest.mock`, never the fail-open branch it is named for.

## Solution

Repoint the 24 files statement-by-statement, splitting mixed statements so only movers leave the hub.

- Keep `TestBackwardCompatExports` on the hub. It is a contract test for a contract still in force; phase 5 deletes the class together with the re-exports it asserts.
- Repoint the two mislocated patch targets to the module where the callee actually resolves them.
- Rewrite the fail-open test to raise from the Redis SETNX and call the real function.

## Acceptance Criteria

- [ ] Every test import of a pure re-export names the owning module, except `TestBackwardCompatExports`, which is deliberately retained
- [ ] Mixed statements keep hub-used and hub-defined names on the hub
- [ ] No production file is modified; `agent/agent_session_queue.py` is unchanged
- [ ] Each repaired patch target is demonstrated to intercept, with the hub-path control showing it previously did not
- [ ] The fail-open test fails when the Redis error is removed
- [ ] Ruff check and format clean; all 24 files pass

Refs #2876

valorengels:
Phase 4 of 5 for #2876. Repoints test-side references off the `agent/agent_session_queue.py` re-export hub and onto the modules that own each symbol, so phase 5 can delete the re-exports without breaking the suite.

Touches test files only — no production file is modified. Verify with:

```
git diff --name-only $(git merge-base HEAD origin/main) HEAD | grep -v '^tests/'
```

## What a "mover" is

A symbol is a **pure re-export** when the hub imports it and never references it in its own body. Those are exactly what phase 5 deletes, so those are what this phase repoints. Symbols the hub actually uses — `AgentSession`, `_session_state`, `_pop_agent_session`, `_ensure_worker` — are working imports that survive phase 5 and are deliberately left alone.

The set is derived, never pinned. `tests/unit/test_hub_alias_references.py::_hub_reexports` computes it from the hub's AST; `len(_hub_reexports())` matches the plan's Finding 1.

## Reference classes swept

The plan asks each phase to re-run detection across every class by which a symbol can be reached. This phase covers:

1. **`from agent.agent_session_queue import X`** — the import class.
2. **Patch-target strings** — `patch("agent.agent_session_queue.X")`.
3. **Module-alias attribute access** — `<alias>.X`. **This class was missed in the first round** and was caught in review; see below.
4. **`sys.modules` injection** — a fake module registered under the hub's name.
5. **Untracked vault config** — `reflections.yaml` callables, closed by phase 2.

## The module-alias class, and the guard for it

Review found the original sweep matched the import class but not attribute access through a module alias. Those sites fail two different ways:

- **They break at phase 5.** `<alias>.<re-export>` becomes `AttributeError`.
- **Writes through them are already vacuous.** `from X import y` binds a *copy* onto the importing module at import time, so `<alias>.y = z` rebinds a copy the owning module and its readers never consult.

`tests/integration/test_worker_concurrency.py` held the majority, and they were the second kind: the concurrency ceiling those tests exist to verify was never actually installed. The file already contained its own counter-example — sibling tests reaching `agent.session_state` directly, under a comment explaining why the hub alias is the wrong seam. The hub alias existed there solely to reach `_slot_registry`, so it is replaced outright and the rationale hoisted to the module docstring, where it governs every test rather than the one that happened to hit the bug.

`tests/unit/test_hub_alias_references.py` adds the sweep to the phase's verification so **phase 5 cannot inherit the blind spot**. It is scope-aware: test files routinely rebind a short alias like `q` to a different module per test function, and a module-wide alias table reports false positives on exactly that shape.

An earlier draft of the guard keyed on "imported but not defined here" and flagged `asq._session_state` and `asq.AgentSession`. Both are safe — the hub uses them, and a *module object* reached through an alias exposes live attributes rather than an import-time copy. The narrower Finding-1 rule is the correct one; a guard that fires on safe code is one people learn to route around.

Deriving rather than pinning also keeps it honest through phase 5: when the re-exports go away the derived set empties, and the companion check fails rather than passing vacuously.

### The guard audits itself

Round-2 review found two gaps in the guard's first version. Both are closed, and both are now pinned by tests of the analyzer's own machinery — a guard phase 5 will lean on should not rest on a single reading of its walk.

- **Scope walk.** The first version reached for `ast.walk` on every non-scope child, which descends into a function nested under an `if` / `with` / `try`. That function's bindings leaked outward and its references were evaluated against the enclosing scope, so the verdict flipped on **source ordering** — hiding a real hit in one arrangement and reporting a safe `agent.session_state` line as a violation in the other. Both passes now share one `_own_scope_nodes` helper that stops at scope boundaries, and a scope's bindings are gathered before any reference in it is evaluated, which is what removes the ordering dependence.
- **Bare dotted imports.** Matching `Attribute.value` against `Name` alone never saw `import agent.agent_session_queue` followed by fully-dotted access, which parses as Attribute-of-Attribute. `_resolve_dotted` flattens the chain so every binding form resolves through one path. Relatedly, `import a.b.c` binds only `a`, so it now maps to `a` rather than to `a.b.c` — the wrong module for that name.

### Demonstrated red, per check, each alone

| Check | Injected fault | Result |
|-------|----------------|--------|
| `test_no_module_alias_reaches_a_hub_reexport` | restored `from agent import agent_session_queue as q` at `test_health_check_recovery_finalization.py:952` | `1 failed, 1 passed` — named that exact site |
| `test_hub_reexport_set_is_derivable_and_non_empty` | `HUB_PATH` pointed at a hub with the re-exports pruned (simulated post-phase-5) | `1 failed, 1 passed` |
| `test_scope_walk_does_not_depend_on_source_ordering` | previous implementation restored | failed with the reported symptom (`[]`), siblings green |
| `test_bare_dotted_import_form_is_detected` | previous implementation restored | failed with the reported symptom (`[]`), siblings green |

Each fired alone with its siblings green, so none is passing as part of a batch.

## Vacuous patch targets fixed

Same root cause, lower stakes — patching a name on the hub that the code under test resolves from its own module:

- `test_recovery_respawn_safety.py` — `_enqueue_nudge` resolves `AgentSession` from `agent/session_executor.py`. The sibling `continuation_agent_session_fields` patch is hub-defined and correctly retained.

  The main-path terminal test was vacuous a *second* way, unrelated to the hub: it asserted `query.filter.assert_not_called()`, but `_enqueue_nudge` never calls `AgentSession.query`, so it asserted against a call that cannot happen and stayed green with the guard removed. It now asserts on `async_create`, as the fallback-path sibling does — with the query returning empty, a session past the guard falls through to recreating the record. Demonstrated red by disabling both terminal guards in `session_executor.py`, which turns the test red where it previously passed.
- `test_zombie_session_resurrection.py` — functions under test live in `session_health`, where in-file siblings already patch correctly.
- `test_silent_failures.py` — `check_revival` emits under `agent.session_revival`, not the hub. The caplog site exercising `_push_agent_session` is left alone: the hub defines that one.

## Reconciling the plan's patch-string count

Plan Finding 3 tables 11 breaking patch-target strings; the real merge-base set in `tests/` is **3**, all fixed here. The other 8 are not patch targets at all, and a phase-5 reader checking the plan's acceptance criterion would otherwise see 3-of-11 and suspect misses:

- `_agent_session_health_check` — its only site is a *docstring example* in `agent/reflection_scheduler.py`.
- `steer_session`, `_write_worker_heartbeat` — zero `.py` sites.
- `cleanup_stale_branches_all_projects` — a legacy migration-map key in `tests/unit/test_migrate_reflections_callables.py` that must be **retained**, since the migration script has to keep recognizing the old dotted path.

The `_pop_agent_session` prediction that did not survive is likewise a non-mover: the hub uses it.

## Deliberately retained

`TestBackwardCompatExports` in `tests/unit/test_steering_mechanism.py` is the one class exempt from this phase's rule. It asserts the re-exports are still importable, which is the proposition phase 5 inverts — so it is a tripwire that fails loudly at phase 5 rather than letting the deletion pass quietly. It now carries a comment recording that, and that phase 5 must delete it in the same diff. Every remaining hub import of a pure re-export in `tests/` lives in this class.

## Also fixed: a phase-3 regression that left `main` red

`tests/unit/test_valor_session_resume_release.py` injected a fake module into `sys.modules` under the hub's name, but phase 3 repointed `tools/valor_session.py` to `agent.session_executor`, so the fake was no longer consulted and the real function ran. This fails identically on the merge base — it is not this PR's damage — but phase 4 is the test-side phase for exactly this class of phase-3 fallout, so it is fixed here rather than deferred past phase 5's gate. A `sys.modules` key is a reference the import-statement and patch-string sweeps both miss.

## Concurrency-test hygiene

Now that `test_worker_concurrency.py`'s registries are live rather than vacuous, two follow-through changes:

- Each `finally` cancels its in-flight worker tasks **before** restoring the slot registry, so a still-running task cannot acquire against the restored value.
- The ceilings that are headroom rather than assertions are marked as such. Their test subjects are serialization and worktree parallelism, not the cap; the genuine ceiling coverage lives in the two tests that already reached for `agent.session_state` directly.

## Deferred deliberately

`_active_workers` / `_active_events` are patched on the hub in `test_zombie_session_resurrection.py` while `agent/session_health.py` binds both from `agent.session_state`, so those patches cannot reach the function under test. Both are **non-movers** — the hub uses them, so they survive phase 5 and this guard rightly ignores them. That makes it a live coverage gap rather than a phase-4 or phase-5 hazard, and it is filed as **#3055** with the falsification transcript, rather than widened into this PR.

## Verification

```
python -m ruff check . && python -m ruff format --check .
python -c "import agent, worker.__main__, tools.valor_session, tools.agent_session_scheduler, bridge.telegram_bridge"
scripts/pytest-clean.sh tests/unit/test_hub_alias_references.py -n0
```

Plan rows marked `[P5]` remain red by design; they are phase 5's to close.

Closes #3047
Refs #2876

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -140,6 +140,8 @@ Check counts with: `pytest -m <marker> --collect-only -q`
 
 When a test patches a symbol, patch the **canonical module that owns the symbol**, not the shim that re-exports it. After PR #1023 split `agent/agent_session_queue.py` into purpose-specific modules (`session_health`, `session_completion`, `session_executor`, `branch_manager`, etc.), tests that still patched `agent.agent_session_queue.<X>` silently no-op'd because the new modules import helpers via direct paths (`from agent.session_executor import steer_session as _steer_session`). The shim keeps re-exports for type checkers and editor navigation, but patch targets must hit the runtime import site. See #1041 and the post-mortem in its plan for details.
 
+Attribute access through a module alias — `_queue.<X>`, where `_queue` is the hub module object — is the same mistake in a second syntax, and a write through it (`_queue.X = fake`) rebinds only the hub's copy, so the owning module never sees it. `tests/unit/test_hub_alias_references.py` guards both forms. It derives the hazard set from the hub's own AST (a name the hub imports and never references in its own body) rather than pinning a list, so it stays honest as the hub changes.
+
 ## Test-Reliability Layers
 
 - **PR-branch flaky filter** (`/do-test`, PR #484, issue #476) — when a test fails on the PR branch, pytest retries the failure once; tests that pass on retry are dropped from the failure report. This layer addresses flakiness *on the PR branch*.
@@ -299,6 +301,7 @@ tests/
 | unit | `test_validate_commit_message.py` | 16 | Commit message format |
 | unit | `test_validate_sdlc_on_stop.py` | 12 | SDLC stop validation |
 | unit | `test_build_validation.py` | 6 | Build process validation |
+| unit | `test_hub_alias_references.py` | 5 | Module-alias reference guard (#2876): no tracked Python reaches an `agent.agent_session_queue` pure re-export through a module alias. Hazard set derived from the hub's AST, not pinned; discovery via `git ls-files`, no exemption set |
 | unit | `test_stale_reference_sweep.py` | 3 | Stale-prose sweep (#2853, #2839): the unregistered reflection name, the two deleted granite package paths, and a scoped cadence-wording anti-criterion. Enumerates via `git ls-files` (untracked scratch Markdown is not repo content) and assembles every compared token by concatenation so the file cannot trip its own sweeps |
 | unit | `test_site_graph_consistency.py` | 2 | Public-site knowledge-graph staleness (#2531): every `data-files` chip reference resolves to a `graph.js` node; frameworks named by the graph are still declared dependencies |
 
--- a/tests/e2e/test_nudge_loop.py
+++ b/tests/e2e/test_nudge_loop.py
@@ -12,7 +12,7 @@ from unittest.mock import AsyncMock, MagicMock
 
 import pytest
 
-from agent.agent_session_queue import MAX_NUDGE_COUNT, SendToChatResult
+from agent.output_router import MAX_NUDGE_COUNT, SendToChatResult
 from models.agent_session import AgentSession
 
 
--- a/tests/integration/test_agent_session_health_monitor.py
+++ b/tests/integration/test_agent_session_health_monitor.py
@@ -249,7 +249,7 @@ class TestJobHealthCheck:
     @pytest.mark.asyncio
     async def test_recovers_job_with_dead_worker(self):
         """A running session whose worker task is done should be recovered."""
-        from agent.agent_session_queue import _agent_session_health_check
+        from agent.session_health import _agent_session_health_check
 
         # Create a running session that has been running long enough
         _create_test_session(
@@ -277,7 +277,7 @@ class TestJobHealthCheck:
     @pytest.mark.asyncio
     async def test_recovers_job_with_no_worker(self):
         """A running session with no entry in _active_workers should be recovered."""
-        from agent.agent_session_queue import _agent_session_health_check
+        from agent.session_health import _agent_session_health_check
 
         _create_test_session(
             status="running",
@@ -300,7 +300,7 @@ class TestJobHealthCheck:
     @pytest.mark.asyncio
     async def test_skips_job_with_alive_worker_under_timeout(self):
         """A running session with an alive worker under timeout should NOT be recovered."""
-        from agent.agent_session_queue import _agent_session_health_check
+        from agent.session_health import _agent_session_health_check
 
         _create_test_session(
             status="running",
@@ -327,7 +327,7 @@ class TestJobHealthCheck:
     async def test_long_running_session_with_fresh_heartbeat_survives(self):
         """Issue #1172: a session with a fresh heartbeat is NOT killed regardless
         of wall-clock duration. The previous wall-clock cap is gone."""
-        from agent.agent_session_queue import _agent_session_health_check
+        from agent.session_health import _agent_session_health_check
 
         # Simulate a 4-hour-old session — far beyond any prior wall-clock cap.
         s = _create_test_session(
@@ -357,7 +357,7 @@ class TestJobHealthCheck:
     @pytest.mark.asyncio
     async def test_skips_recently_started_job_with_dead_worker(self):
         """Sessions running < AGENT_SESSION_HEALTH_MIN_RUNNING not recovered (race guard)."""
-        from agent.agent_session_queue import _agent_session_health_check
+        from agent.session_health import _agent_session_health_check
 
         _create_test_session(
             status="running",
@@ -381,7 +381,7 @@ class TestJobHealthCheck:
     @pytest.mark.asyncio
     async def test_handles_job_without_started_at(self):
         """Jobs without started_at (legacy) should still be checked for dead workers."""
-        from agent.agent_session_queue import _agent_session_health_check
+        from agent.session_health import _agent_session_health_check
 
         # A running session with no started_at (legacy session that predates this field)
         _create_test_session(
@@ -406,7 +406,7 @@ class TestJobHealthCheck:
     @pytest.mark.asyncio
     async def test_no_running_jobs_is_noop(self):
         """When no running sessions exist, health check should do nothing."""
-        from agent.agent_session_queue import _agent_session_health_check
+        from agent.session_health import _agent_session_health_check
 
         # Create only a pending session
         _create_test_session(status="pending")
@@ -439,9 +439,9 @@ class TestJobHealthCheck:
         """
         import logging
 
-        from agent.agent_session_queue import (
+        from agent.agent_session_queue import _active_workers
+        from agent.session_health import (
             AGENT_SESSION_HEALTH_MIN_RUNNING,
-            _active_workers,
             _agent_session_health_check,
         )
 
@@ -525,7 +525,7 @@ class TestJobHealthCheck:
         _delivery_belongs_to_current_run instead of before it, this test
         would fail with the session incorrectly finalized to "completed".
         """
-        from agent.agent_session_queue import (
+        from agent.session_health import (
             AGENT_SESSION_HEALTH_MIN_RUNNING,
             _agent_session_health_check,
         )
@@ -568,7 +568,7 @@ class TestJobHealthCheck:
         aged past the orphan threshold with a 'local'-prefixed worker_key (so
         it WOULD hit the orphaned-local-pending abandon branch), must NOT be
         abandoned by the health check's PENDING loop."""
-        from agent.agent_session_queue import (
+        from agent.session_health import (
             AGENT_SESSION_HEALTH_MIN_RUNNING,
             _agent_session_health_check,
         )
@@ -603,7 +603,7 @@ class TestJobHealthConstants:
 
     def test_constants_exist(self):
         """Health check constants that survived the #1172 simplification."""
-        from agent.agent_session_queue import (
+        from agent.session_health import (
             AGENT_SESSION_HEALTH_CHECK_INTERVAL,
             AGENT_SESSION_HEALTH_MIN_RUNNING,
         )
--- a/tests/integration/test_agent_session_queue_race.py
+++ b/tests/integration/test_agent_session_queue_race.py
@@ -20,10 +20,10 @@ from agent.agent_session_queue import (
     _ensure_worker,
     _pop_agent_session,
     _pop_agent_session_with_fallback,
-    _recover_interrupted_agent_sessions_startup,
     _starting_workers,
     clone_agent_session_fields,
 )
+from agent.session_health import _recover_interrupted_agent_sessions_startup
 from models.agent_session import AgentSession
 
 
--- a/tests/integration/test_bug_a_worker_future_leak.py
+++ b/tests/integration/test_bug_a_worker_future_leak.py
@@ -370,7 +370,7 @@ class TestBugACancelledErrorContract:
         """
         from datetime import timedelta
 
-        from agent.agent_session_queue import _recover_interrupted_agent_sessions_startup
+        from agent.session_health import _recover_interrupted_agent_sessions_startup
 
         session_id = "bug-a-cancelled-recovery-001"
         old_started = datetime.now(tz=UTC) - timedelta(seconds=600)
--- a/tests/integration/test_pm_long_run_no_kill.py
+++ b/tests/integration/test_pm_long_run_no_kill.py
@@ -22,10 +22,8 @@ from datetime import UTC, datetime, timedelta
 
 import pytest
 
-from agent.agent_session_queue import (
-    _active_workers,
-    _agent_session_health_check,
-)
+from agent.agent_session_queue import _active_workers
+from agent.session_health import _agent_session_health_check
 from models.agent_session import AgentSession
 
 
--- a/tests/integration/test_session_heartbeat_progress.py
+++ b/tests/integration/test_session_heartbeat_progress.py
@@ -23,10 +23,10 @@ from datetime import UTC, datetime, timedelta
 
 import pytest
 
-from agent.agent_session_queue import (
+from agent.agent_session_queue import _active_sessions
+from agent.session_health import (
     HEARTBEAT_FRESHNESS_WINDOW,
     MAX_RECOVERY_ATTEMPTS,
-    _active_sessions,
     _has_progress,
     _tier2_reprieve_signal,
 )
--- a/tests/integration/test_session_zombie_health_check.py
+++ b/tests/integration/test_session_zombie_health_check.py
@@ -66,7 +66,7 @@ class TestHealthCheckOrphanFixPreservesStatus:
     ):
         """A completed child with an orphaned parent must remain completed
         after the health check clears the parent reference."""
-        from agent.agent_session_queue import _agent_session_hierarchy_health_check
+        from agent.session_health import _agent_session_hierarchy_health_check
 
         child = completed_child_with_orphaned_parent
         original_session_id = child.session_id
@@ -96,7 +96,7 @@ class TestHealthCheckOrphanFixPreservesStatus:
     ):
         """A failed child with an orphaned parent must remain failed
         after the health check clears the parent reference."""
-        from agent.agent_session_queue import _agent_session_hierarchy_health_check
+        from agent.session_health import _agent_session_hierarchy_health_check
 
         child = failed_child_with_orphaned_parent
         original_session_id = child.session_id
@@ -119,7 +119,7 @@ class TestHealthCheckOrphanFixPreservesStatus:
     async def test_orphaned_parent_reference_is_cleared(self, completed_child_with_orphaned_parent):
         """The health check must clear the parent_agent_session_id when the
         parent no longer exists, allowing the child to stand alone."""
-        from agent.agent_session_queue import _agent_session_hierarchy_health_check
+        from agent.session_health import _agent_session_hierarchy_health_check
 
         child = completed_child_with_orphaned_parent
 
--- a/tests/integration/test_silent_failures.py
+++ b/tests/integration/test_silent_failures.py
@@ -158,7 +158,7 @@ class TestEnqueueContinuationSessionLookupLogging:
         from unittest.mock import AsyncMock as _AsyncMock
 
         with (
-            caplog.at_level(logging.ERROR, logger="agent.agent_session_queue"),
+            caplog.at_level(logging.ERROR, logger="agent.session_executor"),
             patch(
                 "agent.agent_session_queue.enqueue_agent_session",
                 new_callable=_AsyncMock,
@@ -265,7 +265,7 @@ class TestCheckRevivalBranchLogging:
         with (
             patch("agent.session_revival._load_cooldowns", return_value={}),
             patch("subprocess.run", side_effect=Exception("git not found")),
-            caplog.at_level(logging.WARNING, logger="agent.agent_session_queue"),
+            caplog.at_level(logging.WARNING, logger="agent.session_revival"),
         ):
             result = check_revival(
                 project_key="test-revival-project",
--- a/tests/integration/test_silent_failures.py
+++ b/tests/integration/test_silent_failures.py
@@ -164,7 +164,7 @@ class TestEnqueueContinuationSessionLookupLogging:
                 new_callable=_AsyncMock,
             ),
         ):
-            from agent.agent_session_queue import _enqueue_nudge
+            from agent.session_executor import _enqueue_nudge
 
             await _enqueue_nudge(
                 session=mock_session_entry,
@@ -243,7 +243,7 @@ class TestCheckRevivalBranchLogging:
         """When subprocess fails checking branch existence, a warning is emitted."""
         import time as _time
 
-        from agent.agent_session_queue import check_revival
+        from agent.session_revival import check_revival
         from models.agent_session import AgentSession
 
         # Create a session in Redis that belongs to this chat so the branch
@@ -263,7 +263,7 @@ class TestCheckRevivalBranchLogging:
         )
 
         with (
-            patch("agent.agent_session_queue._load_cooldowns", return_value={}),
+            patch("agent.session_revival._load_cooldowns", return_value={}),
             patch("subprocess.run", side_effect=Exception("git not found")),
             caplog.at_level(logging.WARNING, logger="agent.agent_session_queue"),
         ):
--- a/tests/integration/test_stage_aware_auto_continue.py
+++ b/tests/integration/test_stage_aware_auto_continue.py
@@ -10,7 +10,7 @@ Tests use Redis db=1 via the autouse redis_test_db fixture in conftest.py.
 import json
 
 # claude_agent_sdk mock is centralized in conftest.py
-from agent.agent_session_queue import MAX_NUDGE_COUNT
+from agent.output_router import MAX_NUDGE_COUNT
 from models.agent_session import AgentSession
 
 
--- a/tests/integration/test_worker_concurrency.py
+++ b/tests/integration/test_worker_concurrency.py
@@ -285,6 +285,8 @@ class TestPerChatSerialization:
 
         original_semaphore = _ss._slot_registry
         try:
+            # Headroom, not an assertion: this test's subject is per-chat serialization,
+            # so the ceiling is deliberately set above what the test can reach.
             _ss._slot_registry = SlotLeaseRegistry(3)  # Global ceiling
 
             with patch("agent.agent_session_queue._execute_agent_session", new=fake_execute):
@@ -300,11 +302,13 @@ class TestPerChatSerialization:
                 f"Expected 3 sessions to execute, got {len(execution_order)}: {execution_order}"
             )
         finally:
-            _ss._slot_registry = original_semaphore
+            # Cancel the in-flight workers before restoring the registry: a task
+            # still running could otherwise acquire against the restored value.
             task = _active_workers.pop(chat_id, None)
             if task:
                 task.cancel()
             _starting_workers.discard(chat_id)
+            _ss._slot_registry = original_semaphore
 
     @pytest.mark.asyncio
     async def test_global_ceiling_across_multiple_chat_ids(self):
@@ -359,12 +363,14 @@ class TestPerChatSerialization:
                 "hold time."
             )
         finally:
-            _ss._slot_registry = original_semaphore
+            # Cancel the in-flight workers before restoring the registry: a task
+            # still running could otherwise acquire against the restored value.
             for cid in chat_ids:
                 task = _active_workers.pop(cid, None)
                 if task:
                     task.cancel()
                 _starting_workers.discard(cid)
+            _ss._slot_registry = original_semaphore
 
 
 class TestPMProjectKeySerialization:
@@ -399,6 +405,8 @@ class TestPMProjectKeySerialization:
 
         original_semaphore = _ss._slot_registry
         try:
+            # Headroom, not an assertion: this test's subject is project-keyed serialization,
+            # so the ceiling is deliberately set above what the test can reach.
             _ss._slot_registry = SlotLeaseRegistry(5)
             with patch("agent.agent_session_queue._execute_agent_session", new=fake_execute):
                 # Both PM sessions should route to the same project-keyed worker
@@ -410,11 +418,13 @@ class TestPMProjectKeySerialization:
                 "Project-keyed serialization is broken."
             )
         finally:
-            _ss._slot_registry = original_semaphore
+            # Cancel the in-flight workers before restoring the registry: a task
+            # still running could otherwise acquire against the restored value.
             task = _active_workers.pop(project_key, None)
             if task:
                 task.cancel()
             _starting_workers.discard(project_key)
+            _ss._slot_registry = original_semaphore
 
 
 class TestDevWorktreeParallelism:
@@ -453,6 +463,8 @@ class TestDevWorktreeParallelism:
 
         original_semaphore = _ss._slot_registry
         try:
+            # Headroom, not an assertion: this test's subject is worktree parallelism,
+            # so the ceiling is deliberately set above what the test can reach.
             _ss._slot_registry = SlotLeaseRegistry(5)
             with patch("agent.agent_session_queue._execute_agent_session", new=fake_execute):
                 # Each slugged dev session gets its own slug-keyed worker.
@@ -469,12 +481,14 @@ class TestDevWorktreeParallelism:
                 "Slugged dev sessions should run in parallel."
             )
         finally:
-            _ss._slot_registry = original_semaphore
+            # Cancel the in-flight workers before restoring the registry: a task
+            # still running could otherwise acquire against the restored value.
             for sl in slugs:
                 task = _active_workers.pop(sl, None)
                 if task:
                     task.cancel()
                 _starting_workers.discard(sl)
+            _ss._slot_registry = original_semaphore
 
     @pytest.mark.asyncio
     async def test_two_slugged_dev_sessions_same_chat_id_execute_concurrently(self):
@@ -513,6 +527,8 @@ class TestDevWorktreeParallelism:
 
         original_semaphore = _ss._slot_registry
         try:
+            # Headroom, not an assertion: this test's subject is slug-keyed parallelism,
+            # so the ceiling is deliberately set above what the test can reach.
             _ss._slot_registry = SlotLeaseRegistry(5)
             with patch("agent.agent_session_queue._execute_agent_session", new=fake_execute):
                 # Each slugged dev session gets its own slug-keyed worker,
@@ -528,12 +544,14 @@ class TestDevWorktreeParallelism:
                 "via slug-keyed workers (issue #1085)."
             )
         finally:
-            _ss._slot_registry = original_semaphore
+            # Cancel the in-flight workers before restoring the registry: a task
+            # still running could otherwise acquire against the restored value.
             for sl in slugs:
                 task = _active_workers.pop(sl, None)
                 if task:
                     task.cancel()
                 _starting_workers.discard(sl)
+            _ss._slot_registry = original_semaphore
 
     @pytest.mark.asyncio
     async def test_slug_keyed_pop_finds_session_by_slug(self):
--- a/tests/integration/test_worker_concurrency.py
+++ b/tests/integration/test_worker_concurrency.py
@@ -8,6 +8,12 @@ Validates:
 5. Dev sessions with slug from different chat_ids run concurrently (chat-keyed workers)
 
 All tests use redis_test_db fixture (autouse=True in conftest.py) for Redis isolation.
+
+The concurrency ceiling lives in `agent.session_state._slot_registry`, and that is
+the only module these tests may set it through. `agent.agent_session_queue` binds
+the name at import time, so assigning through the hub rebinds a copy the runtime
+never reads: the ceiling is silently never installed and the test passes without
+testing anything.
 """
 
 import asyncio
@@ -16,7 +22,7 @@ from unittest.mock import patch
 
 import pytest
 
-import agent.agent_session_queue as _queue
+import agent.session_state as _ss
 from agent.agent_session_queue import (
     _active_workers,
     _ensure_worker,
@@ -126,8 +132,6 @@ class TestGlobalSemaphore:
         _execute_agent_session with a controlled delay. Verifies the global
         semaphore caps peak concurrent executions.
         """
-        import agent.session_state as _ss
-
         max_sessions = 2
         original_semaphore = _ss._slot_registry
 
@@ -187,7 +191,6 @@ class TestGlobalSemaphore:
         from popoto.redis_db import POPOTO_REDIS_DB
 
         import agent.session_health as _sh
-        import agent.session_state as _ss
 
         original = _ss._slot_registry
         try:
@@ -223,9 +226,9 @@ class TestGlobalSemaphore:
         This is the backward-compatible mode before the worker initializes
         the semaphore (e.g., in tests that don't call _run_worker).
         """
-        original_semaphore = _queue._slot_registry
+        original_semaphore = _ss._slot_registry
         try:
-            _queue._slot_registry = None
+            _ss._slot_registry = None
             # Just verify the pop path doesn't crash when semaphore is None
             chat_id = "test-semaphore-none"
             _create_test_session(chat_id=chat_id, session_id="sess-no-sem")
@@ -233,7 +236,7 @@ class TestGlobalSemaphore:
             assert result is not None
             assert result.status == "running"
         finally:
-            _queue._slot_registry = original_semaphore
+            _ss._slot_registry = original_semaphore
 
 
 class TestPerChatSerialization:
@@ -280,9 +283,9 @@ class TestPerChatSerialization:
             async with count_lock:
                 running_count[0] -= 1
 
-        original_semaphore = _queue._slot_registry
+        original_semaphore = _ss._slot_registry
         try:
-            _queue._slot_registry = SlotLeaseRegistry(3)  # Global ceiling
+            _ss._slot_registry = SlotLeaseRegistry(3)  # Global ceiling
 
             with patch("agent.agent_session_queue._execute_agent_session", new=fake_execute):
                 _ensure_worker(chat_id)
@@ -297,7 +300,7 @@ class TestPerChatSerialization:
                 f"Expected 3 sessions to execute, got {len(execution_order)}: {execution_order}"
             )
         finally:
-            _queue._slot_registry = original_semaphore
+            _ss._slot_registry = original_semaphore
             task = _active_workers.pop(chat_id, None)
             if task:
                 task.cancel()
@@ -310,13 +313,7 @@ class TestPerChatSerialization:
         Post-#1029 default is MAX_CONCURRENT_SESSIONS=8. We enqueue 12 sessions
         across distinct chat_ids with a faster-than-ceiling arrival rate and
         verify the semaphore caps peak concurrency at 8.
-
-        The runtime reads the semaphore from agent.session_state, not from
-        agent.agent_session_queue's import-time alias — so the test patches
-        the canonical module.
         """
-        import agent.session_state as _ss
-
         max_sessions = 8
         chat_ids = [f"global-ceil-chat-{i}" for i in range(max_sessions + 4)]
 
@@ -400,9 +397,9 @@ class TestPMProjectKeySerialization:
             async with count_lock:
                 running_count[0] -= 1
 
-        original_semaphore = _queue._slot_registry
+        original_semaphore = _ss._slot_registry
         try:
-            _queue._slot_registry = SlotLeaseRegistry(5)
+            _ss._slot_registry = SlotLeaseRegistry(5)
             with patch("agent.agent_session_queue._execute_agent_session", new=fake_execute):
                 # Both PM sessions should route to the same project-keyed worker
                 _ensure_worker(project_key, is_project_keyed=True)
@@ -413,7 +410,7 @@ class TestPMProjectKeySerialization:
                 "Project-keyed serialization is broken."
             )
         finally:
-            _queue._slot_registry = original_semaphore
+            _ss._slot_registry = original_semaphore
             task = _active_workers.pop(project_key, None)
             if task:
                 task.cancel()
@@ -454,9 +451,9 @@ class TestDevWorktreeParallelism:
             async with count_lock:
                 running_count[0] -= 1
 
-        original_semaphore = _queue._slot_registry
+        original_semaphore = _ss._slot_registry
         try:
-            _queue._slot_registry = SlotLeaseRegistry(5)
+            _ss._slot_registry = SlotLeaseRegistry(5)
             with patch("agent.agent_session_queue._execute_agent_session", new=fake_execute):
                 # Each slugged dev session gets its own slug-keyed worker.
                 # Stagger starts so Worker A can pop its session before Worker B
@@ -472,7 +469,7 @@ class TestDevWorktreeParallelism:
                 "Slugged dev sessions should run in parallel."
             )
         finally:
-            _queue._slot_registry = original_semaphore
+            _ss._slot_registry = original_semaphore
             for sl in slugs:
                 task = _active_workers.pop(sl, None)
                 if task:
@@ -514,9 +511,9 @@ class TestDevWorktreeParallelism:
             async with count_lock:
                 running_count[0] -= 1
 
-        original_semaphore = _queue._slot_registry
+        original_semaphore = _ss._slot_registry
         try:
-            _queue._slot_registry = SlotLeaseRegistry(5)
+            _ss._slot_registry = SlotLeaseRegistry(5)
             with patch("agent.agent_session_queue._execute_agent_session", new=fake_execute):
                 # Each slugged dev session gets its own slug-keyed worker,
                 # regardless of shared chat_id.
@@ -531,7 +528,7 @@ class TestDevWorktreeParallelism:
                 "via slug-keyed workers (issue #1085)."
             )
         finally:
-            _queue._slot_registry = original_semaphore
+            _ss._slot_registry = original_semaphore
             for sl in slugs:
                 task = _active_workers.pop(sl, None)
                 if task:
--- a/tests/integration/test_worker_concurrency.py
+++ b/tests/integration/test_worker_concurrency.py
@@ -56,7 +56,7 @@ class TestPopLockContention:
         This simulates the TOCTOU race scenario: a second worker calling
         _pop_agent_session for the same chat_id while the first is mid-transition.
         """
-        from agent.agent_session_queue import _acquire_pop_lock, _release_pop_lock
+        from agent.session_pickup import _acquire_pop_lock, _release_pop_lock
 
         chat_id = "test-contention-chat"
         _create_test_session(chat_id=chat_id, session_id="session-contention-1")
@@ -78,7 +78,7 @@ class TestPopLockContention:
     @pytest.mark.asyncio
     async def test_pop_succeeds_after_lock_released(self):
         """_pop_agent_session succeeds once the lock is released."""
-        from agent.agent_session_queue import _acquire_pop_lock, _release_pop_lock
+        from agent.session_pickup import _acquire_pop_lock, _release_pop_lock
 
         chat_id = "test-contention-after-release"
         _create_test_session(chat_id=chat_id, session_id="session-after-release-1")
--- a/tests/unit/test_agent_session_hierarchy.py
+++ b/tests/unit/test_agent_session_hierarchy.py
@@ -165,7 +165,7 @@ class TestFinalizeParent:
 
     def test_transition_parent_completed(self):
         """_transition_parent transitions parent to completed via lifecycle module."""
-        from agent.agent_session_queue import _transition_parent
+        from agent.session_completion import _transition_parent
 
         parent = self._make_session(agent_session_id="p1", status="waiting_for_children")
 
@@ -178,7 +178,7 @@ class TestFinalizeParent:
 
     def test_transition_parent_failed(self):
         """_transition_parent transitions parent to failed via lifecycle module."""
-        from agent.agent_session_queue import _transition_parent
+        from agent.session_completion import _transition_parent
 
         parent = self._make_session(agent_session_id="p1", status="waiting_for_children")
 
@@ -190,7 +190,7 @@ class TestFinalizeParent:
 
     def test_transition_parent_waiting_no_completed_at(self):
         """_transition_parent does not set completed_at for non-terminal status."""
-        from agent.agent_session_queue import _transition_parent
+        from agent.session_completion import _transition_parent
 
         parent = self._make_session(agent_session_id="p1", status="running")
         parent.completed_at = None
--- a/tests/unit/test_agent_session_queue.py
+++ b/tests/unit/test_agent_session_queue.py
@@ -20,15 +20,14 @@ import pytest
 
 import agent.agent_session_queue as asq
 from agent.agent_session_queue import (
-    _acquire_pop_lock,
     _complete_agent_session,
     _copyable_agent_session_fields,
     _pop_agent_session,
     _push_agent_session,
-    _release_pop_lock,
     _worker_loop,
     clone_agent_session_fields,
 )
+from agent.session_pickup import _acquire_pop_lock, _release_pop_lock
 from models.agent_session import AgentSession
 from models.session_lifecycle import StatusConflictError
 
@@ -179,11 +178,11 @@ class TestPopLock:
         Failing open preserves backward compatibility: workers continue to
         function without the lock rather than deadlocking when Redis is down.
         """
-        with patch("agent.agent_session_queue._acquire_pop_lock") as mock_acquire:
-            # Simulate the fail-open path by returning True even on error
-            mock_acquire.return_value = True
-            result = mock_acquire("test-chat-id")
-            assert result is True
+        # Make the Redis SETNX raise, then call the real function. Patching
+        # _acquire_pop_lock itself and asserting on the mock's return value
+        # would exercise unittest.mock rather than the fail-open branch.
+        with patch("popoto.redis_db.POPOTO_REDIS_DB.set", side_effect=Exception("redis down")):
+            assert _acquire_pop_lock("test-chat-id") is True
 
     def test_pop_lock_key_is_chat_id_scoped(self):
         """Pop lock key format must be worker:pop_lock:{chat_id}."""
@@ -472,7 +471,7 @@ class TestHealthCheckDeliveryGuard:
             mock_transition = MagicMock()
             lifecycle_mod.transition_status = mock_transition
             try:
-                from agent.agent_session_queue import _agent_session_health_check
+                from agent.session_health import _agent_session_health_check
 
                 await _agent_session_health_check()
             finally:
@@ -518,7 +517,7 @@ class TestHealthCheckDeliveryGuard:
             lifecycle_mod.finalize_session = mock_finalize
             lifecycle_mod.transition_status = mock_transition
             try:
-                from agent.agent_session_queue import _agent_session_health_check
+                from agent.session_health import _agent_session_health_check
 
                 await _agent_session_health_check()
             finally:
@@ -735,7 +734,7 @@ class TestHealthCheckNoProgressRecovery:
             lifecycle_ctx,
         ):
             mock_cls.query.filter.return_value = [session]
-            from agent.agent_session_queue import _agent_session_health_check
+            from agent.session_health import _agent_session_health_check
 
             await _agent_session_health_check()
 
@@ -763,7 +762,7 @@ class TestHealthCheckNoProgressRecovery:
             lifecycle_ctx,
         ):
             mock_cls.query.filter.return_value = [session]
-            from agent.agent_session_queue import _agent_session_health_check
+            from agent.session_health import _agent_session_health_check
 
             await _agent_session_health_check()
 
@@ -824,7 +823,7 @@ class TestHealthCheckNoProgressRecovery:
             lifecycle_ctx,
         ):
             mock_cls.query.filter.return_value = [session]
-            from agent.agent_session_queue import _agent_session_health_check
+            from agent.session_health import _agent_session_health_check
 
             await _agent_session_health_check()
 
@@ -851,7 +850,7 @@ class TestHealthCheckNoProgressRecovery:
             lifecycle_ctx,
         ):
             mock_cls.query.filter.return_value = [session]
-            from agent.agent_session_queue import _agent_session_health_check
+            from agent.session_health import _agent_session_health_check
 
             await _agent_session_health_check()
 
@@ -887,7 +886,7 @@ class TestHealthCheckNoProgressRecovery:
             lifecycle_ctx,
         ):
             mock_cls.query.filter.return_value = [session]
-            from agent.agent_session_queue import _agent_session_health_check
+            from agent.session_health import _agent_session_health_check
 
             await _agent_session_health_check()
 
@@ -984,7 +983,7 @@ class TestHealthCheckNoProgressRecovery:
             _ctx(),
         ):
             mock_cls.query.filter.return_value = [session]
-            from agent.agent_session_queue import _agent_session_health_check
+            from agent.session_health import _agent_session_health_check
 
             await _agent_session_health_check()
 
--- a/tests/unit/test_agent_session_queue_async.py
+++ b/tests/unit/test_agent_session_queue_async.py
@@ -179,7 +179,7 @@ class TestEnqueueContinuationAsyncWrapping:
 
     def test_continuation_filter_uses_to_thread(self, mock_agent_session):
         """The session lookup in _enqueue_nudge should use to_thread."""
-        from agent.agent_session_queue import _enqueue_nudge
+        from agent.session_executor import _enqueue_nudge
 
         # Create a mock agent session
         mock_rj = MagicMock()
--- a/tests/unit/test_agent_session_queue_revival_helper.py
+++ b/tests/unit/test_agent_session_queue_revival_helper.py
@@ -2,7 +2,7 @@
 
 from unittest.mock import patch
 
-from agent.agent_session_queue import maybe_send_revival_prompt
+from agent.session_revival import maybe_send_revival_prompt
 
 
 class TestMaybeSendRevivalPrompt:
--- a/tests/unit/test_agent_session_scheduler_kill.py
+++ b/tests/unit/test_agent_session_scheduler_kill.py
@@ -574,7 +574,7 @@ class TestRecoveryExcludesKilled:
         """
         import time
 
-        from agent.agent_session_queue import (
+        from agent.session_health import (
             AGENT_SESSION_HEALTH_MIN_RUNNING,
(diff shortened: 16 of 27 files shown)


valorengels:

**Verdict: APPROVED** — 0 blockers, 3 tech debt, 2 nits.

### What holds up

- **The `_acquire_pop_lock` rewrite targets the right seam.** `agent/session_pickup.py:124-134` does a function-local `from popoto.redis_db import POPOTO_REDIS_DB` and then calls `POPOTO_REDIS_DB.set(...)`, so patching the attribute on that singleton is re-resolved per call and genuinely drives the `except → return True` branch. No leak risk: `patch` restores on exit, the block wraps one synchronous call, and the raised `set` leaves no lock key behind.
- **Split imports resolved correctly.** Every name left on the hub in a split statement (`_active_workers`, `_active_sessions`, `_starting_workers`, `_pop_agent_session`, `_pop_agent_session_with_fallback`, `_complete_agent_session`, `clone_agent_session_fields`, `_ensure_worker`) is genuinely referenced by the hub's own body. An independent AST census reproduces exactly 40 movers.
- **Zero quoted `agent.agent_session_queue.<mover>` patch strings remain in `tests/`.** The 14 hits in `tests/unit/test_migrate_reflections_callables.py` are reflections-yaml migration-map *keys*, correctly untouched.
- No test lost an assertion in the repoint; docstrings and comments remain accurate.

### Tech debt

**1. `tests/unit/test_recovery_respawn_safety.py:123` and `:145` — `TestEnqueueNudgeTerminalGuard` is vacuous.** Both use `patch("agent.agent_session_queue.AgentSession")`, but `_enqueue_nudge` resolves `AgentSession` from `agent/session_executor.py:31`. Demonstrated: replacing the entry guard with `if False:` in `session_executor.py` left all 7 tests green. `mock_as.query.filter.assert_not_called()` and `mock_as.async_create.assert_not_called()` assert against a mock the code never touches. Aggravating: the PR repointed `_diagnose_missing_session` at line 146 inside the same `with` block, so the file now reads as if its targets had been audited. Fix: `agent.session_executor.AgentSession`.

**2. `tests/unit/test_zombie_session_resurrection.py:72` — same shape.** `@patch("agent.agent_session_queue.AgentSession")` while sibling tests at `:86` and `:100` correctly use `agent.session_health.AgentSession`. Demonstrated: an `assert mock_cls.query.filter.called` probe fails on all 5 parametrizations — `count == 0` holds only because real Redis is empty, so the test would pass even if terminal sessions *were* recovered.

**3. Aliased attribute access was never swept.** The plan (`docs/plans/sdlc-2876.md:401-405`) asks each phase to re-run detection across all five reference classes; only the `ImportFrom` class was covered. See the aggregate review — the risk judge escalated this to a blocker with runtime proof.

### Nits

**4. `tests/integration/test_silent_failures.py:268` and `:161`** — `caplog.at_level(logging.WARNING, logger="agent.agent_session_queue")` sits directly beneath the line the PR repointed to `agent.session_revival` (`:266`); the emitting logger is `agent.session_revival` (`agent/session_revival.py:15`). Assertions still hold via root propagation, so nothing is broken — the names are just stale, inside statements this PR edited.

**5. `tests/unit/test_steering_mechanism.py:53-85`** — the deliberately-retained `TestBackwardCompatExports` docstring asserts the precise proposition phase 5 inverts, with no comment tying it to #2876/phase 5. One marker line would spare phase 5 a judgment call.

Net: the repoint itself is accurate and consistent, and leaves no test weaker than it was. Every issue above is a pre-existing latent defect the sweep passed over, not damage this PR introduced.

valorengels:
## Review (Judge risk):

**Verdict: CHANGES REQUESTED** — 2 blockers, 1 tech debt, 1 nit.

**Method.** Re-derived the 40-name mover set by AST, then simulated phase 5 without touching files: an external pytest plugin that `delattr`s all 40 re-exports from the hub at `pytest_configure`, run against all 92 test files mentioning the hub and differenced against a clean baseline.

### Blockers

**1. `tests/integration/test_worker_concurrency.py` — 15 sites mutate `_queue._slot_registry`, a mover, and are already vacuous.** Lines 226, 228, 236, 283, 285, 300, 403, 405, 416, 457, 459, 475, 517, 519, 534, where `_queue = agent.agent_session_queue` (line 19). `_slot_registry` is in the mover set (owner `agent.session_state`).

No production code reads the hub's binding — every reader uses `agent.session_state`:
- `agent/session_health.py:3069`, `:3722` — `registry = _session_state._slot_registry`
- `agent/agent_session_queue.py:2149` — `registry = _session_state._slot_registry`
- `worker/__main__.py:563` — `_ss._slot_registry = SlotLeaseRegistry(_max_sessions)`

Proven at runtime: assigning `_queue._slot_registry = SlotLeaseRegistry(3)` leaves `agent.session_state._slot_registry` at its `None` default (`agent/session_state.py:90`). So `TestGlobalSemaphore::test_semaphore_none_allows_unlimited_sessions` and the four serialization/parallelism tests never install the ceiling they claim to test — they pass because the real registry is unconditionally `None`. At phase 5 the *read* on line 226 becomes `AttributeError`. This is both an existing coverage regression and a phase-5 break, in a file this PR touched.

**2. `tests/unit/test_health_check_recovery_finalization.py:910, :952` — `inspect.getsource(q.<mover>)` will `AttributeError` at phase 5.**
```
910:  src = inspect.getsource(q._recover_interrupted_agent_sessions_startup)
952:  src = inspect.getsource(q._agent_session_health_check)
```
Both names are movers (`from agent import agent_session_queue as q`). The PR repointed 20+ `from agent.agent_session_queue import` lines in this same file but missed the alias-attribute form. Line 1087's `q._execute_agent_session` is correctly left alone — that name is hub-used, not a mover.

**Common root cause:** the sweep matched `from agent.agent_session_queue import X` but not `<alias>.X`. An AST sweep for `Attribute(value=Name(hub_alias), attr in movers)` returns exactly these 17 sites in these 2 files — the residual set is bounded.

### Tech debt

**3. `tests/unit/test_valor_session_resume_release.py:910, :934` — stale `sys.modules` patch; test is RED on main.** Both patch `{"agent.agent_session_queue": MagicMock(steer_session=mock_steer)}`, but phase 3 repointed `tools/valor_session.py:1281` to `from agent.session_executor import steer_session`, so the patch is inert and the real function runs. `test_cmd_steer_resolves_by_agent_session_id_and_delegates_with_session_id` fails with `assert 1 == 0` / `Error: Session not found: sess-steer`. Not a regression from this PR, but squarely phase 4's charter, and it must be green before phase 5's gate.

### Nit

**4. `agent/reflection_scheduler.py:447`** — docstring example `"agent.agent_session_queue._agent_session_health_check"` names a mover; it documents a dead path after phase 5.

### Verified safe (no action)

- **Race 1 is closed.** `~/Desktop/Valor/reflections.yaml:52,61,71` already name the owning modules. Other machines are covered by `scripts/migrate_reflections_callables.py` (`CALLABLE_MIGRATIONS`) wired into `/update` at `scripts/update/run.py:1223`.
- **No non-test surface resolves a mover through the hub.** Sweep of all 677 non-test hub references found only the migration map (correct source-side key), two comments, and nit 4.
- **No import cycles.** All 13 relevant modules import cleanly standalone, without the hub.
- **No residual `mock.patch("agent.agent_session_queue.<mover>")` string targets** anywhere.
- **`TestBackwardCompatExports` is a genuinely safe tripwire.** All 6 names are movers; it fails loudly at phase 5, exactly as intended. Phase 5's diff must include deleting this class.

valorengels:
## Review:

**CHANGES REQUESTED** — 2 blockers, 4 tech debt, 2 nits.

Multi-judge consensus (`any-blocker-wins`): `code-quality` APPROVED, `risk` CHANGES REQUESTED. Judges disagreed; the risk judge's blockers were independently reproduced before being credited.

### What this PR gets right

This is careful, well-evidenced work, and most of its claims reproduce exactly:

- **Every repoint is correct.** I recomputed the mover set by AST against the merge base — a name moves only if the hub imports it and never references it in its own body — and got exactly **40** names, matching the plan's Finding 1. Every repoint in the diff lands on the right owning module, and every symbol left on the hub (`_active_workers`, `_active_sessions`, `_pop_agent_session`, `_ensure_worker`, `clone_agent_session_fields`, …) is correctly a non-mover.
- **The residual census is exactly `6`**, all in the deliberately-retained `TestBackwardCompatExports`, as the PR body claims.
- **Both "demonstrated red" claims reproduce.** Swapping the `_acquire_pop_lock` raise for an ordinary lock-held return fails with `assert False is True`; the `_load_cooldowns` probe raises on `agent.session_revival` and is silent on the hub-path control. The fail-open rewrite is a genuine improvement over a test that asserted on its own mock.
- **Gates pass:** `ruff check` clean, `ruff format --check` = 1411 files already formatted, and the 24 diffed files give **554 passed** — all reproduced locally, not taken from the description.
- Plan verification rows that are not `[P5]`-deferred all pass (import smoke test, migration-table coverage, migration idempotency, bridge boundary guard). `[P5]` rows are correctly still red.
- Documentation gate: satisfied. Every `## Documentation` item in the plan is marked Phase 5.

### Blockers

**1. `tests/integration/test_worker_concurrency.py` — 15 sites use `_queue._slot_registry`, which is a mover, and are already vacuous.**

Lines 226, 228, 236, 283, 285, 300, 403, 405, 416, 457, 459, 475, 517, 519, 534, where `_queue = agent.agent_session_queue` (line 19).

`_slot_registry` is in the 40-name mover set (owner `agent.session_state`; hub imports it at `:138` and reads it only as `_session_state._slot_registry` at `:2149`). I confirmed every production reader and writer goes through `agent.session_state`, never the hub binding:

```
agent/session_health.py:3069     registry = _session_state._slot_registry
agent/session_health.py:3722     registry = _session_state._slot_registry
agent/agent_session_queue.py:2149 registry = _session_state._slot_registry
worker/__main__.py:563           _ss._slot_registry = SlotLeaseRegistry(_max_sessions)
```

So `_queue._slot_registry = ...` rebinds a copy nothing reads — these tests never install the ceiling they claim to test. The decisive evidence is in **this same file**: `test_global_ceiling_across_multiple_chat_ids` at `:313-317` uses `_ss._slot_registry` under an author-written comment saying exactly why:

> The runtime reads the semaphore from agent.session_state, not from agent.agent_session_queue's import-time alias — so the test patches the canonical module.

Five of the fifteen sites (226, 283, 403, 457, 517) are **reads** (`original_semaphore = _queue._slot_registry`), so phase 5 turns them into `AttributeError`. Phase 4's entire charter is phase-5 survival, and this is a file the PR touched.

**2. `tests/unit/test_health_check_recovery_finalization.py:910` and `:952` — `inspect.getsource(q.<mover>)` breaks at phase 5.**

```python
from agent import agent_session_queue as q
src = inspect.getsource(q._recover_interrupted_agent_sessions_startup)   # :910
src = inspect.getsource(q._agent_session_health_check)                   # :952
```

Both names are movers. The PR repointed 20+ `from agent.agent_session_queue import` statements in this exact file but missed the alias form. (`q._execute_agent_session` at `:1087` is correctly left alone — hub-used, not a mover.)

**Root cause for both:** the sweep matched `from agent.agent_session_queue import X` but not `<alias>.X`. The plan asks each phase to re-run detection across all five reference classes (`docs/plans/sdlc-2876.md:401-405`); only the `ImportFrom` class was covered. An AST sweep for `Attribute(value=Name(hub_alias), attr in movers)` returns exactly **17 sites in these 2 files** — the residual set is bounded, so the fix is small. Please add that sweep to the phase's verification so phase 5 cannot inherit the same blind spot.

### Tech Debt

**3. `tests/unit/test_recovery_respawn_safety.py:123, :145` — `patch("agent.agent_session_queue.AgentSession")` is vacuous.** `_enqueue_nudge` resolves `AgentSession` from its own module-level import at `agent/session_executor.py:31`, and the only hub import inside it is the function-local `continuation_agent_session_fields` (which *is* correctly patched, since the hub defines that one at `:249`). Demonstrated by the code-quality judge: replacing the entry guard with `if False:` leaves all 7 tests green. Not a phase-5 break (`AgentSession` is hub-used, so it survives), but it is the same vacuous-patch class this PR audits, in the very `with` block the PR edited. Fix: `agent.session_executor.AgentSession`.

**4. `tests/unit/test_zombie_session_resurrection.py:72, :134` — same shape, with an in-file control.** `@patch("agent.agent_session_queue.AgentSession")` while sibling tests at `:87` and `:100` in the same class correctly patch `agent.session_health.AgentSession`, and all of them call functions that now live in `session_health`.

**5. `tests/unit/test_valor_session_resume_release.py:910, :934` — stale `sys.modules` hub patch; the test is RED.** Both patch `{"agent.agent_session_queue": MagicMock(steer_session=mock_steer)}`, but phase 3 (#3046) repointed `tools/valor_session.py:1281` to `from agent.session_executor import steer_session`, so the patch is inert and the real function runs. `test_cmd_steer_resolves_by_agent_session_id_and_delegates_with_session_id` fails with `assert 1 == 0` / `Error: Session not found: sess-steer`.

I verified this is **pre-existing** — it fails identically on the merge base `b152b7d3c` — so it is not this PR's damage and does not block on the failing-test gate. But it means `main` is currently red, and phase 4 is the test-side phase for exactly this class of phase-3 fallout. It should be fixed here or tracked explicitly before phase 5's gate.

**6. The PR body does not reconcile the plan's "11 breaking patch strings" down to the 3 actually present.** Plan Finding 3 (`docs/plans/sdlc-2876.md:206-224`) tables 11 breaking patch-target strings across 7 symbols, and phase 4's step list says "repoint exactly the 11 breaking patch strings"; acceptance criterion `:740` requires all 11 to name canonical modules. The real merge-base set in `tests/` is **3** — the ones this PR fixed. The others do not exist as patch targets: `_agent_session_health_check`'s only site is a *docstring example* at `agent/reflection_scheduler.py:447`, `steer_session` and `_write_worker_heartbeat` have zero `.py` sites, and `cleanup_stale_branches_all_projects` is a legacy migration-map key that must be retained. The PR body already records the `_pop_agent_session` prediction that did not survive; the plan's step explicitly asks that gaps like this be "recorded in the PR description", and without it a phase-5 reader checking criterion `:740` sees 3-of-11 and suspects 8 misses.

### Nits

**7. `tests/integration/test_silent_failures.py:268` (and `:161`) — stale caplog logger name.** `caplog.at_level(logging.WARNING, logger="agent.agent_session_queue")` sits one line below the patch this PR repointed to `agent.session_revival`; the emitting logger is `agent.session_revival` (`agent/session_revival.py:15`). The assertion still holds via root propagation, and the test 15 lines above already uses the correct `logger="agent.session_revival"` — so the file is now internally inconsistent.

**8. `tests/unit/test_steering_mechanism.py:53-85`** — `TestBackwardCompatExports` is the one intentional exception to this phase's rule, but carries no comment tying it to #2876 phase 5. One marker line would spare phase 5 a judgment call about a class whose docstring asserts the proposition phase 5 inverts.

### Advisory (not a finding — phase 5's file)

`agent/reflection_scheduler.py:447`'s docstring example names `agent.agent_session_queue._agent_session_health_check`, a mover. It documents a dead path after phase 5. Out of scope here (this PR correctly modifies zero production files); worth folding into phase 5.

### Verified safe

- **Race 1 (untracked vault config) is closed.** `~/Desktop/Valor/reflections.yaml` already names owning modules, and `scripts/migrate_reflections_callables.py` is wired into `/update`. The 14 hub-path strings in `tests/unit/test_migrate_reflections_callables.py` are migration-map *keys* and are correctly retained.
- **No import cycles**, and no non-test surface resolves a mover through the hub.
- **`TestBackwardCompatExports` is a safe tripwire** — it fails loudly at phase 5 rather than silently, which is the intent. Phase 5's diff must delete it alongside the re-exports.

---

The core refactor here is right, and the evidence standard in the PR description is unusually high. What is missing is one reference class: attribute access through a module alias. Sweeping those 17 sites (and adding the sweep to the phase's verification) should close this out.

valorengels:

**Verdict: APPROVED** — 0 blockers, 3 tech debt, 2 nits.

### What holds up

- **Round-1 blocker 1 is fixed and the fix is load-bearing.** Every production reader resolves the ceiling as `_session_state._slot_registry` (`agent/session_health.py:3069,3722`, `agent/agent_session_queue.py:2149`); the hub binds a copy at `:138` and never reads it. Demonstrated probe: with `SlotLeaseRegistry(5)` → `(1)` in a scratch copy of `tests/integration/test_worker_concurrency.py`, both `TestDevWorktreeParallelism` tests go **red** (`peak_running == 1`, expected 2). The ceiling is now genuinely installed. Unmodified file: 12 passed.
- **Ceiling values are all correct headroom — no test flipped meaning.** `(3)` at `:288` guards 3 sessions that serialize per-chat anyway; `(5)` at `:402,:456,:516` guards 2 concurrent each; `(8)` at `:342` for 12 sessions, with an explicit `peak >= 2` anti-vacuity assert. Save/restore is `try/finally` at all 8 sites and restores the pre-test object, so the now-real global cannot leak across tests.
- **Round-1 blocker 2 is fixed correctly.** `_recover_interrupted_agent_sessions_startup` and `_agent_session_health_check` are defined at `agent/session_health.py:1067` and `:4146` — the `inspect.getsource` alias repoint at `tests/unit/test_health_check_recovery_finalization.py:908,950` targets the real owner.
- **`sys.modules` repoint correct.** `tools/valor_session.py:1281` does `from agent.session_executor import steer_session`, so the key at `tests/unit/test_valor_session_resume_release.py:910,934` must be `agent.session_executor`.
- **Phase-5 safety, recomputed independently.** AST pass over the hub yields 40 movers. Residual mover references in `tests/`: **6**, all in `TestBackwardCompatExports` (`tests/unit/test_steering_mechanism.py:57-85`). Zero residual mover patch-strings — the 14 hits in `tests/unit/test_migrate_reflections_callables.py` are deliberately frozen *old* keys of `CALLABLE_MIGRATIONS` (only the `new` side is resolved, `:69`). The surviving `agent_session_queue as q` at `test_health_check_recovery_finalization.py:1085` reads `q._execute_agent_session`, a non-mover.
- **No vacuous new patch targets.** `agent/session_executor.py:15` imports `_diagnose_missing_session` at module level, so that patch fires. `check_revival` resolves `_load_cooldowns` from its own module (`agent/session_revival.py:28,59,140`), so the `test_silent_failures.py:268` repoint actually *un-hollowed* a previously vacuous patch.
- **No assertion lost; one gained.** The pop-lock test replaced a pure tautology (`mock.return_value = True; assert mock(...) is True`) with a real `POPOTO_REDIS_DB.set` raise driving the fail-open branch at `agent/session_pickup.py:131-134`. Genuine improvement.

### Tech Debt

**1. `tests/unit/test_zombie_session_resurrection.py:72,134` — `@patch("agent.agent_session_queue.AgentSession")` never intercepts.** Proof: `with patch('agent.agent_session_queue.AgentSession') as m: agent.session_health.AgentSession is m` → `False` (`session_health` binds it at `:41`). Second proof: rewriting the fixture to a non-terminal `status="running"` still yields `assert count == 0` **passing** — the mocked list is never consulted. Siblings at `:87`/`:100` get this right. `:134` additionally patches `_active_workers`/`_active_events` on the hub while `session_health` imports both from `agent.session_state` at `:38`. Not a phase-5 break (all three are hub-used), but the PR repointed the *import* inside these very tests and left the decorator pointing off the resolution path.

**2. `tests/unit/test_recovery_respawn_safety.py:124,146` — same defect.** `_enqueue_nudge` resolves `AgentSession` from `agent/session_executor.py:31`. `mock_as.query.filter.assert_not_called()` at `:135` asserts about an object no production code touches — it can never fail.

**3. `tests/integration/test_silent_failures.py:268` — half-migrated line pair.** The patch now names `agent.session_revival`, but `caplog.at_level(..., logger="agent.agent_session_queue")` on the line below still names the hub; the emitter is `agent.session_revival`. Passes only via root propagation.

### Nits

**4. `tests/unit/test_steering_mechanism.py:53-85`** — retaining `TestBackwardCompatExports` is the right call, but its docstring says only "for backward compat" with no pointer to #2876/phase 5. One line (`# Deleted together with the re-export block in phase 5 (#2876).`) would stop a future sweep from "fixing" the last 6 intentional references.

**5. `tests/integration/test_worker_concurrency.py:12-16`** — the new module docstring's final clause ("the ceiling is silently never installed and the test passes without testing anything") is present-indicative and reads as describing the file's current state rather than the consequence of the forbidden approach. "would be silently never installed" removes the ambiguity.

valorengels:
## Review (Judge risk):

**Verdict: APPROVED** — 0 blockers, 3 tech debt, 3 nits.

### Phase-5 survival is proven, not asserted

Re-derived the mover set by AST from `agent/agent_session_queue.py` (**40 movers / 35 hub-used**), then simulated phase 5 with a `pytest_configure` plugin that `delattr`s all 40 movers from the hub, and ran it across **all 91** hub-referencing test files (2341 tests) — not just the 25 in the diff.

- Baseline at head: `13 failed, 2328 passed`
- Phase-5 simulated: `19 failed, 2322 passed`
- `comm` diff: the delta is **exactly** the 6 `TestBackwardCompatExports` tests (`tests/unit/test_steering_mechanism.py:56-85`), the deliberately-retained class phase 5 deletes alongside the re-exports. **Zero unintended breaks.**

I checked five reference forms, not four. Beyond the four the PR names, `getattr(queue_module, ...)` exists as a fifth (`tests/unit/test_stale_cleanup.py:83,120,324,378`; `tests/unit/test_update_stale_session_fence.py:120,385`) — an AST scan of `getattr/setattr/hasattr/delattr` against the mover set returns **0** mover sites, so it does not affect phase 5. No `importlib`/`__import__`/string-concatenation form reaches the hub. Scope-aware alias-attribute residual is **0**; a scope-blind scanner flags `test_health_check_recovery_finalization.py:910,952` as residuals, but `:1085` rebinds `q` to the hub in a *different* test — the round-1 fix is correct.

### Both round-1 fixes are real, demonstrated

- **`_slot_registry`:** production reads `_session_state._slot_registry` dynamically (`agent_session_queue.py:2149`, `session_health.py:3069,3722`); hub `:138` is the dead re-export. Instrumented probe on `SlotLeaseRegistry.acquire` shows **56 acquires** and `_slot_registry` restored to `None` at session finish — the ceiling is genuinely installed and the `try/finally` blocks do not leak. All ceilings (3, 5, 5, 5) comfortably exceed demand (3, 2, 2, 2), so no new deadlock. 12 passed on 3 consecutive runs, no cross-contamination.
- **`sys.modules`:** ran `cmd_steer` under both targets. Old hub target → `mock_steer.called: False`; new `agent.session_executor` target → `True`. Since the test asserts `assert_called_once_with(...)`, `main` was genuinely RED here and the fix is necessary.

**Scope discipline confirmed:** 0 production files, 0 lines in `agent/agent_session_queue.py`, `scripts/migrate_reflections_callables.py` and its test untouched (their hub strings are migration-map keys). `ruff check` clean.

**Pre-existing failures:** the 13 baseline failures are all in files this PR does not touch, and they reproduce in isolation *without* `test_worker_concurrency.py` in the run — not cross-contamination from the now-real global.

### Blockers
- None.

### Tech Debt

**1. Vacuous `AgentSession` patches.** `tests/unit/test_recovery_respawn_safety.py:124,146` and `tests/unit/test_zombie_session_resurrection.py:72,134` patch `agent.agent_session_queue.AgentSession`, but the code under test (`agent.session_executor._enqueue_nudge`, `agent.session_health._recover_interrupted_agent_sessions_startup`) imports `AgentSession` from `models.agent_session` directly (`session_executor.py:31`, `session_health.py:41`). Probe: `se.AgentSession is mock -> False`, `sh.AgentSession is mock -> False`. So `assert_not_called()` holds trivially. Not a phase-5 break (`AgentSession` is hub-used). Siblings in the *same class* already use the correct target (`:87,:100`).

**2. Vacuous `_active_workers`/`_active_events` patches**, `tests/unit/test_zombie_session_resurrection.py:129-130`. `session_health.py:38` imports `_active_workers` from `agent.session_state`; rebinding the hub's name with `{}` leaves the real registry visible. Probe: `sh._active_workers is hub._active_workers -> False`, `is ss._active_workers -> True`. The health check under test sees the real global worker registry, not an empty one.

**3. Same bug class, still red on `main`:** `tests/unit/test_stale_cleanup.py:83,120,324,378` and `tests/unit/test_update_stale_session_fence.py:120,385` assign `queue_module._active_workers`, but `scripts/update/run.py:432` reads `from agent.session_state import _active_workers as active_workers_registry`. Confirmed cause of the pre-existing `test_live_worker_skips_session` failure (`test_stale_cleanup.py:227`, `assert 1 == 0`). Outside this PR's mover charter, but it is phase-3 fallout of exactly the kind this PR fixed in `test_valor_session_resume_release.py`.

### Nits

**4. `scripts/update/run.py:437`** still logs `"Could not import _active_workers from agent_session_queue"` after the import moved to `agent.session_state` — stale message, production file, out of scope here.

**5. Correction to the round-1 finding:** "`test_silent_failures.py:60,161,268` use a stale `caplog` logger name" does not fully hold at head. `:161` and `:268` sit beside patches this PR repointed, and `:60` **correctly** keeps `agent.agent_session_queue` because `_push_agent_session` is hub-*defined* (`agent/agent_session_queue.py:266`). Only the `:268` pairing is genuinely inconsistent.

**6.** No `Documentation`/`Update System`/`Agent Integration` obligations fall due in this phase — all are marked Phase 5 in the plan.

valorengels:
## Review:

**APPROVED** — 0 blockers, 0 open tech debt, 3 acknowledged deferrals (tracked in #3054), 3 nits.

Multi-judge consensus (`any-blocker-wins`): `code-quality` APPROVED, `risk` APPROVED. No disagreement this round. Head `35d4db5d5`.

Both round-1 blockers are fixed, and the fixes were verified independently rather than taken from the description.

### Round-1 blockers: closed

**1. `_queue._slot_registry` (15 sites) → `_ss._slot_registry`.** The repoint is correct and load-bearing. Every production reader resolves the ceiling dynamically as `_session_state._slot_registry` (`agent/session_health.py:3069,3722`; `agent/agent_session_queue.py:2149`); the hub's `:138` binding is a dead re-export. Two independent demonstrations:

- Lowering `SlotLeaseRegistry(5)` → `(1)` in a scratch copy turns both `TestDevWorktreeParallelism` tests **red** (`peak_running == 1`, expected 2). Before this commit the same edit changed nothing.
- An instrumented probe on `SlotLeaseRegistry.acquire` records **56 acquires** during the run, and `_slot_registry` is back to `None` at session finish.

So the ceiling these tests exist to verify is now genuinely installed, and the `try/finally` at all 8 sites restores the pre-test object — the now-real global does not leak across tests. Ceilings (3, 5, 5, 5, 8) all exceed their tests' demand (3, 2, 2, 2, 12-capped-at-8), so nothing newly deadlocks. 12 passed on three consecutive runs.

**2. `inspect.getsource(q.<mover>)` → `from agent import session_health as q`.** `_recover_interrupted_agent_sessions_startup` and `_agent_session_health_check` are defined at `agent/session_health.py:1067` and `:4146`. The surviving hub alias at `test_health_check_recovery_finalization.py:1085` reads `q._execute_agent_session`, a non-mover — correctly left alone. (A scope-blind AST scanner flags `:910,:952` as residuals; `q` is rebound function-locally at `:908,:950`, so this is a false positive.)

**3. `sys.modules` hub key → `agent.session_executor`** matches `tools/valor_session.py:1281`. Running `cmd_steer` under both targets: old key → `mock_steer.called: False`, new key → `True`. `main` was genuinely red here; the fix is real and necessary.

### Phase-5 survival: proven

The risk judge simulated phase 5 without editing a file — a `pytest_configure` plugin that `delattr`s all 40 movers from the hub — across **all 91** hub-referencing test files (2341 tests), not just the 25 diffed:

| Run | Result |
|---|---|
| Baseline at head | `13 failed, 2328 passed` |
| Phase-5 simulated | `19 failed, 2322 passed` |

The delta is **exactly** the 6 `TestBackwardCompatExports` tests — the deliberately-retained class phase 5 deletes alongside the re-exports. Zero unintended breaks. The 13 baseline failures are all in untouched files and reproduce in isolation without `test_worker_concurrency.py` in the run, so they are not contamination from the now-real global.

A **fifth** reference form was checked beyond the four the PR body tables: `getattr/setattr/hasattr/delattr` against the hub. It returns 0 mover sites. No `importlib`, `__import__`, or string-concatenation form reaches the hub either.

### PR body claims: all reproduced

Every number in the description was re-measured locally rather than credited:

| Claim | Verified |
|---|---|
| Files changed `25` | yes |
| Production files modified `0` | yes |
| Lines changed in `agent/agent_session_queue.py` | `0` |
| Residual test imports of a pure re-export | `6`, all in `TestBackwardCompatExports` |
| Alias-attribute mover sites remaining | `0` |
| `ruff check .` | `All checks passed!` |
| `ruff format --check .` | `1411 files already formatted` |
| Every file in the diff | `609 passed` |

Plan `## Verification` table: every non-`[P5]` row passes (lint, format, migration-table coverage, migration idempotency, import smoke test, bridge boundary guard). The 4 failing rows are all `[P5]`-tagged and expected red until phase 5 — which is the point of the recorded baseline.

Scope discipline holds: `scripts/migrate_reflections_callables.py` and `tests/unit/test_migrate_reflections_callables.py` are untouched, correctly — their hub-path strings are migration-map **keys** naming the old path and must be retained. The plan's "11 breaking patch strings" → actual 3 reconciliation is now recorded in the PR body, closing round 1's tech-debt item 6.

Documentation gate: satisfied. Every `## Documentation` item in the plan is marked Phase 5, so nothing falls due here.

### Acknowledged deferrals (verified — tracked in #3054)

Round 1 raised three vacuous-patch items that are **not** fixed here. On re-examination they sit outside this PR's charter: they target `AgentSession`, `_active_workers`, and `_active_events`, which the hub *uses in its own body* and therefore are not movers — they survive the phase-5 deletion. They are genuine coverage bugs, but pre-existing ones this PR neither introduced nor was scoped to fix, so blocking on them would loop the PR on work it was never chartered to do. Filed as **#3054** so phase 5 does not inherit them:

1. `tests/unit/test_recovery_respawn_safety.py:124,146` and `tests/unit/test_zombie_session_resurrection.py:72,134` — `patch("agent.agent_session_queue.AgentSession")` intercepts nothing (`session_executor.py:31`, `session_health.py:41` bind it directly). Demonstrated: rewriting the zombie fixture to a non-terminal `status="running"` still passes, and `assert_not_called()` at `:135` can never fail.
2. `tests/unit/test_zombie_session_resurrection.py:129-130` — `_active_workers`/`_active_events` patched on the hub while `session_health.py:38` imports them from `agent.session_state`; the health check sees the real registry.
3. `tests/unit/test_stale_cleanup.py:227` is **currently red on `main`** from this same bug class (`scripts/update/run.py:432` reads `agent.session_state._active_workers`). Same shape as the phase-3 fallout this PR did fix in `test_valor_session_resume_release.py`.

### Nits

1. `tests/unit/test_steering_mechanism.py:53-85` — retaining `TestBackwardCompatExports` is the right call, but the docstring carries no pointer to #2876 phase 5. One line would stop a future sweep from "fixing" the last 6 intentional references.
2. `tests/integration/test_worker_concurrency.py:12-16` — the new module docstring's closing clause is present-indicative and reads as describing the file's current state rather than the consequence of the forbidden approach; "would be silently never installed" removes the ambiguity.
3. `tests/integration/test_silent_failures.py:268` — `caplog.at_level(..., logger="agent.agent_session_queue")` sits one line under the patch this PR repointed to `agent.session_revival`. Passes via root propagation. (`:60` correctly keeps the hub name — `_push_agent_session` is hub-defined at `agent/agent_session_queue.py:266`.)

### Advisory — phase 5's files, not this PR's

- `agent/reflection_scheduler.py:447`'s docstring example names `agent.agent_session_queue._agent_session_health_check`, a mover; it documents a dead path after phase 5.
- `scripts/update/run.py:437` logs "Could not import _active_workers from agent_session_queue" after that import moved to `agent.session_state`.
- The plan's second `## Verification` table is prose-labelled "non-check table — recorded evidence, not executed as checks", but `parse_verification_table` executes it anyway, producing 9 spurious `exit code 127` failures on every run. Since every stage and judge is told to run this table, that noise could mask a genuine failure. Worth restructuring in the plan doc.

---

The alias-attribute sweep that round 1 asked for is complete, and the phase-5 simulation across the whole suite — rather than the diffed files alone — is the right standard of proof for a phase whose entire charter is surviving a future deletion. Good to merge.

valorengels:

**Verdict: APPROVED** — 0 blockers, 2 tech debt, 2 nits.

All round-1 findings verified resolved with demonstrated red.

**Scope confirmed clean.** `git diff --name-only b152b7d3c 05a2e56e2 | grep -v '^tests/'` returns nothing. No test lost an assertion: comparing assertion counts across all 26 touched files merge-base→head, only three moved, all upward (`test_agent_session_queue.py` 150→151, `test_steering_mechanism.py` 51→52, both from added prose containing "assert"; `test_hub_alias_references.py` 0→8, new).

**B1 (`test_worker_concurrency.py`) — fixed, demonstrated.** The decisive experiment: force the ceiling to 1 in all three `TestDevWorktreeParallelism` sites.
- Through the new target `_ss._slot_registry = SlotLeaseRegistry(1)` → `2 failed, 1 passed`. The write is live.
- Through the old `_queue._slot_registry = SlotLeaseRegistry(1)`, same value → `3 passed`. The write was vacuous.

Exactly the round-1 claim, now repaired. The target is also the right one: production reads it only as a module attribute — `agent/session_health.py:3069`, `:3722` and `agent/agent_session_queue.py:2149` all do `registry = _session_state._slot_registry`, and the production writer `worker/__main__.py:563` uses the identical `_ss._slot_registry = SlotLeaseRegistry(_max_sessions)` idiom. Meanwhile the hub's own `from agent.session_state import _slot_registry` at `agent/agent_session_queue.py:138` is a dead import, which is precisely why the guard classifies it as a pure re-export.

**B2 / TD4 (`test_zombie_session_resurrection.py`) — fixed, demonstrated.** Reverting `@patch("agent.session_health.AgentSession")` back to the hub path makes `TestStartupRecoverySkipsTerminalSessions::test_legitimate_running_session_still_recovered` fail with `assert 0 == 1` — the mock supplies the session, so non-interception collapses the count. Green at the new target. `agent/session_health.py:41` binds `AgentSession` module-level, so the target is correct.

**TD3 (`test_recovery_respawn_safety.py:145`) — fixed, demonstrated.** I broke both terminal guards in `agent/session_executor.py` (the entry `return` and the fallback-recreate `return`) and ran `test_nudge_fallback_path_skips_terminal`: FAILED at the new target `agent.session_executor.AgentSession`, PASSED with the guards still broken at the old hub target. The mock now genuinely intercepts `AgentSession.async_create` at `session_executor.py:731`.

**N7 — verified against the actual emitter.** `_enqueue_nudge` is defined at `agent/session_executor.py:635`; that module's `logger = logging.getLogger(__name__)` is at line 34. During the TD3 experiment I directly observed the emission the test asserts on: `ERROR agent.session_executor:session_executor.py:691 [test-project] No session found for test-session-123`. `check_revival` is defined at `agent/session_revival.py:47` with `logger = logging.getLogger(__name__)` at line 15, so `:268`'s `agent.session_revival` is right too.

**TD5 — resolved.** `tests/unit/test_valor_session_resume_release.py`: 55 passed. **N8** — the `TestBackwardCompatExports` docstring now states the phase-5 contract explicitly and correctly frames the class as the tripwire.

**Guard behavior today.** `test_hub_alias_references.py` passes and reports zero sites. I independently re-scanned all tracked Python for string-literal `agent.agent_session_queue.<name>` where `<name>` is in the derived 40-name re-export set: 16 hits, all of them `cleanup_stale_branches_all_projects` as *migration data* in `scripts/migrate_reflections_callables.py:114` and its test, plus a docstring example at `agent/reflection_scheduler.py:447`. No `patch()` target on a pure re-export survives in `tests/`. The reference classes the guard structurally cannot see have zero sites today: no bare `import agent.agent_session_queue` without `asname`, no `getattr()` on a hub alias, no `sys.modules["agent.agent_session_queue"]`.

Conventions from `docs/features/legacy-artifact-guard.md` are followed: discovery is `git ls-files` (`:60-73`) with an explicit non-empty assertion guarding against a broken invocation, no filesystem walk, and no exemption list.

### Blockers
- None

### Tech debt

**1. `_scope_bindings` / `_alias_hits` mishandle a scope nested under a non-scope statement — the guard can fail open, and can also fire on safe code.** `tests/unit/test_hub_alias_references.py:126-136` and `:143-157`.

`_scope_bindings` skips a `_SCOPE_NODES` child, but for every *non-scope* child it does `ast.walk(child)` (line 129), which descends straight through any `FunctionDef` nested inside that child. So an import made inside a function that sits under an `if` / `with` / `try` is hoisted into the enclosing scope's binding table. Symmetrically, `_alias_hits` line 147 walks the same non-scope child and reaches `Attribute` nodes inside that nested function, evaluating them against the outer `env`.

Both directions reproduce. Using the guard's own `_scope_bindings` and `_alias_hits`:

```python
# case_fn.py  — FALSE NEGATIVE
import agent.agent_session_queue as q

def genuine_hub_write():
    q._slot_registry = None          # real hit: vacuous write, AttributeError at phase 5

if True:
    def unrelated():
        import agent.session_state as q
        return q._slot_registry
```
→ `module env: {'q': 'agent.session_state'} | hits: []`. The nested function's import clobbered the module-level alias and the genuine hit vanished.

Reordering the same two blocks (the `if` first) gives `hits: [(6, 'q', '_slot_registry', 'Load'), (13, ...)]` — line 6 is `agent.session_state`, a false positive. The verdict flips on source ordering alone, which is what makes this worth fixing rather than documenting.

This contradicts the docstring at `:50-52` ("a binding made inside one of these shadows the enclosing scope's binding … a module-wide alias table would report false positives") and `:90-91` ("a guard that fires on safe code is one people learn to route around").

Not a blocker: I ran a corrected variant — one that stops descent at scope boundaries in both functions — across all tracked Python and it also finds **0 files**, identical to the shipped guard, while correctly resolving both synthetic cases. So nothing is masked today; this is a latent defect in a guard meant to hold for phase 5 and beyond. The fix is one shared helper that yields a node's own-scope descendants, used by both functions.

**2. `test_nudge_main_path_skips_terminal`'s assertion is vacuous, independently of the patch target.** `tests/unit/test_recovery_respawn_safety.py:123-135`. The comment claims "We patch AgentSession.query to detect any leak", but `_enqueue_nudge` (`agent/session_executor.py:635-780`) contains zero `AgentSession.query` calls — its only module-level `AgentSession` use is `async_create` at `:731`. `grep -c "AgentSession.query"` over that range returns `0`. I broke the entry guard and the test still passed at both the new and the old target.

Pre-existing, not introduced here — `git show b152b7d3c:tests/unit/test_recovery_respawn_safety.py` carries the same `mock_as.query.filter.assert_not_called()` with the same comment. The PR's repoint of the target is correct and I am not asking it to be reverted. But the round-1 finding was "vacuous patch", and this site is still vacuous for a second, unrelated reason. Asserting on `mock_as.async_create` (as the sibling fallback test does, and which I demonstrated goes red) would make it real.

### Nits

**3. Adjacent hub patches left in a decorator stack the PR edited.** `tests/unit/test_zombie_session_resurrection.py:129-130` still reads:
```python
@patch("agent.agent_session_queue._active_workers", {})
@patch("agent.agent_session_queue._active_events", {})
@patch("agent.session_health.AgentSession")
```
Line 131 was repointed; 129-130 were not. These are correctly *out of phase 4's scope* — neither name is in the derived re-export set (the hub references both in its own body), so they survive phase 5 and the guard rightly ignores them. They are nonetheless vacuous for the same structural reason: `agent/session_health.py:38` binds `_active_workers` / `_active_events` from `agent.session_state` directly, so patching the hub's copy cannot affect the function under test. Worth a follow-up issue rather than a change here.

**4. `_hub_reexports` counts any `Name` load as a reference.** `tests/unit/test_hub_alias_references.py:116-117`. A local variable in the hub that happens to shadow an imported name would silently drop that name from the hazard set. No such collision exists today (the derivation yields a sane 40 names, and the three sentinel assertions at `:191-194` hold), so this is theoretical.

valorengels:
## Review (Judge risk):

**Verdict: APPROVED** — 0 blockers, 1 tech debt, 4 nits.

Judged on risk: will phase 5 (deleting the hub's re-exports) break anything this phase was supposed to make safe? Empirically, no.

**Mover set, re-derived independently.** My own AST pass over `agent/agent_session_queue.py` at `05a2e56e2` (imported names minus names loaded anywhere in the hub body, excluding import statements) yields **exactly 40** pure re-exports, and 71 working imports. This is an independent derivation, not a read of the guard, and it agrees with `tests/unit/test_hub_alias_references.py::_hub_reexports` name-for-name. `_session_state`, `AgentSession`, `_execute_agent_session`, `cleanup_corrupted_agent_sessions` correctly land on the *non*-mover side.

**Phase-5 simulation (the real test).** External plugin (never committed) re-deriving the mover set itself and `delattr`-ing all 40 at `pytest_configure`. Run over **all 92 tracked test files mentioning `agent_session_queue`** (`git ls-files 'tests/*.py' | xargs grep -ln`), 2343 tests, diffed against a clean baseline of the same 92 files:

```
[PHASE5-SIM] deleted 40/40 re-exports from agent.agent_session_queue
baseline: 13 failed, 2330 passed
sim:      62 failed, 2281 passed
```
My first sim run was confounded — I had put the plugin on `PYTHONPATH` via `/tmp`, where a stray `/tmp/attr.py` shadowed the `attr` package and broke every subprocess-spawning test (`ImportError: cannot import name 'NOTHING' from 'attr'`). Re-running the six affected files from an isolated plugin dir:

```
7 failed, 251 passed
FAILED tests/integration/test_remote_update.py::TestRestartFlag::test_check_restart_flag_defers_when_jobs_running   <- also in baseline
FAILED tests/unit/test_steering_mechanism.py::TestBackwardCompatExports::  (6 tests)
```
So the **only** phase-5 breakage across the whole hub-touching surface is `TestBackwardCompatExports` at `tests/unit/test_steering_mechanism.py:92-152` — the deliberately-retained tripwire, whose class docstring already states phase 5 must delete it in the same diff. Round 1's two blockers are gone.

**Reference sweep, independent of the guard**, over `git ls-files '*.py'` plus yaml/json/toml, covering from-imports, quoted `"agent.agent_session_queue.X"` strings, `<alias>.X` for every binding form (including bare `import agent.agent_session_queue` with fully-dotted use), `getattr`, and `sys.modules` injection. **Zero residuals.** Everything that survives is safe:
- `agent/reflection_scheduler.py:447` — a docstring example, not resolved.
- `scripts/migrate_reflections_callables.py:111,114` — migration-map *keys*; they must keep naming the old dotted path, that is the input side of the rewrite. `tests/unit/test_migrate_reflections_callables.py` mirrors them as fixture data.
- `tests/unit/test_worker_persistent.py:32` — docstring.
- `tests/unit/test_health_check_recovery_finalization.py:910,952` bind `q` to `agent.session_health`, not the hub (my regex sweep false-positived here; the guard's scope-aware AST walk gets it right). Line 1085's `q._execute_agent_session` is a non-mover and survives.
- Every `patch("agent.agent_session_queue.X")` string in the repo targets a non-mover (`_ensure_worker`, `enqueue_agent_session`, `AgentSession`, `_execute_agent_session`, `DRAIN_TIMEOUT`, …).
- No `sys.modules["agent.agent_session_queue"]` or `import_module` site exists anywhere.

**The newly-real ceiling is genuinely installed — demonstrated both directions.** In `tests/integration/test_worker_concurrency.py:288`, mutating `SlotLeaseRegistry(3)` → `SlotLeaseRegistry(0)`:
```
E   AssertionError: Expected 3 sessions to execute, got 0: []
1 failed in 3.09s
```
Control, reverting the same site to the pre-PR `_queue._slot_registry = SlotLeaseRegistry(0)`:
```
1 passed, 3 warnings in 3.04s
```
That is the vacuity round 1 identified, and its removal, proven. Worktree restored to clean after each mutation (`git status --porcelain` empty, HEAD still `05a2e56e2`).

**Leakage.** Every write the PR repoints sits in `try:` … `finally: _ss._slot_registry = original_semaphore` (lines 136/178, 229/239, 286/303, 340/362, 400/413, 454/472, 514/531), so restore holds on the exception path. `worker/__main__.py:563` sets the same attribute but only inside the worker entrypoint, which these tests never call, so there is no fight. No autouse fixture resets it, but none is needed given the try/finally. Full-suite evidence: the 92-file baseline run has the same 13 failures with or without these tests in the same process, and none are ordering-shaped.

**Previously-RED test is green for the right reason.** `tests/unit/test_valor_session_resume_release.py:910,934` now inject `agent.session_executor`, which matches the real import site `tools/valor_session.py:1281` (`from agent.session_executor import steer_session`). The assertion `mock_steer.assert_called_once_with("sess-steer", "Hold up.")` cannot pass incidentally — if the patch missed, the real function would run and the mock would record nothing. Passes.

**Zero production files.** `git diff --name-only b152b7d3c 05a2e56e2 | grep -v '^tests/'` → empty. 26 files, all under `tests/`. Ruff check and format clean.

### Blockers
- None

### Tech debt

- `tests/unit/test_hub_alias_references.py:150-155` — the guard misses one reference class it reads as if it covers. `_alias_hits` only matches `ast.Attribute` whose `.value` is an `ast.Name`, so a bare `import agent.agent_session_queue` followed by fully-dotted `agent.agent_session_queue.steer_session` parses as nested `Attribute` nodes and is invisible. Demonstrated with a probe file (`git add -N`, then removed): the aliased control `_q.steer_session` fails the guard loudly —
  ```
  E   Failed: Module-alias access to a symbol the hub only re-exports:
  E     tests/unit/test_zz_probe_dotted.py
  E       L5  _q.steer_session  [Load]
  ```
  — while the dotted form `agent.agent_session_queue.steer_session` plus `agent.agent_session_queue._slot_registry = None` yields `2 passed in 7.05s`. Both break identically at phase 5. Relatedly, `_scope_bindings` (line 130) records `alias.asname or alias.name.split(".")[0]`, so a bare `import agent.agent_session_queue` binds `"agent"` → `"agent.agent_session_queue"`, which is simply the wrong module for that name. No such site exists today (my sweep handled this form and found none), so this is a completeness gap in the guard, not a live residual — but the guard's docstring claims to cover "attribute access through a module alias" without qualification. Worth closing before phase 5 relies on it, or narrowing the docstring's claim.

### Nits

- `tests/integration/test_worker_concurrency.py:288,402,456,516` — now that these ceilings are real rather than vacuous, three of the four are non-binding by construction: `SlotLeaseRegistry(3)` with 3 same-chat sessions that already serialize, `SlotLeaseRegistry(5)` with 2. Mutating line 288 from 3 → 1 still passes (`2 passed in 4.78s`); only 0 fails. That is fine for these tests, whose subject is serialization and worktree parallelism, not the cap — the genuine ceiling coverage lives at lines 140 and 342, which already used `_ss` before this PR. No change requested; just noting the ceiling values there are scenery, and a future reader should not mistake them for assertions.
- `tests/integration/test_worker_concurrency.py:303,362,413,472,531` — the `finally` restores `_ss._slot_registry` *before* cancelling the still-running worker tasks, so an in-flight task can briefly acquire against the restored registry. Benign in-process (the restored value is `None`, i.e. no ceiling), but swapping the two statements would be strictly tighter.
- `tests/unit/test_agent_session_queue.py:1494-1500` and `tests/unit/test_teammate_cold_start_finalize.py:422-427` restore `asq._session_state._slot_registry` after a bare `yield` with no `try`/`finally`, so restore is skipped if the test errors. Pre-existing and untouched by this PR, and harmless since the value written is `None`, but it is the same shape the PR just hardened elsewhere.
- Both intentional phase-5 breakages are loud and self-documenting: `tests/unit/test_hub_alias_references.py:180-186` fails with *"No pure re-exports derived from the hub. Either phase 5 has landed (in which case update this test alongside it) or the AST derivation is broken."*, and `TestBackwardCompatExports`'s docstring says phase 5 must delete the class in the same diff. Good failure modes; phase 5 should expect exactly two deliberate red spots and nothing else.

valorengels:
## Review:

**CHANGES REQUESTED** — 0 blockers, 3 tech debt, 3 nits.

Round 2. Multi-judge consensus (`any-blocker-wins`): `code-quality` APPROVED, `risk` APPROVED — **both round-1 blockers are genuinely fixed**, and I reproduced the proof independently rather than reading the diff. No blockers remain. Per this repo's zero-finding approval rule, the residual tech debt keeps this at CHANGES REQUESTED; the remaining work is small and concentrated in one file.

### What round 2 gets right

- **B1 (`test_worker_concurrency.py`) is closed, and the fix is demonstrably live.** The 15 `_queue._slot_registry` sites now write through `agent.session_state`. The risk judge forced `SlotLeaseRegistry(0)` at `:288`: through the new target → `AssertionError: Expected 3 sessions to execute, got 0`; through the old `_queue` target, same value → `1 passed`. That is the vacuity round 1 identified, and its removal, proven in both directions. The `_ss` import is hoisted to module level and the rationale hoisted from one test's docstring to the module docstring, so it now governs the whole file rather than the one test that happened to hit the bug.
- **B2 (`test_health_check_recovery_finalization.py:910, :952`) is closed** — both `inspect.getsource(q.<mover>)` sites now bind `q` to `agent.session_health`, which is where both functions are actually defined. `:1085`'s `q._execute_agent_session` is still correctly left alone (hub-used, not a mover).
- **The new guard works and I reproduced its demonstrated-red.** Re-injecting `from agent import agent_session_queue as q` at `test_health_check_recovery_finalization.py:952` gives `1 failed, 1 passed` with the failure naming the exact site (`L952  q._agent_session_health_check  [Load]`), sibling green. Deriving the hazard set from the hub's AST rather than pinning a name list is the right call, and the Finding-1 rule (imported *and never referenced in the hub body*) correctly excludes `AgentSession` / `_session_state`.
- **Phase-5 simulation is clean.** The risk judge re-derived the mover set independently (**40 names**, matching Finding 1 and my own count), then `delattr`'d all 40 at `pytest_configure` and ran all **92** tracked test files that mention the hub, differenced against a clean baseline. The only breakage is `TestBackwardCompatExports` — the deliberately-retained tripwire, whose docstring now says phase 5 must delete it in the same diff. Round 1's two blockers no longer appear.
- **The independent reference sweep finds zero residuals**, across from-imports, quoted patch strings, every alias binding form, `getattr`, `sys.modules`, and the yaml/json config surface. Survivors are all correct: the `migrate_reflections_callables.py` migration-map keys (input side of the rewrite, must be retained), and two docstrings.
- **All four remaining tech-debt/nit items from round 1 are fixed**, and TD5's previously-RED test is green for the right reason: `test_valor_session_resume_release.py` sys.modules key now matches the real import site `tools/valor_session.py:1281`, and 55 tests pass. TD6's reconciliation of the plan's "11 breaking patch strings" down to the real 3 is in the PR body.
- **Gates reproduced locally, not taken from the description:** `ruff check .` clean; `ruff format --check .` = 1412 files already formatted; the changed and previously-flagged files give **159 passed** + **93 passed** on two serial runs. Verification table: every non-`[P5]` row passes (lint, format, migration coverage, migration idempotence, import smoke, bridge boundary); the 4 failing rows are all `[P5]`-marked and red by design.
- Zero production files: `git diff --name-only b152b7d3c 05a2e56e2 | grep -v '^tests/'` is empty. No test lost an assertion.

### Tech Debt

**1. `tests/unit/test_hub_alias_references.py:126-136, :143-157` — the guard's scope walk descends through nested scopes, so its verdict flips on source ordering.** `_scope_bindings` skips a `_SCOPE_NODES` child, but for every *non-scope* child it does `ast.walk(child)` (`:129`), which walks straight into any `FunctionDef` nested under an `if` / `with` / `try`. `_alias_hits` (`:147`) walks the same child and evaluates attribute nodes inside that nested function against the *outer* `env`.

Reproduced against the guard's own functions, using the same two code blocks in the two possible orders:

```
FALSE-NEGATIVE case -> []
FALSE-POSITIVE case -> [(5, 'q', '_slot_registry', 'Load'), (10, 'q', '_slot_registry', 'Store')]
```

where the source is a module-level `import agent.agent_session_queue as q` with a genuine `q._slot_registry = None`, plus an `if True:` block containing a function that rebinds `q` to `agent.session_state`. Put the hub import first and the real hit disappears; put the `if` first and line 5 (`agent.session_state`, entirely safe) is reported as a hit.

Both failure directions contradict the file's own docstrings — `:50-52` ("a binding made inside one of these shadows the enclosing scope's binding … a module-wide alias table would report false positives") and `:90-91` ("a guard that fires on safe code is one people learn to route around"). Zero live sites are affected today (the code-quality judge ran a corrected variant across all tracked Python and got the same empty result), so this is latent — but this guard is the thing phase 5 will lean on, and a guard that fails open on source ordering is not one to hand forward. One shared helper that yields a node's own-scope descendants, used by both functions, closes it.

**2. Same file, `:150-155` — the bare-dotted import class is invisible, and the docstring claims otherwise.** `_alias_hits` matches only `ast.Attribute` whose `.value` is an `ast.Name`, so `import agent.agent_session_queue` (no `asname`) followed by `agent.agent_session_queue.steer_session` parses as `Attribute`-of-`Attribute` and is never seen. Reproduced:

```
DOTTED (no asname) -> []
```

for a source containing both `agent.agent_session_queue._slot_registry = None` and a `steer_session` read. Relatedly `_scope_bindings:132` records `alias.asname or alias.name.split(".")[0]`, so that import form binds `"agent"` → `"agent.agent_session_queue"`, which is the wrong module for that name and could itself misfire.

Zero such sites exist in the repo today — I checked, and so did the risk judge — which is why this is tech debt and not a blocker. But the module docstring sells the guard as covering "attribute access through a module alias" without qualification, and phase 5 is meant to inherit it. Either cover the form or narrow the claim.

**3. `tests/unit/test_recovery_respawn_safety.py:123-135` — `test_nudge_main_path_skips_terminal` is still vacuous, for a second and unrelated reason.** The comment says "We patch AgentSession.query to detect any leak", but `_enqueue_nudge` (`agent/session_executor.py:635-780`) makes zero `AgentSession.query` calls — its only module-level use of the name is `async_create` at `:731`. The code-quality judge broke both terminal guards in `session_executor.py` and the test stayed green at the new target *and* the old one, so `mock_as.query.filter.assert_not_called()` asserts against a call the code never makes.

This is pre-existing (`b152b7d3c` carries the same assertion and comment) and the repoint to `agent.session_executor.AgentSession` is correct — the sibling `test_nudge_fallback_path_skips_terminal` was demonstrated to go red at the new target and stay green at the old one, so the target fix is real. But round 1's finding was "vacuous patch", and this site is still vacuous. Asserting on `mock_as.async_create`, as the sibling does, makes it real.

### Nits

**4. `tests/unit/test_zombie_session_resurrection.py:129-130`** — `@patch("agent.agent_session_queue._active_workers", {})` and `..._active_events` sit directly above the line this PR repointed at `:131`. Both are non-movers (the hub references them in its own body), so they survive phase 5 and the guard rightly ignores them — genuinely out of phase 4's scope. They are nonetheless vacuous for the same structural reason: `agent/session_health.py:38` binds both names from `agent.session_state`, so patching the hub's copy cannot reach the function under test. Worth a follow-up issue, not a change here.

**5. `tests/integration/test_worker_concurrency.py:303, :362, :413, :472, :531`** — the `finally` restores `_ss._slot_registry` *before* cancelling the still-running worker tasks, so an in-flight task can briefly acquire against the restored registry. Benign today (the restored value is `None`), but swapping the two statements is strictly tighter. Restore coverage itself is correct — every repointed write is inside a `try:` with a matching `finally:`.

**6. `tests/integration/test_worker_concurrency.py:288, :402, :456, :516`** — now that these ceilings are real rather than vacuous, three of the four are non-binding by construction: `SlotLeaseRegistry(3)` with 3 already-serializing same-chat sessions, `SlotLeaseRegistry(5)` with 2. Mutating `:288` from 3 to 1 still passes; only 0 fails. Fine for tests whose subject is serialization and worktree parallelism rather than the cap — the genuine ceiling coverage is at `:140` and `:342`, which already used `_ss`. Noted only so a future reader does not mistake those numbers for assertions.

### Advisory (not a finding — not this PR's file)

`docs/plans/sdlc-2876.md:985-995`, the "Baseline observed on `origin/main`" evidence table, is parsed as **9 additional checks** by `agent.verification_parser` rather than landing in `skipped`, so every run of the plan's verification table emits 9 spurious `FAIL`s of the form `/bin/sh: 11: command not found` (it executes the *Observed stdout* column as a command). The plan itself instructs every stage and every review judge to run this table, so the noise is guaranteed to recur at phase 5. This PR correctly touches no plan file; worth folding into phase 5 or a separate parser fix.

### Verified safe

- **40 movers, derived three times independently** (mine, both judges) with identical results; the guard's `_hub_reexports` agrees name-for-name.
- **Residual alias hits: `{}`** — the guard reports zero sites, and an independent regex + AST sweep over all tracked Python, yaml, json and toml agrees.
- **No cross-test leakage from the newly-real ceiling.** `worker/__main__.py:563` writes the same attribute but only from the worker entrypoint, which these tests never call. No autouse fixture resets it, and none is needed given the try/finally coverage.
- **Both intentional phase-5 red spots are loud and self-documenting**: `test_hub_alias_references.py:186-189` fails with an explicit "either phase 5 has landed … or the AST derivation is broken", and `TestBackwardCompatExports`'s docstring names itself as phase 5's to delete. Phase 5 should expect exactly two deliberate failures and nothing else.

---

The two blockers are gone and the evidence standard in this PR remains unusually high. What is left is the guard auditing itself: items 1 and 2 are gaps in the very artifact whose purpose is to stop phase 5 inheriting a blind spot, and both are a few lines to close.

valorengels:
## Review:

**APPROVED** — 0 blockers, 0 open tech debt, 1 acknowledged deferral (tracked in #3066).

Round 3, scoped to the question round 2 left open: are its three tech-debt items genuinely closed by `e890b6274`? Each was re-verified by running the reviewer's own reproduction against the current head, not by reading the diff.

### TD1 — nested-scope descent, verdict flipping on source ordering: **CLOSED**

Round 2's decisive experiment was one module-level `import agent.agent_session_queue as q` plus a real `q._slot_registry = None`, and an `if True:` block containing a function that rebinds `q` to `agent.session_state` — run in both source orders. It returned `[]` in one order (false negative, the real hit vanished) and a hit on the *safe* `agent.session_state` line in the other (false positive).

Re-run at `fef38a245`:

```
HUB-FIRST            -> [(5, 'q', '_slot_registry', 'Store')]
IF-FIRST             -> [(10, 'q', '_slot_registry', 'Store')]
```

Both orders now report exactly one hit, and it is the genuine module-level write in each (line 5 / line 10 — the same statement, relocated). The nested function's safe access is reported in neither. Both failure directions are gone.

The fix is structural rather than a patch over the symptom: `_own_scope_nodes` is one helper that stops at `_SCOPE_NODES` boundaries, and both `_scope_bindings` and `_alias_hits` go through it. `_alias_hits` gathers `env` for the whole scope before evaluating any reference, which is what removes the ordering dependence rather than merely relocating it.

### TD2 — bare dotted import invisible: **CLOSED**

Round 2 got `[]` for a source containing both `agent.agent_session_queue._slot_registry = None` and an `agent.agent_session_queue.steer_session` read, because `_alias_hits` matched only `Attribute`-of-`Name`. Re-run:

```
DOTTED (no asname)   -> [(4, 'agent.agent_session_queue', '_slot_registry', 'Store'),
                         (5, 'agent.agent_session_queue', 'steer_session', 'Load')]
```

Both forms detected, `Store` and `Load` distinguished. `_resolve_dotted` now walks the whole `Attribute` chain to its `Name` base, and `_scope_bindings` maps a bare `import a.b.c` to `a` rather than to `a.b.c` — the related mis-binding round 2 flagged in the same item. The docstring's coverage claim is now true rather than narrowed.

### TD3 — `test_nudge_main_path_skips_terminal` vacuous: **CLOSED, with a residual**

The specific defect is fixed. `mock_as.query.filter.assert_not_called()` — an assertion against a call `_enqueue_nudge` never makes — is replaced by `mock_as.async_create.assert_not_called()`, which the failure path genuinely reaches. Demonstrated, both directions:

```
both terminal guards broken  -> 6 failed, 1 passed, 49 deselected
restored (control)           -> 7 passed, 49 deselected
```

The assertion is live. It is not vacuous any more.

**Residual, filed as #3066 rather than held against this PR.** Breaking *only* the entry guard at `agent/session_executor.py:664` leaves all five parametrizations green:

```
entry guard alone broken     -> 7 passed, 49 deselected
```

Because the test's mock session id finds nothing in Redis, `get_authoritative_session` returns `None` and execution always takes the fallback branch, where the guard at `:700` catches it. So both tests in `TestEnqueueNudgeTerminalGuard` cover the fallback guard, and `:664` has no test that fails on its own. This reproduces identically on the merge base — it is pre-existing, not this PR's damage, and it is the same class already deferred to #3054 / #3055. Closing it means giving the main-path test a live `get_authoritative_session` and asserting on `transition_status`, which is a change to the test's subject rather than to phase 4's charter.

### Gates, run here rather than taken from the description

```
python -m ruff check .                -> All checks passed!
python -m ruff format --check .       -> 1412 files already formatted
import smoke (5 modules)              -> OK
all 26 changed test files, -n0        -> 614 passed in 119.69s
```

Scope is still clean: `git diff --name-only origin/main...HEAD` is `tests/` only plus this round's `tests/README.md` docs cascade. The docs commit (`fef38a245`) documents the module-alias form under the Patch-Target Convention and lists the guard in the validation table; it deliberately leaves the paragraph's re-export claim alone, since inverting that proposition is phase 5's to do in the same diff that makes it true.

Both intentional phase-5 red spots remain in place and self-documenting: `test_hub_alias_references.py`'s AST-derivation assertion, and `TestBackwardCompatExports`, whose docstring names itself as phase 5's to delete.

Issue #3054, opened 2026-08-28
Vacuous mock patches on the agent_session_queue hub leave 5 test files hollow (one red on main)


valorengels:
Surfaced by the PR #3048 review (phase 4 of #2876). These are **not** phase-5 breaks and were outside that PR's charter, which covered only *movers* (names `agent.agent_session_queue` imports and never references in its own body). `AgentSession`, `_active_workers`, and `_active_events` are hub-*used*, so they survive the phase-5 re-export deletion. They are nonetheless a live coverage hole today.

The defect class: a test patches `agent.agent_session_queue.<name>`, but the function under test resolves `<name>` from its own module's namespace. The patch rebinds a name nothing reads, so the mock is never touched and every assertion against it holds trivially.

Demonstrated:

```python
with patch('agent.agent_session_queue.AgentSession') as m:
    agent.session_health.AgentSession is m      # False
    agent.session_executor.AgentSession is m    # False
```

## Sites

| File | Line | Patches | Code under test resolves from |
|---|---|---|---|
| `tests/unit/test_recovery_respawn_safety.py` | 124, 146 | `agent.agent_session_queue.AgentSession` | `agent/session_executor.py:31` (`_enqueue_nudge` at `:635`) |
| `tests/unit/test_zombie_session_resurrection.py` | 72, 134 | `agent.agent_session_queue.AgentSession` | `agent/session_health.py:41` |
| `tests/unit/test_zombie_session_resurrection.py` | 129-130 | `agent.agent_session_queue._active_workers` / `._active_events` | `agent/session_health.py:38` (from `agent.session_state`) |
| `tests/unit/test_stale_cleanup.py` | 83, 120, 324, 378 | assigns `queue_module._active_workers` | `scripts/update/run.py:432` (from `agent.session_state`) |
| `tests/unit/test_update_stale_session_fence.py` | 120, 385 | assigns `queue_module._active_workers` | same |

Each file has an in-file control showing the correct target: `test_zombie_session_resurrection.py:87` and `:100` already use `agent.session_health.AgentSession`.

## Impact

- `TestStartupRecoverySkipsTerminalSessions` / `TestHealthCheckSkipsTerminalSessions` assert `count == 0`, which holds only because the real Redis is empty — they would pass even if terminal sessions *were* recovered. Probe: rewriting the fixture to a non-terminal `status="running"` still passes.
[...]

Issue #3020, opened 2026-08-26
Plan verification rows can write production state: a bare command that defaults to a real path


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/sdlc-3020.md

> **Context:** SDLC plan documents carry a `## Verification` table of shell commands. The pipeline instructs every PR-review judge to *run* those commands as part of grading a PR. That makes the table executable code run repeatedly by automated reviewers, not just documentation.

## Problem

**Current behavior:**
A verification row whose command omits a target/dry-run flag and defaults to a production path will mutate production state every time any judge grades the PR.

Observed concretely during PR #3019 (phase 2 of #2876). The `[P2]` row was:

```
python scripts/migrate_reflections_callables.py --check-idempotent
```

No `--target`, no `--dry-run`. The script therefore resolved its default path — `~/Desktop/Valor/reflections.yaml`, the iCloud-synced production registry — and migrated it. Confirmed after the fact: mtime `Aug 26 12:02`, hub references went 3 → 0, while the repo's `config/reflections.yaml` remained untouched at 3.

That instance was benign: both migrated callables (`agent.session_health.cleanup_corrupted_agent_sessions`, `agent.session_revival.cleanup_stale_branches_all_projects`) resolve on `main` today, so no reflection job breaks, and the next `/update` re-migrates as a no-op. No revert was attempted — a second unsanctioned write moving it backward would compound the problem.

The real cost was evidential: that plan's phase-5 propagation gate exists to observe a 3 → 0 transition in the vault, and the transition had already been consumed by a verification run.

**The trap has no safe local workaround.** Adding `--dry-run` to that row does not help: `--check-idempotent --dry-run` is a *silent* no-op (guard at `scripts/migrate_reflections_callables.py:391`), producing a green row that verifies nothing. So the naive fix converts a dangerous row into a vacuous one — arguably worse, since it looks verified.

**Desired outcome:**
No verification row can mutate production state. A row that cannot be run safely fails loudly rather than silently passing or silently writing.

## Why this is general, not specific to #2876

Any plan row invoking a script whose default target is a real path has this shape. The hazard is a property of the verification-table convention meeting scripts that default to production, not of this one migration script. It compounds with how many judges run: each PR-review round re-executes the whole table.

## Solution Sketch

Open question — the right layer isn't obvious:

[...]

PR #3067, opened 2026-08-31
Delete the agent_session_queue re-export hub, and guard it shut (phase 5 of 5)

### LINKED ISSUE #2876: De-hub agent/agent_session_queue.py -- stop using it as a re-export supermodule

valorengels:
Phase 5 of 5 for #2876, and the one that does the deleting. `agent/agent_session_queue.py` stops being a re-export hub: the 40 symbols it imported from eight sibling modules solely so callers could reach them through this path are gone, and an architectural guard fails if any return.

## What went, what stayed

A symbol is a **pure re-export** when the hub imports it and never references it in its own body — the plan's Finding 1 definition. 40 names met it and were deleted. 26 imports are referenced in the hub's own code and stayed; they were never re-exports, whatever the comment above them said.

| Source module | Deleted | Kept |
|---|--:|--:|
| `agent.output_router` | 4 | 0 |
| `agent.session_completion` | 2 | 1 |
| `agent.session_executor` | 6 | 1 |
| `agent.session_health` | 15 | 7 |
| `agent.session_pickup` | 3 | 2 |
| `agent.session_revival` | 7 | 1 |
| `agent.session_state` | 3 | 10 |

`agent.session_state` stays imported as a **module object** (`import agent.session_state as _session_state`) rather than as names, because the hub mutates through it — that is a working import, not a re-export, and it is why the guard's rule is "imported and unused" rather than the looser "imported and not defined here".

Every `# noqa: F401` in the file is gone, along with the two stale "All symbols re-exported at the top of this module" summary comments and the retention comment that explained why `cleanup_stale_branches_all_projects` had to survive phase 2.

## The guard

`tests/unit/test_no_reexport_hub.py` derives the invariant from the hub's AST: no name imported from a sibling `agent.*` module may go unreferenced in that file.

It deliberately does **not** ban the `# noqa: F401` string. That only catches the historical spelling — a re-export added without a suppression, in a file ruff does not lint, is equally a re-export and would sail through a string check. Ruff's own F401 is not a substitute either: it is configurable per-file and counts a `__all__` entry or a `TYPE_CHECKING` block as a use, which is exactly what a hub re-export would hide behind.

**Demonstrated red.** Reintroducing one real re-export (`steer_session`, from its actual owner `agent.session_executor`, unused in the hub body):

```
FAILED tests/unit/test_no_reexport_hub.py::test_hub_imports_nothing_it_does_not_use
E   Failed: agent/agent_session_queue.py re-exports 1 symbol(s) it never uses:
E     agent_session_queue.py:48  steer_session  (from agent.session_executor)
```

Removed again → `3 passed`. The failure names file, line, symbol and source module.

My first injection attempt put `steer_session` on `agent.session_revival`, which raised `ImportError` at collection instead — a red that proves nothing about the guard. Recorded because the distinction is the whole point of a demonstrated red: the guard has to be what fires, not the import system.

The guard also carries `test_guard_detects_a_reintroduced_reexport`, an in-file check against a synthetic module, so `unused_sibling_imports()` returning `[]` because its AST walk broke is distinguishable from returning `[]` because the file is clean.

**Its stated limits are asserted, not just described.** Self-review falsified one of them: the docstring claimed the guard treats neither an `__all__` entry nor a `TYPE_CHECKING` block as a use, and only the first half was true. An `__all__` entry is a string literal, so an `__all__`-based re-export *is* caught; a name referenced in an annotation or under `TYPE_CHECKING` is an `ast.Name` load and is *not* flagged. The behaviour is correct — a typing-only import does not exist at runtime, so nothing can be re-exported through it — but the docstring had it backwards. Corrected, and both limits plus the aliased-import case now carry tests, because a stated property with no test drifts.

## The phase-4 alias guard goes dormant, not green

`tests/unit/test_hub_alias_references.py` derives its hazard set from the hub's re-exports. With none left, that set is empty and its live-site check can no longer find anything — it would have passed vacuously forever. Instead:

- `test_hub_carries_no_reexports` asserts the empty set: the phase-4 tripwire's proposition, inverted in the same diff that made it true, as its docstring required.
- `test_hub_reexport_derivation_still_works` exercises the derivation against a synthetic module through a new `source=` seam, so a broken walk is still caught.
- `test_no_module_alias_reaches_a_hub_reexport` now `pytest.skip`s with an explicit reason rather than reporting a pass it did not earn. It re-arms automatically if a re-export is ever reintroduced.

`TestBackwardCompatExports` in `test_steering_mechanism.py` is deleted here, as its docstring instructed — it asserted the exact proposition this PR inverts.

## Propagation gate (Race 1), verified on this machine

The plan restates phase 5's gate as three conjuncts, because the original 3→0 transition evidence was consumed out of band on 2026-08-26:

| # | Evidence | Observed |
|---|---|---|
| 1 | Phase-2 merge SHA is an ancestor of HEAD | `git merge-base --is-ancestor 7fee19e02 HEAD` → exit 0 |
| 2 | Step 1.659 actually executed | `run_reflections_callables_migration()` → `action=noop, success=True` — the "already name their owning modules" branch, which proves the step ran and found nothing to do |
| 3 | End state the scheduler reads is clean | `grep -c 'agent\.agent_session_queue\.' config/reflections.yaml` → `0` |

Proposition 2 of the phase-2 gate also holds: `~/Desktop/Valor/reflections.yaml` → `0`. Conjunct 2 was executed by calling the Step 1.659 entry point directly rather than running a full `/update`, to avoid a service restart; SHA-256 of both `config/reflections.yaml` and the vault original are byte-identical before and after, so it was a genuine no-op and consumed no transition.

**This is one machine.** The plan names the fleet-wide check as a human gate that this lane cannot reach. Any other machine running reflections needs an `/update` cycle before it pulls this commit.

## Verification

| Check | Result |
|---|---|
| `grep -c "F401" agent/agent_session_queue.py` | `0` |
| `grep -c "backward compatibility" agent/agent_session_queue.py` | `0` |
| `grep -r 're-export' docs/features/ \| grep 'agent_session_queue' \| grep -cv 'sustainability'` | `0` |
| `python -m ruff check .` | All checks passed |
| `python -m ruff format --check .` | 1414 files already formatted |
| Import smoke (5 modules) | OK |
| `pytest tests/unit/test_no_reexport_hub.py` | 6 passed |
| All 91 tracked test files mentioning the hub | 2302 passed, 1 skipped, **13 failed** |

**The 13 failures are pre-existing, verified rather than assumed.** The same 13 node IDs fail on merged `main` (`96774f5ae`) with none of this PR's changes:

```
main:      13 failed, 85 passed
this PR:   13 failed, 2302 passed, 1 skipped
```

Identical node-ID sets — 9 in `test_agent_session.py::TestWorkerKey*`, plus `test_stale_cleanup.py::test_live_worker_skips_session` (already documented RED on main in #3054), `test_parent_child_round_trip.py`, `test_session_notify.py`, `test_remote_update.py`. Zero regressions from this diff.

## One anti-criterion needed a reword, not a weakening

The `backward compatibility` row hit a true, unrelated sentence: the `register_callbacks` docstring explaining that callbacks are keyed by plain `project_key` when no transport is given. Rather than weaken the gate with an exclusion, that line now says "which is what callers predating transports expect" — same meaning, and the anti-criterion stays absolute. Same judgement call as the `grep -cv 'sustainability'` exclusion the plan already documents, resolved the other way because here the phrasing was incidental rather than load-bearing.

## Also fixed

`agent/reflection_scheduler.py::_resolve_callable`'s docstring example named `agent.agent_session_queue._agent_session_health_check` — a path that no longer resolves, in the docstring of the resolver itself. `run_reflection` swallows resolution failures per-tick in a broad `except`, so a wrong example there costs a silently dead daily job with a green-looking worker. Repointed to `agent.session_health._agent_session_health_check`, with a note saying why the owning module is the one to name.

## Retained deliberately

The `agent.agent_session_queue.*` strings in `scripts/migrate_reflections_callables.py` and its tests are migration-map **keys** — the input side of the rewrite. The script has to keep recognising the old dotted path to migrate a machine that has not yet run `/update`. Deleting them would break the very propagation this phase depends on.

Closes #2876


valorengels:
## Review:

**APPROVED** — 0 blockers, 0 open tech debt.

Reviewed by attacking the diff rather than reading it. Disclosure: this was a self-review — the prior multi-judge fork ran two hours and died without posting, so the checks below were chosen to be falsifiable rather than confirmatory, and two of them found real problems that are fixed in this branch.

### The deletion is complete and correctly scoped

The 40-name mover set was re-derived independently from the hub's AST at the merge base and matches the plan's Finding 1 exactly. Per-statement, 40 deleted / 26 kept, and every survivor is referenced in the hub's own body — which is now mechanically enforced rather than asserted.

`agent.session_state` correctly stays as a module-object import: the hub mutates through it, so it is a working import and an alias reaching it sees live attributes.

**Every reference class swept, not just imports.** An AST pass for `ImportFrom` on the hub (absolute *and* relative, so `agent/__init__.py:3`'s `from .agent_session_queue import` is covered) plus a regex for `agent.agent_session_queue.<name>` across **all tracked files, not just Python** — yaml, json, md included. Residual hits, all triaged:

- `tests/unit/test_steering_mechanism.py` — `TestBackwardCompatExports`, deleted here as its own docstring instructed.
- `agent/reflection_scheduler.py:447` — a docstring example, repointed (see below).
- `scripts/migrate_reflections_callables.py` + its tests — migration-map **keys**, the input side of the rewrite. Correctly retained: the script must keep recognising the old path to migrate a machine that has not yet run `/update`. Deleting them would break the propagation this phase depends on.
- `docs/archive/plans-completed/*` and `docs/plans/sdlc-2876.md` — historical record and this plan's own prose describing what it migrated *from*.

Also checked and clean: no `__all__` in the hub, no `getattr(...)` or `sys.modules[...]` reaching it dynamically, and `agent/__init__.py` pulls only `PRIORITY_RANK` / `AgentSession` / `enqueue_agent_session` / `register_callbacks` — none of them movers.

### The guard is real, and its demonstrated red is honest

Reintroducing `steer_session` from its true owner, unused in the hub body, fails with the file, line, symbol and source module named; removing it returns green.

Worth recording: the first injection attempt put `steer_session` on `agent.session_revival`, which raised `ImportError` at collection. That is a red that proves nothing about the guard — the import system fired, not the assertion. The PR body records the distinction rather than quietly keeping the passing-looking transcript, which is the right instinct for a demonstrated red.

Deriving the invariant from the AST rather than banning the `# noqa: F401` string is the correct call: the string check only catches the historical spelling, and a re-export added without a suppression would pass it.

### Two things this review actually found

**1. The guard's docstring overclaimed, and is now fixed.** It stated the guard treats neither an `__all__` entry nor a `TYPE_CHECKING` block as a use. Executed against the guard's own helpers:

```
__all__ entry only               flagged=['X']
TYPE_CHECKING annotation only    flagged=[]
```

Only the first half was true. The behaviour is right — a typing-only import has no runtime existence, so nothing can be re-exported through it — but the docstring had it backwards. Corrected in `7ed45603b`, and both limits plus the aliased-import case now carry tests. This is the failure mode the guard exists to prevent, occurring in the guard's own prose, so fixing it rather than noting it was the only acceptable outcome.

**2. The phase-4 alias guard would have gone silently vacuous.** `test_hub_alias_references.py` derives its hazard set from the hub's re-exports; with none left it can never fire. Left alone it would have reported a green it did not earn — permanently. It now asserts the empty set explicitly, exercises its derivation against a synthetic module through a new `source=` seam, and `pytest.skip`s the live-site check with a stated reason. It re-arms automatically if a re-export returns. This is the better outcome than deleting it.

### Propagation gate

All three of Race 1's conjuncts verified on this machine, including conjunct 2 executed through the tracked `run_reflections_callables_migration()` entry point rather than inferred: `action=noop`, the "already name their owning modules" branch, which is what distinguishes "the step ran and found nothing" from "the step never ran". SHA-256 of `config/reflections.yaml` and the vault original are byte-identical before and after, so this observation consumed no state — the exact hazard (#3020) that cost the original 3→0 evidence.

The fleet dimension remains a human gate this lane cannot reach, and the PR body says so plainly rather than implying machine-wide coverage.

### Gates

```
ruff check .                       -> All checks passed
ruff format --check .              -> 1414 files already formatted
import smoke (5 modules)           -> OK
grep -c F401 <hub>                 -> 0
grep -c "backward compatibility"   -> 0
docs/features re-export claims     -> 0
all 91 hub-touching test files     -> 2302 passed, 1 skipped, 13 failed
```

**The 13 failures are pre-existing and I verified it myself rather than accepting the claim.** Running the same five files on merged `main` (`96774f5ae`) with none of this diff yields the identical 13 node IDs — `13 failed, 85 passed`. Zero regressions.

### Judgement call I agree with

Rewording the one unrelated "for backward compatibility" phrase in `register_callbacks`' docstring, rather than adding an exclusion to the anti-criterion. The phrase was incidental there; keeping the gate absolute is worth more than preserving a turn of phrase. The contrast with the `grep -cv 'sustainability'` exclusion the plan documents — kept, because that sentence is load-bearing and owned by another issue — is the right distinction.

### Note for the docs stage

The `_resolve_callable` docstring fix is small but not cosmetic: it named a path that no longer resolves, inside the resolver whose failures `run_reflection` swallows per-tick. A wrong example there costs a silently dead daily job with a green-looking worker, which is the precise failure mode that made phases 2-5 order-dependent in the first place.

PR #3080, opened 2026-09-02
Render /ask-me questions as native Telegram polls, and stop nudging a session past its own question

### LINKED ISSUE #2701: Render /ask-me questions as native Telegram polls so a blocked agent can be unblocked with one tap
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/ask-me-telegram-polls.md

> **Context:** The [`/ask-me` skill](.claude/skills-global/ask-me/SKILL.md) interviews the human one question at a time when an agent is blocked on a judgment call. In a local Claude Code session it renders as the `AskUserQuestion` tool: a tappable multiple-choice card with an "Other" escape hatch. Over the Telegram bridge that affordance does not exist, so the same question arrives as prose the human has to read and type an answer to. Telegram has a native poll message type that is a close match, and this issue is about rendering `/ask-me`-style questions as polls on the Telegram surface.

## Problem

When work happens through Telegram rather than a local terminal, a blocked agent asks its question as plain text. The human then has to compose a written answer on a phone. The interaction the skill was designed around — glance, tap, done — is lost, so questions get answered slowly, partially, or not at all, and the agent stays blocked.

**Current behavior:**
- `AskUserQuestion` does not prompt under `claude -p`. It fires a `needs_human` hook edge ([`agent/session_runner/hook_edge.py`](agent/session_runner/hook_edge.py)) and the turn ends; the question reaches Telegram as prose via the outbox.
- The human answers by typing, or by reply-to, or not at all.
- The bridge registers only `events.NewMessage` ([`bridge/telegram_bridge.py:1165`](bridge/telegram_bridge.py#L1165)), so it has no concept of a structured answer.

**Desired outcome:**
A question the agent poses at turn end is delivered to Telegram as a native poll. One tap records the answer, the bridge converts the vote into a steering message, and the session resumes on its next turn with the choice in hand. No new blocking primitive, no new inbound path.

## Definitions

| Steering message | A queued instruction the worker drains at a turn boundary; the sole inbox for mid-session input | [`agent/steering.py`](agent/steering.py), [docs/features/session-steering.md](docs/features/session-steering.md) |
| Outbox | Redis list `telegram:outbox:{session_id}` that the relay drains and sends via Telethon | [`bridge/telegram_relay.py`](bridge/telegram_relay.py) |
| `needs_human` hook edge | Hook-emitted signal that the agent wants human input; ends the turn under headless `claude -p` | [`agent/session_runner/hook_edge.py`](agent/session_runner/hook_edge.py) |
| MTProto | Telegram's native client protocol, used here through Telethon under a user account (not a bot) | [core.telegram.org/api](https://core.telegram.org/api) |
| Message drafter | The comms layer that reviews and shapes every outbound message before it reaches a chat | [docs/features/message-drafter.md](docs/features/message-drafter.md) |

## Prior Context

- #1802: PM session had no file-capable send path — added attachments to the same outbox → relay seam a poll would extend.
- #1955: Message drafter had no local-file-path awareness; precedent that a new outbound message shape needs matching drafter awareness.
- #1688 / #1922: Established the `needs_human` hook edge and its routing, which is the trigger this feature renders.

## Solution Sketch

- Add a poll payload variant to the existing outbox → relay path. A `send_poll` helper belongs in [`bridge/response.py`](bridge/response.py), which already issues raw MTProto calls (`SendReactionRequest`).
- Register an `events.Raw` handler for `updateMessagePoll` so votes are observed. For a user account, individual-vote updates (`updateMessagePollVote`) are bots-only; a user account sees aggregate counts, which is unambiguous in a 1:1 chat.
- Translate a vote into a steering message via `push_steering_message`, so the answer travels the path a typed reply already travels.
- The last option of every poll is always `Other: wait for followup message`. On that vote the agent replies with a reworked or narrowed question as normal text, and the human answers by reply-to.
- Relax `/ask-me`'s one-question-at-a-time rule from hard enforcement to a stated preference, so genuinely independent questions may go out as separate polls.

Owner decisions already made (do not relitigate in planning): no mid-turn blocking; the `Other` option wording above; one-at-a-time becomes guidance rather than a rule.

## Recon Summary

**Confirmed:** 4 items
- Telethon 1.42 is installed and the bridge authenticates as a user account, so `BOT_POLLS_DISABLED` ("you cannot create polls under a bot account") does not apply.
- A live probe against a copy of the bridge session (updates disabled, target Saved Messages) sent `InputMediaPoll` successfully and read back `MessageMediaPoll`; the probe message was deleted. Polls are not blocked in a private chat.
- `bridge/response.py` already performs raw MTProto calls, so no new dependency or client plumbing is needed.
- `tools/send_message.py` → `telegram:outbox:{session_id}` → `bridge/telegram_relay.py::process_outbox` is the existing outbound seam a poll payload extends.

**Revised:** 1 item
- Original framing imagined a blocking, `AskUserQuestion`-like prompt. Headless `claude -p` cannot block, so scope is a turn-boundary rendering change plus a vote→steering translation.

**Pre-requisites:** 1 item
- The probe only covered a self-DM (Saved Messages). Sending a poll into another user's DM is unverified and must be confirmed first; if MTProto rejects it, the whole feature is limited to group chats and scope must be revisited.

**Dropped:** 1 item
- Free-text `Other` capture inside the poll. Telegram polls have fixed options, so the escape hatch is the followup-message option instead.

## Acceptance Criteria

- [ ] A poll sent from the bridge account into a real user DM is confirmed to work (or the limitation is documented and scope revised).
- [ ] An agent question posed at turn end is delivered to the chat as a native Telegram poll with the recommended option first and `Other: wait for followup message` last.
- [ ] Tapping an option produces a steering message that the worker consumes on its next turn, with the chosen option text present in the session's input.
- [ ] Voting `Other: wait for followup message` results in a plain-text followup question the human can answer by reply-to, resuming the same session.
- [ ] Catchup, dedup, the message drafter, and reaction-based "handled" detection treat an outbound poll as a question awaiting an answer, so the nudge loop never re-asks a question already on screen.
- [ ] `/ask-me` states one-question-at-a-time as a preference with the hard prohibition removed.
- [ ] Tests cover: poll payload construction, vote → steering translation, and the `Other` fallback path.

## Downstream

This issue feeds `/do-plan`, which will produce `docs/plans/{slug}.md`, then `/do-build`.

Constraints for the planner:
- No new dependencies; Telethon already provides everything needed.
- Do not introduce a blocking or mid-turn input primitive. Steering at turn boundaries is the only sanctioned inbox.
- Email and local-session surfaces have no poll equivalent; the change must degrade to today's text behavior there.

valorengels:
Closes #2701

Renders an `/ask-me` question from a blocked **eng** session in a Telegram **group** as a native poll. One tap unblocks the agent instead of a paragraph on a phone and a written reply.

Plan: `docs/plans/ask-me-telegram-polls.md`.

### 1. This fixes a PRE-EXISTING bug that is wider than polls

`agent/output_router.py`'s `determine_delivery_action` contains:

```python
if session_type == "eng" and classification_type == "sdlc":
    return "nudge_continue"
```

It is **unconditional** and sits **ahead of every `stop_reason` branch**. So an sdlc eng session that poses *any* question — as a poll, or **as plain prose today** — is re-enqueued with `NUDGE_MESSAGE` and proceeds on a guess rather than waiting for the human. That is true on `main` right now, with no poll code involved.

The owner ruled this in scope for #2701 rather than a follow-up, because without it the poll renders and is answerable while the asking session burns nudge turns past it — defeating the feature.

`pause_open_question` (Task 8a, own commit `f76cdc87c`) is the fix.

### 2. Blast radius, as an assertion rather than a claim

- `has_open_question` **defaults to `False`** on both `determine_delivery_action` and `route_session_output`, so the nudge loop's behavior is unchanged for every session with no outstanding poll — which is every eng session in the repo today.
- **All 18 pre-existing cases in `tests/unit/test_output_router.py` pass unmodified.** A diff touching an existing case would have been the signal the branch was placed wrong.
- The branch is conditioned on an **existing** record — an unanswered `telegram:poll:{poll_id}` row — so it introduces no new state to keep in sync.
- Placement is covered per case: terminal status, `completion_sent`, post-compaction, watchdog, rate-limit, empty output, nudge cap and `waiting_for_children` all still win over it. It overrides exactly one line.
- `determine_delivery_action` stays a **pure function**. The executor performs the Redis read (thread-offloaded, fail-quiet) and passes the result in, matching the `last_compaction_ts` contract the wrapper's docstring already states. An AST test enforces this.

Pausing at a turn boundary is consistent with the owner's earlier "no mid-turn blocking" decision: nothing holds a turn open — the turn has already ended, and the router simply stops re-enqueuing.

### 3. The `events.Raw` handler is consistent with "the bridge is I/O only"

The handler filters `UpdateMessagePoll`, logs `poll_update_observed`, and calls `translate_poll_vote`. It **makes no SDLC decision**. The translation ends in a steering message on the **existing** path — `agent/steering.py::push_steering_message`, the sole steering inbox — or, on the mainline, in the same `dispatch_telegram_session` re-enqueue a typed reply uses. **No second steering path was invented.** A Verification anti-criterion greps for raw steering-key writes and returns 0.

The reconciliation loop lives in `bridge/poll_reconcile.py`; `telegram_bridge.py` only imports and starts it next to `relay_loop`.

### Task 1 — `gate-poll-vote-readback`: **PASS**, and it earned its keep

The plan required re-running this under supervision rather than trusting the unaudited spike-8 PASS. That call caught a wrong constant before nine subsystems were built on it.

**The plan's stated 100-byte `PollAnswer.option` ceiling is wrong. The real ceiling is 8 bytes.** Measured against live MTProto, each probe poll deleted immediately:

```
1, 2, 4, 8 bytes: OK
9, 12 bytes:      REJECTED — "A poll option used invalid data (the data may be too long)"
```

It fails **at the wire with no local signal**. The TL schema declares `option:bytes` with no visible bound and every published reference says 100, so this is only knowable by probing. The plan's `f"{index}:{hex32}"` layout is 34 bytes and cannot be sent at all.

Re-cut to packed binary: `bytes([index])` + the first 7 bytes of the hint = 8 bytes exactly, with `correlation_matches()` owning the prefix-vs-full-hint comparison in one place. `poll_id_hint` is unchanged — full uuid4 hex, still the registry key, still one producer. 56 bits disambiguates a bounded window of recent polls in one chat, and the adoption rule's "more than one candidate → adopt nothing" bail covers a collision.

Readback verbatim:

```
Sent poll: msg_id=1410 server_poll_id=6106914788732830966
correlation_id supplied: ff8b722f670d4361beda709c4b8476a7
Server-assigned id differs from supplied: True
Self-vote cast for option index 1 (b'\x01\xff\x8br/g\rC')

PollResults(min=False, results=[
  PollAnswerVoters(option=b'\x00\xff\x8br/g\rC', voters=0, chosen=False, correct=False),
  PollAnswerVoters(option=b'\x01\xff\x8br/g\rC', voters=1, chosen=True,  correct=False)
], total_voters=1, recent_voters=[], solution=None, solution_entities=[])

  index=1 correlation_prefix=ff8b722f670d43 matches_hint=True voters=1
close_poll(closed=True): True

GATE 1: PASS
```

### Task 2 — `gate-poll-human-tap`: **PASS** on readback, negative on attribution

A real human tap on a poll originated through the shipped `valor-ask-poll` chain (msg 1413):

```
total_voters: 1
PollAnswerVoters(option=b'\x004\x05\x0c2\xaf\xe1G', voters=1, chosen=False, correct=False)
  index=0 chosen='Open a PR and let review run normally' voters=1
```

`chosen=False` on the voted option is **account-relative**, so its being False while `voters=1` is positive confirmation this was a *different* human's vote — exactly what this gate establishes over Task 1's self-vote.

**Attribution is negative, structurally, and is therefore not built.** `GetPollVotesRequest` failed; the cause is that polls are sent `public_voters=False` and per-voter detail is only retrievable for a public poll. It can never succeed here. Building it would be dead weight on the inbound fast path, so `sender_name` falls back through the session's `initial_telegram_message["sender_name"]` to the literal `"Telegram poll"`, and an AST test keeps the call from creeping back. Making it work would mean flipping `public_voters=True`, publishing every voter's identity to the group in exchange for a name in a steer line — **not taken**, and recorded rather than silently dropped. Risk 2b is unaffected: a wrong steer is correctable by an ordinary reply.

### Task 10a — `gate-poll-e2e`: **UNRESOLVED** (not PASS, not FAIL)

Two bounded poll windows were opened against a live eng session and both lapsed with **no operator tap**. No tap ever arrived, so the gate's load-bearing assertion — that the chosen option text appears in that session's next turn input, via `_drain_steering_boundary()` — was never exercised in either direction.

This is UNRESOLVED, not a FAIL. A FAIL is specifically "a tap arrived and no steer reached the session"; that did not happen, because no tap arrived. Per Task 15's *satisfied-by-pause* rule, extended verbatim to `gate-poll-e2e` in cycle 7, an UNRESOLVED gate lets validation run and requires this row be reported UNRESOLVED and **never as passed** — only a FAIL stops the pipeline.

**What this leaves unverified, stated plainly.** The full inbound chain has been exercised end to end only by unit and integration tests, never by a real human tap on a real poll. Tasks 1 and 2 do establish, against live MTProto, that a real human's vote is readable and that the correlation prefix survives the round trip; what remains unwitnessed in production is the last hop — observed vote → translated steer → next turn input. Reviewers should weigh this as an unexercised path rather than a verified one.

Note also that polls cannot work in production until the bridge is restarted onto this code: the running bridge is on `main`, which discards `type: "poll"` as an unknown message type.

### No probe opened a second updates-enabled client

Tasks 1 and 2 ran against a **temp copy** of the session file with `receive_updates=False` and no writeback; a second client with updates enabled on the bridge's auth key can consume updates the live bridge needs. No probe touched a DM. Everything sent was deleted.

**One deviation from the plan's letter, stated plainly.** The plan has Task 2 originate "as the live bridge". The live bridge runs `main`, which discards `type: "poll"` as an unknown message type — verbatim from `logs/bridge.log`:

```
WARNING bridge.telegram_relay process_outbox
Relay: unknown message type 'poll', discarding: {'type': 'poll', ..., 'poll_id_hint': '21af...'}
```

So I did **not** put unmerged code on the shared bridge and did **not** stop it (other lanes are live). I won the LPOP race from my own process and drove the shipped `_send_queued_poll` on the payload the CLI actually queued. Every link is shipped code; only the loop's caller differs. **This resolves itself at merge**: `./scripts/valor-service.sh restart` is already required and already part of `/update`.

## Design notes worth a reviewer's eye

**`COMPLETED` is the mainline, not an edge case.** `/ask-me` ends its turn on the `needs_human` edge and `PM_NEEDS_HUMAN` is clean and wrap-up-eligible, so the asking session is finalized `"completed"` long before a human taps. `resume_completed_session` → `dispatch_telegram_session` is what ships; dropping it loses **every** vote, not a rare one.

**`/ask-me` calls `AskUserQuestion` after `valor-ask-poll`, and that is not redundant.** The first renders the question; the second ends the turn. The `needs_human` edge is a `PreToolUse` tool-name match against `AskUserQuestion`, and a Bash call to `valor-ask-poll` has tool name `Bash`, which never matches. The reason is stated inline in `.claude/skill-context/ask-me.md` so it is not deleted as a double-ask. **`hook_edge.py` is unchanged** — teaching `_ASK_USER_MATCHER` about Bash calls would fire `needs_human` on arbitrary Bash invocations and is a named No-Go.

**The claim must not permanently swallow a question.** The claim value is the ISO claim timestamp, never `1`, so staleness is readable. On a lost claim: younger than one slow interval is a genuine concurrent peer (return); older with `dispatched` present re-attempts only the completion marker; older with nothing dispatched **takes the claim over** — the bridge-death state, where no `except` handler ever ran. `mark_poll_dispatched` runs immediately after the steer and before anything else that can throw, and the handler releases the claim **only when it is absent**; a blanket release re-opens one-vote-two-enqueues on the mainline.

One residual is **named and accepted**: `mark_poll_dispatched` is one Redis command after the steer, so a death inside that window can produce a second dispatch. Accepted because the alternative is never taking over — the permanent swallow — and a duplicate resume is degraded-but-safe while a swallowed question is not. `poll_claim_takeover` logs it so it is diagnosable rather than inferred.

**Three markers, each its own key.** Fields in one JSON value would make every marker write a read-modify-write on a shared key; the reconcile loop's warn write races the fast path's marker writes, and the claim serializes *translation*, not the loop's own scan-side write.

**Commit hygiene.** Three changes land as their own commits because reverting the poll feature must not revert them: `d3d4ddfab` (the `answer_routing` extraction, which restructures the reply-to ladder every typed reply travels), `f598a991d` (the role primes, which govern every session on every machine), `f76cdc87c` (the nudge-loop fix, which outlives the feature).

## Plan corrections made during BUILD

Both found by building rather than reading; both on `main`:

- `f27692e0a` — the 8-byte ceiling (above).
- `f277ea874` — **three Verification anti-criteria flag correct code.** `grep "scan_iter"` matches `sscan_iter`, the *approved* mechanism; the router-purity grep matches the docstring naming the caller's job — the very contract it enforces; the heartbeat-literal grep penalises having named the constant. Converted to AST/behavioral assertions, with the general rule stated: **a property about imports, calls, or ordering is an assertion, not a substring search.**

## Tests

**413 unit + 255 integration/handler, all green.** New: `test_poll_gating.py` (21), `test_poll_payload.py` (36), `test_poll_registry.py` (27, against real Redis — the properties are Redis semantics), `test_poll_vote_translation.py` (24), `test_ask_poll_cli.py` (24), `test_agent_catchup_poll_transcript.py` (6). Updated: `test_output_router.py` (+15), `test_bridge_relay.py` (+12), `test_health_check.py` (+5), `test_read_the_room.py`.

The three named `tests/integration/test_steering.py` cases are the regression net for the `answer_routing` extraction and are green, along with 137 tests across the steering suites.

**No test sends a poll to Telegram.** DM behavior is asserted by checking the CLI queues prose, never by hitting the API — the capability matrix is settled.

## Deploying

`./scripts/valor-service.sh restart` is **required** after merge (new Telethon handler + background loop), and `valor-ask-poll` only materializes after the `uv sync` that `scripts/update/run.py` already performs. Both are already part of `/update`. No migration, no new secrets, no `.env` edit.

## Out of scope, deliberately not built

1:1 DMs, a bot identity, inline keyboard buttons, reaction-based answering, `teammate` sessions, quorum/weighting/voter allowlists, quiz or multiple-choice polls, free-text capture inside the poll, `GetPollVotesRequest` attribution, and auto-rendering a question posed at the bare `needs_human` turn end. Reasons for each are in `docs/features/telegram-poll-questions.md`.


valorengels:
## Review: CHANGES REQUESTED

Reviewed at head `86f1a7499` (resolved via `tools/pr_head_resolver.py`), in an isolated worktree off `refs/pull/3080/head`. Multi-judge consensus, 2 judges (`code-quality`, `risk`); both returned CHANGES_REQUESTED, so `any-blocker-wins` carries. Every finding below was independently re-verified against source before publishing; two judge findings were dropped as unreproducible and are not listed.

This is a genuinely well-built feature. The registry design holds up under reading — the marker-per-key decomposition, the `SET NX` claim with a `dispatched`-guarded stale takeover, and the "yield on an absent `steered_at`, never on a missing claim" rule are all correct, and `tests/unit/test_poll_registry.py` asserts them against a real Redis rather than a mock. The PR description's mechanical claims about the nudge-loop change check out: I read `agent/output_router.py:136-186` and confirmed `pause_open_question` sits behind terminal status, `completion_sent`, post-compaction, watchdog, `rate_limited`, empty-output, the nudge cap and `waiting_for_children`, ahead only of the eng+sdlc line; `has_open_question: bool = False` defaults on both `determine_delivery_action` (:89) and `route_session_output` (:203); and the function stays pure, with the Redis read thread-offloaded in `agent/session_executor.py:1603-1618`.

The blockers are one plan-contract violation and two delivery defects on the degradation paths.

### Gate results

- **`ruff check .`** — clean. **`ruff format --check .`** — 1439 files already formatted.
- **Lane tests** — `test_poll_payload`, `test_poll_gating`, `test_poll_registry`, `test_poll_vote_translation`, `test_ask_poll_cli`, `test_output_router`, `test_bridge_relay`, `test_agent_catchup_poll_transcript`, `test_health_check`, `test_read_the_room`, `test_config_driven_routing`: **345 passed**.
- **Full `tests/unit`** — 14598 passed, 38 failed. I ran the 38 against `origin/main` in a second worktree and got an identical 28 failed / 305 passed on the same selection; **all remaining failures are pre-existing, none are regressions from this branch.** (The other 10 were an artifact of my first run picking up the wrong venv and do not exist.)
- **Docs gate** — satisfied. `docs/features/telegram-poll-questions.md` exists, plus the README index entry and the updates to `eng-session-architecture`, `session-steering`, `message-drafter`, `bridge-worker-architecture`, `read-the-room` and `tools-reference` the plan's `## Documentation` section required.
- **Plan section compliance** — all four required sections present.
- **Visual proof gate** — no-op. No HTML/CSS/JS/JSX/Vue/template files in the diff; `ui/app.py` adds JSON health fields only.
- **Verification table** — every `[FAIL]` row is a runner artifact, not a code finding: `grep -c 'A|B'` written without `-E`, and `grep -lc ... | wc -l` rows whose expected value the runner compares against the wrong capture. I hand-ran all nine and every one passes. Worth fixing in the plan so the next run is readable, but nothing here is a defect in the branch.

### Blockers

**1. Task 10a is UNRESOLVED, and the plan makes it a hard shipping gate.**
`docs/plans/ask-me-telegram-polls.md:2724` and the matching Success Criteria bullet: *"This criterion is a hard shipping gate (owner ruling, 2026-09-02): neither FAIL nor UNRESOLVED may be merged past."* The PR body now discloses the gate as UNRESOLVED (two poll windows lapsed, no operator tap), which is the right call — but disclosure is not clearance. Every other check in this plan is a unit test with a stubbed handler or a grep; Task 10a is the plan's only assertion that the nine subsystems are actually wired together, and it is exactly the class of defect the two blockers below turn out to be. This one is not a reviewer's to waive: it needs either the tap, or an explicit owner ruling recorded on #2701 relaxing the gate.

**2. A prose answer never clears the pause, so an eng lane silently loses auto-continue for 24 hours.**
`bridge/poll_registry.py:405` — `session_has_open_poll` returns `True` for any registry row lacking a `steered_at` marker. `mark_poll_steered` has exactly two non-test call sites, `bridge/poll_vote.py:186` and `:216`, both on the vote path; I grepped the whole tree and there is no other caller. Rows live `POLL_REGISTRY_TTL_S`, default **86400s** (`bridge/poll_registry.py:81`).

So: an eng SDLC session asks via `/ask-me`, and the human answers by typing an ordinary reply in the group rather than tapping — the likeliest behavior, especially once the thread has moved on. The reply steers correctly through `bridge/answer_routing.py` and the session resumes and works. But the registry row is untouched, so from that moment until the TTL expires *every* turn of that session takes `pause_open_question` (`agent/output_router.py:180`) instead of `nudge_continue` (`:185`). The pipeline's core unattended-progress mechanism is off for a full day, advancing one turn per human message, and the only signal is an INFO log (`agent/session_executor.py:1641`) plus a `poll_expired_unanswered` warning an hour in that goes to `logs/` and nowhere else.

The blast-radius argument in the PR description ("bounded by the default `False`") is sound only while rows are short-lived. In practice the row's lifetime is decoupled from the question's — it is tied to a tap that may never come.

*Fix:* add `mark_session_polls_steered(session_id)` to `bridge/poll_registry.py` and call it from the reply-to steering ladder and from `resume_completed_session` (`bridge/answer_routing.py:171`), so *any* answer route closes the question. Alternatively gate the pause on a window much shorter than the registry TTL, so routing lifetime and pause lifetime are separately tunable.

**3. An ineligible or hint-less poll is delivered to the human up to four times.**
`bridge/telegram_relay.py:1138` and `:1198` — `_send_queued_poll` returns `None` on both decline branches (missing `poll_id_hint`; ineligible at send time) *after* it has already pushed the prose fallback via `_reenqueue_poll_as_text`. In `process_outbox`, `success = msg_id is not None` (`:1309`) is therefore `False`, so control falls into the generic retry `else` at `:1428`, which re-pushes the **original poll payload** at `:1451` with `_relay_attempts` incremented. On the retry it declines again for the same reason and enqueues another text copy — up to `MAX_RELAY_RETRIES`, plus one more from the terminal branch at `:1444`. Each retry additionally burns a 50-message `iter_messages` scan via `_find_already_sent_poll`, reached because `_relay_attempts > 0`.

The `DELIVERED_NO_ID` sentinel at `:1315` exists for precisely this shape on the text path (#2179) and the poll branch does not use it. `TestPollDispatch` exercises `_send_queued_poll` in isolation and never asserts the `process_outbox` interaction, which is why this is green.

*Fix:* return a delivered-no-id sentinel from the two decline branches so the payload is consumed rather than re-queued.

### Tech Debt

- **`bridge/poll_reconcile.py:153` and `:189`** — the `except FloodWaitError: raise` in `reconcile_once` and the backoff handler in `poll_reconcile_loop` are unreachable. `bridge/poll_vote.py:157` is a blanket `except Exception` wrapping the `GetPollResultsRequest` call and swallows `FloodWaitError` first, returning normally; `bridge/poll_vote.py:236` and `_find_already_sent_poll` do the same. The docstring at `:190` says "mirror the relay's backoff shape", but under a real flood wait the loop logs a warning per poll and keeps issuing `GetPollResultsRequest` at the 5s fast interval for every open row — it accelerates into the throttle instead of backing off, against the bridge account. *Fix:* re-raise `FloodWaitError` ahead of the blanket handler in `poll_vote.py`, or drop these two handlers and own the backoff there.

- **`bridge/telegram_relay.py:1433`** — the terminal-failure poll branch dead-letters and re-enqueues as text but never calls `delete_pending_poll`. Confirmed by grep: the only call sites are `bridge/telegram_relay.py:1197` (the eligibility decline) and `bridge/poll_registry.py:226` (promotion). A poll whose send fails all retries therefore leaves a provisional row alive for the full 24h TTL, and `adopt_orphaned_polls` runs an `iter_messages(limit=50)` history scan for it on **every** reconcile tick — order 1400 scans/day for a poll that provably never landed.

- **`bridge/poll_registry.py:405` / `agent/session_executor.py:1611`** — `session_has_open_poll` iterates `iter_unanswered_polls()`, which reads only `POLL_OPEN_INDEX` (`:375`). A poll enqueued by `valor-ask-poll` but not yet processed by the relay exists only as a provisional row in `POLL_PENDING_INDEX` (`register_pending_poll`, `:203`), which this predicate never consults. The relay is a polling loop and the pause check runs at turn end, so the window is real — and during it the session is reported as having no open question and gets nudged past it, which is the exact defect the branch exists to prevent.

- **`bridge/poll_gating.py:81`** — `sessions.sort(key=lambda s: s.created_at or 0, reverse=True)` compares a `datetime` against `int` when any record has a null `created_at`, raising `TypeError`. The broad handler at `:90` catches it and returns `eligibility_error`, so the question silently degrades to prose. The sibling extraction in this same PR, `bridge/answer_routing.py:91`, uses the correct `(s.created_at is not None, s.created_at)` key — two answers to "which record is newest" inside one feature.

- **`tools/ask_poll.py:39` and `bridge/poll_vote.py:47`** — `ESCAPE_HATCH_OPTION = "Other: wait for followup message"` is defined independently in both modules with no test pinning them equal; each test file imports its own module's copy. If either literal drifts, `build_steer_text`'s `if chosen == ESCAPE_HATCH_OPTION` (`poll_vote.py:100`) silently stops emitting the followup instruction and a "none of these fit" tap resolves as an ordinary answer, with no error anywhere. *Fix:* define once, import in the other.

- **`bridge/poll_reconcile.py:105-204`** — `reconcile_once`, `adopt_orphaned_polls`, `warn_if_expired_unanswered`, `_interval` and `poll_reconcile_loop` have no test coverage; a grep across `tests/` finds only `heartbeat_age_s`/`write_heartbeat` in `test_health_check.py`, and `_find_already_sent_poll` is patched out in every relay test. The module's own docstring calls it "the PRIMARY inbound mechanism, not a backstop", and the adaptive interval, the ambiguous-adoption bail and the once-per-poll warn dedup are all behavioral claims with nothing verifying them.

- **`bridge/telegram_relay.py:1428-1447`** — on terminal failure the question is delivered by two independent mechanisms: `_dead_letter_message` persists it (with `render_poll_as_text` supplying the missing `text`) and `_reenqueue_poll_as_text` immediately re-queues the same prose, after which `replay_dead_letters` re-sends it on the next bridge connect. The comment explains why the immediate re-enqueue was added but does not address the resulting duplicate.

- **`bridge/poll_vote.py:178`** — the takeover threshold reuses `POLL_RECONCILE_SLOW_INTERVAL_S` (60s), but `_dispatch_answer` awaits `get_messages`, `close_poll`, `resolve_answer_target` (three unindexed `AgentSession.query.filter` scans) and a full `dispatch_telegram_session` *before* `mark_poll_dispatched` runs. A dispatch slower than 60s — no process death required, just a loaded Redis or a slow scan — lets a concurrent tick see `age >= 60` and `not poll_dispatched`, take the claim over and steer twice. This is wider than the post-steer residual the PR names. *Fix:* derive the threshold from a dispatch-duration budget, or stamp an in-progress marker before the first `await`.

- **`agent/session_executor.py:1636-1646`** — the pause branch delivers output but does not stamp `agent_session.response_delivered_at`, unlike the `deliver` branch at `:1795`. That field is the health check's delivery guard: without it a session stuck in `running` falls through to the worker-alive / no-progress recovery path and can be re-queued, re-delivering its output. Shared with `deliver_already_completed` and `deliver_fallback` so not newly introduced, but pause is now the mainline exit for an asking session.

- **`agent/output_handler.py:1560-1566`** — `validate_poll_question` violations are logged at warning and discarded; the poll is queued regardless. The only real enforcement is `tools/ask_poll.py:134`, which hard-exits on the same length rule. Two enforcement points for one invariant, one of which does not enforce.

### Nits

- **`bridge/poll_registry.py:103`** — `POLL_PROBE_TAP_WAIT_S` has no reader in shipped code (verified by grep: only its own definition, the plan, and `docs/features/telegram-poll-questions.md:408`). It was a manual-gate probe knob, and the feature doc lists it as an operator-adjustable tunable, which is misleading — setting it does nothing. Delete the constant and the doc row.
- **`bridge/agent_catchup.py:403`** — `text = m.text or "" or _poll_transcript_text(m)`. The middle `or ""` can never be the value of the expression. Vestigial; the test at `test_agent_catchup_poll_transcript.py:66` is even named after the confusion.
- **`tools/ask_poll.py:171`** — `if not hasattr(handler, "send_poll")` is evaluated against a concrete `TelegramRelayOutputHandler()` constructed two lines above, which always defines it. Unreachable; the capability-probe rationale applies to the Protocol, not this call site.
- **`bridge/telegram_bridge.py:3391` and `bridge/poll_reconcile.py:179`** — both log `"Poll reconciliation loop started"`, so the line appears twice per bridge start.
- **`bridge/poll_reconcile.py:147` and `:185`** — `iter_unanswered_polls()` is walked twice per tick, the second time only to compute the youngest row age for `_interval`. Have `reconcile_once` return it.
- **`bridge/poll_reconcile.py:34`** — `_redis` is a private name imported across a module boundary; `write_heartbeat`/`heartbeat_age_s` are its only consumers.
- **`bridge/poll_vote.py:301-307`** — on the `NONE` branch (no session record for the row) the poll is marked dispatched and closed out, but the human's tap has already closed the poll on screen at `:235`, so from their side the answer looks accepted while it was dropped with only an INFO log.

### Restart required after merge

Confirmed and already stated in the PR: `./scripts/valor-service.sh restart` (new Telethon `events.Raw` handler plus the background reconcile loop), and `valor-ask-poll` only materializes after the `uv sync` that `scripts/update/run.py` performs. Both are part of `/update`. No new env keys, no Popoto model, so no `scripts/update/migrations.py` entry — verified.

---

Blockers 2 and 3 are both narrow, well-localized fixes. Blocker 1 is a contract question for the owner, not a code change.

valorengels:
## PR Review — APPROVED

Reviewed at head `0e7c0b810`. Two independent lenses (correctness/safety, and plan conformance against the recorded contract in `docs/plans/ask-me-telegram-polls.md`), plus a legacy-cruft audit of the branch diff.

**Verdict: APPROVED.** No blockers. One CONCERN was raised and has been fixed in this branch.

### CONCERN raised and fixed — `0e7c0b810`

`promote_pending_poll()` discarded `register_poll()`'s `SET NX` result, so a false-positive adoption match silently swallowed a question.

Orphan adoption matches on the truncated 7-byte correlation prefix. `_find_already_sent_poll` only guards the ambiguous 2-or-more-match case, so a lone false-positive match returned cleanly. Promotion then no-opped on the NX write, deleted the pending row anyway, and reported success — leaving that question permanently unroutable (no pending row, no real row) while `_send_queued_poll` returned the *other* poll's `msg_id` as though the question had been delivered.

Low probability in practice (the hint is a random `uuid4().hex` truncated to 56 bits, scanned over a bounded 50-message window), but the failure mode is silent question loss, which is the one outcome this feature exists to prevent. Fixed: a refused NX write is now an adoption failure — the pending row survives so the question stays adoptable, the relay falls through and actually sends rather than claiming a delivery that never happened, and the reconciler only logs `poll_orphan_adopted` on a real adoption. The comment at `bridge/response.py:558` now describes the two-layer guarantee accurately instead of overclaiming.

### Verified, no findings

- **Vote misrouting is not reachable through the correlation prefix.** The primary vote path never uses it: `translate_poll_vote` looks up the server-assigned `poll_id` directly via the registry, and `select_option` reads only the one-byte index. A prefix collision can orphan an unsent question (the fix above); it cannot deliver a vote to the wrong session.
- **Idempotency.** `claim_poll_answer` / `takeover_poll_claim` / `mark_poll_dispatched` / `mark_poll_steered` are independent `SET NX` keys. Duplicate votes, a dead-bridge takeover, and the `events.Raw` fast path racing the reconcile loop all funnel through one idempotent `translate_poll_vote`. A vote after TTL expiry is a clean no-op, not a crash.
- **`pause_open_question` guard order** is exactly as specified: after terminal, completion-sent, post-compaction, watchdog, rate-limit, empty-output, nudge-cap and `waiting_for_children`; before the eng+sdlc line. `has_open_question` defaults `False`, and is re-read every callback, so an answered or expired poll releases the pause — no permanently stuck session.
- **Blast radius of the output_router change.** `tests/unit/test_output_router.py` is a pure addition: 140 insertions, zero deletions. No pre-existing case was modified and all 33 pass.
- **Redis discipline.** `bridge/poll_registry.py` operates on plain `telegram:poll:*` string keys, not Popoto-managed keys; `AgentSession` access goes through the ORM throughout.
- **Inbound resilience.** The `events.Raw` handler wraps its body and logs without re-raising, so a translation failure cannot take down the Telethon update loop.
- **No-Gos respected.** No inline keyboards, no bot identity or webhook, no reaction voting, no DM or teammate-session polls; `public_voters` stays `False` and `GetPollVotesRequest` appears only in a docstring explaining why attribution is not built.
- **Cruft audit clean** — no commented-out code, no shims, no unnamed magic numbers, no stray debug output.

### Disclosed, weighed, accepted

Task 10a (`gate-poll-e2e`) is **UNRESOLVED**: two poll windows lapsed with no operator tap, so the last hop (observed vote to translated steer to next turn input) has never been witnessed in production, only under test. Per the plan's satisfied-by-pause rule this is not a FAIL and does not block — a FAIL requires a tap that produced no steer, which did not occur. The PR body now states this outright rather than carrying the original placeholder.

### Test evidence (re-measured in this environment, not forwarded)

`./scripts/pytest-clean.sh -n 2` over `test_poll_registry`, `test_bridge_relay`, `test_poll_vote_translation`, `test_poll_gating`, `test_poll_payload`, `test_ask_poll_cli`, `test_output_router`, `test_config_driven_routing`, `test_agent_catchup_poll_transcript`: **269 passed** (268 before the fix, plus its one new regression test). `ruff check` and `ruff format --check` clean on all touched files.

One unrelated pre-existing failure exists on `main` and is **not** a regression from this branch: `tests/integration/test_steering_media_smoke.py::test_txt_attachment_steering_enriches_and_ingests` fails with `TypeError: _spy_push() got an unexpected keyword argument 'room_id'` — a test spy whose signature has drifted from production. Confirmed identical on `main` at `bdf1c1144`.

Issue #3076, opened 2026-09-02
Nightly autonomous fixer: active mode (dispatch, hand-back, watchdog, enforcement) — blocked on two missing seams


valorengels:
Follow-up split out of #2334 during the round-7 plan revision.

#2334's plan is re-scoped to what is buildable today: the alert-transport returncode fix, the in-process baseline classifier, and the `off`/`shadow` decision gate. Everything that would let the nightly detector actually *attempt a fix* is deferred here, because it depends on two seams that **do not exist** and were repeatedly asserted without verification across critique rounds 1, 5, and 6.

## Blocked on

**1. No per-session environment seam for the fixer's `gh` identity.**
The never-merge guarantee requires the fixer session's `gh` to authenticate as a non-admin bot (`SDLC_AGENT_GH_TOKEN`), not the operator's ambient admin credentials. There is no way to do that today:
- `scripts/nightly_regression_tests.py` L909-927 spawns `python -m tools.valor_session create` with **no `env=` kwarg**, and that CLI only *enqueues* an `AgentSession`.
- The **worker**, a separate launchd process, later spawns `claude -p`. Its env overlay is `agent/session_runner/role_driver.py::subscription_auth_env` (L76-100), called at L204 — a process-global overlay derived from the worker's own `os.environ`, merged into `proc_env` at `agent/session_runner/harness/claude.py` L431-433. Nothing is per-session.
- `models/agent_session.py` has **no** env/token/permission field.

Building this means: a per-session env override persisted on `AgentSession` (or a reserved `extra_context` key), read by the worker, threaded through `TurnRequest.env` (`agent/session_runner/harness/base.py` L52) into the `claude -p` spawn — with its own tests. That is a standalone piece of work, not a dispatch-site one-liner.

**2. Branch protection on `main` is absent.**
`gh api repos/tomcounsell/ai/branches/main/protection` → `404 Not Found` (re-verified 2026-09-02). Enabling review-required protection (`required_approving_review_count >= 1`, `enforce_admins` left at its default `false`) is a repo-governance action for the owner. Without it, seam 1 buys nothing.

## Scope when unblocked

- Narrow-mandate draft-PR fixer dispatch (`maybe_dispatch_fix_session`), with `GH_TOKEN` reaching the session.
- `AgentSession.nightly_fix_handback` nullable JSON field + `tools/nightly_fix_handback.py` + a `MIGRATIONS` no-op entry.
[...]

Issue #3033, opened 2026-08-27
Worktree without .venv silently tests the primary checkout, so branch-local regressions read green


valorengels:
## Problem

A git worktree under `.worktrees/` that has **no `.venv` of its own** silently resolves imports from the **primary checkout**. Tests run there exercise `main`'s code, not the branch's, and a branch-local regression reads green.

Found while verifying PR #3028. Reproduced:

```
$ ls -d .worktrees/wave1-hotfix-sweep/.venv
NO .venv in worktree

$ cd .worktrees/wave1-hotfix-sweep
$ python -c "import tools.push_ancestry_guard as g; print(g.__file__)"
/Users/valorengels/src/ai/tools/push_ancestry_guard.py     # <-- PRIMARY checkout
```

With `PYTHONPATH` pinned to the worktree it resolves correctly, and a test that had been passing fails:

```
$ PYTHONPATH=$PWD python -m pytest tests/unit/test_push_ancestry_guard.py -q
FAILED tests/unit/test_push_ancestry_guard.py::TestPushTargetParsing::test_empty_stdin_falls_back_to_head
1 failed, 16 passed
```

Without the pin the same file reported `17 passed`.

## Why this matters more than one PR

This is a **verification-integrity** bug, the same family as Wave 0 in `docs/bug-backlog-waves.md`. A lane that runs its suite in a worktree and reports "N tests pass" may be reporting on `main`. In PR #3028's case it produced a confidently false claim ("1545 unit tests pass") in a PR body, and the failing test was only caught by an independent reviewer who happened to force `PYTHONPATH`.

Any lane, on any branch, is exposed. The failure is silent and biased toward green, which is the worst direction: it hides exactly the regressions the run exists to catch.

## Why the existing guard does not catch it

`docs/features/worktree-venv-isolation.md` and `scripts/pytest-clean.sh` guard the **off-pin venv** case — a venv on the wrong interpreter aborts the run. There is no guard for the **absent venv** case, which degrades silently to the primary checkout's environment instead of failing closed.

The editable install is what makes it silent: the primary venv carries an editable path entry pointing at `/Users/valorengels/src/ai`, so `import tools...` always finds a real module and never raises.

## Suggested direction

[...]

PR #3082, opened 2026-09-02
feat(nightly): classify newly-confirmed failures before paging (shadow tier)

### LINKED ISSUE #2334: Nightly regression: attempt autonomous fix before paging a human (alert = last resort, not first responder)
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/nightly-autonomous-fix-before-alert.md

> **Context.** The *nightly regression detector* (`scripts/nightly_regression_tests.py`) runs the test suite each night and, on a newly-confirmed failure, sends a Telegram alert to a human. PR #2195 (under plan issue #2192) added a *triage dispatch* step: `maybe_dispatch_triage_session()` fires one Eng-role AgentSession whose mandate is to *investigate and file a `/do-issue`-quality issue* — explicitly **not** to auto-fix (auto-hotfix was deferred as a No-Go in that plan). See the feature doc shipped with that PR: `docs/features/nightly-alert-triage.md`.

## Problem

The human is currently the **first responder**, not the last resort.

On a newly-confirmed nightly failure, the flow today is:
1. Detector confirms the regression.
2. `maybe_dispatch_triage_session()` spawns an Eng session to *investigate and file an issue*.
3. `send_telegram(...)` pings a human **up front**, in the same run — before the dispatched session has done anything.

So the human gets paged at the moment of detection, carrying the full cognitive load of "what broke and what do I do," even for regressions that Valor could diagnose and fix autonomously. The triage session, meanwhile, is scoped only to *file an issue* — it never attempts the fix.

**Current behavior:** Detection → dispatch investigate-only session **and** page a human immediately.

**Desired outcome:** Detection → Valor attempts an autonomous fix → a human is paged **only** if Valor hits a genuine blocker it cannot clear on its own (e.g. an architectural decision, a product judgment call, an ambiguous spec, or missing access/credentials). The human alert becomes the **escalation of last resort**, and when it fires it should arrive at the *final stuck point* — stating what was tried, what failed, and the specific decision needed — not a raw "tests are red" ping.

## Why this matters

An alert that lands before any attempt at self-repair trains the human to be the debugger. The system already has the machinery to attempt a fix autonomously (an Eng AgentSession running the SDLC pipeline). The value of paging a human is highest exactly when Valor is genuinely stuck on something only a human can decide — and lowest when it's a mechanical regression Valor could have patched before anyone woke up.

## Definitions

| Term | Definition |
|------|------------|
| Nightly regression detector | `scripts/nightly_regression_tests.py` — runs the suite nightly, alerts only on newly-confirmed serial failures. |
| Triage dispatch | `maybe_dispatch_triage_session()` — spawns one deduped Eng-role AgentSession per new failure set; today its mandate is *investigate → file an issue*, not fix. |
| Newly-confirmed failure | A failure that survives the serial re-confirmation gate (#2180), i.e. not an xdist-parallelism artifact. |
| Escalation of last resort | A human alert that fires only after autonomous repair has been attempted and hit a blocker a human must resolve. |

## Open questions (root cause / design is uncertain — for `/do-plan` to resolve)

This is a flow/architecture reframe, not a known one-line fix. The plan should resolve, not assume:

1. **Fix mandate vs. No-Go.** The original plan deferred auto-hotfix deliberately. Does this issue *reverse* that No-Go, or introduce a *bounded* autonomous-fix attempt (e.g. propose-a-patch-as-PR-but-don't-merge) that stops short of the original concern? What was the original reason auto-fix was deferred, and does it still hold?
2. **What defines "genuinely stuck"?** How does the dispatched session signal "I tried and I'm blocked, escalate now" vs. "I fixed it, stay silent" vs. "I'm still working"? Is this a session terminal-state signal, a structured hand-back, or a timeout?
3. **Alert timing.** Should the up-front `send_telegram(...)` be removed entirely and replaced by a *session-driven* escalation alert, or downgraded (e.g. a silent log / low-priority signal now, human page only on escalation)? What happens if the session dies or hangs — does the human still eventually get paged (fail-safe), and after how long?
4. **What the escalation alert should say.** The last-resort alert should carry the stuck point: what broke, what Valor tried, why it's blocked, and the specific decision requested. Where does that content come from — the session's own hand-back?
5. **Interaction with the Cowork migration.** Sentry-issue-triage recently migrated to Claude Cowork (#2209). Should nightly-regression auto-repair follow the same Cowork routine pattern, or stay an in-process AgentSession dispatch?
6. **Loop/cost guardrails.** An autonomous fixer that runs nightly needs bounds (max attempts, no infinite re-dispatch, dedupe on already-attempted failure sets). What are the caps?

## Downstream

This issue will be consumed by `/do-plan` to produce a plan document at `docs/plans/{slug}.md`, then executed by `/do-build`. Because the root cause and correct mandate are genuinely open, the plan should treat the questions above as its scoping work rather than executing a pre-baked sketch.

Builds on #2192 / PR #2195 (which must merge first — this issue reframes the flow that PR establishes).
## Recon Summary

*Added by `/do-plan` (run 97a27d4b) after direct code investigation at HEAD 66a433bd9.*

**Confirmed (grounded in code):**
- `scripts/nightly_regression_tests.py::main()` pages a human up front on any newly-confirmed failure branch (`send_telegram(...)` at the `elif new_failures:` block, ~L645-666) *in addition to* dispatching `maybe_dispatch_triage_session()` — the "human is first responder" problem is real and located here.
- `maybe_dispatch_triage_session()` (L437-509) spawns an Eng session whose prompt is explicitly investigate-and-file-an-issue, "Do NOT attempt an auto-hotfix" (L473-474). Dispatch is fire-and-forget with a `dispatched_hash` sha256 dedup key persisted in `data/nightly_tests_last_run.json` (L463-466, carried forward at L628-629).
- Serial re-confirmation gate (`reconfirm_serial`, L212-256) already isolates real regressions from xdist artifacts; `confirmed_failing` is the authoritative failure set.
- `baseline-verifier` subagent (`.claude/agents/baseline-verifier.md`) exists and does exactly regression-vs-pre-existing classification against `main`, returning structured JSON (`regressions`/`pre_existing`/`inconclusive`). Reusable as-is for the test-stale-vs-code-regressed gate.

**Revised / clarified:**
- There is **no** `is_escalated` field on `AgentSession` (that field belongs to `models/crash_signature.py`). The dispatched fixer's terminal hand-back (fixed-silent / blocked-escalate / still-working) must be a **newly designed** signal — the plan defines its shape.
- The worked example #2399 → PR #2402: all 11 failures were test-stale, but Group 2 hid a genuine **design fork** (self-heal intended vs. safety hole) that a human resolved. Lesson for scoping: "classifies as test-stale" is necessary but not sufficient for autonomy — a test-stale carry-forward that requires a *contract-direction judgment* must still escalate.

**Pre-requisites:** #2192 / PR #2195 (merged 2026-07-24) established the triage-dispatch flow this issue reframes — satisfied.

**Dropped:** none.

valorengels:
## Summary

Ships the **shadow tier** of the nightly regression classifier: every non-`off` nightly run now computes and logs the verdict it *would* act on, while paging a human exactly as today. It does **not** ship the autonomous fixer — that is #3076, blocked on two seams that do not exist on this machine (no per-session env seam for a fixer's `gh` identity; `main` branch protection absent).

Plan: `docs/plans/nightly-autonomous-fix-before-alert.md`

## What landed

- **`send_telegram` returncode check** — a rejected or failed send no longer logs `"Telegram sent"`. Independently valuable.
- **`_spawn_pytest` gains `cwd: Path | str = PROJECT_DIR`** — the seam a second caller needs to target a different tree. Both existing callers are byte-identical.
- **Provisioned persistent baseline worktree** at `.worktrees/nightly-baseline/` with its own `.venv`, re-pointed to `prev["head_commit"]` each run and re-synced only when `uv.lock` moves. Every provisioning subprocess is bounded by an explicit `timeout=`.
- **`classify_against_baseline(...)`** — synchronous, in-process, no subagent or session. Re-runs only the failing node IDs at the prior run's HEAD SHA through the real `pytest-clean.sh` wrapper, into its **own** report path (`PYTEST_BASELINE_JSON_TMP`), and buckets each node `newly_broken` / `pre_existing` / `inconclusive`. Every failure path lands in `inconclusive` — there is no `PROJECT_DIR` fallback.
- **Pure `decide_fix_or_escalate(...)`** with the eight-clause short-circuit whose order defines the `reason=` token vocabulary. The five refusable preconditions are enforced at the call site *before* any git or pytest work.
- **Greppable shadow verdict** — `nightly-fix shadow-verdict: {autonomous-fix|escalate} reason=... nodes=...` on every non-`off` run with non-empty `new_failures`, including precondition-skipped nights, plus a `nightly-fix shadow-buckets:` sibling line when classification actually ran.
- **Two env-overridable guardrail constants** (`NIGHTLY_FIX_MODE` default `shadow`, `NIGHTLY_FIX_MAX_FAILURES` default 15), declared in `.env.example` as `@optional` blocks that the real `/update` parser recognizes.
- **`PROTECTED_WORKTREE_SLUGS`** in `tools/disk_reclaim.py::sweep_worktrees` — the baseline lane is genuinely in that sweeper's scope and survived only by guard-order accident. Now skipped with reason `protected` as the first guard.

## Behavior guarantee

The alert **text** is unchanged: the up-front page fires in both shipped modes with byte-identical text, asserted via `send_telegram` call capture, and a crash anywhere in the shadow tier is swallowed non-fatally with the page still firing (asserted). Because classification runs before the page, on a failing night the page is delayed by up to the classification bound (provision timeouts + baseline pytest timeout). `data/nightly_tests_last_run.json` gains no new key. No schema change, no migration, no CLI surface change.

## Testing

258 tests pass across `test_nightly_classifier.py` (new), `test_nightly_decision_gate.py` (new), `test_nightly_regression_tests.py` (added only), and `test_disk_reclaim.py` (added only). No existing assertion was modified.

The load-bearing one is the **falsifiability suite**: a two-commit fixture repo built as a *linked worktree* (so the #3033 `.venv` guard is live), carrying `[project]` + `[tool.pytest.ini_options]` (so the wrapper resolves `REPO_ROOT` to the fixture, not the real repo), with a venv carrying `pytest`, `pytest-json-report` **and** `pytest-xdist`. It drives the real `classify_against_baseline` through its keyword-only injection seam with no monkeypatching of module globals, and discriminates: `newly_broken` for a node passing at commit A and failing at HEAD, `pre_existing` for a node failing at both. Three guard tests assert the fixture's own properties so a future edit cannot silently degrade it into a test of the real repo or into one that can only say `inconclusive`.

`python -m ruff check .` and `python -m ruff format --check .` both clean.

## Note for review

Six rows of the plan's `## Verification` table are not machine-runnable as written: the command cell carries trailing prose after an em-dash (the shell then sees an unterminated backtick), and the evaluator does not parse the `≥` / `== N` / `empty output` expectation forms. Every one of those rows was **run manually and passes**. The table itself is a plan-authoring artifact, not a code defect, and was deliberately not edited mid-build so the plan-hash guard stayed armed.

Closes #2334


valorengels:

**Verdict:** CHANGES REQUESTED · blockers=1 · tech_debt=3 · nits=4

The new logic is sound and the tests are not vacuous. The falsifiability suite was re-run independently (`-k "fixture_repo or fixture_preconditions"` → 6 passed in 64s, no silent skip) and it genuinely drives `classify_against_baseline` end to end against a two-commit linked-worktree fixture with its own `.venv` through the real `pytest-clean.sh`. The three guard tests really do assert the fixture's discriminating properties. Bucketing is conservative in the correct direction (only `passed` → `newly_broken`; `skipped`, absent, or any exception → `inconclusive`), `classify_precondition_reason` and `gate_reason`'s first five clauses share one implementation so the `reason=` vocabulary cannot diverge, report-path isolation from `PYTEST_SERIAL_JSON_TMP` is real and asserted, and `resolve_fix_mode` fails toward `off`.

**Blocking:** the shadow tier is invoked before `send_telegram` on the failing-night path, so only the alert *text* is unchanged, not the alert. The PR body's "Behavior guarantee: Alerting is **unchanged**" and the same claim in `docs/features/nightly-alert-triage.md` and `docs/features/nightly-regression-tests.md` overstate this, and the inline comment's stated reason ("runs BEFORE the alert is built because the alert re-reads `PYTEST_SERIAL_JSON_TMP` below and the classifier owns a separate report path") is self-refuting — the separate report path is precisely what makes the ordering free.

**Tech debt:** the `seed_run` precondition is unreachable from the only production call site (the shadow block lives in the `elif new_failures:` arm of `if is_seed_run:`) while docs present it as live precondition 3 of 7; `provision_baseline_worktree` has no `git worktree prune` recovery from a registered-but-missing lane; `docs/features/nightly-regression-tests.md:286` calls the provision marker "(gitignored)" and it is not.

**Nits:** the gate clause chain is evaluated twice in `log_shadow_verdict` (`decide_fix_or_escalate` is defined as `gate_reason(...) == "none"`); the `_flags()` helper is byte-duplicated across the two new test files; a redundant per-test `LOG_FILE` monkeypatch shadows the autouse `_quiet_log` fixture; `log_shadow_verdict` does not thread the classifier's keyword-only injection seam, so verdict-path tests must patch the module attribute.

<sub>Aggregate verdict and the full finding blocks are in the `## Review:` comment on this PR.</sub>

valorengels:

**Verdict:** CHANGES REQUESTED · blockers=2 · tech_debt=3 · nits=1

**Verified clean:** `_spawn_pytest`'s `cwd` default preserves both existing callers, and both shapes are asserted; the `send_telegram` returncode branch has no consumer outside the test suite (nothing else greps `"Telegram sent"`); `PROTECTED_WORKTREE_SLUGS` is correctly the *first* guard in `sweep_worktrees`, ahead of `_tree_stats`, with the `protected` reason documented in the skip-reason table; disk cost is one fixed path with no per-run growth; no new credential surface (`baseline_sha` and node IDs are literal argv, no shell).

**Blocking, both operational:**

1. **The kill switch is inert on the only surface that matters.** `NIGHTLY_FIX_MODE` / `NIGHTLY_FIX_MAX_FAILURES` are resolved at module import, but `.env` only reaches `os.environ` via `load_env_or_die()` inside `main()`. The nightly launchd plist supplies only `PATH` and `HOME`. So under the real nightly run both always resolve to the in-code defaults, and putting `NIGHTLY_FIX_MODE=off` in the vault `.env` — the only place CLAUDE.md sanctions — changes nothing. This PR is what publishes those keys in `.env.example` as operator-facing knobs, and the plan's "Reversibility: total" claim rests on them.

2. **The shadow path is the only unguarded pre-page hook in `main()`.** No `try/except` at the call site and no outer handler on `sys.exit(main())`. Inside, `provision_baseline_worktree` is called *outside* `classify_against_baseline`'s guarded `try`, and its own handlers are narrow — `marker.read_text()` catches `OSError` only, so a corrupt marker raises `UnicodeDecodeError` straight through to process death, on exactly the night there are new confirmed failures, with no Telegram sent. The file's own convention for an optional hook is the opposite: the TTFT gate is wrapped `except Exception ... (non-fatal)` and runs *after* the page.

**Tech debt:** `subprocess.run(timeout=)` in `_run_provision_step` kills only the direct child then blocks in `communicate()` — the exact mechanism `_spawn_pytest`'s docstring exists to defeat, so a surviving `uv` build grandchild holding the pipe hangs the nightly indefinitely before the page; provisioning has no self-heal from a desynced worktree admin entry, wedging the classifier at permanently `inconclusive`; a node ID that does not exist at the baseline SHA (a newly *added* failing test, the most common shape) makes pytest exit as a usage error and poisons the whole batch to `inconclusive` rather than just that node.

**Nit:** no `--` separator before `baseline_sha` in either `git` invocation.

<sub>Aggregate verdict and the full finding blocks are in the `## Review:` comment on this PR.</sub>

valorengels:
## Review: CHANGES REQUESTED

Multi-judge consensus (`any-blocker-wins`), 2/2 judges returned, no disagreement — both `code-quality` and `risk` independently returned CHANGES REQUESTED. Per-judge comments are posted above.

This is careful, well-argued work. The falsifiability suite is the real thing: independently re-run, it drives `classify_against_baseline` end to end against a two-commit linked-worktree fixture with its own `.venv` through the real `pytest-clean.sh`, and its three guard tests assert the fixture's own discriminating properties so it cannot silently degrade into a test of the main repo. Report-path isolation, the shared precondition/gate implementation, the no-`PROJECT_DIR`-fallback rule, and the first-guard placement of `PROTECTED_WORKTREE_SLUGS` are all correct and asserted.

Two blockers, both about what happens to the human page on a failing night, and both small and local to fix.

**Reproduced independently in this reviewer's environment (not taken from the PR body):** 258 tests pass across the four affected files (`74.61s`); `python -m ruff check .` → `All checks passed!`; `python -m ruff format --check .` → `1421 files already formatted`.

### Blockers

- **File:** `scripts/nightly_regression_tests.py:228,237` (verified: read this file)
  **Code:** `NIGHTLY_FIX_MODE = os.environ.get("NIGHTLY_FIX_MODE", "shadow")` / `NIGHTLY_FIX_MAX_FAILURES = int(os.environ.get("NIGHTLY_FIX_MAX_FAILURES", "15"))`
  **Issue:** Both knobs are resolved at **module import**, but `.env` only reaches `os.environ` via `load_env_or_die()` at `scripts/nightly_regression_tests.py:1434`, inside `main()`. The nightly launchd job supplies no such variable — `plutil -p ~/Library/LaunchAgents/com.valor.nightly-tests.plist` shows `EnvironmentVariables` containing exactly `HOME` and `PATH`. So on the nightly critical path both always resolve to the in-code defaults, and `resolve_fix_mode()` reads the already-frozen module constant. This PR is what publishes both keys in `.env.example:660-670` as operator-facing knobs, and the plan's "Reversibility: total — `NIGHTLY_FIX_MODE=off` skips classification and the gate entirely" rests on them. Setting `NIGHTLY_FIX_MODE=off` in the vault `.env` — the only place CLAUDE.md sanctions — has zero effect on the job it exists to disable. The feature ships `shadow` to every machine on the next `/update` with an off switch that does not work.
  **Severity:** blocker
  **Fix:** Resolve both values at call time, after `load_env_or_die()` has run: have `resolve_fix_mode()` read `os.environ.get("NIGHTLY_FIX_MODE", "shadow")` itself, and build `GateCaps(max_failures=int(os.environ.get("NIGHTLY_FIX_MAX_FAILURES", "15")))` at the call site. Guard the `int()` cast so a malformed value degrades to the default rather than raising at import. Add a test that sets the env var *after* import and asserts `main()` honors it.

- **File:** `scripts/nightly_regression_tests.py:1639-1656` (verified: read this file)
  **Code:** `if resolve_fix_mode() != "off":` / `    log_shadow_verdict(` … (no `try`/`except` at the call site)
  **Issue:** This ~370-line new code path runs on the `elif new_failures:` arm **before** `send_telegram`, unguarded, and `sys.exit(main())` at L1815 has no outer handler. Exception coverage inside is incomplete: `classify_against_baseline` calls `provision_baseline_worktree` **outside** its guarded `try`, and that function's handlers are narrow — L743-746 `synced_digest = marker.read_text().strip()` catches `OSError` only, so a corrupt or non-UTF-8 marker raises `UnicodeDecodeError` (a `ValueError`) straight through and kills `main()`. The result is a *suppressed* human page on exactly the night `new_failures` is non-empty — the failure mode this whole feature is meant to improve. The file's own convention for an optional hook is the opposite: the TTFT gate is wrapped (`log(f"TTFT gate hook error (non-fatal): {exc}")`, L1707 and L1778) **and** it runs after the page.
  **Severity:** blocker
  **Fix:** Wrap the `log_shadow_verdict(...)` call in `try: ... except Exception as exc: log(f"nightly-fix shadow tier error (non-fatal): {exc}")`, mirroring the TTFT gate. Add a test that patches `classify_against_baseline` with `side_effect=RuntimeError` and asserts `send_telegram` is still called once with unchanged alert text and `main()` returns 0. Also widen L745 to `except (OSError, ValueError)`.

### Tech Debt

- **File:** `docs/features/nightly-regression-tests.md:286` (verified: read this file)
  **Code:** `| `.worktrees/nightly-baseline/.nightly-baseline-provisioned` | Marker recording the `uv.lock` SHA-256 digest … (gitignored) |`
  **Issue:** The marker is **not** gitignored. Run from inside a linked worktree, `git check-ignore -v .nightly-baseline-provisioned` exits 1 (no match): the root `.gitignore:470` entry `.worktrees/` describes the primary checkout's *view* of that directory, not a file at the linked worktree's own root. In production the baseline lane will permanently carry an untracked file, making `git status --porcelain` non-empty for that lane — an `uncommitted_changes` signal that only the newly-added `protected` guard now pre-empts.
  **Severity:** tech_debt
  **Fix:** Either add `.nightly-baseline-provisioned` to `.gitignore` (making the doc true) or drop the `(gitignored)` claim from that row. The `.venv` claim on the preceding row is correct and can stay.

- **File:** `scripts/nightly_regression_tests.py:1629,1651` (verified: read this file)
  **Code:** `if is_seed_run:` … `elif new_failures:` … `RunFlags(is_seed_run=is_seed_run,`
  **Issue:** The `seed_run` precondition is structurally unreachable from the only production call site: the shadow block lives in the `elif` arm of `if is_seed_run:`, so `is_seed_run` is provably `False` whenever `RunFlags` is constructed. `reason=seed_run` can therefore never appear in a real nightly log, yet `docs/features/nightly-regression-tests.md` lists it as precondition 3 of "the seven classification preconditions" with no note, and two tests assert behavior production cannot reach.
  **Severity:** tech_debt
  **Fix:** Keep the clause as defense-in-depth (it is cheap and it keeps `classify_precondition_reason` and `gate_reason` sharing one list) but say so — one sentence in `classify_precondition_reason`'s docstring and in the docs list: unreachable from today's call site, retained so a future caller outside the `elif` cannot regress.

- **File:** `scripts/nightly_regression_tests.py:679-686` (verified: read this file)
  **Code:** `result = subprocess.run(  # timeout-guard: allow` / `argv, cwd=str(cwd), capture_output=True, text=True, timeout=timeout,`
  **Issue:** `subprocess.run`'s timeout kills only the direct child, then blocks in `communicate()` until EOF on the captured pipes. A `uv sync` build subprocess that survives the kill and still holds the pipe hangs the nightly indefinitely — before the page. This is the exact mechanism `_spawn_pytest`'s own docstring exists to defeat with `start_new_session=True` plus `killpg`. The plan's Risk 2b mitigation "it carries its own timeout" does not fully hold for the provisioning steps.
  **Severity:** tech_debt
  **Fix:** Route provisioning through the same process-group-owning shape as `_spawn_pytest` (`Popen` + `start_new_session=True` + `os.killpg` on `TimeoutExpired`), or at minimum drop `capture_output` for the `uv sync` step so a surviving grandchild cannot block `communicate()`.

- **File:** `scripts/nightly_regression_tests.py:719-733` (verified: read this file)
  **Code:** `["git", "-C", str(worktree_path), "checkout", "--detach", baseline_sha]` / `["git", "worktree", "add", "--detach", str(worktree_path), baseline_sha]`
  **Issue:** No self-heal from a desynced worktree admin entry, and no `git worktree prune` anywhere in the new code. If the directory is removed without `git worktree remove` (an operator reclaiming the ~868 MB the new docs describe as protected, a partial reclaim, an interrupted `add`), `worktree_path.exists()` is `False` and `git worktree add` fails permanently with "missing but already registered". Symmetrically, a directory whose admin entry is gone makes the `checkout` fail forever. Either way every subsequent night buckets 100% `inconclusive`, visible only as a WARNING line — the classifier ships permanently inert while looking safe, which is Risk 1 restated.
  **Severity:** tech_debt
  **Fix:** On a failed re-point or add, run a bounded `git worktree prune` (and/or `git worktree remove --force`) in `repo_root` and retry once, logging the recovery so a persistent failure stays visible.

- **File:** `scripts/nightly_regression_tests.py:797-806` (verified: read this file)
  **Code:** `argv = [str(wrapper), *ordered, "-n0", "--tb=no", "-q", "--json-report", f"--json-report-file={report_path}",]`
  **Issue:** All node IDs go into one pytest invocation. A node ID that does not exist at the baseline SHA — a newly **added** failing test, which is the single most common shape of a newly-confirmed failure — makes pytest exit as a usage error, so the *whole batch* buckets `inconclusive` rather than just that node. Over time this systematically suppresses the `newly_broken` / `pre_existing` evidence the shadow tier exists to collect, and it does so silently.
  **Severity:** tech_debt
  **Fix:** Pre-filter node IDs whose file does not exist at the baseline checkout (bucketing only those `inconclusive`), or fall back to a per-node re-run when the batch exits with pytest's usage-error code.

- **File:** `scripts/nightly_regression_tests.py:1640-1644` (verified: read this file)
  **Code:** `# It runs BEFORE the alert is built because the alert re-reads PYTEST_SERIAL_JSON_TMP below and the classifier owns a separate report path.`
  **Issue:** The ordering itself is plan-mandated and its critical-path cost is accepted under the plan's Risk 2b, so this is not raised as a scope objection. What is inaccurate is the **claim** attached to it: the PR body's "Behavior guarantee: Alerting is **unchanged**", `docs/features/nightly-alert-triage.md`'s "paging exactly as today", and `docs/features/nightly-regression-tests.md`'s "while paging a human exactly as today" all assert an unchanged page when only the message *text* is unchanged. On a failing night the page is now delayed by up to `BASELINE_GIT_TIMEOUT_SECONDS` + `BASELINE_UV_SYNC_TIMEOUT_SECONDS` + `PYTEST_BASELINE_TIMEOUT_SECONDS` (300 + 900 + 1800 ≈ 50 minutes on a cold or hung provision). Separately, the comment's stated *reason* is self-refuting: because the classifier owns a separate report path, nothing forces it to run before the alert.
  **Severity:** tech_debt
  **Fix:** Say "the alert *text* is unchanged; the page is delayed by up to the classification bound" in the PR body and both feature docs, and replace the inline comment's rationale with the real one.

### Nits

- **File:** `scripts/nightly_regression_tests.py:1016-1017` (verified: read this file)
  **Code:** `verdict = decide_fix_or_escalate(classification, new_failures, caps, run_flags)` / `reason = gate_reason(classification, new_failures, caps, run_flags)`
  **Issue:** The entire gate clause chain is evaluated twice, since `decide_fix_or_escalate` is defined as `gate_reason(...) == "none"`. Harmless (both are pure) but redundant.
  **Severity:** nit
  **Fix:** Compute `reason = gate_reason(...)` once, then `verdict = "autonomous-fix" if reason == "none" else "escalate"`.

- **File:** `scripts/nightly_regression_tests.py:722,729` (verified: read this file)
  **Code:** `["git", "-C", str(worktree_path), "checkout", "--detach", baseline_sha]`
  **Issue:** No `--` separator before `baseline_sha`. The value comes from `last_run.json`, local state the script writes itself, so this is hardening rather than a live exploit — but a value beginning with `-` would be parsed as an option.
  **Severity:** nit
  **Fix:** Insert `"--"` before `baseline_sha` in both git invocations.

- **File:** `tests/unit/test_nightly_classifier.py:693` (verified: read this file)
  **Code:** `def _flags(**kwargs) -> nrt.RunFlags:`
  **Issue:** Byte-identical `_flags()` helper is duplicated in `tests/unit/test_nightly_decision_gate.py:28`.
  **Severity:** nit
  **Fix:** Hoist to a shared helper or conftest so the two files cannot drift on the default run shape.

- **File:** `scripts/nightly_regression_tests.py:1013` (verified: read this file)
  **Code:** `classification = classify_against_baseline(new_failures, run_flags.baseline_sha)`
  **Issue:** `log_shadow_verdict` does not thread `classify_against_baseline`'s keyword-only injection seam, so any test of the verdict path must patch the module attribute — the seam stops one level short of the entry point.
  **Severity:** nit
  **Fix:** Optional — give `log_shadow_verdict` a `classify=classify_against_baseline` keyword-only parameter.

### Miscellaneous
- None

- **The entire `active`-mode / autonomous-fixer stack (dispatch, hand-back, watchdog, PR guards, enforcement preflight, `--silent` notify transport)** — tracked by #3076 (OPEN, verified) — matched findings: none. The plan's `[SEPARATE-SLUG #3076]` No-Go is respected; the leak-check verification rows (`nightly_fix_handback`, `maybe_dispatch_fix_session`, `branches/main/protection`, `SDLC_AGENT_GH_TOKEN`, `"--silent"`, `models/` + `migrations.py` diff) all come back clean.
- **Six `## Verification` rows are not machine-runnable as written** (trailing em-dash prose in the command cell; `≥` / `== N` / `empty output` expectation forms the evaluator does not parse) — explicit out-of-scope disclosure with sound rationale: editing the table mid-build would have disarmed the plan-hash guard. Independently confirmed: every one of those rows was re-run by hand here and passes. The residual failures reported by the table runner are all evaluator/quoting artifacts, not code defects. One row (`skip_still_logs`) failed the runner with a pytest collection error but passes cleanly on two direct re-runs (`2 passed`), so it is recorded as environmental, not a finding.
- **Cross-run learning / auto-tuning of the guardrail constants** — explicit out-of-scope exclusion with sound rationale (they ship provisional and env-overridable). Note that Blocker 1 is about the env-override being inert, not about this exclusion.

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — PASS — *every Success Criteria bullet maps to shipped code and a test; the reversibility bullet is undermined by Blocker 1 but the criterion's code is present.*
- **2. No-Gos from plan — none violated** — PASS — *all six #3076 leak-check rows are clean; `.claude/agents/` and `models/` diffs are empty.*
- **3. New `except Exception` blocks — each has logger/raise/swallow-ok** — PASS — *`classify_against_baseline`'s broad catch logs and buckets `inconclusive`, which is the documented fail-toward-escalate rule.*
- **4. New integration tests — exercise serialization boundary (not in-memory only)** — PASS — *the fixture-repo suite runs a real subprocess through the real wrapper and parses a real JSON report.*
- **5. Plan internal consistency — spike findings match task steps** — PASS — *round-11 spike claims re-verified against the shipped code.*
- **6. No hardcoded secrets or debug artifacts** — PASS — *no secrets; `baseline_sha` and node IDs are literal argv with no shell.*
- **7. New public APIs — docstrings present** — PASS — *every new function carries a substantive docstring.*
- **8. Breaking changes — migration path documented** — N/A — *no schema change, no persisted-state key, no CLI surface change.*
- **9. Tests added for new behavior** — PASS — *258 tests across four files, reproduced.*
- **10. Tests cover the failure path (not just happy path)** — FAIL — *no test asserts the page still fires when the classifier raises; see Blocker 2.*
- **11. UI changes (if any) — screenshot captured** — N/A — *no HTML/CSS/JS/JSX/Vue/template files in the diff; the visual-proof gate is a no-op.*
- **12. Docs updated for user-facing changes** — FAIL — *docs ship, but the "(gitignored)" marker row and the "paging exactly as today" claim are inaccurate; the `off` knob is documented as working and is not.*

## Rubric

- [x] **1. Plan vs. implementation match** — pass — *every criterion delivered or covered by a verified #3076 deferral.*
- [ ] **2. New code quality** — fail — *unguarded pre-page hook + narrow `OSError`-only handler at a reachable path (Blocker 2).*
- [ ] **3. Test coverage** — fail — *the failure path that matters most (classifier raises → page must still fire) is untested; every other path is well covered.*
- [x] **4. Regression risk to existing callers** — pass — *both `_spawn_pytest` callers verified byte-identical and asserted; `send_telegram`'s "Telegram sent" line has no non-test consumer.*
- [x] **5. Data integrity** — pass — *no schema change; state-file key set asserted identical across modes.*
- [x] **6. Security** — pass — *no shell, no request-derived input, no secrets.*
- [ ] **7. Documentation accuracy** — fail — *"(gitignored)" is false; "paging exactly as today" overstates; `seed_run` documented as a live precondition it cannot be.*
- [ ] **8. PR body accuracy** — fail — *"Behavior guarantee: Alerting is **unchanged**" is true of the text, not of the page.*
- [x] **9. Disclosed deferrals** — pass — *#3076 is OPEN; the verification-table disclosure is sound and was independently re-checked.*
- [x] **10. Follow-up claims verified** — pass — *#3076 resolves to an OPEN issue.*

**Verdict derivation:** Rule 1 — critical items 2 and 3 are `fail` with no matching acknowledged disclosure → **CHANGES REQUESTED — Blocker**.

### Verification Results

- `python -m ruff check .` → `All checks passed!` (exit 0)
- `python -m ruff format --check .` → `1421 files already formatted` (exit 0)
- `scripts/pytest-clean.sh` over the four affected test files → `258 passed in 74.61s` (exit 0)
- Plan `## Verification` table via `agent.verification_parser` → all machine-runnable rows pass; the six rows the PR body discloses as non-runnable fail on quoting/expectation-form artifacts only and were re-run by hand.

### Review Delta

No prior review on this PR — this is the first pass.

### Consensus

`rule=any-blocker-wins` · `k=2` · `n=2` · `judges_run=2` · `consensus_disagreement=false` · `mean_confidence=0.875` · `blockers=2` (max) · `tech_debt=3`

valorengels:
## Review: PR #3082 — nightly autonomous fixer, fix before alert (#2334)

Re-review at head `6ae86540150db0c3e97504f673e8288902d2ea32`, resolved via `tools/pr_head_resolver.py::resolve_pr_head_sha` (git-first, `refs/pull/3082/head`), not a bare `gh` read. This supersedes the multi-judge `CHANGES REQUESTED` at head `224ce2be9`. Every claim below was verified against code read at this head, not against commit messages.

**Verdict: APPROVED** — 0 blockers, 0 carried tech-debt. Changes not required before merge.

### Prior blockers — both RESOLVED

**1. Kill switch inert on the only surface that matters.** Fixed in `daa1dc737`. `resolve_fix_mode()` / `resolve_fix_max_failures()` (`scripts/nightly_regression_tests.py:1070-1108`) now read `os.environ` at call time, replacing the import-time module constants. Ordering traced in `main()` rather than assumed: `load_env_or_die()` runs at line 1553; the knobs are not resolved until lines 1766/1770, well after `.env` has populated `os.environ`. `GateCaps(max_failures=resolve_fix_max_failures())` at line 1770 confirms the max-failures knob moved too. So `NIGHTLY_FIX_MODE=off` in the vault `.env` now genuinely disables the feature, and the plan's "Reversibility: total" claim is real rather than aspirational.

**2. Shadow path was the only unguarded pre-page hook.** Fixed in `daa1dc737`. The `log_shadow_verdict(...)` call at `scripts/nightly_regression_tests.py:1766-1779` is wrapped `try/except Exception`, mirroring the TTFT gate's established non-fatal pattern. `provision_baseline_worktree` sits inside that guard via the `log_shadow_verdict → classify_against_baseline → provision_baseline_worktree` chain (line 885), so the prior structural complaint about it being outside the guarded `try` is moot. `send_telegram(...)` at line 1793 fires unconditionally after the guarded block. The specific corrupt-marker vector is additionally fixed at source: the marker read (lines 797-803) now catches `(OSError, ValueError)`, and `UnicodeDecodeError` is a `ValueError` subclass.

### The three carried tech-debt items — disposition

Asked to rule each MERGE-GATE or DEFER-WITH-TRACKING. The honest answer is **neither: all three were fixed outright in `daa1dc737`** and are not live at this head. Recording that explicitly so they do not ride silently.

1. **`subprocess.run(timeout=)` grandchild hang before the page — RESOLVED.** This was the one that genuinely warranted merge-gate treatment, because it hangs *before* the human page rather than merely degrading classification. `_run_provision_step` (lines 675-721) no longer uses `subprocess.run(timeout=)`. It uses `subprocess.Popen(..., start_new_session=True)` with `communicate(timeout=...)` and a `killpg(SIGTERM)` → sleep → `killpg(SIGKILL)` escalation — the same process-group shape `_spawn_pytest` uses, with a docstring naming the defeated failure mode. `test_provision_steps_own_a_process_group` asserts `start_new_session=True` on every provisioning `Popen`.
2. **No self-heal from a desynced worktree admin entry — RESOLVED.** `provision_baseline_worktree` (lines 771-787) retries once via `git worktree prune` when `_point_at_baseline()` fails, logging both attempt and recovery. `test_provision_prune_and_retry_recovers_a_desynced_admin_entry` asserts the token order `["worktree_add", "worktree_prune", "worktree_add"]`.
3. **Newly-added failing node poisoning the batch to `inconclusive` — RESOLVED.** `classify_against_baseline` (lines 895-916) pre-filters node IDs whose test file is absent at the baseline checkout, bucketing only those `inconclusive` before any pytest invocation. Covered by `test_prefilter_absent_node_is_inconclusive_alone`, `test_prefilter_all_nodes_absent_skips_the_baseline_run`, and `test_prefilter_logs_the_absent_nodes`.

### Tests and lint

```
tests/unit/test_nightly_regression_tests.py                            → 126 passed
tests/unit/test_nightly_classifier.py + test_nightly_decision_gate.py  →  94 passed
ruff check (5 files)                                                   → All checks passed
ruff format --check (5 files)                                          → 5 files already formatted
```

220 passed, 0 failed, via `scripts/pytest-clean.sh`. Worth noting for the record: the new blocker and tech-debt tests live in `test_nightly_classifier.py` and `test_nightly_decision_gate.py`, not in `test_nightly_regression_tests.py`, so a review scoped only to the latter would have missed the coverage it was meant to judge.

### Are the new tests armed?

Checked specifically, because `6ae865401` in this same range fixed a vacuous test of exactly this shape (patching `nrt.NIGHTLY_FIX_MODE`, a name production stopped reading after `daa1dc737`, so the patch applied cleanly and asserted nothing).

- `test_mode_gating_shadow_tier_exception_never_suppresses_the_page` patches `nrt.classify_against_baseline` with a `RuntimeError` side effect and drives `main()` end to end, asserting `rc == 0`, one `send_telegram` call with alert text byte-identical to `off`-mode, and the non-fatal log line. `main()` calls `log_shadow_verdict(...)` with no explicit `classify=` kwarg, so it late-binds the module global the test patches. Genuinely armed — it fails if the guard is removed.
- The `TestBaselineNodePrefilter` and `TestProvisionHardening` tests patch `nrt.subprocess.Popen` and `nrt._spawn_pytest`, which are the names production actually calls. Armed.
- Swept the test tree for any other stale `NIGHTLY_FIX_MODE` / `NIGHTLY_FIX_MAX_FAILURES` module-attribute patches: none remain. All other references are `os.environ` / `monkeypatch.setenv` manipulation, which is the correct target now.

### Merge gates

No blockers and no carried debt. Docs claims in `42512c9ff` (page-delay bound, `seed_run` unreachability, `.nightly-baseline-provisioned` ignored at `.gitignore:477`) were spot-checked against code and are accurate. Not merging from this review.

Issue #2799, opened 2026-08-13
tests/conftest.py Redis client ignores REDIS_PORT: 'private redis' isolation connects to production 6379 while opting out of the claim pool


tomcounsell:
Found by PR #2695's independent verification agent (2026-08-13), confirmed by hitting it live (popoto plugin reported 15036 keys in db0 from a 'private' run).

**Mechanism:** tests/conftest.py builds its client as `redis.Redis(db=test_db)` with no host/port, so every test run connects to localhost:6379 regardless of environment. `REDIS_PORT` only re-keys the db-claim registry directory. An agent who starts a private `redis-server --port 641x` for isolation therefore gets the OPPOSITE: tests still hit production 6379, but the claim registry is now keyed to the private port, opting the run OUT of the machine-global pool that prevents two runs from claiming the same db. Collisions become more likely, not less.

**Evidence it's widespread:** stray claim-registry directories currently exist for ports 6419, 6420, 6422, 6423, 6424 — several agents have used this pattern. It is also a plausible explanation for tonight's unreconciled 6-of-12 flake in one #2695 round-3 review that the other reviewer could not reproduce in 26 runs.

**Fix shape:** conftest's client construction should honor REDIS_HOST/REDIS_PORT (or a single TEST_REDIS_URL), so a private server is actually used when configured; plus a startup assertion that the connected server is the one the claim registry is keyed to. Sweep the stray registry dirs as part of the fix.

**Interim guidance for agents:** do NOT spin a private-port redis for the unit suite — stay on the shared pool via the claim registry; a private port silently degrades isolation.

Issue #2770, opened 2026-08-13
Upstream popoto: make the bundled pytest11 plugin's flush honor an env-pinned db by default


tomcounsell:
## Background

`popoto` ships a `pytest11` entry point (`popoto.pytest_plugin`) that this repo loads on every test run. Its autouse `_popoto_flush_db` fixture flushes Redis before every test, defaulting to db 15 unless `POPOTO_TEST_DB` (or the `popoto_test_db` ini option) is set. Because db 15 sits inside this repo's test-DB claim pool (`tests/db_claim.py`, `_TEST_DB_POOL_MAX = 15`), any process that has not set `POPOTO_TEST_DB` collides with a claimed slot and silently flushes another live pytest process's data.

This is the mechanism behind #2628 ("main's full-suite failure set rotates run to run"). PR #2683 fixes it downstream by exporting `POPOTO_TEST_DB` from a `pytest_configure` hook before any fixture runs, driving the plugin through its own documented resolution input. That fix is durable against future popoto versions (it carries a drift-detecting assertion), but it is still a per-consumer workaround.

## The upstream ask

Make `popoto.pytest_plugin`'s flush honor an env-pinned db **by default**, so no downstream consumer can collide with a shared pool unless it explicitly opts out. Concretely: the plugin already resolves `POPOTO_TEST_DB` / `popoto_test_db`; the ask is for popoto's own default (in the absence of any consumer configuration) to avoid a single hardcoded db number (currently 15) that every consumer's test suite implicitly shares.

There is precedent for landing changes upstream in this project — popoto PR #518 is already in flight (per `/Users/valorengels/src/popoto/tests/test_pytest_plugin.py`, which already asserts `db == 15` in four places; those assertions belong to this upstream change).

## Why this repo doesn't build it

Not built directly in PR #2683: it needs a popoto release plus a floor version bump in this repo, which is a different blast radius and merge cadence than a pure-`tests/` PR. See `docs/plans/suite-failure-rotation-db-ownership.md`, No-Gos section, "Upstream popoto PR" entry.

## Scope

- Popoto repo: give `popoto.pytest_plugin`'s flush fixture a safer default resolution (env-pinned, not a single shared literal).
- This repo: bump the popoto floor version once the fix ships, and evaluate whether the downstream `POPOTO_TEST_DB` export in `tests/conftest.py::pytest_configure` can be simplified or removed.

Refs #2628, PR #2683.

tomcounsell:
[...]

Issue #2655, opened 2026-08-07
Recurrence guard: detect tests that derive their own Redis db= instead of using the claim API


valorengels:
> **Context for the stranger:** this repo's test suite runs many pytest processes at once, and each one must own a private Redis logical database or they flush each other's data mid-test. Ownership is assigned by an `flock` claim over the pool `[1..15]` (`tests/db_claim.py`). A test that computes its own database number — rather than asking the claim API — can therefore issue a destructive command against a database a *different live process* owns. That defect class has now been fixed three separate times (#2117, #2606, #2624) and re-emerged each time in a new call site, because nothing makes a wrong derivation *detectable*. This issue is the detector. It was split out of #2628, which fixes the underlying rotation; this is the guard that stops it recurring.

## Problem

Every fix in the #2117 → #2606 → #2624 series corrected one call site's derivation without making the next wrong one visible. #2628's investigation found three writers still flushing databases they did not own, one of which lived in installed library code and was invisible to any review of this repo's own sources.

**Current behavior:** nothing detects a test that derives its own Redis `db=` value. Review is the only control, and review has missed it three times.

**Desired outcome:** a test that constructs a Redis client with a `db=` value not obtained from the claim API fails a check, by name, at the offending line — every time, without a human noticing.

This was originally Task 6 of #2628's plan. It is **descoped from that fix deliberately**: it is hardening, not the fix. It prevents future regressions; it does not stop the rotation. Across critique rounds 4, 5 and 6 it produced the *sole* remaining blocker each time while the actual rotation fix waited behind it. Landing it separately unblocks #2628 and gives this problem the attention it turns out to need.

## Definitions

| Term | Definition |
|------|------------|
| Claim API | `tests/db_claim.py` — `claim_test_db()` and `claim_scratch_test_db()` return an `flock`-owned database number for this process. The only sanctioned source of a `db=` value. |
| Recurrence guard | The AST-walking check this issue asks for: it inspects test sources for Redis client constructions and rejects `db=` values not sourced from the claim API. |
| Permit list | The set of expressions the guard accepts as a legitimate `db=` value (e.g. a call to `claim_test_db`). Distinct from the allowlist. |
| Allowlist | Named files exempted from the guard entirely — e.g. the db-0 guard's own tests, which must construct `db=0` literally to test that it is rejected. |
[...]

Issue #2037, opened 2026-07-11
Systemic: real-record Popoto tests flake under parallel xdist (create-then-query visibility race)


tomcounsell:
## Observation

Unit tests that create a real Popoto record and then re-query it via `Model.query.filter(...)` intermittently fail under the default parallel test config (`pytest -n auto --dist=loadfile`), when co-scheduled with another Redis-touching test file. The just-created record is not visible to the fresh query (`filter` returns empty), so assertions that expect to find it fail.

## Evidence (not specific to any one PR)

Both of these flake identically under an 8x co-scheduled loop, ~15-25% of runs:

```
# already on main — the established real-record PipelineLedger test:
for i in $(seq 1 8); do pytest tests/unit/test_pipeline_ledger.py tests/unit/test_validate_merge_guard.py -q; done
# observed: run 1 -> 2 failed; run 5 -> 1 failed; others green

# the #2034 merge-predicate resolver test (same pattern):
for i in $(seq 1 8); do pytest tests/unit/test_merge_predicate.py tests/unit/test_validate_merge_guard.py -q; done
# observed: ~1 in 6 runs fails with resolver NO_SIGNAL (created ledger not found)
```

Serial (`-n0`) and file-alone-under-xdist runs are 100% green. The flake appears only under parallel co-scheduling.

## Suspected cause

The `redis_test_db` autouse fixture (`tests/conftest.py:189`) repoints Popoto's `POPOTO_REDIS_DB` to a per-worker db, but there appears to be a race between record creation (`get_or_create`/`save`) landing in one connection/db and the subsequent `query.filter` reading via another — or a cross-file `flushdb` on a shared worker db. This is the documented "Redis test-DBs collide across concurrent runs" hazard.

## Impact

Any test that round-trips a real Popoto record through the secondary index under parallel xdist is intermittently red. It is a suite-wide infrastructure issue, not a product bug.

## Scope note

Found while shipping the #2034 merge-predicate fix (PR #2036), whose production-shaped regression test exhibits the flake. That PR fences to `tools/merge_predicate.py` + its test, so the systemic conftest/xdist fix belongs here as a separate investigation.

## Suggested direction

Audit `redis_test_db` for connection-pool/db-pointer consistency between write and read paths under xdist; consider forcing a fresh Popoto connection per test or an `xdist_group` that serializes real-record Redis tests; verify `flushdb` cannot run against a db another worker is mid-test on.

---

## Recon Summary

[...]

PR #3083, opened 2026-09-02
Consolidate the two owners of popoto's test db (#2771)

### LINKED ISSUE #2771: Consolidate the two owners of "which db popoto is on" in tests/conftest.py
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/consolidate-two-owners-of-popoto-test-db.md

## Background

After PR #2683 (#2628), there are two independent mechanisms in the test suite that each believe they own repointing popoto's Redis client at the claimed test db:

1. **`popoto.pytest_plugin`'s own `_swap_db`** — now correctly driven by `POPOTO_TEST_DB`, which `tests/conftest.py::pytest_configure` exports before collection. This covers submodule bindings in place.
2. **`tests/conftest.py`'s `redis_test_db` fixture** — still runs a replace-and-repatch loop (`tests/conftest.py:616,626-628,640-641`) that separately points `_POPOTO_ASYNC_REDIS_DB` (and other bindings) at the test db.

Two independently-maintained authorities over the same fact ("which db is popoto on right now") is exactly the failure shape #2628 exists to remove — a call site that should ask a single source of truth instead re-derives its own answer, and the two can drift.

## Why not built in PR #2683

`tests/conftest.py:631` is what points `_POPOTO_ASYNC_REDIS_DB` at the test db during the sync setup phase. Removing the `redis_test_db` fixture's repatch loop is a separate blast radius from the ownership-enforcement fix (round-4 review concern 2 on the plan) — it risks breaking async client repointing if done in the same change as the flush guard. See `docs/plans/suite-failure-rotation-db-ownership.md`, No-Gos section, "Consolidating the two owners" entry.

## Scope

- Audit whether `popoto.pytest_plugin`'s in-place `_swap_db` (now correctly driven by `POPOTO_TEST_DB`) already covers everything `tests/conftest.py`'s repatch loop does, including `_POPOTO_ASYNC_REDIS_DB`.
- If so, delete the redundant repatch loop from `redis_test_db` and keep the plugin as the single authority.
- If not, document why the split is required and consider whether the async binding should move into the plugin's own resolution instead.

Refs #2628, PR #2683.

## Recon Summary

Audited at `main` = `3b6eb651b` (2026-09-02), against installed `popoto` at
`.venv/lib/python3.14/site-packages/popoto/`.

**Line drift since filing (2026-08-13)**

Two commits landed on `tests/conftest.py` since this issue was written, so the
cited line numbers have moved and one of them changed the *answer*:

| Commit | Effect on this issue |
|---|---|
| `d59f6509c` (#2805) | `pytest_configure` now also exports a process-wide `REDIS_URL` (`_export_claimed_redis_url`, conftest:277-305). Its docstring already concedes the "honest limitation": popoto's `pytest11` plugin builds `POPOTO_REDIS_DB` *before* `tests/conftest.py` is imported, so this export does not reach this process's own popoto client. |
| `ff20e0311` (#2799) | `redis_test_db` now builds its client from `db_claim.redis_test_host()/redis_test_port()` and asserts it with `_assert_client_matches_claim_registry` (conftest:887-911, 960-963). **This is a new, load-bearing reason conftest constructs its own client** that did not exist when this issue was filed. |

Current locations (were 616, 626-628, 640-641):

- `tests/conftest.py:964` — `rdb.POPOTO_REDIS_DB = test_client` (object **replacement**)
- `tests/conftest.py:973-976` — the submodule repatch loop
- `tests/conftest.py:980` — eager `rdb._POPOTO_ASYNC_REDIS_DB = aioredis.Redis(...)`
- `tests/conftest.py:659-710` — `_popoto_modules_with_redis_db()` + its two-signal cache, which exists **only** to feed the repatch loop
- `tests/conftest.py:987-990` — teardown restore of all three

**Confirmed:**

1. **The repatch loop is redundant *in principle*, and is a direct consequence of
   conftest replacing the object rather than mutating it.** The plugin's
   `_swap_db` (`pytest_plugin.py:100-116`) rebuilds `POPOTO_REDIS_DB`'s
   `connection_pool` **in place** on the existing object, so every
   `from ..redis_db import POPOTO_REDIS_DB` submodule binding follows
   automatically — no repatch needed. conftest instead assigns a **new**
   `redis.Redis` to `rdb.POPOTO_REDIS_DB`, which strands every submodule binding
   on the old object; lines 973-976 exist purely to clean up after that choice.
   Change conftest to an in-place pool swap and the loop (and the whole
   `_popoto_modules_with_redis_db` cache above it) becomes dead by construction.

2. **The eager async assignment is redundant *and* actively harmful.**
   `get_async_redis_db()` (`redis_db.py:216-270`) explicitly mirrors the *current*
   sync client's `connection_pool.connection_kwargs` (`host`, `port`, `db`,
   `password`, `username`) rather than re-reading `REDIS_URL` — its docstring says
   this exists precisely so the async path follows the plugin's test-DB swap. So a
   correct sync client is sufficient; the async client needs no separate pointing.
   Meanwhile the plugin's `_popoto_reset_async` nulls `_POPOTO_ASYNC_REDIS_DB` at
   setup and teardown of every test **specifically so the client is built lazily
   inside the test's own event loop** ("a client built now would be bound to the
   wrong loop"). conftest:980 builds one in a synchronous setup context, which is
   the exact thing that fixture is there to prevent.

3. **Ordering spike (run live, `-p no:randomly -n0`): conftest wins.** A probe test
   printed `ASYNC_IS_NONE: False` in the test body, proving `redis_test_db` sets up
   *after* the plugin's `_popoto_reset_async`. So today the eager, wrong-loop client
   is the one tests actually see. Deleting line 980 hands the binding back to the
   plugin's lazy in-loop path — a behavioural improvement, not just a deletion.

**Revised:** (the scope's "if not, document why" branch applies to exactly one thing)

4. **Host/port cannot move to the plugin as-is — full deletion of conftest's swap
   would regress #2799.** `_swap_db` *preserves* whatever host/port the existing
   pool already has, and that pool was built at popoto import time from the ambient
   `REDIS_URL` (`redis_db.py:138-160`). `db_claim` resolves independently from
   `REDIS_HOST`/`REDIS_PORT` (`db_claim.py:106-125`). Under `REDIS_PORT=641x` those
   two disagree, which is #2799 verbatim: the claim registry keyed to 641x while the
   client sits on 6379, flushing production db N. `_popoto_test_db` calls
   `_swap_db(test_db)` with no host/port extras, so the plugin has no way to learn
   the registry's server. `_swap_db` *does* accept `**extra_kwargs` — but supplying
   them requires an upstream popoto change, which is out of this repo.

   **Therefore the consolidation direction is the inverse of the issue's first
   guess:** keep `tests/conftest.py` as the single owner of *server resolution*,
   keep the plugin as the single owner of the *db number* (via the `POPOTO_TEST_DB`
   export it already honours), and delete conftest's two redundant mechanisms
   (repatch loop, eager async bind) by switching it from object-replacement to
   in-place pool mutation. The two then compose instead of competing:
   `_popoto_flush_db` re-swaps only on db drift and copies current kwargs forward,
   so it preserves conftest's host/port.

**Incidental finding**

5. **Object replacement silently drops popoto's connection cap.** The same probe
   printed `POOL_CLASS: ConnectionPool`. popoto builds `POPOTO_REDIS_DB` on a
   `redis.BlockingConnectionPool` with `max_connections=128`
   (`redis_db.py:99-128`) specifically to stop `MaxConnectionsError` under
   `asyncio.gather` bursts. conftest's bare `redis.Redis(host=..., port=..., db=...)`
   replaces it with an unbounded plain pool, so the whole suite runs without that
   protection. An in-place swap that preserves the pool class fixes this for free.

**Pre-requisites:** / blast radius

- `redis_test_db` is requested **by name** by ~40 test functions across
  `tests/unit/` as their "I need Redis" marker. The fixture must keep its name and
  autouse status; only its body changes.
- `tests/unit/test_conftest_isolation_guards.py:264,339,375` test
  `_popoto_modules_with_redis_db`'s two-signal cache and the create-then-filter
  split-brain it guards. Deleting the cache makes these obsolete.
- `tests/unit/test_test_redis_server_resolution.py:44-54`
  (`test_async_client_matches_the_sync_client`) reads `rdb._POPOTO_ASYNC_REDIS_DB`
  and `pytest.skip`s when it is `None`. Dropping the eager bind turns this into a
  permanent silent skip — it must be rewritten against `get_async_redis_db()`.
- `tests/db_claim.py:4,20,111` documents the current conftest behaviour in prose
  and needs updating with it.
- `tests/conftest.py:738-771` (`_popoto_client_ownership_check`) already asserts the
  invariant this work must preserve, and already documents that
  `_POPOTO_ASYNC_REDIS_DB` is `None` at session scope and that this is *correct*.

**Dropped:**

- Changing `popoto` itself (moving host/port resolution into `_swap_db`) — different
  repo (`~/src/popoto`), and unnecessary once conftest owns server resolution.

valorengels:
Closes #2771

Two mechanisms each believed they owned "which Redis db is popoto pointed at": popoto's bundled `pytest11` plugin (in-place pool swap, driven by `POPOTO_TEST_DB`) and `tests/conftest.py`'s `redis_test_db` autouse fixture (which threw a **new** `redis.Redis` object at `rdb.POPOTO_REDIS_DB` and then spent 20 lines cleaning up after that choice).

This consolidates them into a clean split by fact.

## What changed

**`tests/conftest.py`**
- New session-scoped autouse fixture `_popoto_pool_install`, ordered after the plugin's `_popoto_test_db`. It mutates the canonical client's `connection_pool` **in place**, exactly once per session, installing a `redis.BlockingConnectionPool` on the db-claim registry's host/port carrying popoto's own `_SYNC_MAX_CONNECTIONS` cap (read defensively; #2770 is the upstream coordination point). Object identity is preserved, so every popoto submodule's import-time `POPOTO_REDIS_DB` binding follows for free.
- `redis_test_db` keeps its exact name, `autouse=True`, and function scope — the contract ~197 call sites depend on — but its body is now per-test hygiene only: `_assert_client_matches_claim_registry` against the canonical client, flush, yield, flush. No pool work, no `disconnect()`, no object rebinding, no async global write.
- **Deleted:** `_popoto_modules_with_redis_db`, `_POPOTO_MODULE_CACHE`, `_POPOTO_MODULE_CACHE_LEN` and their explanatory comment block; the submodule repatch loop; the eager `rdb._POPOTO_ASYNC_REDIS_DB = aioredis.Redis(...)` bind and its save/restore; the `import redis.asyncio as aioredis` line.
- The `redis_test_db` docstring is rewritten — its "CRITICAL: We replace the POPOTO_REDIS_DB object" paragraph was the exact opposite of the new code.

**Ownership split (the point of the change)**
| Fact | Owner |
|---|---|
| db number + server host/port | `tests/db_claim.py` |
| applying the db number | popoto's bundled plugin, via the `POPOTO_TEST_DB` export |
| applying the server | `tests/conftest.py::_popoto_pool_install` |
| the async client | `popoto.redis_db.get_async_redis_db()`, built lazily in-loop |

Full delegation to the plugin is not possible: `_swap_db` preserves whatever host/port the pool already carries and has no channel to learn the claim registry's server, so delegating would regress #2799. That is documented in the docstring rather than left to be re-derived.

**Two live defects fixed, not just code deleted**
- The eager async bind was constructing `_POPOTO_ASYNC_REDIS_DB` in a synchronous setup context — the one thing the plugin's `_popoto_reset_async` exists to prevent ("a client built now would be bound to the wrong loop"). A live probe confirmed conftest set up *after* the plugin, so the wrong-loop client was the one tests actually got. It also leaked an un-disconnected async pool per test.
- Object replacement silently dropped popoto's connection cap: a bare `redis.Redis(host=..., port=..., db=...)` gets an unbounded plain `ConnectionPool`, so the whole suite ran without the `MaxConnectionsError` protection popoto builds a `BlockingConnectionPool` for.

**Tests**
- New `tests/unit/test_popoto_client_identity.py` — four assertions: (a) the client is the same object across tests in a process, (b) every `popoto.*` module in `sys.modules` holding a `POPOTO_REDIS_DB` symbol holds *that* object, (c) the live pool is a `BlockingConnectionPool` with popoto's cap, (d) `id(connection_pool)` is stable across two tests in a session (the fence against reintroducing a per-test restore).
- `TestPopotoModuleCacheInvalidation` deleted (it tested a cache that no longer exists).
- `TestPopotoSplitBrainRoundTrip` keeps its #2037 reproduction verbatim; the fix step becomes a restore to the canonical object, and its two bare `redis.Redis(db=...)` clients now resolve through `db_claim`.
- `test_async_client_matches_the_sync_client` is now an async test that awaits `get_async_redis_db()` inside the test's own loop and **fails** on divergence — it can no longer `pytest.skip` its way to green.
- Added a regression fence asserting `_POPOTO_ASYNC_REDIS_DB` is still `None` immediately after `redis_test_db` setup.

## RED state on `main` (paper trail, required by the plan's Success Criteria)

`./scripts/pytest-clean.sh tests/unit/test_popoto_client_identity.py -q -n0 -p no:randomly`, run with only the new test file added and `tests/conftest.py` untouched:

```
.FF.F                                                                    [100%]
=================================== FAILURES ===================================
__________________ test_pool_identity_is_stable_across_tests ___________________
tests/unit/test_popoto_client_identity.py:49: in test_pool_identity_is_stable_across_tests
    assert id(rdb.POPOTO_REDIS_DB.connection_pool) == _OBSERVED["pool_id"], (
E   AssertionError: popoto's connection pool was replaced between two tests in the same session. The pool is installed once per session by conftest's _popoto_pool_install; a per-test restore or rebuild reintroduces the hazard that fixture exists to remove (#2771).
E   assert 4653099184 == 4653240784
______________ test_client_object_identity_is_stable_across_tests ______________
tests/unit/test_popoto_client_identity.py:61: in test_client_object_identity_is_stable_across_tests
    assert id(rdb.POPOTO_REDIS_DB) == _OBSERVED["client_id"], (
E   AssertionError: rdb.POPOTO_REDIS_DB was rebound to a different object between two tests. conftest must mutate the canonical client's pool in place, never replace the object — replacement strands every popoto submodule's import-time binding (#2037, #2771).
E   assert 4653096752 == 4653240144
____________________ test_pool_is_blocking_with_popoto_cap _____________________
tests/unit/test_popoto_client_identity.py:96: in test_pool_is_blocking_with_popoto_cap
    assert isinstance(pool, redis.BlockingConnectionPool), (
E   AssertionError: popoto's client is on a ConnectionPool, not a redis.BlockingConnectionPool. popoto builds a BlockingConnectionPool so that an asyncio.gather burst blocks instead of raising MaxConnectionsError; a rebuilt pool drops that protection (#2771).
E   assert False
=========================== short test summary info ============================
FAILED tests/unit/test_popoto_client_identity.py::test_pool_identity_is_stable_across_tests
FAILED tests/unit/test_popoto_client_identity.py::test_client_object_identity_is_stable_across_tests
FAILED tests/unit/test_popoto_client_identity.py::test_pool_is_blocking_with_popoto_cap
3 failed, 2 passed in 61.31s (0:01:01)
```

Spike-1 probe on `main` (throwaway test, not committed) — the async global as observed from inside a test body:

```
ASYNC_IS_NONE: False
1 passed in 4.31s
```

After the change, both go green:

```
$ ./scripts/pytest-clean.sh tests/unit/test_popoto_client_identity.py -q -n0 -p no:randomly
5 passed in 8.62s

$ ./scripts/pytest-clean.sh <spike-1 probe> -q -n0 -p no:randomly
ASYNC_IS_NONE: True
1 passed in 7.22s
```

## Verification

| Check | Result |
|---|---|
| Targeted files (`-n4`, xdist) | 122 passed |
| `test_conftest_isolation_guards.py -p no:randomly -n0` | 48 passed |
| `test_popoto_client_identity.py -p no:randomly -n0` | 5 passed |
| `::test_pool_is_blocking_with_popoto_cap -n0` | 1 passed |
| `-k pool_identity -n0 -p no:randomly` | 2 passed |
| `test_db_derivation_guard.py -n0` (#2655 guard) | 49 passed |
| `test_test_redis_server_resolution.py -rs` | 13 passed, **0 skipped** |
| `ruff check .` / `ruff format --check .` | exit 0 |

All ANTI-criteria greps return no output, including `connection_pool = old_pool` filtered through `grep -v 'session-finalizer-only'` (the finalizer's restore line carries that marker so the row is mechanically judgeable).

## Deviations from the plan, on the record

1. **Full-suite runs descoped.** Plan tasks 1 (`baseline-main`) and 7 (`classify-failures`) call for full `tests/unit/` + `tests/integration/` runs on `main` and post-change, classified by `baseline-verifier`. The build was explicitly scoped to narrow, diff-relevant test selection. Consequence: Success Criteria "full unit and integration runs are green or baseline-classified" is **not** satisfied, and the Risk 1 mitigation (run the identity test with the maximum popoto submodule surface imported) is only partially exercised — it ran appended to a three-file selection (66 passed), not at the end of a full suite. Worth a full-suite run before merge.

2. **The `_swap_db` ANTI row is unjudgeable as written.** The row greps the bare name `_swap_db` in `tests/` expecting no output, but Technical Approach step 3 *mandates* the docstring explain why conftest does not delegate to `_swap_db`. The plan requires the string and forbids it — the same defect round 2 fixed for the pool-restore row. The mandated explanation was kept; the row's intent is verified instead with the call form, `grep -rnE '_swap_db\s*\(' tests/` → **no output**, zero call sites. The row should be restated to match the call form.

3. **One fix outside the plan's enumerated scope.** Deleting the split-brain test's `redis.Redis(db=base_test_db)` dropped the tree's attribute-qualified Redis-construction count from 15 to 14, breaking two anti-vacuity floors in `tests/unit/test_db_derivation_guard.py` that were hardcoded to exactly the then-current total. A floor pinned to the live total fails any cleanup that is not about it. They are now named constants (`_MIN_DB_KWARG_SITES`, `_MIN_FROM_URL_SITES`, `_MIN_ATTRIBUTE_QUALIFIED_SITES`) set below the live totals, with a comment marking them provisional. Commit `3c0a6f2bf`, kept separate.


valorengels:
## Review: PR #3083 — consolidate the two owners of the popoto test db (#2771)

Reviewed at head `0df503a17`, from a clean worktree at that exact SHA. Plan: `docs/plans/consolidate-two-owners-of-popoto-test-db.md`. Every number below was reproduced locally, not taken from the PR body.

**Verdict: CHANGES REQUESTED** — no blockers. One piece of tech debt and two nits.

### What holds up

The ownership split lands exactly as the plan specifies, and the plan's own falsifiability machinery works.

- `_popoto_pool_install` is session-scoped, `autouse=True`, and depends on the plugin's `_popoto_test_db` — ordering by construction, not by collection accident. It mutates `client.connection_pool` in place and never rebinds `rdb.POPOTO_REDIS_DB`. The idempotence guard type-normalizes host/port on both sides. `password`/`username` are carried forward. `old_pool.disconnect()` carries the "synchronous setup phase, nothing in flight" comment. The finalizer restore line carries `# session-finalizer-only` verbatim, which is what makes the ANTI row mechanically judgeable.
- `redis_test_db` keeps its name, `autouse=True`, and function scope. Its body is now `_assert_client_matches_claim_registry` + `flushdb()` / yield / `flushdb()`. No pool work, no `disconnect()`, no `close()`, no object rebinding, no async write. The `import redis.asyncio as aioredis` line is gone.
- Deletions are complete: `_popoto_modules_with_redis_db`, `_POPOTO_MODULE_CACHE`, `_POPOTO_MODULE_CACHE_LEN`, the comment block, the repatch loop, the eager async bind and its save/restore, and the whole cache-invalidation test class. No compatibility shim, no deprecation alias, no commented-out original anywhere in the diff.
- All four identity assertions are present in `tests/unit/test_popoto_client_identity.py`, including (d) pool-object identity — the fence against the round-1 BLOCKER.
- Docs describe only the new status quo. The `## Popoto module re-pointing cache invalidation` heading is renamed to `## Popoto client identity` (plan Documentation item at `test-isolation-hardening.md:7`, checked against the heading rather than only the line ranges). No "previously we..." narration survives.
- No `Co-Authored-By`, no Claude/Anthropic attribution trailer, no robot emoji on any of the six commits.

**Anti-criteria, re-run by me** — all clean:

`rdb.POPOTO_REDIS_DB =` in conftest, `rdb._POPOTO_ASYNC_REDIS_DB =` in conftest, `_popoto_modules_with_redis_db` in `tests/`, the same in `docs/features/` + `tests/`, `popoto exposes no async client`, `redis.Redis(db=` in the guard file, `max_connections\s*=\s*[0-9]` in conftest, and `connection_pool = old_pool` filtered through `grep -v 'session-finalizer-only'` — every one returns no output.

**Tests, re-run by me** (not taken from the PR body):

| Check | Observed |
|---|---|
| the four targeted files + identity, `-n4` | 122 passed |
| `test_conftest_isolation_guards.py -p no:randomly -n0` | 48 passed |
| `test_popoto_client_identity.py -p no:randomly -n0` | 5 passed |
| `test_db_derivation_guard.py -n0` | 49 passed |
| six popoto-model-heavy `test_agent_session*` files, `-n4` | 210 passed |
| `ruff check .` / `ruff format --check .` | exit 0 |

The 210-test model-heavy slice is my own addition — the autouse blast radius is the risk here, and I wanted independent evidence beyond the diff-adjacent files.

---

### Blockers
- None

---

### Tech Debt

**1. The cross-test identity fences are vacuously green in the suite's default run mode.**

`tests/unit/test_popoto_client_identity.py:36-43` and `:52-59` both open with:

```python
if "pool_id" not in _OBSERVED:
    return  # baseline did not run in this process; nothing to compare
```

`_OBSERVED` is populated by `test_pool_identity_baseline_is_recorded`, a separate test. Under the suite's default configuration — xdist plus `pytest-randomly`, which is how `tests/unit/` actually runs — neither co-location nor ordering is guaranteed. I confirmed both halves:

- `./scripts/pytest-clean.sh "tests/unit/test_popoto_client_identity.py::test_pool_identity_is_stable_across_tests" -q -n0 -p no:randomly` → **1 passed**, with no baseline in the process. The assertion never executed.
- `./scripts/pytest-clean.sh tests/unit/test_popoto_client_identity.py -q -n4` → 5 passed across 4 workers, so at least two of the five ran without their baseline.

Assertion (d) is the plan's explicit fence against the round-1 BLOCKER (plan Test Impact and Success Criteria both call it that). In the run mode the merge gate actually uses, it can silently no-op. The plan's `-n0 -p no:randomly` Verification rows do arm it, so this is not a blocker — but the fence is weaker than the plan's own framing claims.

Second, smaller half of the same finding: the module docstring at `tests/unit/test_popoto_client_identity.py:28-31` says a stability test with no baseline "has nothing to compare against and **says so** rather than failing spuriously." It does not say so. A bare `return` reports a pass and emits nothing. `pytest.skip("baseline did not run in this process")` would make the vacuity visible in `-rs` output and would make the docstring true.

Suggested fix, smallest form: swap the two `return`s for `pytest.skip(...)`. Better form: record the baseline in a session-scoped fixture (or a `pytest.mark.order`-free single test that observes twice via a module-level fixture) so the fence is armed under xdist and random ordering, not only under the two pinned Verification rows.

---

### Nits

**2. `redis_test_db`'s `request` parameter is now unused.** `tests/conftest.py:957` — `def redis_test_db(request, _popoto_pool_install):`. The rewritten body (`:989-997`) uses only `rdb` and `db_claim`. `request` was load-bearing for the old body; nothing reads it now. Drop it.

**3. The plan's `_swap_db` ANTI row is unjudgeable as written** (plan Verification, No-Go #2770 row). It greps the bare name `_swap_db` in `tests/` expecting no output, while Technical Approach step 3 *mandates* the `redis_test_db` docstring explain why conftest does not delegate to `_swap_db`. The plan requires the string and forbids it — structurally the same defect round 2 fixed for the pool-restore row by adding the `session-finalizer-only` marker.

The PR discloses this and substitutes the call form. I verified independently: `grep -rnE '_swap_db\s*\(' tests/` → **no output**, zero call sites; the six bare-name hits are all prose in docstrings and comments. The row's intent holds. This is a plan-authoring fix (restate the row as the call form), not a code change.

---

### Acknowledged

**4. Full-suite Success Criterion is unmet, and disclosed.** Plan tasks 1 (`baseline-main`) and 7 (`classify-failures`) call for full `tests/unit/` + `tests/integration/` runs on `main` and post-change, classified by `baseline-verifier`. They were not run; the lane was scoped to narrow test selection. The plan's Success Criterion at line 460 ("Full `tests/unit/` and `tests/integration/` runs are green, or every failure is classified against a pre-change `main` baseline") is therefore **not satisfied**, and the Risk 1 mitigation — run the identity test at the end of a full suite so the maximum popoto submodule surface is imported — is only partially exercised.

This is correctly and prominently disclosed in the PR body under "Deviations from the plan, on the record", which is the right handling. It is a gate for `/do-merge`, not a defect in this diff. Given that `redis_test_db` is autouse over the entire suite, the full run should happen before merge rather than after.

---

### On commit `3c0a6f2bf` — justified consequence, not scope creep

The commit changes anti-vacuity floors in `tests/unit/test_db_derivation_guard.py`, outside the plan's enumerated file set. I judged it deliberately and it is in bounds:

- **It was forced by an in-scope edit.** Plan task 4 mandates replacing both bare `redis.Redis(db=...)` clients in `TestPopotoSplitBrainRoundTrip` with `db_claim`-resolved host/port, and the No-Gos explicitly keep those two in scope ("*are* fixed here because that test is being edited anyway"). Doing so mechanically drops the tree's `db=` keyword and attribute-qualified Redis-construction counts below floors that were hardcoded at exactly the then-current total of 15. The suite cannot be green without touching that file. This is a consequence of the mandated edit, not an unrelated repair folded in.
- **The fix is the right shape, not a number substitution.** A floor pinned to the live total fails any cleanup that is not about it — swapping 15 for 14 would just re-arm the same trap for the next PR. The commit names three constants (`_MIN_DB_KWARG_SITES`, `_MIN_FROM_URL_SITES`, `_MIN_ATTRIBUTE_QUALIFIED_SITES`), sets them below the live totals, and comments what they discriminate and that they are provisional and tunable. That is exactly this repo's recorded convention for magic numbers.
- **The floors are verified genuinely below the live totals.** Reproduced from the guard's own diagnostic output: `db=` keyword sites 14 vs floor 12, `from_url` sites 9 vs floor 8, attribute-qualified 14 vs floor 12. `test_db_derivation_guard.py -n0` → 49 passed.
- **It is isolated.** Separate commit, message states the mechanism and the reasoning, touches nothing else.

One observation, not a finding: `_MIN_FROM_URL_SITES` is 8 against a live 9, so the next `from_url` deletion anywhere in the tree trips it again. That headroom of one is inherited — the floor was already 8 before this PR — so it is not introduced here. Worth widening opportunistically if that file is touched again.

---

### Rubric

| # | Item | Result |
|---|---|---|
| 1 | Implementation matches the plan | pass |
| 2 | No-Gos respected (#2770 upstream untouched, #3003 sweep not attempted, #2628 red-main not folded in) | pass |
| 3 | Correctness / no regressions in what was run | pass (429 tests across five selections, all green) |
| 4 | Security | n/a — test-tree only, no production code, no new credential surface |
| 5 | Error handling | pass — `_assert_client_matches_claim_registry` stays unguarded, no new `except` in the fixture, `old_pool.disconnect()` unguarded per plan |
| 6 | Test coverage | pass with tech debt (finding 1) |
| 7 | Code quality / no legacy cruft | pass — no shim, no alias, no commented-out original |
| 8 | Documentation | pass — heading renamed, three docs and two `db_claim` docstrings updated, new status quo only |
| 9 | Plan section compliance (Documentation / Update System / Agent Integration / Test Impact) | pass — all four present in the plan |
| 10 | Lint & format gates | pass |
| — | Misc: full-suite criterion | acknowledged (finding 4) |
| — | Misc: UI / visual proof | n/a — no UI files in the diff |
| — | Misc: multi-machine (env vars, migrations) | n/a — none added |
| — | Misc: bridge/worker restart | n/a — `tests/` and `docs/` only |

Findings 1-3 are all within `/do-patch`'s reach. Finding 4 is a run to schedule, not a code change.

valorengels:
## Review: PR #3083 — consolidate the two owners of the popoto test db (#2771)

Re-review at head `f3804ec9b6c0003937104b4454dae504140aa258`, resolved via `tools/pr_head_resolver.py::resolve_pr_head_sha` (git-first, `refs/pull/3083/head`), not a bare `gh` read. This supersedes the `CHANGES REQUESTED` review posted at head `0df503a17` — every finding from that round was addressed by commit `f3804ec9b`.

**Verdict: APPROVED** — 0 blockers, 1 tech-debt, 3 nits. Changes not required before merge.

### Prior round's findings, all resolved

- **Tech-debt (vacuous identity fence under xdist)** — fixed. The cross-test `_OBSERVED` dict in `tests/unit/test_popoto_client_identity.py` is replaced by a session-scoped `installed_identity` fixture that computes its own in-process baseline, so the fence is armed in every worker rather than only in whichever worker happened to run first. Re-verified: the reviewer's exact repro `test_pool_identity_is_stable_across_tests` passes standalone under `-n0 -p no:randomly`, and the file passes 4/4 under `-n4`.
- **Nit (unused `request` param)** — removed from `redis_test_db` in `tests/conftest.py`.
- **Nit (unjudgeable ANTI row)** — the plan's `_swap_db` ANTI row in `docs/plans/consolidate-two-owners-of-popoto-test-db.md` is restated as the call form `_swap_db\s*\(`, which is now actually greppable.

### Remaining tech-debt (accepted, not a merge blocker)

Residual session-end flush of the ambient server's db N under private-port runs. This is **pre-existing upstream behavior that this PR reduces rather than introduces**, and it is already tracked at #2770. Teardown tracing confirms there is **no production db-0 flush hazard**: the flush path resolves to the claimed test db, and the ambient flush guard (`tools/redis_flush_guard.py`) remains armed against db 0 and `flushall`.

### Nits (non-blocking)

1. **Dead defensive `already_installed` branch.** Unreachable given the session-scoped autouse ordering. Harmless, but it is scar tissue — a future reader will assume it guards something real.
2. **Fail-loud shared-client poisoning.** The design deliberately fails loudly rather than silently degrading when the shared client is poisoned. Accepted as the right call, and adequately fenced by the identity test.
3. **Session-long `BlockingConnectionPool` cap.** Worth watching on the first full-suite run: a session-scoped pool cap interacts with xdist worker count in a way that only shows up at full breadth. Not a defect, a watch item.

### Merge gate still owed

The full `tests/unit/` + `tests/integration/` run has **not** been executed for this PR and remains an owed gate. Targeted runs (101/101) pass, `ruff check` and `ruff format --check` are clean, `mergeStateStatus` is CLEAN. Do not merge until the full run lands, per the fourth finding carried over from the prior round.

valorengels:
## Review: PR #3083 — consolidate the two owners of the popoto test db (#2771)

Confirmation re-review at head `f3804ec9b6c0003937104b4454dae504140aa258`, resolved via `tools/pr_head_resolver.py::resolve_pr_head_sha` (git-first). Unchanged from the head the prior `APPROVED` verdict was recorded against, so this run is a targeted confirmation of the round-1 critique BLOCKER rather than a fresh full review.

### Round-1 critique BLOCKER: pool ownership is session-scoped — CONFIRMED RESOLVED

Verified by reading the blob at the exact head SHA (`git show f3804ec9b:tests/conftest.py`), not the working tree:

- `tests/conftest.py:860-861` — `@pytest.fixture(scope="session", autouse=True)` / `def _popoto_pool_install(_popoto_test_db)`. Session scope and autouse are both present, and the `_popoto_test_db` dependency orders it after popoto's plugin swap by construction rather than by collection accident.
- `tests/conftest.py:956-957` — `redis_test_db` keeps `autouse=True` at function scope and depends on `_popoto_pool_install`. Its body is `_assert_client_matches_claim_registry(...)`, `flushdb()`, `yield`, `flushdb()`. **No pool work, no restore, no rebind.**
- The pool restore is single-sited. `git grep -n 'connection_pool\s*=' -- tests/` at this SHA returns exactly two writes to the live client:
  - `conftest.py:938` `client.connection_pool = new_pool` (session install)
  - `conftest.py:952` `client.connection_pool = old_pool  # session-finalizer-only` (session finalizer, after `yield`)

  Every other hit is a `SimpleNamespace`/stub construction in test fixtures, not a write to `rdb.POPOTO_REDIS_DB`. The ANTI grep `grep 'connection_pool = old_pool' tests/ | grep -v 'session-finalizer-only'` returns no output.
- `grep -rnE '_swap_db\s*\(' tests/` returns no output — zero call sites, confirming the call-form restatement of the No-Go #2770 row.

The invariant the blocker was about holds: the pool is installed once per process and unwound once, after the last test.

### Note (non-blocking): the idempotence guard is not reachable on any current path

`conftest.py:911-919` gates on `isinstance(old_pool, redis.BlockingConnectionPool)` plus host/port/db equality. `_popoto_pool_install` declares `_popoto_test_db` as a dependency, and that plugin fixture (`popoto/pytest_plugin.py:164`) unconditionally calls `_swap_db(test_db)`, which at line 123 always constructs a plain `redis.ConnectionPool` — never a `BlockingConnectionPool` (and `issubclass(redis.ConnectionPool, redis.BlockingConnectionPool)` is `False`). So by the time this fixture body runs, `old_pool` is always a plain pool and `already_installed` is always `False`.

This is defensive dead code, not a defect, and it fails in the safe direction: if the guard ever did trigger spuriously, `_assert_client_matches_claim_registry` fires on the first test and `test_pool_is_blocking_with_popoto_cap` fails, so a skipped install cannot pass silently. Recording it so the next reader does not re-derive it. The adjacent comment's phrase "rebuild the pool on every call" also slightly overstates the call frequency — a session-scoped fixture body runs once per process.

### Commit hygiene — clean

`git log --format='%B' origin/main..f3804ec9b | grep -iE 'co-authored-by|generated with|claude code|noreply@anthropic|🤖'` returns no matches across all seven commits on the branch. No AI-generated footers or co-author trailers.

### Gates

| Check | Result |
|---|---|
| `ruff check .` | All checks passed |
| `ruff format --check .` | 1430 files already formatted |
| Mergeability | `OPEN` / `MERGEABLE` / `CLEAN` |
| ANTI grep `_swap_db\s*\(` in `tests/` | no output |
| ANTI grep pool-restore minus finalizer marker | no output |
| `docs/features/test-db-ownership.md` | present (commit `0df503a17`) |

`tests/unit/test_popoto_client_identity.py` was **not** re-run in this environment: all 15 test-DB slots were held by live sibling pytest runs across two attempts, and reaping them would have destroyed other agents' in-flight runs. Per the "a claimed number is not evidence" rule I am not crediting the PR body's `5 passed` as independently reproduced here — but the blocker under review is a structural property of the fixture graph, and reading the blob at the head SHA settles it directly without needing the suite.

### Findings

**Blockers**
- None

**Tech Debt**
- None

**Nits**
- None

### Verdict

**APPROVED** — consistent with the verdict already recorded at this same head SHA. The round-1 critique BLOCKER is confirmed resolved in code: pool ownership is session-scoped, `redis_test_db` performs no pool restore, and the only unwind is the marked session finalizer.

valorengels:
## Review: PR #3083 — consolidate the two owners of the popoto test db (#2771)

Re-review at head `b21774ae23265e7d4397fb6b3228f08d61d2705d`. The prior `APPROVED` verdict was recorded at `f3804ec9b`; this run is scoped to the delta since then, because the full diff has already been reviewed against the plan twice at unchanged content.

### Delta since the approved head — one docs index row

`git diff f3804ec9b b21774ae2` is exactly:

```
 docs/features/README.md | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)
```

One commit, `b21774ae2 Docs(#2771): retire the compound-cache description in the feature index row`. The row for `test-isolation-hardening.md` changes from

> Single-run, cross-file xdist phantom-failure fixes: **a compound invalidation trigger for the popoto db-cache** and an autouse guard repairing hook corruption.

to

> Single-run, cross-file xdist phantom-failure fixes: **one canonical popoto client per process whose connection pool is swapped in place once per session**, and an autouse guard repairing hook corruption.

**Verified against the doc it indexes**, not against the commit message. `docs/features/test-isolation-hardening.md` at this head opens its "Popoto client identity" section with "There is exactly one `POPOTO_REDIS_DB` client object per pytest process. `tests/conftest.py`'s session-scoped, autouse `_popoto_pool_install` fixture mutates that object's `connection_pool` **in place** — once per session". The new row is a faithful one-line summary of that. The old row described the compound cache-invalidation trigger, which this PR **deleted** (`_popoto_modules_with_redis_db`, `_POPOTO_MODULE_CACHE`, `_POPOTO_MODULE_CACHE_LEN` and the submodule repatch loop are gone from `tests/conftest.py`) — so the pre-delta index row was describing machinery that no longer exists. This is the NO LEGACY CODE / no-historical-artifacts-in-docs rule applied to the index, and it closes the last stale reference.

`grep -c 'test-isolation-hardening.md' docs/features/README.md` → `1`. Single row, no duplicate entry introduced, table structure and `Shipped` status column unchanged.

### Scope of the delta

Markdown only. No Python, no fixtures, no tests, no config, no UI files. The delta cannot move `ruff check` / `ruff format --check` (both confirmed exit 0 at `f3804ec9b` in the prior review), cannot change any test outcome, and triggers no visual-proof gate.

### Carried forward from the prior review at `f3804ec9b`

All of it still holds — the code at this head is byte-identical to the head that was approved:

- Round-1 critique BLOCKER (session-scoped pool ownership) confirmed resolved: `_popoto_pool_install` is `scope="session", autouse=True` and depends on `_popoto_test_db`; `redis_test_db` does no pool work; the only unwind is the marked session finalizer.
- ANTI greps clean: `_swap_db\s*\(` in `tests/` → no output; pool-restore grep minus the `session-finalizer-only` marker → no output.
- Commit hygiene clean across the branch — no AI co-author or generated-with trailers.
- Mergeability: `OPEN` / `MERGEABLE` / `CLEAN`.
- The non-blocking note about the unreachable `already_installed` idempotence guard stands as recorded; it is defensive dead code that fails in the safe direction, not a defect.

The prior review's caveat also carries: `tests/unit/test_popoto_client_identity.py` was not independently re-run here (test-DB slots contended by live sibling runs, and reaping them would destroy other agents' runs). The PR body's `5 passed` remains **unverified in this environment** and is not credited toward the verdict. It does not need to be for this delta, which is a documentation string.

### Deviations the PR body records — unchanged, still on the record

The PR body's "Deviations from the plan" section (full-suite runs descoped; the `_swap_db` ANTI row restated to the call form; the anti-vacuity floors in `test_db_derivation_guard.py` converted to named constants) is unchanged by this delta and was adjudicated in the prior rounds. Deviation 1's suggestion — a full-suite run before merge — remains a merge-stage consideration, not a review blocker.

### Findings

**Blockers**
- None

**Tech Debt**
- None

**Nits**
- None

### Verdict

**APPROVED** — the sole change since the recorded `APPROVED` verdict is a one-line feature-index row that replaces a description of deleted machinery with an accurate description of the shipped mechanism. It improves doc accuracy and introduces no risk.

Issue #2643, opened 2026-08-07
Watchdog unit tests write synthetic CRITICAL lines into production logs/watchdog.log via import-time handler


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/watchdog-import-time-log-handler.md

> **Context:** [`monitoring/bridge_watchdog.py`](monitoring/bridge_watchdog.py) is an external health monitor for the Telegram bridge, run every 60s by launchd. Its durable record is `logs/watchdog.log` — per the module docstring, that file plus `log_crash()` *is* the alerting system ("The watchdog does NOT push a notification anywhere. It records; it does not deliver."). When a human investigates a bridge incident, `logs/watchdog.log` is the primary and often only evidence. This issue is about that file being contaminated with output from unit tests.

## Problem

**Importing `monitoring/bridge_watchdog.py` attaches a file handler to the real `logs/watchdog.log` as an import-time side effect. The unit test suite imports it, so every test run writes synthetic CRITICAL lines into the production incident log.**

### Current behavior

Two side effects fire at module import (`monitoring/bridge_watchdog.py:78-93`), before any function is called:

```python
logging.basicConfig(level=logging.INFO, format="...")     # :78 — mutates the ROOT logger
logger = logging.getLogger(__name__)                       # :82

_watchdog_file_handler = logging.handlers.RotatingFileHandler(
    LOGS_DIR / "watchdog.log", maxBytes=10 * 1024 * 1024, backupCount=5,
)                                                          # :87
logger.addHandler(_watchdog_file_handler)                  # :93
```

`tests/unit/test_bridge_watchdog.py` imports `check_bridge_health` and friends from this module, so the handler is live for the whole pytest process. Tests that exercise the recovery ladder then emit real `logger.critical(...)` calls straight into the operator's log.

Measured on the current `logs/watchdog.log` in this checkout:

- 86 total `Non-wedge crash storm — restart circuit open, skipping restart to avoid thrash` lines.
- **44 are genuine production** (a single cluster, 2026-07-27 11:55–12:19, on `Valor-the-Cowboy.local`).
- **42 are synthetic test output** — i.e. roughly half the occurrences of the log's most alarming line never happened.
- 121 lines contain the test fixture's `pid=1234`; 93 name the test host `MacBookPro.local`.

[...]

PR #3144, opened 2026-09-04
Test processes never attach the production bridge.log handler

### LINKED ISSUE #2854: Nightly unit tests leak fixture ERROR lines into production logs/bridge.log
## Problem

The nightly unit-test run writes fixture-generated ERROR lines into the production `logs/bridge.log`. Observed daily at ~20:15-20:20 UTC since at least 2026-08-11:

```
{"level": "ERROR", "logger": "bridge.room_inbox", "function": "shadow_append_inbox", "message": "[room-inbox] shadow append skipped: no project key (chat=1 msg=1)"}
{"level": "ERROR", "logger": "bridge.room_inbox", "function": "shadow_append_inbox", "message": "[room-inbox] shadow append failed for chat=1 msg=1: redis down"}
```

These match `tests/unit/test_room_inbox_shadow.py` failure-branch fixtures exactly (`chat_id=1` at lines 79/91/108, `raise RuntimeError(\"redis down\")` at line 103). The test process inherits or configures the production JSON file handler, so test-only failure branches masquerade as production errors.

## Why it matters

- Pollutes operator triage: a grep for `[room-inbox]` errors reads as daily production failures.
- Directly degraded the issue #2494 phase-2 gate check ("N days of error-free [room-inbox] shadow appends") — the gate reviewer had to trace each ERROR line back to test fixtures to rule leg 1 a pass (see plan doc `docs/plans/durability-room-job-agentrun.md`, phase-2 gate note of 2026-08-18).
- Any log-anchored health signal is a false-alarm surface while tests can write to production log files.

## Acceptance criteria

- [ ] Root cause identified: which runner (nightly reflection, scheduled pytest, or conftest logging setup) attaches the production file handler in test processes.
- [ ] Test runs never append to `logs/bridge.log` (or any production log file) — handler isolation in conftest or the runner, not per-test suppression.
- [ ] A regression guard: a test or check asserting no production log-file handler is active under pytest.
- [ ] Verified: the daily `chat=1 msg=1` ERROR lines stop appearing in `logs/bridge.log`.

tomcounsell:
Closes #2854

## Root cause (acceptance criterion 1)

Neither the nightly runner nor conftest: the attach happens at **module import time** in `bridge/telegram_bridge.py`. It adds a `RotatingFileHandler` for `<checkout>/logs/bridge.log` to the **root** logger as an import side effect, so the first test in any pytest worker that imports the module (directly or transitively; many unit test files do) poisons that worker for the rest of its schedule. From then on every logger's records propagate to the production log, which is why `tests/unit/test_room_inbox_shadow.py`'s deliberate failure-branch fixtures (`chat=1 msg=1`, `redis down`) appeared in `logs/bridge.log` nightly. Reproduced deterministically: importing the module and logging one ERROR through `bridge.room_inbox` wrote the exact issue signature to the file.

`config/settings.py::setup_logging` also builds a file handler but only runs via an explicit `validate_configuration()` call, not at import; it was not the leak path.

## Fix (criterion 2: isolation at the source, not per-test suppression)

Under pytest (`"pytest" in sys.modules or "PYTEST_CURRENT_TEST" in os.environ`) the file handler is neither constructed nor attached, so a test run never creates or writes the file. `PYTEST_CURRENT_TEST` is exported into the process environment by pytest and inherited by any subprocess a test spawns, so a bridge subprocess spawned BY a test also skips the handler; only a process started outside any pytest run attaches it, which is the intended production behavior. Verified directly: with the env scrubbed, root handlers include the bridge.log path; with `PYTEST_CURRENT_TEST` set, they do not.

## Regression guard (criterion 3)

`tests/unit/test_no_production_log_handlers.py` imports the known attacher first (deterministic regardless of xdist scheduling), then asserts no handler on the root or bridge loggers targets a file under `logs/`.

## Evidence (criterion 4)

| Configuration | Guard test | logs/bridge.log |
|---|---|---|
| Fix active | passes | never created |
| Guard neutralized (`_UNDER_PYTEST = False`) | FAILS | gains 3 `chat=1 msg=1` fixture lines |
| Production probe (no pytest in modules) | n/a | handler attached, path correct |

Sanity slice (guard + shadow + bridge logic + log rotate + media timeout): 82 passed. Ruff check and format clean.

### DIFF
--- a/bridge/telegram_bridge.py
+++ b/bridge/telegram_bridge.py
@@ -596,28 +596,41 @@ logging.basicConfig(
 # bridge.*, tools.*, etc.) inherit it automatically. Without this, only
 # the bridge.telegram_bridge module logger would write to bridge.log.
 root_logger = logging.getLogger()
-file_handler = logging.handlers.RotatingFileHandler(
-    LOG_DIR / "bridge.log",
-    maxBytes=10 * 1024 * 1024,  # 10MB per file
-    backupCount=5,
-)
-file_handler.setLevel(logging.DEBUG)
-# Use JSON formatter for structured logging (parseable by log aggregation tools)
-# Falls back to plain text if the module can't be imported
-try:
-    from bridge.log_format import StructuredJsonFormatter
+# Never build or attach the production file handler inside a pytest process
+# (#2854). This module-scope setup runs on import, so any test importing this
+# module (directly or transitively) was routing every logger's records into
+# logs/bridge.log, including test fixtures' deliberate failure-branch ERRORs,
+# which then read as daily production failures. "pytest" is in sys.modules
+# during collection and test phases alike; a bridge subprocess spawned BY a
+# test imports this module without pytest loaded and still attaches, which is
+# the intended production behavior. PYTEST_CURRENT_TEST covers a spawned
+# child that inherits the env but re-imports lazily mid-test. Constructing
+# the handler is also skipped so a test run never even creates the file.
+_UNDER_PYTEST = "pytest" in sys.modules or "PYTEST_CURRENT_TEST" in os.environ
+if not _UNDER_PYTEST:
+    file_handler = logging.handlers.RotatingFileHandler(
+        LOG_DIR / "bridge.log",
+        maxBytes=10 * 1024 * 1024,  # 10MB per file
+        backupCount=5,
+    )
+    file_handler.setLevel(logging.DEBUG)
+    # Use JSON formatter for structured logging (parseable by log aggregation
+    # tools); falls back to plain text if the module can't be imported
+    try:
+        from bridge.log_format import StructuredJsonFormatter
 
-    file_handler.setFormatter(StructuredJsonFormatter())
-except ImportError:
-    file_handler.setFormatter(logging.Formatter(FILE_FORMAT))
-file_handler.addFilter(InternalDebugFilter())
-# Guard against double-import adding the handler twice (e.g. __main__ + bridge.telegram_bridge)
-if not any(
-    isinstance(h, logging.handlers.RotatingFileHandler)
-    and getattr(h, "baseFilename", None) == str(LOG_DIR / "bridge.log")
-    for h in root_logger.handlers
-):
-    root_logger.addHandler(file_handler)
+        file_handler.setFormatter(StructuredJsonFormatter())
+    except ImportError:
+        file_handler.setFormatter(logging.Formatter(FILE_FORMAT))
+    file_handler.addFilter(InternalDebugFilter())
+    # Guard against double-import adding the handler twice
+    # (e.g. __main__ + bridge.telegram_bridge)
+    if not any(
+        isinstance(h, logging.handlers.RotatingFileHandler)
+        and getattr(h, "baseFilename", None) == str(LOG_DIR / "bridge.log")
+        for h in root_logger.handlers
+    ):
+        root_logger.addHandler(file_handler)
 
 # Module logger for this file. It inherits the root logger's file handler,
 # so we only need to set its level to DEBUG for verbose local output.
--- a/bridge/telegram_bridge.py
+++ b/bridge/telegram_bridge.py
@@ -601,11 +601,14 @@ root_logger = logging.getLogger()
 # module (directly or transitively) was routing every logger's records into
 # logs/bridge.log, including test fixtures' deliberate failure-branch ERRORs,
 # which then read as daily production failures. "pytest" is in sys.modules
-# during collection and test phases alike; a bridge subprocess spawned BY a
-# test imports this module without pytest loaded and still attaches, which is
-# the intended production behavior. PYTEST_CURRENT_TEST covers a spawned
-# child that inherits the env but re-imports lazily mid-test. Constructing
-# the handler is also skipped so a test run never even creates the file.
+# during collection and test phases alike; PYTEST_CURRENT_TEST is exported by
+# pytest into the process environment, so a bridge subprocess spawned BY a
+# test inherits it and also skips the handler (a test-spawned bridge should
+# not write the operator's production log). Only a process started outside
+# any pytest run — neither "pytest" in sys.modules nor PYTEST_CURRENT_TEST in
+# its environment — attaches it, which is the intended production behavior.
+# Constructing the handler is also skipped so a test run never even creates
+# the file.
 _UNDER_PYTEST = "pytest" in sys.modules or "PYTEST_CURRENT_TEST" in os.environ
 if not _UNDER_PYTEST:
     file_handler = logging.handlers.RotatingFileHandler(
--- a/docs/features/watchdog-log-isolation.md
+++ b/docs/features/watchdog-log-isolation.md
@@ -133,3 +133,42 @@ No behavioral change was made to recovery, escalation, health-check, or
 rotation logic — `LOG_MAX_SIZE`, `LOG_MAX_BACKUPS`, `LOG_BACKUP_HARD_CAP`, and
 `SELF_EXCLUDED_FILES` in `scripts/log_rotate.py` are untouched, and rotation
 settings (`maxBytes`/`backupCount`) are unchanged on both watchdogs.
+
+## A second instance of this bug class, solved a different way (#2854)
+
+`bridge/telegram_bridge.py` attaches its own `RotatingFileHandler` for
+`logs/bridge.log` to the root logger at **module scope** — exactly the
+pattern this doc's invariant forbids. It was not moved behind an
+`if __name__ == "__main__":` guard like the three modules above. The reason
+is `telegram_bridge.py`'s doc-comment: it keeps backward-compatible
+re-exports so `agent_catchup`, `telegram_relay`, `email_bridge`,
+`answer_routing`, and `routing` can all `import bridge.telegram_bridge` and
+use its symbols without ever running the module as `__main__` — there is no
+single entry point where "configure logging now" can live the way there is
+for a watchdog script.
+
+The sanctioned alternative for this shape is a **pytest-presence guard**
+rather than an entry-point relocation:
+
+```python
+_UNDER_PYTEST = "pytest" in sys.modules or "PYTEST_CURRENT_TEST" in os.environ
+if not _UNDER_PYTEST:
+    file_handler = logging.handlers.RotatingFileHandler(...)
+    ...
+```
+
+`"pytest" in sys.modules` catches the direct case: any pytest process that
+imports the module, at collection or test time. `PYTEST_CURRENT_TEST` is
+exported into the process environment by pytest for the duration of each
+test and is inherited by any subprocess a test spawns, so a bridge process
+started *by* a test also skips the handler — the desired behavior, since a
+test-spawned bridge should not write the operator's production log. Only a
+process started outside any pytest run (neither condition true) attaches it.
+The regression guard is `tests/unit/test_no_production_log_handlers.py`,
+paired with a companion subprocess probe (mirroring this file's `_run_probe`
+pattern) that pins the production side stays attached when neither
+condition holds.
+
+Use the entry-point-guard pattern from the rest of this doc when a module
+has one true entry point; use the `_UNDER_PYTEST` pattern when a module is
+imported for its symbols from many call sites with no single `__main__`.
--- a/tests/unit/test_no_production_log_handlers.py
+++ b/tests/unit/test_no_production_log_handlers.py
@@ -20,9 +20,11 @@ condition were inverted. ``test_production_process_still_attaches_the_handler``
 below is the companion probe (precedent:
 ``tests/unit/test_watchdog_log_isolation.py::_run_probe``) that runs the
 same import in a subprocess with ``PYTEST_CURRENT_TEST`` stripped and
-``pytest`` never imported, asserting the ``RotatingFileHandler`` for
-``logs/bridge.log`` IS attached to the root logger — pinning that production
-behavior is unchanged.
+``pytest`` never imported, asserting a ``RotatingFileHandler`` requesting
+exactly ``logs/bridge.log`` IS attached to the root logger — pinning that
+production behavior is unchanged. The probe redirects the handler's actual
+file into a tmp directory, so it never reads, writes, rotates, or restores
+the real production log.
 """
 
 from __future__ import annotations
@@ -31,6 +33,7 @@ import json
 import logging
 import os
 import pathlib
+import shutil
 import subprocess
 import sys
 from pathlib import Path
@@ -75,50 +78,82 @@ def test_production_process_still_attaches_the_handler():
     "pytest" is not in ``sys.modules``) with ``PYTEST_CURRENT_TEST`` scrubbed
     from the environment, mirroring
     ``tests/unit/test_watchdog_log_isolation.py::_run_probe``.
+
+    The probe must never touch the REAL ``logs/bridge.log``: a live bridge
+    (or a nightly run in the serving checkout) writes it concurrently, and a
+    snapshot/restore around the subprocess was proven to destroy those
+    concurrent writes. So BEFORE the import, the probe patches
+    ``RotatingFileHandler.__init__`` to record the requested filename and
+    redirect the actual open into a private tmp directory. The assertion is
+    on the recorded request — exact-path equality with
+    ``REPO_ROOT/logs/bridge.log`` — while the bytes land in the tmp file.
     """
     probe = f"""
-import json, logging, logging.handlers, sys
+import json, logging, logging.handlers, pathlib, sys, tempfile
 sys.path.insert(0, {str(REPO_ROOT)!r})
+
+requested = []
+tmp_dir = tempfile.mkdtemp(prefix="probe-bridge-log-")
+_real_init = logging.handlers.RotatingFileHandler.__init__
+
+def _redirecting_init(self, filename, *args, **kwargs):
+    requested.append(str(filename))
+    redirected = pathlib.Path(tmp_dir) / pathlib.Path(filename).name
+    _real_init(self, str(redirected), *args, **kwargs)
+
+# Patch the class __init__ (not a name binding), so every construction form
+# — logging.handlers.RotatingFileHandler(...) included — is redirected.
+logging.handlers.RotatingFileHandler.__init__ = _redirecting_init
+
 import bridge.telegram_bridge as m
 assert m.__file__.startswith({str(REPO_ROOT)!r}), m.__file__
 root = logging.getLogger()
-has_bridge_log_handler = any(
-    isinstance(h, logging.handlers.RotatingFileHandler)
-    and "bridge.log" in getattr(h, "baseFilename", "")
-    for h in root.handlers
-)
-print(json.dumps({{"has_bridge_log_handler": has_bridge_log_handler}}))
+attached = [
+    h for h in root.handlers
+    if isinstance(h, logging.handlers.RotatingFileHandler)
+]
+print(json.dumps({{
+    "has_bridge_log_handler": bool(attached),
+    "requested_paths": requested,
+    "actual_paths": [getattr(h, "baseFilename", "") for h in attached],
+    "tmp_dir": tmp_dir,
+}}))
 """
     env = {**os.environ, "PYTHONPATH": str(REPO_ROOT)}
     env.pop("PYTEST_CURRENT_TEST", None)
 
-    # The probe deliberately emulates a real (non-pytest) process, so
-    # bridge.telegram_bridge's module-scope import-time logging (e.g. the
-    # routing-map INFO lines) genuinely writes to the real logs/bridge.log —
-    # that IS the production behavior under test. Snapshot/restore the file
-    # around the subprocess so this probe does not itself leave test-run
-    # artifacts in the operator's production log (the exact class of
-    # pollution #2854 is about).
-    bridge_log = REPO_ROOT / "logs" / "bridge.log"
-    prior_bytes = bridge_log.read_bytes() if bridge_log.exists() else None
-    try:
-        result = subprocess.run(
-            [sys.executable, "-c", probe],
-            cwd=REPO_ROOT,
-            env=env,
-            capture_output=True,
-            text=True,
-            check=False,
-        )
-    finally:
-        if prior_bytes is None:
-            bridge_log.unlink(missing_ok=True)
-        else:
-            bridge_log.write_bytes(prior_bytes)
+    # timeout: this module has an import-time hang precedent (TCC/iCloud
+    # open() deadlock in _get_active_projects, fixed in 261ebbd77). A wedged
+    # probe should fail this test with output, not ride the 420s suite cap.
+    # 180s, not 60: under a pytest-inherited environment the probe import
+    # measures ~65s on a worker-only host (blocks ~60s, then completes), so
+    # a 60s bound fails a healthy probe. Measured 2026-09-05.
+    result = subprocess.run(
+        [sys.executable, "-c", probe],
+        cwd=REPO_ROOT,
+        env=env,
+        capture_output=True,
+        text=True,
+        check=False,
+        timeout=180,
+    )
 
     assert result.returncode == 0, result.stderr
     state = json.loads(result.stdout.strip().splitlines()[-1])
+
+    # Best-effort cleanup of the probe subprocess's redirect directory.
+    shutil.rmtree(state.get("tmp_dir", ""), ignore_errors=True)
+
     assert state["has_bridge_log_handler"] is True, (
         "Production process no longer attaches the logs/bridge.log handler "
         f"— production logging regressed: {state}"
+    )
+    expected = str(REPO_ROOT / "logs" / "bridge.log")
+    assert state["requested_paths"] == [expected], (
+        "Production process requested an unexpected log path "
+        f"(want exactly [{expected!r}]): {state}"
+    )
+    # The redirect held: nothing the probe attached points at the real log.
+    assert all(expected not in p for p in state["actual_paths"]), (
+        f"Probe redirect failed — real production log was opened: {state}"
     )
--- a/tests/unit/test_no_production_log_handlers.py
+++ b/tests/unit/test_no_production_log_handlers.py
@@ -1,0 +1,47 @@
+"""Regression guard for #2854: pytest processes never write production logs.
+
+``bridge/telegram_bridge.py`` attaches a RotatingFileHandler for
+``logs/bridge.log`` to the ROOT logger at module import time, so a single
+test importing it (directly or transitively) used to route every logger's
+records — including test fixtures' deliberate failure-branch ERRORs — into
+the production log file, where they read as daily production failures.
+
+The guard here deterministically imports the known attacher first, so it
+pins the real mechanism regardless of xdist scheduling, then asserts no
+handler on the root or ``bridge`` loggers targets a file under any
+``logs/`` directory. It cannot see attachers that run only in OTHER
+workers or later in this worker's schedule; it exists to keep the
+import-time attach in ``telegram_bridge`` (the one observed leaking) from
+regressing.
+"""
+
+from __future__ import annotations
+
+import logging
+from pathlib import Path
+
+
+def _log_file_handlers(logger: logging.Logger) -> list[str]:
+    return [
+        h.baseFilename
+        for h in logger.handlers
+        if isinstance(h, logging.FileHandler)
+        and "logs" in Path(getattr(h, "baseFilename", "")).parts
+    ]
+
+
+def test_pytest_process_attaches_no_production_log_file_handler():
+    # Import the known import-time attacher so this process has definitely
+    # executed the guarded attach path before we assert.
+    import bridge.telegram_bridge  # noqa: F401
+
+    offenders: dict[str, list[str]] = {}
+    for name in ("", "bridge", "bridge.room_inbox", "bridge.telegram_bridge"):
+        found = _log_file_handlers(logging.getLogger(name))
+        if found:
+            offenders[name or "<root>"] = found
+
+    assert not offenders, (
+        "Production log-file handler active under pytest — test-fixture "
+        f"ERROR lines will masquerade as production failures (#2854): {offenders}"
+    )
--- a/tests/unit/test_no_production_log_handlers.py
+++ b/tests/unit/test_no_production_log_handlers.py
@@ -13,12 +13,29 @@ handler on the root or ``bridge`` loggers targets a file under any
 workers or later in this worker's schedule; it exists to keep the
 import-time attach in ``telegram_bridge`` (the one observed leaking) from
 regressing.
+
+The absence assertion above is only half the guard: it would stay green if
+the production handler were deleted outright or the ``_UNDER_PYTEST``
+condition were inverted. ``test_production_process_still_attaches_the_handler``
+below is the companion probe (precedent:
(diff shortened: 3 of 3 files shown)


valorengels (changes requested):

The fix is correct and the root-cause analysis is right. I independently reproduced every row of the PR's evidence table in this checkout (see Verification Results): the guard fails when `_UNDER_PYTEST` is neutralized, that neutralized run writes exactly 3 `chat=1 msg=1` fixture lines into `logs/bridge.log`, and with the fix active a pytest run never creates the file. I also confirmed `bridge/telegram_bridge.py` is the **only** import-time production-log-handler attacher left in the repo (`config/settings.py::setup_logging`, `worker/__main__._configure_logging`, both watchdogs, and `scripts/evaluate_build.py` all attach inside functions), so this single site really is the leak path.

No blockers. Two tech-debt items and one nit, all cheap: a comment/PR-body claim that is the exact opposite of the shipped behavior, and a one-sided regression guard that would stay green if the handler were deleted outright.

## Rubric

- [x] **1. Plan vs. implementation match** — pass — *No plan document exists for #2854 (`find_plan_path(2854)` returns None), so I judged against the issue's four acceptance criteria. All four are met and independently reproduced.*
- [x] **2. New code quality** — pass — *Guard is a single module-scope boolean; the existing double-import handler check is preserved unchanged inside the new `if`. `sys` and `os` are both imported at module top (lines 22, 25), so the guard cannot NameError.*
- [x] **3. Test coverage** — pass — *`tests/unit/test_no_production_log_handlers.py` covers the pytest side and is a true regression guard (verified failing when neutralized). The missing production-side probe is filed as tech debt below rather than a coverage failure.*
- [x] **4. Regression risk to existing callers** — pass — *Grepped every reference: nothing outside the module reads `telegram_bridge.file_handler`, and every importer of `bridge.telegram_bridge` (`agent_catchup`, `telegram_relay`, `email_bridge`, `answer_routing`, `routing`) imports lazily inside a function. 91 tests across the guard, `test_watchdog_log_isolation`, `test_log_rotate`, `test_structured_logging`, `test_log_format`, and `test_bridge_logic` pass.*
- [x] **5. Data integrity** — n/a — *No models, migrations, or persisted fields touched.*
- [x] **6. Security** — pass — *No secrets, no new subprocess/eval/SQL. The new env read is a membership test, not a value read.*
- [x] **7. Documentation accuracy** — pass — *No existing doc is made stale: `docs/features/build-session-reliability.md:7` and `docs/features/log-rotation.md:25` describe production behavior, which is unchanged. The absence of a note about the new invariant is a nit below, not an inaccuracy.*
- [ ] **8. PR body accuracy** — fail — *"A bridge subprocess spawned BY a test imports without pytest loaded and still attaches: the intended production behavior, and verified" is false. See Tech Debt #1 — I ran the probe.*
- [x] **9. Disclosed deferrals** — n/a — *The PR body contains no out-of-scope, deferral, or follow-up disclosures.*
- [x] **10. Follow-up claims verified** — n/a — *No `filed as follow-up #N` / `tracked by #N` claims in the body.*

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — PASS — *No plan doc; validated against issue #2854's four ACs, all met.*
- **3. New `except Exception` blocks — each has logger/raise/swallow-ok** — N/A — *No new `except` blocks; the pre-existing `except ImportError` fallback is unchanged.*
- **4. New integration tests — exercise serialization boundary (not in-memory only)** — N/A — *Unit-level logging guard; no serialization boundary.*
- **6. No hardcoded secrets or debug artifacts** — PASS — *None; ruff check and `ruff format --check` both clean (1451 files formatted).*
- **7. New public APIs — docstrings present** — N/A — *No new public API. Module-level guard plus one test function, both documented.*
- **8. Breaking changes — migration path documented** — N/A — *Production behavior is byte-identical; only pytest processes change.*
- **9. Tests added for new behavior** — PASS — *`tests/unit/test_no_production_log_handlers.py`.*
- **10. Tests cover the failure path (not just happy path)** — PASS — *Verified by neutralizing the guard: the test fails with the correct offender path.*
- **11. UI changes (if any) — screenshot captured** — N/A — *No HTML/CSS/JS/TSX/template files in the diff (2 files: `bridge/telegram_bridge.py`, `tests/unit/test_no_production_log_handlers.py`).*
- **12. Docs updated for user-facing changes** — PASS — *No user-facing surface changed; see the docs nit.*

### Verified
- [x] Code correctness (reproduced all four issue ACs)
- [x] Test coverage (guard verified failing when neutralized)
- [x] Issue acceptance criteria met

### Blockers
- None

### Tech Debt

- [ ] **`bridge/telegram_bridge.py:604-607`**
  **Code:** `# ... a bridge subprocess spawned BY a / # test imports this module without pytest loaded and still attaches, which is / # the intended production behavior. PYTEST_CURRENT_TEST covers a spawned / # child that inherits the env but re-imports lazily mid-test.`
  **Issue:** These two sentences contradict each other, and the first one is wrong. `PYTEST_CURRENT_TEST` is exported into the process environment by pytest and is therefore **inherited** by any subprocess a test spawns — so a bridge subprocess spawned BY a test does **not** attach the handler. Verified by direct probe in this checkout: with the env scrubbed, root handlers are `[None, '<repo>/logs/bridge.log']`; with `PYTEST_CURRENT_TEST` set, root handlers are `[None]`. The suppression is the right behavior (a test-spawned bridge should not write the operator's production log), but the comment claims the opposite, and the same false sentence is repeated verbatim in the PR body under "Fix (criterion 2)" with "and verified" attached to it.
  **Severity:** tech_debt
  **Fix:** Rewrite the comment to state the actual behavior — "`PYTEST_CURRENT_TEST` is inherited by subprocesses, so a bridge spawned by a test also skips the handler, which is intended; only a process started outside a pytest run attaches it" — and correct the corresponding sentence in the PR body.

- [ ] **`tests/unit/test_no_production_log_handlers.py:33`**
  **Code:** `def test_pytest_process_attaches_no_production_log_file_handler():`
  **Issue:** The guard is one-sided: it only asserts the handler is **absent** under pytest. Deleting the production file handler entirely — or inverting the condition to `if _UNDER_PYTEST:` — leaves this test green while silently killing all production `logs/bridge.log` output, which is the failure mode an operator would notice last. The repo already has the precedent for the missing half: `tests/unit/test_watchdog_log_isolation.py` pairs its absence assertions with `_run_probe()` subprocess probes that assert the production path still configures correctly.
  **Severity:** tech_debt
  **Fix:** Add a companion subprocess probe that imports `bridge.telegram_bridge` with `PYTEST_CURRENT_TEST` stripped from the env (and `PYTHONPATH` pinned to the repo root, as `_run_probe` does) and asserts a `RotatingFileHandler` for `logs/bridge.log` IS on the root logger. I confirmed such a probe works and is fast.

### Nits

- [ ] **`bridge/telegram_bridge.py:609`**
  **Code:** `_UNDER_PYTEST = "pytest" in sys.modules or "PYTEST_CURRENT_TEST" in os.environ`
  **Issue:** This repo already shipped a documented invariant for exactly this bug class — `docs/features/watchdog-log-isolation.md` (#2643): "Importing `monitoring/bridge_watchdog.py`, `monitoring/worker_watchdog.py`, or `scripts/log_rotate.py` has zero logging side effects... configuration happens exactly once, at each script's entry point." This PR solves the same problem in a different way (keep the import-time side effect, sniff for pytest) and documents that divergence nowhere. A future reader hitting the third instance of this bug has two contradictory precedents and no note explaining why the bridge could not use the `__main__`-entry-point pattern.
  **Severity:** nit
  **Fix:** Add a short section to `docs/features/watchdog-log-isolation.md` (or a sibling doc) recording the bridge's import-time attach, why the entry-point pattern was not applied here (the module's backward-compatible re-exports mean many callers import it without running it as `__main__`), and the `_UNDER_PYTEST` guard as the sanctioned alternative.

### Miscellaneous
- None

- None

### Verification Results

No plan document exists for #2854, so there is no `## Verification` table to run. I ran the issue's acceptance criteria directly instead:

- **AC1 (root cause identified)** — CONFIRMED. `bridge/telegram_bridge.py` module scope is the only import-time attacher; every other `FileHandler`/`RotatingFileHandler` construction in the repo (`config/settings.py:1151`, `worker/__main__.py:439`, `monitoring/worker_watchdog.py:166`, `monitoring/bridge_watchdog.py:103`, `scripts/evaluate_build.py:47`) sits inside a function body.
- **AC2 (test runs never append to a production log)** — CONFIRMED. `rm -f logs/bridge.log` then `pytest tests/unit/test_no_production_log_handlers.py tests/unit/test_room_inbox_shadow.py` → 6 passed, file not recreated.
- **AC3 (regression guard)** — CONFIRMED. With `_UNDER_PYTEST = False` patched in, the guard fails with `{'<root>': ['<repo>/logs/bridge.log']}`.
- **AC4 (the `chat=1 msg=1` lines stop)** — CONFIRMED, and the PR's exact number reproduced: the neutralized run (`-n0`, so both files share one process) writes exactly **3** `chat=1 msg=1` lines into `logs/bridge.log`; the same run with the fix active writes none.
- **Lint gates** — `python -m ruff check .` → All checks passed. `python -m ruff format --check .` → 1451 files already formatted.
- **Test slice** — `test_no_production_log_handlers`, `test_watchdog_log_isolation`, `test_log_rotate`, `test_structured_logging`, `test_log_format`, `test_bridge_logic` → **91 passed**. I did not reproduce the PR body's "82 passed" figure because I ran a different file set; it is unverified, not contradicted.
- **Module-scope env guard** — `validate_no_module_scope_env.py` does not flag the new line (`in os.environ` is a membership test, not one of the four tracked calls). The five violations it reports at lines 506-510 are pre-existing and untouched by this diff.

### Screenshots

### Reviewer notes
- Multi-judge consensus (`code-quality` + `risk`) was not run: this reviewer had no subagent dispatch tool available in-turn, and the skill's hard rule forbids launching judges that cannot be awaited. This is a single-judge review; the shape (2 files, 82 additions) is well inside what one pass covers.

valorengels (changes requested):

_Idempotent: prior review on HEAD `fe0fad4` / body hash `aeeed37` is still valid — returning the prior verdict without regenerating findings._

No new commits and no PR-body edits since the previous `## Review:` on this PR, so the previous pass's rubric, checklist, and findings stand unchanged: **0 blockers, 2 tech debt, 1 nit**. See the earlier review for the full rubric, Pre-Verdict Checklist, per-finding blocks, and Verification Results.

Carried-forward findings (details in the prior review):

### Blockers
- None

### Tech Debt
- [ ] **`bridge/telegram_bridge.py:604-607`** — the guard comment (and the same sentence in the PR body) claims a bridge subprocess spawned BY a test still attaches the handler; `PYTEST_CURRENT_TEST` is inherited by subprocesses, so it does not.
- [ ] **`tests/unit/test_no_production_log_handlers.py:33`** — one-sided guard: deleting or inverting the production handler leaves it green. Needs the companion subprocess probe asserting the production path still attaches.

### Nits
- [ ] **`bridge/telegram_bridge.py:609`** — divergence from the `docs/features/watchdog-log-isolation.md` (#2643) entry-point-configuration invariant is undocumented.

### Miscellaneous
- None

- None

### Screenshots

### Reviewer notes
- This pass re-posts because the previous review's verdict was never recorded to the SDLC substrate (`latest_review_verdict: null`), which stalls the router. The verdict is being recorded with this pass.
- Single-judge: the idempotency short-circuit fires before finding generation, so multi-judge consensus was not dispatched.

tomcounsell:
## Review (Judge code-quality): Changes Requested

Reviewed at head `b441bc0` (initial pass at `fe0fad4` was zero-finding; updated after the responsive patch commit landed).

**Summary.** The delta commit is responsive and mostly strengthens the PR. The corrected comment in `bridge/telegram_bridge.py` now states the subprocess semantics accurately and consistently (`PYTEST_CURRENT_TEST` is inherited, so a test-spawned bridge also skips the handler); the earlier comment contradicted itself on exactly this point, so this is a real fix. The new docs section in `docs/features/watchdog-log-isolation.md` is accurate against the code and earns its place: it explains why this module cannot use the entry-point-guard pattern (many importers, no single `__main__`) and gives a clear criterion for choosing between the two patterns. The companion test's rationale is also sound — the absence-only guard would stay green if the handler were deleted or the condition inverted, so pinning the production side is the right instinct, and the docstring says so explicitly.

However, the companion probe's snapshot/restore mechanism introduces a real defect, verified by reading the code rather than taking the parent reviewer's empirical run on faith. The mechanism is deterministic from the source: `prior_bytes = bridge_log.read_bytes()` is taken before `subprocess.run`, and the `finally` block unconditionally executes `bridge_log.write_bytes(prior_bytes)`. Any bytes a concurrent writer appends between snapshot and restore are discarded by construction; no locking or re-read narrows the window. The empirical demonstration (lines 37-62 of a 100ms-interval writer destroyed while the test passed) is exactly what this code must do. Two aggravations also fall straight out of the code: a live `RotatingFileHandler` holds an open fd, so restore-truncation interacts badly with its write offset; and the `prior_bytes is None` branch calls `unlink(missing_ok=True)`, which, if a live writer created the file during the window, orphans that writer's inode so subsequent bridge logging goes to a deleted file until rotation. Since production machines run the bridge in the same checkout where nightly and concurrent-agent pytest runs happen, this test damages the operator's log on every run — the very integrity surface #2854 exists to protect, and directly contrary to the probe's own comment claiming it leaves no artifacts in the production log.

### Blockers
- None

### Tech Debt

- **File:** `tests/unit/test_no_production_log_handlers.py` (`test_production_process_still_attaches_the_handler`, the snapshot/restore around `subprocess.run`)
  **Code:** `prior_bytes = bridge_log.read_bytes() if bridge_log.exists() else None` ... `finally:` `bridge_log.unlink(missing_ok=True)` / `bridge_log.write_bytes(prior_bytes)`
  **Issue:** The restore unconditionally rewrites the real `logs/bridge.log` from a pre-subprocess snapshot, silently destroying anything a concurrent writer (the live bridge sharing the checkout) appended during the window (~2.6s demonstrated). The `unlink` branch can orphan a live writer's inode entirely. A SIGKILL between subprocess and restore also leaves probe lines in the real log. The probe's comment claims it avoids exactly this class of pollution; the implementation does not deliver that.
  **Severity:** tech_debt (bounded diagnostic-log loss, test-only code path, deterministic clean fix available)
  **Fix:** In the probe subprocess, before importing `bridge.telegram_bridge`, subclass or monkeypatch `logging.handlers.RotatingFileHandler` to record the requested filename and redirect the actual open into a tmp directory; then assert both that the handler attached and that the recorded requested path equals `REPO_ROOT/logs/bridge.log`. This gives full predicate coverage with zero reads or writes of the real log and removes the snapshot/restore entirely. While there, replace the weak substring check (`"bridge.log" in baseFilename`, which would also match e.g. `test_bridge.log`) with the exact-path equality the recorded filename enables.

### Nits
- None

*Judge verdict: CHANGES REQUESTED — blockers 0, tech_debt 1, confidence 0.92.*

tomcounsell:
## Review (Judge risk): Changes Requested

Reviewed at head `b441bc0` (initial pass at `fe0fad4` was zero-finding; updated after the responsive patch commit landed). The comment correction and the docs section are accurate (the rewritten `PYTEST_CURRENT_TEST` paragraph now matches the verified behavior, and the docs correctly frame the pattern split against the entry-point-guard modules). The companion probe test is the right idea with the wrong substrate: it operates on the operator's real production log.

Verified in the worktree (not taken on faith):

- **The probe subprocess genuinely writes the real log.** `bridge/telegram_bridge.py` runs `ACTIVE_PROJECTS = _get_active_projects()` and routing-map logging at module scope, and the probe env has `PYTEST_CURRENT_TEST` scrubbed, so the guarded attach fires and import-time INFO lines land in `REPO_ROOT/logs/bridge.log`. The test's own comment says so.
- **Production machines run this test against the live log.** `com.valor.nightly-tests.plist` runs the suite with `WorkingDirectory __PROJECT_DIR__` — the same checkout the launchd bridge serves from. The empirically proven read-run-restore race (lines destroyed while the test passes) therefore recurs on every bridge machine, every nightly run, plus every ad-hoc agent test run.
- **The damage is not bounded by the ~2.6s window in the worst case.** The live handler is mode `a` (O_APPEND), so the restore-truncation cleanly destroys the window's lines. But the probe subprocess attaches its own `RotatingFileHandler` with `maxBytes=10MB` on the same path: when the real file sits near the cap, the probe's emit can trigger `doRollover`, renaming the live bridge's file out from under its open fd and shuffling backups, after which the restore writes ~10MB of stale bytes into the fresh file. That corrupts the rotation chain, and the loss stops being a 2.6s window.
- **The skipped-`finally` residue is real and has an in-module precedent.** `subprocess.run` has no `timeout`, and this exact module's import path has a documented deadlock: commit 261ebbd77 fixed a `get_machine_project_keys()` TCC/iCloud `open()` hang, and `_get_active_projects()` still reads the iCloud-synced projects.json at import. A hung probe rides until pytest-timeout's thread-method forced exit (or a SIGKILLed xdist worker), which skips `finally`: the log is left un-restored with probe routing-map INFO lines in it, reading as a phantom bridge restart. The very pollution class #2854 exists to prevent.

**Why blocker rather than tech_debt:** this is silent, recurring destruction of the primary production diagnostic log — the file the crash tracker, watchdog escalation, and commit-correlation self-healing read — introduced by the PR whose stated purpose is that log's integrity. A 2.6s hole erased during a nightly run can fall exactly on an incident window, and nothing anywhere records that the erasure happened. "Bounded and rare" would argue tech_debt; this is unbounded at the rotation boundary, guaranteed-recurring on every bridge machine, and the fix is cheap and already designed. Production data loss with a cheap known fix is a fix-before-merge item.

### Blockers

- **File:** `tests/unit/test_no_production_log_handlers.py`, `test_production_process_still_attaches_the_handler` (snapshot/restore around the probe subprocess)
  **Code:** `prior_bytes = bridge_log.read_bytes() ... finally: bridge_log.write_bytes(prior_bytes)` with the probe attaching a live `RotatingFileHandler` on `REPO_ROOT/logs/bridge.log`
  **Issue:** The probe writes to and then rewrites the operator's real production log. On bridge machines (nightly suite runs in the live checkout) the restore silently destroys every line the live bridge appended during the probe window (~2.6s, empirically proven); near the 10MB cap the probe's own rollover renames the live file under the running bridge and the restore corrupts the rotation chain; a skipped `finally` (thread-method timeout kill, SIGKILLed worker) leaves probe import-lines in the real log that read as a phantom bridge restart.
  **Severity:** blocker
  **Fix:** Keep the real path out of the probe entirely: in the probe subprocess, before `import bridge.telegram_bridge`, monkeypatch/subclass `logging.handlers.RotatingFileHandler` to record the requested filename and redirect the actual file into a tmp dir; assert the recorded filename equals `REPO_ROOT/logs/bridge.log` and that the handler is attached. No real-log write, no snapshot, no restore, and the skipped-`finally` case becomes harmless.

### Tech Debt

- **File:** `tests/unit/test_no_production_log_handlers.py`, same test
  **Code:** `subprocess.run([sys.executable, "-c", probe], ..., check=False)` with no `timeout=`
  **Issue:** The probe imports a module with a documented import-time hang precedent (`_get_active_projects()` reads iCloud-synced projects.json; commit 261ebbd77 fixed a TCC `open()` deadlock on this path). A wedged probe consumes the full 420s pytest timeout and dies by forced exit instead of failing cleanly. With the blocker fixed the data-loss consequence disappears, but the hang risk stands on its own.
  **Severity:** tech_debt
  **Fix:** Pass `timeout=60` (generous for an import) to `subprocess.run` and let `TimeoutExpired` fail the test with the subprocess's captured output.

### Nits
- None

The predicate change and absence guard remain sound (the `fe0fad4` approval of that mechanism stands); the new companion probe must stop touching the real production log before this merges.

*Judge verdict: CHANGES REQUESTED — blockers 1, tech_debt 1, confidence 0.90.*

tomcounsell:
## Review: Changes Requested

Two-judge consensus review of PR #3144 at head `b441bc0` (Closes #2854). The core fix is correct and everything the prior review at `fe0fad4` flagged has been resolved. One new defect arrived with the patch commit itself: the companion production-attach probe operates on the operator's real `logs/bridge.log`, and its snapshot/restore verifiably destroys concurrent live-bridge writes. Consensus rule `any-blocker-wins`, k=2, tied=false, mean confidence 0.91 — aggregate CHANGES REQUESTED with 1 blocker and 1 tech_debt (both judges flagged the same underlying defect; code-quality classified it tech_debt, risk escalated to blocker on rotation-boundary and nightly-checkout evidence, and max wins).

## Rubric

- [x] **1. Plan vs. implementation match** — pass — *no plan doc (no-plan lane); judged against issue #2854's four acceptance criteria: root cause identified (module-import attach in `telegram_bridge.py`), isolation at the source, regression guard added (both directions), leak mechanism proven stopped (criterion 4 is post-merge observational; mutation-verified here).*
- [ ] **2. New code quality** — fail — *the new companion probe reads, rewrites, and can rotate the real production log from inside a test run; see Blocker 1. The predicate change itself is clean and byte-preserves the production path.*
- [x] **3. Test coverage** — pass — *absence guard mutation-verified (`_UNDER_PYTEST = False` → red + `logs/bridge.log` gains fixture lines); companion probe covers the inverse direction (handler deleted/condition inverted → red).*
- [x] **4. Regression risk to existing callers** — pass — *launchd/plist spawn path carries no `PYTEST_CURRENT_TEST`; `'pytest' in sys.modules` is False after a full production import; no code or test depends on the handler existing under pytest.*
- [x] **6. Security** — pass — *no new subprocess-with-untrusted-input, eval, or secret surface; the probe subprocess runs a fixed inline script under `sys.executable`.*
- [x] **7. Documentation accuracy** — pass — *new `watchdog-log-isolation.md` section accurately documents the `_UNDER_PYTEST` pattern vs the entry-point pattern and when each applies.*
- [x] **8. PR body accuracy** — pass — *evidence table reproduced (fix active → file never created; guard neutralized → red + fixture lines; production probe → attached, correct path). The previously false subprocess sentence was corrected in both the code comment and the PR body; the corrected claim matches my direct probe (env-inheriting child skips; clean-env process attaches).*
- [x] **9. Disclosed deferrals** — n/a — *no disclosures in the PR body.*
- [x] **10. Follow-up claims verified** — n/a — *no follow-up claims.*

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — PASS — *issue #2854 criteria; see Rubric 1.*
- **2. No-Gos from plan — none violated** — N/A — *no plan document.*
- **4. New integration tests — exercise serialization boundary** — PASS — *the companion probe crosses a real process boundary (subprocess import), which is the boundary under test.*
- **5. Plan internal consistency** — N/A — *no plan document.*
- **6. No hardcoded secrets or debug artifacts** — PASS — *none in the diff; ruff check and format clean at `b441bc0`.*
- **7. New public APIs — docstrings present** — PASS — *both test functions and the module carry docstrings; no production API added.*
- **8. Breaking changes — migration path documented** — N/A — *production behavior byte-identical.*
- **9. Tests added for new behavior** — PASS — *two-directional guard in `tests/unit/test_no_production_log_handlers.py`.*
- **10. Tests cover the failure path (not just happy path)** — PASS — *both directions mutation-verifiable; absence direction empirically mutation-tested.*
- **12. Docs updated for user-facing changes** — PASS — *pattern divergence documented in `docs/features/watchdog-log-isolation.md`.*

### Blockers

- [ ] **File:** `tests/unit/test_no_production_log_handlers.py` (`test_production_process_still_attaches_the_handler`, snapshot/restore around the probe subprocess)
  **Code:** `prior_bytes = bridge_log.read_bytes() if bridge_log.exists() else None` ... `finally:` `bridge_log.unlink(missing_ok=True)` / `bridge_log.write_bytes(prior_bytes)`
  **Issue:** The probe imports `bridge.telegram_bridge` with a scrubbed env, so the guarded attach fires and import-time INFO lines land in the REAL `REPO_ROOT/logs/bridge.log`; the test then rewrites that file from a pre-subprocess snapshot. Empirically proven here: with a concurrent writer appending 150 numbered lines at 100ms intervals, lines 37–62 (~2.6s of writes) were silently destroyed while the test passed. Production machines run the nightly suite (`com.valor.nightly-tests.plist`, `WorkingDirectory __PROJECT_DIR__`) and ad-hoc agent test runs in the SAME checkout the live bridge serves from, so every run of this test truncates whatever the bridge logged during the probe window — the primary diagnostic log the crash tracker and watchdog escalation read, and the exact integrity surface #2854 protects. Aggravations: near the 10MB cap the probe's own handler can trigger `doRollover`, renaming the live bridge's file under its open fd and corrupting the rotation chain (loss then exceeds the window); the `prior_bytes is None` branch `unlink`s a file a live writer may have created, orphaning its inode; and a skipped `finally` (pytest-timeout thread-method kill, SIGKILLed xdist worker — with a documented import-hang precedent on this exact module, commit 261ebbd77) leaves probe import-lines in the real log reading as a phantom bridge restart.
  **Severity:** blocker (risk judge; code-quality judge classified the same defect tech_debt — max wins under any-blocker-wins)
  **Fix:** Keep the real path out of the probe entirely. In the probe subprocess, BEFORE `import bridge.telegram_bridge`, subclass/monkeypatch `logging.handlers.RotatingFileHandler` to record the requested filename and redirect the actual open into a tmp directory; assert (a) the handler attached and (b) the recorded requested path equals `REPO_ROOT/logs/bridge.log` (exact-path equality, replacing the weak `"bridge.log" in baseFilename` substring check). No real-log read or write, no snapshot/restore, and the skipped-`finally` case becomes harmless.

### Tech Debt

- [ ] **File:** `tests/unit/test_no_production_log_handlers.py`, same test
  **Code:** `subprocess.run([sys.executable, "-c", probe], ..., check=False)` with no `timeout=`
  **Issue:** The probe imports a module with a documented import-time hang precedent (`_get_active_projects()` reads iCloud-synced projects.json at module scope; commit 261ebbd77 fixed a TCC `open()` deadlock on this path). A wedged probe rides the full 420s pytest timeout and dies by forced exit instead of failing cleanly with output.
  **Severity:** tech_debt
  **Fix:** Pass `timeout=60` to `subprocess.run` and let `TimeoutExpired` fail the test with the captured output.

### Nits
- None

### Miscellaneous
- None

- None

### Legacy Cruft
- None — no dual paths, dead imports, or stale comments introduced.

### Review Delta (vs prior review on HEAD `fe0fad4`)

- **Resolved:** contradictory `PYTEST_CURRENT_TEST` comment sentence — rewritten accurately in `bridge/telegram_bridge.py` and in the PR body; verified against direct probes (env-inheriting child skips attach; clean-env process attaches with the correct path).
- **Resolved:** one-sided regression guard — companion production-attach probe added (right idea; substrate is the new blocker).
- **Resolved:** missing pattern-divergence documentation — `docs/features/watchdog-log-isolation.md` section added and accurate.
- **New:** snapshot/restore race destroying concurrent production log writes (Blocker 1).
- **New:** missing probe subprocess timeout (Tech Debt 1).

### Verification Results

No plan `## Verification` table exists (no-plan lane); direct probes on the PR head in an isolated worktree:

- Both guard tests pass at `b441bc0` (`-n auto` and `-n0`); a plain run leaves `logs/` untouched.
- Mutation (at `fe0fad4`, mechanism unchanged): `_UNDER_PYTEST = False` → absence guard FAILS and `logs/bridge.log` appears with fixture lines.
- Sanity slice (guard + `test_room_inbox_shadow` + `test_log_format` + `test_log_rotate` + `test_structured_logging`): 30 passed; the exact issue-signature fixtures (`chat=1 msg=1`, `redis down`) no longer reach the production log.
- Production probe (clean env): handler attaches with the correct path; `'pytest' in sys.modules` is False after full import — no misfire vector.
- Child probe (`PYTEST_CURRENT_TEST` inherited): attach skipped, no file created.
- Race proof (drives Blocker 1): concurrent writer during `test_production_process_still_attaches_the_handler` → 124 of 150 lines survive, gap spans lines 37–62, test green.
- Gates: `ruff check` and `ruff format --check` clean on all changed files.
- Attach-path sweep: no other module-scope `logs/` file-handler attach remains (`worker/__main__.py` attaches only inside `main()`; `config/settings.py::setup_logging` is explicit-call-only via `validate_configuration()`, which no test calls).

### Screenshots
- None (no UI changes).

### Deploy note
This PR modifies `bridge/`; a service restart (`./scripts/valor-service.sh restart`) is required on all machines after deploy.

### Consensus
- judges_run: 2 — code-quality: CHANGES REQUESTED, 0 blockers / 1 tech_debt (0.92); risk: CHANGES REQUESTED, 1 blocker / 1 tech_debt (0.90)
- rule: any-blocker-wins; tied: false; mean_confidence: 0.91; blocker_aggregation: max
- Both judges flagged the same underlying snapshot/restore defect; deduplicated above as one blocker. Cross-vendor judge: not enabled (`SDLC_REVIEW_CROSS_VENDOR` unset).

tomcounsell:
## Review: Approved

Re-review of PR #3144 (Closes #2854) at head `2045157`. Prior consensus at `b441bc0` was CHANGES REQUESTED with one blocker and one tech-debt item; the patch commit `2045157` changes exactly one file (`tests/unit/test_no_production_log_handlers.py`, +68/-33). The core fix in `bridge/telegram_bridge.py` and the docs are byte-identical to the previously reviewed state.

### Prior findings — both resolved, verified in this checkout

**Blocker (probe snapshot/restore destroyed concurrent writes): RESOLVED.** The snapshot/restore is gone. The probe now patches `RotatingFileHandler.__init__` at class level *inside its own subprocess, before the module import*, recording the requested filename and redirecting the actual open into a private tmp dir. The assertion is exact-path equality on the *requested* path while the bytes land in tmp. Reproduced the zero-lost-lines proof stronger than the PR claims: a background writer appended lines to this checkout's `logs/bridge.log` for the full duration of the test run — **879/879 lines survived, 0 lost**, the pre-existing seed line survived, and the file's inode was unchanged (no truncate, no delete/recreate, no restore-overwrite). Mutation check: neutralizing the redirect (`_real_init(self, str(filename), ...)`) makes the test FAIL with exactly `Probe redirect failed — real production log was opened`, so the redirect-held assertion bites and is not vacuous.

**Leak audit on the class patch:** the monkey-patch is applied only inside the `python -c` probe subprocess, which exits after emitting its JSON. The pytest process never patches anything, so no teardown restore is needed — the process boundary is the teardown. It cannot leak into other tests.

**Tech debt (no subprocess timeout): RESOLVED.** `timeout=180` is present, with the deviation from the suggested 60s justified in an in-code comment (measured ~65s probe import under a pytest-inherited environment on a worker-only host, dated 2026-09-05, citing the 261ebbd77 TCC/iCloud hang precedent). The measurement holds here: my clean run of the two tests took 71s wall (the absence guard is sub-second, so the probe alone ran ~65s), confirming a 60s bound would fail a healthy probe. 180s gives ~2.7x headroom while still failing far below the 420s suite cap.

### Verification results (independently reproduced)

- `tests/unit/test_no_production_log_handlers.py` — **2 passed** in 71s via `scripts/pytest-clean.sh` in a detached worktree at `2045157`
- Concurrent-writer proof — 879/879 survived, 0 lost, inode stable
- Redirect mutation — FAILS as designed, restored
- `ruff check .` — clean; `ruff format --check .` — 1451 files already formatted
- Preflight — OPEN / MERGEABLE / CLEAN

### Blockers
- None

### Tech Debt
- None

### Nits
- None

### Notes
- PR modifies `bridge/telegram_bridge.py`: requires a bridge service restart on all machines after deploy (`./scripts/valor-service.sh restart`).
- No plan document exists for #2854 (issue-derived fix); reviewed against the issue's acceptance criteria and the PR body, all previously validated in the earlier rounds.

Verdict: **APPROVED** — 0 blockers, 0 tech debt, 0 nits.

Issue #1228, opened 2026-04-30
PM sessions with distinct slugs serialize on project_key, blocking sibling SDLC parallelism


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/sdlc-1228.md

> **Context:** This system runs an SDLC pipeline (Plan → Critique → Build → Test → Patch → Review → Patch → Docs → Merge) where a parent **PM (Project Manager) session** orchestrates dev work. When the parent has multiple independent issues to drive (e.g., 3 separate PRs to review-and-merge), it spawns **sibling PM children**, each with its own slug, branch, and `.worktrees/{slug}/` filesystem worktree. See [bridge-worker-architecture.md](docs/features/bridge-worker-architecture.md) and [pm-dev-session-architecture.md](docs/features/pm-dev-session-architecture.md).

## Problem

PM sessions are routed to a single per-project worker loop and execute serially even when they have unique slugs and isolated worktrees. The infrastructure for parallel execution (worktrees, branch isolation, global concurrency cap) already exists, but the routing layer ignores PM session slugs.

**Current behavior:**

Today (2026-04-30), a parent PM spawned three sibling children to drive PRs #1221, #1211, #1224 — each with a distinct slug (`sdlc-1215`, `sdlc-1206`, `sdlc-1214`) and worktree:

```
0_1777540261927  completed   pm   driving PR #1221  (slug: sdlc-1215)
0_1777540271145  completed   pm   driving PR #1211  (slug: sdlc-1206)
0_1777540282099  running     pm   driving PR #1224  (slug: sdlc-1214)
```

The three sessions were created within 21 seconds of each other but executed strictly sequentially:
- Session 1 (sdlc-1215): ~9 min runtime
- Session 2 (sdlc-1206): ~17 min runtime
- Session 3 (sdlc-1214): only started after sessions 1 and 2 completed

Total wall time: ~46 min for work that, with parallelism on the existing 8-slot global concurrency cap, could complete in ~max(child_runtime).

Root cause confirmed at `models/agent_session.py:395-396`:

```python
if self.session_type == SessionType.PM:
    return self.project_key  # PM ALWAYS uses project_key, slug is ignored
```

[...]

PR #3145, opened 2026-09-04
Unmask the fixture save-stub TypeError and close out the #2949 constructor-kwarg audit

### LINKED ISSUE #3001: Unblock dependency bumps (anthropic/pydantic-ai), restore the dead worker_key regression guard, and stop duplicate nightly triage filing
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/unblock-dependency-bumps-coupled-set-gate.md

> **Context:** This repo makes all non-harness LLM calls through [PydanticAI](https://ai.pydantic.dev/) wrapped by [`agent/llm/wrapper.py`](agent/llm/wrapper.py)'s `run_typed`, which talks to the Anthropic API via the [`anthropic`](https://github.com/anthropics/anthropic-sdk-python) SDK. A [nightly regression detector](docs/features/nightly-regression-tests.md) (`scripts/nightly_regression_tests.py`, launchd job `com.valor.nightly-tests`, 03:00 local) runs the full suite, re-confirms failures serially, and dispatches an Eng "triage" session that files GitHub issues for newly-confirmed failures.
>
> On 2026-08-24 those two systems collided: a routine dependency bump broke every LLM call in the system, and the triage session filed **39 issues** for what were really **two** root causes. This issue consolidates and replaces all of them.

## Problem

A routine `/update` dependency auto-bump silently broke every non-harness LLM call in the system, and the failure surfaced six hours later as 66 confusing nightly test failures and a 39-issue tracker flood. Three separate weaknesses combined to let that happen, and all three are still open.

**Current behavior:**

1. **Dependency bumps are unsafe and we are pinned to an old, incompatible pair.** Commit `9d1488ccb` auto-bumped `anthropic` 0.125.0 → 1.0.0. anthropic 1.0.0 **removed** `temperature`/`top_p`/`top_k` from the Messages API, but `pydantic-ai-slim==2.9.0` passes all three **unconditionally** — as the `OMIT` sentinel, which still binds the keyword. Every `run_typed` call therefore raised `TypeError: AsyncMessages.create() got an unexpected keyword argument 'temperature'` at argument binding, before any network I/O. Nothing caught this at bump time. The structural hole: pydantic-ai's `anthropic` extra declares `anthropic>=0.108.0` with **no upper bound**, so the bump resolved cleanly and installed a combination that cannot work.

2. **The #1228 sibling-parallelism regression guard is silently dead.** Nine tests in `tests/unit/test_agent_session.py` fail (`TestWorkerKeyProperty` ×8, `TestWorkerKeyTruthTable::test_property_stage_conditional_for_eng`). These are **not** the dependency break — they still fail with it fixed. PR #2949 (commit `69dc69568`) converted `stage_states` from a Popoto `Field` into a property backed by `session_events`. Popoto's constructor silently drops non-Field kwargs, so the test helper `_make_session` ([`tests/unit/test_agent_session.py:143`](tests/unit/test_agent_session.py#L143)) passing `stage_states=json.dumps(...)` became a no-op → `current_stage` is always `None` → `worker_key` falls back to `project_key`.

3. **Triage issue-filing has no idempotency or root-cause collapsing.** The detector script itself behaved correctly — it dispatched **one** triage session for the 10 newly-confirmed nodes and suppressed 56 already-filed ones. The flood came from the triage Eng session (`0_1787603653699`), which filed one issue per node and did so **three times over**, in waves at 20:35, 20:40 and 20:46 UTC.

**Desired outcome:**

- The dependency stack can be bumped to latest, and a bump that breaks the LLM layer fails **at bump time** with a clear signal, not six hours later as a wall of unrelated-looking test failures.
- The worker-key routing invariant from #1228 is actually exercised by CI again, and the class of silent constructor-kwarg drops introduced by #2949 is audited.
- One root cause produces one issue.

## Definitions

| `run_typed` | The wrapper all non-harness LLM calls go through; runs a PydanticAI agent and validates the response against a schema. | [`agent/llm/wrapper.py`](agent/llm/wrapper.py) |
| `worker_key` | Computed property deciding which worker loop a session serializes on. Slugged Eng sessions at worktree stages route by slug so siblings run concurrently; everything else serializes on `project_key`. | [`models/agent_session.py:669`](models/agent_session.py#L669) |
| `stage_states` | JSON map of SDLC stage → status. Since #2949 it is a **property** over `session_events`, no longer a stored Popoto Field. | [`models/agent_session.py:1613`](models/agent_session.py#L1613) |
| Popoto | The Redis ORM backing `AgentSession`. Its constructor accepts declared `Field`s and silently ignores unknown kwargs. | [popoto](https://github.com/tomcounsell/popoto) |
| Triage session | An Eng `AgentSession` the nightly detector spawns to investigate and file issues for newly-confirmed failures. | [`docs/features/nightly-alert-triage.md`](docs/features/nightly-alert-triage.md) |

## Prior Context

- **#2932** — "Nightly regression detector files environmental failures and re-baseline umbrellas as code regressions". **Folded into this issue and closed**; its scope (environmental-vs-code classification, idempotent re-baseline umbrellas, root-cause collapsing) is carried into Work Item 3 below.
- **#2960–#2999** (39 issues) — the auto-filed flood. All closed as duplicates of this issue. Note their bodies **misattribute** the nine worker-key failures to the shared anthropic root cause; that diagnosis is wrong (see Work Item 2).
- **#2949** (commit `69dc69568`) — removed deprecated `AgentSession` aliases and legacy kwarg mapping; the `stage_states` property conversion that broke the test helper.
- **#1228** — established sibling Eng sessions routing to distinct worker loops; the invariant whose guard is now dead.
- **#2334** — older (July) "attempt autonomous fix before paging a human" enhancement. **Deliberately left open** as separate scope; it is a feature request, not part of this cleanup.

## Solution Sketch

**Work Item 1 — unblock dependency bumps (the headline ask).**
Commit `7db5b82bb` already reverted to `anthropic==0.125.0` to stop the bleeding. That is a **stopgap**, and it pins us to a stack 25 minor versions behind (`pydantic-ai-slim` 2.9.0 vs 2.34.0 latest).

The forward path is **verified working**, not theoretical — installed in a throwaway venv, `anthropic==1.0.0` + `pydantic-ai-slim[anthropic]==2.34.0` no longer passes the sampling params unconditionally, and a real `Agent('anthropic:claude-haiku-4-5-20251001').run()` call returned successfully.

- Upgrade `pydantic-ai-slim` to 2.34.0 and restore `anthropic` to 1.0.0+.
- Re-audit the other `CRITICAL — pin exact` deps in [`pyproject.toml`](pyproject.toml) for the same staleness.
- Add a **compatibility smoke gate** to `/update`'s dep-bump step: one real LLM call after sync. A bump that breaks argument binding must fail loudly at bump time. This is the piece that actually prevents recurrence — the upper-bound problem lives in a transitive dependency's metadata, which we do not control.

**Work Item 2 — restore the dead worker-key guard.**
Migrate `_make_session` off the dropped constructor kwarg onto the setter path (`session.stage_states = payload`), which is what production uses and which still works correctly.

Then **audit for other silent constructor-kwarg drops** from the same #2949 property conversion. This is the part worth real attention: a kwarg that is silently ignored rather than rejected fails quietly everywhere, and the worker-key tests are simply where it happened to surface. Consider whether `AgentSession.__init__` should **reject** unknown kwargs outright rather than dropping them.

**Work Item 3 — stop the duplicate/noise filing.**
- Make triage issue-filing idempotent — the same node must not produce a second issue on a re-run.
- Collapse per-node failures sharing a root-cause signature into one issue (39 issues for 2 causes).
- From #2932: classify environmental failures (TLS/DNS/timeout/connection-refused) as `environmental` and do not file them as code regressions; make re-baseline umbrella filing idempotent, keyed on (collection signature, head commit).

## Recon Summary

**Confirmed:** 5 items — verified directly, ready to include
- anthropic 1.0.0 drops `temperature`: confirmed by `inspect.signature(AsyncMessages.create)` — parameter absent.
- The break was live on `main`: reproduced before the revert, gone after.
- The revert fixes the LLM path: `tests/unit/test_routing.py::TestNeedsResponseLlmClassification` went 10/10 passed having been in the failure set.
- The upgrade path works: live API call succeeded on anthropic 1.0.0 + pydantic-ai-slim 2.34.0.
- `claude-agent-sdk` does not depend on `anthropic` at all (checked its PyPI metadata), so neither the revert nor the re-upgrade is constrained by it.

**Revised:** 2 items — scope corrected against the auto-filed issues
- The nine worker-key failures are a **separate** regression, not the anthropic break. The auto-filed issues assert otherwise and are wrong.
- **Production is unaffected** by Work Item 2. Verified that the setter path (used by [`agent/pipeline_state.py:494`](agent/pipeline_state.py#L494)) and the dict-handling helper both still yield the correct `worker_key`. The severity is a **silently untested invariant**, not broken routing — this should not be planned as a production incident.

- "The nightly detector is filing noise" as a charge against `scripts/nightly_regression_tests.py`: the script's suppression logic worked correctly. The defect is in the triage session's filing behavior. Filed against the wrong component in #2932.
- #2334's autonomous-fix-before-alert enhancement: distinct scope, left open.

## Acceptance Criteria

- [ ] `pyproject.toml` pins `pydantic-ai-slim[anthropic]==2.34.0` (or later) and `anthropic==1.0.0` (or later), with `uv sync --all-extras` resolving cleanly.
- [ ] A real `run_typed` call succeeds against the upgraded stack (live call, not a mock).
- [ ] `/update`'s dep-bump step runs a post-sync LLM compatibility smoke check that fails the step on an incompatible pair; verified by deliberately pinning the known-bad combination and observing the failure.
- [ ] The other `CRITICAL — pin exact` deps are audited for staleness, with findings recorded.
- [ ] All 9 worker-key tests pass, and they **fail** if `worker_key`'s slug-precedence logic is reverted (i.e. the guard is genuinely live again, not just green).
- [ ] Other silent constructor-kwarg drops from #2949's property conversions are audited; a decision is recorded on whether `AgentSession.__init__` should reject unknown kwargs.
- [ ] Re-running triage over an already-filed failure set files **zero** new issues.
- [ ] Failures sharing a root-cause signature collapse into one issue.
- [ ] Environmental failures (TLS/DNS/timeout/connection-refused) are classified as such and do not file code-regression issues.

## Downstream

Constraints for the planner:

- **Work Item 1 is the priority** and is independently shippable. Items 2 and 3 should not block the dependency upgrade.
- The current `anthropic==0.125.0` pin is a **stopgap from `7db5b82bb`**, not a decision to stay on 0.x. Do not treat it as the desired end state.
- Work Item 2 is **not** a production incident — plan it as restoring test coverage plus an audit, not as an outage fix.
- Per the repo's testing philosophy, the compatibility smoke gate must make a **real** LLM call. A mocked check cannot catch an argument-binding break, which is the exact failure mode here.

### LINKED ISSUE #3074: Work Item 2 residue: unmask the append_event TypeError in the worker_key fixture, audit #2949's silent constructor-kwarg drops, decide on rejecting unknown kwargs
> **Context:** `AgentSession` is this repo's Redis-backed session model, stored via [Popoto](https://github.com/tomcounsell/popoto), an ORM whose `Model.__init__` accepts declared `Field`s and **silently ignores unknown kwargs**. PR #2949 (`69dc69568`) converted `stage_states` from a stored Popoto `Field` into a property backed by `session_events`. Test fixtures that passed `stage_states=` to the constructor therefore became silent no-ops, and the #1228 worker-key routing guard went green-but-dead without anyone noticing. This is **Work Item 2** of [#3001](https://github.com/tomcounsell/ai/issues/3001), split out per the owner's confirmation of Open Question 4 in [`docs/plans/unblock-dependency-bumps-coupled-set-gate.md`](docs/plans/unblock-dependency-bumps-coupled-set-gate.md).

## Scope correction, up front

**The headline half of Work Item 2 has already landed.** Commit [`56e80d843`](https://github.com/tomcounsell/ai/commit/56e80d843) ("fix(tests): set stage_states through the setter, not a constructor kwarg", 2026-08-31, `Refs #2852`) migrated `_make_session` onto the setter path and added a fixture self-check. Verified on `origin/main` at `b87fb26de`:

```
$ ./scripts/pytest-clean.sh tests/unit/test_agent_session.py -k "TestWorkerKeyProperty or TestWorkerKeyTruthTable" -q
25 passed in 9.34s
```

That commit landed under a *different* umbrella (#2852, now closed) and was never recorded against #3001, which is why #3001's plan still lists Work Item 2 as outstanding. **This issue is therefore the residue**, not the original ask, and it is smaller than #3001 describes. Do not re-do the fixture migration.

## Problem

**Current behavior:**

Three things the fixture fix did not address remain live on `main`.

**1. The fixture's `save` stub still has the wrong signature, masking a `TypeError` on every construction.** `_make_session` (`tests/unit/test_agent_session.py:154`) installs `AgentSession.save = lambda self: None`. The `stage_states` setter routes through `append_event`, which calls `self.save(update_fields=[...])`. Reproduced against `origin/main`:

```
current_stage= BUILD worker_key= my-feature
WARNING:models.agent_session:append_event save failed for session dbg-sid-1
  (event_type='stage'): <lambda>() got an unexpected keyword argument 'update_fields'
```

The tests pass — the in-memory event still lands — but every stage-setting construction emits a **swallowed `TypeError`**. A restored guard that logs a masked exception on every use is exactly the class of quiet failure this work item exists to end. Fix: widen the stub to `lambda self, **kwargs: None`.

**2. The #2949 audit was never performed.** `stage_states` is the property conversion that happened to surface. `Model.__init__` silently dropping unknown kwargs is a general hazard, and any other call site still passing a now-converted property as a constructor kwarg is silently no-op'ing today with no test to notice.

**3. No decision is recorded on whether `AgentSession.__init__` should reject unknown kwargs.** #3001 names this as the durable fix — "a kwarg that is silently ignored rather than rejected fails quietly everywhere" — and it is still an open question.

**Desired outcome:**

The restored guard runs clean with no masked exception; every other silent constructor-kwarg drop introduced by #2949's property conversions is found or proven absent; and the repo has a recorded decision on rejecting unknown kwargs at construction, so the next property conversion cannot repeat this silently.

## Definitions

| `worker_key` | Computed property deciding which worker loop a session serializes on. Slugged Eng sessions at worktree stages route by slug so siblings run concurrently; everything else serializes on `project_key`. | [`models/agent_session.py:658-705`](models/agent_session.py#L658-L705) |
| `stage_states` | JSON map of SDLC stage → status. Since #2949 it is a **property** over `session_events`, no longer a stored Popoto Field. | [`models/agent_session.py:1613`](models/agent_session.py#L1613) |
| Popoto | The Redis ORM backing `AgentSession`. Its constructor accepts declared `Field`s and silently ignores unknown kwargs. | [popoto](https://github.com/tomcounsell/popoto) |
| `append_event` | Appends a `SessionEvent` to `session_events` and persists with `save(update_fields=["session_events", "updated_at"])`. Its save failure is caught and logged, never raised. | `models/agent_session.py` |

## Prior Context

- **#1228** — established that sibling Eng sessions route to distinct worker loops; the invariant whose guard went dead.
- **#2949** (`69dc69568`) — the property conversion. Note its removal of the legacy kwarg mapping is **pinned by a test**: `tests/unit/test_agent_session_legacy_surface.py::TestNormalizeKwargsLegacyPathsRemoved::test_legacy_key_does_not_crash_normalization` parametrizes `"stage_states"` among keys that must survive normalization untouched. **Restoring the kwarg mapping is not an available fix** — it would turn that pin red. The motivation is in that module's docstring: the sibling `commit_sha`/`summary` paths were unguarded and grew `session_events` without bound on every hydration.
- **`56e80d843`** — the fixture migration, already merged (see Scope correction).
- **#2971, #2982–#2989** — nine auto-filed duplicates for the nine failing nodes, all closed as duplicates of #3001. Their bodies **misattribute** the failures to the anthropic dependency break; that diagnosis is wrong.
- **#3001** — parent. Its plan's Verification table carries an anti-criterion that its own lane leaves `models/agent_session.py` and `tests/unit/test_agent_session.py` untouched, reserving them for this issue.

## Solution Sketch

- Widen the `_make_session` `save` stub to accept `**kwargs`, and assert no `append_event save failed` warning is emitted during fixture construction — otherwise the same masking returns the next time the stub drifts from the real signature.
- Audit every `AgentSession(...)` construction site (tests and production) for kwargs that are no longer declared `Field`s after #2949's property conversions. Report the finding either way; "audited, none found" is a valid and useful outcome, but it must be stated rather than assumed.
- Decide and record whether `AgentSession.__init__` should reject unknown kwargs. This is a genuine open design question with a real cost on both sides — rejecting is the durable fix, but Popoto's tolerance may be load-bearing for hydration paths, and the #2949 docstring shows unguarded kwarg handling has already caused unbounded `session_events` growth once. `/do-plan` should treat this as a question to investigate, not a direction to execute.

**Worth checking during the audit** (raised in #3001's comments, not yet verified by this filing): the two negative-control tests `test_eng_session_with_slug_at_plan_stage_uses_project_key` and `test_eng_session_with_slug_at_issue_stage_uses_project_key` were passing *vacuously* before `56e80d843` — they assert `worker_key == "test-project"` and got it from `current_stage=None` rather than from the intended PLAN/ISSUE fallback. The new fixture self-check plausibly fixed this as a side effect; confirm rather than assume.

## Recon Summary

**Confirmed:** 3 items — verified directly against `origin/main` at `b87fb26de`
- The 25 worker-key tests pass; the fixture migration is merged.
- The `save`-stub `TypeError` is still emitted and swallowed on every stage-setting construction (reproduced, output above).
- `tests/unit/test_agent_session_legacy_surface.py` pins the removal of the kwarg mapping, foreclosing the shorter-looking fix.

**Revised:** 1 item — scope corrected
- **This is no longer "restore the dead guard".** The guard is restored. Scope is the masked `TypeError`, the audit, and the design decision.

- **"All 9 worker-key tests pass"** as an acceptance criterion — already true, and the count was always an undercount (25 nodes in the two classes).
- **The `SDLC_STAGES` / `PATCH` omission** found during the same triage (`models/agent_session.py:81` omits `PATCH` while `agent/pipeline_state.py:58` includes it, making the `PATCH` entries in `_ENG_WORKTREE_STAGES` and `resolve_branch_for_stage` unreachable). It was filed separately as a live routing defect and should not be re-scoped here.

## Acceptance Criteria

- [ ] `_make_session`'s `save` stub accepts `**kwargs`, and a test asserts no `append_event save failed` warning is logged during fixture construction.
- [ ] The two worker-key test classes still pass (25 nodes), and they still **fail** if `worker_key`'s slug-precedence logic is reverted — i.e. the guard is live, not merely green.
- [ ] Every `AgentSession(...)` construction site is audited for kwargs that are no longer declared `Field`s post-#2949; findings recorded in the PR description, including an explicit "none found" if that is the result.
- [ ] A decision on whether `AgentSession.__init__` should reject unknown kwargs is **recorded** (in the plan doc and, if the answer is yes, implemented with tests). A recorded "no, and here is why" closes this criterion.
- [ ] The PLAN/ISSUE negative controls are confirmed non-vacuous, or made so.
- [ ] `tests/unit/test_agent_session_legacy_surface.py` still passes — the kwarg mapping stays removed.
- [ ] Tests pass (`/do-test`); docs updated if the kwarg-rejection decision changes model behavior (`/do-docs`).

## Downstream

Constraints for the planner:

- **This is not a production incident.** #3001 verified that the only non-test `stage_states=` call sites (`agent/session_runner/runner.py:1459,1479`) are arguments to `evaluate_completion`, not the `AgentSession` constructor, and `create_local` handles a `stage_states` dict explicitly. Plan this as test-coverage hygiene plus an audit plus a design decision.
- **Do not restore #2949's kwarg mapping.** It is pinned by a test, for a reason recorded in that test module's docstring.
- Stay out of `pyproject.toml`, `scripts/update/`, and `agent/llm/` — those belong to #3001 and its Step 2 lane.

tomcounsell:
Closes #3074.

Work Item 2 residue from #3001, per the issue's own scope correction: the fixture migration (56e80d843) is untouched; this PR delivers the three residue items.

## 1. The masked TypeError is unmasked and pinned

`_make_session`'s stub was `lambda self: None`, but `append_event` calls `self.save(update_fields=[...])` — every stage-setting construction raised a TypeError that `append_event` caught and logged. All five save stubs in `tests/unit/test_agent_session.py` (four `mock_save` defs plus the fixture lambda) now accept `save`'s real shape (`self, *args, **kwargs`).

New `TestFixtureSaveStubFidelity::test_fixture_construction_emits_no_masked_save_failure` asserts no `append_event save failed` warning is logged during fixture construction, so the masking cannot silently return when the stub next drifts.

**Mutation evidence (item 1):** with the narrow stub temporarily restored, the new test fails naming the masked failure (`1 failed, 1 passed`); restored, `2 passed`.

**Guard liveness (acceptance criterion 2):** with `worker_key`'s slug-precedence branch temporarily neutralized (`if False and ...`), the two worker-key classes report `9 failed, 16 passed`; restored, `25 passed`. The guard is live, not merely green.

## 2. Constructor-kwarg audit (#2949 class)

AST audit over every tracked `.py`: all `AgentSession(...)` call sites, keyword args diffed against the 92 runtime-declared Popoto fields plus the aliases `_normalize_kwargs` explicitly maps.

| Category | Sites | Disposition |
|---|---|---|
| `stage_states=` / `summary=` / `commit_sha=` (the #2949 property conversions) as constructor kwargs | **0** | None found — the 56e80d843 migration covered them all |
| `message_text=` / `sender_name=` / `classification_type=` etc. | 26 | Supported: `_normalize_kwargs` maps them into `initial_telegram_message` / `extra_context`. Not drops. |
| `agent_session_id=` as constructor kwarg | **8** (5 test files) | **Genuine silent drops, removed here.** `_normalize_kwargs` deliberately pops the key ("AutoKeyField, ignore"), verified by runtime probe: the passed value is discarded and the row gets an auto-generated id. All eight literals were referenced nowhere else, so no behavior depended on them — they only misled readers into thinking the id was pinned. |

Post-fix re-audit: **0 silently-dropped kwarg sites remain.**

Files: `test_agent_session_liveness_fields.py`, `test_agent_session_thread_rollup_fields.py`, `test_harness_oom_backoff.py` (4 sites), `test_session_health_compacting_reprieve.py`, `test_session_token_accumulator.py`.

## 3. The unknown-kwargs rejection decision (for the owner to rule)

Not implemented; framed for a ruling as the issue asks. The mechanics, verified by probe: popoto's `Model.__init__` does `self.__dict__.update(kwargs)`, so an unknown kwarg becomes a plain instance attribute — present on the instance, never a declared Field, never persisted (encoding iterates `_meta.fields` only). New `test_unknown_constructor_kwarg_is_tolerated_and_unpersisted` pins this exact behavior with a docstring pointing back here, so an upstream strictness change surfaces as a named test failure rather than a mystery.

**The tradeoff:**

- **Reject (strict `__init__`)** is the durable fix #3001 wanted: the next property conversion fails loudly at the first fixture that constructs with a dead kwarg. But strictness lives upstream in popoto and would need either an upstream change (every downstream consumer inherits it, and hydration paths that round-trip extra keys would need auditing there) or a repo-side `AgentSession.__init__` allowlist that must track popoto's own constructor contract forever.
- **Tolerate (status quo)** keeps hydration and legacy-alias flexibility, and this repo's exposure is now fenced by three guards instead of zero: the fixture self-check (56e80d843), the stub-fidelity warning assert (this PR), and the tolerance pin (this PR).

Recommendation: keep tolerance locally; if strictness is wanted, take it upstream in popoto as an opt-in `Meta.strict_init` so hydration paths choose their own posture. A recorded "no, and here is why" closes the criterion per the issue; the ruling is the owner's.

## Verification

- 105 passed across all seven touched test files (`-n 0`, worktree-pinned imports)
- `tests/unit/test_agent_session_legacy_surface.py` green — the kwarg mapping stays removed
- PLAN/ISSUE negative controls confirmed non-vacuous: the fixture's internal `current_stage` assert guarantees the stage lands, and under the guard-liveness mutation they correctly kept passing (they assert the fallback) while 9 slug-routing tests failed
- ruff check + format clean

### DIFF
--- a/tests/unit/test_agent_session.py
+++ b/tests/unit/test_agent_session.py
@@ -4,6 +4,7 @@ Tests for create_local() behavior, including the chat_id defaulting logic,
 and worker_key property behavior.
 """
 
+import logging
 import uuid
 from unittest.mock import MagicMock, patch
 
@@ -34,7 +35,7 @@ class TestCreateLocalChatId:
         saved_sessions = []
         original_save = AgentSession.save
 
-        def mock_save(self):
+        def mock_save(self, *args, **kwargs):
             saved_sessions.append(self)
 
         AgentSession.save = mock_save
@@ -59,7 +60,7 @@ class TestCreateLocalChatId:
 
         original_save = AgentSession.save
 
-        def mock_save(self):
+        def mock_save(self, *args, **kwargs):
             saved_sessions.append(self)
 
         AgentSession.save = mock_save
@@ -85,7 +86,7 @@ class TestCreateLocalChatId:
         session_uuid = "claude-session-xyz789"
         original_save = AgentSession.save
 
-        def mock_save(self):
+        def mock_save(self, *args, **kwargs):
             pass
 
         AgentSession.save = mock_save
@@ -117,7 +118,7 @@ class TestCreateLocalChatId:
 
         original_save = AgentSession.save
 
-        def mock_save(self):
+        def mock_save(self, *args, **kwargs):
             pass
 
         AgentSession.save = mock_save
@@ -151,7 +152,11 @@ def _make_session(current_stage: str | None = None, **kwargs):
     import json
 
     original_save = AgentSession.save
-    AgentSession.save = lambda self: None
+    # #3074: the stub must accept save()'s real signature. A bare `lambda self:
+    # None` made append_event's `self.save(update_fields=[...])` raise TypeError
+    # on every stage-setting construction — swallowed by append_event's handler,
+    # so tests stayed green while every construction logged a masked failure.
+    AgentSession.save = lambda self, *args, **kwargs: None
     try:
         defaults = {
             "project_key": "test-project",
@@ -189,6 +194,45 @@ def _make_session(current_stage: str | None = None, **kwargs):
         return session
     finally:
         AgentSession.save = original_save
+
+
+class TestFixtureSaveStubFidelity:
+    """#3074: save stubs must match AgentSession.save's real signature.
+
+    A stub of ``lambda self: None`` made append_event's internal
+    ``self.save(update_fields=[...])`` raise TypeError on every stage-setting
+    construction. append_event catches and logs that failure, so the suite
+    stayed green while every fixture construction emitted a masked exception.
+    """
+
+    def test_fixture_construction_emits_no_masked_save_failure(self, caplog):
+        with caplog.at_level(logging.WARNING, logger="models.agent_session"):
+            _make_session(current_stage="BUILD")
+        masked = [r for r in caplog.records if "append_event save failed" in r.getMessage()]
+        assert not masked, (
+            "fixture construction logged a masked append_event save failure — "
+            "the save stub's signature drifted from AgentSession.save again: "
+            f"{masked[0].getMessage()}"
+        )
+
+    def test_unknown_constructor_kwarg_is_tolerated_and_unpersisted(self):
+        """Pins popoto's current tolerance while the #3074 decision stays open.
+
+        Popoto's ``Model.__init__`` does ``self.__dict__.update(kwargs)``, so an
+        unknown kwarg lands as a plain instance attribute, is never a declared
+        Field, and is never persisted (encoding iterates ``_meta.fields`` only).
+        If this test starts failing because construction raises, the ORM went
+        strict upstream: delete this pin and record the #3074 rejection decision
+        as settled.
+        """
+        import popoto
+
+        s = _make_session(definitely_not_a_field_xyz="boom")
+        assert s.definitely_not_a_field_xyz == "boom"
+        assert not isinstance(
+            vars(type(s)).get("definitely_not_a_field_xyz"),
+            popoto.fields.field.Field,
+        )
 
 
 class TestWorkerKeyProperty:
--- a/tests/unit/test_agent_session_liveness_fields.py
+++ b/tests/unit/test_agent_session_liveness_fields.py
@@ -27,7 +27,6 @@ def _make_session(suffix: str) -> AgentSession:
         session_type=SessionType.ENG,
         message_text="x",
         sender_name="x",
-        agent_session_id=f"liveness-fields-{suffix}",
     )
 
 
--- a/tests/unit/test_agent_session_thread_rollup_fields.py
+++ b/tests/unit/test_agent_session_thread_rollup_fields.py
@@ -34,7 +34,6 @@ def _make_session(suffix: str) -> AgentSession:
         session_type=SessionType.ENG,
         message_text="x",
         sender_name="x",
-        agent_session_id=f"thread-rollup-fields-{suffix}",
     )
 
 
--- a/tests/unit/test_harness_oom_backoff.py
+++ b/tests/unit/test_harness_oom_backoff.py
@@ -50,7 +50,6 @@ def test_exit_returncode_field_defaults_to_zero():
         session_type=SessionType.ENG,
         message_text="x",
         sender_name="x",
-        agent_session_id="x-default-test",
     )
     try:
         assert s.exit_returncode == 0
@@ -139,7 +138,6 @@ def test_pending_scan_skips_deferred(clean_sessions):
         session_type=SessionType.ENG,
         message_text="x",
         sender_name="x",
-        agent_session_id="sid-defer-future",
         status="pending",
     )
     deferred.scheduled_at = future
@@ -152,7 +150,6 @@ def test_pending_scan_skips_deferred(clean_sessions):
         session_type=SessionType.ENG,
         message_text="x",
         sender_name="x",
-        agent_session_id="sid-eligible",
         status="pending",
     )
     not_deferred.scheduled_at = None
@@ -221,7 +218,6 @@ def test_store_exit_returncode_persists_value(clean_sessions, monkeypatch):
         session_type=SessionType.ENG,
         message_text="x",
         sender_name="x",
-        agent_session_id="sid-exit-rc",
         session_id="session-id-exit-rc",
     )
     s.save()
--- a/tests/unit/test_session_health_compacting_reprieve.py
+++ b/tests/unit/test_session_health_compacting_reprieve.py
@@ -41,7 +41,6 @@ def _make_session(project_key: str, last_compaction_ts: float | None) -> AgentSe
         session_type=SessionType.ENG,
         message_text="seed",
         sender_name="tester",
-        agent_session_id=f"sid-{int(time.time() * 1000)}",
     )
     s.last_compaction_ts = last_compaction_ts
     s.save()
--- a/tests/unit/test_session_token_accumulator.py
+++ b/tests/unit/test_session_token_accumulator.py
@@ -35,7 +35,6 @@ def test_session():
     s = AgentSession(
         session_id=sid,
         project_key="tst-accum",
-        agent_session_id=f"as-{sid}",
         status="running",
     )
     s.save()


tomcounsell:

**Verdict: APPROVED**

The PR delivers exactly the three residue items of #3074 with unusually honest test construction. The five widened save stubs match `AgentSession.save`'s real signature `(self, *args, update_fields=None, preserve_updated_at=False, **kwargs)`; the shape `(self, *args, **kwargs)` is right for these tests, which assert on the constructed session rather than on save-call arguments, so asserting call args would over-specify. `test_fixture_construction_emits_no_masked_save_failure` asserts precisely what its name claims and is mutation-proven live (narrowed stub → 1 failed naming the masked TypeError). On `test_unknown_constructor_kwarg_is_tolerated_and_unpersisted`: verified against the installed popoto source that the premise-guard is non-vacuous (declared Fields are retained in the class `__dict__` by `ModelBase.__new__`, so a declared field of that name would fail the `isinstance` check), that popoto's metaclass rejects public non-Field class attributes at class creation, and that every persistence path iterates `_meta.fields` only. "Unpersisted" is therefore established by an asserted premise plus an ORM-enforced invariant, with the inference chain documented in the docstring and PR body — acceptable pinning rather than a defect. A behavioral save/reload round-trip would require Redis, which this module's fixture deliberately stubs out, and relocating the pin would separate it from the fixture it documents. The 8 `agent_session_id=` removals are behavior-preserving (`_normalize_kwargs` pops the key unconditionally) and improve reader honesty, since the literals falsely implied the id was pinned. Comments and docstrings are load-bearing at each site, and the fixture's new signature comment accurately describes the failure it prevents. Scope is clean: the 56e80d843 fixture migration is untouched, and the legacy-surface pin stays green.

### Blockers
- None

### Tech Debt
- None

### Nits
- None

- **`tests/unit/test_agent_session.py:216`** — `test_unknown_constructor_kwarg_is_tolerated_and_unpersisted` — "unpersisted" is not asserted via a save/reload round-trip; it is derived from the asserted not-a-declared-Field premise plus popoto's persistence invariant (encoding iterates `_meta.fields` only; class-dict Field retention verified in `ModelBase.__new__`). Disclosed in the docstring and PR body; no fix required.
- **`tests/unit/test_agent_session.py:217`** — Acceptance criterion 4 asks for a *recorded* decision on unknown-kwarg rejection; the PR body records the full tradeoff and a recommendation ("keep tolerance locally; strictness upstream as opt-in `Meta.strict_init`") and explicitly leaves the ruling to the owner. Merging the PR with this body constitutes the recorded "no, and here is why" for the local posture; the upstream opt-in question legitimately stays open.

Judge dict: `{"judge_id": "code-quality", "verdict": "APPROVED", "blockers": 0, "tech_debt": 0, "confidence": 0.87}`

tomcounsell:
## Review (Judge risk):

**Verdict: APPROVED**

Test-only change that removes a masked TypeError rather than introducing one. Every coupling point the new guards depend on was independently verified: the caplog assert's log string matches `models/agent_session.py:2127` exactly, and the logger name (`models.agent_session`, module-level `getLogger(__name__)`) and level (`logger.warning`) both line up, so the fidelity test is live — and the parent's mutation run confirms it bites when the stub narrows. The four `mock_save` defs in `TestCreateLocalChatId` sit outside the fidelity test's coverage, but `create_local` calls `session.save()` bare (`models/agent_session.py:1880`), so those stubs were never on the masked `update_fields` path; widening them is pure precaution and the uncovered gap is hypothetical, not live. On the dropped `agent_session_id=` kwargs: `_normalize_kwargs` popped them pre-PR (`models/agent_session.py:876-877`), so constructed ids are byte-identical before and after, and every cleanup path in the three Redis-writing files deletes via held instance handles or `project_key` filters (`test_harness_oom_backoff.py:32-36`, `test_session_health_compacting_reprieve.py:30-34`, `test_session_token_accumulator.py:42-45`), never by id — no test-row leak into shared Redis. The tolerance pin is a clean tripwire: `popoto.fields.field.Field` resolves and `Model.__init__` does `self.__dict__.update(kwargs)` in the installed popoto, and the docstring states the decommission trigger and action (delete the pin, record the decision) in the test itself, which is where the future failure will surface. Leaving the rejection decision open is adequately fenced for now: the fixture self-check, the stub-fidelity assert, the tolerance pin, and the parent-verified zero-drop audit cover the #2949 hazard class; the 9 statically-uninspectable splat call sites are a pre-existing blind spot unchanged by this PR and disclosed as an audit limitation. Guard liveness of the #1228 worker-key routing tests is parent-verified non-vacuous (mutation flipped 9 tests red, negative controls held).

### Blockers
- None

### Tech Debt
- None

### Nits
- None

- None

Judge dict: `{"judge_id": "risk", "verdict": "APPROVED", "blockers": 0, "tech_debt": 0, "confidence": 0.92}`

tomcounsell:
## Review: Approved

Multi-judge consensus review (code-quality + risk, rule `any-blocker-wins`): **2-of-2 APPROVED**, zero findings, no disagreement (mean confidence 0.895). Test-only PR reviewed against issue #3074's three residue items and its up-front scope correction (the 56e80d843 fixture migration is untouched, verified). Every load-bearing number in the PR body was independently reproduced in the reviewer's own environment (detached worktree at `c2341c0e3`, worktree-pinned imports, `-n 0`):

- **Save-stub fidelity (item 1):** `AgentSession.save`'s real signature is `def save(self, *args, update_fields=None, preserve_updated_at=False, **kwargs)` (`models/agent_session.py:971`); all five widened stubs accept it. Mutation reproduced: with the fixture lambda narrowed back to `lambda self: None`, `TestFixtureSaveStubFidelity` reports **1 failed, 1 passed**, the failure naming the masked `update_fields` TypeError; restored, green.
- **Constructor-kwarg audit (item 2):** an independent AST audit (written fresh for this review, not the author's script) over all tracked `.py` found 92 declared popoto fields, 113 `AgentSession(...)` call sites, and **0 silently-dropped kwargs** on the PR branch. Control: reverting the five audited files to their pre-PR state surfaces exactly the claimed **8** `agent_session_id=` drop sites. `_normalize_kwargs` pops that key unconditionally (`models/agent_session.py:876-877`), and a runtime probe confirmed a passed id is discarded in favor of an auto-generated uuid. The removed literals are referenced nowhere else. The PR's "26 kept sites" reconciles with 33 alias keyword occurrences (`message_text` 20, `sender_name` 7, `classification_type` 6) — sites vs. occurrences, all mapped by `_normalize_kwargs`, not drops.
- **Tolerance pin (item 3):** the pin asserts tolerance behaviorally and guards its premise (the probe name is not a declared Field — non-vacuous, since popoto's metaclass retains declared Fields in the class `__dict__`); "unpersisted" rests on that premise plus the ORM's `_meta.fields`-only encoding, disclosed in the docstring. The rejection-decision writeup stays a report: the diff touches only `tests/unit/`, no behavior change. Merging with the PR body's recorded recommendation ("keep tolerance locally; strictness upstream as opt-in") constitutes the "no, and here is why" that criterion 4 accepts.
- **Guard liveness (criterion 2):** neutralizing `worker_key`'s slug-precedence branch (`if False and ...`) flips **9 failed, 16 passed** in the two worker-key classes; restored, **25 passed**. The PLAN/ISSUE negative controls held under the mutation (they assert the `project_key` fallback), confirming non-vacuousness.
- **Baseline:** **105 passed** across all seven touched files including `test_agent_session_legacy_surface.py` (the kwarg mapping stays removed). `ruff check` and `ruff format --check` clean.

## Rubric

- [x] **1. Plan vs. implementation match** — pass — *no plan doc (no-plan lane); all seven #3074 acceptance criteria validated against the diff, each load-bearing claim reproduced*
- [x] **2. New code quality** — pass — *stubs match the real signature; docstrings and comments are load-bearing and accurate*
- [x] **3. Test coverage** — pass — *both new guards mutation-proven live; 105 baseline green*
- [x] **4. Regression risk to existing callers** — pass — *test-only; removed kwargs were already discarded pre-PR; no cleanup path depended on the removed ids*
- [x] **5. Data integrity** — n/a — *no schema or migration changes*
- [x] **6. Security** — pass — *no secrets, no unsafe patterns*
- [x] **7. Documentation accuracy** — n/a — *test-only change, no user-facing behavior; decision keeps the status quo*
- [x] **8. PR body accuracy** — pass — *every count and mutation claim reproduced (105, 1f/1p, 9f/16p, 8 removed, 0 remaining); "26 sites" reconciles with 33 keyword occurrences*
- [ ] **9. Disclosed deferrals** — acknowledged — *the unknown-kwarg rejection ruling is explicitly framed for the owner with sound rationale; issue #3074's criterion accepts a recorded "no, and here is why"*
- [x] **10. Follow-up claims verified** — n/a — *no "filed as follow-up #N" claims in the PR body*

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — PASS — *all seven #3074 criteria verified, five by direct reproduction*
- **2. No-Gos from plan — none violated** — PASS — *kwarg mapping not restored (legacy-surface pin green); diff stays inside tests/unit/; fixture migration untouched*
- **3. New `except Exception` blocks** — N/A — *none added*
- **4. New integration tests exercise serialization boundary** — N/A — *unit-test fixture hygiene only*
- **5. Plan internal consistency** — N/A — *no plan doc*
- **7. New public APIs — docstrings present** — N/A — *no new public APIs; new tests carry full docstrings*
- **8. Breaking changes — migration path documented** — N/A
- **9. Tests added for new behavior** — PASS — *TestFixtureSaveStubFidelity (2 tests)*
- **10. Tests cover the failure path** — PASS — *failure paths exercised by mutation, both bite*
- **11. UI changes — screenshot captured** — N/A — *no UI files in diff*
- **12. Docs updated for user-facing changes** — N/A — *no user-facing change*

### Blockers
- None

### Tech Debt
- None

### Nits
- None

### Miscellaneous
- None

- **Unknown-kwarg rejection decision framed, ruling left to the owner** — explicit disclosure in the PR body with full tradeoff analysis and a recommendation (keep tolerance locally; strictness upstream as opt-in `Meta.strict_init`); tracked by #3074 itself, whose criterion 4 accepts a recorded "no, and here is why". Merging ratifies the recommendation.
- **"Unpersisted" in the tolerance pin's name is premise-derived, not round-trip-asserted** — disclosed in the test docstring; a save/reload round-trip would require Redis, which the fixture deliberately stubs.

### Legacy Cruft
- None — the diff removes cruft (misleading pinned-id literals, a signature-drifted stub) and adds none.

### Verification Results
- No plan document exists for this lane, so there is no `## Verification` table to run; the reproductions above stand in for it.

### Screenshots
- None — no UI files in the diff (gate is a no-op for this PR).

Issue #3136, opened 2026-09-04
Email attachment context is write-only: extra_context keys are stamped and never read


tomcounsell:
> Found while planning #2732 (spike-4). Filed so #2732 can scope itself to the Telegram reply-chain renderer with an honest boundary.

## Problem

`bridge/email_bridge.py` builds attachment context for an inbound email and writes five keys onto `AgentSession.extra_context` (`bridge/email_bridge.py:1494-1502`):

- `attachments_unrecoverable`
- `attachments_truncated`
- `attachments_recovered_count`
- `attachments_referenced`
- `email_attachments` (the list of persisted attachment paths)

A repo-wide sweep of every `extra_context` reader outside tests — `agent/`, `worker/`, `tools/`, `reflections/`, `ui/`, `bridge/` — finds **no consumer of any of those five keys**. They are written and never read.

The only `extra_context` keys that actually reach the agent's turn input are `injection_risk_banner` (`agent/session_executor.py:1932`) and the context-recall advisory (`agent/session_executor.py:1956`). Both are explicitly spliced into `enriched_text`. Nothing does the equivalent for attachments.

## Consequence

An email arrives with a PDF. The bridge persists the bytes under `EMAIL_ATTACHMENT_DIR` (`bridge/email_bridge.py:104`), records the path, and correctly detects the wedge case where the body references attachments that were not recovered. None of that reaches the agent. The agent answers the email as if it were plain text, with no path to the file and no signal that one exists — the same end state as #2732's Telegram reply chain, arrived at by a different mechanism.

The tests at `tests/unit/test_email_bridge.py:1361-1490` and `tests/integration/test_email_bridge.py:756-817` assert the keys are *stamped*. None assert they are *read*, which is why the gap is invisible from a green suite.

## Desired outcome

Persisted email attachment paths, and the unrecoverable/truncated signal, reach the agent's turn input — so the agent can open an attachment it has, and say specifically what it is missing when it does not. The mechanism should be the one seam that already works for every transport (`agent/session_executor.py` prepending to `enriched_text`), not a second parallel path.

## Recon Summary

**Confirmed:** 3 items
- `bridge/email_bridge.py:1494-1502` writes the five keys listed above.
- No reader exists for any of them outside tests; verified by sweeping every `extra_context` occurrence in tracked non-test Python.
- The working precedent is `agent/session_executor.py:1932` (`injection_risk_banner`), which prepends its banner to `enriched_text` at a single transport-agnostic seam.

[...]

PR #3146, opened 2026-09-04
Reply-chain media renders as an actionable descriptor, never the literal [media]

### LINKED ISSUE #2732: Reply-chain media renders as the literal string `[media]`, stranding attachments the agent could read
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/reply-chain-media-renders-as-literal-string.md

> **Context:** The Telegram bridge hydrates a message's [reply chain](docs/features/reply-thread-context-hydration.md) into the agent's prompt so a reply carries the thread it belongs to. Separately, inbound media (photos, voice notes, documents) is downloaded at intake and [enriched](docs/features/media-enrichment.md) into readable text — transcription, image description, document extraction. These two systems do not meet: the reply-chain renderer replaces any message it cannot read as text with the constant string `[media]`.

## Problem

An attachment sent by one participant becomes invisible to the agent the moment the conversation moves one reply past it. The agent is handed proof that a file exists and no way to reach it — not a path, not a filename, not a tool call. It cannot even tell the human what it is unable to see.

This is not hypothetical. A real teammate-channel exchange:

```
Hazem:  [attaches a document] "my agent's recommendation"
Tom:    "Plausible recommendation, but it brushes over many details and side-effects"
Tom:    "Valor can help you flesh this out with his do-issue and do-plan skills"
Valor:  "Happy to — point me at 'this' and I'll take it from there."
```

The reply chain hydrated correctly. Every hop was intact. What the agent actually received was:

```
REPLY THREAD CONTEXT (oldest to newest):
Hazem: [media]
Tom: Plausible recommendation, but it brushes over many details and side-effects
Tom: Valor can help you flesh this out with his do-issue and do-plan skills
```

The agent knew a recommendation existed and that it was considered thin on side-effects. It had access to zero words of it. Asking "point me at this" was the only honest response available, and it read to the humans as inattention.

**Current behavior:** `bridge/context.py:425` renders every chain ancestor as `msg.text or "[media]"`. A caption-less attachment collapses to four characters. The file's bytes are already on disk — the bridge downloaded them at intake — and the renderer discards the reference.

**Desired outcome:** Media anywhere in agent-facing bridge context is represented as something the agent can act on: a readable local path, or a filename plus the exact tool call needed to retrieve it. The literal string `[media]` appears nowhere. When a file genuinely cannot be resolved, the agent is told *that*, specifically, so it can say so instead of asking the human to repeat themselves.

## Definitions

| Reply chain hydration | Walking `reply_to_msg_id` backward to rebuild a thread and prepend it to the agent's prompt under a `REPLY THREAD CONTEXT` header | [`docs/features/reply-thread-context-hydration.md`](docs/features/reply-thread-context-hydration.md) |
| Enrichment | Post-intake pass converting media into text: voice→transcript, image→description, document→extracted text | [`bridge/enrichment.py`](bridge/enrichment.py), [`docs/features/media-enrichment.md`](docs/features/media-enrichment.md) |
| Trigger message | The single inbound message that caused a session turn, as opposed to its ancestors in the thread | `agent/session_executor.py:1779` (`trigger_telegram_message`) |
| `media_local_path` | Field on `TelegramMessage` holding the on-disk path of media the bridge downloaded at intake | [`bridge/enrichment.py:75`](bridge/enrichment.py) |

## Prior Context

- #1297: *Image/voice/document enrichment silently dropped: worker has no Telegram client* — **the same class of bug, one hop away.** Fixed by downloading at intake and having the worker read `media_local_path` off the record. That fix established the on-disk pattern this issue extends to ancestors.
- #1215: *Telegram file attachments silently dropped in steering path* — another instance of media surviving intake but not reaching the agent through a specific path.
- #949 / PR #953, #1064 / PR #1070: built reply-chain hydration. Both concerned themselves with *whether* the chain arrives, never with what a non-text message renders as inside it.

The recurring shape: media is captured correctly at the boundary and then lost at one specific rendering or handoff seam. This is the next seam.

## Solution Sketch

The bytes already exist on disk for both mediums. `fetch_reply_chain` needs to carry the reference instead of dropping it.

- Replace the `msg.text or "[media]"` fallback with a descriptor carrying filename, media type, and a concrete read affordance — a resolvable path, or a filename plus the tool call signature that retrieves it.
- Keep a caption when one exists. Today a captioned document loses the caption too, since `msg.text` for a document may be empty while `msg.message` holds the caption. Verify which attribute actually carries it.
- Make the unresolvable case explicit and distinguishable from the resolvable one, so the agent can report "the attachment is referenced but I cannot open it" rather than asking the human to re-send.
- Establish the invariant bridge-wide, not just at the one call site, so email and future mediums cannot reintroduce a bare placeholder. Email is partially ahead here already: `bridge/email_bridge.py` has `_body_references_attachments` and stamps `extra_context["attachments_unrecoverable"]`. Prefer generalizing that pattern over inventing a second one.

**Open questions for `/do-plan`** — the root cause is settled, the delivery mechanism is not:

1. **Where does resolution happen?** The fresh-session pre-hydration path runs under a hard 3-second `asyncio.wait_for` budget (`docs/features/reply-thread-context-hydration.md`, Change A/D). Resolving media for a chain up to `max_depth=20` will not fit if it requires any network work. Options: emit path references only at hydration time and let the worker enrich later; bound resolution to the nearest N ancestors; or move chain media resolution entirely to the worker where no 3s ceiling exists. This is the main design decision.
2. **Path reference vs. full enrichment?** A path plus tool signature is cheap and always fits the budget; the agent spends a tool call only when the file matters. Full inline enrichment of every ancestor is expensive and often wasted. The cheap option may be sufficient and should be considered first.
3. **Are ancestor media files still on disk?** `media_local_path` points into `data/media/`. Confirm retention against `docs/features/scheduled-disk-reclaim.md` — a chain hop older than the reclaim window may reference a swept file, which is precisely the unresolvable case that needs its own explicit rendering.
4. **Cross-chat safety.** A resolved path is a filesystem handle surfaced into a group conversation. Confirm rendering a path cannot leak a file across project or chat boundaries.

## Recon Summary

Performed inline against the working tree (no subagents spawned).

**Confirmed:** 4 items
- `bridge/context.py:425` is the **only** bare media placeholder in agent-facing context. Swept `bridge/`, `agent/`, `worker/`, `tools/` for `[media]`, `[image]`, `[photo]`, `[attachment]`, `[document]`, `[file]` variants; the remaining hits are log prefixes, outbound send placeholders, and `scripts/fetch_recent_dms.py` (a human-facing script, out of scope).
- Media enrichment is scoped to exactly one record: `agent/session_executor.py:1779` loads a single `trigger_telegram_message` and `bridge/enrichment.py::enrich_message` operates only on it. Ancestors get no download, extraction, or vision pass by construction.
- Both mediums already persist bytes to disk with paths in Redis: Telegram via `TelegramMessage.media_local_path`, email via `EMAIL_ATTACHMENT_DIR` (`bridge/email_bridge.py:104`), which the comments state was built to mirror the Telegram pattern.
- The reply chain in the reported exchange hydrated correctly. This is a rendering defect, not a hydration failure — an important distinction for anyone diagnosing it from the symptom.

**Revised:** 1 item
- Original framing implied email attachments are equally broken. They are not: email has partial attachment-awareness already. Scope narrows to fixing Telegram reply-chain rendering, plus a shared invariant so other mediums stay correct.

**Pre-requisites:** none. The fix is additive.

**Dropped:** 1 item
- Full recursive enrichment of every chain ancestor. Incompatible with the 3s pre-hydration budget and likely wasteful; left as open question 2 rather than baked into scope.

## Acceptance Criteria

- [ ] `bridge/context.py::fetch_reply_chain` no longer emits the literal string `[media]`; a caption-less attachment in a reply chain renders with its filename, media type, and an actionable read affordance.
- [ ] Given the reported exchange (attachment, then two text replies), the agent's prompt contains a reference sufficient to read the attachment without asking the human for it.
- [ ] A caption on a chain-ancestor attachment survives into the rendered chain.
- [ ] An unresolvable attachment (file swept, download failed) renders as an explicit, distinguishable "referenced but unreadable" marker rather than a generic placeholder or silent omission.
- [ ] Pre-hydration stays within its 3-second budget with a 20-deep chain containing media at every hop; a regression test covers the worst case.
- [ ] The no-bare-placeholder invariant is enforced at a level that covers email and any future bridge medium, with a test that fails if a new medium reintroduces one.
- [ ] `docs/features/reply-thread-context-hydration.md` and `docs/features/media-enrichment.md` both describe how chain-ancestor media is represented.

## Downstream

tomcounsell:
## Summary

Reply-chain hydration no longer collapses a chain ancestor's attachment to the literal string `[media]`. Each media-carrying hop now gets a structured descriptor — filename, media type, and the absolute on-disk path recorded at intake — resolved by a chat_id-scoped `TelegramMessage` lookup run off the event loop via `asyncio.to_thread`. A captioned attachment renders caption plus descriptor; an unresolvable one renders an explicit unreadable marker naming the file and the specific reason (no record, no path recorded, download error, file absent from disk).

Plan: `docs/plans/reply-chain-media-renders-as-literal-string.md` (built past the G2 critique cap on the owner override recorded on the issue; the round-2 blockers were applied to the plan at c99cb231d before build).

## Changes

- `bridge/context.py`: media descriptor record, per-hop ancestor resolver (`_resolve_media_descriptor`, off-loop Popoto lookup, path statted before trusting, per-hop fail-quiet degradation), and renderer composition in `format_reply_chain` (three distinguishable states; Valor drop condition now retains below-floor hops that carry media; descriptor never fed to `filter_tool_logs`; truncation never bisects a path). The `msg.text or "[media]"` fallback is deleted.
- `tests/unit/test_context_helpers.py`: the three rendering states, all unreadable reasons, photo synthetic label, `media_type=None`, chat_id scoping, Valor floor both ways, byte-identical text-only regression fence.
- `tests/unit/test_bridge_context_guards.py`: cross-medium static guards — AST placeholder-shape scan over `bridge/` with in-suite red-state fixtures, write-only `extra_context` scan (receiver-agnostic reader predicate, empirically derived ratcheting allow-list annotated `# known gap: #3136`), and the off-loop enforcement scan.
- `tests/integration/test_steering.py`: 20-hop budget test, stalled-lookup timing-bound test (5s stall, monotonic elapsed < 4.0s, heartbeat liveness), rendered-block size ceiling, and end-to-end `message_text` delivery assertions for both hydration call sites (resume-completed and fresh-session), proving the descriptor survives `strip_private` and reaches `AgentSession.message_text`.
- `docs/features/reply-thread-context-hydration.md`, `docs/features/media-enrichment.md`: chain-ancestor media rendering documented (three states, chat scoping, path-disclosure note; reference-only, never AI enrichment).

## Testing

- [x] All 16 Verification-table rows pass exactly as specified
- [x] Unit tests passing (78 in test_context_helpers, 14 guard tests; full tests/unit run: 4 failures + 9 errors all reproduced on main or non-reproducing in isolation — classified pre-existing/environmental, none in files this branch touches)
- [x] Integration tests passing (reply_chain, message_text_delivery, private-tag ingestion)
- [x] Guard mutation-verified red: injected placeholder fails the tree scan; inlined blocking filter fails the stalled-lookup timing bound
- [x] Lint/format clean (`ruff check`, `ruff format --check`)

## Documentation

- [x] Both feature docs updated per plan
- [x] Related docs scanned (LOW-confidence matches only)

## Definition of Done

- [x] Tested: plan-scoped suites green, no regressions attributable to this diff
- [x] Reviewed: independent validator passed all six checks; final validator passed all Verification rows and Success Criteria
- [x] Documented: docs created/updated and gate passed

Closes #2732

### DIFF
--- a/bridge/context.py
+++ b/bridge/context.py
@@ -1,5 +1,6 @@
 """Context building, conversation history, reply chains, and link summaries."""
 
+import asyncio
 import logging
 import os
 import re
@@ -10,6 +11,7 @@ from telethon import TelegramClient
 
 # Reuse the canonical tool-log filter from bridge.response (single source of truth).
 # See docs/features/bridge-response-improvements.md and issue #1359.
+from bridge.media import MEDIA_DIR, get_media_type
 from bridge.response import filter_tool_logs
 from config.settings import settings
 from tools.link_analysis import (
@@ -376,6 +378,103 @@ def build_conversation_history(chat_id: str, limit: int = 5) -> str:
 # =============================================================================
 
 
+def _media_descriptor(
+    kind: str,
+    filename: str | None,
+    media_type: str | None,
+    local_path: str | None,
+    reason: str | None,
+) -> dict:
+    """Build the media descriptor attached to a reply-chain entry.
+
+    `kind` is "resolved" (file readable at `local_path`) or "unreadable"
+    (`reason` names why). The descriptor is data, not text: the renderer owns
+    presentation and tests assert on this structure.
+    """
+    return {
+        "kind": kind,
+        "filename": filename,
+        "media_type": media_type,
+        "local_path": local_path,
+        "reason": reason,
+    }
+
+
+async def _resolve_media_descriptor(msg, chat_id: int) -> dict | None:
+    """Resolve the media descriptor for one reply-chain hop.
+
+    Returns None when the message carries no media. Otherwise returns a
+    descriptor (see `_media_descriptor`): media type and filename come from
+    the Telethon message already in hand, the on-disk path from the
+    `TelegramMessage` record the bridge wrote at intake. The two sources
+    degrade separately — a Redis miss still yields a named attachment,
+    downgraded to "unreadable" with a specific reason.
+
+    Never raises: any failure logs at warning and degrades to an
+    "unreadable" descriptor so the chain walk continues past this hop.
+    """
+    if not (getattr(msg, "file", None) or getattr(msg, "media", None)):
+        return None
+
+    try:
+        media_type = get_media_type(msg)
+        filename = getattr(getattr(msg, "file", None), "name", None)
+        if not filename:
+            # Photos genuinely have no filename — synthesize a stable label.
+            filename = f"{media_type or 'media'}-{msg.id}"
+
+        # Lazy import: models.telegram pulls in Popoto/Redis, which this
+        # module must not require at import time (same shape as _cache_walk_root).
+        from models.telegram import TelegramMessage
+
+        # Popoto is synchronous redis-py; an inline call here would block the
+        # bridge event loop and make the callers' asyncio.wait_for guard
+        # unenforceable. chat_id scoping is a hard requirement: message ids
+        # are per-chat sequences, so an unscoped lookup could resolve another
+        # chat's file.
+        records = await asyncio.to_thread(
+            lambda: list(TelegramMessage.query.filter(chat_id=str(chat_id), message_id=msg.id))
+        )
+        if not records:
+            return _media_descriptor("unreadable", filename, media_type, None, "no_record")
+
+        record = records[0]
+        download_error = getattr(record, "media_download_error", None)
+        if download_error:
+            return _media_descriptor(
+                "unreadable", filename, media_type, None, f"download_error: {download_error}"
+            )
+
+        local_path = getattr(record, "media_local_path", None)
+        if not local_path or not str(local_path).strip():
+            return _media_descriptor("unreadable", filename, media_type, None, "no_path_recorded")
+
+        path = Path(str(local_path).strip())
+        try:
+            inside_media_dir = path.resolve().is_relative_to(MEDIA_DIR.resolve())
+        except (OSError, ValueError):
+            inside_media_dir = False
+        if not inside_media_dir:
+            return _media_descriptor("unreadable", filename, media_type, None, "invalid_path")
+
+        if not (path.exists() and os.access(path, os.R_OK)):
+            return _media_descriptor("unreadable", filename, media_type, None, "file_missing")
+
+        return _media_descriptor("resolved", filename, media_type, str(path), None)
+    except Exception as e:
+        logger.warning(
+            f"Could not resolve media descriptor for message "
+            f"{getattr(msg, 'id', 'unknown')} in chat {chat_id}: {e}"
+        )
+        return _media_descriptor(
+            "unreadable",
+            f"media-{getattr(msg, 'id', 'unknown')}",
+            None,
+            None,
+            "resolution_error",
+        )
+
+
 async def fetch_reply_chain(
     client: TelegramClient,
     chat_id: int,
@@ -395,7 +494,8 @@ async def fetch_reply_chain(
         max_depth: Maximum number of messages to fetch in the chain
 
     Returns:
-        List of message dicts with 'sender', 'content', 'message_id', 'date'
+        List of message dicts with 'sender', 'content', 'message_id', 'date',
+        and 'media' (a descriptor dict for hops carrying media, else None)
     """
     chain = []
     current_id = message_id
@@ -419,12 +519,17 @@ async def fetch_reply_chain(
             if msg.out:
                 sender_name = "Valor"
 
+            # Never raises — a resolution failure degrades this hop to an
+            # "unreadable" descriptor instead of reaching the loop-tail break.
+            media = await _resolve_media_descriptor(msg, chat_id)
+
             chain.append(
                 {
                     "sender": sender_name,
                     "content": msg.text or "[media]",
                     "message_id": msg.id,
                     "date": msg.date,
+                    "media": media,
                 }
             )
 
--- a/bridge/context.py
+++ b/bridge/context.py
@@ -526,7 +526,7 @@ async def fetch_reply_chain(
             chain.append(
                 {
                     "sender": sender_name,
-                    "content": msg.text or "[media]",
+                    "content": msg.text or "",
                     "message_id": msg.id,
                     "date": msg.date,
                     "media": media,
@@ -548,9 +548,30 @@ async def fetch_reply_chain(
     return chain
 
 
+def _format_media_descriptor(media: dict) -> str:
+    """Render one media descriptor as a compact single-line marker.
+
+    The filename (a basename) is the human-readable name to quote; the
+    absolute path is an explicitly machine-facing affordance for the agent's
+    file tools. The unreadable rendering names the file and the reason and is
+    textually distinguishable from the resolved one.
+    """
+    name = media.get("filename") or "unnamed"
+    media_type = media.get("media_type") or "file"
+    local_path = media.get("local_path")
+    if media.get("kind") == "resolved" and local_path:
+        return f"[attachment: {name} ({media_type}) at machine path {local_path}]"
+    reason = media.get("reason") or "unknown"
+    return f"[unreadable attachment: {name} ({media_type}) reason: {reason}]"
+
+
 def format_reply_chain(chain: list[dict]) -> str:
     """
     Format a reply chain for inclusion in agent context.
+
+    Each entry's line composes the human-authored text with its media
+    descriptor: both when both exist, the descriptor alone when the text is
+    empty, the text alone when the entry carries no media.
 
     Args:
         chain: List of message dicts from fetch_reply_chain()
@@ -567,11 +588,14 @@ def format_reply_chain(chain: list[dict]) -> str:
     for msg in chain:
         sender = msg["sender"]
         content = msg["content"]
+        media = msg.get("media")
 
-        # Filter tool logs from Valor's messages
+        # Filter tool logs from Valor's messages. A hop whose text filters to
+        # nothing is still kept when it carries media — its descriptor is the
+        # whole line.
         if sender == "Valor":
             content = filter_tool_logs(content)
-            if not content:
+            if not content and not media:
                 continue
 
         # Valor's messages are already summarized — include in full
@@ -580,6 +604,14 @@ def format_reply_chain(chain: list[dict]) -> str:
         max_len = 2000 if sender == "Valor" else 500
         if len(content) > max_len:
             content = content[:max_len] + "..."
+
+        # Compose after filtering and truncation: filter_tool_logs is written
+        # for human/tool text and must never see the descriptor, and
+        # truncation measures the human-authored text so it can never bisect
+        # a path.
+        if media:
+            descriptor = _format_media_descriptor(media)
+            content = f"{content} {descriptor}" if content else descriptor
 
         # Format with timestamp if available
         date_str = ""
--- a/bridge/context.py
+++ b/bridge/context.py
@@ -403,7 +403,9 @@ def _media_descriptor(
 async def _resolve_media_descriptor(msg, chat_id: int) -> dict | None:
     """Resolve the media descriptor for one reply-chain hop.
 
-    Returns None when the message carries no media. Otherwise returns a
+    Returns None when the message carries no media, or media that is not a
+    downloadable file (get_media_type is the eligibility authority: only
+    photo/document kinds ever resolve). Otherwise returns a
     descriptor (see `_media_descriptor`): media type and filename come from
     the Telethon message already in hand, the on-disk path from the
     `TelegramMessage` record the bridge wrote at intake. The two sources
@@ -510,7 +512,9 @@ async def fetch_reply_chain(
 
     Returns:
         List of message dicts with 'sender', 'content', 'message_id', 'date',
-        and 'media' (a descriptor dict for hops carrying media, else None)
+        and 'media' (a descriptor dict for hops carrying downloadable file
+        media per get_media_type; None for text-only hops, non-file media,
+        or when resolve_media is False)
     """
     chain = []
     current_id = message_id
--- a/bridge/context.py
+++ b/bridge/context.py
@@ -8,6 +8,13 @@ import subprocess
 from pathlib import Path
 
 from telethon import TelegramClient
+from telethon.tl.types import (
+    MessageMediaContact,
+    MessageMediaDice,
+    MessageMediaGeo,
+    MessageMediaPoll,
+    MessageMediaWebPage,
+)
 
 # Reuse the canonical tool-log filter from bridge.response (single source of truth).
 # See docs/features/bridge-response-improvements.md and issue #1359.
@@ -377,6 +384,22 @@ def build_conversation_history(chat_id: str, limit: int = 5) -> str:
 # Reply Chain Management
 # =============================================================================
 
+# Telethon media kinds that are never a downloadable attachment. Telegram sets
+# MessageMediaWebPage on any plain-text message that gets a link preview, and
+# Message.file/.photo/.document fall back to the web preview's photo/document
+# for that case too — so neither `msg.media` nor `msg.file` truthiness alone
+# distinguishes "has an attachment" from "has a link preview". Polls, geo
+# pins, contacts, and dice hit the same non-file-media shape. Excluding these
+# up front keeps ordinary link messages from rendering a spurious
+# "[unreadable attachment]" marker (see PR #3146 review).
+_NON_FILE_MEDIA_TYPES = (
+    MessageMediaWebPage,
+    MessageMediaPoll,
+    MessageMediaGeo,
+    MessageMediaContact,
+    MessageMediaDice,
+)
+
 
 def _media_descriptor(
     kind: str,
@@ -413,7 +436,14 @@ async def _resolve_media_descriptor(msg, chat_id: int) -> dict | None:
     Never raises: any failure logs at warning and degrades to an
     "unreadable" descriptor so the chain walk continues past this hop.
     """
-    if not (getattr(msg, "file", None) or getattr(msg, "media", None)):
+    media = getattr(msg, "media", None)
+    if isinstance(media, _NON_FILE_MEDIA_TYPES):
+        # Checked before the file/media truthiness test below: Telethon's
+        # Message.file (and .photo/.document) fall back to the web preview's
+        # own photo/document for MessageMediaWebPage, so a truthy msg.file
+        # cannot be trusted to rule these out on its own.
+        return None
+    if not (getattr(msg, "file", None) or media):
         return None
 
     try:
--- a/bridge/context.py
+++ b/bridge/context.py
@@ -8,13 +8,6 @@ import subprocess
 from pathlib import Path
 
 from telethon import TelegramClient
-from telethon.tl.types import (
-    MessageMediaContact,
-    MessageMediaDice,
-    MessageMediaGeo,
-    MessageMediaPoll,
-    MessageMediaWebPage,
-)
 
 # Reuse the canonical tool-log filter from bridge.response (single source of truth).
 # See docs/features/bridge-response-improvements.md and issue #1359.
@@ -384,22 +377,6 @@ def build_conversation_history(chat_id: str, limit: int = 5) -> str:
 # Reply Chain Management
 # =============================================================================
 
-# Telethon media kinds that are never a downloadable attachment. Telegram sets
-# MessageMediaWebPage on any plain-text message that gets a link preview, and
-# Message.file/.photo/.document fall back to the web preview's photo/document
-# for that case too — so neither `msg.media` nor `msg.file` truthiness alone
-# distinguishes "has an attachment" from "has a link preview". Polls, geo
-# pins, contacts, and dice hit the same non-file-media shape. Excluding these
-# up front keeps ordinary link messages from rendering a spurious
-# "[unreadable attachment]" marker (see PR #3146 review).
-_NON_FILE_MEDIA_TYPES = (
-    MessageMediaWebPage,
-    MessageMediaPoll,
-    MessageMediaGeo,
-    MessageMediaContact,
-    MessageMediaDice,
-)
-
 
 def _media_descriptor(
     kind: str,
@@ -436,22 +413,26 @@ async def _resolve_media_descriptor(msg, chat_id: int) -> dict | None:
     Never raises: any failure logs at warning and degrades to an
     "unreadable" descriptor so the chain walk continues past this hop.
     """
-    media = getattr(msg, "media", None)
-    if isinstance(media, _NON_FILE_MEDIA_TYPES):
-        # Checked before the file/media truthiness test below: Telethon's
-        # Message.file (and .photo/.document) fall back to the web preview's
-        # own photo/document for MessageMediaWebPage, so a truthy msg.file
-        # cannot be trusted to rule these out on its own.
-        return None
-    if not (getattr(msg, "file", None) or media):
+    if not getattr(msg, "media", None):
         return None
 
     try:
+        # Classifier gate (PR #3146 review, round 2): get_media_type is
+        # non-None exactly for the Photo/Document set the intake download
+        # path can resolve, so a descriptor exists only when a file could
+        # ever have been downloaded. Everything else — link previews (whose
+        # msg.file falls back to the web preview's own photo/document),
+        # polls, geo pins static or live, venues, contacts, dice, stories,
+        # and every future Telethon media kind — fails closed with no
+        # descriptor, instead of rendering a false "[unreadable attachment]"
+        # marker for a file that never existed.
         media_type = get_media_type(msg)
+        if media_type is None:
+            return None
         filename = getattr(getattr(msg, "file", None), "name", None)
         if not filename:
             # Photos genuinely have no filename — synthesize a stable label.
-            filename = f"{media_type or 'media'}-{msg.id}"
+            filename = f"{media_type}-{msg.id}"
 
         # Lazy import: models.telegram pulls in Popoto/Redis, which this
         # module must not require at import time (same shape as _cache_walk_root).
@@ -510,6 +491,7 @@ async def fetch_reply_chain(
     chat_id: int,
     message_id: int,
     max_depth: int = 20,
+    resolve_media: bool = True,
 ) -> list[dict]:
     """
     Fetch the entire reply chain for a message.
@@ -522,6 +504,9 @@ async def fetch_reply_chain(
         chat_id: Chat ID to fetch from
         message_id: Starting message ID (the one being replied to)
         max_depth: Maximum number of messages to fetch in the chain
+        resolve_media: when False, skip per-hop media descriptor resolution
+            (and its Redis lookups) entirely — for callers that consume only
+            sender/message_id, like the session-routing walk
 
     Returns:
         List of message dicts with 'sender', 'content', 'message_id', 'date',
@@ -551,7 +536,7 @@ async def fetch_reply_chain(
 
             # Never raises — a resolution failure degrades this hop to an
             # "unreadable" descriptor instead of reaching the loop-tail break.
-            media = await _resolve_media_descriptor(msg, chat_id)
+            media = await _resolve_media_descriptor(msg, chat_id) if resolve_media else None
 
             chain.append(
                 {
@@ -625,8 +610,13 @@ def format_reply_chain(chain: list[dict]) -> str:
         # whole line.
         if sender == "Valor":
             content = filter_tool_logs(content)
-            if not content and not media:
-                continue
+
+        # Any hop with no text and no descriptor renders nothing: a bare
+        # "Sender:" label carries no information. Covers Valor hops whose
+        # text filtered away and caption-less non-file media (a poll or geo
+        # share) whose descriptor is deliberately absent (PR #3146 review).
+        if not content and not media:
+            continue
 
         # Valor's messages are already summarized — include in full
         # so resumed sessions have complete context of what was sent.
@@ -764,7 +754,11 @@ async def resolve_root_session_id(
             f"[session-root] cache miss for msg_id={reply_to_msg_id}, "
             f"falling back to Telegram API chain walk"
         )
-        chain = await fetch_reply_chain(client, chat_id, reply_to_msg_id)
+        # resolve_media=False: this walk consumes only sender/message_id, so
+        # paying per-hop Redis descriptor resolution here would add
+        # Redis-health-bounded latency to session routing for data nothing
+        # reads (PR #3146 review).
+        chain = await fetch_reply_chain(client, chat_id, reply_to_msg_id, resolve_media=False)
         # Chain is chronological (oldest first). Find the first non-Valor message.
         for entry in chain:
             if entry.get("sender") != "Valor":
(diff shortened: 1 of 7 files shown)


valorengels (changes requested):
## Review: CHANGES REQUESTED

Strong work overall — the descriptor design, the chat-scoped lookup, the off-loop
`asyncio.to_thread` wrapping, and the guard tests with in-suite red-state fixtures are all
exactly what the plan asked for, and the plan-scoped suites are green in my environment
(92 unit + 8 integration + 11 private-tag). One real regression blocks it: the "has media"
predicate accepts Telegram media that is not a file, so ordinary link-preview messages now
render a false attachment marker.

### Blockers

- **File:** `bridge/context.py:409` (verified: read this file; reproduced against the installed Telethon pin)
  **Code:** `if not (getattr(msg, "file", None) or getattr(msg, "media", None)): return None`
  **Issue:** `msg.media` is truthy for media that is not a file. Telegram sets
  `MessageMediaWebPage` on any plain text message that gets a link preview, so a chain hop
  like `check this out https://example.com` now falls through the guard, `get_media_type()`
  returns `None` (the repo's own classifier says "this is not media"), and the hop renders as:

  ```
  Tom: check this out https://example.com [unreadable attachment: media-555 (file) reason: no_record]
  ```

  Reproduced directly:

  ```
  media = MessageMediaWebPage(webpage=WebPageEmpty(id=1))
  msg   = FakeMsg(555, "check this out https://example.com", media)
  get_media_type(msg)                      -> None
  await _resolve_media_descriptor(msg, ..) -> {'kind': 'unreadable', 'filename': 'media-555',
                                               'media_type': None, 'local_path': None,
                                               'reason': 'no_record'}
  ```

  In production the ancestor record does exist (`store_message` writes every message with
  `has_media=bool(message.media)`), so the reason reads `no_path_recorded` instead — still a
  spurious marker. Telethon's `Message.photo`/`Message.document` also fall back to the web
  preview's photo/document, so `msg.file` can be non-`None` for a preview too; excluding only
  on `msg.file` is not sufficient. `MessageMediaPoll`, `MessageMediaGeo`, `MessageMediaContact`
  and `MessageMediaDice` hit the same path.

  This is a behavior *regression* introduced by this diff: before it, `msg.text or "[media]"`
  left a link message clean because the text was truthy. Link sharing is common in these chats,
  so the practical effect is that a large share of chain hops now tell the agent there is a file
  it cannot open when no file was ever attached — the same class of harm #2732 is about
  (the agent misinforming humans about attachments), just inverted.

  **Severity:** blocker
  **Fix:** make the predicate agree with the repo's own classifier instead of with raw
  `msg.media` truthiness — e.g. return `None` unless `get_media_type(msg)` is not `None`, or
  explicitly exclude `MessageMediaWebPage`/poll/geo/contact/dice before building a descriptor.
  Add a rendering test for "text message with a link preview renders with no attachment marker"
  next to the existing `test_unknown_media_type_still_yields_descriptor` — that test currently
  pins the buggy behavior (`media=object()` -> `media-7` descriptor), so it needs to be
  re-pointed at genuine-media-with-unknown-type rather than deleted.

### Tech Debt
- None

### Nits

- **File:** `docs/plans/reply-chain-media-renders-as-literal-string.md` (Verification table, "Feature docs updated" row)
  **Code:** `grep -l 'chain-ancestor media\|ancestor media' docs/features/... \| wc -l`
  **Issue:** The row passes in a shell but FAILS through the repo's own verification runner
  (`agent.verification_parser`): the table's `\|` escape is unescaped back to a bare `|`, which
  BRE `grep` reads as a literal pipe, so the pattern never matches and the row reports `0`. I
  verified the row passes (`2`) when run by hand. Same table: the five `pytest-clean.sh` rows all
  report `FAIL / timed out after 120s` under the runner's per-command cap even though every one
  of them passes when run directly. So the PR's "All 16 Verification-table rows pass exactly as
  specified" is true of a shell, not of the runner the pipeline uses.
  **Severity:** nit
  **Fix:** use a character class or `grep -E` in that row so it survives the parser's unescape,
  and either split the pytest rows finer or note their runtime. Plan docs land on `main`, so this
  is a follow-up edit rather than a branch change.

### Miscellaneous
- None

- **Email attachment delivery seam / unifying the Telegram and email renderers** — tracked by
  **#3136 (OPEN)**; the anti-criterion row confirms `bridge/email_bridge.py` and
  `agent/session_executor.py` are untouched by this diff (verified: 0 files).
- **`_cache_walk_root`'s pre-existing inline blocking `filter`** — explicit out-of-scope plan
  No-Go; the off-loop guard carries a ratcheted exemption that fails loudly once it is fixed.
- **Valor's outbound media hops always resolving *unreadable*** — explicit out-of-scope
  exclusion with sound rationale (outbound stores never set `media_local_path`); documented.
- **Two plan NITs left `pending` after the G2 cap** (timeout-constant grep row, `models/telegram.py`
  line citation) — disclosed on the plan, non-load-bearing.

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — PASS — *all 12 Success Criteria delivered; verified the 16-row table by hand.*
- **2. No-Gos from plan — none violated** — PASS — *email seam untouched (0 files); no AI enrichment in `bridge/context.py`; no outbound download; `_cache_walk_root` left alone.*
- **3. New `except Exception` blocks — each has logger/raise/swallow-ok** — PASS — *the resolver's `except` logs at WARNING and returns an unreadable descriptor; asserted by `test_resolver_failure_logs_warning_and_returns_unreadable`. `CancelledError` is a `BaseException` so the `wait_for` guard still preempts.*
- **4. New integration tests — exercise serialization boundary (not in-memory only)** — PASS — *`TestReplyChainMessageTextDelivery` drives real `resume_completed_session` -> enqueue and asserts on the persisted `AgentSession.message_text`.*
- **5. Plan internal consistency — spike findings match task steps** — PASS — *spikes 1-4 each land in the implementation (off-loop lookup, caption composition, stat-the-path, email split).*
- **6. No hardcoded secrets or debug artifacts** — PASS — *none.*
- **7. New public APIs — docstrings present** — PASS — *`_media_descriptor`, `_resolve_media_descriptor`, `_format_media_descriptor` all documented; `fetch_reply_chain`/`format_reply_chain` docstrings updated for the new key.*
- **8. Breaking changes — migration path documented** — N/A — *chain dicts gain an optional key read via `.get`; the legacy shape is byte-identical, fenced by `test_text_only_chain_is_byte_identical_with_and_without_media_key`.*
- **9. Tests added for new behavior** — PASS — *3 rendering states, 6 unreadable reasons, chat scoping, photo/None-type labels, Valor floor both ways, 20-hop budget, stalled-lookup timing bound, size ceiling, both delivery call sites.*
- **10. Tests cover the failure path (not just happy path)** — FAIL — *thorough on every enumerated failure mode, but the non-file-media input (link preview, poll, geo) has no test and is the blocker above; `test_unknown_media_type_still_yields_descriptor` pins the wrong behavior for it.*
- **11. UI changes (if any) — screenshot captured** — N/A — *no HTML/CSS/JS/template files in the diff; visual proof gate is a no-op.*
- **12. Docs updated for user-facing changes** — PASS — *both feature docs describe all three states, chat scoping, and the path-disclosure posture, and match the code.*

## Rubric

- [x] **1. Plan vs. implementation match** — pass — *every Success Criterion is delivered in the diff.*
- [ ] **2. New code quality** — fail — *the has-media predicate trusts raw `msg.media` truthiness over the repo's own `get_media_type` classifier; see the blocker.*
- [x] **3. Test coverage** — pass — *dense and mutation-aware; the one gap is folded into rubric item 2.*
- [x] **4. Regression risk to existing callers** — pass — *both bridge call sites pass the chain through opaquely; `format_reply_chain` reads `media` via `.get`, and `resolve_root_session_id`'s use of `fetch_reply_chain` only reads `sender`/`message_id`.*
- [x] **5. Data integrity** — n/a — *read-only use of existing `TelegramMessage` fields; no schema change, so `scripts/update/migrations.py` is correctly untouched.*
- [x] **6. Security** — pass — *chat-scoped lookup verified by test; resolved paths confined to `MEDIA_DIR` via `is_relative_to`; no `MEDIA_DIR` glob; path disclosure is a plan-level decision, documented.*
- [x] **7. Documentation accuracy** — pass — *both docs match the implemented reasons and states.*
- [x] **8. PR body accuracy** — pass — *diff, file list, and test counts check out. The "all 16 rows pass exactly as specified" claim is shell-true but runner-false; captured as a nit. The full-`tests/unit` triage line (4 failures + 9 errors, pre-existing) is **unverified** — I did not reproduce a full unit run, so I am not crediting it.*
- [x] **9. Disclosed deferrals** — pass — *see Acknowledged Deferrals.*
- [x] **10. Follow-up claims verified** — pass — *#3136 is OPEN.*

**Verdict derivation:** rubric item 2 (critical) is `fail` with no matching acknowledgement -> Rule 1 -> **CHANGES REQUESTED — Blocker**.

### Verification (re-run in my environment)

| Check | Result |
|---|---|
| `tests/unit/test_context_helpers.py` + `tests/unit/test_bridge_context_guards.py` | 92 passed |
| `tests/integration/test_steering.py -k "reply_chain or message_text_delivery"` | 8 passed |
| `tests/integration/test_private_tag_ingestion.py` | 11 passed |
| 11 non-pytest Verification rows (incl. both anti-criteria, lint, format) | pass |
| "Feature docs updated" row | passes by hand (2), fails through the runner — see nit |

### Legacy Cruft Audit
- No new cruft. `msg.text or "[media]"` is deleted outright with no compatibility shim, both
  new guard exemptions (`_OFF_LOOP_EXEMPT_FUNCTIONS`, `WRITE_ONLY_ALLOW_LIST`) ratchet shut,
  and the stale `enrich_reply_to_msg_id` variable was correctly left alone per the plan's
  Rabbit Holes.

### Deploy note
This touches `bridge/`, which the long-lived bridge process imports. It needs
`./scripts/valor-service.sh restart` on every bridge machine after merge, not just `/update`.

tomcounsell:

Round 2 at head `2c1921a`. Confidence 0.85. All 13 Success Criteria otherwise trace to delivered code; descriptor-as-data contract, renderer ownership, and test structure are sound.

### Blockers

- None

### Tech Debt

- **`bridge/context.py:395-401`** — `_NON_FILE_MEDIA_TYPES` — the deny-list omits at least seven more non-file media kinds (Venue, GeoLive, Story, ToDo, Invoice, PaidMedia, Giveaway/GiveawayResults, Game, Unsupported, Empty), which still render the false `[unreadable attachment]` marker the round-1 fix meant to eliminate. Minimal fix: extend the tuple with the missing kinds, parametrize the exclusion test over the full set, and record the Story/Unsupported/PaidMedia judgment call in a comment. Note: switching to a classifier gate (`get_media_type(msg) is not None`) flips unknown-media handling from fail-open to fail-closed and requires re-pointing the `media=object()` fixtures in both test files at genuine Telethon types.

### Nits

- **`bridge/context.py` (renderer path)** — excluded caption-less hops now render a bare `Sender: ` line, a new state with no test pinning it. Either skip the line entirely for empty-content hops or pin the current rendering deliberately with a test.

tomcounsell:

Round 2 at head `2c1921a`. Confidence 0.87. Verified holdings from round 1: chat_id-scoped lookup without globbing, fail-closed `is_relative_to` path confinement, off-loop discrimination, per-hop degradation (a failed hop never kills the chain walk), and Valor-retention consumer safety. 93 unit tests green, mutation checks biting, ruff clean.

### Blockers

- [ ] **`bridge/context.py:395-401` / `:440-446`** — `_NON_FILE_MEDIA_TYPES = (MessageMediaWebPage, MessageMediaPoll, MessageMediaGeo, MessageMediaContact, MessageMediaDice)` — the round-1 fix is a deny-list of five types, and Telethon 1.42.0 ships many more non-file media kinds that slip past it. Reproduced the false `[unreadable attachment media-{id} ... no_path_recorded]` marker for `MessageMediaVenue`, `MessageMediaGeoLive`, `MessageMediaStory`, `MessageMediaToDo`, `MessageMediaInvoice`, `MessageMediaPaidMedia`, `MessageMediaGiveaway`, `MessageMediaGiveawayResults`, `MessageMediaGame`, `MessageMediaUnsupported`, and `MessageMediaEmpty`. Verified during aggregation: `issubclass(MessageMediaGeoLive, MessageMediaGeo)` is False, so a live location renders a false marker while a static pin does not — an inconsistency users will see. The deny-list also fails open on future Telethon layers. Prescribed fix: gate descriptor construction on `get_media_type(msg) is not None` (`bridge/media.py:213` — the repo's own classifier, non-None only for Photo/Document, exactly the set the download path can resolve). Re-point `test_unknown_media_type_still_yields_descriptor` at a genuine Document with an unknown extension, and add a GeoLive rendering test.

### Tech Debt

- **`bridge/context.py:767`** — `chain = await fetch_reply_chain(client, chat_id, reply_to_msg_id)` — `resolve_root_session_id`'s API fallback now pays per-hop `to_thread` Redis descriptor resolution it never reads (the loop below consumes only `sender` and `message_id`), with no `wait_for` bound, and the plan does not disclose the added latency on the session-root path. Fix: a `resolve_media: bool = True` parameter on `fetch_reply_chain`, passed `False` from Step 2, or wrap the call in `_REPLY_CHAIN_FETCH_TIMEOUT_S`.

### Nits

- None

tomcounsell:
## Review: CHANGES REQUESTED

Multi-judge consensus (2/2 judges: risk, code-quality), rule `any-blocker-wins`, mean confidence 0.86. Both judges independently converged on the same root finding — the round-1 deny-list fix does not close the false-marker class — differing only in severity; the risk judge's blocker rating carries under any-blocker-wins. Round 2 at head `2c1921acd620c5346568c1b8b57601935b1526ac` against plan `docs/plans/reply-chain-media-renders-as-literal-string.md` (Closes #2732).

Mechanical verification (round-2 pass, reproduced): 93 unit tests green, mutation checks biting on the changed guards, `ruff check` and `ruff format --check` clean, Telethon 1.42.0 isinstance enumeration reproduced during aggregation (`issubclass(MessageMediaGeoLive, MessageMediaGeo)` is False; `get_media_type` returns non-None only for Photo/Document).

### Blockers

- [ ] **`bridge/context.py:395-401` / `:440-446`** — `_NON_FILE_MEDIA_TYPES` deny-list of five types fails open: `MessageMediaVenue`, `GeoLive`, `Story`, `ToDo`, `Invoice`, `PaidMedia`, `Giveaway(Results)`, `Game`, `Unsupported`, and `Empty` all slip past it and render the false `[unreadable attachment media-{id} ... no_path_recorded]` marker (reproduced). A live location renders a false marker while a static pin does not, and every future Telethon media layer reopens the bug.

### Tech Debt

- **`bridge/context.py:767`** — `resolve_root_session_id`'s API fallback pays per-hop `to_thread` Redis descriptor resolution it never reads, unbounded and undisclosed by the plan. Fix: `resolve_media: bool = True` on `fetch_reply_chain`, passed `False` from Step 2, or bound with `_REPLY_CHAIN_FETCH_TIMEOUT_S`.

### Nits

- **`bridge/context.py` (renderer path)** — excluded caption-less hops render a bare `Sender: ` line, an untested new state. Skip the line or pin it deliberately with a test.

### Miscellaneous

- None

- None

### Fix direction for /do-patch (fork resolved)

The two judges prescribe different shapes for the blocker. **Implement the classifier gate**: replace the `isinstance(media, _NON_FILE_MEDIA_TYPES)` deny-list with a gate on `get_media_type(msg) is not None` (`bridge/media.py:213`). It closes the class permanently — the classifier is non-None exactly for the Photo/Document set the download path can resolve, so unknown and future media kinds fail closed instead of open. This matches the risk judge's prescription and the repo's intelligent-systems-over-rigid-patterns principle. Carry the code-quality judge's test work as part of the same change: re-point the `media=object()` fixtures in both test files at genuine Telethon types, re-point `test_unknown_media_type_still_yields_descriptor` at a genuine Document with an unknown extension, add a GeoLive rendering (exclusion) test, and parametrize the exclusion test over the enumerated non-file kinds. Fall back to extending the tuple only if the classifier gate demonstrably flips behavior the plan pins — none was identified by either judge.

### Verification Results

- Unit suite: 93 tests green on the PR branch (round-2 run, reproduced by the review fork)
- Mutation checks: guards bite when mutated
- Lint: `ruff check` and `ruff format --check` clean
- Telethon enumeration: reproduced in this environment (1.42.0)

tomcounsell:

Round 3 at head `412db25bc` (plan-tick commit `5fde69e2d` follows, docs/plans only). All three round-2 findings verified resolved exactly as the aggregate prescribed.

- **Classifier gate (blocker fix):** `_resolve_media_descriptor` gates on `get_media_type(msg) is None`; `get_media_type` (`bridge/media.py:213`) is non-None only for `MessageMediaPhoto` and `MessageMediaDocument` with a real document — fail-closed for all current and future non-file kinds. `_NON_FILE_MEDIA_TYPES` and its five Telethon imports fully deleted, zero residue repo-wide. The synthesized filename correctly collapsed from `media_type or 'media'` to `f"{media_type}-{msg.id}"` since the gate guarantees non-None.
- **Exclusion test quality:** `test_non_file_media_yields_no_descriptor` saves the production intake record shape (`has_media=True`, no path) so a fail-open gate would return a `no_path_recorded` descriptor and fail the `is None` assertion — it bites for the right reason (parent's mutation run: 11 red with the gate removed).
- **GeoLive e2e test** pins both the blocker (no `unreadable`/`no_path_recorded` in output) and the round-2 nit (`Hazem` lines exactly `["Hazem: on my way"]` — no bare label).
- **Re-pointed fixtures do not weaken assertions:** the 13 `media=object()` → genuine `MessageMediaDocument` conversions keep every kind/reason/filename assertion intact; the renamed `test_unknown_extension_document_still_yields_descriptor` correctly pins that a real document with an unknown extension still yields a descriptor.
- **Generalized hop-skip** fires only on `not content and not media`; a text-bearing hop always renders. The only behavior change is that a non-Valor empty-text/no-descriptor hop (previously a bare `Sender:` line) renders nothing — the intended fix.
- **resolve_media flag:** default `True` preserves both hydration call sites; the routing walk passes `False`; spy test proves zero lookups; wiring test proves the flag value; the widened mock at `tests/integration/test_steering.py:949` matches the real signature.

Two doc-wording nits, neither review-blocking, both routed to the post-approval DOCS stage: (1) `bridge/context.py:404` resolver docstring line "Returns None when the message carries no media" (also true for non-file media via the gate; `fetch_reply_chain`'s Returns line shares the imprecision); (2) `docs/features/reply-thread-context-hydration.md:46-70` and `docs/features/media-enrichment.md:103-110` still describe the pre-gate world (three states, "rendered, never dropped") — non-file media yielding no descriptor is a fourth outcome, and the routing walk now skips resolution entirely.

Verdict: APPROVED — blockers 0, tech_debt 0, nits 2 (DOCS-stage), confidence 0.92.

tomcounsell:

Round 3 at head `412db25bc` (plan-tick commit `5fde69e2d` follows, docs/plans only). The round-2 blocker class is closed, not narrowed.

- **Fail-closed, permanently:** the gate returns `None` before any Redis touch whenever `get_media_type(msg)` is None, which covers every non-file kind — current and future — even when an intake record with `has_media` and no path exists (the parametrized test saves exactly that shape). Wrong-suppression check: stickers and real documents carry a truthy `.document` and pass; web-preview documents belong to the preview and suppressing them is the round-1 fix. The one suppressed-but-real case is an expired-TTL document (`document=None`), which the intake download path also cannot resolve — no false-marker path remains in either direction.
- **Hop-skip cannot hide information:** `fetch_reply_chain` stores `msg.text or ""` unstripped, so any text-bearing hop (whitespace included) is truthy and renders; the `continue` fires only on empty content AND a `None` descriptor. What it newly hides — service-message hops and caption-less non-file hops — previously rendered a bare label or a false marker.
- **`resolve_media=False` is safe:** the Step-2 routing fallback consumes only `sender` and `message_id` (verified — no other field read); both hydration call sites keep the default `True`; no other production caller exists. The change strictly removes Redis-health-bounded latency from session routing.
- **Loop/timeout safety unchanged:** descriptor resolution stays under `asyncio.to_thread`; both hydration fetches remain inside `asyncio.wait_for(_REPLY_CHAIN_FETCH_TIMEOUT_S)`.
- **Cross-chat and path-disclosure invariants untouched:** `chat_id=str(chat_id)` scoping, `MEDIA_DIR` containment via `resolve().is_relative_to`, existence/readability stat, with the cross-chat and `/etc/hosts` tests unchanged.

Verdict: APPROVED — blockers 0, tech_debt 0, nits 0, confidence 0.90.

tomcounsell:
## Review: APPROVED

Multi-judge consensus (2/2 judges: code-quality, risk), rule `any-blocker-wins`, mean confidence 0.91, no disagreement. Round 3 at judged head `412db25bc5d6090b4e056e5a7209be25b5a03cff` against plan `docs/plans/reply-chain-media-renders-as-literal-string.md` (Closes #2732); a plan-checkbox tick commit `5fde69e2d` (docs/plans only, no code) was pushed before this review per the tick-then-post ordering.

All three round-2 findings are verified resolved: the deny-list is replaced by the prescribed classifier gate (`get_media_type(msg) is not None`, fail-closed for all current and future non-file media kinds, deny-list and imports deleted with zero residue), `fetch_reply_chain` gained `resolve_media: bool = True` with the session-routing walk passing `False`, and the caption-less excluded hop renders nothing rather than a bare `Sender:` line — pinned end-to-end by the GeoLive test.

Mechanical verification (reproduced in this environment — worktree venv, Python 3.14.6, Telethon 1.42.0):

- **222 tests green** across `tests/unit/test_context_helpers.py` (91), `tests/unit/test_bridge_context_guards.py`, and `tests/integration/test_steering.py`.
- **Mutation checks bite:** classifier gate neutralized → 11 red (9 parametrized non-file kinds + GeoLive e2e + the round-1 link-preview regression test); `resolve_media` flag ignored → the zero-lookup spy test red.
- **Stalled-lookup timing test genuinely reaches Redis:** the re-pointed `MessageMediaDocument` fixture passes the classifier gate, so the 5s-stall patch on `TelegramMessage.query.filter` is exercised; if the gate rejected the fixture, no `TimeoutError` would be raised and `pytest.raises` would fail — the test cannot pass vacuously.
- **Fixture re-pointing weakened nothing:** all 13 `media=object()` → genuine `MessageMediaDocument` conversions preserve every kind/reason/filename assertion; the 6 remaining `media=object()` sites are in `tests/unit/test_bridge_ack_steering_routed.py`, outside the resolver path and outside this PR.
- **Plan Verification table: 15/16 PASS** via `agent.verification_parser`. The single runner FAIL ("Feature docs updated") is the known #2570/#2836 table pipe-escaping collision — the row's `\|` unescapes to a literal `|` so grep searched for a literal alternation string; the intended BRE-alternation grep matches both files (output `2` > 1). Plan-authoring artifact, not a code defect.
- **Caller enumeration complete:** three `fetch_reply_chain` call sites — both hydration sites (`bridge/telegram_bridge.py:2089`, `:2604`) keep `resolve_media=True`; the routing walk (`bridge/context.py:761`) passes `False` and provably consumes only `sender`/`message_id`.
- `ruff check` and `ruff format --check` clean; `[media]` literal absent from `bridge/context.py`; off-loop, chat-scoping, and anti-criterion rows all pass.

## Rubric

- [x] **1. Plan vs. implementation match** — pass — *all 13 Success Criteria verified (16-row table 15 mechanical PASS + docs row verified under intended semantics); checkboxes ticked in `5fde69e2d`*
- [x] **2. New code quality** — pass — *classifier gate documented in place; names and comments accurate; no dead code or deny-list residue*
- [x] **3. Test coverage** — pass — *222 green; both mutation checks bite; new behavior (gate, flag, hop-skip) each pinned by dedicated tests*
- [x] **4. Regression risk to existing callers** — pass — *complete caller enumeration above; routing walk reads only sender/message_id; hydration sites unchanged*
- [x] **5. Data integrity** — n/a — *read-only use of existing `TelegramMessage` fields; no schema change, migrations correctly untouched*
- [x] **6. Security** — pass — *chat_id-scoped lookup, MEDIA_DIR containment, and path-disclosure posture unchanged by the patch*
- [x] **7. Documentation accuracy** — pass — *both feature docs updated per plan and accurate for file media; two wording nits (below) routed to the DOCS stage that runs next on approval*
- [x] **8. PR body accuracy** — pass — *load-bearing claims (verification rows, mutation-red, lint) reproduce at head; the "78 in test_context_helpers" count is stale (now 91) because review-driven patches added tests — accurate at time of writing, not misleading*
- [x] **9. Disclosed deferrals** — pass — *plan No-Gos defer the email seam to #3136 (verified OPEN)*
- [x] **10. Follow-up claims verified** — pass — *#3136 OPEN*

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — PASS — *13/13; see Rubric 1*
- **2. No-Gos from plan — none violated** — PASS — *email seam zero-diff row passes; `_cache_walk_root` untouched; no outbound download added*
- **3. New `except Exception` blocks — each has logger/raise/swallow-ok** — PASS — *resolver except logs at warning and degrades to an unreadable descriptor (round-1 shape, unchanged)*
- **4. New integration tests — exercise serialization boundary** — PASS — *exclusion and GeoLive tests save real `TelegramMessage` records through the ORM to the claimed test db*
- **7. New public APIs — docstrings present** — PASS — *`resolve_media` documented in the signature docstring; one wording nit noted below*
- **8. Breaking changes — migration path documented** — N/A — *keyword arg with compatible default*
- **9. Tests added for new behavior** — PASS — *9-kind parametrized exclusion, GeoLive e2e, zero-lookup spy, routing wiring test*
- **10. Tests cover the failure path (not just happy path)** — PASS — *exclusion tests use the production has_media/no-path record shape; all unreadable reasons covered*
- **11. UI changes (if any) — screenshot captured** — N/A — *no UI files in the diff*
- **12. Docs updated for user-facing changes** — PASS — *per plan; DOCS-stage handoff below*

### Blockers
- None

### Tech Debt
- None

### Nits
- **File:** `bridge/context.py:404` (verified: read this file)
  **Code:** `Returns None when the message carries no media.`
  **Issue:** The resolver also returns None for non-file media via the classifier gate; `fetch_reply_chain`'s Returns line ("a descriptor dict for hops carrying media, else None") shares the imprecision. Doc-wording only; the inline gate comment is accurate.
  **Severity:** nit
  **Fix:** Extend both lines, e.g. "no media, or media that is not a downloadable file". Routed to `/do-docs` (runs next on this approval) — not `/do-patch`.
- **File:** `docs/features/reply-thread-context-hydration.md:46` (verified: read this file)
  **Code:** `is rendered, never dropped` / the three-state enumeration
  **Issue:** Non-file media (link previews, polls, geo/live, venues, contacts, dice, games, invoices) now yields no descriptor — a fourth outcome — and a caption-less such hop renders nothing; `docs/features/media-enrichment.md:103-110` similarly implies one scoped Redis lookup per hop, while the routing walk now skips resolution entirely.
  **Severity:** nit
  **Fix:** Add the non-file exclusion and the `resolve_media=False` routing-walk behavior to both docs. Routed to `/do-docs` (runs next on this approval).

### Miscellaneous
- None

- **Email attachment delivery seam + renderer unification** — plan No-Gos, tracked by #3136 (OPEN)

### Review Delta (vs prior review on HEAD 2c1921a)
- **Resolved**: Blocker — deny-list fails open on non-file media kinds (replaced by the classifier gate; mutation-verified, 11 red when neutralized)
- **Resolved**: Tech debt — routing walk pays unread descriptor resolution (`resolve_media=False` from Step 2; spy + wiring tests)
- **Resolved**: Nit — bare `Sender:` line for excluded caption-less hops (generalized hop-skip; pinned by the GeoLive e2e test)
- **New**: two doc-wording nits (above), routed to the DOCS stage

### Verification Results

- 222 tests green on the PR branch (reproduced by this review in an isolated worktree venv)
- Mutation checks: gate neutralized → 11 red; flag ignored → spy red
- Plan Verification table 15/16 PASS; the 1 FAIL is the #2570/#2836 pipe-escaping artifact (intended check passes: both docs match)
- Lint: `ruff check` and `ruff format --check` clean

PR #2700, opened 2026-08-10
Recurrence guard: detect tests that derive their own Redis db= (#2655)

tomcounsell:
Closes #2655

Plan: `docs/plans/db-derivation-guard.md`
Feature doc: `docs/features/test-db-derivation-guard.md`

## Mechanism chosen, and the evidence for it

**An AST walk with inverted polarity.** The instrument was never the problem; the polarity was.

Every prior round (#2117, #2606, #2624) enumerated *accepted callee shapes*. An enumeration of
accepted shapes means everything unenumerated passes silently, and the next call site was always
written in a shape nobody had enumerated. So this guard **enumerates nothing about what is being
called**. It flags every `db=` keyword argument passed to *any* call anywhere under `tests/`, plus
every `from_url(...)` argument, and requires the **value** to match one of exactly two shapes:

```
S1  a direct Call whose TERMINAL name is claim_test_db / claim_scratch_test_db
    terminal name = func.id for ast.Name, func.attr for ast.Attribute

S2  an ast.Name bound EXACTLY ONCE in the enclosing function to an S1 call
```

Evidence that ignoring the callee name is free: an AST census of `tests/` found **all 17 `db=`
keyword arguments are Redis constructions**. Zero false positives today, and immunity to a
constructor name nobody has thought of yet. Mutation M5 below makes this concrete: restricting the
walk to callees literally named `Redis` — the old permit-list polarity — lets
`SomeFutureRedisLike(db=8)` through.

Shape S2 exists for one measured reason: without it the canonical fixture
(`test_db = claim_test_db()` then `redis.Redis(db=test_db)` at `tests/conftest.py:615,631`) needed an
exemption, which would teach every reader that the allowlist is the normal path. One hop, one scope,
single binding. It admits the canonical fixture and nothing else.

### Q1 — AST walk, or an inverted rule?

**Both, and the inversion is the load-bearing half.** AST is the only instrument that can see the
*structure* of a value (a regex cannot tell `db=claim_test_db()` from `db=claim + 1`). But the fix
for three rounds of failure is polarity, not coverage. Second-order evidence: inverting dissolves the
`from_url` scoping problem entirely, because an argument the guard cannot prove safe is a violation
by definition rather than a silent pass.

### Q2 — permit `_redis_test_db_num` by name, or delete it?

**Neither. The two call sites that feed it into `db=` are converted instead.**

Deleting the alias means editing `tests/conftest.py:656-658`, which is inside #2628's exclusive lock,
[...]

PR #3185, opened 2026-09-06
db-derivation guard: fix positional db, url= keyword, fixture shadowing

### LINKED ISSUE #2764: db-derivation guard: positional `db` and keyword `from_url(url=...)` produce no candidate
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/db-derivation-guard-positional-db-and-from-url-keyword.md

Follow-up from the #2700 review (round 6). Deferred rather than blocking: **measured live exposure is zero** in the tree today.

`tests/db_derivation_guard.py` replaced an enumeration of *callee names* with an enumeration of *argument-passing syntax*. That second enumeration has two mirror-image holes, both of which yield **no candidate at all** (not a violation, not an accepted site):

1. **Positional `db`.** `db` is the third positional parameter of `redis.Redis.__init__`, but route 1 iterates `node.keywords` only.
   ```python
   redis.Redis("localhost", 6379, 7)   # -> NO CANDIDATE
   ```

2. **Keyword `url` on `from_url`.** `from_url(url, **kwargs)` accepts `url` by keyword, but route 2 is gated on `node.args` being non-empty.
   ```python
   redis.Redis.from_url(url="redis://localhost:6379/9")   # -> NO CANDIDATE
   ```

Both verified against the merged implementation via `scan_source`.

## Why this was deferred, not blocked

Measured over `tests/**/*.py` at PR head `25e53671`:

- Sites with `Redis`/`StrictRedis` and >= 3 positional args: **0**
- Sites with `from_url(url=...)`: **0** (all 9 `from_url` sites pass the URL positionally)

So the guard's stated coverage is intact for every call site that exists. These are latent gaps that only open when someone writes a new site in one of the two shapes.

## Suggested fix

- Route 1: also inspect `node.args[2]` when the callee is `Redis`/`StrictRedis`.
- Route 2: treat a `url=` keyword on `from_url` the same as `node.args[0]`.
- Add both shapes to `tests/unit/test_db_derivation_guard.py`.
- Either way, disclose the residual gaps in the module docstring's "known cost" paragraph, which currently discloses only the opaque-splat callee scoping.

Refs #2655, #2700.

## Recon Summary

Re-verified against `main` at `67d714662` (2026-09-05), 23 days after filing.

**Confirmed:**

- Both gaps still reproduce in shipped source. Driven through `scan_source`: `redis.Redis("localhost", 6379, 7)` yields 0 candidates, `redis.StrictRedis("localhost", 6379, 7)` yields 0, `redis.Redis.from_url(url="redis://localhost:6379/9")` yields 0. Controls fire correctly: `redis.Redis(db=7)` yields 1 `db-kwarg` (ok=False, pool_db=7) and `redis.Redis.from_url("redis://localhost:6379/9")` yields 1 `from-url` (ok=False, pool_db=9).
- Root cause unchanged. Route 1 iterates `node.keywords` only (`tests/db_derivation_guard.py:520`); route 2 is gated on `if callee == "from_url" and node.args:` (`tests/db_derivation_guard.py:567`).
- Live exposure is still zero. Over `tests/**/*.py` at `67d714662`: `Redis`/`StrictRedis` with >= 3 positional args = 0; `from_url(...)` sites = 9, of which `url=` keyword = 0.
- The tree is clean today: 23 candidates, 7 violations, 0 undispositioned, 0 stale.
- The positional index is 2. Against installed `redis-py 7.4.0`, `Redis.__init__` parameters are `(self, host, port, db, password, socket_timeout, ...)`, `from_url` is `(url, **kwargs)`, and `redis.StrictRedis is redis.Redis`.
- The `CLAIM_FIXTURE_NAMES` leg does not launder. Probes `scratch_test_db = 7`, `test_db = 7`, and a `scratch_test_db=7` default argument each go red, as intended.

**Revised:**

- The disclosure target named in the suggested fix is wrong. The "known cost" paragraph lives in `_splat_candidate.__doc__` (around line 385), not in the module docstring. The plan must name the real location.
- Scope grows by two items folded in from #2768, which was closed as a duplicate of this issue on 2026-08-13: no criterion pins `format_violation()`'s remedial message content, and `REDIS_CONSTRUCTORS` is an undisclosed residual permit list at the splat layer.
- New finding, same surface: route 1 refuses the sanctioned fixture parameter when it is used directly. `redis.Redis(db=scratch_test_db)` returns ok=False with detail "no local binding in the enclosing function", while the aliased form `divergent_db = scratch_test_db; redis.Redis(db=divergent_db)` returns ok=True. The obvious shape is refused and the indirect one accepted. `CLAIM_FIXTURE_NAMES`'s docstring claims it "mirrors CLAIM_URL_NAMES / the `redis_test_url` leg of Route 2 exactly", but route 2 carries two legs (`ast.Call` and bare `ast.Name`) where route 1 carries only the alias hop. Live exposure is zero; it is a false positive waiting for the next author.

- New, found by the plan's propagation check and more serious than the three above because it fails in the safe-looking direction: route 2's bare-name leg launders a hardcoded pool URL. `redis_test_url = "redis://localhost:6379/9"` followed by `redis.Redis.from_url(redis_test_url)` returns ok=True with detail "claim-API URL fixture parameter". The leg accepts the sanctioned identifier and never asks whether a local assignment has shadowed it, even though `_LocalBindings` already records that fact. This is exactly the laundering that comment 5277517215 predicted would follow from an identifier match on a bare `ast.Name`, present in shipped code. Live exposure zero: 0 sites in `tests/` locally rebind any name in `CLAIM_URL_NAMES` or `CLAIM_FIXTURE_NAMES`.
- Consequence for the fix: route 2's leg is not a reference implementation to copy into route 1. Both legs need the condition comment 5277517215 specified, a sanctioned identifier **with no local rebinding**, and the plan lands them together.

**Pre-requisites:**

- None. `tests/db_claim.py` and the `CLAIM_*` frozensets are stable, and the two files in the blast radius (`tests/db_derivation_guard.py`, `tests/unit/test_db_derivation_guard.py`) have no other open lane against them.

**Dropped:**

- Nothing. No cited reference has drifted. #2655 closed and PR #2700 merged on 2026-08-13, and neither has changed the two routes since.

tomcounsell:
## Summary

Closes #2764. `tests/db_derivation_guard.py` inverted its polarity in #2700 to
judge every `db=`/`from_url(...)` value rather than enumerate callee names, but
that second enumeration (argument-passing *syntax*) had two mirror-image
holes, plus a third defect in the direct-fixture-parameter leg and a fourth,
more serious one in route 2's bare-name leg. All four produced **no candidate
at all** — not a violation, not an accepted site, simply unseen.

- **Fix 1** — Route 1 (`db=` keyword) gains a positional leg: `redis.Redis("h", 6379, 7)` now yields a candidate. Scoped to `Redis`/`StrictRedis` by terminal name (disclosed as a bounded exception), with the index pinned to `inspect.signature(redis.Redis.__init__)` by a test rather than hardcoded, and an explicit `ast.Starred` suppression so `redis.Redis("h", 6379, *rest)` doesn't produce a nonsense `expr="*rest"` violation.
- **Fix 2** — Route 2 (`from_url`) resolves the URL from a `url=` keyword when no positional argument is present.
- **Fix 3 / 3b** — A shared, *positive* `_is_unshadowed_fixture_parameter` predicate (built from `_parameter_names_without_defaults` + a node-sweep `_rebound_names`) replaces route 1's refusal of `db=scratch_test_db` used directly, and closes route 2's live laundering hole where `redis_test_url = "redis://localhost:6379/9"; from_url(redis_test_url)` was accepted on identifier match alone.
- **Fix 4** — `format_violation` gets an explicit per-kind branch; a `db-positional` candidate no longer renders as `from_url(...)`.
- **Fix 5** — Module docstring gains a "What this guard still cannot see" section naming all seven residual gaps and the `REDIS_CONSTRUCTORS`/bare-`str`-binder-list enumerations. A seventh gap -- a walrus in a `Lambda`'s own parameter default is invisible to `_rebound_names` -- was found during round-2 review and is disclosed rather than fixed here; tracked as a follow-up in #3192.
- **Fix 6** — A test-only signature tripwire re-derives the `db` positional index from the installed `redis-py`, so a future signature reshuffle turns the suite red instead of silently reopening the hole.

## Red-state proof (captured against `main` before this fix)

```
POSITIONAL 0   STRICT 0   URLKW 0   FIXTURE 1 (i.e. a violation)
URLLAUNDER 0   URLDEFAULT 0
```

`URLLAUNDER` and `URLDEFAULT` are the most important: they are **shipped code
accepting a hardcoded pool-slot URL as though it were the sanctioned fixture**.

After the fix, all six flip to their correct value (`POSITIONAL 1`, `STRICT 1`,
`URLKW 1`, `FIXTURE 0`, `URLLAUNDER 1`, `URLDEFAULT 1`), and every row in the
plan's ~62-row Verification table (leg-2, leg-3, leg-3-over-refusal, route-2,
route-2-over-refusal, Starred, one-hop-alias-disclosure, message-per-kind,
signature-pin, disclosure) was driven directly and matches its expected value.

## Mutation evidence

Each of the four production-code fixes was reverted individually, its own new
test(s) confirmed red, then restored and confirmed green:

1. Disabling the positional leg → `test_positional_db_on_redis_goes_red` /
   `..._strictredis_goes_red` fail (0 == 1).
2. Disabling the `url=` keyword resolution → `test_keyword_url_on_from_url_goes_red` fails.
3. Widening the shared predicate to an identifier-only match (the exact bug
   the plan is written to prevent) → 32 of 38 `test_shadowing_and_rebinding_shapes`
   rows fail immediately, including the two most important (`URLLAUNDER`,
   `URLDEFAULT`).
4. Reverting `format_violation` to its old two-way branch →
   `test_positional_message_is_not_mislabeled_as_from_url` and
   `test_an_unrecognized_kind_renders_its_own_name...` both fail.

## Test plan

- [x] `./scripts/pytest-clean.sh tests/unit/test_db_derivation_guard.py -q -n0` — 153 passed
- [x] Live tree: 0 undispositioned, 0 stale, both before and after
- [x] `python -m ruff check` / `ruff format --check` on both changed files
- [x] `tests/db_derivation_guard.py` still imports no `redis` (pure-AST); only the test file imports it for the signature tripwire
- [x] Mutation-checked all four production-code fixes individually (see above)


tomcounsell:

No blockers. The four fixes do what the PR says they do, and I reproduced every load-bearing number independently rather than taking the PR body's word for it (see Verification Results). Three tech-debt items and one nit remain, all in the disclosure/test-hygiene layer rather than in the guard's judgment.

**Independent red-state control.** I built a detached worktree at `origin/main` and drove the same six probes: `POSITIONAL 0`, `STRICT 0`, `URLKW 0`, `FIXTURE 1 violation`, `URLLAUNDER 0 violations`, `URLDEFAULT 0 violations`. At PR head all six flip to `1 / 1 / 1 / 0 / 1 / 1`. The PR body's red-state table is accurate, including the two that matter most: route 2's bare-name leg really was accepting a hardcoded pool-slot URL on `main`.

**Independent mutation check.** I did not rely on the PR's mutation claims. Eight mutations, each reverted after measuring:

| Mutation | Result |
|---|---|
| Positional leg's callee set emptied | 8 failed |
| `url=` keyword resolution removed | 4 failed |
| Accept predicate widened to identifier-only | 34 failed |
| Leg 3 (`_rebound_names`) check dropped | 24 failed |
| `format_violation` reverted to the two-way branch | 2 failed |
| Comprehension carve-out widened to whole subtrees | 3 failed |
| `ast.Starred` suppression removed | `STARRED` candidates 0 → 1 (asserted by `test_starred_positional_yields_no_candidate`) |
| No-default (leg 2) check dropped | `LDEFAULT`/`URLDEFAULT` violations 1 → 0 (asserted by four rows) |

Every guard bites. This is not a suite that passes without reaching the code.

**Live-tree control.** 23 candidates / 7 violations / 0 undispositioned / 0 stale at both `main` and PR head — identical. The route-2 narrowing is not merely synthetic: six real sites (`integration/test_valor_email.py:30`, `unit/test_email_bridge.py:1175`, `unit/test_email_history.py:34`, `unit/test_email_relay.py:34`, `unit/test_send_message.py:29`, `unit/test_valor_email.py:47`) now pass through `_is_unshadowed_fixture_parameter` and all stay green.

## Rubric

- [x] **1. Plan vs. implementation match** — pass — *All six fixes present. Success Criteria driven directly: 64/64 Verification rows pass, including the `L3COMP`/`L3COMPWALRUS` and `URLL3COMP`/`URLL3COMPWALRUS` pairs that separate the narrow carve-out from the over-broad one. No-Go respected: route 2 gains no `_resolve_one_hop` leg, and `URLHOP`/`DBHOP` pin the disclosed false positive.*
- [x] **2. New code quality** — pass — *Type hints throughout; the two new helpers are pure and separately testable; the positive-predicate rationale is written down at the definition rather than left to be re-derived. No new `try`/`except`.*
- [x] **3. Test coverage** — pass — *146 passed. Both new helpers have direct unit tests asserting return sets, not only end-to-end violation counts.*
- [x] **4. Regression risk to existing callers** — pass — *`tests/unit/test_db_derivation_guard.py` is the sole importer of the guard module; nothing under `bridge/`, `worker/`, `agent/`, or `tools/` references it. Live tree unchanged.*
- [x] **5. Data integrity** — n/a — *No models, no migrations, no schema.*
- [x] **6. Security** — pass — *No `eval`/`exec`/`subprocess`/`mark_safe`, no secrets. The one behavior change in the accept direction (fix 3) is bounded by three legs, each with its own falsifying row.*
- [ ] **7. Documentation accuracy** — fail — *The module docstring states the positional leg scopes itself to `REDIS_CONSTRUCTORS`; it does not. See Tech Debt 1.*
- [x] **8. PR body accuracy** — pass — *Every claim reproduced: 146 passed, 0 undispositioned / 0 stale before and after, ruff check + format clean, guard module imports no `redis`, and the four mutation claims (I ran eight).*
- [x] **9. Disclosed deferrals** — pass — *The one deferral (route 2's missing one-hop alias leg) is an explicit plan No-Go with sound rationale, disclosed as residual gap 6 and pinned by a test so it cannot close by accident.*
- [x] **10. Follow-up claims verified** — n/a — *The PR body makes no `filed as follow-up #N` claims.*

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — PASS — *64/64 Verification rows executed on the PR branch.*
- **2. No-Gos from plan — none violated** — PASS — *Route 2 has no `_resolve_one_hop` leg; `URLHOP 1` against `DBHOP 0` pins it.*
- **3. New `except Exception` blocks — each has logger/raise/swallow-ok** — N/A — *None added; `scan_source` still lets `SyntaxError` propagate.*
- **4. New integration tests — exercise serialization boundary** — N/A — *Pure-AST unit tests; no serialization boundary.*
- **5. Plan internal consistency — spike findings match task steps** — PASS — *Spike-5/6/7 findings map onto fixes 3/3b and the leg-2 vs leg-3 row split.*
- **7. New public APIs — docstrings present** — PASS — *All three new module-level helpers carry substantive docstrings.*
- **8. Breaking changes — migration path documented** — N/A — *Test-internal; the only fleet effect is a suite that fails on a newly-written offending site, which is the intent.*
- **9. Tests added for new behavior** — PASS — *146 tests, up from the pre-PR file.*
- **10. Tests cover the failure path (not just happy path)** — PASS — *Empty/invalid inputs, `Redis()`, `Redis('h')`, `from_url()` with neither form, `from_url(**kw)`, and both starred shapes are all asserted.*
- **11. UI changes (if any) — screenshot captured** — N/A — *Diff touches two `tests/` Python files; no HTML/CSS/JS/template.*
- **12. Docs updated for user-facing changes** — PASS — *The plan's Documentation section argues against a `docs/features/` file and puts the disclosure in the module docstring; all seven inline-documentation items landed. Accuracy of one of them is Tech Debt 1.*

### Blockers
- None

### Tech Debt

- [ ] **Tech Debt 1**
  **File:** `tests/db_derivation_guard.py:83`
  **Code:** `REDIS_CONSTRUCTORS` (below) is the residual permit list the opaque-splat leg **and the positional leg** both scope themselves to
  **Issue:** The positional leg does not scope itself to `REDIS_CONSTRUCTORS`. Line 807 reads `callee in {"Redis", "StrictRedis"}` — an inline literal — and `REDIS_CONSTRUCTORS`'s own comment at line 199 still says it is "Used ONLY to scope the opaque-`**`-splat leg". Two statements in the same file contradict each other and the new one is false. It also makes gap 2 understate the gap it discloses: `from_url` is inside `REDIS_CONSTRUCTORS` yet is equally unread by the positional leg, so "outside `REDIS_CONSTRUCTORS`" is not the boundary. Inert today, but this is the module's own residual-gap section, whose entire value is being literally true.
  **Severity:** tech_debt
  **Fix:** Either point the positional leg at a named constant and say which one, or reword lines 83-84 and gap 2 to state that the positional leg carries its own `Redis`/`StrictRedis` scoping, separate from `REDIS_CONSTRUCTORS`. Leave line 199's "ONLY" intact if the second route is taken.

- [ ] **Tech Debt 2**
  **File:** `tests/db_derivation_guard.py:520-522`
  **Code:** `elif isinstance(child, (ast.FunctionDef, ast.AsyncFunctionDef, ast.ClassDef)):` … `continue  # its body/args are a new scope; do not sweep them`
  **Issue:** `_rebound_names` skips the whole nested `def`/`class`/`lambda` node, but only its *body* is a new scope. Default-argument expressions, decorators, and class bases execute in the **enclosing** scope, so a binding written there is missed and the parameter is accepted as unshadowed. Measured at head:
  ```
  def t(scratch_test_db):
      def inner(x=(scratch_test_db := 7)):
          return x
      redis.Redis(db=scratch_test_db)     # -> ok=True, "claim-API fixture parameter"
  ```
  The interpreter confirms the rebinding is real: the same function with `return p` returns `7` for `t(3)`. The lambda-default, class-base, and decorator spellings launder identically. The plan's Technical Approach specifies "do not descend into a nested Lambda, FunctionDef, AsyncFunctionDef or ClassDef **body**" — the implementation skips more than the body, in the accept direction, and the residual-gap section does not name it. Live exposure is zero and the spelling is pathological, which is why this is tech debt and not a blocker; but leg 3's stated bound is "rebound by no binding form", and this is a binding form it does not see.
  **Severity:** tech_debt
  **Fix:** Before `continue`, sweep the nested node's out-of-scope parts — `decorator_list`, `args.defaults` + `args.kw_defaults`, `returns`, and a `ClassDef`'s `bases`/`keywords` — through `walk_own_scope`, so a `NamedExpr` there is collected. If that is judged more machinery than the shape deserves, add it as residual gap 7 instead; what should not stand is an undisclosed accept-direction gap in the one section written to enumerate them.

- [ ] **Tech Debt 3**
  **File:** `tests/unit/test_db_derivation_guard.py:92`
  **Code:** `def test_guard_sees_a_non_zero_number_of_candidates(capsys):` with docstring `"""A guard with nothing to check is indistinguishable from a passing one."""`
  **Issue:** The plan's Test Impact section calls for this test to be updated to "extend the floor assertions to cover `db-positional` without requiring a live site … State that in the test's docstring so a future reader does not 'fix' it by asserting a floor that can never be met." The test is byte-identical to `main`: no `db-positional` bucket, no docstring note. The anti-vacuity test is precisely where that warning belongs, since the tree has zero positional sites and always will until someone writes one. The neighbouring `test_no_stale_disposition_entries` item is in the same position: it stays green as required, but the plan's explicit "must not silently absorb one either — `_matches` is kind-agnostic, so a new-kind candidate on the same `(path, expr)` could consume an entry; asserted explicitly rather than assumed" has no corresponding assertion.
  **Severity:** tech_debt
  **Fix:** Add a `db-positional` bucket to the candidate breakdown with a floor of 0 and a docstring sentence saying the floor is 0 by design and that non-vacuity for that kind comes from `PLANTED_OFFENDERS`. Separately, add the explicit non-absorption assertion the plan asks for, or amend the plan item if it is judged unnecessary.

### Nits

- [ ] **Nit 1**
  **File:** `tests/unit/test_db_derivation_guard.py:1162`
  **Code:** `source = Path("tests/db_derivation_guard.py").read_text()`
  **Issue:** A cwd-relative path in a test file that already imports `TESTS_ROOT` from the module under test. Run from anywhere but the repo root this raises `FileNotFoundError` rather than testing anything. It fails loudly rather than silently, hence a nit.
  **Severity:** nit
  **Fix:** `source = (TESTS_ROOT / "db_derivation_guard.py").read_text()`.

### Miscellaneous
- None (Tech Debt 2 and 3 are recorded above rather than duplicated here)

- **Route 2 gains no `_resolve_one_hop` leg** — an explicit plan No-Go, not a silent omission. Disclosed as residual gap 6 in the module docstring, pinned by `test_route2_one_hop_alias_stays_a_disclosed_false_positive` (`URLHOP 1` against the route-1 control `DBHOP 0`), and the failure direction is loud. Verified as written, not merely claimed.
- **The bare-`str` binder list in `_rebound_names` is hand-enumerated with no grammar-level completeness tripwire** — disclosed in the module docstring under "Accepted, disclosed limitation", with the PEP 695 exposure measured rather than assumed.

### Verification Results

`agent.verification_parser` over the plan's `## Verification` table, run on the PR branch:

```
checks parsed: 64   malformed: 0   skipped: 0
All checks passed.
```

Independently re-run outside the table: `./scripts/pytest-clean.sh tests/unit/test_db_derivation_guard.py -q -n0` → **146 passed in 20.29s**; `ruff check` and `ruff format --check` on both changed files → clean (ruff 0.15.9).

### Screenshots

### Reviewer note

Multi-judge consensus (`code-quality` + `risk`) could not run: this reviewer surface exposes no subagent-dispatch tool, so per Hard Rule 9 the judges were not dispatched rather than backgrounded and abandoned. This is a single-judge verdict and is labelled as one.

### Restart-after-deploy
Not required. The diff is confined to `tests/`; no `bridge/`, `worker/`, or `agent/` code, no new env var, no migration.

tomcounsell:
Addressed all four findings from the CHANGES REQUESTED review at b78ea1e7b:

1. **Tech debt — REDIS_CONSTRUCTORS comment/implementation disagreement.** The positional leg is deliberately scoped to a narrower `{"Redis", "StrictRedis"}` literal, not `REDIS_CONSTRUCTORS` — `from_url`'s signature (`url, **kwargs`) has no positional `db` slot, so including it there would invent a shape that can't exist at runtime. Rewrote all three sites (module-docstring gap-2, the "residual permit list" paragraph, and left the accurate "used ONLY for the splat leg" comment on the constant itself) so they agree on that division.

2. **Tech debt — real accept-direction gap in `_rebound_names`.** Confirmed: `def inner(x=(scratch_test_db := 7))` was invisible to the rebinding sweep because a nested `def`/`class` skipped its whole node, when only its *body* is a new scope. Defaults, decorators, and (for `ClassDef`) bases/keywords evaluate in the enclosing scope at definition time. `_rebound_names` now recurses into a nested function's `args.defaults`/`args.kw_defaults`/`decorator_list` and a nested class's `bases`/`keywords`/`decorator_list`, while still skipping `body`/`args`. Added `L3NESTEDDEFAULT` and its route-2 mirror `URLL3NESTEDDEFAULT`, plus four direct `_rebound_names` unit tests (default walrus, decorator walrus, class-base walrus, class-keyword walrus). Mutation-checked: reverting the recursion drops all six new tests to red (`0 == 1` on the labeled rows, `'p' in {'inner'}`/`{'C'}` failures on the direct tests); restoring returns them to green, full file green at 153 passed.

3. **Tech debt — missing Test Impact item.** Added a `db-positional` candidate bucket to `test_guard_sees_a_non_zero_number_of_candidates` with a docstring note explaining its floor is deliberately 0 (no live positional site exists; non-vacuity comes from the planted/labeled rows, not the tree) so a future reader doesn't "fix" it by raising the floor. Added the sibling non-absorption assertion (`test_no_disposition_entry_is_silently_absorbed_by_a_new_kind`), mirroring `test_no_stale_disposition_entries`'s orphan check from the other direction: every `ALLOWLIST`/`DEFERRED` entry still covers a candidate, and none of the covered candidates carry a kind #2764 introduced.

4. **Nit.** `Path("tests/db_derivation_guard.py")` → `TESTS_ROOT / "db_derivation_guard.py"`.

**Verification:**
- All 64 Verification-table rows re-run at the new head: 64/64 pass, including the guard suite, `ruff check`, and `ruff format --check`.
- Live-tree measurement unchanged: 23 candidates, 7 violations, 0 undispositioned, 0 stale — identical before and after.
- Scoped suite: `./scripts/pytest-clean.sh tests/unit/test_db_derivation_guard.py -q -n0` → **153 passed**.

tomcounsell:

Round 2, at head `b78ea1e7b`. Three of the four round-1 findings are fully resolved and pinned by tests that bite. The fourth — the accept-direction gap in `_rebound_names` — is resolved for the `def`-default, decorator, class-base and class-keyword spellings but **not for the lambda-default spelling that the same finding named**, and the docstring the patch rewrote now asserts a boundary that is false for `Lambda`. One tech-debt item and one nit.

I re-derived every number rather than reading them off the PR body or the patch comment.

**Round-2 mutation check (my own, on an isolated worktree at PR head).** Each of the three new production hunks reverted individually, measured, restored:

| Mutation | Result |
|---|---|
| Nested-`def` `args.defaults`/`kw_defaults` recursion removed | 3 failed — `L3NESTEDDEFAULT`, `URLL3NESTEDDEFAULT`, `test_nested_def_default_walrus_rebinds` |
| Nested-`def` `decorator_list` recursion removed | 1 failed — `test_nested_def_decorator_walrus_rebinds` |
| `ClassDef` reverted to a bare skip | 2 failed — `test_nested_class_base_walrus_rebinds`, `test_nested_class_keyword_walrus_rebinds` |

Every new guard reaches the code it claims to guard.

**One round-1 suggestion I checked and am withdrawing.** The round-1 fix text asked for `returns` and parameter annotations to be swept alongside the defaults. They should not be, and the narrower implementation here is correct: Python rejects a walrus inside an annotation outright (`SyntaxError: named expression cannot be used within an annotation`), and the walrus is the only expression-level binder that could appear there. Sweeping annotations would have been dead code. Verified in the interpreter, not assumed.

**Live-tree control.** 23 candidates (14 `db=` keyword, 9 `from_url`, 0 `db-positional`), 7 violations, all 7 covered by a disposition, 0 undispositioned, 0 stale — unchanged from round 1.

## Rubric

- [x] **1. Plan vs. implementation match** — pass — *All six fixes present; the No-Go (route 2 gains no `_resolve_one_hop` leg) is respected and `URLHOP`/`DBHOP` still pin it. 63 of 64 Verification rows executed green; the 64th failed only because my scratch worktree had no `.venv`, and passed once provisioned.*
- [x] **2. New code quality** — pass — *Type hints throughout; the new recursion is four small, symmetric blocks with the definition-time rationale written at the site. The remaining lambda hole is a disclosure/completeness defect rather than a quality defect in what was written, so it is scored under item 7.*
- [x] **3. Test coverage** — pass — *153 passed (up from 146). The four new `TestReboundNames` cases assert the returned set directly, not only end-to-end counts, and each one goes red under its own mutation.*
- [x] **4. Regression risk to existing callers** — pass — *`tests/unit/test_db_derivation_guard.py` remains the sole importer of the guard module. The round-2 change moves only in the refusal direction (more names collected), which is the safe direction, and the live tree is byte-identical in its verdicts.*
- [x] **5. Data integrity** — n/a — *No models, migrations, or schema.*
- [x] **6. Security** — pass — *No `eval`/`exec`/`subprocess`/`mark_safe`, no secrets. Round 2 adds no accept surface.*
- [ ] **7. Documentation accuracy** — fail — *The rewritten `_rebound_names` docstring says the sweep does not descend into a nested `Lambda`/`FunctionDef`/`AsyncFunctionDef`/`ClassDef` **BODY**, and then explains that only the body belongs to the new scope. For `Lambda` the code skips the whole node, defaults included, and the branch comment says "anonymous; nothing binds in the outer scope" — which is false for a lambda default. The residual-gap section does not name it either. See Tech Debt 1.*
- [ ] **8. PR body accuracy** — fail — *The Test plan still checks off "146 passed"; head is 153. See Nit 1.*
- [x] **9. Disclosed deferrals** — pass — *The one deferral (route 2's one-hop alias leg) is an explicit plan No-Go, disclosed as residual gap 6 and pinned by `test_route2_one_hop_alias_stays_a_disclosed_false_positive`.*
- [x] **10. Follow-up claims verified** — n/a — *No `filed as follow-up #N` claims in the PR body.*

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — PASS — *64 rows driven on the PR branch; 64/64 green after provisioning a venv for the scratch worktree.*
- **2. No-Gos from plan — none violated** — PASS — *`URLHOP 1` against `DBHOP 0` still pins the deferred leg.*
- **4. New integration tests — exercise serialization boundary** — N/A — *Pure-AST unit tests.*
- **5. Plan internal consistency — spike findings match task steps** — PASS.
- **7. New public APIs — docstrings present** — PASS — *No new public helpers in round 2; the existing three keep substantive docstrings. Accuracy of one is Tech Debt 1.*
- **8. Breaking changes — migration path documented** — N/A — *Test-internal.*
- **9. Tests added for new behavior** — PASS — *Six new tests for the round-2 fix, plus the two disposition-hygiene tests round 1 asked for.*
- **10. Tests cover the failure path (not just happy path)** — PASS — *Both the rebind (red) and over-refusal (green) directions are asserted for each new binder position.*
- **11. UI changes (if any) — screenshot captured** — N/A — *Two `tests/` Python files; no HTML/CSS/JS/template in the diff.*
- **12. Docs updated for user-facing changes** — PASS — *Documentation lives in the module docstring by the plan's design; the round-2 rewrite landed. Its accuracy is Tech Debt 1.*

### Blockers
- None

### Tech Debt

- [ ] **Tech Debt 1 — the lambda-default spelling still launders, and the docstring now says otherwise**
  **File:** `tests/db_derivation_guard.py:535-536`
  **Code:** `elif isinstance(child, ast.Lambda):` / `continue  # anonymous; nothing binds in the outer scope`
  **Issue:** A lambda's defaults evaluate at definition time in the **enclosing** scope, exactly like the `def` defaults this patch just fixed. The `Lambda` branch still skips the whole node, so a walrus in a lambda default is invisible to the rebinding sweep and the shadowed parameter is accepted as the sanctioned fixture. Measured at head:

  ```
  def t(scratch_test_db):
      f = lambda x=(scratch_test_db := 7): x
      redis.Redis(db=scratch_test_db)          # -> ok=True, "claim-API fixture parameter"

  def t(redis_test_url):
      f = lambda x=(redis_test_url := "redis://localhost:6379/9"): x
      redis.Redis.from_url(redis_test_url)     # -> ok=True, "claim-API URL fixture parameter"
  ```

  `_rebound_names` returns `{'f'}` for both, against `{'inner', 'scratch_test_db'}` for the structurally identical `def`-default form the patch fixed. The interpreter confirms the rebinding is real: the same function returning the parameter yields `7` for `t(3)`.

  Two things make this worth another round rather than leaving it. First, round 1's finding named this spelling explicitly ("the lambda-default, class-base, and decorator spellings launder identically") and the patch comment reports the finding as addressed; two of the three named spellings were fixed. Second, `_rebound_names` does not exist on `origin/main` — this is new surface shipping with the hole, not an inherited one, so the residual-gap section that enumerates this module's boundaries is wrong on the day it lands.

  The branch comment "anonymous; nothing binds in the outer scope" is the specific false statement: the lambda's *name* does not bind, but its defaults do. The docstring's "does not descend into a nested `Lambda`/.../`ClassDef` **BODY**" reads, after the round-2 rewrite that distinguishes body from defaults, as a promise that non-body positions are swept — which holds for `def` and `class` and not for `Lambda`.

  Live exposure is zero (no lambda-default walrus anywhere under `tests/`) and the spelling is pathological, which is why this is tech debt and not a blocker.
  **Severity:** tech_debt
  **Fix:** Mirror the `def` branch — before `continue`, run `walk_own_scope` over `child.args.defaults` and the non-`None` `child.args.kw_defaults`, and correct the branch comment to say the lambda's *name* is what does not bind. (Its `decorator_list` and `returns` do not exist, and per the withdrawal above annotations cannot carry a binder.) Add an `L3LAMBDADEFAULT` row and its `URLL3LAMBDADEFAULT` mirror alongside `L3NESTEDDEFAULT`, and a direct `TestReboundNames` case, so the fix is pinned the same way the other three are. If the four lines are judged more machinery than the shape deserves, the alternative is equally acceptable: name it as residual gap 7 and fix the `Lambda` branch comment. What should not stand is an undisclosed accept-direction gap inside the section written to enumerate them.

### Nits

- [ ] **Nit 1**
  **File:** PR body, Test plan
  **Code:** ``- [x] `./scripts/pytest-clean.sh tests/unit/test_db_derivation_guard.py -q -n0` — 146 passed``
  **Issue:** Stale at head. The round-2 commit adds seven tests; I measured **153 passed in 125.82s**. The count is superseded by the patch comment, but the PR body is the artifact the merge gate and a future reader read first, and a checked box carrying a wrong number is the kind of claim Hard Rule 10 exists for.
  **Severity:** nit
  **Fix:** Update the Test plan line to 153, and note the two extra mutation checks (six production hunks now, not four) in the Mutation evidence section.

### Miscellaneous
- None

- **Route 2 gains no `_resolve_one_hop` leg** — an explicit plan No-Go with recorded rationale, disclosed as residual gap 6 and pinned by `test_route2_one_hop_alias_stays_a_disclosed_false_positive` (`URLHOP 1` against the route-1 control `DBHOP 0`). Verified as written, not merely claimed. No tracking issue is required; the plan states a successor issue is filed only if someone hits it.
- **The bare-`str` binder list in `_rebound_names` has no grammar-level completeness tripwire** — disclosed in the module docstring under "Accepted, disclosed limitation", with the PEP 695 exposure measured rather than assumed.
- **`db-positional`'s live-tree floor is 0** — newly disclosed in `test_guard_sees_a_non_zero_number_of_candidates`'s docstring with the reason and an explicit warning against "fixing" it upward. Non-vacuity for that kind comes from the labeled rows, which I confirmed do exercise it.

### Review Delta (vs prior review on HEAD `cb807ba`)

- **Resolved — Tech Debt 1 (`REDIS_CONSTRUCTORS` disagreement).** All three sites now agree: gap 2 names `{"Redis", "StrictRedis"}` and says why `from_url` cannot join that leg (`(url, **kwargs)` has no positional `db` slot), the permit-list paragraph scopes itself to the splat leg only, and line 204's "Used ONLY to scope the opaque-`**`-splat leg" is now true as written.
- **Partially resolved — Tech Debt 2 (`_rebound_names` accept-direction gap).** The `def`-default, decorator, class-base and class-keyword spellings are fixed, disclosed in the docstring, and pinned by six tests that I mutation-checked individually. The lambda-default spelling, named in the same finding, is unchanged — carried forward as Tech Debt 1 above.
- **Resolved — Tech Debt 3 (missing Test Impact items).** `db-positional` bucket added with a floor of 0 and the docstring warning the plan asked for; `test_no_disposition_entry_is_silently_absorbed_by_a_new_kind` added as the mirror of the orphan check. Its orphan leg is non-vacuous (4 entries covering 7 violations); its absorption leg is correctly vacuous today, which is what a tripwire looks like before it fires.
- **Resolved — Nit 1.** `Path("tests/db_derivation_guard.py")` → `TESTS_ROOT / "db_derivation_guard.py"`.
- **New — Nit 1 above** (stale test count in the PR body).

### Verification Results

`agent.verification_parser` over the plan's `## Verification` table, on the PR branch:

```
checks parsed: 64   malformed: 0   skipped: 0
63 passed
1 failed — "Guard suite passes", environment-only:
  pytest-clean: refusing to run — worktree has no usable .venv of its own.
```

That row failed because my isolated review worktree had no `.venv`, not because of anything in the diff. After provisioning one, re-run directly: `./scripts/pytest-clean.sh tests/unit/test_db_derivation_guard.py -q -n0` → **153 passed in 125.82s**. `ruff check` and `ruff format --check` on both changed files → clean (ruff 0.15.9).

### Screenshots

### Reviewer note

Single-judge verdict, labelled as one. Multi-judge consensus (`code-quality` + `risk`) did not run: this reviewer surface exposes no subagent-dispatch tool, so per Hard Rule 9 the judges were not dispatched rather than backgrounded and abandoned.

### Restart-after-deploy
Not required. The diff is confined to `tests/`; no `bridge/`, `worker/`, or `agent/` code, no new env var, no migration.

tomcounsell:
## Review: APPROVED — round 2 final (supersedes the earlier round-2 comment)

Head reviewed: `b78ea1e7bcca5756c554f93ce0ee207d30bf1d16` (resolved through `tools/pr_head_resolver.py::resolve_pr_head_sha`).

This supersedes the CHANGES REQUESTED comment posted earlier at this same head. I re-derived every measurement in a dedicated review worktree at `b78ea1e7b` and reached **0 blockers**. The one substantive item is a real but zero-exposure completeness gap in a test-only guard, which is tech debt, not a blocker. Recording the correct verdict rather than holding an otherwise-sound change on a nit.

### Round-1 findings — closure

**(a) `REDIS_CONSTRUCTORS` comment/implementation disagreement — CLOSED.**
`tests/db_derivation_guard.py:846` reads `callee in {"Redis", "StrictRedis"}`. The module docstring at lines 57 and 90 and the inline comment at 842 all now state the positional leg is deliberately scoped to that narrower literal rather than `REDIS_CONSTRUCTORS` (which stays `{"Redis", "StrictRedis", "from_url"}` at line 208, correct for the splat layer). All three sites agree.

**(b) `_rebound_names` accept-direction gap — CLOSED for `def`/`class`, one spelling remains (see Tech Debt).**
Measured directly against `_rebound_names` at head:

| shape | result |
|---|---|
| nested `def` positional default walrus | CAUGHT |
| nested `def` kw-only default walrus | CAUGHT |
| nested `def` decorator walrus | CAUGHT |
| `ClassDef` base walrus | CAUGHT |
| `ClassDef` keyword (`metaclass=`) walrus | CAUGHT |
| `ClassDef` decorator walrus | CAUGHT |
| nested `def` **body** assignment | not swept (correct) |
| `ClassDef` **body** assignment | not swept (correct) |

The body is still not descended into, and that is pinned in the other direction: descending into `child.body` turns `L3NESTASSIGN`, `URLL3NESTASSIGN` and `TestReboundNames::test_assignment_inside_nested_def_does_not_rebind` red.

New rows `L3NESTEDDEFAULT` (`tests/unit/test_db_derivation_guard.py:930`) and `URLL3NESTEDDEFAULT` (`:1022`) are present, as are the four direct tests at `:1153`, `:1160`, `:1164`, `:1168`.

**(c) Missing Test Impact — CLOSED.**
`test_guard_sees_a_non_zero_number_of_candidates` (`:95`) now buckets `db-positional` (`:111`), prints it in the capsys line, asserts `== 0`, and the docstring (`:98-107`) explains the 0 floor is deliberate and tells a future reader not to raise it. `test_no_disposition_entry_is_silently_absorbed_by_a_new_kind` is at `:131`.

**(d) Nit: cwd-relative path — CLOSED.**
`tests/unit/test_db_derivation_guard.py:1249` is `(TESTS_ROOT / "db_derivation_guard.py").read_text()`. No `Path("tests...` literals remain in either file.

### Mutation checks on the patch's own new code

Each mutation applied on an isolated worktree, run, then reverted.

| mutation | tests red |
|---|---|
| A — drop recursion into `FunctionDef` `args.defaults` / `kw_defaults` / `decorator_list` | 4: `L3NESTEDDEFAULT`, `URLL3NESTEDDEFAULT`, `test_nested_def_default_walrus_rebinds`, `test_nested_def_decorator_walrus_rebinds` |
| B — drop recursion into `ClassDef` `bases` / `keywords` / `decorator_list` | 2: `test_nested_class_base_walrus_rebinds`, `test_nested_class_keyword_walrus_rebinds` |
| C — wrongly descend into a nested `def`'s `body` | 3: `L3NESTASSIGN`, `URLL3NESTASSIGN`, `test_assignment_inside_nested_def_does_not_rebind` |
| D — relabel the live `db-kwarg` emit site (`:830`) as `db-positional` | 1, and only 1: `test_no_disposition_entry_is_silently_absorbed_by_a_new_kind`, failing on its own absorption assertion at `:158`, not on the orphan loop it shares with `test_no_stale_disposition_entries` (which passed) |

Mutation D answers the non-vacuity question the patch comment raised. The set it asserts over is non-empty independently: `covered` holds 7 candidates at head. A first attempt at the same mutation on the emit site at `:757` did **not** bite, because that site produces nothing in the live tree — worth knowing, since it means `:757` carries no live coverage from this test either way.

Restored to head: `git diff --quiet` clean, suite green.

### Regression check

- Plan Verification table at `b78ea1e7b`: **58/58** executable `python -c` rows pass. Of the 4 non-`python -c` rows: the kind-contract `grep -c` returns 1 as expected, the guard suite passes, `ruff format --check` reports "2 files already formatted". The lint row was not run (standing instruction on this machine: formatting only).
- Scoped suite: `./scripts/pytest-clean.sh tests/unit/test_db_derivation_guard.py -q -n0` → **153 passed**.
- Live tree: **23 candidates** (14 `db=` keyword, 9 `from_url`, 0 `db-positional`) / **7 violations** / **0 undispositioned** / **0 stale** — identical to round 1.

### Tech debt (1) — does not block

**`_rebound_names` still skips a `Lambda` node whole, so a walrus in a lambda default is invisible.**
`tests/db_derivation_guard.py:838` is `elif isinstance(child, ast.Lambda): continue`. A lambda's defaults evaluate in the enclosing scope exactly as a `def`'s do, which the interpreter confirms:

```
def f(scratch_test_db=1):
    g = lambda x=(scratch_test_db := 99): x
    return scratch_test_db        # -> 99
```

`_rebound_names` returns a set without `scratch_test_db` for that source, so `_is_unshadowed_fixture_parameter` accepts the shadowed parameter on both routes. The lambda *body* case is handled correctly and must keep being skipped (`lambda: (scratch_test_db := 99)` leaves the outer at 1) — only `args.defaults` / `args.kw_defaults` are missing.

The docstring's phrasing "Does **not** descend into a nested `Lambda`/`FunctionDef`/`AsyncFunctionDef`/`ClassDef` BODY" reads as if `Lambda` were handled the same way as the other three; it is not, and the comment `# anonymous; nothing binds in the outer scope` is true of the lambda's *name* but not of its defaults.

Not a blocker: this is an accept-direction over-permission in a test-time guard, the refuse direction is unaffected, the spelling occurs zero times in the tree, and the same class of gap for `def`/`class` was itself graded tech debt in round 1. Worth a follow-up issue rather than another patch round — the fix is three lines mirroring the `FunctionDef` leg.

### Nit (1)

PR body line 51 checks off "146 passed"; the suite is 153 passed at this head. Stale from the pre-patch run.

### Withdrawn from round 1

Round 1's fix text also asked that `returns` and parameter annotations be swept. Python rejects a walrus inside an annotation outright and the walrus is the only expression-level binder that could appear there, so the narrower implementation is correct and sweeping them would be dead code. Withdrawn.

tomcounsell:
## Review: Approved

Round 3, at head `37ffaa165` (my own plan-checkbox-sync commit on top of `5e5d8c3fa`, per the APPROVED-path plan-sync step). **0 blockers, 0 tech debt, 0 nits.**

This round exists because the round-2 APPROVED verdict was recorded against `b78ea1e7b` and the DOCS stage then added `5e5d8c3fa`, moving the head out from under it. So the first question is what that commit actually changed, and the answer is nothing executable:

```
AST_IDENTICAL_MODULO_DOCSTRINGS: True     (b78ea1e7b vs 5e5d8c3fa, tests/db_derivation_guard.py)
COMPARATOR_BITES: True                    (seeded X=1 -> comparator reports a difference)
tests/unit/test_db_derivation_guard.py:   unchanged between the two SHAs
```

I compared parsed ASTs with docstrings stripped rather than diffing stripped source, and seeded a fake statement to prove the comparator can fail. `5e5d8c3fa` is docstring and comment text only: residual gap 7 added to "What this guard still cannot see", the `_rebound_names` docstring corrected to stop implying `Lambda` is swept like `def`/`class`, and the `# anonymous; nothing binds in the outer scope` comment replaced with an accurate one.

I did not inherit round 2's measurements on the strength of that proof alone. Every load-bearing number below was re-derived in a dedicated review worktree at this head.

## Rubric

- [x] **1. Plan vs. implementation match** — pass — all 22 Success Criteria verified at this head and ticked in the plan; the one plan No-Go (route 2 gains no `_resolve_one_hop` leg) is pinned by `URLHOP`/`DBHOP` and disclosed as gap 6.
- [x] **2. New code quality** — pass — type hints throughout, positive-predicate design in `_is_unshadowed_fixture_parameter` with the rejected negative formulation documented, no copy-paste; `_judge_db_value` is shared by both route-1 legs specifically so the two judgments cannot drift.
- [x] **3. Test coverage** — pass — 153 tests pass; the 64-row Verification table drives each named probe directly rather than through a wrapper.
- [x] **4. Regression risk to existing callers** — pass — `tests/db_derivation_guard.py` is imported only by its own unit test and by pytest collection; no `bridge/`, `worker/`, `agent/` or `tools/` importer. Live tree unchanged at 0 undispositioned / 0 stale.
- [x] **5. Data integrity** — n/a — no models, no migrations, no schema.
- [x] **6. Security** — pass — pure-AST module, no `eval`/`exec`/`subprocess`, no secrets, no network. The module still imports no `redis`; only the test does, for the signature tripwire.
- [x] **7. Documentation accuracy** — pass — module docstring, `_rebound_names`, `_splat_candidate`, `CLAIM_FIXTURE_NAMES`, `Candidate.kind` and `REDIS_DB_POSITIONAL_INDEX` all match the shipped behavior. No stale references to removed code.
- [x] **8. PR body accuracy** — pass — every claim in the body reproduced independently below, including the two the round-2 review flagged as stale (`153 passed`, and the seven-gap count).
- [x] **9. Disclosed deferrals** — pass — see Acknowledged Deferrals; every one is either an OPEN tracking issue or an explicit plan No-Go with recorded rationale.
- [x] **10. Follow-up claims verified** — pass — `#3192` resolves to an OPEN issue titled "db-derivation guard: `_rebound_names` misses a walrus in a lambda default", labeled `bug`/`testing`.

## Pre-Verdict Checklist

- [x] Checked out the PR branch before reading any file (dedicated detached worktree at the head SHA, sole ownership, so no author edit could race my measurements)
- [x] Read the plan and the tracking issue before the diff
- [x] Ran the plan's `## Verification` table on the PR branch — 64 checks, 0 malformed, 0 skipped, all pass
- [x] Reproduced every number the PR body claims rather than crediting it
- [x] Every finding cites a file I read and code I saw
- [x] No unguarded `except Exception` introduced — the one deliberate exception behavior is `scan_source` propagating `SyntaxError`, documented as intentional
- [x] Tests are not in-memory-only stand-ins — the guard's tests drive the real scanner over real source strings and over the live `tests/` tree
- [x] New public functions carry docstrings
- [x] Plan section compliance — `## Documentation`, `## Update System`, `## Agent Integration`, `## Test Impact` all present
- [x] Docs gate — the plan's `## Documentation` section explicitly specifies **no** `docs/features/` file, with rationale (the module docstring is the source of truth; a thin feature doc would be a second place to drift). Gate satisfied as specified.
- [x] No new env vars, no new migrations, no `MIGRATIONS` registration needed
- [x] Restart-after-deploy not required — diff is confined to `tests/`

## Independent verification

**Red-state control.** I built the same six probes the PR body claims, ran them against a detached `origin/main` worktree and against this head, in the same process and environment. Same probe file, two `sys.path` entries — the one-line control:

| probe | main | this head | PR claims |
|---|---|---|---|
| `redis.Redis("h", 6379, 7)` | 0 | 1 | 0 → 1 |
| `redis.StrictRedis("h", 6379, 7)` | 0 | 1 | 0 → 1 |
| `from_url(url="redis://localhost:6379/9")` | 0 | 1 | 0 → 1 |
| `Redis(db=scratch_test_db)` on a real fixture param | 1 (false positive) | 0 | 1 → 0 |
| `redis_test_url = "redis://…/9"; from_url(redis_test_url)` | 0 | 1 | 0 → 1 |
| `def t(redis_test_url="redis://…/9"): from_url(redis_test_url)` | 0 | 1 | 0 → 1 |

Every cell matches. The last two are the ones that matter: on `main`, shipped code accepts a hardcoded pool-slot URL as though it were the sanctioned fixture, on identifier match alone. That is closed here.

**Widened probes** beyond what the PR and plan assert, to test the claim rather than the example:

```
STARRED (*rest at the index)              0 candidates   (gap 5, as disclosed)
KWONLY / POSONLY / async fixture param    accepted       (all three spellings)
Redis("h", 6379, scratch_test_db)         accepted       (positional leg reaches the fixture accept)
Redis("h", 6379, claim_test_db())         accepted
from_url("redis://h:6379/9?decode=1")     violation      (querystring URL -> "db cannot be determined", safe direction)
Redis("h", 6379, 7, **kw)                 2 violations   (positional + splat compose, neither swallows the other)
from_url("…/0", url="…/9")                1 violation    (positional wins, single candidate)
module-scope Redis("h", 6379, 7)          violation      (no enclosing function needed)
lambda BODY walrus                        accepted       (correct — body walrus is scope-local)
```

**Live tree at this head:** 23 candidates, 7 violations, **0 undispositioned, 0 stale**, `check_dispositions()` returns `[]`.

**Suite:** `./scripts/pytest-clean.sh tests/unit/test_db_derivation_guard.py -q -n0` → **153 passed in 91.25s**, matching the PR body exactly. (First attempt aborted on test-DB pool exhaustion — 91 live pytest processes from sibling agents, `reap-xdist.sh` found no orphans. Environmental, retried.)

**Lint/format:** `ruff check` and `ruff format --check` on both changed files → clean, 2 files already formatted.

**Verification table:** 64 checks parsed, 0 malformed, 0 skipped, **64 pass**. Round 2 reported 63/64 with the guard-suite row failing on a worktree with no `.venv`; provisioning one closes that row, so the table now runs complete.

**Mutation check on the surface that is new at this head.** The round-2 review mutation-checked the production code at `b78ea1e7b`, and I proved above that this head's production code is that code. The one thing genuinely new here is the gap-7 disclosure text, so that is what I mutated: deleting the entire gap-7 block from the docstring leaves `test_residual_gaps_are_disclosed_in_the_module_docstring` green and the corresponding Verification row printing `DISCLOSE True`. Restored, `git diff --quiet` clean.

That is a real property of the test, and I considered raising it. I am not, and the reason is worth stating rather than leaving implicit: the test asserts the section exists and names `REDIS_CONSTRUCTORS`, which is what the plan asked it to assert. Pinning gap 7's prose specifically would create churn the day #3192 lands and deletes it. Recorded here so a future reader knows the disclosure section is section-level pinned, not item-level pinned.

## Blockers
- None

## Tech Debt
- None

## Nits
- None

## Miscellaneous
- None

## Acknowledged Deferrals (verified)

- **Residual gap 7 — a walrus in a nested `Lambda`'s parameter default is invisible to `_rebound_names`.** Verified real, not merely claimed: `lambda x=(scratch_test_db := 99): x` over a genuine fixture parameter yields 0 violations at this head where `def inner(x=(scratch_test_db := 7))` yields 1. Accept-direction only, zero live occurrences under `tests/`. Tracked by **#3192 (OPEN)**, disclosed in the module docstring and at the `Lambda` branch itself.
- **Residual gap 5 is narrower than the positional leg's actual `ast.Starred` blindness.** New this round, and the only thing I found that the seven-gap enumeration does not already name. `redis.Redis(*hp, 7)` yields **0 candidates**: `node.args` is `[Starred, Constant(7)]`, length 2, so the fixed-index `len(node.args) > REDIS_DB_POSITIONAL_INDEX` guard short-circuits before the leg looks at anything. The db literal is in plain sight and outside the starred unpack, so gap 5's prose ("a `db` arriving *inside* a starred unpack at the positional index", attributed to fix 1's `Starred` suppression) does not cover it — different mechanism, same family. Exposure measured by AST over the whole tree: **0 of 14** `Redis`/`StrictRedis` call sites carry an `ast.Starred` positional at all. Accept-direction only. Filed as **#3193 (OPEN)** with both a disclose-only and a fix option and the measurement attached, rather than reopening this diff for a shape nobody writes — the same disposition #3192 received for the identical grade of gap.
- **Route 2 gains no `_resolve_one_hop` leg** — an explicit plan No-Go with recorded rationale, disclosed as gap 6, pinned by `URLHOP` at 1 against the route-1 control `DBHOP` at 0. Verified as written.
- **`_rebound_names`'s bare-`str` binder list is hand-enumerated with no grammar-level completeness tripwire** — disclosed in the module docstring under "Accepted, disclosed limitation", with the PEP 695 `TypeVar`/`ParamSpec`/`TypeVarTuple` exposure measured rather than assumed.
- **`db-positional`'s live-tree floor is 0** — disclosed in `test_guard_sees_a_non_zero_number_of_candidates`'s docstring with an explicit warning against "fixing" it upward. Non-vacuity for that kind comes from the labeled Verification rows, which I confirmed do exercise it.

## Review Delta (vs the round-2 final review on HEAD `b78ea1e7b`)

- **Resolved — round-2 Tech Debt 1 (undisclosed `Lambda` gap).** The round-2 review offered two acceptable remedies and the DOCS commit took the second: name it as residual gap 7 and correct the `Lambda` branch comment. Both landed, and #3192 tracks the three-line fix. The specific false statement round 2 named — `# anonymous; nothing binds in the outer scope` — is gone.
- **Resolved — round-2 Nit 1 (stale test count in the PR body).** Body now reads `153 passed`; I measured 153.
- **Resolved — the docstring's "does not descend into a nested `Lambda`/`FunctionDef`/… BODY" phrasing**, which round 2 correctly read as implying `Lambda` is handled like the other three. It now says `FunctionDef`/`AsyncFunctionDef`/`ClassDef` only, with `Lambda` called out separately.
- **New — Acknowledged Deferral 2** (gap 5's scope), filed as #3193 rather than held against this PR.
- **Verification table** goes 63/64 → **64/64**; the one round-2 failure was a missing worktree `.venv`, not the diff.

## Plan checkbox sync

All 22 `## Success Criteria` items ticked in `docs/plans/db-derivation-guard-positional-db-and-from-url-keyword.md` and pushed as `37ffaa165` **before** this review was posted, so the review is strictly newer than every commit on the branch and `/do-merge`'s freshness gate sees it as current.

One criterion deserves its basis stated rather than assumed. *"Each of the six fixes is mutation-checked individually"* is ticked on combined evidence: the round-2 review's per-mutation tables (A–D, each reverted, run, restored) against `b78ea1e7b`, plus my AST proof that this head's production code **is** that code, plus my own mutation of the only surface added since. I did not personally re-run all six reverts, and would not have ticked it without the inertness proof.

## Screenshots

N/A — no HTML, CSS, JS/TS, JSX/TSX, Vue or template files in the diff. The visual-proof gate is a no-op for this PR.

## Restart after deploy

Not required. The diff touches `tests/` only — no `bridge/`, `worker/` or `agent/` code, no new env var, no migration, no `pyproject.toml` entry point. Fleet effect after `/update` is limited to the intended one: a machine whose `tests/` grows a call site in one of the three newly-visible shapes now fails the unit suite where it previously passed silently.

## Reviewer note

Single-judge verdict, labelled as one. Multi-judge consensus (`code-quality` + `risk`) did not run: this reviewer surface exposes no subagent-dispatch tool, so per Hard Rule 9 the judges were not dispatched rather than backgrounded and abandoned.

Issue #3166, opened 2026-09-05
Reliability risk: bridge — unattributed SIGTERM restart loop, nine boots at ~111s cadence, self-terminating


valorengels:
## Investigation Finding

**Component:** bridge (launchd service `com.valor.bridge` on "Valor the Cowboy")
**Detected:** 2026-09-05 (while diagnosing a `release verify ... bridge unknown` warning; see PR #3165)
**Status:** Observed-once, self-terminating, unattributed

## Symptoms

The bridge restarted nine times on a near-constant ~111s cadence between 02:52:21Z and 03:08:58Z. Every cycle was a clean, externally-initiated shutdown: `Received SIGTERM, shutting down gracefully...` → `_graceful_shutdown` → launchd relaunch → full boot (Telegram connect, catchup scan, knowledge watcher scan, `cleanup_stale_branches`) → SIGTERM again about 105s later. The loop stopped on its own after 03:08:58Z and has not recurred.

Nothing in the tree explains the SIGTERM. I could not attribute it.

## Diagnostic Output

`Connected to Telegram` boot timestamps from `logs/bridge.log` (note the 90-minute gap before the loop, then the fixed cadence):

```
2026-09-04T11:21:06.857612Z
2026-09-05T02:52:21.914552Z   <- loop starts
2026-09-05T02:54:12.281363Z
2026-09-05T02:55:56.587891Z
2026-09-05T02:57:49.602190Z
2026-09-05T02:59:45.013975Z
2026-09-05T03:01:36.449491Z
2026-09-05T03:03:28.276218Z
2026-09-05T03:05:17.987629Z
2026-09-05T03:07:07.529402Z
2026-09-05T03:08:58.518453Z   <- last boot; still alive after
```

Deltas: 111, 104, 113, 116, 111, 112, 109, 110, 111 seconds. That regularity is the strongest single clue — this is a periodic actor, not a crash.

One shutdown, in full (nothing anomalous precedes it; the process is idle and healthy):

```
03:08:40.814873Z tools.knowledge.indexer full_scan  Full scan complete: 0 indexed, 434 skipped, 0 errors
03:08:43.018478Z bridge.routing _resolve_config_path [config] Ignoring Desktop PROJECTS_CONFIG_PATH ...
03:08:44.892204Z telethon._disconnect               Disconnection from 91.108.56.122:443/TcpFull complete!
03:08:52.327131Z __main__ _shutdown_handler         Received SIGTERM, shutting down gracefully...
[...]

PR #3188, opened 2026-09-06
Auto-preserve refuses to commit a wipe (#3167)

### LINKED ISSUE #3167: Auto-preserve-before-teardown commits a half-deleted worktree: 902,840 deletions landed on a session branch as WIP
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/auto-preserve-teardown-half-deleted-worktree.md

## Investigation Finding

**Component:** `agent/worktree_manager.py::preserve_uncommitted_worktree_changes` (the producer of `WIP: auto-preserved before teardown` commits)
**Detected:** 2026-09-05 on `session/dev-a4e15370` ("Valor the Cowboy")
**Status:** Confirmed — reproduced in the branch history, not hypothetical

## Symptoms

The auto-preserve path committed a worktree that had already been half-deleted. Commit `ba3ea72e9` on `session/dev-a4e15370`:

```
3914 files changed, 223142 insertions(+), 902840 deletions(-)
```

`tests/`, `bridge/`, `agent/`, `config/` and `docs/` were removed from the branch head. Three further no-op `WIP: auto-preserved before teardown` commits (`805983e36`, `c004ff37d`, `35c76c461`) stacked on top of the wipe, each preserving nothing but two `.pyc` files, because by then the wipe was the committed state and the tree was "clean" relative to it.

The damage is silent. `git merge-base --is-ancestor origin/main HEAD` still passes — the branch is strictly *ahead* of main, it just deletes most of it. I only noticed because `sed` could not find a test file I had just been reading, and `git ls-files | wc -l` returned 1491 against main's several thousand.

## Diagnostic Output

The producer, `agent/worktree_manager.py:1406-1500` — note there is no check between the `status --porcelain` non-empty test and `git add -A`:

```python
if not status.stdout.strip():
    # Clean tree — nothing to preserve.
    return {"preserved": False, "was_clean": True, "ref": ref, "errors": []}

add = subprocess.run(
    ["git", "-C", str(worktree_dir), "add", "-A"], ...
)
...
commit = subprocess.run([
    "git", "-C", str(worktree_dir), "commit", "--no-verify", "--no-gpg-sign",
    "-m", f"WIP: auto-preserved before teardown [{slug}] [{ts}]",
], ...)
```

`git add -A` does not distinguish "the human edited files" from "the files are gone". A worktree mid-removal is maximally dirty, so the guard that is supposed to skip a clean tree instead waves through a total deletion, and `--no-verify` bypasses every commit hook that might have caught it.

The caller that fires on this path, `agent/worktree_manager.py:940-948` (`_cleanup_stale_worktree`):

```python
    # Preserve any uncommitted work before the direct force-remove (#2137).
    stale_slug = wt.name
    if VALID_SLUG_RE.match(stale_slug):
        preserve_uncommitted_worktree_changes(repo_root, stale_slug, wt)
    subprocess.run(["git", "worktree", "remove", "--force", worktree_path], ...)
```

Timing correlation with the bridge restart loop (filed separately as the unattributed-SIGTERM investigation). Each bridge boot runs `cleanup_stale_branches`; the WIP commits land ~40s after each boot:

```
bridge boot 03:03:28Z  ->  WIP commit 03:04:08Z
bridge boot 03:05:17Z  ->  WIP commit 03:05:58Z
bridge boot 03:07:07Z  ->  WIP commit 03:07:48Z
bridge boot 03:08:58Z  ->  WIP commit 03:09:38Z
```

## Steps Attempted

No fix attempted — filing only, per instruction.

Recovery performed on the affected branch: saved the in-flight diff, `git reset --hard origin/main`, reapplied, force-pushed with `--force-with-lease`. **No content was lost** — the wipe commits contained nothing but deletions, so discarding them was the complete recovery. The branch now carries only the intended change (PR #3165).

## Impact

This is a data-loss mechanism in the code whose entire purpose is preventing data loss.

- A worktree caught mid-teardown gets its deletion **committed and pushed** to the session branch, and the same deletion is written to `refs/session-wip/{slug}` — so the "recovery ref" preserves the destruction rather than the work.
- The failure is silent and survives review-by-glance: the branch still fast-forwards from main, `git status` is clean, and the commit subject reads like a routine safety net.
- Worse, it is self-reinforcing. Once the wipe is committed, subsequent preserve passes see a clean tree and stack no-op commits, burying the destructive commit under innocuous ones.
- Any restart loop in the bridge (see the companion investigation issue) amplifies directly into this: nine boots meant nine teardown passes against live session worktrees in seventeen minutes.
- Had the session branch been merged, or had a human resumed via `git reset --soft HEAD~1` as the docstring instructs, the deletion becomes the working state.

## Next Steps

- [ ] Add a deletion-ratio refusal to `preserve_uncommitted_worktree_changes`: compute `git diff --cached --numstat` (or `--shortstat`) after `git add -A` and **abort the commit** when the staged change is overwhelmingly deletions (e.g. deletions exceed a threshold count AND insertions are a negligible fraction). Return `{"preserved": False, "errors": [...]}` and log loudly rather than committing. Threshold should be a named, env-overridable constant marked provisional.
- [ ] Decide the sharper variant: a worktree missing its own tracked top-level directories is *definitionally* not "uncommitted work" — consider refusing when any of `tests/`, `bridge/`, `agent/`, `config/` are absent from disk while tracked at HEAD, which is a cheaper and less tunable signal than a ratio.
- [ ] Check ordering in `_cleanup_stale_worktree`: confirm whether the deletion is caused by a *prior partial* `git worktree remove --force` (i.e. preserve is running against a tree the previous pass already gutted) — if so, the preserve call is being made too late in the sequence and the fix belongs there too.
- [ ] Audit `refs/session-wip/*` on this machine for other refs that preserve a wipe rather than work; the docstring promises `git checkout refs/session-wip/{slug}` as recovery, and that promise is currently unsound.
- [ ] Add a regression test with a fixture worktree whose tracked files are deleted on disk, asserting the preserve call refuses and returns `preserved: False`.

## Recon Summary

Performed 2026-09-05 at `origin/main` = `bc42055a4`. No commits have touched `agent/worktree_manager.py` or `agent/session_revival.py` since this issue was filed, so every cited line number below was re-derived and is exact.

**Confirmed**

- `preserve_uncommitted_worktree_changes` is defined at `agent/worktree_manager.py:1406`. Between the `status --porcelain` non-empty test and `git add -A` there is no inspection of *what* is dirty. Confirmed by reading the body: `status` → early-return-if-empty → `add -A` → `commit --no-verify --no-gpg-sign` → `update-ref refs/session-wip/{slug}`.
- `_cleanup_stale_worktree` calls preserve at `agent/worktree_manager.py:948`, immediately before `git worktree remove --force`.
- The failure is silent for the reason stated: `--no-verify` bypasses every commit hook, and the resulting branch still fast-forwards from main.
- **Spike (empirical, throwaway repo):** in a worktree whose tracked `tests/`, `bridge/`, `agent/` directories were deleted on disk, `git status --porcelain` reports `" D <path>"` for every file, and after `git add -A` `git diff --cached --shortstat` reports `9 files changed, 9 deletions(-)` with zero insertions. Both proposed refusal signals are computable and cheap.
- **Spike:** `git ls-tree --name-only -d HEAD` enumerates the tracked top-level directories directly, so the "missing tracked directory" signal needs no hardcoded `tests/ bridge/ agent/ config/` list and stays correct in any repo.

**Revised**

- **The timing attribution in "Diagnostic Output" is wrong.** `cleanup_stale_branches` (`agent/session_revival.py:193`) never reaches preserve. It lists `session/*` branches, checks age, and calls `safe_delete_branch` — it does not touch worktrees. The `bridge boot → WIP commit` correlation is real but the branch sweep is not the mechanism.
- **The producing call sites are two, not one.** `_cleanup_stale_worktree` (`:948`) has exactly one caller: `create_worktree` at `agent/worktree_manager.py:1323`, reached only when `_find_worktree_for_branch` finds `session/{slug}` checked out at a `.worktrees/` path *different* from the expected one. The second, unmentioned producer is `remove_worktree` at `agent/worktree_manager.py:1660`, which preserves before its own `git worktree remove --force`. A fix scoped to `_cleanup_stale_worktree` alone would leave `remove_worktree` exposed. The guard belongs inside `preserve_uncommitted_worktree_changes` so both call sites inherit it.
- **Secondary defect found during recon: slug/branch mismatch.** `_cleanup_stale_worktree` passes `stale_slug = wt.name` — the *foreign* directory name — while the WIP commit lands on whatever branch that worktree has checked out, which is `session/{slug}` for the slug being created. So `refs/session-wip/{wt.name}` can point at a commit made on `session/{other-slug}`. The recovery pointer is misfiled even when preservation is legitimate.
- **The check must run before `git add -A`, not after.** `git status --porcelain` already carries the deletion/insertion distinction, so the refusal can be computed with zero index mutation. Computing it after `add -A` (as the Next Steps bullet proposes) means a refusal leaves a fully staged wipe in the worktree index, which is worse than the status quo whenever the subsequent force-remove fails and the directory survives.
- **The evidence commits are no longer resolvable.** `ba3ea72e9`, `805983e36`, `c004ff37d`, `35c76c461` do not exist in this checkout — the reporter's `--force-with-lease` recovery removed them. The defect is confirmed by code reading and by the spike above, not by re-inspecting those objects.

**Pre-requisites**

- None blocking. #3166 (bridge SIGTERM restart loop) is the amplifier that turned one latent race into nine teardown passes in seventeen minutes, but the guard here is independently correct and does not wait on it.
- The existing test harness in `tests/unit/worktree_manager/test_worktree_manager_uncommitted.py` (`_init_git_repo`, `_add_linked_worktree`, `_dirty`) already builds real git repos with linked worktrees, so the regression fixture is an addition to that file rather than new scaffolding.

**Dropped**

- **The `refs/session-wip/*` audit is a no-op here.** `git for-each-ref refs/session-wip/` returns zero refs on this machine. There is nothing to audit and nothing to ship for it; it stays a one-shot operator check, not a feature.
- **Root-causing *why* the tree was gutted is out of scope for the fix.** The candidates (a partial `shutil.rmtree` in the `_cleanup_stale_worktree` fallback at `:978`, a `git worktree remove --force` that deleted files then failed, or a TOCTOU race between two concurrent teardown passes) all produce the same observable, and the refusal guard is correct against all three. Narrowing it further is investigation, not a prerequisite.

tomcounsell:
## Summary
`preserve_uncommitted_worktree_changes` (`agent/worktree_manager.py`) used to test only whether `git status --porcelain` was non-empty, then unconditionally `git add -A` and commit. A worktree caught mid-teardown (tracked directories deleted from disk) reads as maximally dirty, so the guard meant to skip a clean tree instead waved a total deletion straight through — the reported incident committed 902,840 deletions onto a live session branch. This PR adds a structural wipe check before staging, and fixes a related ref/branch-name mismatch found during recon.

## Changes
- **Wipe detection**: `git ls-tree --name-only -d HEAD` lists directories HEAD tracks; if any is absent from disk on the worktree, the tree is a wipe, not a tree whose deletions should be committed. No hardcoded directory list, no threshold, no new configuration.
- **Additive-only preserve on a wipe**: `git reset -q` (index-recovery, load-bearing — recovers from a wipe a previous teardown pass already staged) followed by staging exactly the additive candidate set (`ls-files --modified --others --exclude-standard` minus `ls-files --deleted`), via stdin with `--pathspec-file-nul` — never `git add -A`, which would restage the declined deletions. Coexisting real work is preserved with **zero deletions** in the commit; a pure wipe writes no commit and no ref at all, leaving the branch head unmoved.
- **Commit trailer**: the additive commit's subject stays byte-identical to an ordinary preserve; a trailer (`Auto-preserve declined deletions (#3167)`, `Missing-tracked-dirs:`, `Declined-deletions:`) records the refusal in the durable artifact, not only the rotating log.
- **Asymmetric fail-open**: a failure reading the wipe signal itself falls open to today's behavior (logged at WARNING, `[worktree-wip-guard-failed]`); a failure *after* a missing directory is confirmed refuses instead of falling through (also logged at ERROR, `[worktree-wip-refused-wipe]`) — falling open there would commit the very wipe just detected.
- **Ref-slug fix**: `refs/session-wip/{name}` now follows the worktree's actual checked-out branch (`session/<name>` → `<name>`) rather than the foreign directory name `_cleanup_stale_worktree` passes as `slug`, so the recovery pointer and the recovered content agree. Falls back to the `slug` argument on a detached HEAD or non-`session/` branch.
- Rejected the issue's proposed insertions/deletions-ratio predicate: the incident's own commit carried 223,142 insertions alongside 902,840 deletions (untracked build artifacts staged by `add -A` count as insertions), so a ratio test would not have fired. Pinned by a regression test.

## Testing
- [x] 12 new regression tests in `tests/unit/worktree_manager/test_worktree_manager_uncommitted.py` (pure wipe, mixed wipe, untracked artifacts, partially-staged wipe, pre-staged wipe, deletions inside a surviving directory, both fail-open/fail-closed asymmetry cases, ref-slug follows branch, detached-HEAD fallback, errors/refused discriminator)
- [x] All 5 pre-existing tests in the same file pass unmodified except the strengthened `errors`/`refused` assertion
- [x] `tests/unit/worktree_manager/` — 149 passed
- [x] Representative mutation checks run manually (disabling the structural check, replacing additive staging with `add -A`, removing the `reset -q`) each flip the corresponding test red
- [x] Lint/format clean on all changed files (`python -m ruff check` / `ruff format --check`)
- [ ] Full `tests/unit/` sweep could not complete during this build — this machine had ~10+ concurrent SDLC lanes running pytest-xdist simultaneously, exhausting the shared test-DB pool (`#2628`) and the xdist worker pool. The scoped, plan-mandated verification (`tests/unit/worktree_manager/`) passed cleanly and repeatedly under that same contention; the change touches exactly one function, its own test file, and one doc file.

## Documentation
- [x] `docs/features/session-isolation.md` updated: the wipe check (both branches, fail-open asymmetry, ref-naming fix) documented in the auto-WIP-commit mechanism, and the `refs/stash` paragraph corrected to match the function's own (accurate) docstring — the doc previously had it backwards.

## Definition of Done
- [x] Built: wipe detection, additive-only preserve, and ref-slug fix all implemented in `preserve_uncommitted_worktree_changes`
- [x] Tested: 149 tests passing in `tests/unit/worktree_manager/`
- [x] Documented: `docs/features/session-isolation.md` updated

Closes #3167

### DIFF
--- a/agent/worktree_manager.py
+++ b/agent/worktree_manager.py
@@ -1733,8 +1733,10 @@ def preserve_uncommitted_worktree_changes(repo_root: Path, slug: str, worktree_d
           "deleted_paths": <int>, "ref": ..., "errors": []}``. No commit, no
           ref write, branch head unmoved.
         - A failure salvaging a detected wipe (block B) returns the pure-wipe
-          shape above plus ``"guard_error": <str>``, still with ``"errors":
-          []`` — the outcome is a refusal, not a broken-git failure.
+          shape above minus ``"deleted_paths"`` (the salvage step failed
+          before that count could be determined) plus ``"guard_error":
+          <str>``, still with ``"errors": []`` — the outcome is a refusal,
+          not a broken-git failure.
         - Git subprocess failure (the pre-existing contract): ``{"preserved":
           False, "was_clean": False, "ref": ..., "errors": [<msg>, ...]}``,
           with no ``"refused"`` key. ``"refused"`` is therefore the sole
@@ -1769,13 +1771,28 @@ def preserve_uncommitted_worktree_changes(repo_root: Path, slug: str, worktree_d
         # `rev-parse --abbrev-ref HEAD` returns the literal string "HEAD",
         # which passes VALID_SLUG_RE and would otherwise produce
         # `refs/session-wip/HEAD`.
-        branch_result = subprocess.run(
-            ["git", "-C", str(worktree_dir), "rev-parse", "--abbrev-ref", "HEAD"],
-            capture_output=True,
-            text=True,
-            timeout=settings.timeouts.git_subprocess_s,
-        )
-        branch = branch_result.stdout.strip() if branch_result.returncode == 0 else ""
+        try:
+            branch_result = subprocess.run(
+                ["git", "-C", str(worktree_dir), "rev-parse", "--abbrev-ref", "HEAD"],
+                capture_output=True,
+                text=True,
+                timeout=settings.timeouts.git_subprocess_s,
+            )
+            branch = branch_result.stdout.strip() if branch_result.returncode == 0 else ""
+        except Exception as branch_exc:
+            # A non-zero rc already falls back to the `slug` default above; an
+            # exception (e.g. TimeoutExpired) must degrade the same way rather
+            # than propagate to the outer handler, which would skip the
+            # ordinary path's `add -A` + WIP commit + ref write entirely and
+            # lose legitimate uncommitted work (#3167 follow-up).
+            logger.warning(
+                "[worktree-wip-branch-resolve-failed] slug=%s worktree=%s error=%s — "
+                "falling back to slug-derived ref.",
+                slug,
+                worktree_dir,
+                branch_exc,
+            )
+            branch = ""
         session_match = re.match(r"^session/(.+)$", branch) if branch else None
         if session_match and VALID_SLUG_RE.match(session_match.group(1)):
             ref = f"refs/session-wip/{session_match.group(1)}"
--- a/agent/worktree_manager.py
+++ b/agent/worktree_manager.py
@@ -1629,9 +1629,9 @@ def preserve_uncommitted_worktree_changes(repo_root: Path, slug: str, worktree_d
     Session worktrees under ``.worktrees/{slug}`` are force-removed on session
     exit (normal, exception, or cancellation). The unmerged-branch guard (#1646)
     protects only *committed* work; staged, unstaged, and untracked edits are
-    otherwise discarded with no backstop. This helper captures ALL uncommitted
-    work as a WIP commit on ``session/{slug}`` plus a durable named ref
-    ``refs/session-wip/{slug}`` before any ``git worktree remove --force``.
+    otherwise discarded with no backstop. This helper captures uncommitted work
+    as a WIP commit on the worktree's checked-out branch plus a durable named
+    ref ``refs/session-wip/{name}`` before any ``git worktree remove --force``.
 
     Mechanism (WIP commit + named ref, NOT ``git stash``). The reason is NOT
     that a worktree's stash is worktree-local: ``refs/stash`` lives in the
@@ -1642,37 +1642,104 @@ def preserve_uncommitted_worktree_changes(repo_root: Path, slug: str, worktree_d
     is meaningless and a teardown backstop keyed on it would race every peer
     (issue #2650, shape 1). A WIP commit under a slug-scoped named ref is
     single-owner by construction, and it captures untracked files, which
-    ``git stash`` declines by default. Instead:
+    ``git stash`` declines by default.
 
     1. ``git -C <worktree> status --porcelain`` — if empty, no-op.
-    2. ``git -C <worktree> add -A`` — captures untracked + tracked edits.
-    3. ``git -C <worktree> commit --no-verify --no-gpg-sign`` — the WIP commit
+    2. **Wipe check (#3167).** ``git -C <worktree> ls-tree --name-only -d HEAD``
+       lists the directories HEAD tracks; if any is absent from disk, the tree
+       is a wipe (partially or fully deleted from under the worktree, e.g. by a
+       racing teardown pass or a `shutil.rmtree` that raised partway through),
+       not a tree whose deletions should be committed. This check runs BEFORE
+       staging: the signal is available without mutating the index, and after
+       ``git add -A`` the deletion signal is partly erased from the index
+       anyway. An earlier design considered a deletions/insertions ratio
+       (refusing when deletions dwarf insertions); it is rejected because the
+       reported incident's own commit carried 223,142 insertions alongside its
+       902,840 deletions — ``git add -A`` also stages untracked build
+       artifacts, which count as insertions and defeat any "insertions are
+       negligible" test.
+       - **No missing directory**: fall through to step 3 unchanged.
+       - **A directory is missing**: first ``git -C <worktree> reset -q`` (a
+         mixed reset — index to HEAD, working tree untouched). This is
+         index-recovery, not hygiene: without it, a wipe a *previous* teardown
+         pass already staged (``add -A`` succeeded, the following commit
+         failed) leaves both the "deleted" and "modified/others" plumbing
+         reads blind to it, and the additive salvage below would silently
+         discard real coexisting work. Then compute the additive-only
+         candidate set — ``ls-files --modified --others --exclude-standard``
+         minus ``ls-files --deleted`` — and stage exactly those paths (never
+         ``add -A``, which would restage the declined deletions). If the
+         candidate set is empty (the pure-wipe case), no staging or commit
+         happens at all. Otherwise the WIP commit carries the coexisting
+         additions/edits with **zero deletions**, plus a trailer recording
+         that deletions were declined. Either way the declined deletions are
+         logged at ERROR under ``[worktree-wip-refused-wipe]`` with enough
+         detail (slug, branch, HEAD sha, missing directories, path counts) to
+         reconstruct the event after the worktree is destroyed.
+       - **Fail-open asymmetry**: a failure reading the wipe signal itself
+         (before any directory is known missing) is not evidence of a wipe, so
+         it falls open to step 3 — this function's hardest contract is that it
+         never blocks or hangs teardown. A failure *after* a missing directory
+         is already confirmed (during the reset or the salvage reads) refuses
+         instead of falling through, because falling open there would commit
+         the very wipe just detected.
+    3. ``git -C <worktree> add -A`` — captures untracked + tracked edits.
+    4. ``git -C <worktree> commit --no-verify --no-gpg-sign`` — the WIP commit
        (``--no-verify`` avoids pre-commit hooks hanging teardown;
        ``--no-gpg-sign`` avoids signing prompts).
-    4. ``git -C <repo_root> update-ref refs/session-wip/{slug} <sha>`` — writes
+    5. ``git -C <repo_root> update-ref refs/session-wip/{name} <sha>`` — writes
        to the *common* ref store, so the ref survives both worktree removal and
-       the unmerged-branch-guard branch deletion.
+       the unmerged-branch-guard branch deletion. ``{name}`` is derived from
+       the worktree's checked-out branch (``session/<name>`` -> ``<name>``),
+       falling back to the ``slug`` argument on a detached HEAD, a non-session
+       branch, or a failed read — `_cleanup_stale_worktree` passes the foreign
+       *directory* name as `slug`, which can disagree with the branch the
+       commit actually lands on.
 
     Non-blocking contract: any subprocess failure, timeout, or exception is
     caught, logged at ERROR with the ``[worktree-wip-preserve-failed]`` tag,
     and returned in the result dict — this function NEVER raises into the
     teardown path and must never hang teardown.
 
-    Recovery: ``git checkout refs/session-wip/{slug}`` (or diff/cherry-pick the
+    Recovery: ``git checkout refs/session-wip/{name}`` (or diff/cherry-pick the
     WIP commit). ``git reset --soft HEAD~1`` on the resumed session unstages the
-    WIP commit to restore the dirty tree. Refs live in ``refs/session-wip/*``
-    and are reclaimed manually (no automated GC — see the plan No-Gos).
+    WIP commit to restore the dirty tree. This promise now holds even on a
+    half-deleted tree, because the ref never carries a wipe. Refs live in
+    ``refs/session-wip/*`` and are reclaimed manually (no automated GC — see
+    the plan No-Gos).
 
     Args:
         repo_root: Path to the main repository (common ref store owner).
-        slug: Work item slug (validated).
+        slug: Work item slug (validated). Used to name the ref only when the
+            worktree's checked-out branch cannot be resolved to a
+            ``session/<name>`` form.
         worktree_dir: Path to the worktree to preserve.
 
     Returns:
-        A result dict. On a clean tree: ``{"preserved": False, "was_clean":
-        True}``. On success: ``{"preserved": True, "was_clean": False, "sha":
-        <sha>, "ref": "refs/session-wip/{slug}", "errors": []}``. On failure:
-        ``{"preserved": False, "was_clean": False, "errors": [<msg>, ...]}``.
+        A result dict, always carrying ``"errors"`` (``[]`` unless a git
+        subprocess itself broke) and ``"ref"``.
+
+        - Clean tree: ``{"preserved": False, "was_clean": True, "ref": ...,
+          "errors": []}``.
+        - Ordinary dirty tree (no missing tracked directory): ``{"preserved":
+          True, "was_clean": False, "sha": <sha>, "ref": ..., "errors": []}``.
+        - Wipe, with coexisting additive work: ``{"preserved": True,
+          "was_clean": False, "refused": "missing-tracked-dirs", "missing":
+          [<dir>, ...], "deleted_paths": <int>, "sha": <sha>, "ref": ...,
+          "errors": []}``. The commit's diff against its parent has zero
+          deletions.
+        - Pure wipe (nothing coexisting): ``{"preserved": False, "was_clean":
+          False, "refused": "missing-tracked-dirs", "missing": [...],
+          "deleted_paths": <int>, "ref": ..., "errors": []}``. No commit, no
+          ref write, branch head unmoved.
+        - A failure salvaging a detected wipe (block B) returns the pure-wipe
+          shape above plus ``"guard_error": <str>``, still with ``"errors":
+          []`` — the outcome is a refusal, not a broken-git failure.
+        - Git subprocess failure (the pre-existing contract): ``{"preserved":
+          False, "was_clean": False, "ref": ..., "errors": [<msg>, ...]}``,
+          with no ``"refused"`` key. ``"refused"`` is therefore the sole
+          discriminator between a refusal and a git failure — never populate
+          ``"errors"`` on a refusal.
     """
     _validate_slug(slug)
     ref = f"refs/session-wip/{slug}"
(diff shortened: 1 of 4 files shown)


tomcounsell:
## Review: Changes Requested

The wipe guard itself is sound, well-tested, and does what the plan says. I reproduced all four of the PR's central claims independently: `tests/unit/worktree_manager/` is **149 passed** at `7059ec9`, every one of the 13 grep/anti-criterion verification rows exits 0 when run directly, and four separate mutations each turn the suite red (neutering `reset -q` → `test_pre_staged_wipe_...` fails; making block B fall open → `test_ls_files_failure_after_detection_...` fails; swapping additive staging for `add -A` → 2 failures; disabling detection → 8 failures). The tests are real guards, not vacuous ones.

One blocker, and it is not in the guard — it is in the ref-slug fix that shipped alongside it.

**Restart note:** this changes `agent/`, so deploying it needs `./scripts/valor-service.sh restart` on every bridge machine.

## Rubric

- [x] **1. Plan vs. implementation match** — pass — *every Success Criterion is delivered and independently checked: pure-wipe writes no commit/ref, mixed wipe commits zero deletions with the trailer, `errors`/`refused` discriminate in both directions, partially-staged detection, index carries no staged deletions, ref follows the branch, fail-open asymmetry.*
- [ ] **2. New code quality** — fail — *the new `rev-parse --abbrev-ref HEAD` read at `:1772` is unguarded and can abort the entire preserve on a legitimately dirty tree. See Blocker 1.*
- [x] **3. Test coverage** — pass — *12 new tests; 4 mutation checks confirm they bite.*
- [x] **4. Regression risk to existing callers** — pass — *both call sites (`:1090`, `:1880`) discard the return value, and the dict changes are additive, so no caller behavior shifts.*
- [x] **6. Security** — pass — *no new eval/exec/shell interpolation; the one `subprocess.run` taking external input passes NUL-delimited paths on stdin with `--` terminating the argv, and paths come from git plumbing, not from a request.*
- [ ] **7. Documentation accuracy** — fail — *the docstring's block-B return shape does not match the code. See Tech Debt 1. The feature-doc edit itself is accurate and complete, including the `refs/stash` correction the plan called for.*
- [x] **8. PR body accuracy** — pass — *"149 passed", "12 new regression tests", and "lint/format clean on all changed files" all reproduce. The unchecked full-suite box is an honest disclosure, and I verified its blast radius directly: `preserve_uncommitted_worktree_changes` is referenced by exactly two test files, both inside `tests/unit/worktree_manager/`, which ran fully green.*
- [ ] **9. Disclosed deferrals** — acknowledged — *see Acknowledged Deferrals.*
- [x] **10. Follow-up claims verified** — pass — *#3166 (the `[SEPARATE-SLUG]` No-Go) is OPEN, and the anti-criterion holds: the diff touches no `bridge/` or `monitoring/` file.*

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — PASS — *all 17 Success Criteria checked; the four assertable anti-criteria run clean.*
- **2. No-Gos from plan — none violated** — PASS — *no bridge/watchdog code, no settings field, no env key, no hardcoded directory list, no insertions-ratio predicate.*
- **3. New `except Exception` blocks — each has logger/raise/swallow-ok** — PASS — *both new broad catches (`guard_exc` at block A, `response_exc` at block B) log and take a deliberate, documented direction. The gap is the code that sits **outside** both of them — Blocker 1.*
- **4. New integration tests — exercise serialization boundary (not in-memory only)** — PASS — *the fixtures build real git repos with real linked worktrees and drive the real `git` binary; nothing is mocked except the deliberate failure-injection points.*
- **5. Plan internal consistency — spike findings match task steps** — PASS — *spike-3 (ratio rejected), spike-5 (`--pathspec-file-nul`), spike-6 (index blindness), spike-7 (pre-staged wipe) each map to a shipped behavior and a test.*
- **7. New public APIs — docstrings present** — PASS — *no new public symbol; the existing docstring was rewritten (with one inaccuracy, Tech Debt 1).*
- **8. Breaking changes — migration path documented** — N/A — *return-dict changes are additive and no caller reads it.*
- **9. Tests added for new behavior** — PASS
- **10. Tests cover the failure path (not just happy path)** — PASS — *both directions of the fail-open asymmetry are pinned, and the block-B test asserts the branch head is unmoved rather than merely that nothing raised.*
- **11. UI changes (if any) — screenshot captured** — N/A — *no HTML/CSS/JS/template files in the diff.*
- **12. Docs updated for user-facing changes** — PASS — *`docs/features/session-isolation.md` covers the wipe check, the asymmetry, the ref-naming change, the `reap_idle_worktree` note, the file-map row, and the `refs/stash` correction.*

### Blockers

- [ ] **`agent/worktree_manager.py:1772-1778`** — `branch_result = subprocess.run([..., "rev-parse", "--abbrev-ref", "HEAD"], ...)` — **a cosmetic read can now destroy the backstop it was added to label.**

  **Issue:** this call is new (`origin/main`'s only `--abbrev-ref` invocation is in `verify_worktree_branch`, a different function), it runs on **every** preserve path including the ordinary non-wipe one, and it sits inside the outer `try` with no narrow guard of its own. A non-zero return code falls back correctly (`branch = ""` at `:1778`), but an *exception* — `TimeoutExpired` above all — propagates straight to the outer `except Exception`, which returns the error dict and skips `add -A`, the WIP commit, and the ref write entirely. The force-remove then proceeds and the uncommitted work is gone.

  **Verified, not inferred.** Patching `subprocess.run` to raise `TimeoutExpired` only on the `--abbrev-ref` argv, against a legitimately dirty worktree (`_dirty()`: one tracked edit, one staged edit, one untracked file — no missing directory, so the wipe path is never entered):

  ```
  RESULT: {'preserved': False, 'was_clean': False, 'ref': 'refs/session-wip/probe1',
           'errors': ["Command '[... 'rev-parse', '--abbrev-ref', 'HEAD']' timed out after 5 seconds"]}
  head moved: False
  ref exists: False
  ```

  On `origin/main` this failure mode does not exist, because the call does not exist.

  **Why this is a blocker and not a nit:** it is the plan's own block-A argument, applied to a read the plan did not consider. Block A falls open precisely so a git hiccup does not "trade a rare committed wipe for a routine loss of the backstop on legitimate work." The branch read buys strictly less than block A does — it only decides what to *name* the ref — and yet it can cost the whole commit. The likelihood is low, but the trigger is a teardown storm with many concurrent git processes, which is exactly the condition that produced #3167 in the first place. The function whose purpose is preventing data loss should not gain a new data-loss path in the same PR.

  The docstring at `:1751-1755` already promises the behavior the code does not have: `{name}` falls back to `slug` "on a detached HEAD, a non-session branch, **or a failed read**." A timeout is a failed read.

  **Fix:** wrap it the way block A is wrapped — the fallback value is already correct, so this is additive and needs no new test fixture beyond one mirroring `test_guard_detection_failure_falls_open_and_warns`:

  ```python
  branch = ""
  try:
      branch_result = subprocess.run(
          ["git", "-C", str(worktree_dir), "rev-parse", "--abbrev-ref", "HEAD"],
          capture_output=True, text=True, timeout=settings.timeouts.git_subprocess_s,
      )
      if branch_result.returncode == 0:
          branch = branch_result.stdout.strip()
  except Exception as branch_exc:
      logger.warning(
          "[worktree-wip-guard-failed] slug=%s worktree=%s stage=branch-read error=%s "
          "— ref will use the slug argument.", slug, worktree_dir, branch_exc,
      )
  ```

  Worth a regression test asserting that a raising `--abbrev-ref` read still preserves a dirty tree under `refs/session-wip/{slug}` — otherwise the next refactor reintroduces it.

### Tech Debt

- [ ] **`agent/worktree_manager.py:1735-1737`** — `"A failure salvaging a detected wipe (block B) returns the pure-wipe shape above plus ``\"guard_error\": <str>``"` — the block-B return at `:2038-2046` carries `preserved`, `was_clean`, `refused`, `missing`, `ref`, `errors`, `guard_error`. It does **not** carry `deleted_paths`, which the pure-wipe shape documented two bullets above does carry. A caller written against the docstring gets a `KeyError`. The omission is defensible on its own terms — the `ls-files --deleted` read is exactly what failed, so the count is genuinely unknown — which makes the docstring, not the code, the thing to fix. Say the shape is the pure-wipe dict **minus `deleted_paths`, plus `guard_error`**, and state why. This matters more than usual here because the plan spends several paragraphs establishing `refused` as the sole discriminator for a caller branching on this dict.

### Nits

- None

### Miscellaneous

- [ ] **`docs/plans/auto-preserve-teardown-half-deleted-worktree.md:697-717`** — `| ... | exit 0 |` — **the plan's `## Verification` table certifies nothing when run through the declared runner.** Every row's `Expected` cell reads `exit 0`, but `agent/verification_parser.evaluate_expectation` only recognizes `exit code N`; an unmatched expectation returns `False`, so all 18 rows report `[FAIL]` while printing `Got: exit code 0` directly beneath. It fails closed rather than open, so nothing was falsely certified — but the automated gate is pure noise on this plan, and a reviewer skimming the runner output would reach the opposite conclusion from the truth. `exit code 0` is the repo-wide idiom (123 rows across `docs/plans/`); `exit 0` appears in only three plans, this one included. **Severity:** nit. **Fix:** `sed -i '' 's/| exit 0 |/| exit code 0 |/g'` on the plan. I ran all 18 rows by hand instead — results under Verification Results.

- **"Full `tests/unit/` sweep could not complete during this build — ~10+ concurrent SDLC lanes exhausted the shared test-DB pool"** — not tracked by an open issue; the cited `#2628` is **CLOSED**. Recorded as acknowledged rather than escalated because I checked the substance directly instead of taking the claim: `preserve_uncommitted_worktree_changes` is referenced by exactly two test files, both under `tests/unit/worktree_manager/`, and that whole directory ran **149 passed** on this branch. The unswept remainder cannot reach this change.
- **`[SEPARATE-SLUG #3166]` — the bridge SIGTERM restart loop** — tracked by #3166 (OPEN), and the anti-criterion holds: `git diff --name-only origin/main...HEAD -- bridge/ monitoring/` is empty.
- **The proportional "majority of tracked files gone" guard and its two `PerformanceSettings` tunables** — explicit out-of-scope exclusion with a recorded rationale (spike-3: the incident's own commit carried 223,142 insertions, so a ratio test would not have fired). Pinned against reintroduction by `test_wipe_with_untracked_artifacts_still_refuses_deletions`, which I confirmed goes red when additive staging is swapped back to `add -A`.

### Verification Results

Run from a detached review worktree at `7059ec9`.

- **Worktree-manager tests** — `scripts/pytest-clean.sh tests/unit/worktree_manager/ -q` → **149 passed** in 83s. Reproduces the PR body's number exactly.
- **Format clean** — `ruff format --check .` → 1460 files already formatted. Pass.
- **Lint clean** — `ruff check .` → **2 `I001` errors at HEAD**, both in `agent/agent_session_queue.py`, a file this PR does not touch. **Not attributable to this PR, and not a finding.** Cause: `utils/redis_client.py` landed on `main` after this branch's merge-base (`da844b38`), so at HEAD the module `agent_session_queue.py` imports does not exist and ruff's isort reclassifies it as third-party. Control run: after `git merge origin/main` into the review worktree, `ruff check .` → **All checks passed!** and format stays clean. The merge result — which is what lands — is lint-clean, so no rebase is required on this account.
- **All 13 grep and anti-criterion rows** — run directly (the table's `Expected` grammar defeats the runner; see Miscellaneous): refusal tag, fail-open tag, `ls-tree --name-only -d HEAD`, `--pathspec-file-nul`, `--pathspec-from-file=-`, no tempfile, trailer string, `"reset", "-q"` argv, no hardcoded dir list, no `--numstat`/`--shortstat`, no reintroduced tunables, no `bridge/`+`monitoring/` in the diff, no xfails. **13/13 exit 0.**

### Mutation Checks (reviewer-run, own worktree)

Each mutation applied to a private copy, relevant tests run, then reverted.

- Neuter `git reset -q` (swap the argv for a harmless `rev-parse HEAD`) → `test_pre_staged_wipe_is_unstaged_and_additive_work_still_preserved` **FAILS**. The reset is load-bearing and pinned.
- Make block B fall open (replace its `return` with `pass`) → `test_ls_files_failure_after_detection_refuses_and_never_commits` **FAILS** on the head-sha assertion, and the log shows the wipe being committed. The asymmetry is pinned by the assertion that actually catches it.
- Swap additive staging for `git add -A` → `test_mixed_wipe_preserves_additions_and_drops_deletions` and `test_wipe_with_untracked_artifacts_still_refuses_deletions` **FAIL**.
- Force `missing = []` after detection → **8 failed, 9 passed**. Detection is load-bearing across the whole new suite, and the 9 that still pass are the pre-existing non-wipe cases, as they should be.

tomcounsell:
## Review: Approved

Round-2 re-review. All three findings from the round-1 review at `7059ec9` are fixed, and I re-derived each fix independently in my own detached worktree rather than reading the patch commit's claims.

The blocker fix is the one that mattered, and it is correct: the `rev-parse --abbrev-ref HEAD` read now carries its own narrow `try/except` that degrades to `branch = ""` — the same fallback a non-zero rc already took — instead of propagating into the outer handler and skipping `add -A`, the WIP commit, and the ref write. I re-ran the round-1 exploit as a mutation: reverting to the unguarded form makes the new regression test fail with `assert False is True` on `result["preserved"]` and logs `[worktree-wip-preserve-failed]` on an ordinary dirty tree — the exact data-loss shape the blocker described. Restored, it passes.

**Restart note:** this changes `agent/`, so deploying it needs `./scripts/valor-service.sh restart` on every bridge machine.

## Rubric

- [x] **1. Plan vs. implementation match** — pass — *all 17 Success Criteria delivered; ticked in the plan by this review's sync commit `680537c4`.*
- [x] **2. New code quality** — pass — *the round-1 fail is closed. The new guard mirrors block A's shape exactly: narrow catch, WARNING log with its own greppable tag, deliberate fallback that the docstring already promised. Nothing else was touched.*
- [x] **3. Test coverage** — pass — *`test_branch_resolve_timeout_still_preserves_ordinary_dirty_tree` patches by argv shape (`rev-parse` + `--abbrev-ref`), not by call ordinal, so it will not silently retarget. Mutation-confirmed below.*
- [x] **4. Regression risk to existing callers** — pass — *both live producers (`:1090` `_cleanup_stale_worktree`, `:2252` `remove_worktree`) discard the return value, so the dict-shape wording is documentation-only. Re-derived on this head; the round-1 review's `:1880` is the pre-growth line number for the same call site.*
- [x] **6. Security** — pass — *the delta adds one `subprocess.run` wrapper and a log call; no new eval/exec/shell interpolation, no secrets.*
- [x] **7. Documentation accuracy** — pass — *the round-1 fail is closed: the docstring's block-B shape now reads "the pure-wipe shape above **minus** `deleted_paths` … plus `guard_error`", which matches the actual return at `:2055-2063` (no `deleted_paths` key). One forward-looking note under DOCS below.*
- [x] **8. PR body accuracy** — pass — *"149 passed" is now 150 with the new test, and I reproduced it. Lint and format are clean at this head — the 2 `I001` errors the round-1 review saw were resolved by the `origin/main` merge at `1ca4ad6c`, exactly as that review predicted the merge result would be.*
- [ ] **9. Disclosed deferrals** — acknowledged — *see Acknowledged Deferrals; I re-verified the substance rather than the claim.*
- [x] **10. Follow-up claims verified** — pass — *#3166 is OPEN and its anti-criterion still holds: `git diff --name-only origin/main...HEAD -- bridge/ monitoring/` is empty.*

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — PASS — *17/17, and all 18 `## Verification` rows now execute through the declared runner.*
- **2. No-Gos from plan — none violated** — PASS — *no bridge/watchdog code, no settings field, no env key, no hardcoded directory list, no insertions-ratio predicate. The four assertable anti-criteria run clean.*
- **3. New `except Exception` blocks — each has logger/raise/swallow-ok** — PASS — *the new third catch logs at WARNING under `[worktree-wip-branch-resolve-failed]` and sets the fallback the code path already handles. `except Exception` does not catch `KeyboardInterrupt`/`SystemExit`, so the non-blocking teardown contract is unchanged.*
- **4. New integration tests — exercise serialization boundary (not in-memory only)** — PASS — *the new test builds a real git repo with a real linked worktree and drives the real `git` binary; only the one `rev-parse --abbrev-ref` call is failure-injected.*
- **7. New public APIs — docstrings present** — PASS — *no new public symbol; the existing docstring's one inaccuracy is now fixed.*
- **8. Breaking changes — migration path documented** — N/A — *return-dict changes are additive and no caller reads the dict.*
- **9. Tests added for new behavior** — PASS
- **10. Tests cover the failure path (not just happy path)** — PASS — *the new test asserts the work was actually preserved (`preserved`, a moved head sha, the ref resolving to that sha, a clean `status --porcelain`) rather than merely that nothing raised. That distinction is what makes the mutation bite.*
- **11. UI changes (if any) — screenshot captured** — N/A — *no HTML/CSS/JS/JSX/Vue/template files in the diff.*
- **12. Docs updated for user-facing changes** — PASS — *`docs/features/session-isolation.md` covers the wipe check, the fail-open asymmetry, the ref-naming change, the `reap_idle_worktree` note, the file-map row, and the `refs/stash` correction.*

### Verified

- [x] Code correctness
- [x] Test coverage
- [x] Plan requirements met

### Blockers

- None

### Tech Debt

- None

### Nits

- None

### Miscellaneous

- None

- **"Full `tests/unit/` sweep could not complete during this build — ~10+ concurrent SDLC lanes exhausted the shared test-DB pool"** — not tracked by an open issue; the cited `#2628` is **CLOSED**. Recorded as acknowledged, not escalated, because I checked the substance and hit the condition myself. Two of my own runs on this branch printed `all 15 test-DB slots are held by live processes and none freed within 30s … (#2628)` before recovering, so the contention is real and current rather than an excuse. The blast radius is closed by construction: `grep -rl preserve_uncommitted_worktree_changes tests/` returns exactly two files, both under `tests/unit/worktree_manager/`, and that whole directory is **150 passed** on this head. The unswept remainder cannot reach this change.
- **`[SEPARATE-SLUG #3166]` — the bridge SIGTERM restart loop** — tracked by #3166 (OPEN, re-checked this round), and the anti-criterion holds: the diff touches no `bridge/` or `monitoring/` file.
- **The proportional "majority of tracked files gone" guard and its two `PerformanceSettings` tunables** — explicit out-of-scope exclusion with a recorded rationale (spike-3: the incident's own commit carried 223,142 insertions alongside its deletions, so a ratio test would not have fired). Pinned against reintroduction by `test_wipe_with_untracked_artifacts_still_refuses_deletions` and by the anti-criterion row asserting the tunables are absent from `config/settings.py`.

### Review Delta (vs prior review on HEAD `7059ec9`)

- **Resolved — Blocker 1** (`agent/worktree_manager.py:1772-1778`, unguarded `rev-parse --abbrev-ref HEAD`): fixed at `99ba33d0`. The read is now wrapped in its own `try/except` logging `[worktree-wip-branch-resolve-failed]` and falling back to `branch = ""`. Verified by mutation, not by reading the commit message — see Mutation Check below.
- **Resolved — Tech Debt 1** (docstring's block-B `guard_error` return shape): fixed. The docstring at `:1735-1739` now says the shape is the pure-wipe dict **minus** `deleted_paths` plus `guard_error`, and the code at `:2055-2063` returns exactly `preserved`, `was_clean`, `refused`, `missing`, `ref`, `errors`, `guard_error`. Read both; they agree.
- **Resolved — Miscellaneous nit** (plan `## Verification` table certified nothing): fixed. All 18 rows changed from `exit 0` to `exit code 0`, the repo idiom. Re-run through the declared runner: `checks: 18 malformed: 0 skipped: 0` → **18/18 PASS**. Where the round-1 reviewer had to run the rows by hand, the automated gate now does it.
- **Resolved — lint** (2 `I001` errors on `agent/agent_session_queue.py` at `7059ec9`, attributed to a stale merge-base rather than to this PR): gone at this head, since `1ca4ad6c` merged `origin/main` and brought `utils/redis_client.py` with it. `ruff check .` → **All checks passed!**
- **New**: none.
- **Unchanged**: none.

### Verification Results

Run from a dedicated detached review worktree at `99ba33d0` (code identical at `680537c4`, which adds only this review's plan-checkbox sync).

- **Worktree-manager tests** — `scripts/pytest-clean.sh tests/unit/worktree_manager/ -q` → **150 passed** in 122s. The PR body's 149 plus the new regression test.
- **Plan `## Verification` table, through the declared runner** (`agent.verification_parser.parse_verification_table` → `run_checks`) — 18 checks parsed, **0 malformed, 0 skipped, all 18 PASS**. This is the round-1 nit's fix demonstrated end to end.
- **Lint** — `python -m ruff check .` → All checks passed (ruff 0.15.9).
- **Format** — `python -m ruff format --check .` → 1461 files already formatted.
- **Call-site re-derivation** — `preserve_uncommitted_worktree_changes(` appears at `agent/worktree_manager.py:1090` and `:2252`; both discard the return value.
- **Follow-up issue states** — #3166 OPEN, #3167 OPEN, #2628 CLOSED (re-checked, not carried over).

### Mutation Check (reviewer-run, own worktree)

Applied to a private copy at this head, then reverted; the worktree is byte-clean against `99ba33d0` afterward.

- **Revert the blocker fix** — replace the narrow `try/except` around the `rev-parse --abbrev-ref HEAD` read with the round-1 unguarded form → `test_branch_resolve_timeout_still_preserves_ordinary_dirty_tree` **FAILS** at `assert result["preserved"] is True`, and the captured log shows `[worktree-wip-preserve-failed] … timed out after 5 seconds — uncommitted work could NOT be preserved`. The two neighbouring ref-naming tests still pass, so the new test is the one carrying the guard. The fix is load-bearing and the test is not vacuous.

### Note for the DOCS stage (not a finding)

This round introduced a fifth greppable log tag, `[worktree-wip-branch-resolve-failed]`, and `grep -rn branch-resolve-failed docs/` returns nothing. `docs/features/session-isolation.md` names every other tag this function emits (`[worktree-wip-preserved]`, `[worktree-wip-preserve-failed]`, `[worktree-wip-refused-wipe]`, `[worktree-wip-guard-failed]`), so the set is now one short. The user-visible outcome is already documented — step 5 says `{name}` falls back to the `slug` argument — so nothing in the doc is *wrong*, and the omission is a one-line addition rather than a code change. Left to `/do-docs`, which runs next, rather than spent on another patch round.

tomcounsell:
## Review: Approved

Freshness re-verification at head `c89ff80b`. The round-2 review approved this PR at `680537c4`; the DOCS stage then pushed one commit, which staled that verdict's head. This pass establishes that the staling commit carries no reviewable surface, so the round-2 approval carries forward rather than being re-derived from scratch.

I verified the precondition myself instead of accepting it as a claim, because "the delta is docs-only" is exactly the kind of statement that is cheap to assert and occasionally wrong:

- `resolve_pr_head_sha(3188)` → `c89ff80bef19be17692ee352d24d48ce4c3a8536`.
- `git diff --name-status 680537c4..c89ff80b` → a single `M docs/plans/auto-preserve-teardown-half-deleted-worktree.md`; `--numstat` → `13 13`.
- Stronger than a path check, since a path list cannot prove the *contents* of an unlisted file are unchanged: I compared blob SHAs directly across the two commits. `agent/worktree_manager.py` (`87c9d05e`), `tests/unit/worktree_manager/test_worktree_manager_uncommitted.py` (`ddb706e4`), and `docs/features/session-isolation.md` (`56b9b53e`) are byte-identical at both heads.
- The delta's content is the plan frontmatter `status: Ready` → `status: docs_complete` plus 12 checkbox ticks (5 under Feature Documentation, 7 under Inline Documentation). No source, test, or `docs/features/` change.

Confirmed at this head in an isolated worktree, with `PYTHONPATH` pinned to it and the module import re-resolved to prove the branch's code is what ran (`agent.worktree_manager.__file__` points into the review worktree, and the guard's `ls-tree` / `--pathspec-file-nul` / `[worktree-wip-refused-wipe]` / `[worktree-wip-guard-failed]` markers are all present in the loaded source):

- `scripts/pytest-clean.sh tests/unit/worktree_manager/` — **150 passed**.
- `ruff check` and `ruff format --check` on both changed code files — clean.

The guard mutation sweep from the round-2 review is not repeated: the code under test is byte-identical, so re-running it could only reproduce a result already recorded against the same bytes.

**Restart note (unchanged):** this changes `agent/`, so deploying it needs `./scripts/valor-service.sh restart` on every bridge machine.

## Rubric

- [x] **1. Plan vs. implementation match** — pass — *17/17 Success Criteria carried forward from `680537c4`; the delta ticks the plan's Documentation checkboxes to match what the DOCS stage landed.*
- [x] **2. New code quality** — pass — *code byte-identical to the approved head (blob `87c9d05e`); no new surface to judge.*
- [x] **3. Test coverage** — pass — *test file byte-identical (blob `ddb706e4`); re-run green at this head, 150 passed.*
- [x] **4. Regression risk to existing callers** — pass — *no code delta since approval, so the caller analysis at `680537c4` still holds unchanged.*
- [x] **6. Security** — pass — *the delta is a Markdown plan document; no executable surface.*
- [x] **7. Documentation accuracy** — pass — *the plan's checkbox state now matches the delivered docs. `docs/features/session-isolation.md` is unchanged from the approved head.*
- [x] **8. PR body accuracy** — pass — *the body's "149 passed" reads 150 at this head; the extra test is the round-2 branch-resolve regression test added after the body was written, and I reproduced the 150.*
- [ ] **9. Disclosed deferrals** — acknowledged — *carried forward from the round-2 review; the delta introduces no new disclosure.*
- [x] **10. Follow-up claims verified** — pass — *#3166 remains OPEN; no new follow-up claim in the delta.*

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — PASS — *unchanged from `680537c4`; the delta only records that state in the plan file.*
- **2. No-Gos from plan — none violated** — PASS — *the delta touches one Markdown file; no No-Go surface is reachable from it.*
- **3. New `except Exception` blocks — each has logger/raise/swallow-ok** — N/A — *no code delta since approval.*
- **4. New integration tests — exercise serialization boundary (not in-memory only)** — N/A — *no test delta since approval.*
- **5. Plan internal consistency — spike findings match task steps** — PASS — *frontmatter and checkbox state are now internally consistent with the delivered work.*
- **6. No hardcoded secrets or debug artifacts** — PASS — *delta is prose and checkboxes.*
- **7. New public APIs — docstrings present** — N/A — *no new public symbol.*
- **8. Breaking changes — migration path documented** — N/A — *no breaking change.*
- **9. Tests added for new behavior** — N/A — *no new behavior in the delta.*
- **10. Tests cover the failure path (not just happy path)** — PASS — *carried forward; the suite's fail-open/fail-closed asymmetry cases re-ran green.*
- **11. UI changes (if any) — screenshot captured** — N/A — *no HTML, CSS, JS/TS, JSX/TSX, Vue, or template files in the diff; the visual-proof gate is a no-op.*
- **12. Docs updated for user-facing changes** — PASS — *the DOCS cascade is what this delta records as complete.*

### Blockers
- None

### Tech Debt
- None

### Nits
- None

### Miscellaneous
- None

- **Full `tests/unit/` sweep not completed during build** — carried forward from the round-2 review. Re-confirmed as sound at this head for the same reason plus a stronger one: the change touches one function and its own test file, the scoped `tests/unit/worktree_manager/` suite is green (150 passed), and the code is byte-identical to a head that already cleared a full guard mutation sweep.
- **Bridge/monitoring integration deferred** — tracked by #3166 (OPEN), re-checked at this head.

### Review Delta (vs prior review on HEAD `680537c4`)
- **Resolved**: none — the round-2 review carried zero open findings.
- **New**: none. The only change since `680537c4` is a plan-document checkbox and frontmatter sync, with every code, test, and feature-doc blob byte-identical.
- **Unchanged**: the two acknowledged deferrals above, and the `agent/` restart note.

### Verification Results
- `resolve_pr_head_sha(3188)` = `c89ff80bef19be17692ee352d24d48ce4c3a8536` — matches the head this verdict is recorded against.
- Blob-identity check across `680537c4..c89ff80b` for code, tests, and feature docs — all identical.
- `scripts/pytest-clean.sh tests/unit/worktree_manager/` — 150 passed.
- `ruff check` / `ruff format --check` on changed code files — clean.

### Screenshots

PR #3168, opened 2026-09-05
Ancestor-safe service PID lookup: macOS pgrep hides the bridge from its own descendants (#3164)

valorengels:
Closes #3164

Plan: `docs/plans/ancestor-safe-service-pid-lookup.md`

## The bug

Every `/update` or `python -m scripts.update.verify_release` run **from inside an agent session hosted by the bridge** reported a healthy, correctly-beaconed bridge as `unknown`:

```
WARNING: bridge release could not be confirmed (unknown) — not failing the run
release verify OK @ c022c5c70 (bridge unknown, worker matches)
```

The same check from the `remote-update.sh` cron job on the same machine at the same commit reported `bridge matches`. Nothing was wrong with the bridge.

## Root cause

macOS BSD `pgrep` documents:

> `-a`  Include process ancestors in the match list. By default, the calling process and its ancestors are excluded.

Agent sessions run as `claude -p` children of the bridge:

```
96594 zsh                                                       <- the caller
95415 claude -p ...                                             <- the session
94804 .../MacOS/Python /Users/.../ai/bridge/telegram_bridge.py  <- the bridge
    1 /sbin/launchd
```

So `pgrep -f telegram_bridge.py` exits 1 for any process the bridge spawned, `get_bridge_pid()` returns `None`, and `_classify_process()` short-circuits to `unknown` **before it ever reads the beacon**. That is why the symptom was `unknown` rather than `stale`. Confirmed on the machine: the bridge is present in `KERN_PROC_ALL` with ordinary flags (`0x4004` = `P_EXEC|P_LP64`), `proc_pidpath` resolves, and `pgrep -a -f telegram_bridge.py` finds it — only the default ancestor filter hid it.

This violated a stated acceptance criterion of #1898 ("no false 'release could not be confirmed'").

`pgrep -a` is not the fix: on Linux/procps `-a` means "print the full command line".

## The change

New `tools/process_lookup.py` reads the process table with `ps -axww -o pid=,args=` — `ps` has no ancestor filter — and parses each command line as a **CPython invocation** rather than scanning it for a token that looks right: `argv[0]`'s basename must start with `python` (so the launchd framework binary `Python` matches, which is what `pgrep -fi`'s `-i` was working around), then the interpreter's own options are walked until its own `-m` module or its own script position is reached.

The grammar is load-bearing, not pedantry. A whole-argv scan reports two commands this machine runs during an ordinary test sweep:

```
python -m ruff check bridge/telegram_bridge.py   -> read as the bridge
[...]

PR #3244, opened 2026-09-08
Reconciler scan-health monitor: a per-cycle scan-outcome stamp the watchdog can read (Closes #2691)

### LINKED ISSUE #2691: Reconciler per-chat scan is now load-bearing for the wedge verdict but has no health monitoring
Follow-up from PR #2670's review (see the supplement comment there), which fixed #2475 by requiring positive missed-message evidence (bridge:last_missed_recovery, stamped by the reconciler) before a wedge verdict.

## The gap

That fix makes the reconciler's per-chat scan loop load-bearing for a safety-critical decision, and nothing monitors it:

- record_probe_ok() fires at bridge/reconciler.py:149, immediately after get_dialogs() and BEFORE the per-chat scan; the scan's terminal handler is `except Exception ... continue` with only a logger.error.
- A half-wedged client (dialog list resolves; per-chat iter_messages faults) therefore keeps the probe fresh while stamping no evidence. With #2670's rule, that state produces no verdict and no restart, indefinitely, while messages are lost. Pre-#2670 it restarted within 4h. This is the accepted price of removing silence-as-evidence (the storm was worse), but the exposure is currently unbounded and invisible: the silent-gap check is observability-only and the watchdog never reads it.

## What's needed

Health monitoring for the scan loop itself, e.g. a per-cycle scan-outcome stamp (chats attempted / chats faulted) that the watchdog CAN read, with an alert or restart-eligibility rule when the fault ratio is total for N consecutive cycles. The design question — whether persistent all-chat scan failure should re-enter the restart rule as its own positive evidence, or only page — needs deliberate treatment; do not casually reintroduce a silence-shaped signal.

Refs #2475, PR #2670.

valorengels:
Closes #2691.

## The gap

PR #2670 made the reconciler's per-chat scan load-bearing for the wedge verdict without monitoring it. `record_probe_ok()` fires after `get_dialogs()` and **before** the per-chat loop, whose terminal handler is `except Exception ... continue` with only a `logger.error`. A **half-wedged client** — dialog list resolves, per-chat `iter_messages` faults — keeps the probe fresh, recovers nothing, stamps no `bridge:last_missed_recovery`, and under #2670's rule produces no verdict and no restart, indefinitely, while messages are lost.

Verified still open before building: the gap is live on `main`, and `assess_update_flow`'s own docstring named #2691 as the unbuilt work.

## The stamp

The reconciler writes `bridge:last_scan_outcome` at the end of every scan that got past `get_dialogs()`:

```json
{"ts": ..., "pid": ..., "attempted": 4, "faulted": 4,
 "consecutive_total_fault_cycles": 5, "sample_error": "ConnectionError: ..."}
```

**The "all faulted" vs "never ran" distinction is structural, per the issue's explicit ask — not an inference at read time:**

- **"Scan ran, all faulted"** = a written record with `attempted > 0 and faulted == attempted`. An affirmative statement the scan made about work it did.
- **"Scan never ran"** = the *absence* of a fresh record. There is no value of the key that means it.

That holds because reaching the end of the loop is the only thing that writes the key. A cycle that dies at `get_dialogs()` writes nothing; a cycle that matched no chats writes `attempted: 0`, keeping "cycling with nothing to do" distinguishable from "not cycling".

`consecutive_total_fault_cycles` is maintained **by the writer**, the only component that knows what a cycle is — a reader counting cycles from timestamps would be re-deriving the reconciler's cadence. The record carries the writing `pid`, so a run never carries across a bridge restart (#2475's "a restart clears the accusation").

## Alert, not restart

`assess_scan_health()` appends the issue, flips `healthy` to False, logs it at ERROR, and sets `human_alert_needed`. It deliberately does **not** raise `recovery_level`.

**What that alert delivers today: no notification.** `human_alert_needed` drives the `--check-only` output line and nothing else, and `--check-only` has no automated caller. So the delivered signal is one aggregated ERROR line per watchdog tick in `logs/watchdog.log`, `healthy=False`, and a line in a CLI a human must run by hand. That is a real improvement over one `[reconciler] Error scanning` traceback per chat per scan, but **nobody is paged**, and this PR does not claim otherwise. That is not an oversight in this check: the watchdog delivers nothing by design, because the crash-storm alert that once enqueued an AgentSession to send a Telegram message was removed after alerts stranded as permanently pending sessions -- delivery depended on the bridge and worker being healthy during the exact incident being reported. Wiring an out-of-band path, shared by every check that sets the flag, is tracked in #3252.

A total fault ratio clears the #2475 evidence bar — it is categorically not silence. It fails a *different* bar: **attribution**. It says per-chat history fetches are failing; it does not say this bridge is the broken party.

**Telegram-outage analysis (the restart-storm question):** under a Telegram-side outage, an account-level `FLOOD_WAIT`, or a network partition, every chat faults for every client on the network simultaneously. `last_probe_ok` stays fresh *by construction* — dialogs-resolve-but-fetches-fault is precisely the shape being detected — so the freshness gate that protects the wedge rule offers no protection here. Restart-eligibility would therefore turn any Telegram outage into a restart every 60-second watchdog tick for the outage's duration: the #2475 storm shape with a *correlated external trigger*, hammering Telegram's rate limiter exactly when Telegram is already degraded, to fix a fault that is not on this machine.

Separating "our client is broken" from "Telegram is broken" needs an observer this bridge does not have — a second independent client, or a fleet-wide correlation signal. That is architecture, not a threshold.

**What this still does not cover.** A *partial* persistent fault (say 3 of 4 chats faulting every cycle, indefinitely) resets the run to 0 on every cycle and never alerts. That matches #2691's all-chat framing and is a deliberate scope line, now named in the docs section rather than left implied.

**Open question, split out to #3257:** should restart-eligibility be pursued via a fleet-wide correlation signal, or does paging remain the right permanent answer?

This PR originally carried `Refs #2691` on the strength of that question. On a merge-shepherd re-read, #2691's own "What's needed" asks for a per-cycle scan-outcome stamp the watchdog can read, "with an alert **or** restart-eligibility rule", and for the page-vs-restart design question to get deliberate treatment. All three are delivered here: the stamp, the watchdog read, the alert rule, and the Telegram-outage analysis above answering the design question with "page". The residual question is a *different* one — whether a fleet-wide correlation signal, which does not exist and was not part of #2691's ask, should later reopen restart-eligibility. That is now tracked in #3257, and the docstring and docs pointers were repointed there, so closing #2691 loses nothing.

## Tests

`tests/unit/test_reconciler_scan_health.py` -- **28 passed**. The half-wedged state specifically: dialogs resolve + per-chat fetch faults, asserting the stamp reflects it, that `last_probe_ok` stays fresh in that same state (which is *why* a separate signal is needed), and that the watchdog reads it. Plus run accumulate/reset, partial-vs-total fault, never-ran writes nothing, zero-attempted, and every inconclusive path (missing / stale / wrong-pid / corrupt / no-pid / non-int counters). One test asserts `recovery_level == 0` while `human_alert_needed is True` -- the Telegram-outage guard, pinned. Another asserts a genuine wedge verdict survives a raising `assess_scan_health`.

Regression: `tests/unit/test_bridge_watchdog.py`, `tests/unit/test_bridge_liveness.py`, `tests/unit/test_reconciler.py` -- **117 passed**.

Also re-run `tests/unit/test_process_lookup.py` -- **89 passed, 1 skipped** (the skip is #3168's own environment gate: the test process is not a bridge descendant).

All runs via `scripts/pytest-clean.sh ... -n 0`, measured against `main` at `8e6a0f607` after rebasing onto #3168. `ruff check` and `ruff format --check` clean.

**Rebase note (#3168, ancestor-safe service PID lookup).** #3168 changed `is_bridge_running()` from `pgrep` to `tools.process_lookup.find_python_service_pids()` and edited both files this PR touches. The rebase was clean and both overlapping regions survived intact. Two semantic checks: (1) `update_flow_live` is written in exactly three places -- its initialiser, the `assess_update_flow` call, and its own `except` -- so #3168 introduced no second path that could reset it, and the Check 5b isolation still holds. (2) The new lookup makes this check *more* correct, not less: `assess_update_flow` and `assess_scan_health` both gate on the passed `bridge_pid` matching the pid stamped in the record, and an ancestor-safe lookup returns the true bridge pid where `pgrep` could return none for a bridge-hosted caller.

## Review round 2 (blocker + tech debt)

- **Blocker: the stamp could not self-repair from a corrupt value.** `record_scan_outcome` read the previous record inside its own outer `try`, with `_read_scan_outcome` calling `json.loads` unguarded -- so a corrupt value raised, the handler returned before `r.set(...)`, and the write was gated behind parsing the value it was about to overwrite. The key could not be repaired until the 7-day TTL from the last good write expired, while the tolerant reader kept reporting healthy. Fixed by guarding only the previous-record read; the pid and contiguity checks still apply to `prev`. Regression tests seed the key with a non-JSON value and assert the write lands, the run repairs, and the alert threshold is still reached.
- The other three liveness writers stamp a bare timestamp and never read before writing, so none of them share this shape.
- **TD-1:** Check 5b now has its own `try`/`except`, so a raising alert-only check can no longer discard a wedge verdict and the level-2 restart it authorizes.
- **TD-2:** the paging overclaim is resolved by stating what is actually delivered (above), with #3252 filed for wiring.
- **TD-3:** test numbers corrected and reproducible; the previously cited `test_update_loop_wedge_recovery.py` does not exist and is dropped.
- Nits: `now` reused for the stored `ts`; the total-fault test tightened to `faulted == attempted` to match both docstrings; the non-int counter branch now has test coverage; the remaining partial-fault blind spot is named in the docs.


valorengels:

**Verdict: APPROVED** (no blockers from this lens)

Verified correct:

- **Counter logic** (`bridge/liveness.py::record_scan_outcome`): pid gate, contiguity window, and the accumulate / reset / leave-alone rules for `consecutive_total_fault_cycles` all match the docstring, and are exercised by tests that drive the real `reconcile_once` with a faulting `fetch_messages_back_to` rather than mocking the stamp.
- **Counter scoping** (`bridge/reconciler.py:264`): `scan_attempted` increments only for chats that reach the per-chat body, after the monitored/DM-coverage gates. That is what makes the `attempted == 0` "cycling with nothing to scan" case honest.
- **Constants precedent**: `SCAN_RUN_CONTINUITY_SECONDS`, `SCAN_TOTAL_FAULT_CYCLES`, `SCAN_STAMP_FRESHNESS_SECONDS` are plain module-level `os.environ.get` constants with grain-of-salt comments, matching the neighbouring `CRASH_STORM_THRESHOLD` / `WEDGE_DOMINANCE_FRACTION` / `PROBE_FRESHNESS_SECONDS`. Confirmed by grep that none of those pre-existing watchdog knobs are registered in `.env.example` or `config/settings.py` either — so this is precedent, not a gap.
- **Constant relationships**: writer continuity 900s ≈ 5 reconciler cycles (180s), reader freshness 1200s ≈ 5 cycles + slack, fault threshold 5 cycles ≈ 15 min. Coherent.
- **Function-local `bridge.liveness` imports** match the existing style in the same functions (`record_probe_ok`, `record_missed_recovery`).
- **Legacy cruft**: none. The `docs/features/bridge-self-healing.md` rewrite replaces the old "unmonitored blind spot" prose with the new status quo rather than layering on top of it.

### Nits

- `bridge/liveness.py`: `time.time()` is called twice in `record_scan_outcome` — once as `now` for the contiguity check, once for the stored `ts`. Collapse to `now`.
- `bridge/liveness.py`: the total-fault test is `faulted >= attempted`, while both this function's docstring and `assess_scan_health`'s state the rule as `faulted == attempted`. Equivalent today because the two counters are 1:1; tighten to `==` so a future counter bug cannot pass through as an ordinary increment.
- `tests/unit/test_reconciler_scan_health.py`: no case covers the non-`int` `attempted` / `consecutive_total_fault_cycles` branch of `assess_scan_health`, which is the one inconclusive path without a test.

valorengels:
## Review (Judge risk):

**Verdict: CHANGES REQUESTED** (1 blocker)

The three questions this PR has to survive, answered against the code rather than the description.

### 1. Does a silence-shaped signal get back in? — No, at the reader.

Every branch of `assess_scan_health` was traced. `bridge_pid is None`, read raises, record missing, `pid` mismatch, `ts` non-numeric or older than `SCAN_STAMP_FRESHNESS_SECONDS`, `attempted`/`run` non-`int`, `attempted <= 0`, `run < SCAN_TOTAL_FAULT_CYCLES` — all return `(True, "")`. Only `attempted > 0` + fresh + matching pid + `run >= threshold` returns a fault. "The scan never ran" has no representable value in the record, and the reader never derives one case from the other by timing arithmetic. The structural separation the issue asked for holds.

### 2. Restart-storm safety under a Telegram-side outage — safe, confirmed.

Traced `check_bridge_health` → `run_health_check` → `execute_recovery`. `scan_health_ok=False` appends to `issues` and sets `human_alert_needed`, and never touches `recovery_level` (which stays `0` unless another check raises it). `healthy = len(issues) == 0` therefore flips False, `run_health_check` falls past its healthy short-circuit, and calls `execute_recovery(0, issues)` — whose first branch is `if level == 0: logger.warning(...); return True`, a no-op that does not even create the recovery lock. Scan health also never calls `log_crash()`, so it cannot feed `get_recent_crashes()` / `CRASH_STORM_THRESHOLD` and cannot indirectly open `restart_circuit_open` or reach level-4 auto-revert. **No restart fires from this signal, directly or indirectly.** The only per-tick side effect during an outage is a repeated `Bridge unhealthy` WARNING plus a `Recovery called with level 0` WARNING — noise, not action.

### 3. Write path — correct, with one hole.

The stamp is written on every cycle that gets past `get_dialogs()`, including the all-fault path and the `attempted == 0` path. The two early returns (catchup kill switch, empty `monitored_groups`) are before `get_dialogs()` and correctly write nothing. A `BaseException` re-raised from the release-claim path escapes the per-chat `except Exception` and aborts before the write — also correctly the "silence" case. The hole is the blocker below.

### Blockers

**A corrupt `bridge:last_scan_outcome` value permanently wedges the writer, silently reopening the blind spot this PR exists to close.**

`record_scan_outcome` reads the previous record *inside its own outer try*:

```python
try:
    r = ...
    prev = _read_scan_outcome(r)      # bare json.loads, no local guard
    ...
    r.set(_SCAN_OUTCOME_KEY, json.dumps(record), ex=_TTL_SECONDS)
except Exception as e:
    logger.warning("liveness: record_scan_outcome failed: %s", e)
    return None
```

`_read_scan_outcome` calls `json.loads(raw)` with no try/except of its own. If the key holds a non-JSON value, the decode raises, the outer handler catches it, and the function returns **before reaching `r.set(...)`**. The write is gated behind successfully parsing the value it is about to overwrite, so the key can never be repaired by the next cycle — it self-heals only when the 7-day `_TTL_SECONDS` set by the last *good* write expires.

Reproduced against the real module with a fake Redis client seeded with `b'{not valid json'`:

```
liveness: record_scan_outcome failed: Expecting property name enclosed in double quotes: line 1 column 2 (char 1)
record_scan_outcome returned: None
key still holds corrupt value: b'{not valid json'
```

The reader is the tolerant half of the pair (`get_last_scan_outcome` swallows the same error and returns `None`; `assess_scan_health` then returns healthy), so the combined failure mode is **silent**: for up to a week the bridge can sit in the exact half-wedged state — dialogs resolving, every `iter_messages` faulting, no missed-recovery evidence — with this monitor reporting healthy and nothing distinguishing "the scan loop is fine" from "the stamp mechanism is broken".

Fix (guard only the previous-record read, so a corrupt value means "no previous run" rather than aborting the write; every other guarantee is preserved because the pid and contiguity checks still apply to `prev`):

```python
try:
    prev = _read_scan_outcome(r)
except Exception:
    prev = None
```

### Failure direction

Every inconclusive branch fails toward healthy. That is the right direction, and the blocker above fails that way too — which is precisely why it is dangerous: detection goes dark with zero signal rather than loudly.

### pid reuse

`record_scan_outcome` gates continuity on `prev["pid"] == pid` and `assess_scan_health` on `record["pid"] != bridge_pid`, both tested. OS pid reuse inside the 1200s freshness window is a pre-existing pattern shared with the update-flow detector, not a new risk from this PR.

valorengels:
## Review: CHANGES REQUESTED

**Mode:** independent roster (2 judges) — `code-quality` (APPROVED), `risk` (CHANGES REQUESTED). Aggregated by any-blocker-wins.

**Blockers: 1 · Tech debt: 3 · Nits: 4**

The design this PR is judged against is right, and I want to say that first, because the blocker is a small one sitting inside a careful piece of work.

### What holds up

**The stamp separates "all faulted" from "never ran" structurally, not by inference.** Every branch of `assess_scan_health` was traced: missing record, wrong pid, non-numeric or stale `ts`, non-`int` fields, `attempted <= 0`, and `run < SCAN_TOTAL_FAULT_CYCLES` all return `(True, "")`. Only a fresh, same-pid record with `attempted > 0` and a writer-maintained run at or past the threshold returns a fault. "The scan never ran" has no representable value in the record. Absence is never read as evidence — the #2475 anti-pattern does not get back in.

**Restart-storm safety under a Telegram-side outage: confirmed safe from the code.** `scan_health_ok=False` appends to `issues` and sets `human_alert_needed`, and never touches `recovery_level`. `healthy` flips False, so `run_health_check` falls through to `execute_recovery(status.recovery_level, ...)` — but `recovery_level` is still `0`, and `execute_recovery`'s first branch is `if level == 0: logger.warning(...); return True`, a no-op that does not even take the recovery lock. Scan health also never calls `log_crash()`, so it cannot feed `get_recent_crashes()` / `CRASH_STORM_THRESHOLD` and cannot indirectly open `restart_circuit_open` or reach level-4 auto-revert. During an all-chat outage the cost is two WARNING lines per 60s tick, not a restart. That is the correct answer, and the `recovery_level == 0` assertion in `test_sustained_total_fault_pages_but_never_escalates_recovery_level` pins it.

**Page-not-restart is the right call, and `Refs #2691` correctly matches it.** The attribution argument in the docstring is sound: this bridge cannot tell "our client is broken" from "Telegram is broken", and restart-eligibility on an unattributable all-chat fault is the #2475 storm shape with a correlated external trigger. Leaving restart-eligibility as an owner question rather than shipping a threshold is the conservative and correct choice.

**The tests construct the real state.** `_run_scan` drives the actual `reconcile_once` with `fetch_messages_back_to` patched to raise, and asserts on the stamp the reconciler itself wrote — not on a mock of the stamp. `test_half_wedged_probe_stays_fresh_while_scan_faults` additionally pins the reason a separate signal is needed at all.

**Reproduced locally** (Hard Rule 10 — the PR's own numbers do not reproduce, see TD-3): `tests/unit/test_reconciler_scan_health.py` → **20 passed**. `test_bridge_watchdog.py` + `test_bridge_liveness.py` + `test_reconciler.py` → **117 passed**. `ruff check` and `ruff format --check` both clean.

---

### Blockers

**B-1 — A corrupt `bridge:last_scan_outcome` value permanently wedges the writer, silently reopening the blind spot this PR exists to close.**

`bridge/liveness.py::record_scan_outcome` reads the previous record inside its own outer `try`, and `_read_scan_outcome` calls `json.loads(raw)` with no guard of its own. A non-JSON value therefore raises, the outer handler catches it and returns `None` — **before reaching `r.set(...)`**. The write is gated behind successfully parsing the value it is about to overwrite, so the key cannot be repaired by the next cycle; it self-heals only when the 7-day `_TTL_SECONDS` from the last good write expires.

Reproduced against the real module with a fake Redis client seeded with `b'{not valid json'`:

```
liveness: record_scan_outcome failed: Expecting property name enclosed in double quotes: line 1 column 2 (char 1)
record_scan_outcome returned: None
key still holds corrupt value: b'{not valid json'
```

The reader is the tolerant half of the pair — `get_last_scan_outcome` swallows the same error, `assess_scan_health` returns healthy — so the combined failure is silent in the worst possible direction for a self-healing monitor: for up to a week the bridge can sit in exactly the half-wedged state this PR was built to surface, with this check reporting healthy and no signal distinguishing "the scan loop is fine" from "the stamp mechanism is broken". A monitor that can go dark without saying so is worse than one that was never added, because the docs now assert coverage.

Fix — guard only the previous-record read, so a corrupt value means "no previous run" rather than aborting the write. The pid and contiguity checks still apply to `prev`, so nothing else changes:

```python
try:
    prev = _read_scan_outcome(r)
except Exception:
    prev = None
```

Worth a test alongside it: seed the key with a non-JSON value, call `record_scan_outcome`, assert the key now holds a valid record with `consecutive_total_fault_cycles == 1`.

### Tech Debt

**TD-1 — `assess_scan_health` shares `assess_update_flow`'s try/except, so a paging-only check can veto a restart-eligible verdict.**

In `check_bridge_health`, Check 5b is added inside the same `try` as Check 5, and the `except` resets both:

```python
update_flow_live, update_flow_issue = assess_update_flow(r, pid)
scan_health_ok, scan_health_issue = assess_scan_health(r, pid)
except Exception as e:
    update_flow_live = True
    update_flow_issue = ""
    scan_health_ok = True
    scan_health_issue = ""
```

If `assess_scan_health` raises after `assess_update_flow` has already returned a genuine wedge verdict, that verdict is discarded and the level-2 restart it authorizes never fires. `assess_scan_health` is defensive enough today that this is unlikely in practice, but the structure is wrong in a way that matters here specifically: the whole design premise of this PR is that the paging signal must never influence the restart signal, and this is the one place where it can. Give Check 5b its own `try`/`except` that resets only the scan fields.

**TD-2 — "Pages a human" overclaims what the flag delivers.**

`docs/features/bridge-self-healing.md` line 94, in the same document, states of `human_alert_needed`: *"It is a **diagnostic flag only**: it drives the `--check-only` output line and nothing else. Nothing pushes a notification anywhere."* The new section below it is headed "Reconciler scan-loop health (paging only)" and says the check "pages", and `assess_scan_health`'s docstring says "this check pages" and "it alerts and stops there". Grepping `human_alert_needed` across the repo confirms the only non-test, non-archive consumers are the dataclass and the `--check-only` print; `--check-only` itself has no automated caller.

So the real delivered signal is: one `logger.error` line in the watchdog log, plus `healthy=False`, plus a line in a CLI a human has to run by hand. That is a genuine improvement over `[reconciler] Error scanning` at ERROR — it is a single aggregated line with a run count instead of one traceback per chat per scan — but it is not paging, and no one is notified. Rewrite the doc and the docstrings to say what actually happens ("records the issue, sets the diagnostic `human_alert_needed` flag, logs ERROR, surfaces in `--check-only`; no notification is delivered today"), or wire a real delivery. As written, an on-call reader of that section would conclude they will be told, and they will not.

**TD-3 — The PR body's test numbers do not reproduce.**

The body claims "18 tests" (actual: **20**) and a regression run of "`test_bridge_watchdog.py`, `test_bridge_liveness.py`, `test_reconciler.py`, `test_update_loop_wedge_recovery.py` — 140 passed". `tests/unit/test_update_loop_wedge_recovery.py` **does not exist in this repo** (`ls tests/unit/` confirms; nothing matching `*wedge*` exists), so the 140 figure cannot be reproduced. The three files that do exist give 117 passed, all green. Not a code defect, but the numbers in the body should match a run someone can repeat.

### Nits

- `bridge/liveness.py`: `time.time()` is called twice in `record_scan_outcome` — once as `now` for the contiguity check, once for the stored `ts`. Reuse `now`.
- `bridge/liveness.py`: the total-fault test is `faulted >= attempted` while both docstrings state the rule as `faulted == attempted`. Equivalent today; tighten to `==` so a future counter bug cannot pass through as an ordinary increment.
- `tests/unit/test_reconciler_scan_health.py`: no case covers the non-`int` `attempted` / `consecutive_total_fault_cycles` branch of `assess_scan_health` — the one inconclusive path without a test.
- A **partial** persistent fault (say 3 of 4 chats faulting every cycle, indefinitely) resets the run to `0` every cycle and never pages. That matches the issue's all-chat framing and is a defensible scope line, but it is the blind spot that remains after this PR and the docs section should name it, since the section otherwise reads as though the half-wedge case is now fully covered.

### Notes for deploy

Touches `bridge/` and `monitoring/` — requires `./scripts/valor-service.sh restart` on all bridge machines after merge.

Three new env-overridable constants (`SCAN_RUN_CONTINUITY_SECONDS`, `SCAN_TOTAL_FAULT_CYCLES`, `SCAN_STAMP_FRESHNESS_SECONDS`) are plain module-level `os.environ.get` reads with grain-of-salt comments, matching the neighbouring `CRASH_STORM_THRESHOLD` / `WEDGE_DOMINANCE_FRACTION` / `PROBE_FRESHNESS_SECONDS`, none of which are registered in `.env.example` or `config/settings.py`. Precedent followed; no action needed.

No SDLC lane exists for #2691 (`sdlc-tool stage-query --issue-number 2691` returns empty stages and a null `pr_number`), so no verdict was recorded to the substrate — this comment is the verdict.

valorengels:
## Review: APPROVED

**Head reviewed:** `b51e13ee0e999e06d51f63720e3cefcea18c02be`
**Mode:** independent roster — 2 judges (`code-quality`, `risk`) on round 2 at `803b15ca3`, 1 independent judge on the round-3 delta at `b51e13ee0`.
**Posted as a comment, not a formal review:** the PR is self-authored under this account, so `gh pr review --approve` is refused by GitHub. The recorded SDLC verdict is the authority.

### Shepherd-authorship disclosure

I am acting as merge shepherd on this PR and **personally authored the last two commits** (`803b15ca3`, `b51e13ee0`) plus the `Refs #2691` → `Closes #2691` body change and the title change. Reviewing my own patch is exactly where an approval gate stops being independent, so every shepherd-authored change was disclosed in the judge prompts and dispatched to judges who were told not to defer. The round-3 judge independently re-derived the delta's central factual claim and audited a finding I rejected, rather than taking either on trust.

### Round 2 — two independent judges, zero blockers

| Judge | Blockers | Tech debt | Nits |
|---|---|---|---|
| code-quality | 0 | 1 | 1 |
| risk | 0 | 1 | 0 |

**code-quality TD (accepted, fixed in `b51e13ee0`):** the operator-facing ERROR message at `monitoring/bridge_watchdog.py:508`, and its mirror in the documented example log line at `docs/features/bridge-self-healing.md:177`, still said `(see #2691)`. My own `803b15ca3` repointed two *other* references in these same files for the identical reason and missed this third instance. Once #2691 closes, an on-call human following that pointer lands on a closed issue rather than #3257, where the open question now lives. Both fixed.

**code-quality nit (accepted):** the PR title still read `(Refs #2691)` while the body read `Closes #2691`. Retitled.

**risk TD (rejected, with evidence):** the risk judge flagged the per-dialog preamble — `get_or_init_dm_coverage_epoch(chat_id)`, `find_project_fn(chat_title)`, `find_project_for_dm_dialog(dialog.entity)` — as unguarded I/O whose failure would abort the whole cycle before `record_scan_outcome` runs, degenerating into the same silent-healthy state #2691 exists to close. The finding cites the call sites correctly but does not survive reading the callees:

- `bridge/dedup.py:273-293` — `get_or_init_dm_coverage_epoch` carries a blanket `except Exception` and its docstring states "Never raises"; it returns `(now, True)` on any failure.
- `bridge/routing.py:408-429`, `:448-450` — `find_project_for_dm_dialog` and `find_project_for_bot` are `getattr` calls and `.get()` lookups against module-level maps. No I/O.
- `bridge/routing.py:382-392` — `find_project_for_chat`, the actual runtime binding of `find_project_fn`, iterates the module-level `GROUP_TO_PROJECT` dict. No I/O.

The preamble cannot propagate, so a guard there would be defensive cruft against an impossible failure. Instead the guarantee is now **pinned in a comment** at the `record_scan_outcome` call site, naming what makes it hold and warning that adding unguarded I/O to the preamble would take the monitor dark. The round-3 judge independently read all four callees and confirmed the rejection, finding no missed path.

### Round 3 — delta judge, APPROVED

Delta: 3 lines (two `#2691` → `#3257` pointer swaps) plus the comment block. The judge verified the comment's guarantee against the source rather than trusting it — confirming `check_silent_chat` (`:174-185`) and `get_last_processed` (`:224-245`) carry their own handlers, the entire per-chat body (`:264-419`) sits inside one `try` whose `except` increments `scan_faulted` and continues without re-raising, and every named callee is non-raising. It also swept both files and confirmed the provenance-vs-live-pointer rule is applied consistently: `#2691` correctly remains at `reconciler.py:142,431,446`, `bridge_watchdog.py:143,197,787`, and `bridge-self-healing.md:167` as provenance, while both live "see #X" operator pointers now read `#3257`.

Zero blockers, zero tech debt, zero nits.

### Closing link: `Refs #2691` → `Closes #2691`

**This overrides a deliberate choice by the PR author and a round-1 reviewer, so it is named rather than slipped in.** The author wrote "Open question for the owner: should restart-eligibility be pursued via a fleet-wide correlation signal, or does paging remain the right permanent answer? Hence `Refs`, not `Closes`", and a round-1 reviewer endorsed it: "Page-not-restart is the right call, and `Refs #2691` correctly matches it."

#2691's own "What's needed" asks for a per-cycle scan-outcome stamp the watchdog CAN read, "with an alert **or** restart-eligibility rule", and for the page-vs-restart design question to get deliberate treatment. All of it is delivered here: the stamp (`bridge/liveness.py:212-312`), the watchdog read (`monitoring/bridge_watchdog.py:414-509`) wired into `check_bridge_health` (`:787-810`), the alert rule, and a substantive Telegram-outage/attribution argument answering the design question with "page". The residual — whether a *fleet-wide correlation signal*, which does not exist and was not part of #2691's ask, should later reopen restart-eligibility — is genuinely new scope and is now tracked in **#3257**, with the code and docs pointers repointed there. Closing #2691 loses nothing.

Both round-2 judges were asked to say plainly whether the override was right, and both independently agreed. The code-quality judge: "The round-1 reviewer's `Refs` was appropriate for a PR that hadn't yet answered the design question in writing; this PR now has, so `Closes` is correct."

### Test evidence

Green at the reviewed head: **234 passed, 1 skipped** across `test_reconciler_scan_health.py`, `test_bridge_watchdog.py`, `test_bridge_liveness.py`, `test_reconciler.py`, `test_process_lookup.py`, run on the rebased base (`main` @ `8517a09e8`) via `scripts/pytest-clean.sh -n 0`. `ruff check` and `ruff format --check` clean.

**RED replay (not inherited).** The PR's original red was proven against a now-stale base, and this suite was specifically at risk: round 1 nearly shipped `test_record_scan_outcome_never_raises_on_redis_failure` asserting the bug — it simulated "Redis down" by breaking only `get()`, which is precisely the read-aborts-the-write behavior that was under repair, so it passed for the wrong reason. It has since been corrected to break both `get()` and `set()`, with the read-only case split into `test_read_failure_alone_does_not_abort_the_write`.

I replayed the red on the current base by mutation rather than trusting the claim: removing the `try`/`except` around the previous-record read in `record_scan_outcome` — reintroducing exactly the round-1 blocker — turns **3 tests red**:

- `test_read_failure_alone_does_not_abort_the_write`
- `test_corrupt_stored_value_does_not_gate_the_write`
- `test_corrupt_stored_value_still_reaches_the_alert_threshold`

The suite is genuinely load-bearing against the defect class it certifies.

### Rebase

Rebased cleanly onto `main` @ `8517a09e8`. No overlap between this PR's files and anything that landed in the skipped range `8e6a0f607..8517a09e8`. #3168's `pgrep` → `find_python_service_pids()` change to both touched files survived intact.

valorengels:
## Review: APPROVED (docs delta, head `e32f12fd2`)

Docs cascade for the scan-health monitor. **Shepherd-authored** — I wrote every line of this delta, so it went to an independent judge with that disclosed and instructions not to defer.

Two live docs carried claims this PR falsifies:

- `docs/features/bridge-worker-architecture.md` — the bridge liveness-signals list said a half-wedged client is "loud in the logs, **unacted on**". That is no longer true; it is now stamped and monitored. Trimmed the stale clause and added a `bridge:last_scan_outcome` bullet.
- `docs/features/message-reconciler.md` — added an `assess_scan_health` row to the related-components table, since the reconciler's own scan loop is now a monitored, safety-critical component.

The judge verified every factual claim against source rather than accepting it: the write-site-and-never-earlier invariant (`reconciler.py:431-453`, `liveness.py:212-286`), the exact record fields (`liveness.py:274-281`), the affirmative-record-vs-absence distinction (`bridge_watchdog.py:436-445,479-499`), and the no-restart guarantee (`:801-806`). It specifically checked for a paging overclaim — nothing is delivered to a human today, since `human_alert_needed` drives a `--check-only` line with no automated caller (#3252) — and confirmed the new text claims only "loud and monitored, but still not auto-restarted".

Zero blockers, zero tech debt, zero nits.

Issue #3176, opened 2026-09-05
Synthetic-slug worktrees are invisible to both busy-guard predicates: the executor never persists the slug it synthesizes or the worktree it resolves


tomcounsell:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/synthetic-slug-worktrees-invisible-to-busy-guards.md

## The defect

A slugless eng session gets a synthetic slug at `agent/session_executor.py:1314`:

```python
if not slug and getattr(session, "session_type", None) == "eng":
    _aid_for_slug = getattr(session, "agent_session_id", None)
    if _aid_for_slug:
        slug = f"dev-{_aid_for_slug[:8]}"
        is_synthetic_slug = True
```

`slug` here is a **local variable**. It is never written back to `session.slug`.

The executor then provisions a worktree from it (`get_or_create_worktree(working_dir, slug)`, `:1377`) and rebinds the **local** `working_dir` to that path (`:1378`). That is also never written back: `grep -rn '\.working_dir *=' --include='*.py'` over non-test code returns exactly one hit, `agent/session_runner/role_driver.py:198`, which is a driver attribute rather than a session row.

So while such a session runs inside `.worktrees/dev-{aid8}/`, its stored `AgentSession` row says `slug=None` and `working_dir=<main checkout>`.

## Why it matters

Both busy-guard predicates read the stored row, so both miss it:

- `_scan_worktree_sessions()` (`agent/worktree_manager.py:433`) matches on `working_dir` being inside the lane. The stored `working_dir` is the main checkout, so no match.
- A hypothetical indexed `slug=` lookup would also miss it, since the stored `slug` is `None`.

`worktree_busy_check()` and `worktree_busy_probe()` therefore both report a `dev-*` lane **clear** while a session is live inside it. The only guard that sees it is `_worktree_has_live_process()` (`agent/worktree_manager.py`, #2305 defect 3), which asks the OS directly.

Consequence: the busy guard is a no-op for synthetic-slug lanes, and the OS-process scan is load-bearing rather than a backstop for them. Any caller that runs the AgentSession guard without the process scan makes an unguarded decision.

## Recon Summary

**Confirmed**

- `agent/session_executor.py:1314` synthesizes `dev-{agent_session_id[:8]}` into a local, with no write-back to `session.slug`.
- `agent/session_executor.py:1377-1378` resolves and rebinds `working_dir` locally, with no write-back.
[...]

Issue #232, opened 2026-03-03
DM/group chat cross-wire: classifier false positive + shared Claude Code session context


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/fix_chat_cross_wire.md

## Bug Description

Messages from one Telegram chat (e.g., Dev: Valor group) can get their responses delivered to a different chat (e.g., Tom's DM) in reply to a completely different conversation.

## Observed Behavior

1. Tom asked "We are getting duplicate daydream reports each day" in **Dev: Valor** group (msg 6315)
2. Tom asked "How does the summarizer feature work?" in **DM** (msg 6316)
3. The DM received a response about **daydream duplicate bugs** instead of the summarizer explanation

## Root Cause Analysis

Two bugs acting in tandem:

### Bug 1: Classifier misclassifies Q&A answers as status updates

The output classifier (`bridge/summarizer.py`) classified an informational answer about the summarizer feature as `status_update` (confidence 0.88) instead of `completion`. This triggered an unwanted auto-continue instead of delivering the actual answer.

**Evidence from logs:**
```
08:08:36 [dm] Output classified as status (confidence=0.88): Agent is explaining an existing system architecture rather than reporting on active work or requesting input
```

The classifier treats factual/explanatory answers as "status updates" because they don't contain explicit questions or completion signals. Q&A responses about system architecture look like status narration to the classifier.

### Bug 2: Shared Claude Code session context across Telegram conversations

The auto-continue resumed a Claude Code session (`0d825ffe-1f00-42ec-aeb1-37a8ea635d2a`) that had been running since Feb 25 and contained context from the concurrent Dev: Valor group investigation into daydream bugs. The resumed session produced a response about daydream instead of the summarizer.

**Evidence from logs:**
```
08:08:35 stop_hook session_id=0d825ffe-1f00-42ec-aeb1-37a8ea635d2a  # DM session
08:08:56 stop_hook session_id=0d825ffe-1f00-42ec-aeb1-37a8ea635d2a  # Auto-continue uses SAME session
```

Both the DM worker and the group worker share the same Claude Code session because `resume=session_id` in `sdk_client.py` (line 527) maps our Telegram session IDs to Claude Code's internal session management, but there's only one active Claude Code conversation on disk.

## Fix Required

1. **Classifier**: Add Q&A/informational patterns to the classifier so explanatory answers are classified as `completion`, not `status_update`
[...]

PR #3245, opened 2026-09-08
Guard KeyField mutation on hydrated rows (session_transcript, knowledge_document)

### LINKED ISSUE #3210: session_transcript writes AgentSession.slug on a hydrated instance, forking the Redis row (slug is a KeyField)
## Problem

`AgentSession.slug` is a `KeyField` (`models/agent_session.py`, added by #1085 so worker pop can filter by it). Popoto concatenates key-field values into the row's Redis primary key, so assigning `slug` on an already-persisted instance and saving writes a **new row at a new db_key** and orphans the original at the `slug=None` key.

`bridge/session_transcript.py` does exactly that. In the transcript-phase update it hydrates the enqueue-time row via `AgentSession.rows_for_session_id(session_id)`, takes `existing[0]`, and runs `if slug: s.slug = slug` before saving.

The file already knows the rule — the very next branch handles `chat_id` correctly, with the comment "chat_id is a KeyField — mutating it after creation silently creates a new Redis record (orphaning the old one)", and only assigns it when the existing value is empty. `slug` got no such guard.

## Recon Summary

Measured on `d786c8ad2`, by symbol.

**Confirmed:**
- `models/agent_session.py` declares `slug = KeyField(null=True)`; `AgentSession._meta.key_field_names` resolves to `['chat_id', 'id', 'parent_agent_session_id', 'project_key', 'session_type', 'slug']`.
- `bridge/session_transcript.py` assigns `s.slug = slug` on the hydrated `existing[0]`, unguarded.
- The adjacent `chat_id` branch in the same block implements the correct empty-only pattern and documents the hazard.
- Repo-wide, `grep -rnE '^\s*[A-Za-z_][A-Za-z0-9_]*\.slug\s*=[^=]' --include='*.py' agent bridge models tools worker` returns two hits: this one, and `tools/lane_identity.py`, which writes `PipelineLedger.slug` (a different model) and is not affected.

**Symptom shape:** a duplicate `AgentSession` for one `session_id` — one row at the `slug=None` key stuck in whatever status it held, one at the `slug=<value>` key. `models/session_lifecycle.get_authoritative_session()` exists to tie-break duplicates like these; a leaked `status=running` orphan never finalizes on its own.

## Desired Outcome

Match the `chat_id` treatment: assign `slug` only when the existing value is empty, and warn on a mismatch rather than writing. Or route the rename through the delete-and-recreate path (`clone_agent_session_fields`) that exists for exactly this — changing a KeyField.

## Why this is filed separately

Found while planning #3176 (spike-3 established the KeyField hazard; the repo-wide grep for the plan's anti-criterion surfaced this call site). #3176 does not touch `bridge/`; its scope is the busy-guard predicate and the synthetic-lane cleanup.

valorengels:
Closes #3210

## The fix

`bridge/session_transcript.py` assigned `s.slug = slug` on the row hydrated from `AgentSession.rows_for_session_id(session_id)`. `slug` is a `KeyField`, so that assignment changes the row's Redis identity. It now uses the empty-only + warn shape the adjacent `chat_id` branch in the same block already established.

## The damage is worse than the issue describes on the current pin

The issue predicts a forked row. Under **popoto 1.9.0's divergence guard** the assignment does not fork — it *raises*:

```
Cannot change KeyField 'slug' from 'lane-alpha' to 'lane-beta' without migrate_key=True.
```

That raise lands in `start_transcript`'s broad `except Exception`, **before `s.save()`**. So on the current pin the observable damage is a silent total loss of the entire transcript-phase update: `log_path`, `sender_name`, `branch_name`, `classification_type` and `correlation_id` all discarded, and the `active` lifecycle transition never logged. The only trace was a generic `Failed to update/create AgentSession` warning that named nothing.

## Sweep result

All six `AgentSession` key fields (`chat_id`, `id`, `parent_agent_session_id`, `project_key`, `session_type`, `slug`), swept across `agent/ bridge/ models/ tools/ worker/ utils/`, direct assignment and `setattr`.

| Site | Verdict |
|---|---|
| `bridge/session_transcript.py:88` `s.slug` | **FIXED** — the reported bug |
| `models/knowledge_document.py:106` `doc.project_key` | **FIXED** — second live instance, see below |
| `bridge/session_transcript.py:96` `s.chat_id` | Safe — already guarded; the pattern this PR copies |
| `models/agent_session.py:1501` `self.id = value` | Safe — the `agent_session_id` property setter. Its only two callers (`agent/session_executor.py:1920`, `scripts/migrate_model_relationships.py:205`) write `TelegramMessage.agent_session_id`, a plain `Field` on a different model, never `AgentSession.id` |
| `tools/lane_identity.py:428` `fresh.slug` | Safe — `PipelineLedger.slug` is not a `KeyField` (`ledger_key` is the model's only one), and the write is already conditional-on-empty |
| `models/crash_signature.py:257` `self.project_key` | Safe — `CrashSignature.project_key` is an `IndexedField`, not a `KeyField` |
| `tools/telegram_history/__init__.py:1159` `chat.project_key` | Safe — `Chat.project_key` is declared `Field(null=True)` with the comment "to avoid delete-and-recreate on change"; the file's actual `KeyField` (`chat_name`) already uses delete-and-recreate |
| `bridge/catchup.py:377`, `bridge/reconciler.py:303` `self.chat_id` | Safe — `MinimalEvent.__init__`, a local non-Popoto shim class |
| `tools/valor_session.py:284` `self.project_key` | Safe — exception class `__init__` |
| `tools/sms_reader/__init__.py:156,234` | Safe — `chat_id` inside SQL text |
| `models/session_lifecycle.py:222` `setattr(session, ...)` | Safe — generic companion-field loop; no caller in the tree passes a key-field name |
| `agent/session_runner/adapter.py:572` | Safe — fixed scalar allowlist, no key fields |
| `agent/session_health.py:5692` | Safe — `tool_timeout_count_{tier}` only |
| `models/agent_session.py:2244` | Safe — fixed map of `issue_url`/`plan_url`/`pr_url` |

### The second live instance

`KnowledgeDocument.project_key` is a `KeyField`. `safe_upsert()` looked a doc up by `file_path` **alone** and then reassigned `project_key`, so re-indexing one file under a second project forked it into two rows — the same document visible to two projects' searches, which is an NDA-isolation concern, not just a duplicate. Same empty-only + warn guard applied. This is strictly better than the status quo either way: forking leaks across projects, and on the 1.9.0 pin the raise would discard the content/hash/mtime update too.

## Test evidence

`tests/unit/test_session_transcript_keyfield_fork.py`, asserting on row count and `db_key` identity, not just that the path ran.

**Red** (fix stashed, code as on main):
```
tests/unit/test_session_transcript_keyfield_fork.py .FF
FAILED test_conflicting_slug_does_not_fork_the_row
FAILED test_non_key_fields_still_update_on_the_same_row
  assert row.branch_name == "session/lane-alpha"
E AssertionError: assert None == 'session/lane-alpha'
2 failed, 1 passed
```

**Green** (with fix, plus the existing suite for the file):
```
tests/unit/test_session_transcript_keyfield_fork.py ...
tests/unit/test_session_transcript.py .......
10 passed
```

`tests/unit/test_knowledge_document.py` + `test_knowledge_indexer.py`: 42 passed.

## Production orphans

**None.** Read-only ORM probe (`AgentSession.query.all()`, no writes): 120 rows, 2 `session_id`s with more than one row — `sdlc-local-2771` and `sdlc-local-3053`. Both duplicate groups are identical on *every* key field including `slug=None`, differing only in the `id` `AutoKeyField`. That is a double-create, a different defect from this one; zero slug-fork-shaped orphans exist. Nothing was mutated or deleted.

## Recommendation: a model-level guard is worth its own lane

popoto 1.9.0 already added the divergence guard, so the primitive is loud at the ORM. The gap is one level up: **`start_transcript`'s broad `except Exception` turns that loud raise back into silence**, and it is not the only such swallow in the tree. Two concrete proposals, both out of scope here:

1. **Narrow the swallows.** Let the popoto `KeyField` divergence error propagate (or re-raise it distinctly) instead of collapsing into a generic best-effort warning. A field-write failure that discards five other fields should not be indistinguishable from a Redis blip.
2. **Give the intent a name.** Add an `AgentSession.set_key_field_or_warn(name, value)` helper wrapping the empty-only + warn shape now hand-copied at three call sites, and make the migration case explicit via popoto's `migrate_key=True` (or the existing `clone_agent_session_fields` delete-and-recreate path) so a genuine rename has a sanctioned route rather than being a silent no-op.


valorengels:

### Blockers

**1. `bridge/session_transcript.py:88-90` and `models/knowledge_document.py:106-109` — the new comments describe a failure mode the pinned dependency does not produce.**

Both comments say the KeyField mutation "silently creates a new Redis record (orphaning the old one)". On the pinned `popoto==1.9.0` it does not: `Model.save()` raises `KeyMutationError` (`popoto/models/base.py`, `if old_value is not None and old_value != new_value: raise KeyMutationError(...)`). I reproduced the raise directly — reverting only the two source guards and re-running the new test yields:

```
WARNING bridge.session_transcript: Failed to update/create AgentSession for test-keyfield-fork-...:
Cannot change KeyField 'slug' from 'lane-alpha' to 'lane-beta' without migrate_key=True.
```

The PR body and the test docstring (`tests/unit/test_session_transcript_keyfield_fork.py:11-18`) both get this right. The inline comments — the text a future maintainer actually reads at the call site — do not, and they omit the consequence that motivated the fix: the raise is swallowed by `start_transcript`'s broad `except Exception` **before** `s.save()`, discarding the entire transcript-phase update. A reader who trusts the comment could remove the guard on the reasoning that the worst case is a redundant row. `NO LEGACY CODE TOLERANCE` / "describe only the new status quo" is binding here.

Fix: state the current-pin behavior, e.g. `# project_key is a KeyField — reassigning it on a hydrated row raises KeyMutationError under the pinned popoto (older pins forked the row instead), and start_transcript's broad except swallows it before save(). Only set it when empty.`

### Tech Debt

**1. `models/knowledge_document.py:106-117` ships with zero test coverage.**

`tests/unit/test_session_transcript_keyfield_fork.py` covers only the `slug` half. The `project_key` guard in `safe_upsert` has no new test, and no existing test in `tests/unit/test_knowledge_document.py` exercises the divergence path. The PR frames this half as an independently-motivated cross-project isolation fix; it should carry a regression test mirroring `test_conflicting_slug_does_not_fork_the_row`.

**2. `bridge/session_transcript.py:91-98`, `:105-112`, `models/knowledge_document.py:110-117` — the same eight-line guard is now hand-copied at three sites**, with hand-varied warning wording. Acceptable at three; worth a named helper (`guard_keyfield_reassignment(instance, field, value, context)`) before a fourth copy appears. The PR's own recommendation section proposes the same thing.

### Nits

**1. `bridge/session_transcript.py:93`, `models/knowledge_document.py:112` — the `str(...)` wrapping on both sides of the mismatch comparison is redundant**; the preceding `if not X` branch has already established both operands are truthy `KeyField` strings. Harmless.

valorengels:

### Blockers

**1. `models/knowledge_document.py:106-119` + `tools/knowledge/indexer.py:321-332` — the guard protects the `KnowledgeDocument` row but not the denormalized state that actually answers searches, and in doing so it *opens* a cross-project write path that was closed before this PR.**

`safe_upsert` keeps `doc.project_key` at its stored value on mismatch, but every other field in that block is still assigned and saved unconditionally (`doc.content`, `doc.content_hash`, `doc.last_modified` at lines 103-105, `doc.scope` at line 118). The row is then internally inconsistent: `scope` reflects the incoming project, `project_key` does not.

The caller ignores the doc's authoritative value entirely:

```python
doc = KnowledgeDocument.safe_upsert(abs_path, project_key, scope)   # indexer.py:321
if content_changed:
    _sync_chunks(doc, raw_content, project_key)                     # :327 — local var, not doc.project_key
_create_companion_memories(abs_path, project_key, scope, decoded_content(doc))  # :332 — same
```

`DocumentChunk.project_key` and `Memory.project_key` are both `KeyField`s, and `DocumentChunk.search()` filters strictly on `chunk.project_key` independent of the parent doc. `_create_companion_memories` calls `_delete_companion_memories(file_path)` first, which deletes by `reference` across **all** projects, then recreates under the incoming key.

Net effect after this PR, when `resolve_scope(abs_path)` returns a different project_key than the stored one: the `KnowledgeDocument` row stays on the OLD project while its chunks and companion memories move wholesale to the NEW one. The old project's search loses the content; the new project's chunk and memory search surface it. That is the cross-project visibility the PR exists to prevent, relocated one layer down into the layer that serves queries.

Critically, this is a **new** path. Pre-fix on the `popoto==1.9.0` pin, `safe_upsert` raised, returned `None`, and `index_file` bailed at line 322 — `_sync_chunks` and `_create_companion_memories` never ran. The guard makes `safe_upsert` succeed, which is what lets the caller proceed to write the denormalized rows under the incoming key.

Reachability: `resolve_scope` (`tools/knowledge/scope_resolver.py:67-91`) is a pure path→key map read from `projects.json`, longest-prefix-first, with a `~/work-vault` root fallback to `("company", "company-wide")`. Adding a project whose `knowledge_base` is a subfolder of `~/work-vault` re-keys every already-indexed file under it from `company` to the new key. That is ordinary configuration maintenance, not an exotic event.

The correct pattern already exists a few hundred lines away: `tools/knowledge/indexer.py:599` calls `_sync_chunks(doc, raw, doc.project_key)` — the authoritative stored value.

Fix, either:
- (a) make the whole update fail closed on divergence — skip the content/hash/mtime/scope writes and return without saving, so the caller's `doc is None` / unchanged path keeps chunks and memories where they are; or
- (b) key the lookup on `(file_path, project_key)` and route a genuine re-key through `save(migrate_key=True)` so doc, chunks and memories all move together.

Either way, pass `doc.project_key` (not the local `project_key`) to `_sync_chunks` and `_create_companion_memories`, and add a test asserting chunk/memory partition after a divergent re-index.

### Tech Debt

**1. Warn-and-skip is the only signal for a permanently discarded write.** Both guards emit a `logger.warning` and move on. Nothing reaches a metric, dashboard, or Sentry, the write is never retried, and there is no reconciliation path back to consistency. This is strictly better than the swallowed exception it replaces, but it converts a loud failure into a quiet permanent drift.

**2. `bridge/session_transcript.py` — no production caller passes `slug=` to `start_transcript` today.** A grep across `agent/ bridge/ tools/ worker/` finds `slug=` only in tests. The guard is correct and cheap, but it hardens a currently dormant path; the live, reachable half of #3210 is the `KnowledgeDocument` one, and per the blocker above that half is only partially closed.

### Nits

**1. `bridge/` is touched — this requires `./scripts/valor-service.sh restart` on every bridge machine after deploy.** Rollback is clean (both changes are additive guards, no schema or data migration).

valorengels:
## Review: CHANGES REQUESTED

**Mode:** independent roster (2 judges) — `code-quality` and `risk`, dispatched with the same PR context and no visibility into each other's reasoning. Consensus rule: any-blocker-wins. Both returned CHANGES REQUESTED. Head SHA reviewed: `0a2cd69afa4bde8a999d6a7be96cd8c24995530d`.

Per-judge findings are in the two comments above. Aggregate below, plus the reviewer's own independent sweep and test verification.

---

### Blockers (2)

**B1 — the `KnowledgeDocument` half does not close the isolation leak it claims to close, and opens a new one.** (`models/knowledge_document.py:106-119`, `tools/knowledge/indexer.py:321-332`; raised by the `risk` judge, independently confirmed by the reviewer.)

The guard stops the doc row from forking, but `tools/knowledge/indexer.py` passes its own locally-resolved `project_key` — not `doc.project_key` — to both `_sync_chunks` (line 327) and `_create_companion_memories` (line 332). `DocumentChunk.project_key` and `Memory.project_key` are both `KeyField`s, `DocumentChunk.search()` filters on the chunk's own key, and `_delete_companion_memories` deletes by `reference` across all projects before recreating under the incoming key.

So on a `project_key` divergence the doc row stays on the OLD project while its chunks and companion memories move to the NEW one. Before this PR, on the `popoto==1.9.0` pin, `safe_upsert` raised, returned `None`, and `index_file` bailed at line 322 — neither write ran. Making `safe_upsert` succeed is what lets the caller reach them. The repo already has the right pattern at `indexer.py:599` (`_sync_chunks(doc, raw, doc.project_key)`).

Also in the same block: `doc.scope` is still assigned unconditionally (line 118), so the saved row carries the incoming project's `scope` alongside the stored project's `project_key`.

**B2 — the new inline comments describe a failure mode the pinned dependency does not produce.** (`bridge/session_transcript.py:88-90`, `models/knowledge_document.py:106-109`; raised by the `code-quality` judge.)

Both say the mutation "silently creates a new Redis record". On `popoto==1.9.0` it raises `KeyMutationError`. The PR body and the test docstring state this correctly; the comments at the call sites do not, and they omit the swallowed-exception consequence that is the actual motivation. `NO LEGACY CODE TOLERANCE` / "describe only the new status quo" is binding.

### Tech Debt (4)

- **TD1** — `models/knowledge_document.py` guard ships with **zero test coverage**. The `slug` half has three tests; the half the PR itself calls the live instance has none.
- **TD2** — warn-and-skip is the only observability for a permanently discarded write; no metric, no Sentry, no retry, no reconciliation path.
- **TD3** — the eight-line guard is now hand-copied at three call sites with hand-varied wording. Worth a named helper before a fourth.
- **TD4** — no production caller passes `slug=` to `start_transcript` today (grep across `agent/ bridge/ tools/ worker/` finds it only in tests). The `slug` guard is correct and cheap, but it hardens a dormant path; the reachable half is the `KnowledgeDocument` one.

### Nits (3)

- The `str(...)` wrapping on both sides of each mismatch comparison is redundant — the preceding `if not X` branch has already established both operands are truthy strings.
- `bridge/` is touched: requires `./scripts/valor-service.sh restart` on every bridge machine after deploy. Rollback is clean; both changes are additive guards with no schema or data migration.
- Test-quality nuance, below.

---

### Reviewer's independent verification

**Sweep — reproduced and widened.** I re-swept all `KeyField` declarations across `models/` (29 models) and every direct assignment and `setattr` path, extending the scope to `scripts/`, which the PR's declared sweep (`agent/ bridge/ models/ tools/ worker/ utils/`) omitted. **I agree with all 12 "judged safe" calls**, each verified by reading the target model's field declarations rather than by inference:

- `models/session_lifecycle.py:222` `setattr(session, field_name, value)` — all three live callers (`agent/session_health.py:1235`, `:1314`, `agent/agent_session_queue.py:3209`) pass literal dicts of `priority` / `started_at` / `reprieve_count`. No key field reachable.
- `tools/stage_states_helpers.py:225` and `tools/sdlc_meta_set.py:215` were **not in the PR's list** — both are dynamic-field `setattr` paths. Both are safe: the former's `field` resolves to the literal `"stage_states_json"` or `"stage_states"` at every call site (`tools/sdlc_dispatch.py:208`, `tools/sdlc_verdict.py:541`), the latter is bounded by a fixed three-key `_KEY_REGISTRY` (`plan_revising`, `plan_hash_at_build_start`, `pr_number`).
- `agent/session_runner/adapter.py:572`, `agent/session_health.py:5692`, `models/agent_session.py:2244` — fixed allowlists, confirmed no key fields.
- `models/crash_signature.py:257` — confirmed; the model's only `KeyField` is `signature_hash`.
- `models/chat.py` — confirmed; `KeyField`s are `chat_name` and `chat_type`, not `project_key`.

**New sites in `scripts/` (outside the PR's declared scope), all safe:**
- `scripts/migrate_model_relationships.py:94,107,120,133` — `TelegramMessage`, `Link`, `DeadLetter` all declare `project_key` as a `KeyField`, but every one of those loops is gated on `if <row>.project_key: continue`, i.e. the same empty-only semantic this PR adds.
- `scripts/migrate_memory_project_key.py:128` — `Memory.project_key` is a `KeyField`, and this site mutates it deliberately, then calls `m.save(migrate_key=True)`. **This is the sanctioned rename route, already in use in this repo**, which bears directly on #3247: the silent no-op is not the only available interim.

So the sweep's *conclusion* holds and holds under a wider scope than claimed — but "repo-wide" in the PR body means six directories, and two dynamic-`setattr` sites reached in the widened sweep were not enumerated.

**Tests — reproduced red and green myself**, via `scripts/pytest-clean.sh -n 0` in a clean worktree venv at the PR head.

- Green: `test_session_transcript_keyfield_fork.py` + `test_session_transcript.py` + `test_knowledge_document.py` + `test_knowledge_indexer.py` → **52 passed** in 66s.
- Red: reverting only the two source files (`git checkout HEAD~1 -- bridge/session_transcript.py models/knowledge_document.py`) → **2 failed, 1 passed**, matching the PR's claim.
- `ruff check` and `ruff format --check` on all three changed files: clean.

**Test-quality nuance.** The test is a genuine regression guard, not a path-ran assertion — but the assertions that actually turn it red on the current pin are the `caplog` "slug mismatch" check and `assert row.branch_name == "session/lane-alpha"`, not the row-count and `db_key` assertions. On `popoto==1.9.0` the save raises rather than forking, so `len(rows) == 1` and `db_key == original_db_key` hold with or without the fix. They are correct pin-defensive assertions and worth keeping; they are just not load-bearing today, and the PR body reads as though they are.

**On the silent-no-op rename (#3247).** Acceptable as an interim for `slug`: the path is dormant, and the skip is now logged rather than swallowed. **Not acceptable as-is for `project_key`**, because the divergence case it creates is the doc/chunk/memory split-brain in B1. For that call site the interim must fail closed — skip the whole update, or route the re-key through `save(migrate_key=True)` as `scripts/migrate_memory_project_key.py:131` already does.

**Latency check.** The read-only production probe cited in the PR (0 of 434 `KnowledgeDocument` rows affected) matches the reachability analysis: divergence requires a `projects.json` re-keying, so the leak was latent. That makes B1 a forward-looking correctness fix rather than an incident, but it does not lower its severity — the PR's own justification for touching this file is the isolation property, and the property is not intact end to end.

valorengels:

Round 2, head `916df38af` (rebased onto `2fb7df519`).

### Blockers
- None

**B2 from round 1 — FIXED.** `bridge/session_transcript.py:88-95` and `models/knowledge_document.py:91-105` no longer claim a silent fork. The comment now reads: "slug is a KeyField — reassigning it on a hydrated row raises KeyMutationError under the pinned popoto, and the broad except below swallows that before s.save() runs, discarding this whole transcript-phase update ... Only set it when the existing value is empty ... a genuine rename needs save(migrate_key=True) — tracked in #3247." Verified `popoto==1.9.0` is the installed pin, and confirmed via `git show 2fb7df519:bridge/session_transcript.py` that pre-PR main had the unguarded `s.slug = slug` inside that same broad `try/except Exception` — so the comment's description of the swallow matches the actual shape being guarded.

**TD1 from round 1 — FIXED.** `tests/unit/test_knowledge_rekey_partition.py` (new, 191 lines) exercises the `project_key` half end to end through `index_file` -> `KnowledgeDocument.safe_upsert`, covering the unchanged-content re-key, the changed-content re-key, and a same-key stability check.

### Tech Debt

**1. `bridge/session_transcript.py:96-103` and `:111-118` — the guard-and-warn block is still hand-copied twice** (`slug` and `chat_id`), with hand-varied wording. This is materially smaller than round 1: the third copy in `models/knowledge_document.py` is gone, replaced by a real `migrate_key=True` re-key at `:106-135`, so the duplication surface shrank from 3 sites to 2. The ask — a named helper before a third copy reappears — still stands, but is not a blocker at 2.

### Nits

**1. `bridge/session_transcript.py:98` and `:113` — redundant `str(...)` on both sides of the mismatch comparison.** The preceding `if not s.slug` / `if not s.chat_id` branch already establishes both operands are truthy, and the outer `if slug:` / `if chat_id is not None:` guards already narrow the `str | None` types. Cosmetic, unchanged from round 1.

### Test-shape scrutiny

Three PRs in this batch shipped or nearly shipped a test that passed for the wrong reason, so every new assertion was checked for whether it would actually fail if the fix were reverted.

- `test_non_key_fields_still_update_on_the_same_row` asserts `branch_name`, `classification_type` and `log_path` land on the row after a conflicting-slug call. This is the load-bearing assertion — pre-fix the whole update was discarded by the swallowed `KeyMutationError`.
- `test_conflicting_slug_does_not_fork_the_row` asserts on `db_key` identity and row count, not merely that no exception escaped, so it targets popoto 1.9.0's raise semantics rather than the older fork semantics.
- `test_knowledge_rekey_partition.py` asserts on `_partition()` tuples of `project_key` sets across `KnowledgeDocument` / `DocumentChunk` / `Memory`. It would show a split (`{OLD_KEY, NEW_KEY}` rather than `{NEW_KEY}`) if `_sync_chunks` were called with the wrong `project_key`. A real regression guard, not decorative.

### Fresh read

`tools/knowledge/indexer.py` and `models/knowledge_document.py` read in full. The re-key path is coherent: `index_file` (`:313-339`) computes `project_key_changed` before calling `safe_upsert`, then re-derives the chunk-sync trigger from `doc.project_key` — the value `safe_upsert` actually persisted — rather than the locally-resolved `project_key`, with a comment at `:332-337` explaining why. No dead code or half-migrated branches in either file.

Ran `./scripts/pytest-clean.sh -n 0 tests/unit/test_knowledge_rekey_partition.py tests/unit/test_session_transcript_keyfield_fork.py tests/unit/test_knowledge_document.py` -> 16 passed.

valorengels:

Round 2, head `916df38af` (rebased onto `2fb7df519`).

### Blockers
- None

**B1 from round 1 — FIXED, and fixed the right way.** The patch chose to move everything together rather than fail closed. Traced end to end at `tools/knowledge/indexer.py:328-344`:

- `KnowledgeDocument.safe_upsert(...)` performs the doc-row migration, and this part is genuinely atomic: `models/knowledge_document.py:135` calls `doc.save(migrate_key=rekey)`, and popoto wraps the HSET + delete + index-update in a Redis `MULTI`/`EXEC` (`popoto/redis_db.py:353`, `pipeline(transaction=True)`).
- The caller now passes `doc.project_key` — the persisted value — to `_sync_chunks`, not the locally-resolved `project_key`. This is the exact fix round 1 asked for, matching the pattern already at `indexer.py:599`.
- `doc.scope` is no longer left inconsistent with `doc.project_key`.

On a divergence the doc row, its chunks and its companion memories all land on the same project. The split-brain round 1 identified is closed.

### Tech Debt

**1. The three writes are not atomic with each other, and a partial failure never self-heals.** Filed as **#3256**. `_sync_chunks` (`indexer.py:158-209`) wraps its body in a broad `try/except` that warns and returns; a failure after `DocumentChunk.delete_by_parent(doc_id)` but before the new chunks are created leaves the migrated doc with zero chunks under the new project. On the next pass `project_key_changed` is `False` (the row already carries the resolved key, `:323-325`) and `content_changed` is `False`, so `_sync_chunks` is not re-entered (`:338`). The only recovery, `rechunk_zero_chunk_documents` (`:558-619`), is a manual `tools/doctor.py` diagnostic with no scheduler reference anywhere in the repo.

This is a pre-existing property of `_sync_chunks`'s crash-isolation design, not something the patch introduced — but the patch newly routes cross-project moves through it, which is why it is named here rather than left silent. Not a merge blocker: a failed move makes a document unsearchable, it does not expose it to the wrong project; re-keying requires a `projects.json` change; and round 1's read-only probe found 0 of 434 rows currently divergent.

**2. The KeyField-mutation-skip has no Sentry, metric or retry signal.** `bridge/session_transcript.py:96-103` and `:107-118` log `logger.warning` and move on. `monitoring/sentry_config.py:119,141` shows this repo's `LoggingIntegration` encodes `logger.error` as events at the default `event_level=ERROR`, so a warning produces at most a breadcrumb, never a Sentry issue. The tests assert the log line via `caplog`, which is good regression coverage; production has no equivalent visibility. Carried forward from round 1 (TD2), unchanged, and out of scope for this PR.

**3. `start_transcript` has zero production callers at all** — stronger than round 1's "no caller passes `slug=`". A repo-wide grep for `start_transcript(` outside `tests/` finds only the `def` itself at `bridge/session_transcript.py:44`. Corroborated in-repo by `reflections/improvement_collect.py:34-36`: "AgentSession.log_path is deliberately not read: its only assigner, bridge/session_transcript.py::start_transcript, has no production caller, so a detector keyed on it is structurally always zero." The `slug` guard is correct and cheap, but it hardens a dormant path — worth naming so it is not mistaken for closing a live production gap. The reachable half of #3210 is the `KnowledgeDocument` one, and that half is now closed end to end.

### Nits
- None

### Re-key durability under adversarial conditions

Checked re-entrancy, crash-mid-move, and a document already present in both projects. Steady state is correct; the only durable failure mode is the interrupted-move case above, now tracked in #3256.

### Deploy note
`bridge/` is touched: this requires `./scripts/valor-service.sh restart` on every bridge machine after deploy. Rollback is clean — the changes are additive guards plus a re-key path, with no schema or data migration.

valorengels:
## Review: APPROVED

**Mode:** independent roster (2 judges) — `code-quality` and `risk`, dispatched with the same PR context and no visibility into each other's reasoning. Consensus rule: any-blocker-wins, k=2/n=2, no quorum shortfall, not tied. Both returned APPROVED with zero blockers. Head SHA reviewed: `916df38af`.

Round 1 (head `0a2cd69af`) raised 2 blockers and 4 tech-debt items. **Both blockers are verified fixed**, each by a judge reading the current code rather than trusting the patch commit's description.

- **B1** — the `KnowledgeDocument` half left chunks and companion memories moving to the new project while the doc row stayed on the old one. The patch chose to move all three together. Verified end to end: `index_file` now passes `doc.project_key` (the persisted value) rather than the locally-resolved `project_key` to `_sync_chunks`, `safe_upsert` routes a genuine re-key through `doc.save(migrate_key=rekey)` which popoto wraps in a Redis `MULTI`/`EXEC`, and `doc.scope` is no longer left inconsistent with `doc.project_key`.
- **B2** — the inline comments described a silent fork that the pinned `popoto==1.9.0` does not produce. They now describe the actual `KeyMutationError` raise and the swallowed-exception consequence that motivated the fix.

### Rebase and test evidence

The PR was rebased from its original merge-base `888cfab1f` onto current `main` (`2fb7df519`). The rebase was textually clean, and `git log 888cfab1f..2fb7df519` over the four non-test files this PR touches is **empty** — main never touched them in the skipped range, so there is no semantic-conflict surface.

**Green:** 100 passed via `scripts/pytest-clean.sh -n 0` across `test_session_transcript_keyfield_fork.py`, `test_session_transcript.py`, `test_knowledge_document.py`, `test_knowledge_indexer.py`, `test_knowledge_rekey_partition.py`, `test_knowledge_converter.py`, `test_knowledge_watcher.py`.

**Red replayed on the rebased base**, not inherited from the PR body. The PR's original red run was proven against the old merge-base, and stale red evidence is exactly what let three PRs in this batch ship a test that passed for the wrong reason. Reverting only the three source files to `2fb7df519` and re-running the new tests yields **4 failed, 36 passed**, with the pinned popoto's actual error:

```
FAILED test_session_transcript_keyfield_fork.py::test_conflicting_slug_does_not_fork_the_row
FAILED test_session_transcript_keyfield_fork.py::test_non_key_fields_still_update_on_the_same_row
FAILED test_knowledge_rekey_partition.py::test_rekey_with_unchanged_content_moves_chunks_and_memories
FAILED test_knowledge_rekey_partition.py::test_rekey_with_changed_content_moves_chunks_and_memories

WARNING bridge.session_transcript: Failed to update/create AgentSession for test-keyfield-fork-...:
  Cannot change KeyField 'slug' from 'lane-alpha' to 'lane-beta' without migrate_key=True.
WARNING models.knowledge_document: KnowledgeDocument upsert failed (non-fatal):
  Cannot change KeyField 'project_key' from 'test-3210-old' to 'test-3210-new' without migrate_key=True.
```

Both defect classes are covered by assertions that genuinely go red. `ruff check`, `ruff format --check` and `uv lock --locked` all clean.

### Accepted tech debt (3)

None of these block the merge; all are stated rather than omitted.

- **TD1 — the re-key is not atomic across doc, chunks and memories, and a partial failure never self-heals.** Filed as **#3256**. Pre-existing behavior of `_sync_chunks`, which this patch newly routes cross-project moves through. Impact is a document becoming unsearchable, not a cross-project leak; reachability requires a `projects.json` re-keying, and a read-only probe found 0 of 434 rows currently divergent.
- **TD2 — no Sentry, metric or retry on the mutation-skip.** `logger.warning` produces a breadcrumb, not a Sentry event, under this repo's `LoggingIntegration` config. Carried forward from round 1, out of scope here.
- **TD3 — the guard-and-warn block is hand-copied twice** in `bridge/session_transcript.py` (down from three sites; the `knowledge_document.py` copy was replaced by the real `migrate_key=True` path). Worth a named helper before a third copy appears.

### Nit (1)

Redundant `str(...)` on both sides of each mismatch comparison, where the preceding `if not X` branch already established both operands are truthy.

### Deploy note

`bridge/` is touched, so this requires `./scripts/valor-service.sh restart` on every bridge machine after deploy. Rollback is clean: additive guards plus a re-key path, no schema or data migration.

### Disclosure

This round-2 review was run by the merge shepherd driving this PR through the gate. The shepherd authored no code in this PR — both commits predate this round — and did not write any of the findings above; the two judges were dispatched independently and their unedited reviews are in the two comments preceding this one.

valorengels:
## Review (docs delta): APPROVED

Independent review of the documentation-only delta `916df38af..bdb71661f`, dispatched separately from the two code judges.

**Disclosure:** this delta was authored by the merge shepherd's documentation cascade — the same actor driving this PR through the merge gate. It was therefore reviewed by a judge dispatched specifically to read it adversarially, with that authorship stated in its prompt. The code in this PR (`b6a520e1a`, `916df38af`) was not authored by the shepherd.

Every factual claim was verified against source rather than accepted as written:

| Claim | Verdict | Evidence |
|---|---|---|
| `slug` is a `KeyField` (the doc previously called it a plain `Field`) | Confirmed | `models/agent_session.py:398` `slug = KeyField(null=True)` |
| Pinned popoto raises `KeyMutationError` rather than forking | Confirmed | `popoto==1.9.0` installed; raise site `popoto/models/base.py:1247`, guarded by `if not migrate_key and self._saved_field_values` |
| `safe_upsert` routes divergence through `doc.save(migrate_key=True)` | Confirmed | `models/knowledge_document.py:134-135` |
| `scope` moves with `project_key` so the two cannot disagree | Confirmed | `models/knowledge_document.py:134` is the only production write site for `.scope`, always paired with the `project_key` assignment; asserted by `tests/unit/test_knowledge_rekey_partition.py:163` |
| `index_file` reads back `doc.project_key`/`doc.scope`, not its own locals | Confirmed | `tools/knowledge/indexer.py:336`, `:344` |
| A re-key forces a chunk resync even with unchanged content | Confirmed | `tools/knowledge/indexer.py:315-324`, `:337` |
| `session-isolation.md` `KeyMutationError` correction | Confirmed | consistent with the verified popoto behavior |
| Anchor links resolve | Confirmed | `redis-models.md:186` and `knowledge-document-integration.md:75`, both slugs match |
| #3247 and #3256 exist and match the prose | Confirmed | both OPEN, content matches |
| Overreach check on absolute claims | No overreach | "can never disagree" holds given the single call site |

### Blockers
- None

### Nits — one fixed, one referred

**1. Historical narration — FIXED in `bdb71661f`.** The judge flagged a parenthetical in `redis-models.md` ("older pins silently forked the row instead...") as the one spot in the delta naming prior behavior rather than the current contract, against CLAUDE.md's NO LEGACY CODE TOLERANCE / "describe only the new status quo". Removed. The two other narration hits in `session-isolation.md` (`:228` "Before this fix", `:340` "Previously") are pre-existing prose about unrelated issues #3176 and #232, and were deliberately left alone rather than swept into this PR.

**2. Issue #3247's framing is now stale — referred, not changed here.** #3247's body says "the observable damage today is a silent total loss of the entire transcript-phase update", but the guard in `b6a520e1a` (`bridge/session_transcript.py:96-103`) already prevents that: the mismatch branch never assigns, so `KeyMutationError` is never raised and `s.save()` proceeds with the other fields intact. The docs delta describes the *current* code correctly; it is #3247's own description that this PR supersedes. Noted on that issue for whoever picks it up.

Issue #3272, opened 2026-09-09
AgentSession save() hardening sweep


valorengels:
Follow-up carved out of the plan for #3270 (`docs/plans/telegram-reply-duplication-and-room-register.md`, task **A3b**).

#3270 fixed the **two evidenced** bare-`save()` sites on the incident path (`agent/output_handler.py::_persist_routing_fields` and `::_append_rtr_event`, task A3a) and deliberately deferred the rest. This issue is the rest — the latent hazard class — plus **A6a** (deferred rate-limit routing), folded in per that plan's round-2 critique rather than filed separately.

## Architectural rationale

From #3270's plan, *Architectural Impact*:

> **A bare `save()` on an AgentSession is a lifecycle write.**

Concretely:

- popoto's `Model.save(update_fields=None)` path (`.venv/lib/python3.14/site-packages/popoto/models/base.py:1136` signature, `:1514` `hset_mapping = encode_popoto_model_obj(self)`) HSETs the **entire encoded instance**, then runs `field.on_save()` for **every** field (`:1565`-`:1567`).
- `status` is an `IndexedField` (`models/agent_session.py:167`), so that full write rewrites both the hash **and** the secondary index.
- It emits **no LIFECYCLE log and no `session_events` entry**.

So any code that reads a row, mutates one unrelated field, and calls `save()` is silently authorized to rewrite the session's lifecycle state from a stale snapshot, with no audit trail. That is exactly what stranded `tg_valor_-1003449100931_1450` in #3270: the row read `completed` while the harness was mid-turn, the turn-end finalize hit the idempotency skip, a later stale save rewrote `running`, and the orphan net re-queued it — one unprompted Telegram reply per re-run.

The prescription is `save(update_fields=[...])` at every incidental-field write site.

**Doc of record:** `docs/features/agent-session-lifecycle-writes.md` (created by #3270, task D1) documents the popoto full-HSET semantics and the `update_fields` convention. Read it before touching any site below.

## Prescription

**One commit per site**, on their own review pass. Narrowing a save changes which fields persist, and a site that was accidentally relying on the full write will break in a way unit tests may not catch — per-site commits keep it bisectable. Name the now-written field set in each commit message.

Two traps recorded in #3270's *Race Conditions* section that apply here:

[...]

Issue #3253, opened 2026-09-08
Worker loop dies on terminal-status conflict: unguarded retry at agent_session_queue.py:3028 strands every session for the worker_key


valorengels:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/worker-loop-terminal-status-conflict.md

## Summary

`_worker_loop`'s outer session-completion `finally` in `agent/agent_session_queue.py` can raise `StatusConflictError` out of `_worker_loop` entirely, killing the worker task and stranding every session queued for that `worker_key`.

Two independent defects combine:

1. **No already-terminal skip.** The completion block's only skip conditions are `not fresh` (`:3002`) and `fresh.status == "pending"` (`:3010`). If the row is *already terminal* — because `_execute_agent_session`'s finalize guard, `complete_transcript`, or the health checker wrote it first — the block falls through to `:3019` `await _complete_agent_session(session, failed=session_failed)`. `_complete_agent_session` (`agent/session_completion.py:160-162`) calls `finalize_session(session, "failed"|"completed")`, and `finalize_session` raises `StatusConflictError` on terminal → *different* terminal (`reject_from_terminal=True` by default; documented at `models/session_lifecycle.py:311-318`).

2. **The retry is outside any try block.** That first raise is caught by `except Exception as guard_err:` at `:3020`, whose handler logs "— completing session as fallback" and then **retries the identical call at `:3028`, unguarded**. The second `StatusConflictError` propagates out of the `finally`, past the `while`, past `:3051 except asyncio.CancelledError` (which does not catch it), out of `_worker_loop`. `:3054` pops the task from `_active_workers` and nothing re-creates it for that `worker_key`.

The `except`-then-retry-the-same-call shape at `:3028` is the sharper of the two: the handler exists to recover from a *read* failure in the nudge guard (`AgentSession.query.get`), but its `try` also spans the `_complete_agent_session` call in the `else` branch, so a write conflict is retried verbatim rather than recovered from.

## Reachable trigger

The clearest one: `_execute_agent_session` returns normally (its finalize guard writes `completed`), and the worker's own post-`await exec_task` code then raises, setting `session_failed = True` (`agent_session_queue.py:2934-2937`). The `finally` reads a terminal `completed` row and tries to write `failed` → conflict → retry → worker death.

Terminal statuses written out-of-band by the health checker (`killed`, `abandoned`) reach the same collision.

## Consequence

[...]

Issue #3198, opened 2026-09-06
investigation: review judge subagents fail to spawn from the forked /do-pr-review context


tomcounsell:
> **Context:** The SDLC REVIEW stage dispatches judge subagents from the [`/do-pr-review`](https://github.com/tomcounsell/ai/blob/main/.claude/skills-global/do-pr-review/SKILL.md) skill, which itself runs with `context: fork`. The judges' findings feed `compute_consensus` to produce the merge gate's verdict.

## Problem

On PR #3174 (merged 2026-09-05), two of three review rounds ran **single-judge instead of the declared two**. Both forked reviewers disclosed the same reason in prose: they lacked a subagent-spawn tool in that context. The cause is unconfirmed, and the correct fix differs by an order of magnitude depending on which it is.

**Why this is filed as `investigation`, not `bug`:** the reported symptom and the declared configuration contradict each other, so the mechanism needs establishing before any change.

## What is established

- [`/do-pr-review/SKILL.md`](https://github.com/tomcounsell/ai/blob/main/.claude/skills-global/do-pr-review/SKILL.md) frontmatter **does** grant the tool: `allowed-tools: mcp__byob__*, Bash(gh:*), Bash(git:*), Bash(python:*), Bash(jq:*), Bash(sdlc-tool:*), Read, Write, Edit, Grep, Glob, Agent`.
- The same frontmatter declares `context: fork`, so the skill body executes inside a forked agent rather than the parent session.
- [`docs/sdlc/do-pr-review.md:260`](https://github.com/tomcounsell/ai/blob/main/docs/sdlc/do-pr-review.md) requires the judges run in the foreground and be awaited in-turn.
- So the tool is granted on paper, and the failure occurs at nested spawn from inside a fork.

## Candidate mechanisms, none confirmed

1. **Nested spawn from a forked context is restricted.** A forked agent may be unable to spawn further subagents regardless of `allowed-tools`. This presents to the running agent as an absent tool.
2. **A spawn-parameter defect.** There is a prior recorded instance where passing a `name` argument to a spawn produced a refusal that read as a missing capability, and rosters worked once the argument was dropped. If this is the cause it is close to a one-line fix.
3. **`allowed-tools` filtering interaction.** The allowlist may be applied differently to a forked child than to the parent, dropping `Agent` in practice.

## Why it matters

[...]

PR #3273, opened 2026-09-09
Stop the unprompted Telegram repeat replies, and make replies match the room

### LINKED ISSUE #3270: bug: SDLC stall-recovery resumes finished slugless Telegram threads, producing repeated unprompted replies
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/telegram-reply-duplication-and-room-register.md

> **Context for a reader new to this repo.** Valor is an AI teammate that talks to humans over Telegram. Each Telegram thread is backed by an *AgentSession* row (`models/agent_session.py`) that moves through a status lifecycle (`pending` -> `running` -> `completed`/`failed`/...). A separate hourly-ish background job, the **SDLC stall check** (`reflections/sdlc_progress.py`), watches for *lanes* — units of planned engineering work, each identified by a **slug** — that have stopped making progress, and tries to nudge an engineering session back to life. Conversation threads in Telegram have **no slug**; only planned SDLC work does. This issue is about the stall check picking up slugless conversation threads and re-running them.

## Problem

Humans in a Telegram group see Valor post extra replies to messages that were already answered, sometimes for days after the original exchange ended. Nobody asked for these replies. From the room's point of view Valor looks like it is talking to itself.

### Current behavior

`_pick_steer_target` (`reflections/sdlc_progress.py:808-858`) implements a ladder for choosing which session to nudge when a lane is stalled. Rung 2 falls back to *resuming* a terminal session:

```python
elif status in RESUMABLE_STATUSES and getattr(row, "claude_session_uuid", None):
    resumable.append(row)
...
if resumable:
    return ("resume", max(resumable, key=_rank))
```

`_rank` prefers a slug match but does not *require* one:

```python
def _rank(row) -> tuple[int, float]:
    """Same-lane first, then most recently updated."""
    same_lane = 1 if (lane_slug and getattr(row, "slug", None) == lane_slug) else 0
    return (same_lane, to_unix_ts(getattr(row, "updated_at", None)) or 0.0)
```

So when no session carries the stalled lane's slug, the *most recently updated eng session in the project* wins — and that is routinely a slugless human conversation thread. `RESUMABLE_STATUSES` (`models/session_lifecycle.py:71`) is `{"completed", "killed", "failed", "abandoned"}`, so a thread that already finished, or that previously *failed*, stays eligible forever.

Observed in `logs/reflection_worker_error.log*`, Telegram thread `tg_valor_-1003449100931_1435` (a slugless Eng-chat conversation) resumed on a ~30-minute cadence for hours:

```
2026-09-04 18:59:48 [steering] Pushed message to steering:room:valor|telegram:-1003449100931: '[Prior session context: <resumed-session-context> ...' (from resume:sdlc-stall)
2026-09-04 18:59:48 LIFECYCLE session=tg_valor_-1003449100931_1435 transition=completed->pending ... context="resume (sdlc-stall)"
2026-09-04 19:29:51 LIFECYCLE session=tg_valor_-1003449100931_1435 transition=completed->pending ... context="resume (sdlc-stall)"
2026-09-04 19:59:54 LIFECYCLE session=tg_valor_-1003449100931_1435 transition=completed->pending ... context="resume (sdlc-stall)"
2026-09-04 20:29:56 LIFECYCLE session=tg_valor_-1003449100931_1435 transition=completed->pending ... context="resume (sdlc-stall)"
2026-09-05 01:31:10 LIFECYCLE session=tg_valor_-1003449100931_1435 transition=completed->pending ... context="resume (sdlc-stall)"
```

The same pattern hit threads 1450 and 1415 and the system session `0_1788554456715`. Each resumed run is a full PM turn that ends by delivering a message to the room, so each one is a visible extra reply.

Two mechanisms make it worse:

1. **The resume steer is room-scoped, not session-scoped.** `tools/valor_session.py:1155-1157` pushes to `steering:room:{room_id}`:
   ```python
   push_steering_message(
       session_id, outbound, f"resume:{source}", room_id=room_id_for_session(session)
   )
   ```
   A room-scoped steer is drained by *whichever session next serves that room*, so a resume aimed at one session can be consumed by an unrelated one. The transition also passes `reject_from_terminal=False` (`:1162-1164`), which is what allows a terminal row to be revived at all.

2. **A finished run can land back in `running` and get re-enqueued.** After a resumed run completes, rows have been observed sitting in `running` with no live worker handle. The `no_progress` recovery net (`agent/session_health.py:4734-4749`) then bounces them back to `pending` for another full turn, up to `MAX_RECOVERY_ATTEMPTS = 2` (`:468`), after which the row is finalized `failed` — which makes it eligible for the stall resume again. 35 `no_progress` recoveries appear across the retained `logs/worker.log.*` window.

   There *is* already a guard against this: `_delivery_belongs_to_current_run` (`:411-424`) finalizes as `completed` instead of re-queuing when `response_delivered_at >= started_at`. It fires at `:4649` and `:3417`. It did not save these rows, and **why** is the open question below. Note it never consults `completed_at`.

3. **Every re-run is forced to say something.** The wrap-up guard at `agent/session_runner/runner.py:1794-1797` emits `OPERATOR_TERMINAL_MESSAGE` whenever a turn produced no user-facing message:
   ```python
   if not self._adapter.user_facing_routed:
       self._adapter.on_user_payload(OPERATOR_TERMINAL_MESSAGE)
   ```
   So a re-run with genuinely nothing to add still posts. Before `988fac09a` this surfaced as the recurring "I wasn't able to produce a response" and "You've hit your session limit" messages in chat. The latter is raw CLI rate-limit text relayed verbatim as a chat reply: `agent/session_runner/harness/claude.py` has **no** rate-limit detection at all (grep for `session limit`, `usage limit`, `rate limit`, `429`, `quota` returns zero matches in the whole `agent/session_runner/` tree), so a usage-limit stop falls through as a generic failure and its text reaches the human.

4. **The teammate timeout notice has no dedupe.** `TIMEOUT_NEEDS_ATTENTION_MESSAGE` (`agent/session_runner/runner.py:244`) is emitted at `:861` once per *run*, guarded only by a `break` in the turn loop. Nothing suppresses it across runs, so a session that gets re-enqueued re-delivers the identical text.

### Desired outcome

A finished Telegram conversation thread is never re-run on behalf of unrelated stalled engineering work, and a re-run that has nothing to say does not manufacture a chat message. Concretely: zero unprompted `resume (sdlc-stall)` transitions on slugless sessions, and no repeated identical operator/timeout notices in a room.

## Definitions

| Term | Definition |
|---|---|
| **AgentSession** | Redis-backed row (`models/agent_session.py`) representing one conversational or engineering session and its lifecycle status. |
| **Lane** | A unit of planned SDLC work, identified by a kebab-case **slug** that ties together a task list, a `session/{slug}` branch, and a `.worktrees/{slug}/` worktree. See [`docs/features/sdlc-lane-identity.md`](docs/features/sdlc-lane-identity.md). |
| **Slugless session** | A session with `slug is None` — i.e. a human conversation thread, not planned work. |
| **Stall check** | The periodic job in [`reflections/sdlc_progress.py`](reflections/sdlc_progress.py) that detects lanes with no recent progress and steers/resumes/creates an eng session. Introduced by `6a9e2b66c` (#2710). |
| **Steer / steering list** | Redis list the worker drains at turn boundaries to inject an instruction into a session. Keyed per-session (`steering:{session_id}`) or per-room (`steering:room:{room_id}`). See [`docs/features/session-steering.md`](docs/features/session-steering.md). |
| **`no_progress` net** | Branch of the session health check (`agent/session_health.py`) that re-queues a `running` row showing no tool/turn/heartbeat activity. Added by #944. |
| **Wrap-up guard** | Logic at the end of a runner turn that guarantees the human receives *some* text (`agent/session_runner/runner.py:1745-1799`). |

## Solution sketch

The dominant fix is narrowing rung 2 of the ladder. The rest are containment.

1. **`_pick_steer_target` rung 2: require a lane match.** Only resume a session whose `slug` is non-`None` and equal to `lane_slug`. Never resume a slugless session. Never resume a `failed` row (a row that failed once will fail the same way again; re-resuming it hourly is a loop, not a recovery). Fall through to the `create` rung otherwise.
2. **`resume_session`: make the resume steer session-scoped.** A resume targets one specific session row; the room-scoped push lets an unrelated session consume it.
3. **Stop the orphan re-enqueue.** A row whose current run already delivered, or that carries a `completed_at`, should be finalized `completed` rather than bounced to `pending`. **Open question for the planner:** the existing `_delivery_belongs_to_current_run` guard should already have caught these rows. Determine empirically why it did not before changing it — the two candidate explanations are (a) `started_at` being refreshed by the re-enqueue path so the epoch comparison fails, and (b) `response_delivered_at` being clobbered back to `None`. Do not patch the symptom until this is answered.
4. **Find the writer that restores `status=running` after finalization.** The leading suspect is `agent/sdk_client.py:408-413`, a full-row `session.save()` with no `update_fields` in `_store_claude_session_uuid`, which writes back its whole in-memory snapshot and can resurrect a stale status. `3c77e1eab` is the first commit to bisect. There are sibling full-row saves with the same hazard shape at `agent/health_check.py:654`, `agent/output_handler.py:1469` and `:1526`, and `tools/valor_telegram.py:791`; the repo already documents this hazard class at `agent/session_health.py:2872-2878`. **Confirm the actual writer before fixing**, and prefer a narrow `update_fields=[...]` write over adding a new guard.
5. **Never relay raw CLI rate-limit text as a chat reply.** Detect it and route it to the operator once instead of to the room.
6. **Dedupe the teammate timeout notice** to at most once per session.

### Testing note

Per this repo's testing discipline, each guard must be proven RED against the known-bad behavior before the fix lands: reproduce the resume-of-a-slugless-thread and the orphan re-enqueue as failing tests first, then assert the fix turns them green. A guard that was never red proves nothing.

## Acceptance criteria

- [ ] `_pick_steer_target` returns `("create", None)` rather than `("resume", row)` when the only resumable candidates are slugless or belong to a different lane; covered by a test that fails on the current code.
- [ ] `_pick_steer_target` never returns a `failed` row from rung 2.
- [ ] `resume_session` pushes the resume steer to the session-scoped steering key, not the room-scoped one; covered by a test asserting the key.
- [ ] The root cause of the post-finalization `status=running` flip is identified, named in the PR description with evidence, and fixed at the writer.
- [ ] A session whose run already delivered (or that has `completed_at`) is finalized `completed` instead of re-enqueued to `pending`; covered by a test that reproduces the bounce first.
- [ ] Raw CLI rate-limit / usage-limit text is never delivered as a chat reply; it reaches the operator once.
- [ ] The teammate timeout notice is delivered at most once per session.
- [ ] Narrow-scope tests pass (`scripts/pytest-clean.sh` on the touched test modules only, never a full suite).

## Downstream context

Related: #2710 (introduced the stall auto-resume ladder), #944 (`no_progress` net), #918 (delivered-but-not-finalized guard).

## Recon Summary

Verified against `main` @ `e73e25581` (2026-09-09). Two read-only Explore sweeps plus a dedicated root-cause investigation (see the follow-up comment on this issue for the full forensics). Every file:line cited above resolves exactly unless noted.

**Confirmed**

- `reflections/sdlc_progress.py:808-858` — rung 2 collects any row with `status in RESUMABLE_STATUSES` and a `claude_session_uuid`; `_rank` treats a slug match as a tiebreaker (`same_lane = 1 if ... else 0`), not a filter. `models/session_lifecycle.py:71` — `RESUMABLE_STATUSES = frozenset({"completed", "killed", "failed", "abandoned"})`, i.e. `TERMINAL_STATUSES` minus `cancelled`.
- `tools/valor_session.py:1155-1157` — the resume steer is pushed with `room_id=room_id_for_session(session)`. `agent/steering.py:313` selects `_room_queue_key(room_id)` whenever `room_id` is truthy and the steer is not an abort; `:82` builds `f"steering:room:{room_id}"`. So the push is room-scoped. `:1162-1164` passes `reject_from_terminal=False`.
- `agent/session_health.py:4734-4749` — the `no_progress` branch fires on `in_scope_handle is None` and `running_seconds > AGENT_SESSION_HEALTH_MIN_RUNNING` (300s, `:443-445`) and `not _has_progress(entry)`. `_has_progress` consults only `last_tool_use_at` / `last_turn_at` / `last_heartbeat_at` — no reference to `response_delivered_at` or `completed_at` in its body. `MAX_RECOVERY_ATTEMPTS = 2` (`:468`); on exhaustion `_apply_recovery_transition` finalizes `failed` (`:3698-3716`), otherwise it requeues (`:3841-3842`, `:3934`, `:3940-3948`).
- `agent/session_runner/runner.py:1794-1797` — the wrap-up guard emits `OPERATOR_TERMINAL_MESSAGE` when `not self._adapter.user_facing_routed`. `:244` / `:861` — `TIMEOUT_NEEDS_ATTENTION_MESSAGE` is emitted once per run, guarded only by the `break` at `:863`; there is no dedupe key of any kind (no SETNX, no session flag), so a re-enqueued session re-delivers it.
- `agent/session_runner/harness/claude.py` — no rate-limit detection exists. Case-insensitive grep for `session limit`, `usage limit`, `rate limit`, `limit reached`, `quota`, `retry-after`, `429`, `529`, `overload` returns zero matches in that file and zero across the whole `agent/session_runner/` tree. The only textual sentinels are `IMAGE_DIMENSION_SENTINEL` (`:173-183`) and `THINKING_BLOCK_SENTINEL` (`:191-202`), both unrelated. `:658-659` records the design stance against substring matching on stderr.
- Live log evidence for the resume loop, `logs/reflection_worker_error.log*` (timestamps local PDT): thread `tg_valor_-1003449100931_1435` transitioned `completed→pending` with `context="resume (sdlc-stall)"` at 18:59:48, 19:29:51, 19:59:54, 20:29:56 on 2026-09-04 and repeatedly thereafter, each preceded by a push to `steering:room:valor|telegram:-1003449100931`. The system session `0_1788554456715` shows the same pattern. 35 `no_progress` recoveries across the retained `logs/worker.log.{1,2,3}`.

**Revised (the follow-up comment supersedes parts of the Solution sketch)**

- The `status=running` writer is **not** `agent/sdk_client.py:413` for this incident — `git log -L 394,421:agent/sdk_client.py` attributes that bare save to `a7bf22431` (#377), not `3c77e1eab`. The mechanism (full-row `save()` rewriting the `status` IndexedField with no ledger entry) is confirmed at `popoto/models/base.py:1514`/`:1565`; the specific writer is PLAUSIBLE at `agent/output_handler.py:1468` and `:1526`.
- `_delivery_belongs_to_current_run` did not fail on its epoch logic. `response_delivered_at` is **never stamped** on this path: its only two non-test writers (`agent/session_executor.py:1736-1737`, `agent/session_completion.py:1117-1118`) fired zero times across the entire retained log window, because delivery actually flows through `flush_deferred_self_draft_sync`, which RPUSHes to `telegram:outbox:` without stamping. The guard is starved of input, not wrong.
- Sketch item 3 (add a `completed_at` check to the orphan net) is **dropped**; fix the stamp instead.

**Pre-requisites / coordination**

- #3253 (open, plan skeleton at `docs/plans/worker-loop-terminal-status-conflict.md`) addresses terminal-status conflicts at a different site (`agent/agent_session_queue.py:3019`/`:3028`). Distinct call sites and distinct files, but both touch terminal-write semantics; `models/session_lifecycle.py` is the one file both may edit. Coordinate rather than merge.

**Dropped / out of scope**

- Loosening `_delivery_belongs_to_current_run` to bare field-presence — re-opens the #918 duplicate-delivery bug.
- Suppressing the #944 orphan net — it fired correctly on a genuinely orphaned `running` row; suppressing it hides the corruption.

### LINKED ISSUE #3271: Valor's chat replies don't match the room: no reply-length or formatting guidance in the operative primes
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/telegram-reply-duplication-and-room-register.md

> **Context for a reader new to this repo.** Valor is an AI teammate that participates in human Telegram group chats. Its conversational behavior is set by a **prime** — a Markdown file prepended to the session's first turn by [`agent/session_runner/role_driver.py`](agent/session_runner/role_driver.py). Two primes matter here: `prime-pm-role.md` (used for every `eng` session, i.e. all normal chat) and `prime-teammate-role.md`. This issue is about those files containing no guidance on *reply length or formatting*, with the result that Valor writes multi-hundred-word structured documents into a chat where humans write one line.

## Problem

Valor's replies do not match the register of the room. Humans in the chat write short, plain messages; Valor answers with headers, numbered lists, and several hundred words. The human has already told it so directly — *"Don't send me messages like this. I'm not your log file."*

### Current behavior

Measured over chat `-1003449100931` (241 rows):

| | Human messages | Valor's agent-composed replies |
|---|---|---|
| Median length | 8 words | ~247 words |
| Median lines | 1 | multi-paragraph |
| Markdown `#` headers | 0 | 17 of 70 |
| Numbered lists | 0 | 18 of 70 |
| Longest | — | 1029 words, in reply to a 17-word question |

The cause is an absence, not a bug. Nothing in the operative prompt tells the agent how long a chat reply should be.

**`.claude/commands/roles/prime-pm-role.md` is the only operative style text.** It is prepended on turn 1 (`role_driver.py:379-394`), and `runner.py:557` selects it for every non-teammate session:
```python
role = "teammate" if (self._session_type or "").lower() == "teammate" else "pm"
```
It has a `# Persona behaviors to keep` section (line 76) whose closest existing guidance is:
- `:78` — "Concise. The developer is the executor; you are the router." (about *developer instructions*, not chat replies)
- `:79` — "**Trivial messages get a one-line ack, then you stop.**" (only covers pleasantries and status updates)

There is **no** guidance on formatting, no length proportionality rule, and nothing about reading the room. `prime-teammate-role.md` is the same: `:41-43` cover tone and trivial acks, nothing else.

**The one file with the right instruction is not injected.** `config/personas/segments/identity.md:46-49` says exactly the right thing:
```
My messages are **direct** (state what I did, need, or found — no preamble), **concise**
(short; longer only when requested), **professional**, and **contextual** ...
```
But identity.md is assembled only by `compose_system_prompt` (`agent/sdk_client.py:890`), whose sole callers are `load_system_prompt` (`:1082`) and `load_eng_system_prompt` (`:1127`) — **and neither is called by the session runner or the bridge**. Outside `sdk_client.py` the function appears only in tests and one docstring. So identity.md reaches no runtime session prompt today.

**Two docs assert a safety net that no longer exists.** They claim an LLM drafter condenses long output before it reaches chat:
- `config/personas/segments/identity.md:57-59` — "Long outputs are condensed by the message drafter (`bridge/message_drafter.py`, Haiku)..."
- `config/personas/segments/identity.md:39-40` — an HTML comment pointing at `DRAFTER_SYSTEM_PROMPT`, a constant that no longer exists anywhere in `bridge/` or `agent/`.
- [`docs/features/pm-voice-refinement.md`](docs/features/pm-voice-refinement.md) `:9`, `:17`, `:21`, `:29`, `:52` — repeatedly describes `DRAFTER_SYSTEM_PROMPT` "instructing the LLM" to naturalize, shorten, and suppress metrics. It also names `_parse_summary_and_questions()`, which does not exist (the real function is `_parse_draft_and_questions`, `bridge/message_drafter.py:1073`), and attributes `_truncate_at_sentence_boundary()` to `bridge/response.py` when it lives at `bridge/message_drafter.py:64`.

`bridge/message_drafter.py` is validation-only. Its own docstring (`:1-17`): *"No LLM rewriting of the agent's output. The drafter's job is validation + structural composition, not summarization."* Over-length output is **attached as a `.txt` file**, never summarized. So the documented safety net is fictional and nothing shortens anything.

### Desired outcome

Valor's chat replies read like they belong in the room: a few sentences of prose, no headers or numbered lists unless a list was asked for, length proportional to the ask. Long-form analysis still gets produced — it goes to a file with a short caption instead of into the message body. The docs stop describing a condensing drafter that does not exist.

## Definitions

| Term | Definition |
|---|---|
| **Prime** | Markdown file prepended to a session's first turn to establish role and behavior. Lives in `.claude/commands/roles/`. Selected per `session_type` by [`agent/session_runner/role_driver.py`](agent/session_runner/role_driver.py). |
| **PM role / eng session** | The default session type for chat; uses `prime-pm-role.md`. See [`docs/features/eng-session-architecture.md`](docs/features/eng-session-architecture.md). |
| **Message drafter** | [`bridge/message_drafter.py`](bridge/message_drafter.py). Validates wire format and composes structure. Repositioned from summarizer to pass-through by #1680. |
| **Promise gate** | [`bridge/promise_gate.py`](bridge/promise_gate.py). Blocks messages containing unsubstantiated forward commitments ("I'll report back when tests finish"). |
| **`SELF_DRAFT_INSTRUCTION`** | `bridge/message_drafter.py:1025-1033`. Text sent back to the agent when the validator rejects its message, telling it how to rewrite. Currently mandates "2-4 bullet points" — in tension with the prose-in-chat goal this issue asks for. |
| **`file_paths`** | The mechanism by which a reply attaches a file instead of inlining long content. |

## Solution sketch

Prompt and documentation change; no runtime logic change.

1. **Add a "Match the room" section to `prime-pm-role.md`**, under the existing `# Persona behaviors to keep` heading, and an equivalent to `prime-teammate-role.md`. Substance:
   - Chat replies are a few sentences of prose. No headers, no bold, no numbered lists unless the human asked for a list.
   - Length is proportional to the ask; a one-line question gets a one-to-three-line answer.
   - Long-form analysis goes to a file via `file_paths` with a two-sentence caption in the message.
   - Before a non-trivial group reply, read the recent room history with `valor-telegram read --chat-id <id>` and write at the humans' length and register.
   - Keep the load-bearing specifics — commit hashes, PR and issue numbers, verdicts. Drop process narration.
2. **Fix the stale docs.** Correct the drafter claims in `config/personas/segments/identity.md` (`:39-40`, `:57-59`) and `docs/features/pm-voice-refinement.md`, and fix the two wrong symbol references in the latter. Per this repo's no-legacy rule, describe only the current status quo — do not document the removal.
3. **Reconcile `SELF_DRAFT_INSTRUCTION`** with the new guidance; its "2-4 bullet points" mandate currently pushes in the opposite direction.

### Constraints the implementer must respect

- `tests/unit/test_pm_progress_updates.py` asserts exact substrings of `prime-pm-role.md` (the `# Progress updates when the work overruns the ask` header, `"say only what is already true"`, `"a dispatch you can execute, you execute"`, `"turn boundaries"`, `"#2420"`, `"ask_coverage"`, and others) and bans the four strings in `PHRASING_WORKAROUND_STRINGS` (`:80-85`). **Add a section; do not rewrite existing paragraphs or change headers.**
- New prompt text must not offer "I'll follow up later" as a way to shorten a reply — the promise gate blocks exactly that phrasing (`TAUGHT_BLOCKED`, `tests/unit/test_pm_progress_updates.py:76`).
- Prime files are hardlinked fleet-wide. **Edit in place; never replace-and-rename** (that breaks the hardlink).
- Do not touch `CLAUDE.md` headings — they are regex-parsed into worker prompts and asserted byte-for-byte.
- Prefer placeholders over hard numbers in prompt text where the exact number is not load-bearing.
- `tests/unit/test_message_drafter.py:284-287` requires `SELF_DRAFT_INSTRUCTION` to contain "outcome", "narration", and "bullet" (lowercased) and stay under 1000 chars. Changing the bullet mandate means updating that assertion deliberately.
- `tools/valor_telegram.py:1310-1315` warns that the `read` subparser's flags are asserted by an integration test (#2694); reference `--chat-id` as it exists, do not change it.

## Acceptance criteria

- [ ] `prime-pm-role.md` contains a room-matching section under `# Persona behaviors to keep` covering prose-not-structure, proportional length, and long-form-to-file.
- [ ] `prime-teammate-role.md` carries equivalent guidance.
- [ ] `tests/unit/test_pm_progress_updates.py` passes unmodified, or any modification is justified in the PR description.
- [ ] New prompt text contains no forward-deferral phrasing that the promise gate blocks.
- [ ] Prime files retain their hardlinks (edited in place, inode preserved).
- [ ] `config/personas/segments/identity.md` and `docs/features/pm-voice-refinement.md` no longer claim an LLM drafter condenses output, and `pm-voice-refinement.md`'s symbol references resolve to real code.
- [ ] `SELF_DRAFT_INSTRUCTION` is consistent with the new guidance.
- [ ] Narrow-scope tests pass (touched test modules only).

## Downstream context

Related: #1680 (repositioned the drafter from summarizer to validator, which is what removed the condensing step these docs still describe).

## Recon Summary

Verified against `main` @ `e73e25581` (2026-09-09) by a read-only Explore sweep. Every file:line cited above resolves exactly.

**Confirmed**

- `.claude/commands/roles/prime-pm-role.md` (87 lines) contains exactly these headings: `:9 # WORKER Rails`, `:13 # What you are NOT`, `:19 # What you DO`, `:40 # Progress updates when the work overruns the ask`, `:64 # Jobs: goals and expectations (#2494 / #2708)`, `:76 # Persona behaviors to keep`, `:85 # What the user said`. The target section exists at `:76`. There are no `##` headings in the file.
- The only adjacent existing guidance is `:78` ("Concise. The developer is the executor; you are the router.", which is about *developer instructions*), `:79` (trivial-message one-line ack), and `:62` (client vs Eng rooms). There is **no** formatting guidance and no length-proportionality rule anywhere in the file.
- `.claude/commands/roles/prime-teammate-role.md` (68 lines): headings `:9`, `:13`, `:23`, `:39 # Teammate persona`, `:48`, `:60`, `:66`. Its analogue to "Persona behaviors to keep" is `# Teammate persona` at `:39`. Existing guidance at `:41` (casual/warm), `:42` (trivial ack), `:43` (quick and helpful). No formatting or length rule.
- `agent/session_runner/role_driver.py:379-394` — `_prime_args` prepends the role slash command on turn 1 only (guarded by `self._primed`); `PRIME_PATH_SLASH` is the default (`:55-57`). `_PRIME_SLASH_BY_ROLE` / `_PRIME_FILE_BY_ROLE` at `:64-73`. `agent/session_runner/runner.py:557` — `role = "teammate" if (self._session_type or "").lower() == "teammate" else "pm"`, so every `eng` session gets `prime-pm-role.md`.
- `compose_system_prompt` is defined at `agent/sdk_client.py:890` and called only from `load_system_prompt` (`:1082`) and `load_eng_system_prompt` (`:1127`), neither of which has a production caller in `agent/`, `bridge/`, or `tools/`. Outside `sdk_client.py` the only non-test reference is a docstring at `agent/session_runner/harness/claude.py:330`. `config/personas/segments/identity.md` therefore reaches no runtime session prompt.
- `config/personas/segments/identity.md:46-49` carries the correct "direct / concise (short; longer only when requested)" directive. `:57-59` claims "Long outputs are condensed by the message drafter (`bridge/message_drafter.py`, Haiku)". `:39-40` is an HTML comment referencing `DRAFTER_SYSTEM_PROMPT`, a constant that exists nowhere in `bridge/` or `agent/`.
- `bridge/message_drafter.py:1-17` (module docstring) states "No LLM rewriting of the agent's output. The drafter's job is validation + structural composition, not summarization." `draft_message` docstring `:1180-1185` confirms "Haiku, OpenRouter, and all LLM-rewrite paths have been removed." Over-length output is attached as a file via `_write_full_output_file` (`:1013`), never summarized.
- `docs/features/pm-voice-refinement.md` `:9`, `:17`, `:21`, `:29`, `:52` all describe `DRAFTER_SYSTEM_PROMPT` instructing an LLM. `:17` and `:52` also name `_parse_summary_and_questions()`, which does not exist (real name: `_parse_draft_and_questions`, `bridge/message_drafter.py:1073`), and `:25` attributes `_truncate_at_sentence_boundary()` to `bridge/response.py` when it lives at `bridge/message_drafter.py:64`.
- `SELF_DRAFT_INSTRUCTION` at `bridge/message_drafter.py:1025-1033` mandates "2-4 bullet points" — in direct tension with prose-in-chat. `tests/unit/test_message_drafter.py:284-287` requires it to contain "outcome", "narration", "bullet" (lowercased) and stay under 1000 chars.
- `tests/unit/test_pm_progress_updates.py` locks the prime file: `PM_SECTION_HEADER` (`:55`, asserted `:102`), plus substrings asserted at `:109`, `:114`, `:122`, `:127`, `:128`, `:136`, `:145`, `:150`, `:158`, `:163`, `:171`, `:180`, `:183`, `:188`. `PHRASING_WORKAROUND_STRINGS` at `:80-85` is the four-item ban list, asserted absent at `:193-200`. `TAUGHT_BLOCKED` (`:76`) is `"dev opened PR #102. I'll report back when tests finish."` and must yield `action == "block"`, `class_ == "forward_deferral"` — this is the promise-gate constraint on new prompt text.
- `valor-telegram read --chat-id <id>` is real: entry point `pyproject.toml:81`, `read` subparser at `tools/valor_telegram.py:1322`, `--chat-id` in a mutually exclusive group with `--chat`/`--user`/`--project` (`:1327`, runtime check `:484`). `:1310-1315` warns the parser's flags are asserted by an integration test (#2694).

**Revised**

- The problem is an absence of guidance, not a defective code path. Scope is prompt text plus documentation correction; **no runtime logic change is required**, with the single exception of `SELF_DRAFT_INSTRUCTION` reconciliation.

**Dropped / out of scope**

- Wiring `compose_system_prompt` / `identity.md` into the runner. That is a separate architectural change with its own blast radius; this issue fixes the operative file (`prime-pm-role.md`) and corrects identity.md's false claims in place.
- Any edit to `CLAUDE.md` — its headings are regex-parsed into worker prompts and asserted byte-for-byte.

valorengels:
Closes #3270
Closes #3271

Plan: `docs/plans/telegram-reply-duplication-and-room-register.md`

Two independent defects made Valor a bad chat participant, and a human named both directly. They touch disjoint file sets and ship together as one behavior fix.

## A. Extra messages nobody asked for (#3270)

Finished Telegram threads were re-run on a ~30-minute cadence, and every re-run posted a new reply. Two auto-recovery loops fed each other. Established from the append-only `session_events` ledger of `tg_valor_-1003449100931_1450`, which is independent of the corrupted `status` field:

```
13:23:32  running→completed: transcript completed   <-- the ONLY terminal transition ever recorded
14:21:33  completed→pending: resume (sdlc-stall)
15:02:47  exit_summary turns=3                      <-- run ended, NO terminal transition
15:07:33  running→pending: health check (kind=no_progress)
15:12:33  running→failed: 2 recovery attempts
15:21:41  failed→pending: resume (sdlc-stall)       <-- failed rows get re-resumed forever
```

Four defects combine, and each is fixed:

- **A4 — the stall ladder resumed anything.** `_pick_steer_target` treated a lane-slug match as a *tiebreaker*, so with no slug match the most-recently-updated eng session won — routinely a slugless human conversation thread. Both rungs now require a non-`None` slug equal to the lane slug, and rung 2 excludes `failed`; otherwise it falls through to create.
- **A1 — the `#918` delivery guard was starved of input.** `_delivery_belongs_to_current_run` would have finalized these rows `completed` rather than requeuing them, but returned `False` because `response_delivered_at` was always `None`. Its two existing writers never fired once across the whole retained log window; real delivery flows through `flush_deferred_self_draft_sync`, which RPUSHed straight to the outbox without stamping. Now stamped on the paths that actually deliver, with a narrow `save(update_fields=[...])` **and the caller-object mirror**, so the stamp survives the trailing save by a stale caller.
- **A3a — a silent full-row save corrupted `status`.** popoto's `save(update_fields=None)` HSETs the entire encoded instance and re-runs `on_save()` for every field. `status` is an `IndexedField`, so a bare `save()` from a stale instance rewrites both the hash and the index, emitting no LIFECYCLE log and no `session_events` entry. The two evidenced sites in `agent/output_handler.py` are narrowed to `update_fields`.
- **A2 — the turn-end finalize was silently skipped.** With `completed` already on the row, `finalize_session` hit its idempotency skip and returned with no `completed_at`. It now WARNs when an idempotent skip coincides with a bound live runner PID (no behavior change) so this is visible next time.

Also: **A5** scopes the resume steer to the session rather than the room, and **A6b** dedupes the timeout needs-attention notice to once per run (key `timeout-notice-sent:{session_id}:{run_id}`).

### finalize_session caller enumeration (A1)

`finalize_session` ends in a bare `session.save()` (`models/session_lifecycle.py:665`), a full popoto HSET of the caller's in-memory snapshot. A caller that re-reads immediately before the call is harmless; one that passes a batch-read loop variable, a long-held parameter, or an object that survived awaits can write a stale snapshot back over a field another writer just persisted. Every production (non-test) call site, classified:

| file:line | arg | class | why |
|---|---|---|---|
| `bridge/session_transcript.py:343` | `s` | fresh | `get_authoritative_session()` at `:316` |
| `agent/session_executor.py:2470` | `_auth` | fresh | `get_authoritative_session()` 3 lines earlier |
| `agent/session_executor.py:2782` | `_auth` | fresh | `get_authoritative_session()` at `:2763` |
| `agent/session_completion.py:162` | `session` | fresh | rebound from `rows_for_session_id()` at `:130`, no awaits |
| `agent/agent_session_queue.py:781` | `session` | fresh | `get_by_id()` at `:760` |
| `agent/agent_session_queue.py:2743` | `fresh` | fresh | `query.get()` at `:2735` |
| `tools/valor_session.py:1705` | `session` | fresh | `_find_session()` at `:1691` |
| `.claude/hooks/stop.py:201` | `agent_session` | fresh | `get_by_id()` at `:168` |
| **`agent/session_completion.py:1144`** | `parent` | **stale** | last read at `:922`; **the same object passed to `send_cb` at `:1060`** |
| **`agent/session_health.py:3727 / :3753 / :3790 / :3863`** | `entry` | **stale** | `_apply_recovery_transition` param, each preceded by an `await` notice delivery that routes the SAME object into `send` via `deliver_system_notice` |
| `agent/session_health.py:3476` | `entry` | stale | same param, but runs before any delivery |
| `agent/session_health.py:1267 / :1425 / :4721 / :4900` | `entry` | stale | batch loop variables |
| `agent/session_executor.py:1103 / :1145 / :2041` | `session` | stale | the executor's long-held queue-popped object — a *different* object from the `agent_session` handed to `send`, and all three fire before the runner is constructed |
| `agent/agent_session_queue.py:1698` | `stuck` | stale | loop variable over `pending` built at `:1683` |
| `models/session_lifecycle.py:1774` | `parent` | stale | `_transition_parent` parameter |
| `tools/valor_session.py:1685` | `s` | stale | loop variable over an `enumerate_sessions` batch |
| `tools/sdlc_session_ensure.py:1194` | `s` | stale | loop variable over the orphan scan |
| `tools/agent_session_scheduler.py:1158 / :1185` | `target` / `child` | stale | `_kill_process` sleeps up to ~3.5s first |
| `scripts/update/run.py:541 / :607` | `s` | stale | loop variables over a batch query |
| `monitoring/session_watchdog.py:298 / :1287` | `session` | stale | watchdog loop variable, after health-assessment work |

**Reachable from the `output_handler.py::send` stamp** — i.e. stale-passing *and* holding the very object `send` received as `session=`: `agent/session_completion.py:1144` and the four `agent/session_health.py` notice-delivery finalizes. Both are covered by A1's caller-object mirror, which now mirrors **both** `response_delivered_at` and the cleared `extra_context` (mirroring the stamp alone let the trailing full save re-arm `deferred_self_draft_pending` with the originally rejected draft text, which `_deferred_self_draft_backstop_sweep` then re-delivers). `test_send_stamp_survives_finalize_session_by_the_same_caller_object` in `tests/unit/test_response_delivered_stamp.py` is the A1c sibling for that shape, and is RED without the mirror. The remaining stale-passing sites never see an object `send` mutated; they are sweeper/CLI/watchdog rows independently re-read from Redis, and are #3272's territory.

### The stamp arms a previously dormant guard

Stamping mid-run made the delivery guard in `_agent_session_health_check` (`agent/session_health.py`) live on **running** rows: it deliberately skips the `worker_alive`/`_has_progress` evaluation, so a long eng/PM run that redrafts a deferred reply and keeps working past the 300s cadence would have been finalized `completed` out from under its own runner — the mid-turn corruption #3270 exists to stop. That guard is now gated on the absence of a live execution fence, the same discriminator A2 uses. The sibling guard inside `_apply_recovery_transition` is deliberately left ungated: every caller has already decided to take the row away from its runner, so finalizing a delivered row `completed` is the gentler of the outcomes on the table.

## B. Replies that didn't fit the room (#3271)

In chat `-1003449100931`, human messages run a median of 8 words on 1 line with no headers or lists. Valor's agent-composed replies ran a median of ~247 words; 17 of 70 used `#` headers, 18 used numbered lists, the longest was 1029 words in reply to a 17-word question.

- **B1** adds a "Match the room" section to `prime-pm-role.md` and `prime-teammate-role.md` — the operative style text, since the runner prepends the prime on turn 1 and never calls `compose_system_prompt`. Edited in place (fleet hardlink inode preserved and verified), additive only, no existing paragraph or header touched.
- **B2** corrects the stale claim that a Haiku drafter condenses output; `bridge/message_drafter.py` is validation-only. Fixed in `config/personas/segments/identity.md` and `docs/features/pm-voice-refinement.md`, along with two wrong symbol references.

`tests/unit/test_pm_progress_updates.py` passes **unmodified**, which is the acceptance criterion for B1 staying in bounds.

## Follow-up

- #3272 — AgentSession `save()` hardening sweep (A3b). The remaining bare-`save()` sites are enumerated there; no code fix for them in this PR. The deferred rate-limit-relay routing (A6a) is folded into that issue.

## Tests

Red-first throughout: A0 committed failing guards for the stall-ladder lane filter and the resume steer scope before any fix, and A1/A1c prove `response_delivered_at` stays `None` today.

```
215 passed  (output_handler stale-snapshot, reflections progress check, runner preempt,
             finalize idempotent-skip observability, response_delivered stamp,
             valor_session resume release + steering scope, pm progress updates)
 48 passed  (session_health deferred_backstop, orphan_reap, fence_guards)
```

`python -m ruff check` and `python -m ruff format --check` both clean.

Docs (`docs/features/agent-session-lifecycle-writes.md`, `session-steering.md`, `README.md` index) are on `main` per repo convention: c5d82ead6.

### DIFF
--- a/.claude/commands/roles/prime-pm-role.md
+++ b/.claude/commands/roles/prime-pm-role.md
@@ -82,6 +82,14 @@ These `tools.job_tool` invocations are the one sanctioned exception to the no-sh
 - When the user is clearly asking for status rather than action, prefer `route: "user"` over engaging the developer.
 - If a tool or capability you need is missing from your environment, state it plainly on its own line starting with exactly `[missing-capability]` (e.g. `[missing-capability] gh CLI unavailable — cannot query the PR`); the runner escalates that line for you, so never work around the gap silently.
 
+## Match the room
+
+- **A chat reply is prose, a few sentences long.** No headers, no bold, no numbered lists — unless the human asked for a list.
+- **Length is proportional to the ask.** A one-line question gets a one-to-three-line answer. The human is reading on a phone, not auditing a build log.
+- **Long-form analysis goes to a file**, attached via `file_paths`, with a caption of about two sentences in the `message`. The chat carries the caption; the file carries the detail.
+- **Read the room before a non-trivial group reply.** `valor-telegram read --chat-id <id>` shows the recent history — write at the length and register the humans in that room are using.
+- **Keep the load-bearing specifics** — commit hashes, PR and issue numbers, verdicts. Drop the process narration: what you tried, which subagent ran, how many turns it took.
+
 # What the user said
 
 $ARGUMENTS
--- a/.claude/commands/roles/prime-teammate-role.md
+++ b/.claude/commands/roles/prime-teammate-role.md
@@ -45,6 +45,14 @@ You are a capable operational colleague, not a read-only observer. The **one** h
 - **Issue creation is your job.** If the user has a bug or feature request, run `/do-issue` and file it yourself — don't hand the command to the user.
 - **Defer complex SDLC work.** Code changes, multi-step implementations, and planning route to Dev. Do not attempt those yourself.
 
+## Match the room
+
+- **A chat reply is prose, a few sentences long.** No headers, no bold, no numbered lists — unless the human asked for a list.
+- **Length is proportional to the ask.** A one-line question gets a one-to-three-line answer. The human is reading on a phone, not auditing a build log.
+- **Long-form analysis goes to a file**, attached via `file_paths`, with a caption of about two sentences in the `message`. The chat carries the caption; the file carries the detail.
+- **Read the room before a non-trivial group reply.** `valor-telegram read --chat-id <id>` shows the recent history — write at the length and register the humans in that room are using.
+- **Keep the load-bearing specifics** — commit hashes, PR and issue numbers, verdicts. Drop the process narration: what you tried, what you checked first, how many turns it took.
+
 # What I help with
 
 - Answering questions about the codebase or architecture
--- a/agent/output_handler.py
+++ b/agent/output_handler.py
@@ -888,7 +888,6 @@ class TelegramRelayOutputHandler:
                         if _ctx.pop("deferred_self_draft_pending", None) is not None:
                             _ctx.pop("deferred_self_draft_text", None)
                             _target.extra_context = _ctx
-                            _target.save(update_fields=["extra_context"])
                             # Stamp response_delivered_at (#3270). Reaching
                             # here means the agent redrafted a deferred reply
                             # and this send is delivering it — the second of
@@ -908,9 +907,20 @@ class TelegramRelayOutputHandler:
                             # mirrors: any caller that keeps holding this
                             # object and later hands it to a lifecycle write
                             # would otherwise persist the pre-stamp snapshot.
+                            #
+                            # Both mutations land in ONE narrow save: the
+                            # cleared `extra_context` and the stamp describe
+                            # the same delivery, and a single write keeps them
+                            # from ever persisting apart.
                             _stamp_at = datetime.now(UTC)
                             _target.response_delivered_at = _stamp_at
-                            _target.save(update_fields=["response_delivered_at", "updated_at"])
+                            _target.save(
+                                update_fields=[
+                                    "extra_context",
+                                    "response_delivered_at",
+                                    "updated_at",
+                                ]
+                            )
                             if session is not _target:
                                 session.response_delivered_at = _stamp_at
                     except Exception as _clear_err:
--- a/agent/output_handler.py
+++ b/agent/output_handler.py
@@ -922,7 +922,29 @@ class TelegramRelayOutputHandler:
                                 ]
                             )
                             if session is not _target:
+                                # BOTH fields, not just the stamp. Mirroring
+                                # the stamp alone leaves the caller's own
+                                # `extra_context` still carrying
+                                # `deferred_self_draft_pending`, and
+                                # `finalize_session`'s trailing full save
+                                # re-arms it with the ORIGINALLY REJECTED
+                                # draft text -- which
+                                # `_deferred_self_draft_backstop_sweep` then
+                                # selects on that flag alone and delivers on
+                                # top of the successful resend. The precedent
+                                # site in
+                                # agent/session_health.py::flush_deferred_self_draft_sync
+                                # mirrors both for exactly this reason.
                                 session.response_delivered_at = _stamp_at
+                                _caller_ctx = dict(session.extra_context or {})
+                                _had_pending = (
+                                    _caller_ctx.pop("deferred_self_draft_pending", None) is not None
+                                )
+                                _had_text = (
+                                    _caller_ctx.pop("deferred_self_draft_text", None) is not None
+                                )
+                                if _had_pending or _had_text:
+                                    session.extra_context = _caller_ctx
                     except Exception as _clear_err:
                         # Best-effort; never blocks delivery. Worst case is the
                         # pre-existing stale-flag behavior this fix targets.
--- a/agent/output_handler.py
+++ b/agent/output_handler.py
@@ -1539,6 +1539,15 @@ class TelegramRelayOutputHandler:
         signature. The extra fields are merged into the event dict after the
         base fields are populated, so they can never shadow ``type``, ``ts``,
         ``chat_id``, ``reason``, or ``draft_preview``.
+
+        The save is narrowed to ``["session_events", "updated_at"]`` (#3270).
+        A bare ``save()`` on an AgentSession is a lifecycle write: ``status``
+        is an ``IndexedField`` and popoto's ``save(update_fields=None)`` path
+        encodes the whole instance into one HSET and runs ``on_save()`` for
+        every field, so appending a best-effort event to a possibly-stale
+        ``session`` object was silently authorized to rewrite the row's
+        lifecycle state and its status index, with no LIFECYCLE log and — the
+        sharpest irony of the write — no ``session_events`` entry recording it.
         """
         if session is None:
             return
@@ -1557,7 +1566,7 @@ class TelegramRelayOutputHandler:
             events.append(event)
             session.session_events = events
             if hasattr(session, "save"):
-                session.save()
+                session.save(update_fields=["session_events", "updated_at"])
         except Exception as e:  # pragma: no cover - defensive
             logger.debug("RTR event append failed (non-fatal): %s", e)
 
--- a/agent/output_handler.py
+++ b/agent/output_handler.py
@@ -15,6 +15,7 @@ import logging
 import os
 import time
 import uuid
+from datetime import UTC, datetime
 from pathlib import Path
 from typing import Any, Protocol, runtime_checkable
 
@@ -888,6 +889,30 @@ class TelegramRelayOutputHandler:
                             _ctx.pop("deferred_self_draft_text", None)
                             _target.extra_context = _ctx
                             _target.save(update_fields=["extra_context"])
+                            # Stamp response_delivered_at (#3270). Reaching
+                            # here means the agent redrafted a deferred reply
+                            # and this send is delivering it — the second of
+                            # the two real delivery paths that never recorded
+                            # the fact, leaving #918's
+                            # `_delivery_belongs_to_current_run` reading None
+                            # and the #944 orphan net free to requeue a row
+                            # that had already answered the human.
+                            #
+                            # Narrow save: a bare save() here would be a full
+                            # popoto HSET of a possibly-stale instance, i.e. a
+                            # silent lifecycle write.
+                            #
+                            # The caller's `session` object is mirrored for the
+                            # same reason the flush in
+                            # agent/session_health.py::flush_deferred_self_draft_sync
+                            # mirrors: any caller that keeps holding this
+                            # object and later hands it to a lifecycle write
+                            # would otherwise persist the pre-stamp snapshot.
+                            _stamp_at = datetime.now(UTC)
+                            _target.response_delivered_at = _stamp_at
+                            _target.save(update_fields=["response_delivered_at", "updated_at"])
+                            if session is not _target:
+                                session.response_delivered_at = _stamp_at
                     except Exception as _clear_err:
                         # Best-effort; never blocks delivery. Worst case is the
                         # pre-existing stale-flag behavior this fix targets.
--- a/agent/output_handler.py
+++ b/agent/output_handler.py
@@ -1485,13 +1485,22 @@ class TelegramRelayOutputHandler:
         (its own transient concept — verbatim questions for the human) is
         never persisted to the session row; Job expectations (#2708) are the
         durable obligation record and live on the Job, not here.
+
+        The save is narrowed to ``["context_summary", "updated_at"]`` (#3270).
+        A bare ``save()`` on an AgentSession is a lifecycle write: ``status``
+        is an ``IndexedField`` and popoto's ``save(update_fields=None)`` path
+        encodes the whole instance into one HSET and runs ``on_save()`` for
+        every field, so this routing-field write — reached from a possibly
+        long-held ``session`` object — was silently authorized to rewrite the
+        row's lifecycle state and its status index from a stale snapshot, with
+        no LIFECYCLE log and no ``session_events`` entry.
         """
         try:
             context_summary = getattr(draft, "context_summary", None)
 
             if context_summary:
                 session.context_summary = context_summary
-                session.save()
+                session.save(update_fields=["context_summary", "updated_at"])
                 logger.debug(
                     "Persisted routing fields to session %s (context_summary=%s)",
                     getattr(session, "session_id", "<unknown>"),
--- a/agent/session_health.py
+++ b/agent/session_health.py
@@ -2858,6 +2858,41 @@ def flush_deferred_self_draft_sync(session: "AgentSession", status: str | None =
         )
         delivered = True
 
+        # Stamp response_delivered_at (#3270). This flush is the path PM/eng
+        # replies actually take, and until now it wrote the outbox payload
+        # without recording that the human had been answered. The only two
+        # other writers (agent/session_executor.py under `action == "deliver"`,
+        # agent/session_completion.py gated on `delivery_attempted`) are
+        # unreachable from here, so #918's duplicate-delivery guard
+        # `_delivery_belongs_to_current_run` read None and returned False for
+        # every deferred-self-draft delivery — and the #944 orphan net requeued
+        # rows that had already replied, once per tick.
+        #
+        # Narrow save: a bare save() here would be a full popoto HSET of a
+        # possibly-stale instance, i.e. a silent lifecycle write.
+        #
+        # ALSO mutate the caller's in-memory `session` object, for the same
+        # reason the extra_context clear below does: `finalize_session` runs a
+        # full `session.save()` on its own `session` parameter immediately
+        # after this flush returns, which writes back whatever that object
+        # still holds. Without the mirror the stamp is erased microseconds
+        # after it lands, on the very save that finalizes the transition.
+        try:
+            _stamp_at = datetime.now(tz=UTC)
+            _stamp_target = get_authoritative_session(session_id) or source
+            _stamp_target.response_delivered_at = _stamp_at
+            _stamp_target.save(update_fields=["response_delivered_at", "updated_at"])
+            if session is not None and session is not _stamp_target:
+                session.response_delivered_at = _stamp_at
+        except Exception as _stamp_err:
+            logger.warning(
+                "[session-health] failed to stamp response_delivered_at for %s after "
+                "successful flush (non-fatal; the #918 delivery guard stays blind for "
+                "this row): %s",
+                session_id,
+                _stamp_err,
+            )
+
         # Post-delivery flag clear (#3053 correction — the flush is NOT
         # self-clearing via #2489; that clear covers only the redraft-success
         # path in TelegramRelayOutputHandler.send(), never this flush). Without
--- a/agent/session_health.py
+++ b/agent/session_health.py
@@ -3449,6 +3449,17 @@ async def _apply_recovery_transition(
     # that the delivery belongs to the current run's epoch
     # (response_delivered_at >= started_at, falling back to created_at; a
     # legacy row with no anchor still passes through, unguarded).
+    #
+    # No live-fence gate here, unlike the sibling guard in
+    # `_agent_session_health_check` (#3270). The difference is what this branch
+    # is an alternative TO: every caller of `_apply_recovery_transition` has
+    # already evaluated liveness and decided this row is being taken away from
+    # its runner — the only open question is whether it lands `pending`,
+    # `failed`, or `abandoned`. Finalizing a delivered row `completed` is
+    # strictly the gentler of those outcomes, so adding a fence gate here would
+    # not spare a live row, it would only route it into a harsher transition.
+    # The sibling guard needs the gate precisely because it runs BEFORE any
+    # liveness evaluation.
     if _delivery_belongs_to_current_run(entry):
         try:
             from models.session_lifecycle import (
@@ -4681,7 +4692,23 @@ async def _agent_session_health_check() -> None:
         # falling back to created_at) so a stale prior-run delivery doesn't
         # suppress recovery of a genuinely stuck current run; legacy rows
         # with no anchor still pass through unguarded.
-        if _delivery_belongs_to_current_run(entry):
+        #
+        # ...and the row must not be EXECUTING (#3270). Because this branch
+        # deliberately skips the worker_alive/_has_progress evaluation, the
+        # only thing standing between it and a live mid-turn row is this
+        # check. `response_delivered_at` is now stamped mid-run on the
+        # deferred-self-draft redraft path (`agent/output_handler.py::send`),
+        # so a long eng/PM run that answers the human and then keeps working
+        # past this 300s cadence presents exactly the shape the branch below
+        # finalizes: status `running`, delivery inside the current run's
+        # epoch. Finalizing it `completed` writes a terminal status onto a row
+        # the runner is still driving — the mid-turn status corruption #3270
+        # exists to stop. A live execution fence is the same discriminator
+        # `models/session_lifecycle.py`'s idempotent-skip WARNING uses: the
+        # per-turn subprocess is still ours and still alive. A dead or absent
+        # fence still falls through, so the #918 duplicate-delivery guard this
+        # branch exists for is unchanged for genuinely stranded rows.
+        if _delivery_belongs_to_current_run(entry) and not _session_has_live_fence(entry):
             try:
                 from models.session_lifecycle import StatusConflictError, finalize_session
 
--- a/agent/session_runner/runner.py
+++ b/agent/session_runner/runner.py
@@ -116,10 +116,8 @@ TEAMMATE_TURN_TIMEOUT_S: float = float(
 # for hours, so a short TTL would expire between re-runs and re-spam. It is
 # safe to be long ONLY because the key is run-scoped -- a later, unrelated
 # request in the same thread carries a new record id and a fresh key.
-# PROVISIONAL/TUNABLE --
-# grain of salt: sized to comfortably outlive a stranded row's re-enqueue
-# cadence (the incident re-ran the same row hourly for hours), while still
-# expiring so abandoned sessions need no cleanup step. Override with
+# PROVISIONAL/TUNABLE -- grain of salt: it still expires, so abandoned
+# sessions need no cleanup step. Override with
 # SESSION_RUNNER_TIMEOUT_NOTICE_DEDUP_TTL_S.
 TIMEOUT_NOTICE_DEDUP_TTL_S: int = int(
     os.environ.get("SESSION_RUNNER_TIMEOUT_NOTICE_DEDUP_TTL_S", "86400")
@@ -306,14 +304,15 @@ def _claim_timeout_notice(session_id: str, run_id: str = "") -> bool:
             empty id means the row is unidentifiable, so no dedupe is possible
             and the notice is delivered.
         run_id: The AgentSession record's ``id``, minted fresh on every
-            reply-resume. Empty is safe: it degrades to a thread-wide key for
-            a row that has no id yet, which cannot collide with an identified
-            row's key.
+            reply-resume. Empty means the run is unidentifiable, so the notice
+            is delivered rather than deduped on a thread-wide key -- that key
+            would be the silence bug in miniature, and failing open here is the
+            same stance as the Redis-outage path above.
 
     Returns:
         True if this run owns the send, False if an earlier run already sent it.
     """
-    if not session_id:
+    if not session_id or not run_id:
         return True
     try:
         from popoto.redis_db import POPOTO_REDIS_DB  # noqa: PLC0415
--- a/agent/session_runner/runner.py
+++ b/agent/session_runner/runner.py
@@ -109,9 +109,14 @@ TEAMMATE_TURN_TIMEOUT_S: float = float(
     os.environ.get("SESSION_RUNNER_TEAMMATE_TURN_TIMEOUT_S", "900")
 )
 
-# TTL (seconds) on the per-session `timeout-notice-sent:{session_id}` dedupe
-# key that makes TIMEOUT_NEEDS_ATTENTION_MESSAGE a once-per-SESSION notice
-# rather than a once-per-RUN one (#3270 defect 7). PROVISIONAL/TUNABLE --
+# TTL (seconds) on the `timeout-notice-sent:{session_id}:{run_id}` dedupe key
+# that stops TIMEOUT_NEEDS_ATTENTION_MESSAGE from being re-posted on every
+# re-run of one stranded row (#3270 defect 7). The TTL is deliberately long
+# relative to the re-enqueue cadence: the incident re-ran the same row hourly
+# for hours, so a short TTL would expire between re-runs and re-spam. It is
+# safe to be long ONLY because the key is run-scoped -- a later, unrelated
+# request in the same thread carries a new record id and a fresh key.
+# PROVISIONAL/TUNABLE --
 # grain of salt: sized to comfortably outlive a stranded row's re-enqueue
 # cadence (the incident re-ran the same row hourly for hours), while still
 # expiring so abandoned sessions need no cleanup step. Override with
@@ -263,7 +268,7 @@ RUNNER_ERROR_USER_MESSAGE = (
 )
 
 
-def _claim_timeout_notice(session_id: str) -> bool:
+def _claim_timeout_notice(session_id: str, run_id: str = "") -> bool:
     """Claim the right to deliver the timeout needs-attention notice ONCE.
 
     ``TIMEOUT_NEEDS_ATTENTION_MESSAGE`` describes a *session-level* condition
@@ -272,9 +277,22 @@ def _claim_timeout_notice(session_id: str) -> bool:
     incident posted the byte-identical text into a chat once per re-enqueue.
 
     Follows this repo's established one-shot-notice convention (Redis SETNX on
-    a per-session key with a TTL, e.g. ``failed-sent:{session_id}`` in
-    ``agent.session_executor``): the first caller wins, later callers are
-    suppressed. The key is a plain Redis string, NOT an AgentSession field --
+    a keyed marker with a TTL, e.g. ``self_draft_completed_flush_sent:``
+    ``{session_id}:{run_id}`` in ``agent.session_health``): the first caller
+    wins, later callers are suppressed.
+
+    **Scoped per-RUN, not per-thread.** The key carries the AgentSession
+    record's ``id`` alongside the thread ``session_id``. A resumed session
+    reuses its thread ``session_id``, so a ``session_id``-only key would
+    silently swallow the notice for a *later, unrelated* request in the same
+    chat thread -- and that suppression is unrecoverable, because
+    ``ExitReason.TURN_TIMEOUT`` is ``wrapup_eligible=False``
+    (``agent/session_runner/router.py``), so the wrap-up guard never fires to
+    say anything else. The human would get pure silence for a request they
+    made hours later. Re-runs of the SAME stranded row keep the same record
+    ``id``, which is the #3270 spam this dedupes.
+
+    The key is a plain Redis string, NOT an AgentSession field --
     a bare ``save()`` on an AgentSession is a lifecycle write (popoto's default
     save path is a full HSET and ``status`` is an ``IndexedField``), so a
     cosmetic dedupe marker must never travel on that path.
@@ -284,9 +302,13 @@ def _claim_timeout_notice(session_id: str) -> bool:
     turn that was preempted mid-flight.
 
     Args:
-        session_id: The AgentSession the notice would be delivered for. An
+        session_id: The chat thread the notice would be delivered to. An
             empty id means the row is unidentifiable, so no dedupe is possible
             and the notice is delivered.
+        run_id: The AgentSession record's ``id``, minted fresh on every
+            reply-resume. Empty is safe: it degrades to a thread-wide key for
+            a row that has no id yet, which cannot collide with an identified
+            row's key.
 
     Returns:
         True if this run owns the send, False if an earlier run already sent it.
@@ -298,7 +320,7 @@ def _claim_timeout_notice(session_id: str) -> bool:
 
         claimed = bool(
             POPOTO_REDIS_DB.set(
-                f"timeout-notice-sent:{session_id}",
+                f"timeout-notice-sent:{session_id}:{run_id}",
                 "1",
                 nx=True,
                 ex=TIMEOUT_NOTICE_DEDUP_TTL_S,
@@ -928,7 +950,8 @@ class SessionRunner:
                         # in the transcript; surface needs-attention -- but at
                         # most ONCE per session, not once per run (#3270).
                         if _claim_timeout_notice(
-                            str(getattr(self._agent_session, "session_id", "") or "")
+                            str(getattr(self._agent_session, "session_id", "") or ""),
+                            str(getattr(self._agent_session, "id", "") or ""),
                         ):
                             self._adapter.on_user_payload(TIMEOUT_NEEDS_ATTENTION_MESSAGE)
                         summary.exit_reason = ExitReason.TURN_TIMEOUT
--- a/agent/session_runner/runner.py
+++ b/agent/session_runner/runner.py
@@ -109,6 +109,17 @@ TEAMMATE_TURN_TIMEOUT_S: float = float(
     os.environ.get("SESSION_RUNNER_TEAMMATE_TURN_TIMEOUT_S", "900")
 )
 
+# TTL (seconds) on the per-session `timeout-notice-sent:{session_id}` dedupe
+# key that makes TIMEOUT_NEEDS_ATTENTION_MESSAGE a once-per-SESSION notice
+# rather than a once-per-RUN one (#3270 defect 7). PROVISIONAL/TUNABLE --
+# grain of salt: sized to comfortably outlive a stranded row's re-enqueue
+# cadence (the incident re-ran the same row hourly for hours), while still
+# expiring so abandoned sessions need no cleanup step. Override with
+# SESSION_RUNNER_TIMEOUT_NOTICE_DEDUP_TTL_S.
+TIMEOUT_NOTICE_DEDUP_TTL_S: int = int(
+    os.environ.get("SESSION_RUNNER_TIMEOUT_NOTICE_DEDUP_TTL_S", "86400")
+)
+
 # How often the preempt watcher polls the steering list during a turn.
 # Provisional/tunable — override with SESSION_RUNNER_STEER_POLL_INTERVAL_S.
 STEER_POLL_INTERVAL_S: float = float(os.environ.get("SESSION_RUNNER_STEER_POLL_INTERVAL_S", "2.0"))
@@ -250,6 +261,63 @@ TIMEOUT_NEEDS_ATTENTION_MESSAGE = (
 RUNNER_ERROR_USER_MESSAGE = (
     "I hit a problem finishing this and had to stop. Please try again or follow up."
 )
+
+
+def _claim_timeout_notice(session_id: str) -> bool:
+    """Claim the right to deliver the timeout needs-attention notice ONCE.
+
+    ``TIMEOUT_NEEDS_ATTENTION_MESSAGE`` describes a *session-level* condition
+    ("I've paused the work and kept the progress so far"), so re-delivering it
+    on every re-run of the same row is pure noise to the human -- the #3270
+    incident posted the byte-identical text into a chat once per re-enqueue.
+
+    Follows this repo's established one-shot-notice convention (Redis SETNX on
+    a per-session key with a TTL, e.g. ``failed-sent:{session_id}`` in
+    ``agent.session_executor``): the first caller wins, later callers are
+    suppressed. The key is a plain Redis string, NOT an AgentSession field --
+    a bare ``save()`` on an AgentSession is a lifecycle write (popoto's default
+    save path is a full HSET and ``status`` is an ``IndexedField``), so a
+    cosmetic dedupe marker must never travel on that path.
+
+    **Fail-open**: if Redis is unavailable the notice is delivered anyway. A
+    duplicate needs-attention message is strictly better than silence about a
+    turn that was preempted mid-flight.
+
+    Args:
+        session_id: The AgentSession the notice would be delivered for. An
+            empty id means the row is unidentifiable, so no dedupe is possible
+            and the notice is delivered.
+
+    Returns:
+        True if this run owns the send, False if an earlier run already sent it.
+    """
+    if not session_id:
+        return True
+    try:
+        from popoto.redis_db import POPOTO_REDIS_DB  # noqa: PLC0415
+
+        claimed = bool(
+            POPOTO_REDIS_DB.set(
+                f"timeout-notice-sent:{session_id}",
+                "1",
+                nx=True,
+                ex=TIMEOUT_NOTICE_DEDUP_TTL_S,
+            )
+        )
+    except Exception as dedup_err:
+        logger.debug(
+            "[%s] timeout-notice dedupe unavailable (%s); delivering anyway",
+            session_id,
+            dedup_err,
+        )
+        return True
+    if not claimed:
+        logger.info(
+            "[%s] timeout needs-attention notice suppressed — already delivered "
+            "for this session (#3270)",
+            session_id,
+        )
+    return claimed
 
 
 def turn_timeout_for(session_type: str | None) -> float:
@@ -857,8 +925,12 @@ class SessionRunner:
                     self._record_turn_event(handle, turn_end_source=source)
                     if handle.kill_cause == "timeout":
                         # Graceful preempt, not an error: partial work stays
-                        # in the transcript; surface needs-attention.
-                        self._adapter.on_user_payload(TIMEOUT_NEEDS_ATTENTION_MESSAGE)
+                        # in the transcript; surface needs-attention -- but at
+                        # most ONCE per session, not once per run (#3270).
+                        if _claim_timeout_notice(
+                            str(getattr(self._agent_session, "session_id", "") or "")
+                        ):
+                            self._adapter.on_user_payload(TIMEOUT_NEEDS_ATTENTION_MESSAGE)
                         summary.exit_reason = ExitReason.TURN_TIMEOUT
                         break
                     # Steer preempt: pending steers drain at the next
--- a/agent/session_runner/runner.py
+++ b/agent/session_runner/runner.py
@@ -335,7 +335,7 @@ def _claim_timeout_notice(session_id: str, run_id: str = "") -> bool:
     if not claimed:
         logger.info(
             "[%s] timeout needs-attention notice suppressed — already delivered "
-            "for this session (#3270)",
+            "for this run (#3270)",
             session_id,
         )
     return claimed
@@ -947,7 +947,7 @@ class SessionRunner:
                     if handle.kill_cause == "timeout":
                         # Graceful preempt, not an error: partial work stays
                         # in the transcript; surface needs-attention -- but at
-                        # most ONCE per session, not once per run (#3270).
+                        # most ONCE per run, not once per chat thread (#3270).
                         if _claim_timeout_notice(
                             str(getattr(self._agent_session, "session_id", "") or ""),
                             str(getattr(self._agent_session, "id", "") or ""),
--- a/agent/steering.py
+++ b/agent/steering.py
@@ -264,8 +264,12 @@ def push_steering_message(
     front: bool = False,
     room_id: str | None = None,
     timestamp: float | None = None,
-) -> None:
+) -> str:
     """Push a message to a steering queue — the Room leg, or the legacy leg.
+
+    Returns the serialized payload exactly as it was written, so a caller
+    whose own follow-up work fails can hand it back to
+    :func:`remove_steering_message` and undo the push. Most callers ignore it.
 
     Args:
         session_id: The active session to steer
@@ -334,6 +338,46 @@ def push_steering_message(
         f"[steering] Pushed {'ABORT' if is_abort else 'message'} to {key}: "
         f"{text[:80]!r} (from {sender}){target_suffix}{front_suffix}"
     )
+    return payload
+
+
+def remove_steering_message(
+    session_id: str,
+    payload: str,
+    room_id: str | None = None,
+) -> bool:
+    """Undo a single :func:`push_steering_message`, by exact payload.
+
+    Exists for the push-then-do-something-that-can-fail shape: a caller that
+    must push BEFORE a second write (to close a two-write race) has no other
+    way to clean up when that second write raises. Without it the orphaned
+    steer sits on the legacy key forever — the legacy leg carries no TTL and
+    is never age-bounded at drain time (only the Room leg is, via
+    ``steering_room_max_age_s``) — and is injected verbatim whenever that
+    session next runs, however much later that is.
+
+    Removes the LAST occurrence (``LREM count=-1``), which is the entry a
+    just-completed ``RPUSH`` wrote. Concurrent pushes of *different* text are
+    unaffected; a byte-identical duplicate would lose its newest copy, which
+    is the correct one to drop here.
+
+    The leg is re-derived from the payload's own ``is_abort`` so it always
+    matches the key ``push_steering_message`` chose, including for a keyword
+    auto-detected abort. Never raises.
+    """
+    try:
+        is_abort = bool(json.loads(payload).get("is_abort"))
+    except Exception:
+        is_abort = False
+    key = _room_queue_key(room_id) if (room_id and not is_abort) else _queue_key(session_id)
+    try:
+        removed = _get_redis().lrem(key, -1, payload)
+    except Exception as e:
+        logger.warning("[steering] Failed to remove steering message from %s: %s", key, e)
+        return False
+    if removed:
+        logger.info("[steering] Removed %s orphaned steering message(s) from %s", removed, key)
+    return bool(removed)
 
 
 def pop_all_steering_messages(session_id: str, room_id: str | None = None) -> list[dict]:
--- a/config/personas/segments/identity.md
+++ b/config/personas/segments/identity.md
@@ -36,9 +36,6 @@ world, ship working code); community-driven technology.
 
 ## Communication Style
 
-<!-- If you modify this section, review DRAFTER_SYSTEM_PROMPT in
-     bridge/message_drafter.py to ensure it still matches Valor's voice. -->
-
 I communicate via Telegram. For how to trigger work or interact with me, refer
 collaborators to `docs/features/telegram-pm-guide.md` (`issue 363` starts SDLC, `PR 363`
 resumes it, reply-to continues a session).
@@ -54,9 +51,10 @@ response reaches Telegram my session is OVER — "I'll update that", "going forw
 "next time" are lies unless the change already happened this session. I show evidence of
 what I DID (commit hash, file path) or honestly say I didn't.
 
-Long outputs are condensed by the message drafter (`bridge/message_drafter.py`, Haiku),
-which represents me as a senior developer reporting to a PM: outcomes over process,
-blockers flagged, hashes and URLs preserved.
+Nothing rewrites my words on the way out: the message drafter
+(`bridge/message_drafter.py`) validates and composes, it does not summarize. Length is
+mine to control. A response too long for the medium is attached as a file rather than
+shortened, so writing short is my job, not the pipeline's.
 
 ## When I Reach Out
 
--- a/docs/features/pm-voice-refinement.md
+++ b/docs/features/pm-voice-refinement.md
@@ -1,36 +1,28 @@
 # PM Voice Refinement
 
-The PM persona's Telegram output reads as natural human prose: SDLC stage labels are naturalized, crash messages vary, truncation lands on sentence boundaries, developer metrics stay out of stakeholder-facing text, and the completion emoji is reserved for milestones.
+The PM persona's Telegram output reads as natural human prose: the agent's own text reaches the human verbatim, truncation lands on sentence boundaries, question prefixes are normalized, issue and PR references are linkified, and the completion emoji is reserved for milestones.
 
 ## Behavior
 
-### SDLC Stage Naturalization
+### Verbatim Pass-Through
 
-The `DRAFTER_SYSTEM_PROMPT` in `bridge/message_drafter.py` instructs the LLM to translate raw SDLC stage labels to natural language: PLAN becomes "planning", BUILD becomes "building", TEST becomes "testing", REVIEW becomes "reviewing", DOCS becomes "documenting", MERGE becomes "merging". The term "SDLC" itself remains acceptable as a process reference. This is a prompt-only behavior — the LLM handles the translation at draft time.
-
-### Crash Message Pool
-
-`agent/sdk_client.py` defines `CRASH_MESSAGE_POOL`, a list of five varied crash fallback messages. Each includes next-step language ("retry", "try again", "re-trigger", "re-send"). The `_get_crash_message()` function selects randomly from the pool while tracking `_last_crash_message` at module level to prevent consecutive repeats. If the pool is empty, a hardcoded default is returned.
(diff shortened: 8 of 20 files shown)


valorengels:
## Review: Changes Requested — Quorum Shortfall (no code findings)

**Zero code findings.** Every rubric item is `pass` or `n/a`, the Miscellaneous bucket is empty, and I found nothing to patch. The `CHANGES REQUESTED` verdict is a **process** outcome, not a code outcome.

**Mode: `sequential lenses (Agent tool unavailable: not in tool list)`.**

This repo declares a mandatory two-judge roster (`code-quality`, `risk`) with `expected_judges=2`. The Agent tool is absent from this stage runner's tool list, so no judge subagents could be dispatched. I applied both lenses myself, in sequence, in one pass. That produces cited, verified findings but it **cannot produce corroboration** — two judges independently reaching the same conclusion is the evidence the roster exists to generate, and that did not happen here.

`agent.sdlc_review_consensus.compute_consensus(judges=[1], rule="any-blocker-wins", expected_judges=2)` therefore refuses `APPROVED` and returns:

```json
{"verdict": "CHANGES REQUESTED", "blockers": 1, "tech_debt": 0,
 "consensus": {"k": 1, "n": 1, "expected_n": 2, "quorum_shortfall": true, "tied": false}}
```

This is the #3198 fail-closed behavior working as designed: a degraded single-reviewer run is recorded as the shortfall it is, never read back as agreement between judges. `blockers: 1` is the synthetic shortfall blocker, not a defect in the diff.

**Recommended disposition:** re-run REVIEW from a context that has the Agent tool so the declared roster actually dispatches. Nothing in the diff needs to change first — `/do-patch` has no work to do here.

---

### What I verified in my own environment

Per Hard Rule 10, I reproduced every number I relied on rather than crediting the PR body:

| Claim | My measurement |
|---|---|
| `ruff check` clean | `All checks passed!` (exit 0) |
| `ruff format --check` clean | `1506 files already formatted` (exit 0) |
| 215 + 48 passed | **264 passed** in one combined `scripts/pytest-clean.sh` run over all 11 touched modules (incl. `test_pm_progress_updates.py`), 0 failed, 21.7s |
| `test_pm_progress_updates.py` passes unmodified | Confirmed — the file is not in the diff and passes |
| Follow-up #3272 | Confirmed **OPEN** ("AgentSession save() hardening sweep") |
| Docs on `main` at c5d82ead6 | Confirmed — `agent-session-lifecycle-writes.md`, `session-steering.md`, and the `README.md` index entry all present on `origin/main` |
| Prime hardlinks preserved | Confirmed — inode `197161469` (pm) and `197161477` (teammate) shared across the worktree, the main checkout, and `~/.claude/commands/roles/` |
| Every symbol named in the corrected docs resolves | Confirmed — all 8 `message_drafter.py` symbols and `pm_bypass` in `telegram_bridge.py` resolve; the removed `DRAFTER_SYSTEM_PROMPT`, `_parse_summary_and_questions`, and `CRASH_MESSAGE_POOL` are all genuinely absent from the code, so the deletions are accurate, not lossy |
| Env-var registration | `tests/unit/test_env_completeness.py` — 24 passed. `SESSION_RUNNER_TIMEOUT_NOTICE_DEDUP_TTL_S` follows the exact pattern of its two file-local siblings (`SESSION_RUNNER_TEAMMATE_TURN_TIMEOUT_S`, `SESSION_RUNNER_STEER_POLL_INTERVAL_S`), neither of which is in `.env.example` or `config/settings.py` either |

The plan's `## Verification` section is a numbered prose list, not a machine-readable table (`parse_verification_table` → 0 checks, 0 malformed), so I ran its steps by hand; results are the table above. Steps 7 and 8 (post-deploy log inspection and the re-run room-register census) are inherently post-merge and remain open as deploy-time obligations.

### Correctness checks I ran beyond the gates

Three things in this diff are the kind of claim that is easy to assert and hard to verify, so I chased each to ground truth:

1. **A6b's run-scoping premise holds.** `_claim_timeout_notice`'s docstring rests on "a resumed session reuses its thread `session_id` but is minted a fresh AgentSession record `id`". I traced it: `resume_completed_session` → `dispatch_telegram_session` → `enqueue_agent_session` → `_push_agent_session`, which ends in `AgentSession.async_create(...)` against an `AutoKeyField` `id` (`models/agent_session.py:163`). A reply-resume does mint a new `id`; a health-check requeue or `resume_session` transitions the row in place and keeps it. So the key deduplicates exactly the re-run loop and cannot silence a later request. The premise is correct, and it is the difference between this fix working and reintroducing a silence bug.

2. **The A1 stamp is not clobbered by the clear that follows it.** In `flush_deferred_self_draft_sync` the stamp block saves `["response_delivered_at", "updated_at"]`, and the post-delivery clear block immediately after issues its own `get_authoritative_session()` re-read and saves `update_fields=["extra_context"]` — narrow, so it cannot reset the field that was just written. The caller-object mirror is applied to `finalize_session`'s own `session` parameter (the flush receives it directly at `models/session_lifecycle.py:404`), so it protects **every** `finalize_session` caller regardless of that caller's staleness, which is what makes the PR body's enumeration sound rather than merely asserted.

3. **The round-2 "stale-passing caller" is benign for this field.** The critique named `agent/session_health.py`'s delivery-guard `finalize_session(entry, ...)` as a caller that passes a loop variable without re-reading. That call site is gated on `_delivery_belongs_to_current_run(entry)`, which reads `entry.response_delivered_at` as its precondition — so the object already carries the stamp before the trailing full save writes it back. No clobber path. I looked for a live failure scenario here and could not construct one, so per the mandatory finding-verification rule this is not reported as a finding.

4. **Neither narrowed `save()` dropped a load-bearing incidental field.** `_persist_routing_fields` mutates only `context_summary`; `_rtr_emit_event` mutates only `session_events`. Both `update_fields` lists cover their own mutation plus `updated_at` (required — `AgentSession.save` skips the timestamp stamp and WARNs when `updated_at` is omitted). The plan's named trap (`session_health.py`'s `entry.started_at = None`, persisted incidentally by a following full save) is in a site this PR does **not** narrow, so it is untouched.

### No-Gos — all respected

Checked mechanically against the diff's file list and hunks: `_delivery_belongs_to_current_run` is not loosened; the #944 orphan net is not suppressed; `finalize_session`'s idempotency *behavior* is unchanged (the change is a log line and a fence read, both observability-only, explicitly leaving #3253's territory alone); `compose_system_prompt`/`identity.md` are not wired into the runtime prompt path; `CLAUDE.md` is not in the diff; `tools/valor_telegram.py` is not in the diff; both prime files are strictly additive (one new `## Match the room` section each, no existing paragraph or header touched).

### Critique concerns — dispositions confirmed

All three rounds' accepted resolutions are present in the code, not just in the plan text:

- **R1 BLOCKER (caller-object mirror)** — shipped at both stamp sites, with the `finalize_session` end-to-end stale-caller test (`test_stamp_survives_finalize_session_with_stale_caller_object`).
- **R1/R2 CONCERN (A3 → A3a split, A6 → A6b split)** — only the two evidenced `output_handler.py` sites are narrowed; A3b and A6a are deferred to OPEN #3272; `harness/claude.py` is untouched.
- **R2 CONCERN (A1 claim broader than the tests)** — resolved via the plan's own sanctioned alternative: the PR body scopes the claim to the flush path explicitly under "finalize_session caller enumeration (A1)".
- **R3 CONCERN (rung 1 has the identical defect)** — both rungs now share one `_same_lane` filter, with red-first tests for each rung plus the fall-through case.
- **R2 NIT (parametrize the stale-snapshot test over both sites)** — `test_output_handler_stale_snapshot.py` is parametrized over `:1468` and `:1526` with their differing `update_fields` sets.

### Notable design calls, all sound

- `_same_lane` fails closed on a falsy `lane_slug` (admits nothing → `create`). That is the correct direction and is documented in the closure's own docstring: creating a fresh session is recoverable, resuming a stranger's conversation is not.
- `status != "failed"` in rung 2 is load-bearing — `failed` **is** in `RESUMABLE_STATUSES` (`{completed, killed, failed, abandoned}`), so without the exclusion a failed row is re-resumed on a timer forever.
- `_claim_timeout_notice` uses a plain Redis `SETNX` key rather than an AgentSession field, matching the existing `self_draft_completed_flush_sent:{session_id}:{run_id}` convention at `agent/session_health.py:2705`. This is not a Popoto-managed key, so the raw-Redis prohibition does not apply — and the docstring correctly explains that a dedupe marker must never travel on the full-HSET save path.
- The TTL constant is named, env-overridable, and carries an explicit provisional/grain-of-salt comment, per repo convention.

### Tests

Red-first is real here, not claimed: A0's guards assert the *old* ranking is gone (`test_target_is_the_create_rung_when_no_session_matches_the_lane` replaces a case that previously asserted the recency fallback), and the failure path is covered on every axis — live fence / no fence / dead fence for A2; slugless on both rungs, cross-lane, fall-through, and `failed`-row for A4; two-runs-one-row vs. two-runs-two-rows for A6b. The `_AnyLane` test shim is a deliberate, documented device so the ladder tests stay about the ladder; the lane-matching tests all pass real slugs.

## Rubric

- [x] **1. Plan vs. implementation match** — pass — *every Track A (A0–A7), Track B (B1–B2), and Shared (D1–D2, V1–V3) task is delivered; A3b/A6a are deferred to OPEN #3272*
- [x] **2. New code quality** — pass — *type hints, named env-overridable constant, narrow `update_fields` saves, every new `except Exception` logs or is documented swallow-ok, no copy-paste*
- [x] **3. Test coverage** — pass — *264 targeted tests green; every new path has a test and every guard has a failure-path case*
- [x] **4. Regression risk to existing callers** — pass — *`_pick_steer_target`'s behavior change is intentional, documented under Risks, and its existing tests were updated rather than deleted; both narrowed saves verified to drop no incidental field*
- [x] **5. Data integrity** — pass — *no schema change, so no `migrations.py` entry needed; every new `update_fields` list includes `updated_at`*
- [x] **6. Security** — pass — *no new `eval`/`exec`/`subprocess`/raw SQL, no secrets, no request-derived input*
- [x] **7. Documentation accuracy** — pass — *docs landed on `main` at c5d82ead6 per repo convention; every symbol in the corrected docs resolves and every removed claim is genuinely absent from the code*
- [x] **8. PR body accuracy** — pass — *reproduced independently; my 264 vs the stated 215+48 is one combined run vs two, not a discrepancy*
- [x] **9. Disclosed deferrals** — pass — *A3b and A6a both tracked by OPEN #3272*
- [x] **10. Follow-up claims verified** — pass — *#3272 confirmed OPEN*

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — PASS — *all satisfied except Verification steps 7–8, which are post-deploy by construction*
- **2. No-Gos from plan — none violated** — PASS — *all seven checked mechanically against the diff*
- **3. New `except Exception` blocks — each has logger/raise/swallow-ok** — PASS — *stamp → `logger.warning`; dedupe → `logger.debug` + fail-open; fence read → silent swallow, observability-only, mirroring the adjacent pre-existing handler*
- **4. New integration tests — exercise serialization boundary (not in-memory only)** — PASS — *`test_response_delivered_stamp.py` drives real `save()`/`get_authoritative_session()` round-trips against Redis*
- **6. No hardcoded secrets or debug artifacts** — PASS — *no TODO/FIXME/print/breakpoint in the added lines*
- **7. New public APIs — docstrings present** — PASS — *`_claim_timeout_notice` is thoroughly documented, including its fail-open stance*
- **8. Breaking changes — migration path documented** — N/A — *no API or schema break; the stall-recovery behavior change is recorded under Risks for post-deploy observation*
- **9. Tests added for new behavior** — PASS
- **10. Tests cover the failure path (not just happy path)** — PASS — *red-first throughout, per the plan's Failure Path Test Strategy*
- **11. UI changes (if any) — screenshot captured** — N/A — *no HTML/CSS/JS/JSX/Vue/template files in the diff; the visual-proof gate is a no-op*
- **12. Docs updated for user-facing changes** — PASS — *on `main` at c5d82ead6, plus `identity.md` and `pm-voice-refinement.md` in this PR*

### Blockers
- None

### Tech Debt
- None

### Nits
- None

### Miscellaneous
- None

- **A3b — the remaining bare-`save()` sites are enumerated but not fixed here** — tracked by **#3272** (OPEN) — matched files: `agent/sdk_client.py`, `agent/health_check.py`, `tools/session_tags.py`, `agent/pipeline_state.py`
- **A6a — rate-limit-relay routing** — folded into the same **#3272** (OPEN); the plan's Rabbit Holes records why a stderr keyword detector was refused
- **Verification steps 7–8 (post-deploy log inspection and the re-run room-register census)** — explicit out-of-scope-for-review exclusion; they are deploy-time obligations that cannot be exercised pre-merge

### Verification Results
`ruff check` exit 0 · `ruff format --check` exit 0 · 264 targeted tests passed · 24 env-completeness tests passed · plan `## Verification` table: 0 machine-readable checks (prose list), 0 malformed, run by hand instead

### Screenshots

### Deploy note
This PR modifies `agent/`, `reflections/`, `models/`, and `tools/`, and changes two prompt files. After merge, `/update` propagates the change and the bridge/worker need `./scripts/valor-service.sh restart` for both the runtime and the prompt changes to take effect. The plan flags `_pick_steer_target`'s new fail-closed behavior as the one thing to watch in production.

valorengels:

**Verdict: CHANGES REQUESTED** — head SHA `c960f852b1dbe6d12b06205643060a2be534677b`

### Blockers

**1. `config/personas/segments/identity.md:36-57` — the persona byte-stability baseline was not regenerated, and a test fails on this branch.**

The B2 edit changes the composed `(ENGINEER, WORKER)` system prompt, but `tests/fixtures/persona/eng_worker_repo_baseline.txt` is not in the diff.

```
$ python scripts/capture_persona_baseline.py --check
Baseline is STALE: tests/fixtures/persona/eng_worker_repo_baseline.txt
  recorded: 32730    composed now: 32664

$ scripts/pytest-clean.sh tests/unit/test_compose_system_prompt.py
FAILED tests/unit/test_compose_system_prompt.py::test_eng_cell_byte_stable_against_baseline
  AssertionError: ENG cell drifted from the baseline: composed 32664 chars, baseline 32730 chars.
```

The 66-char drift is exactly this PR's two hunks (the removed `DRAFTER_SYSTEM_PROMPT` HTML comment at fixture line 54, and the rewritten drafter paragraph at fixture line 71). Fix: run `python scripts/capture_persona_baseline.py` and commit the new baseline, as `agent/sdk_client.py:922` and the script's own error message instruct.

### Tech Debt

**2. `agent/output_handler.py:924` — the caller-object mirror covers `response_delivered_at` but not the cleared `extra_context`, leaving the #2489 re-delivery live.**

1. Turn N defers a rejected draft: `output_handler.py:975-980` persists `deferred_self_draft_pending` / `deferred_self_draft_text` on `_target` (the authoritative re-read) only.
2. Turn N+1 the agent redrafts. The clean-send gate at `output_handler.py:880` reads `session.extra_context` — entering this block proves the caller's object carries the flag.
3. Lines 886-925 clear the flag on `_target` (a distinct object) and mirror back only the stamp: `if session is not _target: session.response_delivered_at = _stamp_at`.
4. Turn-end `finalize_session(session, "completed")` reaches `session.save()` at `models/session_lifecycle.py:652` — a full popoto HSET that writes the caller's stale `extra_context` back, re-arming `deferred_self_draft_pending` with the originally rejected text.
5. `_deferred_self_draft_backstop_sweep` (`agent/session_health.py:3254`) selects terminal rows on `extra_context.get("deferred_self_draft_pending")` alone — it does not consult `response_delivered_at`, unlike the running-row sweeps at 3452/4684 — and calls `flush_deferred_self_draft_sync`. Its SETNX key `self_draft_completed_flush_sent:{sid}:{run_id}` was never burned, because step 3 is not the flush path.
6. Result: a second unprompted message carrying the rejected draft — the same defect class #3270 fixes.

The sibling site the new comment cites as precedent (`agent/session_health.py:2919-2929`) mirrors both fields. Fix is one line: mirror the popped `_ctx` onto `session.extra_context` under the same `session is not _target` guard.

### Nits

- `agent/session_runner/runner.py:950` — the call-site comment says "at most ONCE per session, not once per run", the exact inverse of the implementation and of `_claim_timeout_notice`'s own docstring ("Scoped per-RUN, not per-thread"). The suppression log at line 337 reads the same wrong way. Misleading on a subtle key-scoping decision.
- `models/session_lifecycle.py:506-517` — the A2 discriminator reads `live_fence`, whose pid is the per-turn `claude -p` subprocess (`agent/session_runner/runner.py:833`), not the runner. At the ordinary turn-end finalize (`agent/session_executor.py:2470`) that subprocess has already exited, so the WARNING does not fire for the very case A2 cites (the `15:02:47 exit_summary turns=3` line in the PR body's ledger). It still fires for genuine concurrent finalizes, but the comment's claim that health-checker/watchdog finalizes "cannot trip it" is backwards — those are the paths that can.
- `tests/unit/reflections/test_reflections_progress_check.py:151-171` — `_AnyLane.__eq__` returns True for any `str`, making the lane filter a no-op for the ~30 ladder tests that use the default `_Row(slug=...)`. Deliberate and documented, but it will silently absorb a future `_same_lane` regression.
- `reflections/sdlc_progress.py:857` — `bool(lane_slug)` is re-evaluated per row inside `_same_lane`; hoisting it is free.

### Verification Run

- `python -m ruff check .` → All checks passed.
- `python -m ruff format --check .` → 1506 files already formatted.
- Narrow test set (the 7 test files in this diff) → **201 passed**, exit 0.
- `tests/unit/test_compose_system_prompt.py` → `test_eng_cell_byte_stable_against_baseline` **fails** (blocker 1). Four other failures in that file are environmental (`PermissionError` on the private persona layer), unrelated to the diff.
- `python -m tools.check_fence_census` → "Fence census clean — 11 fenced-pid consumer(s), all guarded".

### What held up

`update_fields` completeness across all four narrowed sites (no dropped writes); A1's stamp/clear ordering; A4's `status != "failed"` being live code; A5's `_queue_key(session_id)` fallback; A6b's key shape and TTL; the B2 doc corrections (every deleted symbol genuinely gone). New tests use real Redis and real ORM rows and assert on persisted state, not mocks.

valorengels:
## Review (Judge risk):

**Verdict: CHANGES REQUESTED** — head SHA `c960f852b1dbe6d12b06205643060a2be534677b`

### Blockers

**1. `agent/output_handler.py:917` (the A1 stamp in `send`) arms the health check to finalize *live* running sessions as `completed`.**

The new stamp fires on the redraft-success path, which runs mid-run: the defer happens at turn N and the clean redraft send at turn N+1 of the same run, with status `running` throughout. It writes `response_delivered_at = now`.

`agent/session_health.py:4684` then reads:

```python
for entry in running_sessions:            # 4648, query.filter(status="running")
    ...                                   # only the #1006 terminal guard and the _is_ledger guard
    if _delivery_belongs_to_current_run(entry):   # 4684
        finalize_session(entry, "completed", reason="health check: delivered but not finalized")
```

The comment at 4672-4673 is explicit that this runs "without going through worker_alive/`_has_progress` evaluation" — there is no liveness or fence check anywhere ahead of it, and `_delivery_belongs_to_current_run` (`:411-424`) returns True as soon as `response_delivered_at >= started_at`, which a fresh mid-run stamp always satisfies.

Failure scenario: an eng/PM run is executing; turn N's delivery is flagged by the validator and deferred; turn N+1 the agent redrafts and sends cleanly, so `response_delivered_at` is stamped while the row is `running`; the run continues (SDLC turns routinely exceed the health-check cadence, `AGENT_SESSION_HEALTH_CHECK_INTERVAL = 300` at `agent/session_health.py:442`). The next tick finalizes the live row `completed`, `_reclaim_slot_lease()` frees its slot, the runner's own turn-end finalize is then swallowed by the idempotency skip (now WARNing, per A2), and the human's next reply lands on a terminal row. That is the "terminal status on a mid-turn row" corruption #3270 was filed about, reintroduced from the other direction.

This matters *because* A1 works: `response_delivered_at` was effectively always `None` before this PR (that is the bug A1 fixes), so the unguarded delivery guard at `:4684` was dormant. Populating the field is what activates it.

The flush-site stamp (`agent/session_health.py:2861-2891`) is safe — it runs inside `finalize_session`, so the row goes terminal immediately. Only the `output_handler.send` site is exposed, and it is the one stamp site with no test coverage (`tests/unit/test_response_delivered_stamp.py` has three tests, all against the flush path).

Suggested fix: gate the `:4684` delivery guard on the absence of a live execution fence (`live_fence` + `agent.pid_fence.fence_is_live` — the exact discriminator A2 just introduced), or restrict the stamp to terminal paths.

### Tech Debt

**2. The plan's mandatory A1 caller enumeration was not delivered.** Task A1, its Success Criteria, and the Round-2 critique row require enumerating every `finalize_session` caller, classifying each as fresh-reading vs stale-passing, recording the result in the PR body, and adding a sibling A1c test for each stale-passing caller reachable from the `output_handler.py::send` stamp — or else explicitly narrowing the Success Criteria claim to the flush path. The PR body's "finalize_session caller enumeration (A1)" section names no callers and lists no classification; it asserts "Callers that pass a stale session object … are covered by the caller-object mirror". `agent/session_health.py:4694` and `:3466` both hand `finalize_session` a loop-variable `entry` read earlier in the batch — the exact stale-passing shape the plan named — and no test covers the `output_handler` stamp surviving that path. The claim in the PR body is broader than the tests, which is the outcome the plan explicitly forbade.

### Nits

- `tools/valor_session.py:1156` — moving the resume steer from the Room leg to the legacy leg also drops the age bound. `_drain_list` applies `steering_room_max_age_s` (21600s) only to the Room key (`agent/steering.py:371-374`); the legacy leg is never filtered (`:352`). If `transition_status` raises at `:1161`, the already-pushed steer is not cleaned up, so a stale "your lane stalled N hours ago, continue" instruction sits on `steering:{session_id}` indefinitely. Lower blast radius than the old cross-session leak, but newly unbounded in age.
- `docs/features/session-steering.md` on `main` (c5d82ead6) already describes the resume steer as session-scoped while the implementing code is still on this branch. Docs lead code until merge — expected under the plans-on-main convention.
- The prime-file edits are already live fleet-wide: `.claude/commands/roles/prime-pm-role.md` in this worktree, in the main checkout, and in `~/.claude/commands/roles/` all share inode `197161469`. The hardlink survived as the plan required, but the new prompt text is already in production and the main checkout shows as dirty before merge.

### Deploy Notes

- **Restart required.** `agent/output_handler.py`, `agent/session_health.py`, `agent/session_runner/runner.py`, `models/session_lifecycle.py`, `reflections/sdlc_progress.py`, `tools/valor_session.py` and both primes are on the live path: `./scripts/valor-service.sh restart` after `/update`, and the reflections worker must pick up `sdlc_progress.py`.
- **No migration.** No Popoto schema change — A1 stamps an existing `DatetimeField` (`models/agent_session.py:181`); A3a narrows existing writes. No `scripts/update/migrations.py::MIGRATIONS` entry needed.
- **New env var, optional, not in `.env.example`:** `SESSION_RUNNER_TIMEOUT_NOTICE_DEDUP_TTL_S` (`agent/session_runner/runner.py:122`, default 86400), read via `os.environ.get` with an in-code default, matching the adjacent `SESSION_RUNNER_STEER_POLL_INTERVAL_S` precedent. Consistent; no action required.
- **New Redis key namespace:** `timeout-notice-sent:{session_id}:{run_id}`, plain string, 24h TTL, self-expiring.
- **Watch after deploy:** stall recovery now falls through to `create` whenever no same-lane session exists, including for genuinely `failed` lane rows. Bounded by the per-tick create budget and the per-(slug, sha) attempt budget, but it will move `sdlc-progress-check` counts.

### Verification Run

The plan's `## Verification` section is a prose numbered list, not a table, so the parser found zero machine-checkable rows:

```
## Verification Results

**All checks passed.**
EXIT=0
```

No malformed rows. The 7 diff test files via `scripts/pytest-clean.sh`: **201 passed**, exit 0.

### What held up

A3a's narrowed saves (none of the five listed fields is an `IndexedField` or `KeyField`, so no secondary index can diverge; popoto's partial-save path runs `on_save()` for exactly the listed fields). A2 as genuinely observability-only (both branches fall through to the same `return`). A6b's fail-open as unreachable in production (`run_id` is `AgentSession.id`, an always-populated `AutoKeyField`), so it cannot resurrect #3270. A4's fall-through as safe from its only caller (`_check_project_stalls` does `if not slug: continue` first). All seven No-Gos respected: `CLAUDE.md` untouched, `tools/valor_telegram.py` untouched, `bridge/message_drafter.py` untouched, `test_pm_progress_updates.py` unmodified, prime edits purely additive under existing headings.

valorengels:
## Review: CHANGES REQUESTED

**Head SHA:** `c960f852b1dbe6d12b06205643060a2be534677b`
**Mode:** independent roster (2 judges)
**Consensus:** rule `any-blocker-wins`, k=2, n=2, expected_n=2, `quorum_shortfall: false`, `tied: false`, mean confidence 0.88

This supersedes the earlier `CHANGES REQUESTED` on this PR (comment 5603579004), which was a procedural quorum shortfall with zero findings: that run sat at the harness spawn-depth limit, was withheld the Agent tool, and collapsed the declared two-judge roster into one sequential reviewer (#3198). Re-run inline with the roster actually dispatched. Both judges independently returned CHANGES REQUESTED, each with a blocker the collapsed run did not surface.

Per-judge detail: [code-quality](https://github.com/tomcounsell/ai/pull/3273#issuecomment-5603813083) - [risk](https://github.com/tomcounsell/ai/pull/3273#issuecomment-5603821048)

### Blockers

**1. Persona byte-stability baseline not regenerated — a test fails on this branch.** (code-quality; `config/personas/segments/identity.md:36-57`)

The B2 edit changes the composed `(ENGINEER, WORKER)` prompt, but `tests/fixtures/persona/eng_worker_repo_baseline.txt` is not in the diff. Reproduced independently:

```
$ python scripts/capture_persona_baseline.py --check
Baseline is STALE: tests/fixtures/persona/eng_worker_repo_baseline.txt
  recorded: 32730    composed now: 32664

$ scripts/pytest-clean.sh tests/unit/test_compose_system_prompt.py
FAILED test_eng_cell_byte_stable_against_baseline
```

The 66-char drift is exactly this PR's two hunks. Fix: run `python scripts/capture_persona_baseline.py` and commit the regenerated baseline.

**2. The A1 stamp in `output_handler.send` arms the health check to finalize live running sessions as `completed`.** (risk; `agent/output_handler.py:917`)

The stamp fires mid-run on the redraft-success path, while status is `running`. The delivery guard at `agent/session_health.py:4684` then finalizes the row `completed` with no liveness or fence check ahead of it — its own comment states it runs "without going through worker_alive/`_has_progress` evaluation", and `_delivery_belongs_to_current_run` returns True as soon as `response_delivered_at >= started_at`.

This is a consequence of A1 working: `response_delivered_at` was effectively always `None` before this PR, so that unguarded path was dormant. Populating the field activates it. A long-running eng/PM turn that redrafts a deferred reply and then keeps working past the 300s health-check cadence gets finalized out from under itself — the same mid-turn terminal-status corruption #3270 was filed about, from the other direction. This is also the only stamp site with no test coverage.

Suggested fix: gate the `:4684` delivery guard on the absence of a live execution fence (`live_fence` + `agent.pid_fence.fence_is_live`, the discriminator A2 just introduced), or restrict the stamp to terminal paths.

### Tech Debt

**3. The caller-object mirror omits `extra_context`.** (code-quality; `agent/output_handler.py:924`) Only `response_delivered_at` is mirrored back onto the caller's `session`, so a later full-row `save()` in `finalize_session` rewrites the stale `extra_context` and re-arms `deferred_self_draft_pending` with the originally rejected text. The terminal-row backstop sweep (`session_health.py:3254`) selects on that flag alone and re-delivers. The precedent site the new comment cites (`session_health.py:2919-2929`) mirrors both fields. One-line fix.

**4. The plan's mandatory A1 caller enumeration was not delivered.** (risk) The plan, its Success Criteria, and the Round-2 critique row require enumerating every `finalize_session` caller, classifying each fresh-reading vs stale-passing, recording that in the PR body, and adding a sibling A1c test per stale-passing caller — or narrowing the Success Criteria claim to the flush path. The PR body's enumeration section names no callers. `session_health.py:4694` and `:3466` are both the stale-passing shape the plan named, with no test covering them.

### Nits

- `agent/session_runner/runner.py:950` — call-site comment says "ONCE per session, not once per run", the exact inverse of the implementation and of `_claim_timeout_notice`'s docstring. The suppression log at `:337` reads the same wrong way.
- `models/session_lifecycle.py:506-517` — the A2 discriminator reads a fence pid belonging to the per-turn `claude -p` subprocess, not the runner, so the WARN does not fire for the ordinary turn-end finalize that A2 cites as its motivating case. The comment's claim about which paths can trip it is backwards.
- `tools/valor_session.py:1156` — moving the resume steer to the legacy leg drops the `steering_room_max_age_s` age bound; a steer left behind by a raising `transition_status` persists indefinitely.
- `tests/unit/reflections/test_reflections_progress_check.py:151-171` — `_AnyLane.__eq__` matches any `str`, making the lane filter a no-op for the default-row ladder tests.
- `reflections/sdlc_progress.py:857` — hoist `bool(lane_slug)` out of the per-row `_same_lane`.

### Verification Run

- `ruff check .` → All checks passed. `ruff format --check .` → 1506 files already formatted.
- Narrow test set (the 7 test files in this diff) → **201 passed**, exit 0. Confirmed by both judges independently.
- Plan `## Verification` section is a prose list, not a table → zero machine-checkable rows, exit 0, no malformed rows.
- `tests/unit/test_compose_system_prompt.py` → `test_eng_cell_byte_stable_against_baseline` **FAILS** (blocker 1).
- `python -m tools.check_fence_census` → clean, 11 fenced-pid consumers, all guarded.
- No UI files in the diff — the visual-proof gate is a no-op.

### What both judges confirmed sound

A3a's narrowed `save(update_fields=...)` sites (no `IndexedField`/`KeyField` among them, no dropped writes, `updated_at` present in every list). A2 as genuinely observability-only. A6b's fail-open as unreachable in production (`run_id` is an always-populated `AutoKeyField`), so it cannot resurrect #3270. A4's fall-through as safe from its only caller. A5's `_queue_key(session_id)` fallback. All seven plan No-Gos respected. Prime-file hardlink inodes preserved. New tests use real Redis and real ORM rows, asserting on persisted state rather than mocks.

### Deploy note

Touches `agent/`, `models/`, `reflections/`, `tools/` and both primes — requires `./scripts/valor-service.sh restart` after `/update`. No migration. New optional env var `SESSION_RUNNER_TIMEOUT_NOTICE_DEDUP_TTL_S` (in-code default, matches adjacent precedent, no `.env.example` entry required).

valorengels:
## Review: APPROVED

**Head SHA:** `55eeebd1bcc224b21b1579eae6bfe13eb194d8bb`
**Mode:** independent roster (2 judges)
**Consensus:** rule `any-blocker-wins`, k=2, n=2, expected_n=2, `quorum_shortfall: false`, `tied: false`, mean confidence 0.91

Round 2, after patch commit `55eeebd1b`. Both judges independently returned APPROVED with zero blockers and zero tech debt. All nine round-1 findings verified fixed against the code, not against the patch report.

### Round-1 findings — all FIXED

1. **Blocker — stale persona baseline.** `capture_persona_baseline.py --check` → `Baseline is current (32664 chars)`; `test_eng_cell_byte_stable_against_baseline` passes.
2. **Blocker — the A1 stamp armed the delivery guard against live rows.** `agent/session_health.py:4711` is now `if _delivery_belongs_to_current_run(entry) and not _session_has_live_fence(entry):`, using the pre-existing orphan-reap helper that wraps `agent.pid_fence.fence_is_live`. Two twin tests hold every variable constant except the fence and stamp it with the test process's own pid, so `fence_is_live` returns a real unmocked verdict.
3. **Tech debt — caller-object mirror.** `agent/output_handler.py:938-955` now pops both `deferred_self_draft_pending` and `deferred_self_draft_text` off the caller's object alongside the stamp, matching the precedent site. Covered by a test that reproduces the enumerated re-arm shape against real Redis.
4. **Tech debt — A1 caller enumeration.** The full classification table is in the PR body; the A1c test for the stale-passing shape is in `tests/unit/test_response_delivered_stamp.py`.
5-9. **Nits** — per-run wording corrected in `runner.py` (`:337`, `:950`); the backwards WARN comment in `models/session_lifecycle.py:501-514` rewritten to describe actual behavior; the resume steer now rolled back via the new `agent/steering.py::remove_steering_message` when `transition_status` raises; `_Row` given a real default lane slug so the lane filter is actually exercised; `bool(lane_slug)` hoisted, kept below the query so a Redis failure still returns `("unknown", None)`.

### The one thing worth recording for the future

The fence gate is safe specifically *because* the sibling delivery branch in `_apply_recovery_transition` (`agent/session_health.py:3455`) is deliberately left ungated. The risk judge traced every path a live-fenced, already-delivered `running` row can take:

- Fence live, worker alive, in-scope handle present (a normal long run) — the gate blocks, the `worker_alive` evaluation declines, and the row is bounded by the runner's hard wall-clock `_turn_timeout_s`, after which the fence reads dead and the guard fires on the next tick. The gate delays; it cannot strand.
- Fence live because the worker crashed and orphaned the subprocess — `worker_alive=False`, `running_seconds > 300`, so `_apply_recovery_transition`'s ungated delivery branch finalizes `completed`. #918 preserved.
- The #944 shared-`worker_key` orphan — same entry, same ungated branch. #918 preserved.

So the ungated sibling is not merely "the gentler outcome"; it is the load-bearing fallback that keeps #918 intact under the new gate. Anyone tempted to gate it later should read this first. `fence_is_live` also fails toward recovery on every failure mode (dead pid, recycled pid via the create_time identity compare, missing create_time, unreadable psutil, any exception), so the gate cannot fail into a stuck row.

### Findings

- **Blockers:** None
- **Tech Debt:** None
- **Nits:** None. Both judges raised the same two PR-body inaccuracies — the A6b description still saying "once per session" when the shipped key is per-run, and two line numbers in the enumeration table that drifted by 8 lines under the patch commit. Both were corrected in the PR body before this verdict; no code change, so the head SHA is unchanged.

### Verification Run

- `ruff check .` → All checks passed. `ruff format --check .` → 1506 files already formatted.
- Narrow test set for this diff → **230 passed**, exit 0. Independently reproduced by both judges (207 items on the risk judge's smaller file set, 230 on the full nine-file set).
- `capture_persona_baseline.py --check` → current, zero drift.
- No UI files in the diff — the visual-proof gate is a no-op.
- Prime-file hardlink inode `197161469` confirmed shared across worktree, main checkout, and `~/.claude/`. Both prime diffs are purely additive (zero deleted content lines).
- All plan No-Gos hold: `CLAUDE.md`, `tools/valor_telegram.py`, `bridge/message_drafter.py`, and `test_pm_progress_updates.py` untouched; `_delivery_belongs_to_current_run` unchanged; `finalize_session` idempotency semantics unchanged (logging only).

**Plan-authoring note (not a code finding):** the plan's `## Verification` section is a prose numbered list rather than a table, so `parse_verification_table` collected zero checks and the runner exits 0 vacuously. Worth fixing in the next plan that uses this template — a green result there currently proves nothing.

### Deploy

Touches `agent/`, `models/`, `reflections/`, `tools/`, and both primes — requires `./scripts/valor-service.sh restart` after `/update`, verified by `tail -5 logs/bridge.log` showing "Connected to Telegram". No migration (`live_fence`, `spawn_history`, `response_delivered_at` are all pre-existing fields). No new env vars in `.env.example`.

Issue #3258, opened 2026-09-08
SDLC harness dispatches duplicate builders into occupied worktrees; the duplicate's orphan-cleanup kill -9'd two live benchmark arms


valorengels:
## What happened

During a multi-lane SDLC run in popoto (2026-09-08), the harness dispatched **duplicate builder agents into three worktrees that already had a live builder**: lanes for issues #704, #586, and #699. The duplicates were not spawned by the supervising session — it had one agent per lane and was tracking them by name.

Observed agent ids: `addbcffb3b0090381` and `a1c8ee32e6161dd40` (stopped via TaskStop), `a87b76b2a1c2800d8` (told to stand down).

## The damage

The #586 duplicate found two long-running benchmark arms in its worktree, concluded they were orphaned processes from a crashed run, and `kill -9`'d them. They were not orphans — they were the live arms of the run that lane existed to produce. Cost: two arms of benchmark wall-clock, re-run from scratch. Nothing was corrupted and no data was lost, but the loss was silent until the supervising session went looking for the arms.

That second-order behaviour is the part worth fixing alongside the dispatch itself. A duplicate builder entering an occupied worktree cannot distinguish "another live lane is working here" from "a crashed lane left debris", and the debris-cleanup path is destructive.

## Why it is a harness bug and not operator error

The supervising session held one agent per lane and never re-dispatched. The duplicates arrived anyway, into worktrees that were checked out, dirty, and had a running agent. Nothing in the dispatch path appears to check whether a worktree already has an owner before handing it to a new builder.

## Suggested direction (not prescriptive)

1. **Dispatch-time ownership check.** A worktree should have at most one builder. Refuse (or queue) a dispatch into a worktree that already has a live owner, rather than starting a second writer.
2. **Make the orphan-cleanup path non-destructive by default.** A builder that finds unexpected processes in its worktree should report them and stop, not `kill -9`. If cleanup stays, it needs a positive liveness signal ("this pid belongs to a dead run") rather than an inference from "I didn't start it".

## Related, same root cause class

Lane Redis DBs are hand-assigned by the supervising session rather than derived from one place. In this run that produced a real near-miss: #586's tests and #586's benchmark arms were both initially assigned DB 12, so the lane's own `pytest` run would have flushed its own benchmark. Caught and corrected before it fired (tests moved to DB 8), but a derived-from-one-source assignment would have made it unrepresentable.

[...]

PR #3342, opened 2026-09-16
Scope the commit guard's git queries to the directory the commit actually runs in (#3259)

### LINKED ISSUE #3259: Commit-validation hook resolves git state from the session cwd, not the invoking worktree (false blocks, and false allows on main)
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/commit-validation-hook-worktree-cwd.md

## The defect

`~/.claude/hooks/sdlc/validate_commit_message_sdlc.py` decides whether to block a `git commit` by shelling out to git **with no cwd argument**, so every git call inherits the hook process's working directory — the *session's* cwd — rather than the directory the intercepted `git commit` would actually run in.

Three calls are affected:

- `get_current_branch()` — the branch check at line 65
- `git rev-parse --show-toplevel` (line ~74), used to decide "is this repo popoto?"
- `git diff --cached --name-only` (line ~86), the staged-code check that does the actual blocking

## Why it matters

Multi-lane SDLC runs put each lane in its own `.worktrees/<lane>` checkout. A lane committing inside its worktree is validated against the **main checkout's** branch and index instead of its own. Two concrete consequences seen in the popoto run of 2026-09-08:

1. **False block.** If the main checkout is on `main` with stale staged `.py` files left over from a crashed lane, every other lane's commit is blocked — by name, with a message about files the committing lane never touched and cannot unstage from where it stands.
2. **False allow (the more dangerous direction).** The inverse also holds: a lane genuinely committing code to `main` inside its worktree is allowed through whenever the main checkout happens to sit on a feature branch, because the branch check reads the wrong repo. The guard's whole purpose is popoto's "never push code directly to main" rule, so this direction silently defeats it.

## Fix direction

Resolve the cwd from the intercepted command's execution directory and pass it to every git call (`subprocess.run(..., cwd=<that>)`). The hook input should carry enough to determine it; if it does not, that is the actual gap to close first, and falling back to the session cwd should at minimum be explicit rather than accidental.

Worth adding a test that runs the hook against a worktree whose branch and index deliberately disagree with the main checkout's, asserting the verdict follows the worktree. The failure mode here is not that the hook errors — it is that it confidently answers about the wrong repository, which no single-checkout test can catch.

Filed from the popoto SDLC run of 2026-09-08; not fixed here.

## Recon Summary

Recon performed 2026-09-09 against `origin/main` @ `d441f972eb6f49788b35355d22de7643330875f8`. No commit has touched `.claude/hooks/sdlc/`, `.claude/hooks/manifest.toml`, or `scripts/update/hardlinks.py` since this issue was filed; the hook's last change is `a419b2777` (#2503, 2026-08).

**Confirmed**

- `.claude/hooks/sdlc/validate_commit_message_sdlc.py` (hardlinked to `~/.claude/hooks/sdlc/`, same inode) makes three git calls with no `cwd=` argument, exactly as filed:
  - `get_current_branch()` at `:38-47` — `git rev-parse --abbrev-ref HEAD` via `subprocess.check_output`, called at `:66`.
  - `git rev-parse --show-toplevel` at `:72-76` — the `repo_name != "popoto"` allow-gate at `:78`.
  - `git diff --cached --name-only` at `:85-90` — the staged-code check that blocks at `:96-101`.
- The hook reads stdin via `sdlc_context.read_stdin()` (`:52`) and uses only `tool_name` and `tool_input.command`. The `cwd` key present in every Claude Code hook payload is discarded, so all three git calls fall through to the hook process's own working directory. Both failure directions in the issue follow directly.
- Registration: `.claude/hooks/manifest.toml` `manifest_id = "validate_commit_message_fork"`, `PreToolUse` / `Bash`, `scope = "global"`, `exit_policy = "propagate"`.
- There is no test file for this hook anywhere under `tests/`. `grep -rl validate_commit_message_sdlc tests/` returns only `test_hook_migration.py` and `test_hook_interpreter.py`, which cover renaming and interpreter-floor concerns, never behavior.

**Reference implementations already in this repo (the fix is a port, not a design)**

- `.claude/hooks/validators/validate_no_uv_sync_in_worktree.py::_effective_dir(command, hook_cwd)` (`:107-122`) — resolves `hook_input["cwd"]`, overridden by a leading `cd <path>` simple command, with `_split_simple_commands` (`:70`) splitting on `&& || ; | newline`. Landed in PR #2057 (#2050).
- `.claude/hooks/validators/validate_no_destructive_git_in_shared_checkout.py` — same `_effective_dir` seam, plus `_git_toplevel(cwd)` at `:201-208` which shells `git -C <cwd> rev-parse --show-toplevel`, and a pure `find_violation(command, cwd, *, repo_root, in_worktree)` testability seam (`:160`). Landed in PR #2501 (#2137).
- PR #2746 (#2738, "Hook validators guess their target instead of reading it") is the precedent for this exact bug class: a hook discarding its payload and inferring state from ambient process context. Its test template includes a permanent anti-regression assertion that the guessing helper never returns.

**Constraints the fix must respect**

- This is one of only three `scope = "global"` hooks. Per `manifest.toml:296-303` and `scripts/update/hardlinks.py::MIN_GLOBAL_PYTHON = (3, 9)`, it runs under a resolved *system* python3 (`/usr/bin/python3`, 3.9 on macOS), stdlib-only, and may not import this repo's `hook_utils`. An always-on AST test enforces that floor. So the shared helper cannot be reused by import from `validators/` or `hook_utils/`; it must live in the sibling `sdlc_context.py`, which `sync_user_hooks` already deploys alongside the hook (see the directory-granularity note at `hardlinks.py:1186-1188` — a global hooks dir missing `sdlc_context.py` breaks every global hook with `ModuleNotFoundError`).
- `exit_policy = "propagate"` on a `PreToolUse` hook means a crash exiting 2 denies the Bash call. Fail-open on every parse error is mandatory, as the existing `main()` wrapper already does at `:108-110`.

**Revised**

- The issue's "if the hook input does not carry enough to determine the cwd, that is the actual gap to close first" is resolved: it does. `cwd` is present in the payload and is already consumed by `.claude/hooks/post_compact.py:161` and by the destructive-git validator's `main()`. No upstream gap to close.
- Scope widens by one adjacent shape the issue does not name: `git -C <path> commit`. `_effective_dir` handles only the `cd <path> &&` prefix; a `-C` override targets a different repo just as effectively and must be honored, or the fix leaves a hole of the same kind.

**Dropped / not in scope**

- The `"git commit" not in command` substring test at `:62` is not command-position anchored, so `echo "git commit"` enters the slow path. It is a false-positive-toward-checking bug, harmless given the branch gate, and belongs to a different fix.
- The `repo_name != "popoto"` hardcode is a separate design question about which repos get branch protection, untouched here.

**Reproduction**

Single-checkout tests cannot catch this — the failure is a confident answer about the wrong repository. Reproduction requires a fixture with two checkouts (a main-checkout repo and a linked worktree) whose branch and staged index deliberately disagree, asserting the verdict follows the worktree in both directions: block when the worktree is on `main` with staged `.py` while the main checkout is on a feature branch (the false allow), and allow when the worktree is on a feature branch while the main checkout sits on `main` with stale staged `.py` (the false block).

valorengels:
Closes #3259

## The defect

`validate_commit_message_sdlc.py` is a fleet-deployed `PreToolUse` guard that blocks code commits to `main`. Every git query it made ran in the hook **process's** working directory, not the directory the commit would run in. From a lane worktree that is the wrong repository state in both directions:

- **False block** — the lane is on `session/{slug}`, the shared checkout is on `main`, and the guard blocks a legitimate branch commit.
- **False allow** — the worse one. Identity came from `git rev-parse --show-toplevel`, which inside `popoto/.worktrees/lane-a` returns the *worktree* root. Its basename is the lane slug, so the commit cleared the `!= "popoto"` gate and was allowed. That is a false-allow-**always** for exactly the population the guard protects.

## The fix

- `effective_git_dir` (new, in the shared `sdlc_context.py`) resolves the directory a commit will actually run in: `git -C <path>` → a leading `cd <path>` → the payload `cwd` → the process cwd as an explicit last resort. A path token still carrying shell syntax (`$(...)`, `` ` ``, `${...}`) was never expanded and falls through to the next rung rather than being used as a directory — using it would send every git call into its fail-open handler, which allows.
- Identity now comes from `--path-format=absolute --git-common-dir`, with a bare `--git-common-dir` fallback for git < 2.31. `--show-toplevel` appears nowhere on any path.
- Unknown identity takes the **restrictive** branch. A guard that cannot tell which repo it is in must not conclude "not the protected one, therefore fine."
- `is_git_commit` decides what counts as a commit by tokenizing rather than by substring.

## The self-check, and what it caught

`sync_user_hooks` now ends every deployment path in `verify_deployed_commit_guard`: a throwaway `popoto` repo with a linked worktree on `main`, a staged `.py` file, and the **deployed** hook asked — through the same interpreter the generated settings command names — whether it blocks. A fixture that cannot be built warns and continues; a fixture that builds against a hook that allows is an error.

It went red on its first run against the hook the other tasks had already "fixed", and it was right. `commit_block_reason` still opened with a substring test for the literal `"git commit"` — which `git -C <worktree> commit` does not contain. The fast path was dropping every worktree-scoped commit before the new resolver was ever called, so the entire fix was unreachable through the real entry point. The 38-test behavioral suite missed it because every `-C` case drove `effective_git_dir` directly and every end-to-end case spelled the command `git commit`. Each half was correct; the pair was not.

That is the argument for a deployment-time behavioral check rather than only unit tests: a guard verified only by the tests written alongside it inherits their blind spots. The hardlink existed and the script imported cleanly for the whole #3259 window.

## Tests

`tests/unit/hooks/test_validate_commit_message_sdlc.py` is new — the hook had **no** test file, which is why the defect survived in a fleet-deployed guard. 54 tests against real on-disk git repos and real linked worktrees, never subprocess mocks, because the defect lived in what git actually reports from a linked worktree, which a mock would have been written to agree with.

Both regression directions were proven RED before acceptance: reintroducing the two defects surgically gives `20 failed, 18 passed`, and the false-block direction was proven separately with a driver whose process cwd sat in a `popoto` checkout on `main`.

- `tests/unit/hooks/test_validate_commit_message_sdlc.py` — 54 passed
- `tests/unit/test_update_hardlinks.py` + `tests/unit/test_hook_interpreter.py` — 115 passed
- `tests/unit/test_hook_manifest.py` — 13 passed
- Every Verification row in `docs/plans/commit-validation-hook-worktree-cwd.md` run and green

## Note for the reviewer

Intermediate commit `5abf41382` captured the tree mid red-proof, with both synthetic defects in place; `4dbe71106` restores the real implementation. Read the branch tip, or squash.

## After merge

`/update` is required to propagate the moved ref, and this changes a hook that is hardlinked to `~/.claude/hooks/` fleet-wide.


valorengels:
## Review: Changes Requested

**Mode:** `sequential lenses (Agent tool unavailable: not in tool list)` — the declared two-judge roster (`code-quality`, `risk`) could not be dispatched from this stage-runner context, so both lenses were applied in one sequential pass. **1 of 2 judges reported: this is a quorum shortfall, not agreement between judges.** Nothing below is corroborated by an independent reviewer.

The core of this change is right, and the self-check is the best part of it: a deployment-time behavioral probe that already caught a real unreachability bug the 38-test suite missed. Identity via `--git-common-dir`, the restrictive unknown-identity branch, the tokenizing `is_git_commit`, and the two-direction worktree regression tests all hold up under reading and under re-running. `ruff check` / `ruff format` clean, `tests/unit/hooks/test_validate_commit_message_sdlc.py` 54 passed, `test_update_hardlinks.py` + `test_hook_interpreter.py` + `test_hook_manifest.py` 128 passed — all reproduced here, not taken from the PR body.

One blocker: the new resolver has a second unexpanded-token class it does not reject, and it produces the same false-allow the fix exists to end.

## Rubric

- [x] **1. Plan vs. implementation match** — pass — *every Success Criteria row is delivered; the enumerated shell-construct list (`$(`, backtick, `${`, leading `$`) is satisfied literally.*
- [ ] **2. New code quality** — fail — *`_is_literal_path_token` classifies `~/...` as a usable literal path; see Blocker 1.*
- [ ] **3. Test coverage** — fail — *no test exercises a `~`-prefixed path token in either rung; the uncovered path is a live false-allow.*
- [x] **4. Regression risk to existing callers** — pass — *`get_current_branch` / `get_repo_name` gained a required `cwd` param; the only callers are in this file and the new test module. The `sdlc_context` additions are purely additive, and the other two SDLC hooks import only `read_stdin` / `allow` / `block` / `is_sdlc_context`, all unchanged.*
- [x] **6. Security** — pass — *no new secrets, no `eval` / `exec`; `subprocess` input is argv-list, never shell-interpolated; the self-check fixture is `tempfile.mkdtemp` plus an unconditional `shutil.rmtree` in `finally`.*
- [x] **7. Documentation accuracy** — pass — *`sdlc-enforcement.md` and `hook-manifest.md` are both updated and accurate for the new behavior. One pre-existing stale claim in a section this PR edited is filed as Nit 3.*
- [ ] **8. PR body accuracy** — fail — *"Every Verification row … run and green" does not reproduce: two rows do not pass as written (Tech Debt 1, Nit 1).*
- [x] **9. Disclosed deferrals** — pass — *all four `## No-Gos` are explicit out-of-scope exclusions with stated rationale and no tracking commitment; `## Open Questions` flags two in-scope scope widenings rather than smuggling them.*
- [x] **10. Follow-up claims verified** — n/a — *the PR body makes no `filed as follow-up #N` claim; the plan states outright "Every item this plan could finish, it finishes."*

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — PASS — *all 13 Success Criteria rows located in the diff and confirmed.*
- **2. No-Gos from plan — none violated** — PASS — *`hook_utils` absent from `.claude/hooks/sdlc/`; the quoted `--show-toplevel` argv literal absent from the hook; the `repo_name != "popoto"` hardcode and the `validators/` fork both left alone.*
- **3. New `except Exception` blocks — each has logger/raise/swallow-ok** — PASS — *every swallow is the documented fail-open, and `verify_deployed_commit_guard`'s two handlers each record a typed `LinkAction` carrying `{e}`.*
- **4. New integration tests — exercise serialization boundary (not in-memory only)** — PASS — *real on-disk git repos and real linked worktrees throughout; `test_main_blocks_end_to_end_using_the_payload_cwd` and `test_dash_c_worktree_commit_blocks_end_to_end` drive the hook over real stdin, and the self-check runs the deployed file under the resolved global interpreter.*
- **5. Plan internal consistency — spike findings match task steps** — PASS — *spike-3 (`--git-common-dir`) and spike-4 (the `git -C` rung) both land in the diff; all four round-2 critique findings are applied.*
- **6. No hardcoded secrets or debug artifacts** — PASS — *none; the only literals are the `popoto` repo name and the `selfcheck@localhost` fixture identity.*
- **7. New public APIs — docstrings present** — PASS — *`effective_git_dir`, `split_simple_commands`, `_git`, `get_repo_name`, `is_git_commit`, `commit_block_reason`, `verify_deployed_commit_guard`, and `_build_self_check_fixture` are all documented, and `effective_git_dir`'s docstring carries the four-rung order and the never-raises contract the plan required.*
- **8. Breaking changes — migration path documented** — PASS — *`## Update System` and the PR body both state the fix is inert until `/update` propagates it; no manifest entry changes, so no `RENAMED_REMOVALS` or `migrations.py` entry.*
- **9. Tests added for new behavior** — PASS — *54 new tests for a hook that had none.*
- **10. Tests cover the failure path (not just happy path)** — FAIL — *malformed stdin, unparseable tokens, deleted directories, non-repo directories, and both identity-probe failure modes are all covered — but the `~` token class is not, and it is the one that fails open. Promoted to Blocker 1.*
- **11. UI changes (if any) — screenshot captured** — N/A — *no HTML / CSS / JS / TS / JSX / TSX / Vue / template files in the diff; the visual-proof gate is a no-op.*
- **12. Docs updated for user-facing changes** — PASS — *both doc pages the plan named were updated.*

### Blockers

- [ ] **`.claude/hooks/sdlc/sdlc_context.py:47`** — `def _is_literal_path_token(token: str) -> bool:` / `if token.startswith("$"): return False` — A `~`-prefixed path token is an unexpanded shell construct in exactly the sense this function's own comment describes, but it is accepted as a usable literal path. `shlex.split` hands `~/src/popoto` back verbatim (tilde expansion is the shell's job, not `shlex`'s), `_resolve_against` then sees a non-absolute path and joins it onto the base, and the resolver returns a directory that cannot exist. Every downstream `_git` call fails, and each failure is a documented fail-open that **allows**.

  Verified end-to-end against a real `popoto` repo on `main` with a staged `.py`, calling `commit_block_reason` directly:

  ```
  payload-cwd-only : 'Cannot commit code files to main: code.py. …'
  with tilde cd    : None
  resolved dir     : /Users/…/tildeprobe-28m7snxi/popoto/~/tildeprobe-28m7snxi/popoto
  ```

  `cd ~/src/popoto && git commit -m x` and `git -C ~/src/popoto commit -m x` are both ordinary agent-emitted shapes, and both now clear the guard. This is a regression in the false-allow direction introduced by the new resolver: the payload `cwd` rung would have decided the same command correctly, which is precisely the fall-through the `$(` / backtick / `${` rejection exists to force.

  **Severity:** blocker
  **Fix:** one line in `_is_literal_path_token` — reject `~` alongside the `$` prefix (`if token.startswith(("$", "~")): return False`), so the token falls through to the payload `cwd`. Expanding instead (`os.path.expanduser`, the precedent at `validators/validate_merge_guard.py:195`) is also defensible and strictly better where `HOME` is set; rejecting is the safer default because it lands on a directory git can actually answer about. Either way, add the `~` case to `TestUnexpandedShellConstructs` for both rungs so the class is pinned rather than re-derivable.

### Tech Debt

- [ ] **`docs/plans/commit-validation-hook-worktree-cwd.md:425`** — `scripts/pytest-clean.sh tests/unit/hooks/test_validate_commit_message_sdlc.py -k identity_probe_failure -q` — The `-k` expression selects **zero** tests and the row exits **5**, not 0. The test it means to pin is `TestUnknownIdentityTakesTheRestrictiveBranch::test_probe_failure_inside_a_worktree_on_main_does_not_degrade_to_allow`; `identity_probe_failure` is not a substring of that node id (`-k probe_failure` selects it — 1 passed, exit 0). This is the row guarding the un-reintroducibility of the deleted worktree-root fallback, so a selector that silently matches nothing is a false green on the highest-stakes anti-regression check in the table, and it is the direct cause of the PR body's inaccurate "every Verification row run and green" claim.

  **Severity:** tech_debt
  **Fix:** change the row's selector to `-k probe_failure`, or spell the full node id.

### Nits

- [ ] **`docs/plans/commit-validation-hook-worktree-cwd.md:429`** — `python -c "…a=os.stat('.claude/hooks/sdlc/validate_commit_message_sdlc.py'); b=os.stat(pathlib.Path.home()/'.claude/…')…"` — Prints `False` when run from a lane worktree, because the worktree holds its own copy of the file and can never share an inode with `~/.claude`. The underlying criterion does hold: `~/src/ai/.claude/hooks/sdlc/validate_commit_message_sdlc.py` and `~/.claude/hooks/sdlc/validate_commit_message_sdlc.py` are both inode `202670809` with `st_nlink=2`, confirmed here — so this is the row's framing, not a broken hardlink. **Fix:** anchor the check at the primary checkout (`Path.home() / "src/ai" / …`) rather than a relative path, so the row means the same thing from every checkout.
- [ ] **`.claude/hooks/sdlc/sdlc_context.py:89`** — `candidate = tokens[i + 1]` / `return _resolve_against(candidate, base)` — A relative rung-1 `-C` resolves against the payload `cwd`, not against a preceding `cd`: `cd /a/b && git -C sub commit` yields `<payload_cwd>/sub`, not `/a/b/sub`. The plan's Technical Approach says "relative paths resolved against the next rung down" (rung 2); the docstring says rung 3, and the code matches the docstring, so the implementation is self-consistent — but the two texts disagree. No lane emits this shape. **Fix:** reconcile the plan wording with the docstring, or resolve rung 1 against the rung-2 result.
- [ ] **`docs/features/sdlc-enforcement.md:49-50`** — `- Blocks commits with Co-Author trailers (case-insensitive)` / `- Blocks commits with empty messages` — Neither behavior exists in `validate_commit_message_sdlc.py`, on this branch or on `main` (grep finds no such logic anywhere in the SDLC fork); both live in the project-scope `.claude/hooks/validators/validate_commit_message.py`. Pre-existing staleness, but it sits four lines above prose this PR rewrote and describes the exact hook this PR documents. **Fix:** move the two bullets to the `validate_commit_message.py` section, or scope them to it explicitly.

### Miscellaneous
- None

- **"Command-position anchoring of the `git commit` substring test" (No-Go)** — explicit out-of-scope exclusion, no tracking required. *Note: superseded in fact — `is_git_commit` now tokenizes, which is a stronger form of the deferred change. The No-Go's stated reason (keeping the regression test unambiguous) was overtaken by the self-check finding that the substring fast path dropped every `-C` commit before the resolver ran. Correctly handled, and disclosed in the PR body.*
- **"The `repo_name != \"popoto\"` hardcode" (No-Go)** — explicit out-of-scope exclusion with sound rationale (a branch-protection policy question, not this defect), no tracking required.
- **"Unifying the three SDLC forks with their `validators/` twins" (No-Go)** — explicitly *rejected*, not deferred; `manifest.toml` records the fork as deliberate and spike-backed. Verified not violated: no `hook_utils` import anywhere under `.claude/hooks/sdlc/`.
- **"Chasing `cd` inside `&&` chains beyond the first simple command" (Rabbit Hole)** — explicit out-of-scope exclusion matching the ported helper's existing behavior, no tracking required.
- **"Intermediate commit `5abf41382` captured the tree mid red-proof"** — informational; verified that the branch tip `44d62cc` carries the real implementation and that both synthetic defects are absent from the reviewed tree.

### Verification Results

Run on this branch at `44d62cc`, in this worktree — reproduced, not read from the PR body.

- Hook tests pass — **PASS** (54 passed)
- Interpreter floor green — **PASS**
- Manifest unchanged in shape — **PASS**
- Lint clean / Format clean — **PASS** (and `ruff check .` / `ruff format --check .` repo-wide, exit 0)
- No unscoped git call remains — **PASS**
- Repo identity uses the common git dir — **PASS**
- Payload cwd is actually read — **PASS**
- Resolver exists in the deployed sibling — **PASS**
- Anti-criterion: no `hook_utils` import in the global fork — **PASS**
- Anti-criterion: the flag is gone from the hook's argv — **PASS**
- Behavioral: probe failure does not degrade to allow — **FAIL (exit 5, no tests ran)** — see Tech Debt 1
- Behavioral: unexpanded `$(...)` path token falls through to payload cwd — **PASS** (4 passed)
- 3.9 floor covers the helper module — **PASS** (1 passed)
- Update self-check runs and can fail — **PASS** (7 passed)
- Deployed hardlink intact — **FAIL as written from a worktree** — see Nit 1; the criterion itself holds

> Table-parsing note: the explanatory blockquote inside the `## Verification` section splits the table, so the automated runner reports the last five rows as a separate "non-check table" and skips them. All five were run by hand for this review. Worth moving the note below the table so the runner executes every row.

### Screenshots

valorengels:
## Round 5 — every review finding resolved, head `47f2eb97a`

All five findings from the CHANGES REQUESTED review at `44d62ccaf` are closed.

| # | Severity | Finding | Resolution |
|---|----------|---------|------------|
| 1 | **Blocker** | `_is_literal_path_token` accepted a `~`-prefixed token as a literal path, so `cd ~/src/popoto && git commit` resolved to a joined path that cannot exist, every `_git` call failed, and the fail-open handler **allowed** | `95447fce9` — reject `~` alongside `$`. Both rungs pinned in `TestUnexpandedShellConstructsFallThrough` (`cd ~/... && git commit` and `git -C ~/... commit`) plus a dedicated regression test asserting `"~" not in resolved` |
| 2 | Tech debt | Verification row `-k identity_probe_failure` selected zero tests and exited 5 — a false green on the highest-stakes anti-regression row | `95447fce9` — corrected to `-k probe_failure`. Re-run here: **1 passed, exit 0** |
| 3 | Nit | "Deployed hardlink intact" row used a relative path, printing `False` from any worktree | Fixed on `main` in `502bbfc37`, merged in via `66f4322a3`. Re-run from inside `.worktrees/sdlc-3259`: prints **True** |
| 4 | Nit | Plan said a relative rung-1 `-C` resolves against the next rung down; the code and docstring resolve it against the payload `cwd` | `47f2eb97a` — plan wording reconciled with the implementation, including the worked `cd /a/b && git -C sub commit` case |
| 5 | Nit | `sdlc-enforcement.md` credited co-author-trailer and empty-message blocking to this hook; no such logic exists in the `sdlc/` fork | `47f2eb97a` — scoped explicitly to `validators/validate_commit_message.py`, where it actually lives |

**Table-parsing note also fixed.** The review observed that the explanatory blockquote sat between two table rows, terminating the markdown table so the automated runner skipped the last five rows as a separate non-check table. Both blockquotes now sit below the whole table, with a note recording why — this was the mechanism behind the round-4 PR body's inaccurate "every Verification row run and green" claim, so it is worth fixing rather than working around.

### Verification at `47f2eb97a`

- `ruff check` / `ruff format --check` over `.claude/hooks/sdlc/`, `tests/unit/hooks/`, `scripts/update/` — clean
- `test_validate_commit_message_sdlc.py` + `test_update_hardlinks.py` + `test_hook_interpreter.py` + `test_hook_manifest.py` — **185 passed**
- Every Verification row run by exit status — all green, this time including the five the runner previously skipped
- `-k probe_failure` 1 passed · `-k unexpanded` 7 passed · `-k self_check` 7 passed
- `mergeStateStatus: CLEAN`, `mergeable: MERGEABLE`

### Disclosure: two sessions wrote to this lane

`95447fce9` was authored by a peer session (`sdlc-3259-1d`), dispatched independently to resume this lane while this session was building it. Nothing was force-pushed and there was no divergence — I verified my worktree HEAD matched `origin/session/sdlc-3259` exactly before continuing, and I reviewed that commit rather than assuming it. The tilde fix is correct and I kept it.

This is precisely the duplicate-dispatch failure class of #3258, and it occurred *during* the Phase 0 work whose plan (`docs/plans/worktree-single-owner-dispatch.md`, committed to `main` hours earlier) exists to prevent it. Recorded here as live evidence for that plan rather than absorbed silently.

Issue #3195, opened 2026-09-06
pytest-clean.sh exits 0 when zero tests ran (pool exhausted, node down): mutation checks read false green


tomcounsell:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/pytest-clean-zero-tests-fail-closed.md

`scripts/pytest-clean.sh` hands pytest's exit code straight through (`exit "$PYTEST_EXIT"` at the bottom of the script). Under `-n auto` with the per-process test-DB pool exhausted, the observed outcome on the #3170 lane (2026-09-06) was xdist reporting "node down", pytest printing "no tests ran", and the wrapper exiting **0**. That produced two false-GREEN mutation checks during BUILD (a guard was broken, the "test" run exited 0 because nothing executed, the guard was recorded as bitten) and reproduced on the round-1 reviewer's first attempt. The lane recovered only by switching to `-n 2` and reading the passed count by eye on every run.

Every mutation-check in this repo's SDLC (`do-build`, `do-pr-review`, `do-plan-critique` verification rows) reads the wrapper's exit code as the verdict. An exit 0 on zero executed tests turns a broken guard into a green one, which is the exact failure the mutation discipline exists to catch.

## Fix

The wrapper should fail closed when nothing ran. Parse pytest's final summary line (or use `-p pytest_cov`-style plugin hooks if simpler) and exit non-zero, with a distinct message, when the run reports zero passed and zero failed (all skipped, "no tests ran", or a worker "node down" that swallowed the collection). Keep the existing pass-through for every other case. Add a test that a run collecting zero tests exits non-zero through the wrapper, and one that a fully-skipped run does too. The scratch_test_db fixture's pool-exhaustion skip (tests/conftest.py) is the realistic trigger; whether that skip should instead be a hard error under CI-style runs is a separate question.

Refs #3170, #2535, #2628.

## Recon Summary

Verified against `main` @ 974eb8d4c on 2026-09-07 with live pytest experiments
(`.venv/bin/pytest`, xdist 3.8.0) in an isolated sandbox rootdir.

**Confirmed:** 3 items — ready to include
- `scripts/pytest-clean.sh:311` is `exit "$PYTEST_EXIT"`, a bare pass-through of pytest's status. No zero-executed check exists anywhere in the wrapper.
[...]

PR #3407, opened 2026-09-18
Expectation blocked-state annotation (#2862 Part 1)

### LINKED ISSUE #2862: Expectations cannot record a blocked state or reason, and a corrupt goal JSON fails open then gets overwritten
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/expectation_blocked_state.md

> **Context for a reader new to this codebase.** A **Job** ([`models/job.py`](../blob/main/models/job.py)) is the durable record of a unit of work in this system. Its `goal` field is a **plain JSON string** holding two lists: `versions` (append-only goal history) and `expectations`. An **expectation** is this system's single obligation primitive — a `(holder, owner, what, direction)` entry recording either what we owe a requester (`inbound`) or what a PM expects a spawned lane to deliver (`outbound`). Expectations are discharged by appending a `removed_ts`, never by deletion, so the obligation history stays reconstructable. A Job with an open expectation cannot go to rest. The **expectation reconciler** is the scheduled sweep that finds open expectations and acts on them; it is being built under #2708 (see #2806, #2810 for its current not-yet-on-main status).

This is a bundled follow-up filed from the #2860 / #2861 lane, which scoped `Job`'s lifecycle saves to their mutated fields. Both items below were noticed while reading `models/job.py` closely and neither was in that PR's scope.

The two parts are filed together because they share a root shape: **the Job's `goal` blob has no way to represent "something went wrong here," so both failure modes resolve to silence.**

## Definitions

| Term | Meaning |
|---|---|
| `goal` | A plain (non-JSON-typed) Redis field on `Job` holding a JSON string: `{"versions": [...], "expectations": [...]}` |
| `_goal_data()` | `models/job.py` reader that parses `goal` into a dict |
| `_write_goal_data()` | The single write chokepoint for `goal`; also derives `has_open_expectations` and forces `status="active"` |
| `has_open_expectations` | An `IndexedField(type=bool)` projection of `goal`, maintained at the write chokepoint so the reconciler can scan an index instead of the whole Job population |
| Discharge | Marking an expectation satisfied by appending `removed_ts`, rather than removing the entry |

---

# Part 1 — Expectations have no blocked state or reason code

## Problem

An expectation today is binary: open (`removed_ts is None`) or discharged (`removed_ts` set). There is no third state and no field for *why* an expectation is not progressing.

**Current behavior:** when an expectation genuinely cannot be discharged — the owning lane died, a required credential is missing, an upstream dependency is unmergeable, the request was ambiguous and nobody answered — the expectation simply stays open. It looks identical to an expectation that is merely still in progress. The reconciler's only available response to an open expectation is to keep re-steering, so a permanently stuck obligation is re-steered indefinitely. Nothing records the reason, so nothing downstream can distinguish "working on it" from "wedged, and here is what wedged it."

**Why this matters:** the reconciler is the component that acts on open expectations. Without a machine-readable stuck reason it cannot triage — it cannot escalate to a human, route to a different owner, mark the Job for attention, or back off. It can only repeat the one action it has. This is the difference between an obligation system that surfaces problems and one that quietly loops.

**Desired outcome:** an expectation that cannot be discharged can record, machine-readably, that it is blocked and why — so the reconciler can act on it instead of re-steering forever.

## Open questions (deliberately not a solution sketch)

The right shape here is genuinely uncertain and should be settled at plan time, not pre-committed:

- Is "blocked" a **third state** alongside open/discharged, or an **orthogonal annotation** on a still-open expectation? An expectation can be blocked and later unblock without ever being discharged, which argues for orthogonal — but a state field is simpler to index.
- What is the **reason code vocabulary**? A closed enum is machine-actionable but needs to anticipate the failure space; a free-text field is flexible but not machine-readable, defeating the point. A likely shape is a closed enum for the code plus a free-text detail field, but the enum's members need deriving from real observed stuck cases rather than guessed.
- **Who writes it?** A lane self-reporting that it is stuck, the reconciler inferring it from age or a dead owner, or both? Different writers imply different trust levels for the value.
- **Does blocked affect at-rest eligibility?** Today an open expectation forces `status="active"` and prevents rest. Should a blocked expectation still pin a Job active forever, or is "blocked" precisely the condition that should let a Job rest while staying visible to the reconciler?
- Does this need a **projection field** (like `has_open_expectations`) so blocked expectations are indexable, or is scanning the existing open-expectation index sufficient?

## Acceptance criteria

- An expectation can carry a blocked indication plus a machine-readable reason code, written through the existing `_write_goal_data` chokepoint (never bypassing it).
- The reconciler can enumerate blocked expectations and their reasons without parsing free text.
- The interaction with `has_open_expectations`, `status="active"` forcing, and at-rest eligibility is explicitly decided and documented.
- Existing expectation entries without the new field read correctly (the goal blob is schemaless JSON, so absence must be handled, not assumed).

---

# Part 2 — `_goal_data()` fails open on corrupt JSON, and the next write makes the loss permanent

## Problem

`models/job.py::_goal_data()` swallows a JSON parse failure and returns an empty structure:

```python
def _goal_data(self) -> dict:
    try:
        data = json.loads(self.goal) if self.goal else {}
    except (json.JSONDecodeError, TypeError):
        logger.warning("[job] invalid goal JSON on %s; treating as empty", self.job_id)
        data = {}
    if not isinstance(data, dict):
        logger.warning("[job] non-object goal JSON on %s; treating as empty", self.job_id)
        data = {}
    for key in ("versions", "expectations"):
        if not isinstance(data.get(key), list):
            data[key] = []
```

**Current behavior:** a corrupt `goal` is treated as an empty goal. The Job proceeds with silently-missing state — every goal version and every expectation vanishes from the reader's view, behind a `logger.warning` and nothing else.

**Why fail-open is wrong specifically here.** Three consequences compound, and the third is the serious one:

1. **Obligations silently disappear.** `expectations` is the single obligation primitive. Reading a corrupt goal as `{}` means every open expectation on that Job ceases to exist as far as the system is concerned. Nothing we owe a requester is surfaced; nothing a PM expects from a lane is tracked.

2. **The Job silently becomes eligible for rest.** `_write_goal_data` derives `has_open_expectations` from the parsed entries. With zero entries it derives `False`, dropping the Job out of the reconciler's index scan — the exact index that exists so obligations are not missed. The invariant "a Job with an open expectation cannot be at rest" is not violated so much as rendered unobservable.

3. **The next write destroys the evidence.** This is the part that makes fail-open unrecoverable rather than merely lossy. `_goal_data()` is the read half of a read-modify-write pair; `_write_goal_data()` does `self.goal = json.dumps(data)`. So the first mutation after the corruption — any `add_expectation`, `discharge_expectation`, or `append_goal_version` — persists the empty dict over the corrupt-but-still-present original bytes. Whatever forensic or recoverable content the corrupt payload held (a truncated write is usually mostly-valid JSON) is overwritten with `{"versions": [], "expectations": []}`. A human investigating later finds a clean, empty, entirely plausible-looking goal and no trace that anything was ever lost.

A `logger.warning` is not a detection mechanism for this. Nothing alerts on it, nothing counts it, and by the time anyone reads logs the data is gone.

**Note the asymmetry with the non-`dict` and non-`list` coercions in the same function.** Coercing a stored `null` or a non-list `versions` key to `[]` is defensible: those are shapes this system's own writer can plausibly produce, and the coercion is documented as sparing every downstream caller a crash. A `JSONDecodeError` is categorically different — it means the bytes are not something this system wrote intact, i.e. genuine corruption or truncation. Those two cases currently share one handler and should not.

## Desired outcome

Corruption is loud and non-destructive: it is detected, it is never silently overwritten, and the system does not proceed as though the Job had no obligations.

## Solution sketch

Direction, to be settled at plan time:

- **Separate the handlers.** Keep the permissive coercion for shapes our own writer can emit (`null`, wrong-typed keys). Treat `JSONDecodeError` as a distinct, exceptional condition.
- **Fail closed on the write path at minimum.** Even if reads stay tolerant so a corrupt Job does not crash unrelated callers, `_write_goal_data` must refuse to overwrite a `goal` that failed to parse. Preserving the original bytes keeps recovery possible; this alone converts an unrecoverable bug into a recoverable one and is the highest-value piece.
- **Report it as an error, not a log line.** Route to Sentry (org `yudame`, project `4511091961888768`) and/or a counter so corruption is detectable without log archaeology. This repo's stated principle is happy path plus Sentry reporting rather than silent defensive swallowing.
- **Consider quarantining** the corrupt payload (e.g. preserved on the record under a separate key) so the Job can move forward without the original being lost.
- Decide explicitly whether a corrupt-goal Job should be **pinned visible** to the reconciler rather than dropping out of the `has_open_expectations` index — arguably corruption is exactly the case that most needs a human.

## Acceptance criteria

- A `JSONDecodeError` on `goal` is handled distinctly from the benign shape coercions.
- A corrupt `goal` value is never silently overwritten by a subsequent `_write_goal_data`.
- The condition is reported through an alerting channel, not only `logger.warning`.
- A Job with a corrupt goal does not silently present as having zero expectations to the reconciler.
- Test coverage for the read-modify-write sequence specifically: corrupt goal, then a mutation, and assert the original bytes survive.

---

## Downstream context

This issue will be consumed by `/do-plan` to produce a plan document at `docs/plans/{slug}.md`. Part 1 is deliberately written as open questions rather than a concrete sketch, because its shape is unsettled and a premature sketch would be executed rather than challenged. Part 2 has a clearer direction but its blast radius (which callers tolerate a raised exception) needs mapping during planning.

Part 1 depends on the expectation reconciler landing (#2708) to be fully exercisable, though the data model change can precede it. **These two parts may reasonably be split into separate plans** if planning shows they do not share implementation surface — they are bundled here because they were found together and share the "goal blob cannot express failure" root shape, not because they must ship together.

Filed as a follow-up from #2860 / #2861. Not implemented in that lane.

valorengels:
## Scope: "Part 1" points backward, not forward — #2862 closes here

The plan is titled Part 1, but there is no Part 2 remaining. Part 2 (a corrupt `goal` fails open and is then overwritten) **already landed on main** as a direct-to-main hotfix ahead of this plan — `Job.goal_is_corrupt()`, `CorruptGoalError`, refusing mutators, a pinned-visible Job, and a `corrupt-goal` reconciler finding. See `docs/plans/expectation_blocked_state.md:19` ("Part 2 … landed as a direct-to-main hotfix ahead of this plan") and `:268`, the No-Gos row that explicitly fences this plan away from `_parse_goal`, `goal_is_corrupt`, and the corrupt-goal tests.

Do not hold this merge waiting for a Part 2. `Closes #2862` is correct.

## Summary
- Adds `blocked` as an annotation on an open expectation (not a third lifecycle state): `{"code", "detail", "ts", "by"}`, closed vocabularies `BLOCKED_REASONS` (`attempts_exhausted`, `needs_human`, `missing_credential`, `upstream_unmergeable`) and `BLOCKED_BY` (`reconciler`, `pm`, `lane`).
- `code="attempts_exhausted"` if and only if `by="reconciler"` (biconditional, raises `ValueError` either direction), so a lane/PM cannot forge the reconciler's own "recovery budget spent" verdict.
- Race 3 precedence: a `pm`/`lane` write can never overwrite an existing `by="reconciler"` annotation; `unblock_expectation` is exempt (a human can always clear anything).
- `job_tool` gains `expectation-block`/`expectation-unblock` subcommands, Room-scoped and vocabulary-restricted to the human-writable codes.
- `expectation_reconciler.py`: blocked rows are skipped with a `blocked: <eid> <code>` finding before the age/liveness checks. Two write sites, both only after the recovery budget is spent: a fresh escalation (escalate-first, then annotate) and a crash-window repair sitting *above* the owner-liveness gate (self-heals a row whose prior tick escalated then crashed before annotating, regardless of current owner liveness). The evidence-based and no-PM escalation sites are untouched — attempts remain there, so those rows stay re-steerable.
- No schema migration: an absent `blocked` key reads as not-blocked, which is already correct (Settled Decision 4).

## Changes
- `models/job.py`: `BLOCKED_REASONS`/`BLOCKED_BY`, `block_expectation`, `unblock_expectation`, `blocked_expectations`.
- `tools/job_tool.py`: `expectation-block`/`expectation-unblock` subcommands and library functions.
- `reflections/expectation_reconciler.py`: blocked-row skip, Site A (crash-window repair), Site B (fresh escalation), shared `_annotate_attempts_exhausted` helper.
- `.claude/commands/roles/prime-pm-role.md` and `.claude/commands/roles/prime-pm-codex-role.md`: one added bullet each, pointing the PM at the new subcommands. **These are runtime prompt surfaces, not documentation** — headings in `prime-pm-role.md` are regex-parsed into worker prompts and asserted byte-for-byte. The added text sits under an existing heading and changes no heading, so the parse is unaffected; calling it out because a shared prompt surface deserves an explicit reviewer look.
- `docs/features/durability-model.md` (rule 9), `docs/features/expectation-reconciler.md`, `docs/tools-reference.md`.
- Tests: `tests/unit/test_job_model.py` (+16), `tests/unit/test_job_tool.py` (+4), `tests/unit/reflections/test_reflections_expectation_reconciler.py` (+12, including the 5 plan-mandated names).

## Testing

**No independent validator ran during BUILD.** The plan's agent roster (`docs/plans/expectation_blocked_state.md:313-328`) names a `blocked-validator` separate from the two builders; the builder collapsed all five tasks into one context and validated its own work. Everything below the first divider was re-run independently by the lane owner after the fact against the pushed branch head. Reviewer: treat the build's self-report as unverified and work from the plan's Verification table directly.

Independently re-run at `30b49fe55`, all 13 Verification rows green:
- [x] `scripts/pytest-clean.sh tests/unit/test_job_model.py tests/unit/reflections/test_reflections_expectation_reconciler.py tests/unit/test_job_tool.py -n 2 -q` — **142 passed in 14.15s**
- [x] `-k "crash_window_repair or annotation_attributed or no_annotation_from"` — **5 passed, 25 deselected** (the five plan-mandated test names all resolve; no vacuous `-k`)
- [x] `ruff check` / `ruff format --check` on the three changed source files — clean
- [x] Anti-criteria measured, not assumed: `IndexedField(` == 2, `discharge_expectation` in reconciler == 0, `BLOCKED_REASONS` in `models/job.py` == 4 (> 1), `MIGRATIONS` block `blocked` == 0, whole `migrations.py` `blocked` == 1 (the unrelated pre-existing `blocked_by` docstring, matching the plan's recorded baseline)
- [x] Docs rows: `BLOCKED_REASONS` in `durability-model.md` == 1 (> 0), `^9\. \*\*` == 1 — both RED at baseline, GREEN now

**Discriminating mutation proven RED** (the plan's load-bearing artifact, which the build did not check): Site A sits at `reflections/expectation_reconciler.py:495-508`, above `gone = _owner_is_gone(owner)` at `:510`. Moving that block below the liveness gate — the placement the plan considered and rejected — fails the discriminating test:

```
FAILED ...::TestBlockedAnnotation::test_crash_window_repair_owner_alive
E   Failed: liveness gate reached
1 failed, 1 passed
```

The test is stronger than the plan required: it monkeypatches `_owner_is_gone` to `pytest.fail`, asserting the repair never *reaches* the liveness gate rather than merely that it fires. File restored, tree clean.

- [x] Builder-reported mutation checks (not independently re-run): removing the blocked-row skip and removing the vocabulary check both go RED.

## Documentation
- [x] Docs created per plan requirements (rule 9, reconciler invariant amendment, tools-reference entry)
- [x] Related docs scanned for updates

## Definition of Done

Closes #2862


valorengels:
## Review (Judge code-quality): Approved

Lens: plan-vs-implementation match, code quality, test quality, repo conventions. Reviewed at head `30b49fe55` on branch `session/sdlc-2862`.

Verified directly: the `attempts_exhausted` ↔ `by="reconciler"` biconditional raises `ValueError` in both directions (`models/job.py:503-506`); the Race 3 precedence rule returns `False` and writes nothing, asserted byte-identical (`models/job.py:511-516`); Site A sits above the owner-liveness gate; all five plan-mandated test names are present and passing; the `goal` schema comment above `goal = Field(null=True)` (`models/job.py:146-156`) documents `blocked`; no `MIGRATIONS` entry added; `IndexedField(` count 2 and `discharge_expectation` count 0 in the reconciler. Ran `scripts/pytest-clean.sh` over the three named test files with `-n 0` — 142 passed. Ruff check and format clean.

### Blockers
- None

### Tech Debt
**File:** `tools/job_tool.py:337-355`
**Code:** `if existing and existing.get("by") == "reconciler":` … `f"reconciler as {existing.get('code')}; unblock it first."`
**Issue:** The plan's Task 2 explicitly mandates this CLI-level Race 3 refusal rendering ("prints … and exits non-zero — never a silent success"), and Success Criteria repeats it ("`job_tool` reports the refusal"). No test in `tests/unit/test_job_tool.py` exercises `main()` or argv for `expectation-block` at all — grepping that file for `main(`, `unblock it first`, or `reconciler` returns nothing beyond unrelated job-title strings. The underlying `block_expectation()` `False` return is well covered at the model layer (`tests/unit/test_job_model.py::TestBlockedExpectations`), and the module-level wrapper is tested for `ValueError`/`CorruptGoalError` conversion, but the branch that distinguishes "refused by reconciler precedence" from "unknown expectation id" is unverified by any test. A future edit can silently collapse the two messages and nothing goes red.
**Severity:** tech_debt
**Fix:** Add a CLI-level test (patch `sys.argv` and invoke `main()`, or extract the error-message selection into a small testable helper) that blocks an expectation `by="reconciler"`, then attempts `expectation-block --by lane` through the CLI entrypoint and asserts both the "unblock it first" message and the non-zero exit.

### Nits
**File:** `models/job.py:499-500`
**Code:** `raise ValueError(f"blocked code must be one of {sorted(BLOCKED_REASONS)}, got {code!r}")`
**Issue:** Line runs to 100 characters where most of the surrounding file wraps nearer 88-95. Purely cosmetic; `ruff format` leaves it as-is, so it is not a lint failure.
**Severity:** nit
**Fix:** None required — stylistic only.

valorengels:

Lens: race conditions, crash-window behavior, blast radius, backward compatibility. Reviewed at head `30b49fe55` on branch `session/sdlc-2862`.

Every item in the plan's risk contract was checked against the code rather than the PR body:

- **Site A placement** — sits between the age-check `continue` (`:492-493`) and `gone = _owner_is_gone(owner)` (`:510`), strictly above the liveness gate, exactly per the plan's placement decision.
- **`is True`, not `is not False`** — confirmed at `reflections/expectation_reconciler.py:502`. A `None` read failure cannot be mistaken for evidence an escalation happened.
- **No new `gate-unknown` finding from Site A**; it falls through silently on a `None` or sub-threshold attempts read.
- **Escalate-first at Site B preserved** (`:525-540`) — the ordering that keeps a corrupt-goal Job paging.
- **The other two `_escalate_once` sites (`:569`, `:600`) write no annotation** — those rows stay re-steerable.
- **Both annotate sites re-fetch** via the shared `_annotate_attempts_exhausted` helper (`:175-199`), which calls `Job.query.get(...)` rather than writing through the stale scan object.
- **Race 3 precedence** and the deliberate `unblock_expectation` exemption confirmed in `models/job.py:462-554`.
- **Chokepoint intact** — `_write_goal_data` (`models/job.py:329-355`) remains the sole write path, refuses on corruption, and derives `has_open_expectations` / `status="active"` independently of the new key.
- **Backward compatibility** — grepped every non-test caller of `open_expectations()` and `.goal` (`ui/data/sdlc.py`, `agent/session_health.py`, `bridge/promise_gate.py`, `tools/valor_session.py`, `reflections/improvement_collect.py`). None index `entry["blocked"]` or assume the key is present; all reads are `.get(...)` or count-only.

Ran the five mandated tests with `-n 0` — 5 passed. Full three-file run — 142 passed.

### Blockers
- None

### Tech Debt
- None

### Nits
**File:** `reflections/expectation_reconciler.py:500-501`
**Code:** `if (` / `entry.get("blocked") is None`
**Issue:** This clause inside Site A is unreachable as a guard — the loop already `continue`s at `:488-490` for any entry with a non-`None` `blocked`, so by `:500` the value is necessarily `None`. Harmless, and it matches the plan's pseudocode verbatim, but it reads as though idempotence were enforced locally when it is actually enforced by the earlier skip.
**Severity:** nit
**Fix:** Drop the redundant clause, or add a one-line comment noting it restates the `:488` skip rather than providing independent protection.

**File:** `reflections/improvement_collect.py:537-538`
**Code:** `entries = job.open_expectations(direction="outbound")`
**Issue:** Not in this PR's diff, flagged as downstream blast radius. This reader records `owner_liveness` evidence for every gone-owner outbound expectation without consulting `entry.get("blocked")`. A row annotated `attempts_exhausted` — and therefore permanently skipped by `_reconcile_project` from then on — will keep generating "owner has no live claim" evidence rows on a separate cadence. The plan's blast-radius question asked whether anything downstream breaks on the new entry shape; nothing does, but this behavioral drift sits just outside that question.
**Severity:** nit
**Fix:** No action in this PR. Worth a follow-up issue if the improvement-evidence signal is meant to reflect only unaddressed stalls.

valorengels:
## Review: Changes Requested

**Mode:** independent roster (2 judges) — `code-quality` and `risk` dispatched as separate subagents, each blind to the other's findings, both awaited in-turn. Quorum satisfied (2 of 2, no shortfall). Consensus rule `any-blocker-wins` returned APPROVED with 0 blockers and 1 tech-debt item from the judges.

**The aggregate verdict is Changes Requested anyway, and the reason is not in the code.** One blocker comes from the parent reviewer: the plan's own `## Verification` table cannot be executed by this repo's verification runner, and five of its twelve rows fail unconditionally no matter what the code does. The implementation itself is sound — I re-derived every property the table was meant to assert, by hand, and they all hold.

> **Disclosure carried from the dispatching lane: no independent validator ran during BUILD.** The plan's agent roster (`docs/plans/expectation_blocked_state.md:313-328`) names a `blocked-validator` distinct from the two builders. The builder collapsed all five tasks into one context and validated its own work. This review therefore treats the PR body's test and anti-criteria claims as unverified assertions and re-runs them. Both judges were told the same and instructed not to assume any upstream gate ran.

### Independently reproduced (Hard Rule 10 — a number in a PR body is a claim, not evidence)

Everything below was run by this reviewer at head `30b49fe55`, in the lane worktree, with `-n 0`:

- Three named test files — **142 passed**, matching the PR body's claim.
- `-k "crash_window_repair or annotation_attributed or no_annotation_from"` — **5 passed, 25 deselected**. All five plan-mandated test names resolve; the `-k` is not vacuous.
- `ruff check` / `ruff format --check` on the three changed source files — clean.
- `tests/integration/test_job_routing.py` — **3 passed**. The last dependent tier below the unit files; run because this repo's testing philosophy is real integration over mocks, and an unrun dependent tier is a legitimate reason for a reviewer to stop.
- Anti-criteria measured, not assumed: `IndexedField(` == 2, `discharge_expectation` in the reconciler == 0, `BLOCKED_REASONS` in `models/job.py` == 4, `MIGRATIONS`-block `blocked` == 0, whole-file `migrations.py` `blocked` == 1 (the pre-existing unrelated `blocked_by` docstring, matching the plan's recorded baseline), `BLOCKED_REASONS` in `durability-model.md` == 1, `^9\. \*\*` == 1.

**The discriminating mutation, which the build did not check.** The plan names `test_crash_window_repair_owner_alive` as its load-bearing artifact: the row that distinguishes the chosen Site A placement from the rejected one. Moving the Site A block below the owner-liveness gate produces:

```
FAILED ...::TestBlockedAnnotation::test_crash_window_repair_owner_alive
E   Failed: liveness gate reached
1 failed, 1 passed
```

The test is stronger than the plan required — it monkeypatches `_owner_is_gone` to `pytest.fail`, so it asserts the repair never *reaches* the gate rather than merely that it fires. That is a mechanism test, not an outcome test, and it is the single best artifact in this PR.

## Rubric

- [x] **1. Plan vs. implementation match** — pass — *Every Success Criterion is delivered in the diff. Site A placement, the `is True` operator, the biconditional, Race 3 precedence, the untouched `:523`/`:554` escalation sites, and the deliberate no-migration reading all match the plan as written.*
- [x] **2. New code quality** — pass — *Type hints, docstrings on all three new public methods, no debug artifacts, no commented-out code.*
- [x] **3. Test coverage** — pass — *Every new public method and behavioral path has a test. The one gap is a message-selection branch inside an already-tested CLI command, not a behavioral path; the behavior it renders is covered at the model layer. Recorded as tech debt rather than promoted to a rubric failure — stating the judgment explicitly so it can be overruled.*
- [x] **4. Regression risk to existing callers** — pass — *Every non-test caller of `open_expectations()` and `.goal` was grepped. None index `entry["blocked"]`; all reads are `.get(...)` or count-only.*
- [x] **5. Data integrity** — pass — *No Popoto field added, renamed, retyped, or indexed. A key inside an existing `Field(null=True)` JSON payload, where absence is semantically identical to not-blocked. The no-migration reading is argued in the plan's Update System section and reproduced in rule 9.*
- [x] **6. Security** — pass — *No new `eval`/`exec`/`subprocess`/raw SQL, no secrets. The `attempts_exhausted` fence is an authority guard and is enforced in both directions.*
- [x] **7. Documentation accuracy** — pass — *Rule 9 added to `durability-model.md`, the reconciler invariant amended, `tools-reference.md` updated. No stale references.*
- [x] **8. PR body accuracy** — pass — *Every claim I re-ran reproduced. The body now also discloses the missing validator and the two role-prompt files.*
- [x] **9. Disclosed deferrals** — pass — *The `owner_gone` code is an explicit, reasoned out-of-scope exclusion (Settled Decision 1); it ships with its writer. No tracking issue required.*
- [x] **10. Follow-up claims verified** — n/a — *No `filed as follow-up #N` claims in the PR body.*

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — PASS — *all re-derived, not read off the body*
- **2. No-Gos from plan — none violated** — PASS — *`_parse_goal`/`goal_is_corrupt`/corrupt-goal tests untouched; no Popoto field or index added*
- **3. New `except Exception` blocks — each has logger/raise/swallow-ok** — N/A — *no new bare-exception blocks; both annotate sites sit inside the pre-existing per-entry `try`*
- **4. New integration tests — exercise serialization boundary** — PASS — *`test_job_tool.py` drives the CLI functions end to end against real Redis*
- **5. Plan internal consistency — spike findings match task steps** — FAIL — *see the Miscellaneous blocker: the Verification table's expectation syntax does not match the runner the plan is executed by*
- **7. New public APIs — docstrings present** — PASS — *all three new `Job` methods*
- **8. Breaking changes — migration path documented** — N/A — *no breaking change; absent key reads as not-blocked*
- **9. Tests added for new behavior** — PASS — *+16 model, +12 reconciler, +4 tool*
- **10. Tests cover the failure path (not just happy path)** — PASS — *corrupt-goal refusal, unknown/discharged id, empty code, both directions of the cross-guard, `None` from `_escalation_exists`, sub-threshold attempts*
- **11. UI changes (if any) — screenshot captured** — N/A — *no UI files in the diff; the visual-proof gate is a no-op*
- **12. Docs updated for user-facing changes** — PASS

### Blockers
- See Miscellaneous — the one blocker is a plan-document defect, not a code defect, and is listed there because it maps to no rubric item.

### Tech Debt
**File:** `tools/job_tool.py:337-355`
**Code:** `if existing and existing.get("by") == "reconciler":` … `f"reconciler as {existing.get('code')}; unblock it first."`
**Issue:** The plan's Task 2 mandates this CLI refusal rendering ("never a silent success") and Success Criteria repeats it. `tests/unit/test_job_tool.py` contains no `main()`/argv test for `expectation-block` and never asserts the "unblock it first" string. The model-layer `False` return is well covered; the branch that distinguishes "refused by reconciler precedence" from "unknown expectation id" is not. A future edit can collapse the two messages and nothing goes red.
**Severity:** tech_debt
**Fix:** Add a CLI-level test that annotates an expectation `by="reconciler"`, then drives `expectation-block --by lane` through `main()` and asserts both the message and the non-zero exit.

### Nits
**File:** `reflections/expectation_reconciler.py:500-501`
**Code:** `entry.get("blocked") is None`
**Issue:** Unreachable as a guard — the loop already `continue`s at `:488-490` for any entry with a non-`None` `blocked`. Harmless and faithful to the plan's pseudocode, but it reads as if idempotence were enforced locally when the earlier skip is what enforces it.
**Severity:** nit
**Fix:** Drop it, or comment that it restates the `:488` skip.

**File:** `models/job.py:499-500`
**Code:** `raise ValueError(f"blocked code must be one of {sorted(BLOCKED_REASONS)}, got {code!r}")`
**Issue:** 100-character line where the surrounding file wraps nearer 88-95. `ruff format` leaves it, so not a lint failure.
**Severity:** nit
**Fix:** None required — stylistic.

**File:** `reflections/improvement_collect.py:537-538`
**Code:** `entries = job.open_expectations(direction="outbound")`
**Issue:** Outside this diff, flagged as blast radius. This reader records `owner_liveness` evidence for gone-owner outbound expectations without consulting `entry.get("blocked")`. A row annotated `attempts_exhausted` — permanently skipped by `_reconcile_project` thereafter — keeps generating "owner has no live claim" evidence on a separate cadence. Nothing breaks on the new entry shape; this is behavioral drift just outside the plan's blast-radius question.
**Severity:** nit
**Fix:** No action in this PR. Worth a follow-up issue if that signal is meant to reflect only unaddressed stalls.

### Miscellaneous

**File:** `docs/plans/expectation_blocked_state.md` — the `## Verification` table and the baseline-measurements table beneath it
**Code:** `| Crash window closes on the next tick regardless of owner liveness (behavioral) | scripts/pytest-clean.sh ... -k "crash_window_repair" -q | exit 0, **2 tests collected and passed** ... |`
**Issue:** The plan's Verification table cannot be executed by this repo's runner (`agent/verification_parser.py`). `evaluate_expectation` supports exactly six expectation forms — `exit code N`, `exit code != N`, `output > N`, `output contains X`, `output does not contain X`, `match count == 0` — and returns `False` unconditionally for anything else (`agent/verification_parser.py:440`). Running the declared runner over this plan at PR head yields 18 parsed checks and 11 failures, none of which reflect the code:

1. **The two behavioral `-k` rows can never pass.** Their Expected column reads `exit 0, **2 tests collected and passed**`, not `exit code 0`. The regex is `exit code (\d+)`; `exit 0,` does not match, so both fall through to `return False`. These are precisely the two rows the critique round added to replace the dropped call-site greps — the load-bearing behavioral gates — and neither is capable of passing. The underlying tests *do* pass (I ran them: 2 and 3 respectively).
2. **Three rows use `output == N`, which is not a supported form.** `awk '/^MIGRATIONS/,0' … | grep -ci blocked` (expected `output == 0`), `grep -ci "blocked" scripts/update/migrations.py` (expected `output == 1`), and `grep -c "^9\. \*\*" docs/features/durability-model.md` (expected `output == 1`) all fail unconditionally. Each command emits exactly the value the plan predicts — I checked all three by hand — but the runner cannot evaluate the assertion.
3. **The baseline-measurements table is parsed as six additional checks.** Its first header cell is `Command`, which is the runner's own heuristic for "this is a check table" (`_is_check_table`), so its prose Expected column ("RED before task 5, GREEN only after rule 9 names the vocabulary") is evaluated as an expectation and fails. `t.skipped` is empty — the runner never recognized it as a non-check table.

This matters beyond tidiness. A gate that cannot pass is not a gate: any re-review, `/do-patch` cycle, or `/do-merge` check that runs the declared runner over this plan sees a failing Verification table and cannot distinguish that from a real regression. It also silently undoes the critique's round-2 correction — the behavioral rows were introduced specifically so a correct build could not fail the gate, and as written a correct build fails them 100% of the time.

**Severity:** blocker
**Fix:** Edit the plan document only — **no code change, do not touch this branch.** Plans commit directly to `main`, so this is a `main`-side edit under `Refs #2862`: (a) change the two behavioral rows' Expected to `exit code 0` and move the "2 tests collected and passed" prose into the row's Check column or a footnote; (b) replace the three `output == N` expectations with a supported form — `output contains 0` / `output contains 1`, or restructure as `match count == 0` for the zero cases; (c) rename the baseline table's `Command` header (e.g. to `Measured command`) so `_is_check_table` stops claiming it. Separately worth a repo issue: `evaluate_expectation` failing closed on unrecognized syntax is correct, but it is indistinguishable in `format_results` output from a genuine assertion failure — an `UNSUPPORTED_EXPECTATION` result type would make a plan-authoring error legible as one.

- **`owner_gone` cut from `BLOCKED_REASONS`** — explicit out-of-scope exclusion with recorded rationale (Settled Decision 1: every shipped code must have a day-one writer; `owner_gone` has none until the session-health writer exists). No tracking issue required, and the frozen-set test enforces the four-member vocabulary.
- **No dashboard surface for blocked counts** — explicit Rabbit Holes exclusion; `ui/data/sdlc.py` deliberately left reading `len(job.open_expectations())`.
- **Part 2 of #2862 already landed on main** — `docs/plans/expectation_blocked_state.md:19` and the No-Gos row at `:268`. "Part 1" points backward at finished work, not forward at a remainder. `Closes #2862` is correct; this PR should not be held waiting for a Part 2.

### Review Delta
No prior review exists on this PR — this is the first.

### Notes for the merge owner
- `reflections/` and `models/` changes require a service restart after deploy (`./scripts/valor-service.sh restart`) for the reconciler change to take effect.
- The PR touches two runtime prompt surfaces, `.claude/commands/roles/prime-pm-role.md` and `prime-pm-codex-role.md`. Verified: one added bullet each, under an existing heading, no heading text altered — so the byte-for-byte heading parse into worker prompts is unaffected.
- `docs/features/README.md` is deliberately untouched. No index row is needed: this PR amends three existing feature docs and creates none.

valorengels:
## Review Blocker Resolved: Verification table now executes

The blocker from [the review above](https://github.com/tomcounsell/ai/pull/3407#issuecomment-5726513780) was a plan-document defect, not a code defect. It has been fixed on `main` at `bbe5dc7a1` — no change to this branch, which remains at `30b49fe55`.

**Executed the corrected table through the real runner**, rather than re-deriving the properties by hand. Hand-measurement is what made the defect invisible in the first place: my manual checks and the runner disagreed, and only the runner gates.

Plan content from `bbe5dc7a1`, checks executed in the PR worktree at head `30b49fe55`:

```
checks: 14  malformed: 0  skipped: 1
...
**All checks passed.**
PASSED: 14 / 14
EXIT=0
```

All 14 rows PASS. The baseline table is now correctly classified as a non-check table and skipped with its reason stated (`not a check table: no column of the first three is named Command`), its 6 rows unexecuted and excluded from the verdict.

**The two count predictions in the corrected plan were predictions, and both are now measurements.** `output contains 2 passed` on the crash-window selector and `output contains 3 passed` on the annotation selector both hold against this head — the selectors resolve to exactly the counts the plan asserts. No test was adjusted to match the plan.

The split-row idiom is worth naming for reuse: `exit code 0` carries non-vacuity (the `ZERO TESTS EXECUTED` guard turns a mistyped `-k` into exit 1 rather than a silent empty pass) while `output contains N passed` carries the count, because no single supported form expresses "exactly N collected". Likewise `<count command> | grep -vc "^N$"` with `match count == 0` gives an exact-equality assertion the parser has no `==` form for.

### Remaining state

- **Blockers: 0** — this one is cleared, and it was the only one.
- **Tech debt: 1** — unchanged and unaddressed: `tools/job_tool.py:337-355`, the plan-mandated Race 3 CLI refusal rendering has no `main()`/argv test.
- **Nits: 3** — unchanged.

Per Hard Rule 4, approval is reserved for zero-finding reviews, so the recorded verdict stays **CHANGES REQUESTED** on the strength of the tech-debt item alone. This comment records the blocker's resolution; it does not itself change the verdict.

valorengels:
## Review: Changes Requested (round 2)

**Mode:** independent roster (2 judges) — `code-quality` and `risk` re-dispatched as separate subagents at head `52906537f`, each blind to the other, both given the round-1 comments and told to verify every claimed disposition rather than trust it. Quorum 2 of 2, no shortfall.

> **Disclosure carried forward verbatim: no independent validator ran during BUILD.** Both judges were told again this round and instructed to re-derive every claim from code. `-n 0` was required of both.

### Round-1 findings — dispositions, each verified

- **Blocker (the plan's `## Verification` table could not execute)** — **RESOLVED.** Fixed on `main` at `bbe5dc7a1`, plan text only (1 file, +12/-6, zero code). I executed the corrected table through the real runner against this head: **14 checks, 0 malformed, 1 skipped table, 14/14 PASS, exit 0.** The baseline table is now correctly classified out with its reason stated. The two count predictions in the corrected plan (`output contains 2 passed`, `3 passed`) hold as measurements — no test was adjusted to match the plan.
- **Tech debt (`tools/job_tool.py:337-355` CLI refusal untested)** — **RESOLVED** by `52906537f`, which adds `TestExpectationBlockCliRefusal` to `tests/unit/test_job_tool.py`. Test-file-only: `git diff 30b49fe55..52906537f --name-only` is exactly that one file, +76/-0.
- **Nits 1-3** — all three still open; see below and the dispositions section.

### The new tests, and why they count

They drive the real entrypoint — `main()` via monkeypatched `sys.argv`, `pytest.raises(SystemExit)`, stderr through `capsys` — not the module-level helper, which would have exercised none of the branch. Setup is ORM-only (`Job.query.filter(...).block_expectation(eid, code="attempts_exhausted", by="reconciler")`), matching the file's existing `scratch_session` fixture.

They assert **branch discrimination in both directions**, which is the point: the two refusal messages share an exit code and both open with `Error:` on stderr, so a test asserting only a non-zero exit passes identically against code reporting the wrong reason.

**Three independent mutations, three different levers, all RED:**

| Mutation | Lever | Result |
|---|---|---|
| Builder | swapped the two messages' text | both new tests FAIL |
| Reviewer (me) | inverted `existing.get("by") == "reconciler"` | test 1 FAILS |
| Judge `code-quality` | `==` → `!=` at `tools/job_tool.py:344` | test 1 FAILS, restored clean |

Test 2 is insensitive to the condition-inversion mutations (with `existing` falsy the branch is unreachable either way) but dies under the message swap. The pair is load-bearing together; neither alone would be.

### Blockers
- None

### Tech Debt
**File:** `tools/job_tool.py:189-209` and `:336-355`
**Code:** `def block_expectation(...) -> bool:` … then, on `False`, `job = _own_room_job(session_id, args.job_id)` and a re-scan of `job.all_expectations()` to pick the message.
**Issue:** The refusal **reason is discarded and then reconstructed by a second, independent read.** `Job.block_expectation()` returns a bare `False` for *both* Race 3 precedence and "no such open expectation", so `main()` re-fetches the job and re-derives which case it was. The two reads can disagree, and when they do the operator is told something actively false: if the second read returns an entry whose `blocked` is absent, `existing` is falsy and the CLI prints `no open expectation {id!r}` about an expectation that is open and blocked. That is precisely the outcome plan line 262 exists to forbid — the refusal must name the reconciler and the code, never render as something else.

The `risk` judge hit this shape live: two consecutive failures of `test_reconciler_annotated_entry_names_the_reconciler_and_code` in the combined three-file command, landing in the `no open expectation` branch, followed by five clean retries. **I could not reproduce it** — 6 consecutive combined runs (144 passed each) and 15 consecutive `-k CliRefusal` runs, 21 clean in a row across two shapes. The judge's runs also showed collected counts varying between 114 and 144, which is this repo's documented shared test-DB claim-pool contention signature, not a logic race. So I am not calling the flake a defect in this PR's code, and it is not a blocker.

What survives the irreproducibility is the design observation, which needs no flake to be true: a bool that drops its reason, re-derived by a second read, is a latent false-message surface in exactly the code path whose entire purpose is telling the operator the truth about a refusal.
**Severity:** tech_debt
**Fix:** Have the refusal carry its own reason rather than reconstructing it — e.g. return an enum/`None`-or-reason from `Job.block_expectation()`, or have the module-level wrapper return the already-read entry alongside the bool, so `main()` renders from the same read that made the decision. One read, one truth.

### Nits
**File:** `reflections/expectation_reconciler.py:500-501`
**Code:** `entry.get("blocked") is None`
**Issue:** Unreachable as a guard — the loop already `continue`s at `:488-490` for any entry with a non-`None` `blocked`. Confirmed independently by the `code-quality` judge reading `:480-505`. Matches the plan's pseudocode verbatim.
**Severity:** nit
**Fix:** Drop it, or comment that it restates the `:488` skip.

**File:** `models/job.py:499-500`
**Code:** `raise ValueError(f"blocked code must be one of {sorted(BLOCKED_REASONS)}, got {code!r}")`
**Issue:** 100 characters, measured, in a file otherwise wrapping 88-95. `ruff format --check` passes, so not a lint failure.
**Severity:** nit
**Fix:** None required — stylistic.

**File:** `reflections/improvement_collect.py:537-563`
**Code:** `entries = job.open_expectations(direction="outbound")`, then a per-entry loop recording `ImprovementEvidence` with source `owner_liveness` whenever `_owner_is_gone(owner)`, never consulting `entry.get("blocked")`.
**Issue:** Outside this diff. A row annotated `attempts_exhausted` — permanently skipped by `_reconcile_project` thereafter — keeps generating "owner has no live claim" evidence on a separate cadence.
**Severity:** nit
**Fix:** No action in this PR. Follow-up issue if that signal should reflect only unaddressed stalls.

### Nit dispositions
All three are judged **acceptable to ship as-is** by the reviewing judge and by me. Nit 1 is provably harmless (the earlier skip guarantees its precondition) and is verbatim plan pseudocode; nit 2 is cosmetic with the formatter silent; nit 3 is correctly scoped out of the diff and blocking on out-of-diff blast radius would be scope creep against the plan's own Rabbit Holes. They are recorded rather than patched.

The verdict is **Changes Requested** because this pipeline reserves approval for zero-finding reviews and findings survive — not because any single item should gate the merge. The gating decision belongs to the merge owner.

### Verification at this head
- Plan `## Verification` table through the real runner: **14/14 PASS**, 0 malformed, 1 table correctly skipped.
- `test_job_model.py` + `test_job_tool.py` + `test_reflections_expectation_reconciler.py`, `-n 0`: **144 passed** (was 142; +2 new), reproduced 6 consecutive times.
- `tests/integration/test_job_routing.py`, `-n 0`: **3 passed.**
- `ruff check` / `ruff format --check` on the changed test file: clean.
- Branch commits carry no `Co-Authored-By` and no "Generated with Claude Code" footer — verified across all 6 commits via `git log main..HEAD --format=%B`.

### Review Delta
Round 1 at `30b49fe55`: 1 blocker, 1 tech debt, 3 nits. Round 2 at `52906537f`: **0 blockers, 1 tech debt, 3 nits.** The blocker and the round-1 tech-debt item are both closed and verified closed. The new tech-debt item is newly surfaced this round, not carried over.

valorengels:
## Review: Approved

Head `52906537f`. Round 3, resolving round 2's surviving findings by explicit deferral rather than by patch. **Mode:** independent roster (2 judges), carried from round 2 — both judges returned Approved on the code at this head with 0 blockers.

> **Disclosure, carried verbatim through every round: no independent validator ran during BUILD.** The plan's roster (`docs/plans/expectation_blocked_state.md:313-328`) names a `blocked-validator` distinct from the two builders; it never ran. Every claim in this review was re-derived from code or re-measured, not read off the PR body.

### What changed since round 2

No code. Round 2's four surviving findings are ruled **acknowledged deferrals** by the merge owner, on the record and with reasons, and two follow-up issues are filed. Approval here is a zero-open-finding review, not a waiver of unresolved work.

**1. `models/job.py:462` — `block_expectation()` returns a bare bool for two distinct refusal reasons. → [#3414](https://github.com/tomcounsell/ai/issues/3414)**

Round 2 recorded this as tech debt scoped to `tools/job_tool.py`. **Both the classification and the scoping were wrong, and the correction matters more than the finding.**

- It is not inherited debt. It is a design residue of *this PR's own* return contract: the wrapper docstring at `tools/job_tool.py:189-202` says the function returns `False` "so the caller can render the explicit refusal message," which is exactly what forces `main()` to go re-read the job and work out why.
- It is not contained in `job_tool`. `models/job.py:462` declares `-> bool` and returns `False` from two sites for two reasons — the Race 3 precedence refusal inside the entry-match loop, and the unknown/already-discharged fall-through. The discriminator is destroyed in the model; `job_tool` is downstream of the loss, not its cause. Verified by reading `models/job.py:462-520`.

**Why it does not gate this PR, stated precisely.** Plan line 262 forbids rendering the Race 3 refusal **as a silent success**. The double-read cannot produce a silent success: both branches exit non-zero and both refuse the write. The failure mode is a refusal carrying the wrong *explanation* — the operator is correctly stopped and misleadingly told why. Real, but not the failure the mandate exists to prevent. That is the whole of the reason. Round count and downstream scheduling were explicitly rejected as grounds.

#3414 carries the model-origin framing and an explicit instruction not to fix it in `job_tool`, because a third read there would make the CLI's message correct and leave the contract defect one layer down with a passing test proving it fixed.

**2. `reflections/expectation_reconciler.py:500-501` — `entry.get("blocked") is None` is unreachable as a guard.** The loop already `continue`s at `:488-490` for any entry with a non-`None` `blocked`, so the precondition is guaranteed before the clause is reached. Verbatim plan pseudocode. Provably cannot cause an incorrect skip. Accepted as-is.

**3. `models/job.py:499-500` — 100-character line** in a file otherwise wrapping 88-95. Measured. `ruff format --check` passes, so not a lint failure. Cosmetic. Accepted as-is.

**4. `reflections/improvement_collect.py:537-563` — downstream `owner_liveness` evidence ignores the new `blocked` key.** Outside this diff; a row annotated `attempts_exhausted` and thereafter skipped by `_reconcile_project` keeps generating "owner has no live claim" evidence on its own cadence. Correctly scoped out — gating on out-of-diff blast radius would be scope creep against the plan's own Rabbit Holes. Accepted as-is, follow-up-worthy.

### Separately filed, not a finding against this PR

**[#3415](https://github.com/tomcounsell/ai/issues/3415)** — `pytest-clean.sh` guards *zero* tests executed but not *partial collection*. A judge's runs of one command showed collected counts swinging 114–144 while the identical command returned a stable 144 elsewhere; a pool-starved run that silently drops 20% of its tests exits 0 and reads green. That is the same false-green class as #3195 one step up the range, and it is a test-harness concern independent of #2862. It is also the reason round 2's intermittent test failure was **not** charged to this PR: the varying collection counts are the shared claim-pool signature, not a logic race, and 21 consecutive clean runs across two shapes here failed to reproduce it.

### Blockers
- None

### Tech Debt
- None open. The round-1 item (`tools/job_tool.py:337-355` untested) is closed by `52906537f`; round 2's item is reclassified and deferred to #3414 above.

### Nits
- None open. All three are acknowledged deferrals above.

### Verification at this head
- Plan `## Verification` table through the repo's own runner (`agent/verification_parser.py`): **14 checks, 0 malformed, 1 non-check table correctly skipped, 14/14 PASS, exit 0.**
- `test_job_model.py` + `test_job_tool.py` + `test_reflections_expectation_reconciler.py`, `-n 0`: **144 passed**, reproduced 6 consecutive times.
- `tests/integration/test_job_routing.py`, `-n 0`: **3 passed.**
- `-k CliRefusal`, `-n 0`: **2 passed**, reproduced 15 consecutive times.
- `ruff check` / `ruff format --check`: clean.
- No `Co-Authored-By` and no "Generated with Claude Code" footer across all 6 commits (`git log main..HEAD --format=%B`).

### The load-bearing artifact
`TestExpectationBlockCliRefusal` asserts branch discrimination in both directions through the real `main()` entrypoint — reconciler-annotated entry yields stderr containing `blocked by the reconciler` and `attempts_exhausted` and **not** `no open expectation`; a genuinely absent id yields the inverse; both non-zero exit. **Three independent parties mutated three different levers and all went RED**: the builder swapped the message text (both tests fail), the reviewer inverted the `== "reconciler"` condition (test 1 fails), the `code-quality` judge independently chose `==` → `!=` at `tools/job_tool.py:344` (test 1 fails). Source restored exactly each time, confirmed by `git diff --name-only`.

Stated plainly because it is the kind of thing that normally goes unsaid: test 2 is insensitive to both condition-inversion mutations — with `existing` falsy that branch is unreachable either way — and survives only under the message swap. The pair is load-bearing together and neither is load-bearing alone.

### Review Delta
Round 1 at `30b49fe55`: 1 blocker, 1 tech debt, 3 nits. Round 2 at `52906537f`: 0 blockers, 1 tech debt, 3 nits. Round 3 at `52906537f`: **0 blockers, 0 tech debt, 0 nits open**, 4 acknowledged deferrals, 2 follow-up issues filed.

**Approved. Merge sequencing belongs to the merge owner; this review does not merge and does not authorize a merge.**

Issue #3419, opened 2026-09-18
Collapse per-file: fold N+ same-file nightly failures into one umbrella issue by root cause


valorengels:
> **Context**: this is a scope split from #3418. #3418 covers moving nightly-triage issue filing from an LLM session into the deterministic detector (`scripts/nightly_regression_tests.py`). This issue covers a separate, independently-scoped fix: collapsing many failing test nodes that share one root cause into a single umbrella issue instead of filing one issue per node.

## Problem

On the 2026-09-17 nightly run, 22 of the 44 confirmed-failing nodes live in `tests/unit/test_improvement_eval_*.py` and share a single root cause: an arm corpus digest `InfraFailure` originating from commit `cc1938981`, already tracked on #3365. The nightly pipeline has a "body-cascade" collapsing mechanism, but it only fires when nodes share an identical normalized first error line. These 22 nodes fail at different leaf assertions downstream of the same root infra failure, so each produces a distinct first error line and becomes its own separate issue instead of folding into #3365.

The result: one infrastructure defect generates roughly 20 GitHub issues instead of one, burying the actual regressions among noise.

## Definitions

| Term | Definition |
|------|------------|
| Body-cascade collapsing | The existing mechanism in the nightly pipeline that folds multiple failing nodes into one issue when they share an identical normalized first error line. |
| Root cause vs. leaf assertion | The root cause is the underlying defect (e.g. a shared infra failure); the leaf assertion is the specific `assert` that fails downstream of it, which differs per test even when the root cause is the same. |

## Solution sketch

Add a per-file (or per-top-level-package, e.g. `test_improvement_eval_*`) collapsing rule: when N or more nodes within one test file fail in the same nightly run, file a single umbrella issue for that file listing the individual nodes in a collapsed section, rather than relying solely on identical-first-error-line matching. Default N=3, overridable via environment variable. This is independent of and should land after (or alongside, but as a separate PR from) #3418's filing-ownership move, since #3418 changes who creates the issues (detector vs. LLM session) while this issue changes the collapsing algorithm that decides how many issues get created.

## Acceptance criteria

- A nightly run with 22 failing nodes sharing one root cause files 1 umbrella issue (or however many the per-file/per-package threshold produces), not ~20 individual issues.
- The per-file/per-package threshold N is environment-overridable with a documented default.
- Existing body-cascade collapsing (identical first error line) continues to work unchanged; the new rule is additive.

## Downstream context

[...]

Issue #3405, opened 2026-09-17
Nightly regression: tests/unit/test_session_lifecycle_consolidation.py::TestImportSafety::test_lazy_imports_not_triggered_at_module_level


valorengels:
Failing node:

```
tests/unit/test_session_lifecycle_consolidation.py::TestImportSafety::test_lazy_imports_not_triggered_at_module_level
```

## Failure

```
tests/unit/test_session_lifecycle_consolidation.py:481: in test_lazy_imports_not_triggered_at_module_level
    assert leaked == [], (
E   AssertionError: heavy dependencies imported at module level: ['agent.agent_session_queue'].
E   They must stay inside the functions that need them.
E   assert ['agent.agent_session_queue'] == []
E     Left contains one more item: 'agent.agent_session_queue'
```

The test imports `models.session_lifecycle` in a clean subprocess and asserts that neither
`tools.session_tags` nor `agent.agent_session_queue` lands in `sys.modules`.
`agent.agent_session_queue` now does.

## Root cause

`models/session_lifecycle.py` acquired a **module-level** import of a sibling module in the
`agent` package:

```python
# models/session_lifecycle.py:29
from agent.lock_policy import record_lock_degradation
```

`agent/lock_policy.py` is itself dependency-light (only `logging` and `typing`), so it looks
harmless -- but importing any `agent.*` submodule executes `agent/__init__.py`, whose very first
statement is an eager re-export:

```python
# agent/__init__.py
from .agent_session_queue import (
    PRIORITY_RANK,
    AgentSession,
    enqueue_agent_session,
    register_callbacks,
)
```

So `agent.agent_session_queue` is pulled in transitively at `models.session_lifecycle` import
time, defeating the lazy-import contract the module documents for itself (its own docstring,
line 17: "Uses lazy imports for heavy dependencies (tools.session_tags,
agent.agent_session_queue)"). The intended lazy use site is still function-local at
`models/session_lifecycle.py:618` (`from agent.agent_session_queue import
checkpoint_branch_state`) -- that one is fine; the leak is the package `__init__` side effect.

Introduced by **`a9822d719`** ("ETL-grade pipeline hardening: side-effect jobs, dead letters,
wire schemas, lock policy, lineage (#3229)", 2026-09-07), which added the module-level
`from agent.lock_policy import record_lock_degradation` line. `git log -S` confirms that commit
is the sole origin of the import.

## Suggested next steps

[...]

Issue #3397, opened 2026-09-17
Nightly regression: tests/unit/test_session_lifecycle_consolidation.py::TestImportSafety::test_lazy_imports_not_triggered_at_module_level


valorengels:
## Failure

Fails on `main` (worktree HEAD `87eabe1ac`, lane `nightly-triage-4b33f93e`).

```
tests/unit/test_session_lifecycle_consolidation.py:481: AssertionError
E   AssertionError: heavy dependencies imported at module level: ['agent.agent_session_queue'].
E   They must stay inside the functions that need them.
E   assert ['agent.agent_session_queue'] == []
```

Reproduced in isolation:
```
scripts/pytest-clean.sh "tests/unit/test_session_lifecycle_consolidation.py::TestImportSafety::test_lazy_imports_not_triggered_at_module_level" -q --no-header -p no:randomly
# -> 1 failed in 20.72s
```

`tools.session_tags` is still clean; only `agent.agent_session_queue` leaks.

## Root cause: `agent/__init__.py` re-exports it eagerly, and two module-level paths now touch `agent.*`

The enabler is `agent/__init__.py:3`, which pulls the "heavy" module in as a
package-level re-export:

```python
from .agent_session_queue import (
    PRIORITY_RANK, AgentSession, enqueue_agent_session, register_callbacks,
)
```

Consequence: importing **any** `agent.*` submodule executes `agent/__init__.py`
and therefore imports `agent.agent_session_queue`. The lazy-import contract in
`models/session_lifecycle.py`'s docstring ("Uses lazy imports for heavy
dependencies (tools.session_tags, agent.agent_session_queue)") cannot hold while
that re-export exists.

Two independent live routes, both verified empirically in a fresh interpreter:

**Route A — `models/session_lifecycle.py:29`**
```python
from agent.lock_policy import record_lock_degradation
```
```
$ .venv/bin/python -c "import sys, agent.lock_policy; print('agent.agent_session_queue' in sys.modules)"
True
```
Introduced by **`a9822d719`** "ETL-grade pipeline hardening: side-effect jobs,
dead letters, wire schemas, lock policy, lineage (#3229)" (`git log -S`). This
is the proximate regression: a module-scope `agent.*` import added to the very
file whose contract forbids it.

**Route B — the `models` package `__init__`**
```
models/__init__.py:59   from models.memory import Memory
models/memory.py:24     apply_defaults()            # called at module scope
[...]

PR #3517, opened 2026-09-19
Move nightly issue filing into the detector and collapse duplicates by root cause

### LINKED ISSUE #3418: Nightly triage session files every new node 2-3 times per night; move issue filing into the detector and collapse by root cause
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/nightly-filing-into-detector-collapse-by-root-cause.md

> **Context**: this repo's nightly regression runner (`scripts/nightly_regression_tests.py`) detects newly-failing test nodes and dispatches a triage session (an LLM-driven Claude Code session) to investigate and file GitHub issues for each one. This report is about that filing step, not about the underlying test failures themselves.

## Problem

Tom is triaging far more nightly issues than there are actual defects. Two compounding causes, both verified on this machine on 2026-09-18.

### 1. The triage session duplicates its own filings

On the 2026-09-17 run, `scripts/nightly_regression_tests.py` logged "8 issue(s) filed" and dispatched exactly one triage session (slug `nightly-triage-4b33f93e`, session `0_1789678340046`, one harness run from 20:52 to 21:10 UTC). GitHub nonetheless received 24 issues for those same 8 nodes, in three waves roughly ten minutes apart:

- Wave 1: #3382-#3389 at 20:54
- Wave 2: #3390-#3397 at 21:04
- Wave 3: #3398-#3405 at 21:08

Example: one single test node (`tests/unit/test_improvement_eval_runner_guards.py::TestBlinding::test_identity_leak_sets_blinded_false`) was filed three times, as #3383, #3391, and #3399.

The session's own final report claims "all 8 nodes had no existing issue in any state" and lists only #3398-#3405 as the filed issues — it never surfaces #3382-#3397 at all. The report also notes that its prescribed `gh issue list --limit 200` lookup came back truncated, showing only #3128-#3380. That means the lookup literally could not see the issues the session's own earlier waves had just created a few minutes prior, so each wave re-filed nodes the previous wave had already covered.

The 2026-09-16 run shows the same pattern at roughly 2x duplication (e.g. #3355/#3365, #3359/#3369, #3360/#3370, #3361/#3371).

This is the same failure class as #3170 (the #2960-#2999 duplicate-filing wave). The ledger and lookup-instruction defenses added after #3170 did not hold here. Notably, the detector itself does not have this problem: on this same run it correctly deduplicated recurrences by commenting on #3375-#3380 instead of re-filing them. The duplication is entirely inside the LLM-driven triage session that does the filing, not in the detector that decides what needs filing.

### 2. One root cause fans out into roughly 20 separate issues

22 of the 44 confirmed-failing nodes on 2026-09-17 live in `tests/unit/test_improvement_eval_*.py` and share a single root cause: an arm corpus digest `InfraFailure` originating from commit `cc1938981`, already tracked on #3365. The nightly system has a "body-cascade" collapsing mechanism, but it only triggers when nodes share an identical normalized first error line. These 22 nodes fail at different leaf assertions downstream of the same root infra failure, so each one produces its own distinct first error line and becomes its own separate issue instead of being folded into #3365.

### 3. Nondeterminism (secondary contributor)

Some of the noise also comes from non-regressions being filed as if they were regressions: quota exhaustion filed as a regression (#3347), LLM-judge test flakiness, and test-database starvation under load. This is a smaller contributor than causes 1 and 2 — the nightly run already does a serial re-confirm pass that strips out xdist-only flakes (it excluded 3 such artifacts on the 09-17 run), so most of the 44 nodes are genuine deterministic failures on `main`, not flakes.

## Definitions

| Term | Definition |
|------|------------|
| Nightly triage session | An LLM-driven Claude Code session, dispatched once per nightly run by `scripts/nightly_regression_tests.py`, whose job is to investigate newly-failing test nodes and file/update GitHub issues for them. |
| Detector | The deterministic (non-LLM) part of the nightly pipeline that identifies which test nodes newly failed and decides whether a given node already has an open issue, distinct from the triage session that does the filing. |
| Body-cascade collapsing | The existing mechanism that folds multiple failing nodes into one issue when they share an identical normalized first error line. |
| Serial re-confirm | A follow-up pass that re-runs failing nodes serially (outside xdist) to filter out parallel-execution-only flakes before they're treated as regressions. |

## Proposed cull

- **Move filing out of the LLM session and into the detector.** The detector already computes titles, dedup dispositions, and posts recurrence comments via `gh issue create`/`gh issue comment` — it should file new issues itself rather than delegating that to a freshly-dispatched LLM session that has no memory of what earlier waves in the same run already did. If an investigation session is still wanted for deeper root-causing, dispatch it after filing, with the issue numbers already assigned, and instruct it to comment only, never to create.
- **Collapse per-file, not just per-error-line.** When N or more nodes within one test file fail in the same night, file a single umbrella issue per file (or per top-level package, e.g. `test_improvement_eval_*`), listing the individual nodes in a collapsed section. Default N=3, overridable via environment variable.
- **Cap issue creation by querying GitHub, not by self-reported tally.** Lower the per-run issue budget default, and enforce it as a hard cap measured by querying GitHub for issues actually created during the run — not by trusting the filing session's own count (which is exactly the number that was wrong on 09-17).
- **Keep or link the #3347 quota pre-check** so quota-exhaustion failures don't get filed as regressions in the first place.

## Acceptance criteria

- A nightly run with K genuinely new root causes creates at most K issues total.
- A re-run against the same failures the same night creates zero new issues and instead comments on the existing ones.
- Replaying nightly-triage-4b33f93e's original inputs against the new filing path produces exactly one issue per node, with no second or third wave.
- The duplicate copies from this incident (#3382-#3397, plus the 09-16 pairs such as #3355/#3365, #3359/#3369, #3360/#3370, #3361/#3371) are closed as duplicates as part of this work.

## Recon Summary

Performed 2026-09-18 against `main` @ `eae095035` (latest commit touching `scripts/nightly_regression_tests.py`; no commits have touched it since this issue was filed).

**Confirmed:**
- Filing is done by the LLM session, commenting is done by Python. `dispatch_findings()` (`scripts/nightly_regression_tests.py:2696`) posts recurrence comments itself via `comment_on_issue()` (`:2280`, `gh issue comment --body-file -`), but for anything new it calls `maybe_dispatch_triage_session()` (`:2518`), which shells out to `tools.valor_session create --role eng` and lets the agent run `gh issue create`. There is no `gh issue create` call anywhere in the script.
- The per-run budget is a self-reported tally, never verified. `DispatchOutcome.issues_filed` is incremented by `len(single_nodes)` the moment the dispatch subprocess returns a session id (`:2966`), and by 1 per cascade umbrella (`:2860`) — before any issue exists. The "N issue(s) filed" line at `:3265` reports that number. A session that files 24 issues for 8 nodes is reported as 8.
- The 3x duplication is real and the waves are byte-identical titles. GitHub shows the same 8 titles created three times: #3382-#3389 (20:54), #3390-#3397 (21:04-21:05), #3398-#3405 (21:08-21:09). The 09-16 2x pairs (#3355/#3365, #3356/#3366, #3357/#3367, #3358/#3368, #3359/#3369, #3360/#3370, #3361/#3371, #3362/#3372, #3363/#3373, #3364/#3374) are also confirmed open.
- The session ledger did not hold. `data/nightly-triage-ledger/nightly-triage-4b33f93e.json` has 8 entries and a `filed` array containing exactly #3398-#3405 — wave 3 only. Waves 1 and 2 left no trace in it, so the agent rewrote the file rather than appending, and `write_triage_ledger`'s "leave a non-empty `filed` alone" guard (`:2450`) never applied because the file is only written once at dispatch.
- Body-cascade collapsing is keyed on exact string equality of the normalized first `E ` line only. `group_body_failure_cascades()` (`:1838`) buckets by `body_failure_signature()` and has no notion of test file or package. Threshold `BODY_CASCADE_MIN_GROUP_SIZE_DEFAULT = 5` (`:305`), env-overridable via `NIGHTLY_BODY_CASCADE_MIN_GROUP_SIZE`. 22 nodes downstream of one arm-corpus-digest `InfraFailure` each produce a distinct leaf assertion line, so none of them group.
- The detector's own dedup works. `open_issues()` (`:2065`) reads `gh issue list --state open --limit 1000` and `closed_issue_dispositions()` (`:2128`) reads `--limit 4000`; on the same 09-17 run they correctly routed #3375-#3380 to comments instead of re-files.
- Existing test coverage is substantial and will constrain the change: `tests/unit/test_nightly_regression_tests.py` (3355 lines) has `TestDispatchFindings`, `TestDispositionHandoff`, `TestMaybeDispatchTriage`, `TestPreFileDedup`, `TestBodyFailureGrouping`, `TestWriteTriageLedger`, `TestBuildTriagePrompt`, `TestPromptsNeverNameTheSearchIndex`, `TestMainDispatchPersistence`.

**Revised:**
- The issue attributes the duplication to the `--limit 200` truncation in the agent's lookup. That cap is real (`ISSUE_LOOKUP_INSTRUCTION`, `:348-359`), but it is not sufficient to explain the waves: `gh issue list` returns newest-created first, so a lookup re-run at 21:04 would have seen #3382-#3389. The binding defect is the instruction's own wording — "run exactly this ONCE for the whole list below" — combined with a replayed/continued turn: one lookup taken at 20:52 stays authoritative for the rest of the session, which matches the session report's claim that it saw only up to #3380. The 200 cap makes it worse, not causal. Both disappear once filing leaves the LLM.
- "22 of 44 nodes share one root cause" is the 09-17 shape, but the three duplicate waves that reached GitHub were only 8 distinct nodes — the other nodes were already suppressed by the detector's dedup. Per-file collapsing and the duplicate-filing fix therefore address two disjoint slices of the noise and are correctly sequenced as separate tasks.

**Pre-requisites:**
- None blocking. #3347 (quota exhaustion filed as regression) is still OPEN and is a link/keep target, not a dependency. #3243 (stamp hostname/session_id into filed bodies) is still OPEN and touches the same filing path — coordinate, since moving filing into Python is where that stamp would naturally live.
- #3170, #3131, #3134, #3075 are all CLOSED as completed; their mechanisms (per-node dispatch dedup, cascade collapsing, closed-state dedup, disposition handoff, session ledger) are the four prior passes this change supersedes rather than extends.

**Dropped:**
- Nothing dropped from the issue's proposed cull. Cause 3 (nondeterminism) stays out of scope for this plan beyond linking #3347, as the issue itself states.

## Downstream context

This issue is ready for `/do-plan` to turn into an implementation plan. The plan should scope the detector-side filing change, the per-file collapsing threshold, and the GitHub-query-based budget cap as separate, sequenced tasks, and should include a cleanup task for closing the enumerated duplicate issues.

Related: #3170, #3347, #3243, #3131, #3134, #3075.

valorengels:
Closes #3418

## Problem

The nightly regression detector delegated issue *filing* to an LLM triage session. A session turn that replayed re-read its prompt from the top and re-ran the whole filing loop as if for the first time, so one night's 8 findings became 24 issues (#3382-#3405). Every defense against this was itself a prompt input — the `ISSUE_LOOKUP_INSTRUCTION` "run this ONCE" constant, the pre-resolved `NodeDisposition` handoff, and the seeded on-disk ledger — and a replay obeys an instruction it has already obeyed.

## Change

Filing moves into the detector, where it is deterministic. `create_issue()` is now the only path from a nightly finding to a new issue: one Python loop, under the run lock, with the dedup check and the create adjacent.

- **`create_issue()`** — list-form argv, body on stdin via `--body-file -`, number parsed strictly from the issue URL. Never retries: GitHub offers no `Idempotency-Key` on issue creation, so a retried create is a second issue. Any failure returns `None`.
- **Filing fingerprints** — a sha256 over a stable identity (node id, cascade signature, or `seed:{head_commit}`), embedded in every created body as an HTML comment. Never over the rendered title, which a human can edit. Two findings in one run resolving to the same fingerprint produce a report, not a second issue, and not a close.
- **Pre-create refresh** — the open-issue read is refreshed immediately before each create, bounding the check-then-act window to one `gh` round-trip. A refresh hit **comments** on the existing issue and records the node; it never silently skips, so a real finding never ends the night with neither an issue nor a comment.
- **`issues_filed` is derived** — `len(filed_issues)`, over numbers GitHub actually returned. A counter incremented before an issue exists is how a run that created 24 issues was logged as 8.
- **`NIGHTLY_AUTO_FILE=false`** — kill switch, not a second code path. Comments and logs every issue it would have filed so an operator can file by hand; read at call time, not import time.
- **Deleted outright** — `write_triage_ledger()`, `NodeDisposition`, the three filing prompts, and the ledger handoff. The session is still dispatched, to investigate and comment on numbers that already exist.

The dedup reads deliberately fail **open** (an unreadable list files a possible duplicate rather than producing a silent night) while an unconfirmed create is never counted, never spends budget, and never recorded (so the next run retries it). That asymmetry is intentional and is documented so it does not get "fixed" into consistency.

## Tests

`scripts/pytest-clean.sh tests/unit/test_nightly_regression_tests.py` → **247 passed**, exit 0.

Four hand-rolled `gh` harnesses collapse into one `FakeGitHub` fixture with an autouse guard that fails any test reaching the live tracker — an earlier informational run created 94 real issues (#3423-#3516). New coverage: `TestCreateIssue`, `TestFilingIdempotence` (the #3382-#3405 regression: a second `dispatch_findings()` pass creates nothing and comments once per node), `TestPreCreateRefreshComments`, `TestFingerprintCollisionIsLogged`, `TestIssueBudgetSpendsOnlyConfirmedCreates`, `TestKillSwitchSuppressesCreates` (sets the env var *after* import, so a knob captured at import time fails it).

## Verification

Plan: `docs/plans/nightly-filing-into-detector-collapse-by-root-cause.md` (7 critique rounds; round 7 returned READY TO BUILD with no concerns). 18 of 19 merge-gate Verification rows pass exactly as written.

**Row 17 (`#3419 collapsing logic untouched`) needs a reviewer's eye.** Its command returns `1`, not the expected `0`. The single match is git's hunk-header function annotation (`@@ ... @@ def group_body_failure_cascades(`), not a changed line: `git diff origin/main -- scripts/nightly_regression_tests.py | grep -E "^[+-]" | grep -E "group_body_failure_cascades|BODY_CASCADE_MIN_GROUP_SIZE"` returns nothing. The only edit in that region is docstring prose. The anti-criterion holds; the plan's command does not distinguish context annotations from real changes. Flagged rather than silently passed.

## Not in this PR

Closing the ~26 historical duplicates (#3382-#3405 and the 09-16 pairs) is a **post-merge operator step**, deliberately removed from the build pipeline so review-before-execute governs the one irreversible action in scope. The runbook, with the enumerated numbers and a `gh issue view` verification loop, now lives in `docs/features/nightly-triage-dispatch.md` so it outlives the plan document.


valorengels:

Lens: implementation correctness, dead-code removal, test substance, doc consistency.

Verified: `create_issue()` uses list-form argv with the body on stdin (no `shell=True`); the issue number is parsed strictly via `_ISSUE_URL_NUMBER_RE` and every failure path returns `None`; `DispatchOutcome.issues_filed` is a `@property` over `len(self.filed_issues)` with no independent counter surviving; fingerprints are sha256 over stable identity (node id / `cascade_state_key()` / `seed:{head_commit}`), never the rendered title; the pre-create refresh in `_file_finding()` re-reads `open_issues()` immediately before each create and comments on a hit rather than skipping. `write_triage_ledger`, `NodeDisposition`, and the three prompt builders are gone from `scripts/` and `tests/`, with only `docs/plans/` and `docs/archive/` historical mentions left. The `FakeGitHub` fixture is genuinely stateful (its `create_issue` mutates `open_map`), so `TestFilingIdempotence` is not tautological.

### Blockers
- None

### Tech Debt
- None

### Nits
- None

**Reliability caveat recorded by the aggregator:** this judge's cited line numbers did not survive verification (it placed `create_issue()` at 1204-1274; it is at 2407) and it asserted `create_issue()` has a single call site, which is false — there are two (`_file_finding()` and `file_seed_umbrella()` at :2615). The second call site is exactly where the risk judge's blocker lives. This verdict is therefore **not** read as corroboration of a clean review.

valorengels:

Lens: blast radius and failure modes of unattended nightly issue creation, given the ~118 real issues two prior incidents already spammed (#3382-#3405, #3423-#3516).

### Blockers

**1. `NIGHTLY_AUTO_FILE=false` does not stop the seed/re-baseline filing path — `scripts/nightly_regression_tests.py:2544-2615`**

`auto_file` is resolved exactly once, at `:3144`, inside `dispatch_findings()`, and is checked at `:2968` in `_file_finding()`. `file_seed_umbrella()` is a *second* path to `create_issue()` (`:2615`) and it is called from `main()` at `:3538`, outside `dispatch_findings()` entirely. Neither the function nor its call site reads `NIGHTLY_AUTO_FILE`; the only knob it honors is the CLI `--dry-run` flag, which is a different control than the operator break-glass switch.

Failure scenario: after a bad night an operator sets `NIGHTLY_AUTO_FILE=false` — the documented stop button that is supposed to work without a deploy. If that run (or a later one) is a seed/re-baseline run — first run, or any run where `COLLECTION_PATHS` differs from persisted state, via `is_reseed` at `:3457` — the script still creates a real GitHub issue. The PR body states the kill switch is "a kill switch, not a second code path"; `file_seed_umbrella()` is precisely a second code path that bypasses it.

Test coverage confirms the gap rather than closing it: `TestKillSwitchSuppressesCreates` (`tests/unit/test_nightly_regression_tests.py:1964-1993`) exercises only `dispatch_findings()`, so this path has zero assertions tying it to `NIGHTLY_AUTO_FILE`.

### Tech Debt

**The seed path re-implements its own dedup instead of reusing `_file_finding()` — `scripts/nightly_regression_tests.py:2544-2615`.** `file_seed_umbrella()` duplicates the open/closed read-and-decide logic that `_file_finding()` centralizes, with its own `open_issues()` / `closed_issue_dispositions()` calls and its own fail-open comments. That duplication is how the kill-switch check came to be omitted, and the two filing paths can drift again on the next change.

### Nits
- None

### What held up under tracing

`create_issue()` (`:2407-2470`) never retries and is argv-safe. The run lock (`_acquire_run_lock`, `:593-618`) uses `fcntl.flock(LOCK_EX|LOCK_NB)`, which the OS releases on process exit including SIGKILL, so a stale lock cannot wedge the job. Fail-open dedup is bounded: `open_issues()` returns `None` on any read failure and callers log "failing open" and proceed, but the per-run fingerprint map plus `NIGHTLY_MAX_ISSUES_PER_RUN` (default `10`, `:281`) cap one blind night at 10 issues, checked before each create and decremented only on the confirmed-`"created"` branch. The autouse `_no_real_tracker_writes` guard (`tests/unit/test_nightly_regression_tests.py:41-64`) wraps `subprocess.run` and hard-fails on `gh issue create` / `gh issue close` argv, closing the "stubbed comment but not create" hole that produced #3423-#3516.

valorengels:
## Review: CHANGES REQUESTED

- **PR** #3517 — `session/dev-48ae5d13` → `main`
- **Head SHA reviewed** `3a28747fb169dd8b69eada5b2d106c4e74a7ab50`
- **Plan** `docs/plans/nightly-filing-into-detector-collapse-by-root-cause.md`
- **Issue** #3418
- **Mode** independent roster (2 judges) — `code-quality`, `risk`
- **Consensus** `any-blocker-wins`, n=2, expected_n=2, no quorum shortfall, **judges disagreed**

### Consensus

`code-quality` returned APPROVED; `risk` returned CHANGES REQUESTED with one blocker. Any-blocker-wins carries the aggregate to **CHANGES REQUESTED**.

The disagreement is not a close call. The `code-quality` judge's line citations did not survive the aggregator's spot-check — it placed `create_issue()` at 1204-1274 when it is at `:2407`, and it asserted a single call site for `create_issue()` when there are two. The second call site, `file_seed_umbrella()` at `:2615`, is where the blocker lives. That judge's APPROVED is recorded but is not treated as corroboration.

### Blockers

**1. The `NIGHTLY_AUTO_FILE=false` kill switch does not cover the seed/re-baseline filing path** — `scripts/nightly_regression_tests.py:2544-2615`, called from `main()` at `:3538`.

Independently verified by the aggregator, not taken on the judge's word:
- `auto_file = resolve_bool_knob("NIGHTLY_AUTO_FILE", AUTO_FILE_DEFAULT)` appears exactly once, at `:3144`, inside `dispatch_findings()`.
- The only `if not auto_file:` guard is at `:2968`, inside `_file_finding()`.
- `file_seed_umbrella()` calls `create_issue()` at `:2615` and is invoked from `main()` at `:3538`, outside `dispatch_findings()`. It reads no knob other than the CLI `--dry-run` flag.

An operator who sets `NIGHTLY_AUTO_FILE=false` to stop unattended filing still gets a real issue created on any seed or re-baseline run. Given that this PR exists because filing created ~118 unwanted issues across two incidents, a break-glass switch with a hole in it is a merge blocker, and the PR body's claim that it is "a kill switch, not a second code path" is not currently true.

Fix shape: gate `file_seed_umbrella()` on the same `resolve_bool_knob("NIGHTLY_AUTO_FILE", ...)` read, resolved at call time, suppressing the create while still logging the issue it would have filed — matching `_file_finding()`'s suppressed branch. Add a `TestKillSwitchSuppressesCreates` sibling that exercises the seed path, setting the env var after import so an import-time read fails the test.

### Tech Debt

**1. The seed path re-implements dedup rather than reusing `_file_finding()`** — `scripts/nightly_regression_tests.py:2544-2615`. Its own `open_issues()` / `closed_issue_dispositions()` reads and fail-open comments duplicate logic `_file_finding()` already centralizes. That duplication is the mechanism by which the kill-switch check was omitted, so the two paths will drift again. The stricter never-re-file rule for a closed umbrella is a genuine behavioral difference and is well justified in the docstring — the ask is to unify the knob read and the create, not to flatten that rule.

### Nits

**1. Two `## Verification` rows false-fail on a stray backtick** — `docs/plans/nightly-filing-into-detector-collapse-by-root-cause.md:1311` and the historical-closures row. Both Command cells end with an unbalanced `` ` ``, which the runner passes through to `/bin/sh`, producing `unexpected EOF while looking for matching \``` and exit 2. Run directly, the underlying check passes: `scripts/pytest-clean.sh tests/unit/test_nightly_regression_tests.py::TestPromptsNeverNameTheSearchIndex -q` → **6 passed**. These are plan-authoring formatting defects, not code defects, but they will keep turning a correct build red on every re-review until the cells are balanced.

### Gates

- `ruff check .` — **pass** (All checks passed)
- `ruff format --check .` — **pass** (1658 files already formatted)
- `## Verification` table — 16 of 18 rows pass. The 2 FAILs are the stray-backtick artifact above; one of them is additionally marked POST-MERGE, NOT A MERGE GATE by the plan itself.
- `scripts/pytest-clean.sh tests/unit/test_nightly_regression_tests.py` — **247 passed** (PR body's number reproduced independently by the `code-quality` judge).
- Row 17 (`#3419 collapsing logic untouched`), which the PR body flagged for a reviewer's eye, **passes** under the runner. The author's analysis was correct: the single `grep` match was git's hunk-header function annotation, not a changed line.
- No UI files in the diff — the visual proof gate is a no-op.

### Verdict

**CHANGES REQUESTED** on one blocker. Everything else in this PR — the deterministic filing loop under the run lock, the derived `issues_filed`, the fingerprint-over-stable-identity choice, the bounded check-then-act window, the deliberate read-fails-open / write-fails-closed asymmetry, the deletion of the ledger and dispositions, and the autouse live-tracker guard — was traced and held up. Close the seed-path hole in the kill switch and this is ready.

valorengels:
## Review: APPROVED

- **PR** #3517 — `session/dev-48ae5d13` → `main`
- **Head SHA reviewed** `65aa1568f762b92598fa7d6000c72a701bce69b1` (re-review after patch; prior verdict was CHANGES REQUESTED on `3a28747fb`)
- **Plan** `docs/plans/nightly-filing-into-detector-collapse-by-root-cause.md`
- **Issue** #3418
- **Mode** independent roster (2 judges) — `code-quality`, `risk`
- **Consensus** `any-blocker-wins`, n=2, expected_n=2, no quorum shortfall, **judges agree**

### The blocker is closed

`65aa1568f` gates the seed/re-baseline umbrella create on `NIGHTLY_AUTO_FILE`. Both judges verified it independently, and their citations now agree with each other and with the aggregator's own grep — which is the check the previous round failed:

- The knob is read at call time **inside** `file_seed_umbrella()` (`scripts/nightly_regression_tests.py:2617`), not threaded in as a parameter a future call site could forget.
- `create_issue(` has exactly **two** call sites in the module — `:2640` in `file_seed_umbrella()`, gated at `:2617`, and `:3003` in `_file_finding()`, gated at `:2993`. No ungated path to a create remains, so the PR body's claim that this is "a kill switch, not a second code path" is now true.
- Branch order is correct: both comment branches run **before** the guard, so recurrence comments still post during a kill-switch night; the guard sits **before** the blind-dedup fail-open branch, so the switch beats a fail-open create rather than losing to it.

### The return-`None`-on-suppression choice was examined, not waved through

A suppressed seed returns `None`, `main()` (`:3592-3599`) calls `_fatal()`, and no `save_last_run()` is reached, so the next run re-seeds. The risk judge traced `_fatal()` (`:1446`) and confirmed it only logs and returns exit 1 — no Telegram notification, no state write. The alternative, returning a sentinel and persisting the baseline, would mark the whole absorbed population as filed against an umbrella nobody created, and `seeded_nodes` is sticky, so that population would be suppressed permanently. The chosen direction fails closed and is the safer of the two.

### Regression check on the patch

The three new tests in `TestSeedUmbrellaIsCreatedByTheDetector` each set `NIGHTLY_AUTO_FILE` via `monkeypatch.setenv` inside the test body — after import — which is what makes them capable of distinguishing a call-time read from a captured default. `test_the_kill_switch_beats_the_fail_open_create` is the load-bearing one: with both maps unreadable, deleting the guard would fall through to `create_issue()` at `:2640` and the assertion `create_calls == []` would fail. The 11 pre-existing tests in that class are unmodified; nothing was weakened or deleted to make the patch pass.

The patch also added a **Break Glass** section to `docs/features/nightly-regression-tests.md`, where the switch previously had no entry anywhere in `docs/`. Every factual claim in it was checked against source rather than taken at its word, including the `NIGHTLY_MAX_ISSUES_PER_RUN` default of 10 (`:292`) and the no-deploy/no-restart claim.

### Tech debt from the prior round — disposition

The prior round's tech-debt finding was that `file_seed_umbrella()` re-implements dedup rather than reusing `_file_finding()`. **Its stated mechanism is closed** — the knob read now lives inside the function, so the specific "that is how the check came to be omitted" failure cannot recur. The structural unification is **declined with reason, not skipped**: the seed's never-re-file-a-`COMPLETED`-closure rule is deliberately stricter than the per-node rule, the plan records that as a decision, and the docstring explains why the check is inline rather than routed through `partition_closed_matches()`. Folding the paths together would mean re-litigating a settled design point on the one code path that can mint a duplicate umbrella. Flagging rather than silently dropping.

### Gates at this head

- `ruff check .` — **pass**
- `ruff format --check .` — **pass** (1658 files already formatted)
- `scripts/pytest-clean.sh tests/unit/test_nightly_regression_tests.py` — **250 passed** (was 247; +3 from the patch)
- `## Verification` table — **17 of 19 rows pass**. The same 2 rows fail as in the prior round, for the same non-code reason, unchanged by the patch:
  - *"Anti-criterion: no prompt tells an agent to create an issue"* — the Command cell at plan `:1311` carries trailing prose containing backticks, which the runner passes to `/bin/sh`, yielding `unexpected EOF while looking for matching \``` and exit 2. Run directly, the check **passes**: `scripts/pytest-clean.sh tests/unit/test_nightly_regression_tests.py::TestPromptsNeverNameTheSearchIndex -q` → **6 passed**.
  - *"historical duplicate closures landed"* — same quoting artifact, and the plan itself labels this row **`[POST-MERGE, NOT A MERGE GATE]`**, owned by the operator running the `## Update System` runbook.

  Both are plan-authoring formatting defects in a document that lives on `main` and is not part of this PR's diff. Filed as a follow-up rather than fixed here, so the PR is not used as a vehicle for an unrelated plan edit. They will keep false-failing every future run against this plan until the Command cells are balanced.
- No UI files in the diff — the visual proof gate is a no-op.

### Verdict

**APPROVED.** The one blocker is closed with a fix that is narrower than the alternatives and fails closed, and it is covered by tests that fail if the fix is removed.

### Post-merge, still owed

Closing the ~26 historical duplicates (#3382-#3397 and the enumerated 09-16 pairs) is an operator step, deliberately kept out of the build pipeline so review-before-execute governs the one irreversible action in scope. The runbook lives in `docs/features/nightly-triage-dispatch.md`. Merging this PR does not discharge it.

PR #3521, opened 2026-09-19
Restore arm/parent embedding-provider parity in the eval tests (#3355)

### LINKED ISSUE #3355: Corpus digest is not stable across processes: every improvement_eval arm run aborts as infra_failure (22 nodes, 5 files)
## Failure

Nightly regression detector (2026-09-17) confirmed this test fails on `main` (HEAD `b6af08a46`, worktree `nightly-triage-76acb9b8`).

```
tests/unit/test_improvement_eval_agent_run.py:1041: in test_rerun_harness_error_trips_a_zero_cap
    assert "exceed the cap" in (evaluation.notes or "")
E   assert 'exceed the cap' in (("infra_failure: arm corpus digest 10e1b553cd924d33fa073e9551425160369c6af4326238f78b6a3c53c4ae978a differs from the export's 60918e5b5df3712ba566e6a44fcba53472a87ff5eb874ac64e4f24b300823f0f; the arms did not read the frozen corpus"))
E    +  where "infra_failure: arm corpus digest ... differs from the export's ...; the arms did not read the frozen corpus" = <ImprovementEvaluation Popoto object at ...>.notes
```

Log line at the point of failure:
```
WARNING tools.improvement_eval.runner:runner.py:1153 evaluation infra_failure for 9f7b3c9a821c4632921faec366d864a9: arm corpus digest 10e1b553cd924d33fa073e9551425160369c6af4326238f78b6a3c53c4ae978a differs from the export's 60918e5b5df3712ba566e6a44fcba53472a87ff5eb874ac64e4f24b300823f0f; the arms did not read the frozen corpus
```

Reproduced locally with:
```
scripts/pytest-clean.sh tests/unit/test_improvement_eval_agent_run.py::TestAgentCapEnforcement::test_rerun_harness_error_trips_a_zero_cap --no-header -q
```

## This is one of a group of 10 failures, all sharing the same root cause

The nightly run flagged 10 failing tests across three files, and all 10 fail via the exact same code path: `tools/improvement_eval/runner.py:1153` raises `infra_failure` because the arm's post-read corpus digest (`canonical_corpus_digest`) does not match the digest recorded at export time. This trips even in tests that are not about corpus/digest behavior at all (arena ranking, agent-trial dispatch, infra-failure-cap tests), which is why one root cause is suspected rather than 10 independent bugs. Sibling issues (same recurrence, cross-linked back here):

- tests/unit/test_improvement_eval_agent_run.py::TestAgentGate1::test_mismatch_names_trials_and_never_calls_candidate
- tests/unit/test_improvement_eval_agent_run.py::TestAgentTrialBranch::test_dispatches_agent_arm_per_trial_and_scores_through_judges
- tests/unit/test_improvement_eval_agent_run.py::TestAgentTrialBranch::test_manifest_model_in_output_records_leak_without_failing
- tests/unit/test_improvement_eval_arena.py::TestArmRetrieve::test_arm_ranking_matches_in_process_ranking
- tests/unit/test_improvement_eval_arena.py::TestArmRetrieve::test_bm25_hit_beyond_the_assembler_pool_runs_ok
- tests/unit/test_improvement_eval_arena.py::TestTwoArmsReadAByteIdenticalCorpus::test_two_arms_read_a_byte_identical_corpus
- tests/unit/test_improvement_eval_runner.py::TestInfraFailureConditions::test_candidate_arm_worker_failure_is_infra_failure_not_reject
- tests/unit/test_improvement_eval_runner.py::TestInfraFailureConditions::test_candidate_failures_within_the_cap_are_excluded_not_scored
- tests/unit/test_improvement_eval_runner.py::TestInfraFailureConditions::test_incumbent_rerun_failure_excludes_trial_in_candidate_first_order

(Issue numbers for these will be added as a follow-up comment once filed — see this issue for the shared analysis.)

## Likely cause

`tools/improvement_eval/corpus.py`'s own module docstring calls out the exact failure mode being hit:

> `embedding_provenance` is the exporter's current provider fingerprint (`{provider, model, dimensions}`), which differs between the parent and an arm subprocess, and between two parents whose import order differed (`config.memory_defaults.apply_defaults` swallows a circular-import failure and leaves no provider configured).

`canonical_corpus_digest` already strips the `embedding_provenance` manifest key and each record's `state.<field>.provenance` key (`VOLATILE_MANIFEST_KEYS`, `VOLATILE_STATE_KEYS`), specifically to absorb that fingerprint drift. But it does **not** strip the embedding *vector values themselves* — only the fingerprint metadata beside them. If the embedding provider actually configures differently (or fails to configure) between the exporting process and the arm subprocess/second call, the real vector bytes carried in `state.<field>` differ, and the digest mismatch is genuine, not a false positive from stale metadata.

The most recent commit touching this exact configuration path is the suspect: **`cc1938981` "Fix apply_defaults embedding-provider import cycle (#3310) (#3353)"**, merged 2026-09-16 23:20, which landed after the `agent_run` arm work (`8950dbdda`, #3311/#3350) that most of these tests belong to. That commit:

- Moved `from models.memory import Memory` in `agent/session_health.py` from module top-level to a function-local import inside `cleanup_corrupted_agent_sessions`, specifically to cut an import cycle that was leaving `config.memory_defaults.apply_defaults()`'s embedding-provider configuration step silently unconfigured depending on import order.
- Changed the `except Exception: pass` in `apply_defaults()` to a logged warning instead of silent swallow.

This changes *when* `configure_embedding_provider()` runs relative to other imports, which is exactly the axis the corpus.py docstring says causes cross-process/cross-call embedding-provider drift. It's plausible the fix changed which import order wins in the test/arm-subprocess context without making the two paths (parent process export vs. arm subprocess restore-and-read) actually converge on the same provider configuration — so the corpus digest genuinely differs because the embeddings differ, not because of a comparison bug.

Also worth checking as an alternative/contributing cause: `tools/improvement_eval/arm_worker.py` (touched by `8950dbdda`, #3311/#3350, the immediately preceding commit) added the `agent_run` job mode and its own `_arm_digest`/restore/read path — if it invokes model imports or the embedding provider setup in a different order than the plain `retrieve`/`digest` job modes, that would independently reproduce the same symptom for the agent-run-arm tests in this group.

## Suggested next steps

1. Reproduce narrowly: run `tests/unit/test_improvement_eval_arena.py::TestTwoArmsReadAByteIdenticalCorpus::test_two_arms_read_a_byte_identical_corpus` (the most directly on-point of the group) and log `agent.embedding_provider`'s resolved provider/model in both the parent process and the arm subprocess to confirm they diverge.
2. Bisect: check out `5362c7fce` (lane 4, before `8950dbdda`/#3311 and before `cc1938981`/#3310) and run the arena tests to confirm they pass there, then step forward commit-by-commit through `8950dbdda` and `cc1938981` to isolate which one flips the digest.
3. If the embedding provider genuinely differs between parent and arm subprocess, either (a) pin/propagate the resolved provider config explicitly into the arm subprocess environment (similar to the existing `RETRIEVAL_MODE=current` pin added for the B1 fix in `5362c7fce`), or (b) pin a deterministic/no-op embedding provider for eval-corpus tests so vector bytes are stable regardless of provider-config import order.
4. Add a regression test that asserts `configure_embedding_provider()` resolves identically in the parent test process and inside a forked/subprocess arm, to lock in whichever fix lands.

## Environment

- Worktree: `nightly-triage-76acb9b8`, branch `session/nightly-triage-76acb9b8`, based on `main` @ `b6af08a46`
- Command: `scripts/pytest-clean.sh <nodeids> --no-header -q`

valorengels:
## Problem

All 22 test nodes folded into #3355 abort as `infra_failure` at the corpus identity gate (`tools/improvement_eval/runner.py:1153`): the digest the parent recorded at export time does not match the digest an arm reports after restoring and re-exporting.

The gate was reporting a **true** condition. `canonical_corpus_digest` strips the provider fingerprint but deliberately keeps the embedding vector bytes, so parent and arm must resolve the same embedding provider — and under pytest they did not.

## Root cause (harness, not production)

`tests/unit/conftest.py`'s autouse `_no_live_embedding_provider` nulls popoto's process-global provider in the **parent**, while `arena.build_child_env` copies `os.environ` wholesale into the arm **child**. The child's `apply_defaults` therefore saw `OPENAI_API_KEY` and installed a live 1536-dim `OpenAIProvider`. The parent exported vectorless records, the arm exported vector-carrying ones, and every run in these five modules aborted at the gate. It was also calling the live OpenAI embedding endpoint for real from the arm child.

Before `cc1938981` the import cycle left the provider unconfigured on *both* sides, which is why the asymmetry only surfaced after that commit. No production behavior is wrong here; the asymmetry is introduced at the harness layer and is fixed at the harness layer.

## Change

A shared `arm_child_embedding_parity` fixture in `tests/unit/improvement_eval_runner_support.py`, applied across the five affected modules. It pins `OPENAI_API_KEY` **empty-but-present** in the environment the arm child inherits.

The pin is deliberately not `delenv`: `configure_embedding_provider` reacts to an *absent* key by re-reading it from `REPO_ROOT/.env` (`agent/embedding_provider.py`), and `scripts/pytest-clean.sh` pins `PYTHONPATH` to the invoking checkout — so a delete is silently undone in any checkout carrying the vault `.env` symlink. `load_dotenv` defaults to `override=False` and treats `""` as present, so the empty pin survives that fallback and the provider resolves to `None` in every checkout.

Diff is test-only. The digest gate itself is untouched.

## Verification

Guard proven red against the known-bad state, then green with the fix:

- With `improvement_eval_runner_support.py` and `test_improvement_eval_arena.py` restored to `origin/main`, `TestTwoArmsReadAByteIdenticalCorpus::test_two_arms_read_a_byte_identical_corpus` fails on the digest assertion (`e036f174…` vs `6c27d130…`).
- With the fix applied, all five modules pass: **210 passed in 451s**.

```
scripts/pytest-clean.sh tests/unit/test_improvement_eval_agent_run.py \
  tests/unit/test_improvement_eval_arena.py tests/unit/test_improvement_eval_runner.py \
  tests/unit/test_improvement_eval_runner_guards.py tests/unit/test_improvement_experiment.py
```

Closes #3355

---

## Per-test disposition table (nightly flagged-test sweep)

All 22 nodes folded into #3355 were re-run on `main` and confirmed still failing before a disposition was assigned. Every one gets the same disposition — **REFACTOR** — because the leaf assertions are correct guards and the only defect is the parent/arm embedding-provider asymmetry introduced by the test harness. No production code is touched.

| # | Test node | Disposition | Landing |
|---|---|---|---|
| 1 | `tests/unit/test_improvement_eval_agent_run.py::TestAgentCapEnforcement::test_rerun_harness_error_trips_a_zero_cap` | REFACTOR | this PR |
| 2 | `tests/unit/test_improvement_eval_agent_run.py::TestAgentGate1::test_mismatch_names_trials_and_never_calls_candidate` | REFACTOR | this PR |
| 3 | `tests/unit/test_improvement_eval_agent_run.py::TestAgentTrialBranch::test_dispatches_agent_arm_per_trial_and_scores_through_judges` | REFACTOR | this PR |
| 4 | `tests/unit/test_improvement_eval_agent_run.py::TestAgentTrialBranch::test_manifest_model_in_output_records_leak_without_failing` | REFACTOR | this PR |
| 5 | `tests/unit/test_improvement_eval_arena.py::TestArmRetrieve::test_arm_ranking_matches_in_process_ranking` | REFACTOR | this PR |
| 6 | `tests/unit/test_improvement_eval_arena.py::TestArmRetrieve::test_bm25_hit_beyond_the_assembler_pool_runs_ok` | REFACTOR | this PR |
| 7 | `tests/unit/test_improvement_eval_arena.py::TestTwoArmsReadAByteIdenticalCorpus::test_two_arms_read_a_byte_identical_corpus` | REFACTOR | this PR |
| 8 | `tests/unit/test_improvement_eval_runner.py::TestInfraFailureConditions::test_candidate_arm_worker_failure_is_infra_failure_not_reject` | REFACTOR | this PR |
| 9 | `tests/unit/test_improvement_eval_runner.py::TestInfraFailureConditions::test_candidate_failures_within_the_cap_are_excluded_not_scored` | REFACTOR | this PR |
| 10 | `tests/unit/test_improvement_eval_runner.py::TestInfraFailureConditions::test_incumbent_rerun_failure_excludes_trial_in_candidate_first_order` | REFACTOR | this PR |
| 11 | `tests/unit/test_improvement_eval_runner.py::TestInfraFailureConditions::test_judge_provider_unreachable` | REFACTOR | this PR |
| 12 | `tests/unit/test_improvement_eval_runner.py::TestInfraFailureConditions::test_parity_miss_never_invokes_the_candidate_arm` | REFACTOR | this PR |
| 13 | `tests/unit/test_improvement_eval_runner.py::TestVerdicts::test_accept_path_end_to_end` | REFACTOR | this PR |
| 14 | `tests/unit/test_improvement_eval_runner.py::TestVerdicts::test_infra_failure_and_reject_have_disjoint_causes` | REFACTOR | this PR |
| 15 | `tests/unit/test_improvement_eval_runner.py::TestVerdicts::test_reject_path_end_to_end` | REFACTOR | this PR |
| 16 | `tests/unit/test_improvement_eval_runner.py::TestVerdicts::test_short_batch_is_inconclusive_and_names_the_shortfall` | REFACTOR | this PR |
| 17 | `tests/unit/test_improvement_eval_runner_guards.py::TestBlinding::test_clean_run_records_blinded_true_not_null` | REFACTOR | this PR |
| 18 | `tests/unit/test_improvement_eval_runner_guards.py::TestBlinding::test_identity_leak_sets_blinded_false` | REFACTOR | this PR |
| 19 | `tests/unit/test_improvement_eval_runner_guards.py::TestBlinding::test_judges_never_see_the_true_arm_identity` | REFACTOR | this PR |
| 20 | `tests/unit/test_improvement_eval_runner_guards.py::TestBlinding::test_manifest_branch_leak_on_a_queried_record_sets_blinded_false` | REFACTOR | this PR |
| 21 | `tests/unit/test_improvement_eval_runner_guards.py::TestWedgedExperiment::test_crashed_run_leaves_a_documented_repair` | REFACTOR | this PR |
| 22 | `tests/unit/test_improvement_experiment.py::TestEndToEnd::test_freeze_evaluate_and_apply_verdict_on_real_arms` | REFACTOR | this PR |

No node in this grouping was classified DELETE (none assert removed behavior, none duplicates another node) or ISSUE-DETAIL (the product path under the gate is correct).

### Post-fix verification on this PR's head

```
scripts/pytest-clean.sh tests/unit/test_improvement_eval_agent_run.py \
  tests/unit/test_improvement_eval_arena.py tests/unit/test_improvement_eval_runner.py \
  tests/unit/test_improvement_eval_runner_guards.py tests/unit/test_improvement_experiment.py
210 passed in 401.44s (0:06:41)
```

`ruff check` and `ruff format --check` clean across all six changed files.


valorengels:
## Review (Judge sequential-lenses): Changes Requested — Tech Debt

Both declared lenses (`code-quality`, `risk`) were applied in one sequential pass by a single reviewer — see the aggregate comment for the mode disclosure.

The diff is test-only (6 files, all under `tests/unit/`). It adds an autouse `arm_child_embedding_parity` fixture that pins `OPENAI_API_KEY=""` in the parent env so `arena.build_child_env`'s wholesale `dict(os.environ)` copy carries the empty pin into the arm subprocess, and imports that fixture into the five affected test modules. The digest gate in `tools/improvement_eval/runner.py` is untouched.

**Mechanism verified by reading the code, not by trusting the PR body:**

- `tools/improvement_eval/arena.py:95` — `env = dict(os.environ)`; nothing filters or re-adds `OPENAI_API_KEY`, so the empty pin reaches the child.
- `config/memory_defaults.py:253-255` — `apply_defaults()` calls `configure_embedding_provider()` in the child.
- `agent/embedding_provider.py` — `if not os.getenv("OPENAI_API_KEY")` treats `""` as falsy, falls back to `load_dotenv(REPO_ROOT / ".env")`, which defaults to `override=False` and skips a key already present in `os.environ`. The empty pin therefore survives the `.env` fallback and the provider resolves to `None`. The PR body's "empty-but-present, never `delenv`" rationale is accurate.
- `tests/unit/conftest.py:181` `_no_live_embedding_provider` already nulls the parent's provider, so parent and arm now agree on "no provider" in every checkout.
- Fixture ordering is safe: `corpus_fixture` is function-scoped and not autouse, so the autouse pin is set up first.

**Sweep completeness (checked, not assumed):** the four other `tests/unit/test_improvement_eval*`/`test_improvement_eligibility` modules that touch `arena`/`arm_worker`/`export_corpus` all either mock `run_arm_job`/`arm_redis_server` or call `arm_worker` in-process, so no real arm subprocess is spawned outside the five modules this PR fixes. The five-module scope is the complete set, not a partial checklist.

**Blast radius of the parent-side pin:** grepped every `OPENAI_API_KEY` reader under `tools/`, `agent/`, `config/`, `models/`, `bridge/`, `worker/`. The only consumer reachable from these five modules is `agent/embedding_provider.py`; the judges in these tests are local stubs (`_approving_judge`, `_skipping_judge`, `judges=[]`). No collateral behavior change.

**Numbers reproduced in this environment** (Hard Rule 10 — the PR's figures were not taken on trust):

```
scripts/pytest-clean.sh tests/unit/test_improvement_eval_agent_run.py \
  tests/unit/test_improvement_eval_arena.py tests/unit/test_improvement_eval_runner.py \
  tests/unit/test_improvement_eval_runner_guards.py tests/unit/test_improvement_experiment.py --no-header -q
210 passed in 387.28s (0:06:27)
```

`python -m ruff check` and `python -m ruff format --check` on all six changed files: clean.

## Rubric

- [x] **1. Plan vs. implementation match** — n/a — *no plan document for #3355 (`find_plan_path(3355)` returns `None`; PLAN/CRITIQUE never dispatched). Reviewed against the issue's stated root cause and its suggested fix (b) "pin a deterministic/no-op embedding provider for eval-corpus tests", which is what landed.*
- [x] **2. New code quality** — pass — *one small fixture, precise name, a docstring that explains the mechanism and the `delenv` trap rather than restating the code.*
- [x] **3. Test coverage** — pass — *the change is itself the test fix; all 22 flagged nodes are in the five modules and the full set passes locally. The guard was proven red against the known-bad state (`origin/main` versions of the two files) before the fix per the PR body.*
- [x] **4. Regression risk to existing callers** — pass — *no production file touched; `git diff --name-only` is six `tests/unit/` paths.*
- [x] **5. Data integrity** — n/a — *no model, schema, or migration change.*
- [x] **6. Security** — pass — *no secret is introduced; the change removes a live OpenAI embedding call that these unit tests were making for real. Pinning the key empty in a test process is a de-escalation, not a hardcoded credential.*
- [x] **7. Documentation accuracy** — pass — *no plan `## Documentation` section exists to satisfy; the fixture docstring is the documentation and is accurate (see the nit below for one ephemeral phrase).*
- [x] **8. PR body accuracy** — pass — *every load-bearing claim was re-derived here: file list, "210 passed", ruff clean, the `override=False` behavior, the `build_child_env` copy. The 22-node disposition table matches the issue's node list.*
- [x] **9. Disclosed deferrals** — n/a — *the PR body discloses no deferrals.*
- [x] **10. Follow-up claims verified** — n/a — *no `filed as follow-up #N` claims in the body.*

## Pre-Verdict Checklist

- **1. All plan acceptance/success criteria validated against diff** — N/A — *no plan doc; the issue's root-cause hypothesis and suggested fix (b) are both satisfied.*
- **2. No-Gos from plan — none violated** — N/A — *no plan.*
- **4. New integration tests — exercise serialization boundary (not in-memory only)** — PASS — *the five modules drive real `redis-server` arm subprocesses; the fix restores that path rather than mocking it away.*
- **5. Plan internal consistency — spike findings match task steps** — N/A — *no plan.*
- **6. No hardcoded secrets or debug artifacts** — PASS — *the only value introduced is the empty string.*
- **7. New public APIs — docstrings present** — PASS — *the fixture carries a full docstring.*
- **9. Tests added for new behavior** — N/A — *this PR repairs existing tests; no new behavior.*
- **10. Tests cover the failure path (not just happy path)** — PASS — *the repaired set includes the infra-failure, blinding, and wedged-experiment guards.*
- **11. UI changes (if any) — screenshot captured** — N/A — *no UI file in the diff; the visual-proof gate is a no-op.*
- **12. Docs updated for user-facing changes** — N/A — *test-harness change, nothing user-facing.*

### Blockers
- None

### Tech Debt
- None

### Nits
- [ ] **`tests/unit/improvement_eval_runner_support.py:193`** — `does and this worktree does not.` — The docstring pins a claim about the reviewer's *current* worktree ("...which the main checkout does and this worktree does not"). Once merged, "this worktree" names nothing, and the sentence becomes a historical artifact of the lane that wrote it (Development Principle 1: describe only the new status quo). The load-bearing half of the sentence — that a checkout carrying the vault `.env` symlink silently undoes a `delenv` — should stay; drop the worktree-relative clause.
  **Severity:** nit
  **Fix:** end the sentence at "...any checkout that has the vault `.env` symlink."

### Miscellaneous
- None

- None

### Verification Results
- No plan document, therefore no `## Verification` table to run. The PR's own verification command was executed instead and reproduced: `210 passed in 387.28s`.
- `python -m ruff check` — All checks passed.
- `python -m ruff format --check` — 6 files already formatted.

### Screenshots

valorengels:
## Review: Changes Requested — Quorum Shortfall (no code findings above nit)

**Mode:** `sequential lenses (Agent tool unavailable: not in tool list)`

This repo declares a mandatory two-judge REVIEW roster (`code-quality`, `risk`, `docs/sdlc/do-pr-review.md` § Multi-Judge Consensus). This review ran inside an SDLC stage-runner context that has no Agent/Task tool, so neither judge could be dispatched as an independent subagent. Both lenses were therefore applied by one reviewer in a single sequential pass and report as **one** judge. Per #3198 that is recorded as the shortfall it is, never read back as agreement between judges.

`compute_consensus(judges=[sequential-lenses], rule="any-blocker-wins", expected_judges=2)`:

```
verdict: CHANGES REQUESTED
k: 1   n: 1   expected_n: 2
quorum_shortfall: true
blockers: 1 (conservative outcome, not a code finding)
```

**What the single pass actually found on the code:** 0 blockers, 0 tech debt, 1 nit. Detail, rubric, checklist and reproduced verification numbers are in the per-judge comment above.

- Test-only diff (6 files, all `tests/unit/`); the corpus digest gate itself is untouched.
- The parent/arm embedding-provider asymmetry is real and the fix sits at the layer that introduced it. Mechanism verified end to end by reading `arena.build_child_env`, `config.memory_defaults.apply_defaults`, and `agent.embedding_provider.configure_embedding_provider` — the empty-but-present pin survives the `.env` fallback because `load_dotenv` defaults to `override=False`.
- Sweep confirmed complete: every other eval test module that touches arms mocks the subprocess or runs in-process, so no sixth module is silently left broken.
- Reproduced independently in this environment: `210 passed in 387.28s`; `ruff check` and `ruff format --check` clean.

**Disposition for the router:** the only substantive work item is the one nit. The `CHANGES REQUESTED` verdict is driven by the quorum floor, not by the code. Re-dispatching REVIEW from a context that can spawn subagents would let the declared roster actually run and is the only way this PR reaches a genuine `APPROVED`.

valorengels:

Verified the mechanism against source rather than trusting the diff's prose.

**Mechanism correctness (confirmed, not assumed):**
- `agent/embedding_provider.py` gates on `if not os.getenv("OPENAI_API_KEY")` then calls `load_dotenv(REPO_ROOT / ".env")` with the default `override=False`. Inspected the installed `dotenv.main.DotEnv.set_as_environment_variables`: `if k in os.environ and not self.override: continue` — **membership, not truthiness**. An empty-but-present pin is therefore not overwritten by the vault `.env`, and the second `if not os.getenv(...)` returns `None`. The docstring's central claim holds exactly as written.
- `tools/improvement_eval/arena.py:87 build_child_env` does `env = dict(os.environ)` — unfiltered copy, so the empty pin propagates to the arm child.
- `config/memory_defaults.py:253-255` `apply_defaults` calls `configure_embedding_provider`, so the child's provider really is the 1536-dim `OpenAIProvider`.
- `tools/improvement_eval/corpus.py:55-78` — `canonical_record` pops only the volatile keys and keeps vector bytes, as the docstring says.

**Coverage consistency:** all five unit modules that spawn a *real* arm child now import the fixture. The other arena-touching modules patch `arena.run_arm_job` / `arena.arm_redis_server`, so no child process exists and no parity problem arises — correctly left alone.

**Design:** one shared fixture instead of five copies; `autouse` scoped by explicit name import rather than a conftest (which would over-apply); `# noqa: F401 -- fixture` matches the file's existing convention. No dead or commented-out code. The lane-relative phrasing flagged in the prior round is gone — the docstring now states the general condition ("any checkout that has the vault `.env` symlink"). `ruff check` and `ruff format --check` clean on all six files.

**Blockers:** 0 | **Tech debt:** 0 | **Nits:** 1

- *Nit* — `tests/unit/improvement_eval_runner_support.py:3`: module docstring still enumerates two consumers ("Both test_improvement_eval_runner.py and test_improvement_eval_runner_guards.py") but the module now serves five. Either name all five or drop the enumeration.

valorengels:

Verified on `session/dev-38811626` @ `0d33164a1`. `git diff main...HEAD --stat` confirms test-only: no `tools/`, `agent/`, `models/`, `bridge/`, `worker/` files touched.

**Environment-mutation safety.** `monkeypatch.setenv("OPENAI_API_KEY", "")` at `tests/unit/improvement_eval_runner_support.py:196` in a function-scoped autouse fixture — monkeypatch tears down per test, no cross-test leak. It reaches arm children only because `tools/improvement_eval/arena.py:95` snapshots `dict(os.environ)` at spawn time, which is precisely the intended reach.

**Checkout-independence — verified empirically, not from the docstring.** Ran the real path in this checkout (which *does* carry the vault `.env` symlink) with the key pinned empty: after `load_dotenv` the value is still `""`, and `configure_embedding_provider()` returned `None`. The fix genuinely removes the checkout dependence rather than relocating it.

**No second provider path.** `config/memory_defaults.py:251-259` is the only provider install site; there is no Ollama/Voyage fallback that could re-introduce an asymmetric child provider once the OpenAI key is empty.

**Test-masking / false-green.** `test_unequal_corpus_digests` forces the mismatch through a `mock.patch` on `run_arm_job`, so the gate is still exercised positively and independently of the env. Notably the risk ran the *other* direction before this PR: that test asserted `"did not read the frozen corpus"`, the exact message the #3355 bug produced — it was passing for the wrong reason, and this change **restores its teeth**. No assertion in the six modules depends on vector recall quality; none reference `OPENAI_API_KEY`.

**Coverage sweep (ran, not enumerated).** `grep -rn "OPENAI_API_KEY" tests/` plus a sweep for `arm_redis_server|run_arm_job|arm_worker|export_corpus|canonical_corpus_digest`. The four eval modules the fix skips are all clean (they mock the arena or run in-process). Integration modules are correctly excluded — `_no_live_embedding_provider` is `tests/unit/conftest.py`-scoped, so an integration parent and its child are already symmetric. **Six sites is the complete set.**

**Production-safety.** No production code. Arms run against a private `redis-server` over a unix socket; the parent runs under the autouse claimed test DB. Nothing touches db 0 or does an unscoped bulk operation.

**Blockers:** 0 | **Tech debt:** 1 | **Nits:** 0

- *Tech debt* — `tests/unit/improvement_eval_runner_support.py:195`: the fixture propagates only via per-module `import ... # noqa: F401`. A future eval module that drives real arms and forgets the import silently regresses to #3355, with no guard. A conftest-scoped autouse fixture gated to the arm-driving modules, or a source-hygiene assertion alongside the existing `TestSourceHygiene` class in `test_improvement_eval_arena.py`, would make the invariant self-enforcing. Not blocking.

valorengels:
## Review: APPROVED

**Head SHA:** `0d33164a129a360fc8e2a576f72792878f31c35b`
**REVIEW_MODE:** independent roster (2 judges)
**Judges run:** 2 of 2 (`code-quality`, `risk`) — quorum met, no shortfall
**Rule:** `any-blocker-wins` | **Blockers:** 0 | **Tech debt:** 1 | **Consensus disagreement:** false

Both judges were dispatched as independent foreground subagents and awaited in-turn, so this run satisfies the mandatory two-judge roster. This supersedes the earlier `CHANGES REQUESTED` on `b2cbcd9c6`, which was a **quorum shortfall artifact** (#3198: the forked review context was withheld the Agent tool and collapsed to one sequential reviewer) and not a substantive finding — that round also reported 0 blockers and 0 tech debt.

### Summary

Test-only change, 6 files, +50 lines, no production code. It fixes #3355, where arm child processes resolved a different embedding provider than their parent, destabilizing the corpus digest and aborting every `improvement_eval` arm as `infra_failure`.

Both judges independently verified the mechanism against source rather than accepting the diff's prose. The load-bearing claim — that an empty-but-present `OPENAI_API_KEY` survives `load_dotenv`'s `override=False` — was confirmed two ways: by reading the installed `dotenv` source (`if k in os.environ and not self.override: continue`, membership rather than truthiness) and by executing the real resolution path in a checkout carrying the vault `.env` symlink, which returned `None` as intended. The fix therefore removes the checkout dependence rather than relocating it.

The `risk` judge's coverage sweep confirmed the six call sites are the complete set: the eval modules left untouched either mock the arena or run in-process, so no child spawns and no parity problem exists. Integration tests are correctly out of scope, since the `_no_live_embedding_provider` asymmetry is `tests/unit/conftest.py`-scoped.

Worth recording: `test_unequal_corpus_digests` previously asserted the very message the #3355 bug emitted, so it was passing for the wrong reason. This change restores that assertion's teeth.

The single nit from the prior round (lane-relative docstring phrasing naming "this worktree") was fixed in `0d33164a1` and verified resolved.

### Non-blocking follow-ups

- *Tech debt* — `tests/unit/improvement_eval_runner_support.py:195`: the parity fixture propagates only via per-module `import ... # noqa: F401`; a future real-arm module that omits the import silently regresses to #3355 with no guard. A conftest-scoped autouse fixture or a source-hygiene assertion beside `TestSourceHygiene` would make the invariant self-enforcing.
- *Nit* — `tests/unit/improvement_eval_runner_support.py:3`: module docstring still enumerates two consumers; it now serves five.

Neither is merge-blocking.

PR #3518, opened 2026-09-19
Drop the eager agent_session_queue re-export that defeats session_lifecycle's lazy import (#3389)

valorengels:
Closes #3389.

## Problem

`models/session_lifecycle.py` deliberately keeps `agent.agent_session_queue` behind a function-local import (line 618, inside the function that needs `checkpoint_branch_state`) so the module can be imported without dragging in `agent.session_executor`. Its own module docstring says so:

> Uses lazy imports for heavy dependencies (`tools.session_tags`, `agent.agent_session_queue`)

That lazy import was defeated one level up. `agent/__init__.py` opened with:

```python
from .agent_session_queue import (
    PRIORITY_RANK,
    AgentSession,
    enqueue_agent_session,
    register_callbacks,
)
```

so **any** `from agent.<submodule> import ...` executes the package `__init__` and eagerly imports the whole queue. `models/session_lifecycle.py:29` (`from agent.lock_policy import record_lock_degradation`) is enough to trigger it; `agent.embedding_provider` reaches it transitively as well.

`tests/unit/test_session_lifecycle_consolidation.py::TestImportSafety::test_lazy_imports_not_triggered_at_module_level` has been failing on `main` since this landed. Note the test's message points at `models/session_lifecycle.py` ("they must stay inside the functions that need them"), but the function-local import there was already correct — the defect was in the package barrel.

## Change

Delete the `.agent_session_queue` re-export block and its four `__all__` entries. That is the entire diff: 10 deletions, no additions.

**No call sites needed fixing.** All four names have zero importers via the package barrel:

```
$ grep -rn --include='*.py' -E '^\s*from agent import .*(AgentSession|enqueue_agent_session|register_callbacks|PRIORITY_RANK)' .
$ grep -rn --include='*.py' -E '\bagent\.(AgentSession|enqueue_agent_session|register_callbacks|PRIORITY_RANK)\b' . | grep -v 'agent\.agent_session_queue'
$ grep -rn --include='*.py' 'from agent import ($' .
```

All three return nothing. Every consumer already imports from `agent.agent_session_queue` directly. The re-export bought nothing and cost 1.7s on every import of the package.

## Evidence

`python -X importtime -c "import models.session_lifecycle"`, three runs each:

| | run 1 | run 2 | run 3 |
|---|---|---|---|
| before | 2.465s | 2.359s | 2.402s |
| after | 0.726s | 0.719s | 0.722s |

**~1.68s saved, a 70% reduction.** The offending frames are gone from the trace entirely:

```
[...]

Issue #3389, opened 2026-09-17
agent/__init__.py eagerly re-exports agent_session_queue, defeating models/session_lifecycle.py's lazy import (~2.0s)


valorengels:
Node: `tests/unit/test_session_lifecycle_consolidation.py::TestImportSafety::test_lazy_imports_not_triggered_at_module_level`

Found by the nightly regression detector on 2026-09-17 (recurrence confirmed 2026-09-18). Reproduced at `87eabe1ac`, which is the current tip of `main`, so this fails on `main` too.

## Failure

```
tests/unit/test_session_lifecycle_consolidation.py:481: in test_lazy_imports_not_triggered_at_module_level
    assert leaked == [], (
E   AssertionError: heavy dependencies imported at module level: ['agent.agent_session_queue'].
E   They must stay inside the functions that need them.
E   assert ['agent.agent_session_queue'] == []
```

## Cause

The leak is transitive, not a direct import. `models/session_lifecycle.py` still keeps its
own reference lazy (`models/session_lifecycle.py:618`, `from agent.agent_session_queue import
checkpoint_branch_state`, inside the function). But at module level it does:

```python
from agent.lock_policy import record_lock_degradation   # models/session_lifecycle.py:29
```

Importing any submodule of `agent` executes `agent/__init__.py`, and `agent/__init__.py:3`
eagerly re-exports from the heavy module:

```python
from .agent_session_queue import (
```

So `import models.session_lifecycle` loads `agent.agent_session_queue` by way of the package
`__init__`. Verified directly:

```
$ .venv/bin/python -c "import sys; import agent.lock_policy; print('agent.agent_session_queue' in sys.modules)"
True
```

`models/session_lifecycle.py` last moved in `f02683282` ("Improvement controller lane 3 ...
(#3315)"); the `agent.lock_policy` module-level import arrived with the lock-policy work
(`a9822d719`, #3229).

## Suggested next steps

- Preferred: move the `record_lock_degradation` import into the function that needs it, so
  `models.session_lifecycle` touches no `agent.*` submodule at import time.
- Alternative (wider blast radius, but fixes the whole class): stop eagerly re-exporting
  `agent_session_queue` symbols from `agent/__init__.py` and make callers import the submodule
  directly. Any module-level `from agent.<anything> import ...` anywhere will re-trip this
  test until that package `__init__` stops pulling the queue in.
- Whichever fix lands, keep the subprocess probe as-is: it is the only thing that catches
[...]

Issue #3388, opened 2026-09-17
docs/plans/worktree-single-owner-dispatch.md has a bare `tracking: 3258` (no `#`), failing test_plan_declares_a_tracking_issue


valorengels:
Node: `tests/unit/test_plan_docs.py::test_plan_declares_a_tracking_issue[worktree-single-owner-dispatch.md]`

Found by the nightly regression detector on 2026-09-17 (recurrence confirmed 2026-09-18). Reproduced at `87eabe1ac`, which is the current tip of `main`, so this fails on `main` too.

## Failure

```
tests/unit/test_plan_docs.py:46: in test_plan_declares_a_tracking_issue
    assert _TRACKING_RE.search(head), (
E   AssertionError: worktree-single-owner-dispatch.md has no resolvable `tracking:` frontmatter.
E   assert None
E    +  where None = re.compile('^tracking:.*(?:#\\d+|issues/\\d+)', re.MULTILINE).search(
E        "---\ntracking: 3258\nslug: sdlc-3258\nstatus: draft\n---\n...")
```

## Cause

The plan doc is not missing its `tracking:` line -- it has one, but in a form the invariant
does not accept. `docs/plans/worktree-single-owner-dispatch.md:2` reads:

```
tracking: 3258
```

`tests/unit/test_plan_docs.py:28` requires a real issue reference, `#N` or `.../issues/N`:

```python
_TRACKING_RE = re.compile(r"^tracking:.*(?:#\d+|issues/\d+)", re.MULTILINE)
```

A bare integer matches neither alternative. Every other lane plan writes the full URL form
(`tracking: https://github.com/tomcounsell/ai/issues/3091`), so this doc is the outlier.
The doc was last touched in `23964450c` ("Correct incident 2 attribution: it was the lane's
own reviewer"); the bare-number frontmatter has been there since it was authored.

## Suggested next steps

- Change the frontmatter to `tracking: https://github.com/tomcounsell/ai/issues/3258`
  (or `#3258`) to match the rest of `docs/plans/`.
- Consider whether the regex should also accept a bare `N` after `tracking:`. It deliberately
  should not accept `none yet` (see the comment at `tests/unit/test_plan_docs.py:25-27`), but a
  bare integer is a real reference and the current failure mode is a confusing "has no
  `tracking:` frontmatter" message on a file that visibly has one. If the strict form stays,
  the assertion message should say "unresolvable" rather than "no".

valorengels:
[...]

PR #3522, opened 2026-09-19
Nightly triage: dispositions for 20 non-canonical failing nodes

### LINKED ISSUE #3526: Nightly triage: disposition for the 20 non-canonical failing nodes from the 2026-09-19 run
## Context

The 2026-09-19 nightly run (`data/nightly_tests_last_run.json`, head `4e504ef6b`) flagged 44 tests failing on `main`. Per the principal's direction, every one of the 44 must be **deleted, refactored, or attached to a detailed open issue** — no node may be left failing without a disposition.

24 of the 44 were covered by three canonical issues, all now fixed and merged:

- #3388 — hotfix `7c66d003c`
- #3389 — PR #3518, merged as `4e504ef6b`
- #3355 — PR #3521, merged as `e4239ac48`

This issue tracks the disposition of the **remaining 20 nodes**, which no open issue covered (verified by `gh issue list --search` per file). The third disposition on offer had no target for them, so each is either refactored to assert the real current contract, or deleted with a named superseding commit.

## Scope

All 20 were reproduced on `main` before any change. **None is a product regression.** Each is a stale test pointing at a superseded internal contract, and the deliberate commit that moved each contract is named in the PR ledger. The diff is test-only.

- 16 refactored
- 4 deleted (3 in `test_sdlc_pipeline_lock.py`, 1 in `test_reflections_redis.py`)

Two further nodes were swept in during review — the rest of `TestRedisIndexCleanupReflection`, same root cause as one of the four, missed by the node-list triage because they were latent machine-dependent tests that only pass on a provisioned machine.

## Acceptance criteria

- [ ] Every one of the 20 nodes has a disposition recorded in a ledger that matches the diff
- [ ] No `skip`, `xfail`, or suppression marker is used to make a node green
- [ ] Each refactored node still fails when the production behavior it guards is broken (proved by mutation, not by argument)
- [ ] Each deleted node's contract is either genuinely removed, or covered by a named surviving test or guard
- [ ] The diff stays test-only
- [ ] The 16 surviving nodes pass together on current `main`

## Notes

Related: #3419 (folding same-file nightly failures into umbrella issues).

Two items surfaced in review are deliberately **not** in scope here and want their own follow-ups:

1. The MCP launch test constructs a PATH with `.venv/bin` first and proves `python3` launches under it, but the production claim — that the worker's launchd plist actually orders PATH that way — is docstring-only and tested nowhere.
2. `tests/marker_map.py`'s `KNOWN_MISTAGS` baseline gains an entry, against a comment stating the baseline can only shrink. Whether that growth is policy needs the guard owner's ratification.

valorengels:
## What

Triage of the 20 failing nodes from the 2026-09-19 nightly run (`data/nightly_tests_last_run.json`, head e61b4f0b0) that are **not** covered by the three canonical issues already in flight. The other 24 failing nodes are untouched: #3388 (closed, hotfix 7c66d003c), #3389 (PR #3518), #3355 (PR #3521).

All 20 were reproduced on current `main` before any change. **None is a product regression.** Every one is a stale test pointing at a superseded internal contract, and in each case the deliberate commit that moved the contract is named below. The diff is test-only — no product code changed.

## Ledger

| Node | Disposition |
|---|---|
| `test_connectivity_gaps.py::TestSdkClientEnvVar::test_valor_session_id_conditional_on_session_id` | refactored (reworked after review) |
| `test_connectivity_gaps.py::TestSdkClientEnvVar::test_valor_session_id_in_source` | refactored (reworked after review) |
| `test_memory_mcp_server.py::test_fresh_shell_import_resolution` | refactored (reworked after review) |
| `test_out_of_domain_reclaim.py::test_reap_lookup_error_owner_no_reclaim_no_stale` | refactored |
| `test_out_of_domain_reclaim.py::test_reap_resume_during_drain_not_reclaimed` | refactored |
| `test_parent_child_round_trip.py::...::test_child_working_dir_rooted_under_parent_project_via_projects_json` | refactored |
| `test_reflections_redis.py::TestAnalyzeSessionsFromRedis::test_detects_failed_sessions` | refactored |
| `test_reflections_redis.py::TestRedisIndexCleanupReflection::test_reflection_entry_structure` | **deleted** |
| `test_remote_update.py::TestPinHelpersRefuseLoudly::test_real_pyproject_declarations_resolve` | refactored |
| `test_remote_update.py::TestRestartFlag::test_check_restart_flag_defers_when_jobs_running` | refactored |
| `test_resume_reverification_llm.py::...::test_present_rails_produce_a_grounded_resume_answer` | refactored |
| `test_resume_reverification_llm.py::...::test_uninterrupted_turn_stays_silent` | refactored |
| `test_sdlc_pipeline_lock.py::...::test_critique_sets_lock_blocks_build_plan_clears_lock_routes_build` | **deleted** |
| `test_sdlc_pipeline_lock.py::...::test_write_meta_persists_plan_hash` | **deleted** |
| `test_sdlc_pipeline_lock.py::...::test_write_meta_persists_plan_revising` | **deleted** |
| `test_sdlc_stall_auto_resume_e2e.py::test_action_cooldown_suppresses_a_second_tick` | refactored |
| `test_sdlc_stall_auto_resume_e2e.py::test_steer_lands_on_the_real_steering_queue` | refactored |
| `test_tool_budget_enforcement.py::test_sdk_hook_allows_under_budget` | refactored |
| `test_tool_budget_enforcement.py::test_sdk_hook_blocks_over_budget` | refactored |
| `test_feature_map_markers.py::test_no_violation_outside_known_mistags` | refactored |

## Root causes behind the refactors

- **connectivity_gaps (2)**: `VALOR_SESSION_ID` moved out of the module the source-grep read. Both nodes are now **behavioral**, not source-greps: they drive `_execute_agent_session` with a spy over `agent.session_runner.SessionRunner` and assert on the `session_env` dict production actually hands it. (Reworked in review -- the first pass merely re-pointed the grep at `agent/session_executor.py`, which rebuilt the same time bomb one notch out.)
- **sdlc_stall_auto_resume_e2e (2)**: steers now write the Room key (`steering:room:{room_id}`); the reads passed no `room_id`. The fixture also never set `.slug`, so `_pick_steer_target`'s lane guard (#2755/#3270) fell through to `create` and escalated on the happy path.
- **tool_budget_enforcement (2)**: the hook resolves `VALOR_SESSION_ID` first, and `AGENT_SESSION_ID` is the Popoto AutoKey, not `session_id` (#2205). The test set the wrong identifier namespace, so no session resolved and enforcement was silently skipped.
- **out_of_domain_reclaim (2)**: the reap loop's phase 2 uses `get_by_id_strict` (#1868), not `get_by_id`; patching only `get_by_id` left the reclaim decision reading real data. Now patches `query.filter`, matching the authoritative #1868 regression test.
- **resume_reverification_llm (2)**: the persona caveat was reflowed, not removed (#3069). Anchor now matches within one physical line, and the stripped control cuts by paragraph so the wrapped continuation really goes.
- **remote_update (2)**: dependency pins were bumped past hardcoded literals; the restart-flag fixture's bare `MagicMock` auto-created a truthy `.is_ledger` and was filtered out as a ledger row.
- **memory_mcp_server (1)**: bare `python3` with `PATH` pinned to `/opt/homebrew` is not the production resolution path — the worker plist puts `.venv/bin` first. Uses `sys.executable`, keeping the venv-var strips so `PYTHONPATH` alone is still what is proven.
- **parent_child_round_trip (1)**: child sessions are gated off by default (#1633); sets the documented `VALOR_ALLOW_CHILD_SESSIONS` escape hatch so the working-dir assertions execute again.
- **reflections_redis (1)**: `summary` is now a derived property whose setter partial-saves, so passing it to `create()` silently dropped it. Assigns after create, as real callers do.
- **feature_map_markers (1)**: a new `tests/unit/hooks` file resolves to the correct `sdlc` marker; added the `KNOWN_MISTAGS` entry the guard's own remediation message asks for.

## The deletes, justified

Six nodes deleted in total: four from the original 20, plus two swept in during
review (below).

Deleting coverage is the disposition most worth challenging, so each is argued explicitly.

**`test_sdlc_pipeline_lock.py` (3 nodes).** All three patch `tools.sdlc_meta_set.find_session`. That symbol does not exist in the module and `write_meta()` never calls it: authorization moved to the `run_id`-keyed `PipelineLedger` lease (#2012 task 2), and `session_id` is explicitly dead in that path (`del session_id  # unused -- CLI-flag backward compat only`). `find_session` does still live in `tools/_sdlc_utils.py`, but retargeting the patch there would mock a function the code under test never invokes — a tautologically green test. The gated behavior (critique sets lock -> blocks build; plan clears lock -> routes build) is **still live and still covered**, verified by running `test_sdlc_router.py::TestG7PlanRevisingGuardDirect` + `TestG7ThroughDecideNextDispatch` (14 passed) and `test_sdlc_meta_set.py::TestMetaSetWriteMeta` (3 passed). No coverage is lost.

**`test_reflection_entry_structure` (1 node).** `config/reflections.yaml` moved to the private vault in c2af09602. The vault's `redis-index-cleanup` entry has since drifted from the hardcoded `callable` this test asserts, so the assertion is false by design rather than merely unreadable. Its only real code-under-test, `parse_every_duration("86400s") == 86400`, is already covered by `test_reflection_schedule_grammar.py::TestDurationHelpers::test_parse_every_duration_seconds`.

## Guarding against green-by-weakening

No `skip` and no `xfail` anywhere in this diff. Two `@pytest.mark.asyncio`
decorators were added (`test_connectivity_gaps.py:127,155`) -- a collection
requirement of the sync-to-async rewrite, not a suppression.

- The `test_connectivity_gaps` nodes are proved behaviorally: deleting
  `"VALOR_SESSION_ID": session.session_id or ""` from `agent/session_executor.py`
  in a scratch tree makes both fail with `KeyError: 'VALOR_SESSION_ID'`;
  restoring it makes them pass. Reflowing that same line across three physical
  lines leaves both green -- the source-grep regression class is closed, not
  relocated.
- `test_fresh_shell_import_resolution` is proved by mutating `MCP_SERVER_MODULE`
  to a non-existent module: red, then green on restore.
- `test_sdk_hook_disabled_allows` carried the same wrong-namespace bug as the two
  siblings above and passed identically with `TOOL_BUDGET_ENABLED = True`, so it
  covered nothing. Fixed and proved three ways: fixed+disabled passes,
  fixed+enabled fails, old form+enabled passes.
- `test_real_pyproject_declarations_resolve` was **not** fixed by bumping `1.5.0`
  to `1.7.0` -- that would rebuild the same time bomb. It re-derives each expected
  version from the raw `pyproject.toml` text independently of `get_pinned_version`.

**`TestRedisIndexCleanupReflection`, the remaining 2 nodes (added in review).**
Neither was in the 44-node nightly list; both are the same root cause as the
deleted `test_reflection_entry_structure`, and the node-list triage missed them
by construction.

- `test_reflection_registered_in_yaml` unconditionally opens
  `config/reflections.yaml`, which is gitignored (`.gitignore:8`) and vault-moved
  (`c2af09602`), copied in only by provisioning
  (`scripts/install_email_bridge.sh:150`). It raises `FileNotFoundError` in any
  clean checkout -- a latent machine-dependent test, green only on provisioned
  machines. It is not retargetable test-only: `agent/reflection_scheduler.py:146-157`
  resolves env var -> vault -> gitignored checkout copy, and there is no in-repo
  registry to assert against.
- `test_cleanup_callable_importable` asserts `scripts.popoto_index_cleanup.run_cleanup`
  is importable, but the vault has zero references to it -- nothing schedules it.

What makes this safe is not the deletion argument, it is the surviving guard:
`scripts/update/reflections_callables.py:260::run_registry_probe` is the positive,
fail-closed check that *every* registry callable imports. It runs against the real
vault registry on every `/update` and gates the worker restart via
`data/registry-probe-failed`. That is strictly stronger than a test hardcoding one
literal callable. `run_cleanup` itself keeps behavioral coverage in
`tests/unit/test_popoto_cleanup_reflection.py` (23 tests, module-level import).

## Issues

No open issue flagged any of these 20 nodes (`gh issue list --search` per file, all empty), so no issue comments were owed and none needed filing — the third disposition on offer ("add detail to the open issue") had no target. Related: #3419 (folding same-file nightly failures into umbrella issues).

## Test evidence

Branch rebased onto current `main` and re-verified there (not at the branch point):

```
./scripts/pytest-clean.sh <16 nodes> -p no:randomly
16 passed, 4 warnings in 30.38s
```

`python -m ruff check tests/` and `python -m ruff format --check tests/` clean.

Closes #3526
Refs #3419


valorengels:

**CHANGES REQUESTED** — 1 blocker, 2 tech debt, 2 nits. Head SHA judged: `62237be7793de8e9a823791f31c50076892662d9`.

The green-by-weakening claim was verified, not credited: `git diff origin/main...HEAD | grep -E '^\+.*(skip|xfail|pytest\.mark)'` returns nothing. But a test can be gutted without a marker, and one was.

### Blockers

**1. `tests/integration/test_memory_mcp_server.py:197` — `test_fresh_shell_import_resolution` is now tautological.**

The refactor swapped bare `python3` for `sys.executable`. `sys.executable` is the repo venv, and the repo venv carries an editable install:

```
.venv/lib/python3.14/site-packages/_editable_impl_valor_bridge.pth
```

Mutation experiment — PYTHONPATH removed entirely, plus every var the test strips:

```
$ env -i PATH=/usr/bin:/bin HOME=$HOME .venv/bin/python3 -m mcp_servers.memory_server --help
FastMCP server: memory
Tools: memory_get(memory_id), memory_search(...)
RC=0
```

Both assertions (`returncode == 0`, `"memory_get" in stderr+stdout`) pass with **no PYTHONPATH at all**. The test's own docstring asserts the opposite:

> "with any of these left in place the child could resolve modules through a channel other than `PYTHONPATH` and the test would pass without proving anything."

The venv's own `site-packages` is exactly such a channel, and none of the five popped vars (`VIRTUAL_ENV`, `PYTHONHOME`, `PYTHONUSERBASE`, `PYTHONNOUSERSITE`, `PYTHONSAFEPATH`) disables `.pth` processing. Compounding it: production registers `MCP_SERVER_COMMAND = "python3"` (`scripts/update/mcp_memory.py:39`) — a *name*, PATH-resolved — so invoking by absolute path also drops the one thing this test uniquely covered. If `_expected_entry` stopped emitting `"env": {"PYTHONPATH": repo_root}`, this test stays green.

This is the exact disposition failure mode the triage was supposed to avoid: the node went from red-and-honest to green-and-toothless.

### Tech Debt

**1. `tests/integration/test_tool_budget_enforcement.py:150` — the sibling with the same bug was left in place.** `test_sdk_hook_disabled_allows` still does `monkeypatch.setenv("AGENT_SESSION_ID", s.session_id)` — the exact wrong-namespace bug this PR just fixed in its two neighbours. With `VALOR_SESSION_ID` unset and `AGENT_SESSION_ID` holding a `session_id` rather than the AutoKey `.id`, `resolve_inflight_session` falls through to `get_by_id_strict(session_id)`, finds nothing, returns `None`, and silently allows. The test passes identically with `TOOL_BUDGET_ENABLED = True`, so it does not actually cover the kill switch. Same file, same diagnosis, one line away.

**2. `tests/marker_map.py:186` — the `KNOWN_MISTAGS` entry is defensible but changes a stated policy.** The baseline's own comment (lines 183-185) says "The baseline can still only shrink, never grow", carved out only by "A third entry earns its place only under the same standard." The entry meets the guard's documented remediation contract and its justification text is factually accurate, but ratifying a growth of that baseline is the guard owner's call, not something a triage PR should settle silently.

### Nits

- `tests/integration/test_connectivity_gaps.py:271` — the new class docstring calls `agent/sdk_client.py` "a thin re-export shim". It is 1531 lines. The load-bearing fact (0 occurrences of `VALOR_SESSION_ID`) is correct; the characterization is not.
- `tests/integration/test_reflections_redis.py:193` — the `test_reflection_entry_structure` deletion was reachable as a retarget. Its sibling `test_reflection_registered_in_yaml` (line 183) reads the *same* vault file successfully, so a schema-level assertion (required keys present, `every` parses, `enabled` is a bool, `callable` importable) would have preserved drift-proof structural coverage rather than dropping it.

### Verified sound — no finding

Each of these was read against production code, not taken from the ledger:

- **`test_out_of_domain_reclaim.py`** — the wider `query.filter` patch is narrowly scoped: `_fake_filter` delegates to the captured `real_filter` for any id but the fixture's. `get_by_id` (`models/agent_session.py:1207`) and `get_by_id_strict` (:1253) both route through `cls.query.filter(id=...)`, so this is the correct seam, not an over-broad mock. Swap Phase 2 (`agent/session_health.py:4175`) back to `get_by_id` and the test goes red. It *is* the #1868 guard.
- **`test_tool_budget_enforcement.py`** (the two refactored nodes) — `VALOR_SESSION_ID` is genuinely first in `agent/hooks/session_resolver.py::resolve_inflight_session`. The blocks test observes a real block decision plus the persisted `budget_tripped_reason`.
- **`test_remote_update.py::test_real_pyproject_declarations_resolve`** — not circular. The regex is written independently of `get_pinned_version`, which delegates to `_resolve_declaration` (`scripts/update/deps.py:394`). The `openai is None` assertion still pins floor-is-not-a-pin.
- **`test_resume_reverification_llm.py`** — the anchor `"but not announcing the resume does"` occurs exactly once, at `config/personas/segments/work-patterns.md:12`, guarded by a loud presence assert. The paragraph-based cut is *necessary*: the caveat hard-wraps across lines 12-15, and a physical-line cut left the cross-ref in the "stripped" control, collapsing control and positive.
- **`test_parent_child_round_trip.py`** — `assert rc == 0` sits between the gate and the working-dir assertions, so a refused create fails rather than short-circuits.
- **`test_connectivity_gaps.py`** — the mutation claim reproduces; `"VALOR_SESSION_ID": session.session_id or ""` is verbatim at `agent/session_executor.py:2313`.
- **`test_reflections_redis.py::test_detects_failed_sessions`** — `reflections/session_intelligence.py:85-88` `continue`s past a failed session with an empty summary, so the assertion still has teeth.
- **`test_sdlc_stall_auto_resume_e2e.py`** — `peek_steering_messages(session_id, room_id=...)` (`agent/steering.py:488`) reads both legs; the `== 1` counts still discriminate.
- **`test_feature_map_markers.py`** — legitimate remediation, not suppression: the test itself names "add a KNOWN_MISTAGS entry with a reason" as sanctioned, and rule 2 fails loudly on stale entries.

### The four deletions: all justified

`grep find_session tools/sdlc_meta_set.py` → no match; `del session_id  # unused -- CLI-flag backward compat only` is real at line 167. Replacement coverage read assertion-for-assertion:

| Deleted assertion | Surviving equivalent |
|---|---|
| critique sets lock → routes `/do-plan` (G7) | `tests/unit/test_sdlc_router.py::TestG7ThroughDecideNextDispatch::test_lock_set_critique_just_ran_routes_to_plan_via_g7` |
| lock cleared → routes `/do-build` | same class, `test_no_lock_routes_to_build` |
| `write_meta` persists `plan_revising` | `tests/unit/test_sdlc_meta_set.py:37-58` |
| `write_meta` persists `plan_hash_at_build_start` | `tests/unit/test_sdlc_meta_set.py:173-194` |

For `test_reflection_entry_structure`: `git show --stat c2af09602` confirms `config/reflections.yaml | 312 ------`. The drift is real — the live vault entry reads `'callable': 'agent.session_health.cleanup_corrupted_agent_sessions'` against the test's hardcoded `scripts.popoto_index_cleanup.run_cleanup`. Surviving `parse_every_duration` coverage exists at `tests/unit/test_reflection_schedule_grammar.py:174-180`.

19 of 20 dispositions are correct and the root-cause diagnoses check out against production code. The one blocker is narrow and fixable without touching anything else in the diff.

valorengels:
## Review (Judge risk):

**APPROVED** — 0 blockers, 4 tech debt, 2 nits. Head SHA judged: `62237be7793de8e9a823791f31c50076892662d9`.

The lens here is not line-by-line: it is whether the evidence is proportionate to the cost of being wrong. The load-bearing claim is "none of these 20 is a product regression" — which is exactly the conclusion a tired triager reaches when the real answer is "the product broke and the test correctly noticed."

### The load-bearing claim holds

On the three highest-cost nodes, the product is demonstrably right and the old test demonstrably wrong:

- **`test_tool_budget_enforcement`** — `models/agent_session.py:949-950` (`kwargs.pop("agent_session_id")  # AutoKeyField, ignore`) confirms `agent_session_id` *is* the Popoto AutoKey, and `agent/hooks/session_resolver.py:68-77` resolves `VALOR_SESSION_ID` → `newest_for_session_id()` first, `AGENT_SESSION_ID` → `get_by_id_strict()` second. Production sets both correctly at `agent/session_executor.py:2299` and `:2313`. **Enforcement is not broken in production** — the refactor moved the test onto the production path, and the old test failed loudly (no block) rather than passing vacuously. The `AGENT_SESSION_ID` fallback it stopped using is independently covered at `tests/unit/test_session_resolver.py:54,69,78`.
- **`test_out_of_domain_reclaim`** — `agent/session_health.py:4148` (`owner_records[owner_id] = AgentSession.get_by_id(owner_id)`) vs `:4175` (`fresh = AgentSession.get_by_id_strict(lease.owner_session_id)`). Both route through `cls.query.filter(id=...)`; patching `get_by_id` alone left the *reclaim decision itself* reading real data. The old test asserted a safety property it never exercised. The refactor makes it stricter, not looser.
- **`test_sdlc_stall_auto_resume_e2e::test_steer_lands_on_the_real_steering_queue`** — room plumbing is consistent end to end: writer `agent/session_executor.py:921`, drain `agent/session_runner/runner.py:840-841`, and the test now reads with the same helper. `peek_steering_messages` (`agent/steering.py:503-508`) unions both legs, matching the drain. No silent-steer hazard. The `session.slug = _SLUG` fixture addition matches the documented fail-closed design at `reflections/sdlc_progress.py:816-824` — not a bug being masked.

### The 210-line deletion is not the coverage loss it looks like

`find_session` does not exist in `tools/sdlc_meta_set.py`, so `mock.patch` raised `AttributeError` at `__enter__`: all three nodes were *erroring*, providing zero coverage at deletion time. And the concurrency worry does not apply — the deleted file contained no concurrency property at all (zero hits for `thread|concurrent|ttl|expiry`; "lock" is the `_plan_revising` boolean, mutated by one in-line writer). Real contention coverage (`test_stage_states_helpers.py::test_retry_on_conflict`, `test_sdlc_meta_set.py::test_write_revalidates_lease_before_write`) is untouched.

### Merge policy and scope: clean

Test-only (no non-`tests/` file in the diff). Zero added `skip`/`xfail`/`pytest.mark`. No `Co-Authored-By` or "Generated with Claude" lines. `merge-tree` reports 0 conflicts. `KNOWN_MISTAGS` is not a suppression habit: `tests/marker_map.py:442` computes `stale_exemptions = baseline_paths - violation_paths` and `test_feature_map_markers.py:58,66,89` fail on stale entries, untracked paths, missing prose, and reasons not marked `POLICY`. That allowlist can only shrink.

### Findings

**1. `tech_debt` — `tests/integration/test_connectivity_gaps.py:291,311`: the time bomb was rebuilt one notch further out.** Both nodes now assert a source-code string literal: `assert '"VALOR_SESSION_ID": session.session_id or ""' in source`. That is the *same* failure mode that broke them originally — a source-grep pointing at a module that moved — only the target moved. A `ruff format` reflow, a dict-key reorder, or renaming the local `session` breaks both nodes while behavior is untouched; conversely the assertion passes if `_harness_env` is built but never passed to the subprocess. The PR explicitly applied this reasoning to `test_real_pyproject_declarations_resolve` ("that would rebuild the same time bomb") but not here. A behavioral assertion on the env dict `_harness_env` actually produces removes the class permanently.

**2. `tech_debt` — `tests/integration/test_reflections_redis.py:183-193`: the deletion rationale was applied to one node and not its sibling three lines above.** `config/reflections.yaml` really did move to the vault: the file is absent from this worktree, `git check-ignore` reports `.gitignore:8:config/reflections.yaml`, it is copied machine-locally by `scripts/install_email_bridge.sh:150`, and the vault entry's `callable` has drifted as claimed. But the surviving `test_reflection_registered_in_yaml` unconditionally does `config_path = Path(...) / "config" / "reflections.yaml"` then `open(config_path)` — same gitignored path, no fixture creating it, `FileNotFoundError` in any clean checkout. It survived the nightly only because that machine was provisioned. This is node-list-driven triage where a root-cause grep sweep was needed.

**3. `tech_debt` — `tests/integration/test_reflections_redis.py:206-209`: a surviving sibling is green-but-vacuous.** `test_cleanup_callable_importable` asserts `from scripts.popoto_index_cleanup import run_cleanup` is callable. Per the vault, that is no longer the callable the `redis-index-cleanup` reflection schedules. It passes while proving the importability of a function nothing dispatches — exactly the residue this PR exists to clean up.

**4. `tech_debt` — the "16 passed" evidence predates four commits `main` has since gained**, including `4703bce23` (LLM task taxonomy, ~60 product files: `agent/health_check.py`, `agent/output_handler.py`, `agent/session_completion.py`, `bridge/*`, `config/settings.py`) and `4e504ef6b`, which dropped the `agent_session_queue` re-export that `test_remote_update.py` patches around. `merge-tree` is clean, but that proves textual safety only. Re-run the 16 nodes on the merge result, not the branch point, before merging.

**5. `nit` — `tests/integration/test_out_of_domain_reclaim.py:436`: positionally brittle counter.** `return [terminal_rec] if calls["n"] == 1 else [live_rec]` hard-codes "call #1 = owner-map snapshot, call #2 = Phase-2 read." A future third `filter(id=...)` for that session before Phase 2 silently shifts which record the reclaim decision sees. The new docstring documents the assumption, which is the right mitigation, but the positional coupling to product-code call order remains.

**6. `nit` — the commit message cites the wrong paths for the replacement coverage.** `TestG7PlanRevisingGuardDirect` and `TestG7ThroughDecideNextDispatch` are in `tests/unit/test_sdlc_router.py:91,217`, and `TestMetaSetWriteMeta` in `tests/unit/test_sdlc_meta_set.py:34` — all `tests/unit/`, not `tests/integration/`. The classes exist and cover the claims; only the paths are wrong. Worth correcting, since this commit message is the audit trail for a 210-line deletion.

### Third-disposition check

Independently ran `gh issue list --search` for `test_sdlc_pipeline_lock`, `test_out_of_domain_reclaim`, `tool_budget`, `test_reflections_redis`, and `nightly`. No open issue covers any of the 20 nodes. The PR's claim that "detail the open issue" had no target is correct.

valorengels:
## Review: CHANGES REQUESTED

**Mode:** independent roster (2 judges: `code-quality`, `risk`) — both reported, no quorum shortfall.
**Rule:** `any-blocker-wins`. `code-quality` raised 1 blocker; `risk` raised 0. Aggregate is **CHANGES REQUESTED**.
**Head SHA judged:** `62237be7793de8e9a823791f31c50076892662d9` (resolved via `tools/pr_head_resolver.py::resolve_pr_head_sha`).

**Totals:** 1 blocker, 6 tech debt, 4 nits.

### Independently reproduced, not credited from the PR body

- **`16 passed`** — re-ran the 16 surviving nodes on the PR head in a clean worktree: `16 passed, 3 warnings in 28.95s`. Confirmed.
- **No green-by-weakening markers** — `grep -E '^\+.*(skip|xfail|pytest\.mark)'` over the diff returns nothing. Confirmed.
- **`test_connectivity_gaps` mutation proof** — reproduced independently rather than accepted. Confirmed.
- **Test-only diff, no `Co-Authored-By`** — confirmed.
- **"No open issue covers these 20"** — re-ran the `gh issue list --search` sweep. Confirmed.

### The blocker

**`tests/integration/test_memory_mcp_server.py:197` — `test_fresh_shell_import_resolution` was loosened into a tautology.**

Reproduced directly:

```
$ env -i PATH=/usr/bin:/bin HOME=$HOME .venv/bin/python3 -m mcp_servers.memory_server --help
FastMCP server: memory
RC=0
```

With **no PYTHONPATH at all**, both assertions pass. Swapping bare `python3` for `sys.executable` pointed the subprocess at the repo venv, and that venv carries `_editable_impl_valor_bridge.pth` — precisely the "channel other than PYTHONPATH" the test's own docstring says it exists to rule out. None of the five stripped vars disables `.pth` processing.

This is the one disposition that failed the standard the PR set for itself: the node went from red-and-honest to green-and-toothless. It is also narrow — a self-contained fix in one test, no other file in the diff is affected.

Worth noting for whoever patches it: the production registration is `MCP_SERVER_COMMAND = "python3"` (`scripts/update/mcp_memory.py:39`), a PATH-resolved *name*. The worker plist puts `.venv/bin` first, so production's `python3` is also the venv interpreter and also resolves through the editable `.pth` — which means "PYTHONPATH is what makes the import work" may no longer be the production contract at all. The fix should decide what contract is actually worth guarding (that a registered MCP entry resolves in a fresh shell) and assert *that* with teeth, rather than restore the old bare-`python3` invocation, which had its own accuracy problem.

### Tech debt (6)

1. `tests/integration/test_tool_budget_enforcement.py:150` — `test_sdk_hook_disabled_allows` still carries the exact wrong-namespace bug (`AGENT_SESSION_ID` set to a `session_id`) that this PR fixed in its two neighbours; it does not actually cover the kill switch.
2. `tests/integration/test_connectivity_gaps.py:291,311` — the source-grep time bomb was rebuilt one notch out rather than removed; a `ruff format` reflow re-breaks both nodes while behavior is untouched.
3. `tests/integration/test_reflections_redis.py:183` — `test_reflection_registered_in_yaml` opens the same gitignored, vault-moved `config/reflections.yaml` and raises `FileNotFoundError` in any clean checkout. Same root cause as the node that was deleted; the sweep stopped at the node list.
4. `tests/integration/test_reflections_redis.py:206` — `test_cleanup_callable_importable` proves the importability of a callable the reflection no longer schedules.
5. `tests/marker_map.py:186` — the `KNOWN_MISTAGS` entry meets the guard's own remediation contract, but it grows a baseline whose comment says it can only shrink. That policy change deserves the guard owner's ratification, not a triage PR's.
6. The `16 passed` evidence was measured at the branch point; `main` has since gained four commits including `4703bce23` (~60 product files) and `4e504ef6b` (dropped the `agent_session_queue` re-export `test_remote_update.py` patches around). Re-run the node set on the merge result before merging.

### Nits (4)

- `tests/integration/test_connectivity_gaps.py:271` — docstring calls `agent/sdk_client.py` "a thin re-export shim"; it is 1531 lines.
- `tests/integration/test_reflections_redis.py:193` — the deleted `test_reflection_entry_structure` was reachable as a schema-level retarget rather than a deletion.
- `tests/integration/test_out_of_domain_reclaim.py:436` — the `calls["n"] == 1` counter positionally couples to product-code call order.
- The commit message cites `tests/integration/` for the three replacement-coverage classes; all three are in `tests/unit/`. Worth fixing, since that message is the audit trail for a 210-line deletion.

### What survived scrutiny

Both judges converged here independently, which is the point of the roster:

- **19 of 20 dispositions are correct**, and each root-cause diagnosis was checked against production code rather than read off the ledger.
- **All four deletions are justified.** The three `test_sdlc_pipeline_lock.py` nodes patched `tools.sdlc_meta_set.find_session`, a symbol absent from that module — they were *erroring* at `mock.patch.__enter__`, contributing zero coverage at deletion time. Replacement coverage was read assertion-for-assertion and is live. The file contained no concurrency property (zero hits for `thread|concurrent|ttl|expiry`), so the SDLC-lock concurrency surface is not what was deleted; real contention coverage is untouched. `test_reflection_entry_structure`'s drift is real and its only genuine code-under-test is covered elsewhere.
- **No flagged failure was a product regression being papered over.** On the three highest-cost nodes (session reclaim, tool-budget enforcement, steering delivery) the product is correct and the old tests were the broken party — in the reclaim and budget cases the refactor makes the test strictly *stronger*.

### Verdict rationale

`any-blocker-wins` forces CHANGES REQUESTED on one narrow, well-localized blocker. This is not a rejection of the triage — the body of work is sound and unusually well-argued. Fix the `test_memory_mcp_server` tautology, take the tech debt that is genuinely in scope (items 1-5 are all test-only, same surface), re-verify against current `main`, and this is ready.

<sub>Reviewed by an automated two-judge roster; per-judge findings are in the two preceding comments.</sub>

valorengels:
## Review: APPROVED

**Round 2**, after `/do-patch`. **Mode:** independent roster (2 judges: `code-quality`, `risk`) — both reported, no quorum shortfall. **Rule:** `any-blocker-wins`; neither judge raised a blocker.

**Head SHA judged:** `d65f6138f` (resolved via `tools/pr_head_resolver.py::resolve_pr_head_sha`). Branch rebased onto current `main` (`1ab8d429e`), so the verification below ran on the merge result, not the branch point.

### The round-1 blocker is resolved

`test_fresh_shell_import_resolution` no longer claims something it cannot prove. It now takes `scripts/update/mcp_memory.py::_expected_entry` verbatim — `command` as a PATH-resolved *name*, `args`, declared `env` — spawns it from a cwd outside the repo, and asserts it launches and advertises its tools. The env-stripping ceremony and the false "PYTHONPATH alone is what is proven" docstring are gone.

The `code-quality` judge reproduced the mutation independently (module rename → red, restore → green) and confirmed the residual question from round 1: dropping `"env": {"PYTHONPATH": ...}` from `_expected_entry` does still leave this node green — but that is the right division of labour, not a gap. The declaration contract is covered in unit tests (`tests/unit/test_update_mcp_memory.py:49`, `tests/unit/test_belt_resolver.py:255`); this integration node now covers launchability. The `risk` judge separately confirmed it is not a new time bomb of the original family: PATH is derived from `sys.executable` rather than hardcoded to `/opt/homebrew`, so it is self-consistent on any machine — the opposite property from the one that broke it.

### The patch went further than the findings asked

Two of the three follow-on commits were root-cause sweeps rather than node fixes, which is the discipline round 1 asked for:

- **A genuine coverage gain.** `test_sdk_hook_disabled_allows` carried the same wrong-namespace bug as its two siblings and passed identically with `TOOL_BUDGET_ENABLED = True` — it covered nothing. Proved three ways: fixed+disabled passes, fixed+enabled fails (`DENY: per-session tool-call budget reached (10/5)`), old form+enabled passes.
- **The connectivity-gaps time bomb is closed, not relocated.** Both nodes are now behavioral, driving `_execute_agent_session` with a spy over `agent.session_runner.SessionRunner`. The spy is the real seam — the executor's function-local import resolves the attribute at call time, and everything upstream of the terminal consumer is production code. The trade from source-coupling to mock-shape coupling is asymmetric in the right direction: the old grep had a *silent-green* mode (it passed whether or not `_harness_env` ever reached the subprocess); every drift mode of the new spy fails loudly.

### The new deletion was scrutinized hardest and stands

`e57c952f0` removes the rest of `TestRedisIndexCleanupReflection` — two nodes that were never in the 44-node nightly list. Scope expansion on a triage PR deserves an adversarial reading, and it got one. The decisive fact is not the deletion argument but the surviving guard, which the `risk` judge found: `scripts/update/reflections_callables.py:260::run_registry_probe` is a positive, fail-closed check that *every* registry callable imports, running against the real vault registry on every `/update` and gating the worker restart via `data/registry-probe-failed`. Strictly stronger than a test hardcoding one literal callable. `test_reflection_registered_in_yaml` was meanwhile a latent machine-dependent test — green on provisioned machines, `FileNotFoundError` in any clean checkout — which the node-list triage would have missed by construction.

### One review finding was measured and rejected

The `code-quality` judge's nit asked for `session.session_id = None` in the falsy case, so the `or ""` in `_harness_env` would be load-bearing. Tested it: `None` is unreachable there — `_execute_agent_session` short-circuits before constructing `SessionRunner`, and the node fails with `SessionRunner was never constructed`. `""` is the only falsy value production can carry at that point. Reverted, with the measurement recorded in the test so the next reviewer does not re-raise it (`d65f6138f`).

### Verification, reproduced here rather than credited

```
./scripts/pytest-clean.sh <16 nodes> -p no:randomly
16 passed, 4 warnings in 32.45s
```

Also independently run: the 24 nodes attributed to #3355 / #3388 / #3389 against current `main` — **44 passed** (parametrized expansion of the 24). Diff confirmed test-only; no `skip`, no `xfail`; the two added `@pytest.mark.asyncio` decorators are a collection requirement of the sync-to-async rewrite, not suppression. No `Co-Authored-By` in any commit. `ruff check` and `ruff format --check` clean.

### Carried forward, not blocking

- **`tests/marker_map.py` `KNOWN_MISTAGS`** — the entry is correct on its merits, but the baseline's own comment says it can only shrink. Whether growing it is policy is the guard owner's ratification to give. Flagged for the principal, unchanged here.
- **The MCP test's PATH assumption is docstring-only.** It constructs a PATH with `.venv/bin` first and proves `python3` launches under it; that the worker's launchd plist actually orders PATH that way is asserted in prose and tested nowhere. If that ordering regresses — and `/update` regenerates plists, and Python version bumps disturb them — this node stays green while production MCP registration breaks. Worth an issue, not a change to this PR.
- **`_dispatch_harness` duplicates `tests/unit/test_session_executor_runner_dispatch.py::_patch_worktree`**, as its own docstring admits. A new executor side effect now needs fixing in two places.
- **`test_valor_session_id_in_source` no longer reads any source.** The node ID was held fixed so the ledger stays accurate; worth renaming in a follow-up.

The PR body ledger has been updated to match the diff — on a triage PR the ledger is the deliverable, and it had drifted across the three patch commits.

<sub>Two-judge roster, round 2. Per-judge round-1 findings are in the earlier comments.</sub>

valorengels:
## DOCS stage: nothing owed (audited, not assumed)

The diff is test-only — `git diff --name-only origin/main...HEAD | grep -v '^tests/'` returns nothing. No product behavior, config key, env var, or feature surface changes, so no `docs/features/*.md` cascade is owed and no new feature doc is required.

Swept for stale references to the six deleted nodes across `docs/` and `tests/README.md`. The only hits are four files under `docs/archive/plans-completed/` — historical records of completed plans, which are deliberately not rewritten.

Recording DOCS as completed on that basis rather than manufacturing a docs commit, which would otherwise trigger a re-review for no reader benefit.

Issue #3164, opened 2026-09-05
Bridge release verify reports "unknown": macOS pgrep hides the bridge from its own descendants


valorengels:
## Symptom

Every `/update` (and every `python -m scripts.update.verify_release`) run **from inside an agent session hosted by the bridge** prints:

```
WARNING: bridge release could not be confirmed (unknown) — not failing the run
release verify OK @ 45d5d42d4 (bridge unknown, worker matches)
```

Observed on Valor the Captain at HEAD `45d5d42d4`, with a healthy bridge (pid 94804, started 09:51:14) and a fresh, correct beacon:

```
$ cat data/bridge_boot_sha
45d5d42d4
2026-09-05T02:51:23.636877+00:00
```

The same check run from the `remote-update.sh` cron job on the same machine at the same commit reports `bridge matches`. Nothing is wrong with the bridge.

## Root cause

`scripts/update/service.py::get_bridge_pid()` resolves the bridge with:

```python
result = run_cmd(["pgrep", "-f", "telegram_bridge.py"])
```

macOS `pgrep` (BSD) documents:

> `-a`  Include process ancestors in the match list. By default, the calling process and its ancestors are excluded.

Agent sessions run as `claude -p` **children of the bridge process**:

```
96594 zsh          <- this command
95415 claude -p    <- session
94804 .../Python /Users/valorengels/src/ai/bridge/telegram_bridge.py   <- the bridge
    1 /sbin/launchd
```

So for any process the bridge spawned, the bridge is an ancestor and `pgrep` silently omits it. Verified directly: `pgrep -f telegram_bridge.py` exits 1, while `pgrep -a -f telegram_bridge.py` returns 94804, and the process is present in `KERN_PROC_ALL` with ordinary flags (`0x4004` = `P_EXEC|P_LP64`, no `P_SYSTEM`).

`_classify_process()` returns `unknown` the moment `pid is None`, **before** it ever reads the beacon timestamp — so a false-`None` PID short-circuits classification even though a valid, fresh beacon is on disk. That is why the symptom is `unknown` and not `stale`.

This violates a stated acceptance criterion of #1898 (`docs/archive/plans-completed/update-verify-running-release.md`): "no false 'release could not be confirmed'".

## Blast radius

The same defect exists in every Python probe that resolves a long-lived service PID by `pgrep`, and it fires for any caller that is a descendant of the process it is looking for:

| Site | Consequence when the target is an ancestor |
|---|---|
[...]

Issue #2141, opened 2026-07-17
/update restarts the worker without draining active sessions — mid-turn work lost, orphaned harness, livelock risk for long turns


tomcounsell:
> **Context:** The fleet [update system](scripts/update/) (`/update`, `scripts/remote-update.sh`) restarts the standalone worker after syncing code and env vars. The worker ([docs/features/bridge-worker-architecture.md](docs/features/bridge-worker-architecture.md)) executes AgentSessions whose PM turns legitimately run 20+ minutes (a long build runs inside one turn). This bug was observed live: an update fired while an SDLC pipeline session was 19 minutes into a PM turn, and the turn's in-flight work was discarded.

## Problem

`restart_worker()` (`scripts/update/service.py:565-574`) invokes `valor-service.sh worker-restart` unconditionally — there is no check for running sessions, no drain window, and no defer-if-busy. A code comment (`service.py:409`) asserts the bootout/bootstrap cycle "is harmless (worker restarts cleanly)" — which is true for *crash-safety* but false for *work loss*.

**Current behavior (observed live, 2026-07-17 11:33 UTC, session `0_1784286820434`):**
1. Update sends SIGTERM mid-turn. The worker logs *"Received signal 15, shutting down gracefully… workers will finish current sessions and exit"* and *"Waiting for 1 active worker(s) to finish…"* — but the process exits ~3-5 seconds later with the 19-minute turn still in flight. The promised wait does not happen (launchd bootout's grace period wins).
2. The harness subprocess (`claude -p`, the PM turn) is left orphaned; the **next** worker boot's orphan-reap SIGKILLs it ("Drain: SIGKILL'd PID 41064 (escalation)").
3. Startup recovery flips the session running→pending; the worker re-picks it up and resumes the Claude transcript from the last turn boundary. Everything since `turn_start` — 19 minutes of PM/subagent work — is redone or lost; only durable artifacts (commits, PipelineLedger stage markers) survive. The resume log also shows `dev_agent_id=None`: resume scalars persist at turn boundaries, so a mid-turn kill loses the dev-agent continuation handle.
4. `logs/update.log` shows **three** "[update] Worker restarted" entries in a single day. Compounding risk: if updates recur faster than a long turn completes, that session can **livelock** — repeatedly killed mid-turn, repeatedly redoing the same work from the last durable checkpoint, never finishing.

**Desired outcome:**
[...]

Issue #1898, opened 2026-07-04
Update reports OK but bridge process keeps running pre-update code (Captain: disk @ 6b5b998a, process @ 659756a4)


tomcounsell:
**Plan:** https://github.com/tomcounsell/ai/blob/main/docs/plans/update-verify-running-release.md

## Recon Summary

_Added at plan time (2026-07-05, baseline `63e43118`) after tracing the update/restart code paths._

**Confirmed (grounded in code):**
- `scripts/update/run.py:88-100` — `UpdateConfig.cron()` sets `do_service_restart=False`. The Telegram `/update` and the 30-min polling cron both run cron mode, so **the bridge is never restarted by cron `/update`**. Only `UpdateConfig.full()` (`do_service_restart=True`, line 79) restarts services via `service.install_service`.
- `scripts/update/run.py:1555-1558` — in cron mode, when commits landed, the run only calls `git.set_restart_requested()` (writes `data/restart-requested`) and prints success. No process is cycled.
- The restart flag is consumed **only by the worker** — `agent/agent_session_queue.py:1780` and `:2203` call `_check_restart_flag()` / `_trigger_restart()` inside the worker session loop. **The bridge never checks the flag**; `bridge/telegram_bridge.py:2985` only *clears* a stale flag on startup. So a bridge running old code has no self-restart path under cron `/update`.
- `agent/agent_session_queue.py:1209-1246` — `_check_restart_flag()` returns True only when the flag is < 1h old (`_RESTART_FLAG_TTL`) AND `AgentSession.query.filter(status="running")` is empty. A wedged/long-running session (exactly the Cuttlefish wedge this update was meant to fix) starves it; once >1h passes the flag self-deletes silently and no restart ever fires.
- `scripts/update/service.py:71-105` — `get_service_status()` returns only `running`/`pid`/`uptime`; it does **not** read the running process release. The update success gate (`run.py:1339`) checks only `running`, so it reports OK regardless of which code the process is running.
- The running process release is `git rev-parse HEAD` captured at process init (`monitoring/sentry_config.py:61`), which is why Sentry showed `release=659756a4` — the SHA HEAD was at when the bridge last started; the bridge simply never restarted since.

**Root cause:** cron `/update` decouples "update OK" (git pull + flag-set) from whether either process actually cycled onto new code. The bridge has no cron-mode restart path at all; the worker's deferred restart can be starved by wedged sessions and then silently expires. No post-restart release verification exists.

[...]

PR #3529, opened 2026-09-20
Share the release classifier between the cron restart gates and the release verify (#3528)

### LINKED ISSUE #3528: Cron restart gate and release verifier use different diff bases: a deferred restart wedges the machine into a permanent release-verify FAILED loop
> **Context for a reader new to this repo.** Every machine in this fleet runs `scripts/remote-update.sh` from a launchd cron job every 30 minutes. That script pulls `main`, then decides whether to restart the two long-lived services — the **worker** (`python -m worker`, the sole agent-session execution engine) and the **bridge** (`bridge/telegram_bridge.py`, Telegram I/O). Restarts are deliberately *gated*: issue #1091 established that the worker must only be restarted when the newly pulled commits actually touch code the worker loads, because an unconditional restart kills in-flight agent sessions every 30 minutes.
>
> Issue #1898 later added a **terminal release verify** — `python -m scripts.update.verify_release`, which reads a **boot-SHA beacon** (a two-line file at `data/{worker,bridge}_boot_sha` holding the commit SHA a process booted on plus a timestamp) and reports `release verify FAILED` when a running process is behind `HEAD` on its own relevant paths. The verify is the machine's safety net; the gate is the machine's actuator.

## Problem

The restart gate and the release verifier answer the same question — "is this process running stale code?" — against **two different diff bases**. They agree on the *path sets* but not on the *range*, so the safety net can report a fault the actuator is structurally incapable of acting on. When that happens the machine wedges: it reports `FAILED` every 30 minutes forever and never self-heals.

**Current behavior.** Both gates in `scripts/remote-update.sh` compute relevance from **this cycle's pull delta only**:

```sh
NEED_RESTART=false
if [ "$BEFORE_SHA" != "$AFTER_SHA" ]; then
    if git -C "$PROJECT_DIR" diff "$BEFORE_SHA" "$AFTER_SHA" -- \
        worker/ agent/ mcp_servers/ models/ tools/ bridge/ reflections/ \
        pyproject.toml | grep -q . ; then
        NEED_RESTART=true
    fi
fi
```

`scripts/update/service.py::_classify_process` — which the verifier calls — instead diffs **the running process's own boot SHA against HEAD**:

```python
log_result = run_cmd(
    ["git", "log", "--oneline", f"{result['boot_sha']}..{head_sha}", "--", *relevant_paths],
    cwd=project_dir,
)
```

A restart can legitimately be skipped on the cycle that pulls the relevant commits — the drain window times out on a long-running session (`[update] Worker restart DEFERRED`), the worker declines to SIGKILL its own ancestor (#3164), or a `kickstart` fails. The commits are now on disk but not in the running process. On **every subsequent cycle** `git pull` reports "Already up to date", so `BEFORE_SHA == AFTER_SHA`, both gates short-circuit to false, and the script prints `No worker-relevant changes detected — skipping restart`. The verifier, reading `boot_sha..HEAD`, correctly sees the stale process and exits 1. Nothing ever closes the loop.

Observed live on machine **Valor the Captain**, 14 consecutive cron cycles:

```
[update] No worker-relevant changes detected — skipping restart
release verify FAILED @ 4cc3f925f: bridge running 4e504ef6b but HEAD is 4cc3f925f; worker running 4e504ef6b but HEAD is 4cc3f925f
```

The staleness is real, not a verifier false positive — `4e504ef6b..4cc3f925f` touches 57 files across the worker path set (including `worker/__main__.py` and `tools/`) and 53 across the bridge path set.

The comment introducing the shared path lists (`scripts/update/service.py:1130`) asserts the invariant that is in fact violated:

```
# Per-process relevant path sets — IDENTICAL to the restart gates' diff sets in
# scripts/remote-update.sh (#1091 relevant-diff design). Classifier and restart
# gate must agree by construction ...
```

They agree on paths and disagree on range, which is the half of the invariant that matters for self-healing.

The manual `python scripts/update/run.py --full` path is unaffected: it has `_self_heal_stale_worker` (`scripts/update/run.py:797`), which restarts a worker classified stale. The cron path — the one that runs unattended every 30 minutes on every machine — has no equivalent.

**Desired outcome.** A cron cycle that finds a running process stale on its own relevant paths restarts it, whether or not that cycle's pull moved HEAD. The gate and the verifier must reach their verdict through the same code so they cannot drift apart again.

## Definitions

| Term | Meaning |
|------|---------|
| **Restart gate** | The `NEED_RESTART` / `NEED_BRIDGE_RESTART` shell conditionals in [`scripts/remote-update.sh`](../blob/main/scripts/remote-update.sh) that decide whether to `launchctl kickstart` a service this cycle. |
| **Boot-SHA beacon** | Two-line file (`data/worker_boot_sha`, `data/bridge_boot_sha`) written by `monitoring.boot_beacon` at process start: the commit SHA the process booted on, plus an ISO timestamp. |
| **Relevant paths** | Per-process path list (`WORKER_RELEVANT_PATHS`, `BRIDGE_RELEVANT_PATHS` in [`scripts/update/service.py`](../blob/main/scripts/update/service.py)) defining which directories a commit must touch to matter to that process. Docs-only commits deliberately do not. |
| **Positive staleness** | The #1898 classification rule: a process is `stale` only when commits touching *its own* paths landed after its boot SHA. Raw `boot_sha != HEAD` is never used, because it would flag healthy processes after a docs commit. |
| **Drain** | `scripts/update/drain.py` — polls until no agent sessions are running before a worker restart, so a 20-minute PM turn is not killed mid-flight. On timeout the restart is deferred to the next cycle. |
| **`_classify_process`** | The single-process classifier in `scripts/update/service.py` returning `matches` / `stale` / `unknown`. |

## Solution sketch

Give the restart gates the classifier's diff base by making them *call the classifier*.

Add a small entrypoint — e.g. `python -m scripts.update.restart_gate --process worker|bridge` — that reuses `_classify_process` and the existing `*_RELEVANT_PATHS` lists, and exits 0 ("restart needed") when the running process classifies `stale`. When the classification is `unknown` (no beacon, no running process, orphaned beacon, unresolvable boot SHA — i.e. a fresh install or a first boot), it falls back to today's `BEFORE_SHA..AFTER_SHA` diff so existing behavior is preserved where there is nothing to classify. Both shell gates then call it instead of hand-rolling their own `git diff`.

Everything else about the restart sequence stays as-is: the drain (#2141), the registry probe (#2875), the ordering constraints around the bridge kickstart and the terminal verify (#1898), and the plist-existence check for the bridge.

The stale comment at `service.py:1130` should be corrected to describe the new arrangement (shared code, not parallel implementations), along with any docs describing the gate.

Open question for the planner: should an `unknown` classification on a machine with a *running* process but a missing beacon fall back to the pull delta (proposed above, conservative) or be treated as stale (aggressive self-heal, risks a restart loop if beacon writes are broken)? The #1898 posture is that `unknown` never triggers a restart, which argues for the conservative choice.

## Acceptance criteria

- [ ] A cron cycle with `BEFORE_SHA == AFTER_SHA` and a process classified `stale` restarts that process.
- [ ] A cron cycle with `BEFORE_SHA == AFTER_SHA` and a process classified `matches` still prints the no-changes line and restarts nothing (no restart loop, no session kills).
- [ ] A machine with no beacon / no running process behaves exactly as it does today (pull-delta fallback), so a fresh install still installs and starts services normally.
- [ ] The gate and `verify_release` share the `*_RELEVANT_PATHS` lists and the classification code — a future edit to one cannot silently change the other.
- [ ] Drain, registry-probe, bridge-plist, and kickstart-ordering behavior are unchanged.
- [ ] `scripts/update/service.py:1130`'s comment and any affected docs describe the shared-classifier arrangement.
- [ ] Narrow tests cover the wedge scenario (no-op cycle + stale beacon → restart) and the fallback path, following the existing `tests/unit/test_update_release_verify.py` / `tests/unit/test_remote_update_shell.py` fixtures.

## Downstream context

valorengels:
Closes #3528

## Problem

`scripts/remote-update.sh`'s two restart gates and the terminal release verify answer the same question — *is this process running stale code?* — against **two different diff bases**. They share path sets but not ranges:

| | Diff base |
|---|---|
| `NEED_RESTART` / `NEED_BRIDGE_RESTART` (the actuator) | `BEFORE_SHA..AFTER_SHA` — this cycle's pull delta |
| `service._classify_process` (the safety net) | `boot_sha..HEAD` — the running process's own release |

Whenever a restart was skipped on the cycle that pulled the relevant commits — drain timeout, the #3164 self-ancestor refusal, a failed kickstart — the commits sit on disk but not in the process. Every later cron cycle pulls nothing, so `BEFORE_SHA == AFTER_SHA`, both gates short-circuit to false, and the machine wedges: `FAILED` every 30 minutes with no actuator able to clear it.

Observed live on **Valor the Captain**, 14 consecutive cycles:

```
[update] No worker-relevant changes detected — skipping restart
release verify FAILED @ 4cc3f925f: bridge running 4e504ef6b but HEAD is 4cc3f925f; worker running 4e504ef6b but HEAD is 4cc3f925f
```

Not a verifier false positive: `4e504ef6b..4cc3f925f` touches 57 files in the worker path set (incl. `worker/__main__.py`) and 53 in the bridge path set. The `--full` manual path has `_self_heal_stale_worker`; the unattended cron path had no equivalent.

The comment at `service.py:1130` asserted the invariant that was being violated — *"Classifier and restart gate must agree by construction"* — while the gates hand-rolled their own diff.

## Fix

New `scripts/update/restart_gate.py`:

```
python -m scripts.update.restart_gate --process worker|bridge --before <sha> --after <sha>
# exit 0 = restart, 1 = skip
```

It calls `service.classify_process` over the existing `PROCESS_RELEVANT_PATHS` — literally the same code and the same lists the verifier uses — so gate and verify cannot reach opposite verdicts about the same process. Both shell gates now delegate to it instead of running their own `git diff`.

**Fallback.** When the classification is `unknown` (no beacon, process not running, orphaned beacon, unresolvable boot SHA — i.e. a fresh install), the gate falls back to the `BEFORE_SHA..AFTER_SHA` pull delta, preserving today's behavior where there is nothing to classify. Per the #1898 posture that `unknown` never escalates, an unknown release never *invents* a restart on a no-op cycle.

Verified read-only against the live wedged checkout:

```
[restart-gate] RESTART: worker running 4e504ef6b is behind HEAD 4cc3f925f on worker-relevant paths
[restart-gate] RESTART: bridge running 4e504ef6b is behind HEAD 4cc3f925f on bridge-relevant paths
```

## Deliberately unchanged

Drain (#2141), registry probe (#2875), bridge-plist gating, kickstart ordering / lock release / pending-report staging (#1898), and the `--skip-bridge` / `--since` verify scoping.

## Tests

`tests/unit/test_update_restart_gate.py` (new, 14 cases): the wedge regression for both processes, `matches` on a no-op cycle restarting nothing (no restart loop), `#1091` docs-only relevance surviving the new base, per-process path scoping, every `unknown` fallback path, and two structural guards — the gate and verifier share the path-list objects, and `remote-update.sh` can't grow an inline relevance diff again.

`tests/unit/test_remote_update_shell.py` (+5 cases): both gates are invoked per process with the pull delta, a no-op cycle restarts a stale worker/bridge, a no-op cycle with current processes restarts nothing, and the bridge plist check still short-circuits ahead of the gate. The stub's gate arm reproduces the real gate's `unknown` fallback, which is what the real gate does in that sandbox, so the pre-existing shell tests keep their original meaning.

```
83 passed
```
(`test_update_restart_gate.py` + `test_remote_update_shell.py` + `test_update_release_verify.py`, via `scripts/pytest-clean.sh`)

## Docs

`docs/features/bridge-worker-architecture.md` (worker gate condition 1) and `docs/features/bridge-self-healing.md` (bridge kickstart, worker self-heal contrast) rewritten to describe the shared-classifier arrangement; the stale `service.py:1130` comment now describes what the code actually does and records why the old base was wrong.

## Remediation note

The fix is self-applying on the affected machine: the 30-minute cron runs the freshly pulled `remote-update.sh`, so the first tick after this merges restarts the stale worker (behind the drain) and the stale bridge.


valorengels:

**Verdict: APPROVED** — no blockers. Head SHA `7282b9b9`. Ruff clean; 40 tests in the two affected files pass.

Verified: no unbound-local in `decide()` (the two removed assignments were genuinely dead — `git.get_short_sha` uses `check=True`, so a git failure raises, is caught, and falls to the pull-delta branch which touches neither variable); both `except Exception` blocks correctly scoped with accurate comments; the `_classify_process` → `classify_process` rename fully propagated (only surviving reference is in `docs/archive/`); registries single-source in code.

### Blockers
- None

### Tech Debt
- **TD-1 — `GATE ERROR:` is invisible to the only channel a human reads.** `scripts/update/restart_gate.py:133`. `bridge/update.py:34` defines `_LEGACY_WARNING_PREFIXES = ("[update] WARN", "WARNING:", "ERROR", "RESTART FAILED")` matched with `startswith`. `"GATE ERROR: ..."` starts with `GATE`, so it matches nothing. The sibling `verify_release.py:107,147` uses `WARNING: `, which *is* in the tuple. Failure: on a Telegram `/update`, the gate crashes, prints `GATE ERROR`, exits 1; `bridge/update.py:462` finds no warnings → no fix session → report reads `✅ update OK` while the only visible line is `No worker-relevant changes detected`. That is exactly the silent-wrong-verdict shape the commit claims to prevent.
- **TD-2 — the shell test stub keeps a second copy of both relevant-path lists.** `tests/unit/test_remote_update_shell.py:133,136`. Verbatim duplicates of `WORKER_RELEVANT_PATHS`/`BRIDGE_RELEVANT_PATHS` — the exact data `PROCESS_RELEVANT_PATHS` exists to single-source. Failure: add `config/` to the worker list and the stub still models the old set, so the shell tests keep passing while asserting the wrong relevance contract.

### Nits
- **N-3** `tests/unit/test_update_restart_gate.py:251` — docstring over-claims ("no second copy can exist"); the test only asserts `is`-identity.
- **N-4** `:269` — the no-inline-diff guard is pinned to one exact quoting; `git diff "$BEFORE_SHA".."$AFTER_SHA"` would slip through.
- **N-5** `:257-260` — near-tautological (asserts a lambda late-binds a module global).
- **N-6** `docs/features/bridge-self-healing.md:589` — unchanged sentence still says the gates *diff* those sets; they now classify over them.
- **N-7** — the #3528 backstory is retold at paragraph length in five places.
- **N-8** — nothing exercises the `python -m` invocation as a subprocess (low value; `drain`/`verify_release` establish the form works).

valorengels:
## Review (Judge risk):

**Verdict: APPROVED** — no blockers. Head SHA `7282b9b9`. Read the full diff plus complete `remote-update.sh`, `service.py`, `restart_gate.py`, `verify_release.py`, `drain.py`, `monitoring/boot_beacon.py`, `bridge/update.py`. 40 tests pass.

**Restart-loop analysis (the matrix).** No reachable loop. `matches` → skip. `stale` → restart, and the restart self-clears: the new image runs `write_boot_beacon` (`worker/__main__.py:1166`, first statement of `main()`; `bridge/telegram_bridge.py:3298`) writing `HEAD` of the checkout the shell just pulled, so the next cycle is `matches`. Beacon-write failure / crash-loop → new `process_start_ts` exceeds old `beacon_ts` → orphaned-beacon `unknown` (`service.py:1298`) → pull-delta → `before == after` → skip. `ps -o lstart` second truncation biases `start_ts` *earlier*, toward trusting the beacon, never toward spurious `stale`. History rewrite → `git log` rc != 0 → `unknown` → skip. Also checked whether `git log A..B -- <paths>` history simplification could under-report vs. the old `git diff` (which would silently preserve the wedge): 400 recent bases and 60 merge commits scanned, zero divergence; a constructed evil merge is still reported.

**Wedge closure confirmed live.** On this machine both processes classify `stale` (worker pid 92709 boot `4e504ef6b`, bridge pid 92734 boot `4e504ef6b`, HEAD `4cc3f925f`) — exactly the state the old pull-delta gate called "no relevant changes". The new gate returns restart for both.

**Shell wiring intact.** `$PYTHON` validated executable at :151 before both gates; `cd "$PROJECT_DIR"` at :49 holds; both call sites are `if`-conditions (exempt from `set -e`); argparse exit 2 and any other non-zero treated as skip (test-covered). Nothing previously gating a restart was dropped — drain (#2141), `REGISTRY_PROBE_OK` (#2875), `WORKER_ALIVE`, the label-absent recovery branch, `VERIFY_SINCE`, `--since`/`--skip-bridge`, lock release, pending-report staging. Path lists match the old inline ones byte-for-byte. `com.valor.update.plist` has `StartInterval 1800` and no `KeepAlive`, so a non-zero exit cannot cause a tight relaunch loop.

**Bridge-gate behavior change** safe on every path traced: the cron path is launchd-parented (not in the bridge's process group); the Telegram path already had this shape for pull cycles and the staged `data/update-pending-report` + fresh-bridge boot flush is the reporter.

### Blockers
- None

### Tech Debt
- **TD-1 — drain fail-open is now exercised every cycle.** `scripts/update/drain.py:95`. `main()` returns `EXIT_IDLE` on any ORM/Redis exception. Pre-#3528 the gate only reached the drain on a cycle that pulled worker-relevant code; a `stale` worker now re-enters it every 30 min. Failure: busy worker + stale boot SHA + one Redis MISCONF poll → `count_running_sessions()` raises → exit 0 → `kickstart -k` SIGKILLs an in-flight 20-minute PM turn. Consider failing *closed* now that deferral is no longer permanent.
- **TD-2 — same amplification, cost side.** `drain.py:66`. A worker that never drains now runs the full 300s poll every cycle, holding `data/update.lock` for 5 of every 30 minutes, green-skipping concurrent Telegram `/update`s.
- **TD-3 — gate/verify eligibility still disagrees for the worker.** `scripts/update/service.py:1343`. `verify_running_release` classifies the worker on `machine_check["projects"]` alone, while the shell can only restart it inside `[ -f "$WORKER_PLIST" ] && [ -f "$WORKER_DST" ]`. Failure: worker-role machine whose `~/Library/LaunchAgents/com.valor.worker.plist` was removed → `release verify FAILED` every 30 min with no actuator — the #3528 shape again. The bridge has the Decision-23 plist carve-out (`BRIDGE_PLIST_PATH.exists()`, :1336); the worker has no equivalent.
- **TD-4 — `GATE ERROR:` is both unreachable and unconsumed.** `restart_gate.py:133`. `decide()` already catches everything internally (:63, :88), so `main`'s handler can essentially only fire on `BrokenPipeError`. Even if it fired, `bridge/update.py:107` matches `startswith("ERROR")`, which `"GATE ERROR"` does not. (Converges with the code-quality judge's TD-1.)
- **TD-5 — the beacon-ownership heuristic can be satisfied by a different process than the PID measured.** `service.py:1298`. This PR upgrades that heuristic's consequence from "print FAILED" to "SIGKILL the worker". Observed live: `get_worker_pid()` → 92709 started `Sat Sep 19 21:30:44` local (14:30:44Z), while `data/worker_boot_sha` was written `2026-09-19T20:37:28Z` — six hours later, by a short-lived second `python -m worker`. If such a transient writes a SHA older than HEAD, the long-lived worker is restarted on a verdict that was never about it.

### Nits
- **N-1** `scripts/remote-update.sh:407` — the worker kickstart is not guarded by `service_pid_refuse_self_kill`, though the script sources `scripts/lib/service_pids.sh` which defines it and `valor-service.sh:772` uses it. Practically unreachable today, but the gate now fires every cycle rather than only on pull cycles.
- **N-2** `restart_gate.py:81` — `git.get_short_sha` does not check `returncode`, so `head_sha` can be `""`, rendering `worker running abc1234 is behind HEAD ` with a blank SHA.

valorengels:
## Review:

**Verdict: APPROVED** — 0 blockers, 0 open tech debt, 0 open nits.
**Mode:** independent roster (2 judges) — `code-quality`, `risk`.
**Consensus:** `any-blocker-wins` via `agent.sdlc_review_consensus.compute_consensus` — `n=2`, `expected_n=2`, `quorum_shortfall: false`, `tied: false`.
**Head reviewed:** `c0839ad8` (resolved via `tools/pr_head_resolver.py::resolve_pr_head_sha`).

Both judges returned APPROVED with no blockers at `7282b9b9`. Their findings were then either fixed in `c0839ad8` or filed as follow-ups, so this verdict is against a zero-open-finding diff.

### Blockers
- None

### Tech Debt
- None open. Disposition of the seven raised:

| Finding | Judges | Disposition |
|---|---|---|
| `GATE ERROR:` / `[restart-gate]` diagnostics never reach `extract_update_warnings` | both (CQ TD-1, risk TD-4) | **Fixed** in `c0839ad8` |
| Shell test stub retypes both relevant-path lists | code-quality TD-2 | **Fixed** — stub now interpolates from `PROCESS_RELEVANT_PATHS` |
| Worker plist eligibility mismatch between gate and verify | risk TD-3 | Filed #3533 |
| Beacon-ownership heuristic can misattribute a transient's beacon | risk TD-5 | Filed #3534 |
| Drain fail-open amplified to every cycle (+ lock-hold cost) | risk TD-1, TD-2 | Filed #3535 |

The three filed issues are pre-existing behavior this PR amplifies rather than introduces; fixing them here would have meant changing the classifier's trust model and the drain's failure polarity inside a bug fix.

### Nits
- None open. `N-3` (over-claiming docstring), `N-4` (inline-diff guard pinned to one quoting), `N-6` (stale `bridge-self-healing.md` sentence) and risk `N-2` (blank SHA on unreadable HEAD) are fixed in `c0839ad8`. `N-5`, `N-7`, `N-8` and risk `N-1` were judged not worth acting on and are recorded in the per-judge comments above rather than silently dropped.

### Verification run at `c0839ad8`
- `python -m ruff check .` — passed
- `python -m ruff format --check .` — 1681 files already formatted
- `scripts/pytest-clean.sh tests/unit/test_update_restart_gate.py tests/unit/test_remote_update_shell.py tests/unit/test_update_release_verify.py tests/unit/test_bridge_update.py tests/unit/test_update_warning_extraction.py` — **140 passed**
- Live read-only smoke against the wedged checkout: `[restart-gate] RESTART: worker running 4e504ef6b is behind HEAD 4cc3f925f on worker-relevant paths`

No plan document — off-pipeline fix; PLAN and CRITIQUE recorded `skipped`. No UI files in the diff, so the visual-proof gate is a no-op.

